PR #1, opened 2014-01-16
Removing warnings

ScrimpyCat:
Removed two of the warnings surrounding integer precision loss.

I was unsure what the intended behaviour was for the sectionIndex value so I just made the best assumption.


bgerstle:
I think upstream changes have made this obsolete. @ScrimpyCat, please rebase if you feel otherwise.

PR #2, opened 2014-01-16
Non regex getWikiImageFileNameWithoutSizePrefix

ScrimpyCat:
Changed -getWikiImageFileNameWithoutSizePrefix to no longer use
regex for removing the pixel count prefix. This will behave
the same except for input that only contains the pixel count
and no following characters. However that would not be an
appropriate file name so shouldn't matter.


bgerstle:
@ScrimpyCat sorry this wasn't merged, it's a good patch! If you're up for it, please rebase and add some unit tests and we'll take another look.

PR #3, opened 2014-07-16
A first implementation of a MKMapView to display coordinates included in a Wikipedia article

if0xx:
At this stage the implementation catches taps onto geohack extension links, launches a MapViewController and passes the geohack URL on to the MVC.

The MapViewController parses the URL, gathers coordinates and title and displays a native iOS map with an annotation (“pin”) representing the article in question. The map style can be toggled by tapping the gear. Tapping the “Share”-Button opens the geohack page in MobileSafari.

Next steps: use geo api JSON call to retrieve coordinates already while loading article, add map icon to bottom toolbar in main article view if coordinates associated with article can be retrieved.

By Ulf Buermeyer, @if0xx.


Wikipedia.xcodeproj/project.pbxproj
```diff
@@ -1602,7 +1623,7 @@
 				ORGANIZATIONNAME = "Wikimedia Foundation";
 				TargetAttributes = {
 					D4991434181D51DE00E6073C = {
-						DevelopmentTeam = D5RR3ZF62D;
+						DevelopmentTeam = SLUNTGN77L;
```
ollieatkinson: You can include them in a `.xcconfig` file and avoid holding any crucial data in the project, it means it can be versioned controlled too.

See:
> http://jamesdempsey.net/2015/01/31/generating-xcode-build-configuration-files-with-buildsettingextractor-xcodeproj-to-xcconfig/

bgerstle:
Sorry to all involved that this has fallen by the wayside, but I think it's best to close the PR until we resolve the map provider ambiguities.  I'm going to send another email to the [mobile-l list](https://lists.wikimedia.org/mailman/listinfo/mobile-l) and restart the conversation.

PR #4, opened 2014-07-25
MWPageTitle: added property allCapsText

if0xx:
added a new readonly property 'allCapsText' to MWPageTitle that is guaranteed to return an all-caps title string using [NSString capitalizedStringWithLocale:]


bgerstle:
Sorry this is now out of date. @if0xx if you're still interested in this change, please rebase and we'll take another look.

PR #9, opened 2015-04-02
Various project & fastlane tweaks

btgerst:
https://phabricator.wikimedia.org/T94769

Follow-up on some clean-up for Fastfile & Xcode project:

- Remove extra xcconfig & schemes
  - Moved xccconfig settings into build configurations
  - Removed schemes that only set the build configuration (schemes for specific run configs can be created later)
- Update Gemfile
  - Includes `fastlane` as gem
- Remove unneeded setting from Podfile (specifying targets which don't exists)
- Remove duplicate settings from fastlane config files
- Add options to fastlane "configs" which prevent git reset & deployment
  - use `NO_DEPLOY=1` to prevent deployment
  - use `NO_RESET=1` to prevent `git reset`


btgerst:
I give up...

PR #13, opened 2015-05-12
Some clean ups of the Fastfile

KrauseFx:
Also added `fastlane_version` to make sure a certain fastlane version is being used. It's a good thing to have, even if you use `bundle exec`


bgerstle:
merged in b1f3ba4!

PR #14, opened 2015-05-12
Cleaned up the Deliverfile

KrauseFx:
Feel free to merge if you don't need those comments anyway :+1:

### DIFF
--- a/fastlane/Deliverfile
+++ b/fastlane/Deliverfile
@@ -1,3 +1,4 @@
+# vi: ft=ruby
 
 ###################### Available Options ######################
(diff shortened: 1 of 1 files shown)


fastlane/Deliverfile
```diff
@@ -1,21 +1,5 @@
-# vi: ft=ruby
```
bgerstle: this should stay in, since it sets the filetype to Ruby when opened in vim.

KrauseFx:
Sorry, re-added it

bgerstle:
merged in 721f609!

PR #17, opened 2015-07-27
Native article requests only 1st section html and all section titles

montehurd:
When web view is shown full article is requested ONLY if non lead sections are found to be missing (as they will be if only the first section query has happened). If all sections are found to be cached, the cached article is shown w/o making additional network requests.

Known issues:
- gallery seems to not work as it assumes full article has been fetched
- occasional data layer crashes - cause unknown - to reproduce enter various search terms and select a few different articles.

Bug: T105542

Change-Id: I440517cc2546bececb85a6dff0b165a7507ff4b9


coreyfloyd:
Posted some updates… still not quite ready.

Doing some work to make this fit in with https://phabricator.wikimedia.org/T106150

Need to add the new model object for the "article preview" and then check on testing. I'm sure they will be broken any second now…

coreyfloyd:
@bgerstle so apparently you can't test anything with promise kit using tests in Obj-C. So need to re-write article fetcher tests in Swift…

coreyfloyd:
@bgerstle taking a break on these tests - spent a lot of time on trying to get these working last night. 

I basically have them re-written in Obj-C, but have to get them converted to Swift. Filed a separate ticket to get them re-written again.

bgerstle:
> so apparently you can't test anything with promise kit using tests in Obj-C

How so?  PromiseKit has [AnyPromise tests](https://github.com/mxcl/PromiseKit/blob/master/Tests/CorePromise/AnyPromise.test.m).  The promise utilities I wrote for testing are in Swift, but I could write an ObjC wrapper for it as well.

bgerstle:
@coreyfloyd i'll be around tonight if you get to this, ping me if you questions about the tests

coreyfloyd:
@bgerstle a .06% increase? what what?

coreyfloyd:
@bgerstle this is ready to go now I think.

Wikipedia/UI-V5/WMFArticleFetcher.m
```diff
@@ -20,37 +188,45 @@ @implementation WMFArticleFetcher
...
     }
     return self;
 }
 
-- (ArticleFetcher*)fetcher {
-    if (!_fetcher) {
-        _fetcher = [[ArticleFetcher alloc] init];
-    }
+- (id)serializedArticleWithTitle:(MWKTitle*)title response:(NSDictionary*)response {
+    MWKArticle* article = [self.dataStore articleWithTitle:title];
```
bgerstle: why are we reading the article from the data store? this seems to go against what the intent is here: to serialize `response` to the data store as the new state. i.e. we should only be writing in the fetcher, not reading (and even then, we should probably take the data store out of the fetcher entirely, and leave that concern to a higher-level controller object)

coreyfloyd: I didn't change the serialization logic, just moved it. 

Having said that… I think you are right that we do want to totally overwrite any existing article. However, since the article is not immutable, there may be logic that makes changes to the article elsewhere in the app. I am not sure that we can say it is safe to overwrite what is in the store until we make it immutable. 

It also may be "legacy" logic from when it was core data and you would always do a fetch before writing to make sure you don't have duplicates in the database.

bgerstle: we can leave this as-is for now and make refactoring this a separate task, which IMO we should tackle sooner than later. the mixing of responsibilities and the way articles are written to disk makes this difficult to reason about and test.

PR #19, opened 2015-07-28
move build/cov badges & add "prebuild" to makefile

bgerstle:
I mentioned a `prebuild` goal in the `Makefile` when updating the README, but forgot to add it :disappointed:.  This adds the goal, which is intended to install dependencies needed to build the project—only submodules for now.  

I also moved badges to the next line in the README, because it looks better :art:.


coreyfloyd:
Comcastic! :+1:

PR #20, opened 2015-07-28
Removed some old lead image cruft.

montehurd:
- Removed some old lead image cruft. 

- Also fixed small bug related to collapsed tables overlaying the edit pencil.

Note: I had intended to remove the "moveFirstGoodParagraphUp" javascript transform because I reasoned "now that we're collapsing tables it's less necessary to move the first paragraph above the fold". 

This turned out to be incorrect because it doesn't take into account *images*. Many articles have one (or more) images above the first paragraph of text and this can mean one or more screens of images before you encounter any text. (On desktop these images float so it's not like editors intend them to be all you see above the fold on small screens.) 

Since we're widening images, the issue is even more pronounced. The "enwiki > food" article is a good example.

Because of this I recommend we leave the first paragraph transform in place for now as removing it creates more issues than it solves.

Bug: T106168


bgerstle:
@montehurd quick comment: would you please put something human readable in the PR headline? The phab ticket can be linked in the PR description, just like we used to do in commit messages on Gerrit.

I'll review this tomorrow morning :city_sunset:

montehurd:
Oops sorry about that!

montehurd:
Wow... Am I missing something or is the travis log a bit... verbose. I can't make any sense of it.

bgerstle:
@montehurd unfortunately it's _very_ verbose.  in many cases it's _so_ verbose we need to view the [raw log](https://s3.amazonaws.com/archive.travis-ci.org/jobs/72938318/log.txt). one tip for grokking it is to scroll to the bottom to see what went wrong. you can see in this case that there was a test failure:

``` shell
Failures:



  0) -[WMFLegacyImageDataMigrationTests testUnmigratedEntriesCorrectlyReturnsUnmigratedEntry] (WikipediaUnitTests.xctest)
------------------------------------------------------------



/Users/travis/build/wikimedia/wikipedia-ios/WikipediaUnitTests/Utilities/XCTestCase+SwiftDefaults.swift:0: Asynchronous wait failed: Exceeded timeout of 0.1 seconds, with unfulfilled expectations: "testUnmigratedEntriesCorrectlyReturnsUnmigratedEntry():L56".:
```

Specifically: 

> **Asynchronous wait failed: Exceeded timeout of 0.1 seconds, with unfulfilled expectations: "testUnmigratedEntriesCorrectlyReturnsUnmigratedEntry():L56"**

Which is a combination of `XCTestExpectation` and `PromiseKit`.  It's saying that an expectation we created in a test wasn't fulfilled within its timeout of 0.1 seconds.  This is something I've addressed elsewhere, but it slipped through the cracks here. Travis VMs are slow, so we need to raise timeouts in order to give asynchronous tests ample time to complete, otherwise they'll occasionally fail.  

I've raised the default timeout to 1 second, since in most cases we're just dispatching to some background queue to call a quick method.  In some cases though, we're doing file I/O, which can take a while (where I have timeouts of 4-10s as extra security against flaky tests).  This is usually something you want to avoid in unit tests, but our test suite is still small enough to side-step the guideline of separating medium-to-large tests from the small, quick ones.

Hope that sheds some light on the issue!

Wikipedia/assets/preview.html
```diff
@@ -12,7 +12,6 @@
     </head>
     <body>
         <div class="content" id="content">
-            <!-- %@ don't remove unless removed from index.html too -->
```
bgerstle: yay, thanks @montehurd from the past, for the helpful tip!

www/less/collapseTables.less
```diff
@@ -15,6 +15,7 @@ table.infobox caption {
     border-radius: 2px;
     padding:1px;
     box-shadow: inset 0 0 1px gray;
+    clear:both;
```
bgerstle: @montehurd is this an update from upstream? just want to be sure

bgerstle:
@montehurd looks good :+1:, just answer the couple inline questions I wrote and I'll merge

www/less/collapseTables.less
montehurd: No, not upstream.

coreyfloyd:
wow - I was able to follow your conversation… how novel :wink:

PR #21, opened 2015-07-28
Parameterize web style origin to allow testing w/ vagrant

bgerstle:
Change CSS_PREFIX in the Makefile to point the "make web" goal's style
requests at the desired host (i.e. production or localhost).

Also removed node version check since it's too hard to check semver in bash


bgerstle:
@coreyfloyd please review

PR #22, opened 2015-07-28
System history tab

bgerstle:
So... all I initially set out to do was this:

image

but instead spent a bit too much time find/replace/refactor-ing all the glyph string building & offset data.

#### `NSString+WMFGlyphs`
New home for getting glyph characters for a given enum, as well as the baseline offset needed to center it in the frame.  Also provides conveniences on `NSAttributedString` for building attributed strings of a glyph at a given size, color, and offset.

#### On tab bar items
Our glyphs have no power here...  You can only set a single string for the title and a corresponding attributes dictionary, meaning we could only set the glyph w/o a title (i.e. no "Saved" tab w/ a heart icon).  If we want custom icons, we need a way to get a `UIImage`, either by using SVGs as the asset format or rendering one from a string on demand (and caching it).  Ticket for another time...

### DIFF
--- a/Wikipedia.xcodeproj/project.pbxproj
+++ b/Wikipedia.xcodeproj/project.pbxproj
@@ -282,6 +282,7 @@
 		BC5FE5751B1DFF5400273BC0 /* ArticleFetcherTests.m in Sources */ = {isa = PBXBuildFile; fileRef = BC5FE5741B1DFF5400273BC0 /* ArticleFetcherTests.m */; };
 		BC69C3141AB0C1FF0090B039 /* WMFImageInfoController.m in Sources */ = {isa = PBXBuildFile; fileRef = BC69C3131AB0C1FF0090B039 /* WMFImageInfoController.m */; };
 		BC6BF4001B19213600362968 /* XCTestCase+WMFLocaleTesting.m in Sources */ = {isa = PBXBuildFile; fileRef = BC6BF3FE1B19209900362968 /* XCTestCase+WMFLocaleTesting.m */; };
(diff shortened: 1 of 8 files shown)


bgerstle:
@coreyfloyd @montehurd please have a look at the refactor, should have grabbed everything correctly.

coreyfloyd:
@bgerstle yeah - I actually have a ticket to replace the fonts with SVG/PDF images instead. That will make this stuff cake… and also gives us better control over the glyphs in general since we can treat them as images instead of text.

coreyfloyd:
@bgerstle wow -0.4%? I'm disappointed.

coreyfloyd:
@bgerstle no icons for the other tabs? Thought we were just going to do placeholders for all until we gets real ones?

bgerstle:
@coreyfloyd turns out there's no built-in tab bar item for "nearby", but we could use the "Downloads", "Bookmarks", or "Favorites" ones for "Saved." i don't care either way

bgerstle:
feel free to change it and push

bgerstle:
> -0.04%
@coreyfloyd i know :-( soon...

bgerstle:
@coreyfloyd @montehurd i'm ambivalent about the other tabs, especially since this is only temporary. I'd rather leave it as is than try to use an (IMO unfitting) placeholder.

coreyfloyd:
Lets go all or none… maybe i can get some icons real quick

coreyfloyd:
sent an email…waiting to hear back

coreyfloyd:
@bgerstle you want to back out your tab changes and then we can merge the rest of your refactors?

bgerstle:
Yeah, will do that in a bit

coreyfloyd:
:+1:

PR #24, opened 2015-07-30
Fix for iPad constraint warning.

montehurd:
Change-Id: I287c39be77238668c16bc5480f2d2b866f962581


coreyfloyd:
@montehurd you can delete the gerrit hook from .git/hooks now (it is the commit msg hook)

PR #25, opened 2015-07-30
fix threading race condition

bgerstle:
`MWKSiteInfoFetcherTests` were also making the mistake of dispatching back to a background thread, resulting in a race condition that could cause [`fulfill` to be called before we started waiting](https://s3.amazonaws.com/archive.travis-ci.org/jobs/73295822/log.txt).


bgerstle:
@coreyfloyd quick review plz

coreyfloyd:
Another drop in test coverage?

PR #27, opened 2015-07-30
Pin Section Headers in web view to top of screen

montehurd:
https://phabricator.wikimedia.org/T106109


Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/TitleOverlayModel.h
```diff
@@ -0,0 +1,17 @@
+//  Created by Monte Hurd on 7/30/15.
+//  Copyright (c) 2015 Wikimedia Foundation. Provided under MIT-style license; please copy and modify!
+
+#import <Foundation/Foundation.h>
+
+@class TitleOverlayLabel;
+@interface TitleOverlayModel : NSObject
```
bgerstle: missing `WMF` prefix

Wikipedia/View Controllers/WebView/WebViewController.m
```diff
@@ -257,6 +269,35 @@ - (void)viewDidLoad {
     self.webView.scrollView.layer.anchorPoint = CGPointMake((isRTL ? 1.0 : 0.0), 0.0);
 
     [self tocUpdateViewLayout];
+
+    @weakify(self)
```
bgerstle: :+1: for using `KVOControllerNonRetaining`. FYI you can avoid the `weak/strong` dance by using the `observer` argument passed to the block (which is the same as `self` in this case). If you do this, I also recommend you cast `observer` from `id` to whatever type `self` is for more convenient (and safer) type checking.

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/TitleOverlayLabel.h
```diff
@@ -0,0 +1,11 @@
+//  Created by Monte Hurd on 7/30/15.
+//  Copyright (c) 2015 Wikimedia Foundation. Provided under MIT-style license; please copy and modify!
+
+#import <UIKit/UIKit.h>
+#import "PaddedLabel.h"
+
+@interface TitleOverlayLabel : PaddedLabel
```
coreyfloyd: Said on hangout, but just adding here as well:
Lets put this in a nib and add the label as a subview. Also use a standard UILabel for the actual text (can configure in the nib).

Wikipedia/View Controllers/WebView/WebViewController.m
```diff
@@ -51,6 +57,11 @@ @interface WebViewController () <LanguageSelectionDelegate, FetchFinishedDelegat
 
 @property (strong, nonatomic) WMFArticlePopupTransition* popupTransition;
 
+@property (nonatomic, strong) NSMutableArray* nativeTitleLabelModelsArray;
```
coreyfloyd: Instead of the array… we can add a "SectionTitlesViewController" that manages the array of models / views that float on the web view. So that logic is then encapsulated separately.

You can also move the 3 properties below to that class.

Wikipedia/View Controllers/WebView/WebViewController.m
```diff
@@ -2045,4 +2053,142 @@ - (void)promptZeroOff {
     [dialog show];
 }
 
+#pragma mark Native section title label overlays
```
coreyfloyd: Pretty much all of this code can be moved into the new control class that handles the management of the section headers

Wikipedia/View Controllers/WebView/WebViewController.m
montehurd: Fixed!

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/TitleOverlayModel.h
montehurd: Oops. Got it.

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/TitleOverlayLabel.h
montehurd: Will do.

Wikipedia/View Controllers/WebView/WebViewController.m
montehurd: Mostly done w/this last commit. Updating it now to use protocol and/or delegate bits for communication.

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFSectionTitlesViewController.m
```diff
@@ -0,0 +1,231 @@
...
+    }
+    self.topStaticNativeTitleLabel       = [[WMFTitleOverlayLabel alloc] init];
+    self.topStaticNativeTitleLabel.alpha = 0;
+    [self.webViewController.view addSubview:self.topStaticNativeTitleLabel];
+    [self.topStaticNativeTitleLabel mas_makeConstraints:^(MASConstraintMaker* make) {
+        make.left.equalTo(self.webViewController.view.mas_left);
+        make.right.equalTo(self.webViewController.view.mas_right);
+    }];
+
+    self.topStaticNativeTitleLabelTopConstraint =
```
coreyfloyd: Can we switch this to Mantle?

bgerstle: @coreyfloyd *ahem*

bgerstle: @coreyfloyd i can't think of a mnemonic, but how about: "**Ma**sons **ma**ke and Mant**les** make **les**s"

bgerstle: Or,  "**ma** **son** springs and struts around while I burn duplicated code in my fire under the **mantle**."

coreyfloyd:
Just talked with Monte… he is going to do some follow on patches for:
- Removing the need to pass the WebViewController to the TitlesViewController
- Move manual constraint code to Masonry

Checking to see if everything else looks good so we can merge.

cc @bgerstle

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFSectionTitlesViewController.h
```diff
@@ -0,0 +1,16 @@
+//  Created by Monte Hurd on 7/31/15.
+//  Copyright (c) 2015 Wikimedia Foundation. Provided under MIT-style license; please copy and modify!
+
+#import <Foundation/Foundation.h>
+
+@interface WMFSectionTitlesViewController : NSObject
+
+@property (nonatomic, strong) UIWebView* webView;
```
coreyfloyd: Can we init with the needed properties instead of setting them later? Makes it easier to test when we inject on initialization and make them immutable.

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFSectionTitlesViewController.m
```diff
@@ -0,0 +1,231 @@
...
+#import "WMFSectionTitlesViewController.h"
+#import "WMFTitleOverlayLabel.h"
+#import "WMFTitleOverlayModel.h"
+#import "NSString+Extras.h"
+#import "UIView+WMFSearchSubviews.h"
+#import <Masonry/Masonry.h>
+
+@interface WMFSectionTitlesViewController ()
+
+@property (nonatomic, strong) NSMutableArray* models;
```
coreyfloyd: "models" is a bit generic… better to name them for their purpose… like overlayModels or overlayViewModels.

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFSectionTitlesViewController.m
```diff
@@ -0,0 +1,231 @@
...
+
+@end
+
+@implementation WMFSectionTitlesViewController
+
+- (instancetype)init {
+    self = [super init];
+    if (self) {
+        self.models                            = @[].mutableCopy;
+        self.indexOfNativeTitleLabelNearestTop = -1;
```
coreyfloyd: NSNotFound would be more appropriate

bgerstle:
@coreyfloyd 10-4, will skim through to see if i notice anything else major

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFTitleOverlayModel.h
```diff
@@ -0,0 +1,17 @@
...
+@class WMFTitleOverlayLabel;
+@interface WMFTitleOverlayModel : NSObject
+
+@property (nonatomic, strong) NSLayoutConstraint* topConstraint;
+@property (nonatomic, strong) NSString* anchor;
+@property (nonatomic, strong) NSString* title;
+@property (nonatomic) CGFloat yOffset;
+@property (nonatomic) NSNumber* sectionId;
+
+@property (nonatomic, strong) WMFTitleOverlayLabel* label;
```
bgerstle: why does a model have a reference to a label? If we're going to break MVC, there should at least be clear documentation for the rationale.

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFSectionTitlesViewController.m
```diff
@@ -0,0 +1,231 @@
...
+
+- (void)updateOverlaysPositions {
+    NSArray* headingsTopOffsets = [self getSectionTitlesLocationsJSON];
+    if (headingsTopOffsets.count == self.models.count) {
+        for (NSUInteger i = 0; i < self.models.count; i++) {
+            WMFTitleOverlayModel* m = self.models[i];
+            if (m.anchor && m.topConstraint) {
+                NSNumber* topOffset = headingsTopOffsets[i];
+                CGFloat topFloat    = topOffset.floatValue + self.webView.scrollView.contentOffset.y;
+                m.topConstraint.constant = topFloat;
```
coreyfloyd: "touch your friends, but not friends of your friends"

To maintain encapsulation the WMFTitleOverlayModel should really be the one updating the constraint.
For instance: lines 124-126 could be replaced with:
- (void)updateConstraintWithNewYOffset:(CGFloat)yOffset;

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFSectionTitlesViewController.m
```diff
@@ -0,0 +1,231 @@
...
+}
+
+- (void)updateOverlaysPositions {
+    NSArray* headingsTopOffsets = [self getSectionTitlesLocationsJSON];
+    if (headingsTopOffsets.count == self.models.count) {
+        for (NSUInteger i = 0; i < self.models.count; i++) {
+            WMFTitleOverlayModel* m = self.models[i];
+            if (m.anchor && m.topConstraint) {
+                NSNumber* topOffset = headingsTopOffsets[i];
+                CGFloat topFloat    = topOffset.floatValue + self.webView.scrollView.contentOffset.y;
```
coreyfloyd: "+ self.webView.scrollView.contentOffset.y" should be translated to domain knowledge so we understand why this is done.

To increase readability this line could be replaced with: 
```
topOffset.floatValue + [self headerOffSet];

- (CGFloat)headerOffset{
    return self.webView.scrollView.contentOffset.y;
}
```

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFSectionTitlesViewController.m
```diff
@@ -0,0 +1,231 @@
...
+                                        toItem:self.webViewController.topLayoutGuide
+                                     attribute:NSLayoutAttributeBottom
+                                    multiplier:1.0
+                                      constant:0];
+    [self.webViewController.view addConstraint:self.topStaticNativeTitleLabelTopConstraint];
+}
+
+- (void)updateOverlaysPositions {
+    NSArray* headingsTopOffsets = [self getSectionTitlesLocationsJSON];
+    if (headingsTopOffsets.count == self.models.count) {
```
coreyfloyd: Probably should be moved to an assertion and bail out at the top of the method. 

Since this is required and should always be true.

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFSectionTitlesViewController.m
```diff
@@ -0,0 +1,231 @@
...
+                                     attribute:NSLayoutAttributeTop
+                                     relatedBy:NSLayoutRelationEqual
+                                        toItem:self.webViewController.topLayoutGuide
+                                     attribute:NSLayoutAttributeBottom
+                                    multiplier:1.0
+                                      constant:0];
+    [self.webViewController.view addConstraint:self.topStaticNativeTitleLabelTopConstraint];
+}
+
+- (void)updateOverlaysPositions {
```
coreyfloyd: update is a little generic… maybe "recalculateOverlayPositions"

also an extra "s" in there.

Wikipedia/View Controllers/WebView/WebViewController.m
```diff
@@ -257,6 +261,33 @@ - (void)viewDidLoad {
     self.webView.scrollView.layer.anchorPoint = CGPointMake((isRTL ? 1.0 : 0.0), 0.0);
 
     [self tocUpdateViewLayout];
+
+    [self.KVOControllerNonRetaining observe:self.webView.scrollView keyPath:WMF_SAFE_KEYPATH(UIScrollView.new, contentSize) options:NSKeyValueObservingOptionNew block:^(WebViewController* observer, id object, NSDictionary* change) {
```
bgerstle: seems like the section header VC's abstraction is leaking into the webVC. I think you'll cover this by giving the section VC a (weak) reference to the webview it's reacting to.

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFSectionTitlesViewController.m
```diff
@@ -0,0 +1,231 @@
...
+- (void)updateTopOverlayForScrollOffsetY:(CGFloat)offsetY {
+    [self updateIndexOfNativeTitleLabelNearestTopForScrollOffsetY:offsetY];
+    [self nudgeTopStaticTitleLabelIfNecessaryForScrollOffsetY:offsetY];
+}
+
+- (void)updateIndexOfNativeTitleLabelNearestTopForScrollOffsetY:(CGFloat)offsetY {
+    self.topStaticNativeTitleLabel.alpha = (offsetY <= 0) ? 0 : 1;
+
+    CGFloat lastOffset          = 0;
+    NSUInteger newTopLabelIndex = -1;
```
coreyfloyd: NSNotFound

Wikipedia/View Controllers/WebView/WebViewController.m
```diff
@@ -257,6 +261,33 @@ - (void)viewDidLoad {
...
+        [self.sectionTitlesViewController updateOverlaysPositions];
+        // Restrict the web view from scrolling horizonally.
+        [object preventHorizontalScrolling];
```
bgerstle: why do we need to do this here? do we ever want it to scroll horizontally? if not, we might also want to consider handling cases when (after loading) the webview's content size is wider than the view's bounds.

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFSectionTitlesViewController.m
```diff
@@ -0,0 +1,231 @@
...
+        }
+    }
+}
+
+- (void)updateTopOverlayForScrollOffsetY:(CGFloat)offsetY {
+    [self updateIndexOfNativeTitleLabelNearestTopForScrollOffsetY:offsetY];
+    [self nudgeTopStaticTitleLabelIfNecessaryForScrollOffsetY:offsetY];
+}
+
+- (void)updateIndexOfNativeTitleLabelNearestTopForScrollOffsetY:(CGFloat)offsetY {
```
coreyfloyd: You can kill "Native" in names for the most part… native is the rule, not the exception. Maybe just name anything that is web based with "web" is needed.

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFSectionTitlesViewController.m
```diff
@@ -0,0 +1,231 @@
...
+- (id)getJSONFromWebViewUsingFunction:(NSString*)jsFunctionString {
+    NSString* jsonString = [self.webView stringByEvaluatingJavaScriptFromString:jsFunctionString];
+    NSData* jsonData     = [jsonString dataUsingEncoding:NSUTF8StringEncoding];
+    NSError* error       = nil;
+    id result            = [NSJSONSerialization JSONObjectWithData:jsonData options:0 error:&error];
+    return (error) ? nil : result;
+}
+
+static NSString* const WMFJSGetSectionTitlesJSON =
+    @"(function(){"
```
bgerstle: could this be in a JS file (which is lintable, testable, and portable)?

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFSectionTitlesViewController.m
```diff
@@ -0,0 +1,231 @@
...
+    @"  });"
+    @"  return JSON.stringify(nodeArray);"
+    @"})();";
+
+- (NSArray*)getSectionTitlesJSON {
+    return [self getJSONFromWebViewUsingFunction:WMFJSGetSectionTitlesJSON];
+}
+
+static NSString* const WMFJSGetSectionTitlesLocationsJSON =
+    @"(function(){"
```
bgerstle: see [above](https://github.com/wikimedia/wikipedia-ios/pull/27/files#r36116047) about moving into a JS file

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFSectionTitlesViewController.m
coreyfloyd: Seems misnamed… not sure what "IndexOf" is there for:

maybe just:
```
- (void)updateTopVisibleHeaderForScrollOffsetY:(CGFloat)offsetY;
```

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFSectionTitlesViewController.m
```diff
@@ -0,0 +1,231 @@
...
+            self.indexOfNativeTitleLabelNearestTop = newTopLabelIndex;
+            [self updateTopStaticTitleLabelText];
+        }
+    }
+}
+
+- (void)nudgeTopStaticTitleLabelIfNecessaryForScrollOffsetY:(CGFloat)offsetY {
+    NSUInteger pusherIndex  = self.indexOfNativeTitleLabelNearestTop + 1;
+    CGFloat distanceToPushY = 0;
+    if (pusherIndex < (self.models.count)) {
```
bgerstle: this might be a good place for `DDLogWarn` which tells us we tried to nudge/scroll something out of bounds

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFSectionTitlesViewController.m
```diff
@@ -0,0 +1,231 @@
...
+#import "WMFTitleOverlayLabel.h"
+#import "WMFTitleOverlayModel.h"
+#import "NSString+Extras.h"
+#import "UIView+WMFSearchSubviews.h"
+#import <Masonry/Masonry.h>
+
+@interface WMFSectionTitlesViewController ()
+
+@property (nonatomic, strong) NSMutableArray* models;
+@property (nonatomic) NSUInteger indexOfNativeTitleLabelNearestTop;
```
coreyfloyd: I'd get away from naming variables with "…titleLabel" and move to something more agnostic like "sectionHeader". They are really more of a container for the title label and may have other things in there that aren't labels (like an edit button).

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFSectionTitlesViewController.m
```diff
@@ -0,0 +1,231 @@
...
+                                     relatedBy:NSLayoutRelationEqual
+                                        toItem:self.webViewController.topLayoutGuide
+                                     attribute:NSLayoutAttributeBottom
+                                    multiplier:1.0
+                                      constant:0];
+    [self.webViewController.view addConstraint:self.topStaticNativeTitleLabelTopConstraint];
+}
+
+- (void)updateOverlaysPositions {
+    NSArray* headingsTopOffsets = [self getSectionTitlesLocationsJSON];
```
bgerstle: potential follow-on, but could be in scope here. we should cache this until another page is loaded (or perhaps lazily calculated/stored on `MWKArticle`, similar to how face offsets are lazily detected & stored).

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFSectionTitlesViewController.m
```diff
@@ -0,0 +1,231 @@
...
+    [self nudgeTopStaticTitleLabelIfNecessaryForScrollOffsetY:offsetY];
+}
+
+- (void)updateIndexOfNativeTitleLabelNearestTopForScrollOffsetY:(CGFloat)offsetY {
+    self.topStaticNativeTitleLabel.alpha = (offsetY <= 0) ? 0 : 1;
+
+    CGFloat lastOffset          = 0;
+    NSUInteger newTopLabelIndex = -1;
+
+    for (NSUInteger thisIndex = 0; thisIndex < self.models.count; thisIndex++) {
```
coreyfloyd: should try to stay away from low level C stuff, and try to use the high-level abstraction methods like: `enumerateObjects…` and friends.

coreyfloyd: You could also look into even higher level stuff like bk_reduce.

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFSectionTitlesViewController.m
```diff
@@ -0,0 +1,231 @@
...
+
+static NSString* const WMFJSGetSectionTitlesJSON =
+    @"(function(){"
+    @"  var nodeList = document.querySelectorAll('h1.section_heading');"
+    @"  var nodeArray = Array.prototype.slice.call(nodeList);"
+    @"  nodeArray = nodeArray.map(function(n){"
+    @"    var rect = n.getBoundingClientRect();"
+    @"    return {"
+    @"        anchor:n.id,"
+    @"        sectionId:n.getAttribute('sectionId'),"
```
bgerstle: just because i don't want to take this for granted, we should verify none of these values are unexpected (e.g. `undefined` or `null`) and return an explicit error (which we check for in native land). in cases where we rely on the DOM, we should expect the unexpected

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFSectionTitlesViewController.m
```diff
@@ -0,0 +1,231 @@
...
+        if (title) {
+            title = [title wmf_stringByRemovingHTML];
+
+            WMFTitleOverlayLabel* label = [[WMFTitleOverlayLabel alloc] init];
+            label.text      = title;
+            label.sectionId = sectionId;
+
+            [self.webView.scrollView addSubview:label];
+
+            NSLayoutConstraint* (^ constrainEqually)(NSLayoutAttribute) = ^NSLayoutConstraint*(NSLayoutAttribute attr) {
```
bgerstle: would prefer if this used masonry as well

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFSectionTitlesViewController.m
```diff
@@ -0,0 +1,231 @@
...
+        self.indexOfNativeTitleLabelNearestTop = -1;
+    }
+    return self;
+}
+
+- (void)hideTopOverlay {
+    self.topStaticNativeTitleLabel.alpha = 0;
+}
+
+- (void)addOverlays {
```
bgerstle: I think something like `reload` or `reset` would make more sense here instead of `add` (which implies that the operation isn't idempotent and doesn't **remove** anything)

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFSectionTitlesViewController.m
montehurd: I was able to switch the code to use mas_updateConstraints to vary the top constant of topStaticNativeTitleLabel as scrolling happens (i.e. nudging it when a section label needs to "push" it up offscreen). However, mas_updateConstraints  allocs a new MASConstraintMaker each time it runs, which itself allocs a new NSArray. We can't do this though because this code needs to run *each* time the scroll view scrolls so we need it to be as fast as possible, and simply maintaining a handle on "topStaticNativeTitleLabelTopConstraint" means we don't do *any* unnecessary alloc'ing.

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFSectionTitlesViewController.m
```diff
@@ -0,0 +1,231 @@
...
+        } else if ((thisIndex == (self.models.count - 1)) && (offsetY > thisOffset)) {
+            newTopLabelIndex = thisIndex;
+            break;
+        }
+        lastOffset = thisOffset;
+    }
+
+    if (newTopLabelIndex != -1) {
+        if (newTopLabelIndex != self.indexOfNativeTitleLabelNearestTop) {
+            self.indexOfNativeTitleLabelNearestTop = newTopLabelIndex;
```
coreyfloyd: Instead of setting state and then calling `updateTopStaticTitleLabelText` it would be better to just pass the desired value into the method:

```
- (void) updateTopStaticSectionHeaderWithModel:(WMFTitleOverlayModel*)model;
```

Then you can remove the ivar entirely. 

We should always strive to not have an ivar if possible - any state we have to maintain and update is expensive. if you can just calculate it in place and "forget about it" it makes the code easier to maintain.

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFSectionTitlesViewController.m
```diff
@@ -0,0 +1,231 @@
...
+            [self updateTopStaticTitleLabelText];
+        }
+    }
+}
+
+- (void)nudgeTopStaticTitleLabelIfNecessaryForScrollOffsetY:(CGFloat)offsetY {
```
coreyfloyd: "nudge" is a bit colloquial… could we change that to something like "update"? We can probably also reduce the method name down to something like:

```
- (void)updateTopStaticHeaderWithScrollOffsetY:(CGFloat)offsetY {
```

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFSectionTitlesViewController.h
montehurd: Done.

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFSectionTitlesViewController.m
```diff
@@ -0,0 +1,231 @@
...
+- (void)updateTopStaticTitleLabelText {
+    self.topStaticNativeTitleLabel.alpha = 1.0;
+    WMFTitleOverlayModel* m = self.models[self.indexOfNativeTitleLabelNearestTop];
+    self.topStaticNativeTitleLabel.text      = m.title;
+    self.topStaticNativeTitleLabel.sectionId = m.sectionId;
+}
+
+#pragma mark Section title and title location determination
+
+- (id)getJSONFromWebViewUsingFunction:(NSString*)jsFunctionString {
```
coreyfloyd: This method feels like it should be on the web view itself (as a category?)

coreyfloyd: Actually this and all the methods below seem like more of a toolkit that should be in a more centralized js place. Do we have a place where we keep our "js strings" and this type of logic now?

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFTitleOverlayLabel.h
```diff
@@ -0,0 +1,11 @@
+//  Created by Monte Hurd on 7/30/15.
+//  Copyright (c) 2015 Wikimedia Foundation. Provided under MIT-style license; please copy and modify!
+
+#import <UIKit/UIKit.h>
+#import "PaddedLabel.h"
+
+@interface WMFTitleOverlayLabel : PaddedLabel
+
+@property (nonatomic) NSNumber* sectionId;
```
coreyfloyd: Seems unnecessary.  Ideally we should treat these headers as we would table view cells. They should mostly be view information - avoiding any model knowledge if possible.

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFSectionTitlesViewController.m
bgerstle: @montehurd did a bit of looking around, and it appears Masonry attempts to support this via `-[MASConstraint setOffset:]`. you could try:

```objc
// change ref to constraint to be of type `MASConstraint`
@property (nonatomic, weak /*or strong?*/) MASConstraint* topConstraint;
// when setting up labels w/ constraints
[label mas_makeConstraints:^(MASConstraintMaker* make) {
  self.topConstraint = make.label.top.equalTo(self.webView); // if needed: .with.offset(someInitialOffset);
}];
// when you need to update the constant:
[self.topConstraint setOffset:newOffset];
```

bgerstle: It's not mandatory for us to do use Masonry for this, but it might be useful elsewhere.

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFTitleOverlayModel.h
```diff
@@ -0,0 +1,17 @@
+//  Created by Monte Hurd on 7/30/15.
+//  Copyright (c) 2015 Wikimedia Foundation. Provided under MIT-style license; please copy and modify!
+
+#import <Foundation/Foundation.h>
+
+@class WMFTitleOverlayLabel;
+@interface WMFTitleOverlayModel : NSObject
```
coreyfloyd: I know when I mentioned making this class I called it a "Model" class… but it may be helpful to think of it more of a controller class.

It should be "in charge of" the section header that it is assigned. So things like title, label, and constraint should be internal so only this object can modify them. That will make it much easier to debug later on since only one object can make changes.

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFSectionTitlesViewController.m
coreyfloyd: @montehurd yeah - I think the fallback to holding the reference to the constraint and updating the offset as @bgerstle shows is probably the way to go.

montehurd: Oh ok thanks I'll try that.

Wikipedia/View Controllers/WebView/WebViewController.m
```diff
@@ -257,6 +259,34 @@ - (void)viewDidLoad {
...
+- (void)editSection:(NSNotification*)notification {
+    if ([self tocDrawerIsOpen]) {
+        [self tocHide];
+        return;
+    }
+
+    if (self.editable) {
+        WMFTitleOverlayLabel* sender = notification.object;
+        self.sectionToEditId = [sender sectionId];
+        [self showSectionEditor];
```
coreyfloyd: Instead of setting an ivar and then calling a method with no parameters, you can do something like:

```
- (void)showSectionEditorForSection:(NSUInteger)sectionID
```

Then you can get rid of the ivar… similar to another comment I left earlier.

Wikipedia/View Controllers/WebView/WebViewController.m
```diff
@@ -1059,6 +1057,10 @@ - (void)scrollViewDidScroll:(UIScrollView*)scrollView {
     CGFloat distanceScrolled     = self.scrollViewDragBeganVerticalOffset - scrollView.contentOffset.y;
     CGFloat fabsDistanceScrolled = fabs(distanceScrolled);
 
+    if (![self tocDrawerIsOpen]) {
```
coreyfloyd: At this point, you can remove all the "tocDrawer" stuff totally. The no need to respond to changes. You can leave the "TOC" button unhooked, and I can hook it back up in my commit.

Wikipedia/View Controllers/WebView/WebViewController.m
```diff
@@ -2015,7 +2011,7 @@ - (void)languageSelected:(MWKLanguageLink*)langData sender:(LanguagesViewControl
     [self dismissViewControllerAnimated:YES completion:nil];
 }
 
-#pragma Wikipedia Zero alert dialogs
+#pragma mark Wikipedia Zero alert dialogs
```
coreyfloyd: If you add "-" to be `#pragma mark - Wikipedia Zero alert dialogs`, then you get a nice underline as well.

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFSectionTitlesViewController.m
montehurd: This works! Pushing commit...

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFSectionTitlesViewController.m
coreyfloyd: @bgerstle I suggested a category on UIWebView because it is presentation related… it depends on the view port. Not something MWKArticle can know without it being displayed.

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFSectionTitlesViewController.m
coreyfloyd: agreed

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFSectionTitlesViewController.m
montehurd: Done!

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFSectionTitlesViewController.m
montehurd: Done.

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFSectionTitlesViewController.m
montehurd: Ah good point!

montehurd: Done.

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFSectionTitlesViewController.m
montehurd: Killed "native".

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFTitleOverlayLabel.h
montehurd: This was just temporarily added to the label to support the long-press edit.

montehurd: iirc.

Wikipedia/View Controllers/WebView/NativeSectionTitleOverlays/WMFTitleOverlayModel.h
montehurd: This was WIP. Will change based on what Corey and I discussed which he summed up above.

Wikipedia/View Controllers/WebView/WebViewController.m
montehurd: Good point.

Wikipedia/View Controllers/WebView/WebViewController.m
montehurd: This isn't new with this patch.

Wikipedia/View Controllers/WebView/WebViewController.m
montehurd: Fixed.

Wikipedia/View Controllers/WebView/WebViewController.m
montehurd: Fixed.

Wikipedia/View Controllers/WebView/WebViewController.m
montehurd: Ah forgot. Fixed.

PR #26, opened 2015-07-30
Preloading web view when the article is loaded in the ArticleVC

coreyfloyd:
https://phabricator.wikimedia.org/T106107

PR #28, opened 2015-07-31
Clean up tab bar

coreyfloyd:
Based on #26. Please review that first.

Added icons and embedded each tab contents in a navigation controller. 

Needed to embed the navigation controller for https://phabricator.wikimedia.org/T106113 so figured do this with the tab icons.


bgerstle:
@coreyfloyd y u no screenshot? 😥

coreyfloyd:
For brian:

screenshot 2015-08-03 15 45 08

coreyfloyd:
err @bgerstle

bgerstle:
LGTM, but a bookmark instead of a heart? is that from design?

bgerstle:
anyway, :+1:

montehurd:
:-1: 
 
Nearby constraints are way broken. Rotate to landscape and back to portrait to repro.

montehurd:
:+1: 
Opps nevermind! This issue pre-dates your patch and it looks to be an issue with whatever is containing nearby, not nearby itself.

PR #29, opened 2015-07-31
Show links in article summary

bgerstle:
### Checklist
- [x] Render HTML natively (DTCoreText)
- [x] Custom XPath for summary
  - [x] filter out coordinates (see taj mahal snapshot reference)

<br/><strike>- [ ] filter out IPA?</strike>
<strike>- [ ] Interactions w/ native content</strike>
_Going to do interaction in a separate patch_

#### Amendments
- [x] change `UIColor+WMFPalette` to the un-artistic `UIColor+WMFStyle`
- [x] remove premature `@autoreleasepool` optimization
- [x] remove `#define WMFParagraphSelector` in favor of vanilla string formatting
- [x] move "minimal content controller" into a custom attributed text cell, encapsulating all logic in one place

Using DTCoreText to render pages natively, doing a different XPath to grab section content.  See **Extraction** section in `MWKArticle` for implementation details and the new `WikipediaUnitTests/ReferenceImages` folder for visual examples.


bgerstle:
@montehurd don't know if this is a known issue, but headers "stack" and multi-line ones are visible beneath single-line (observed on Albert Einstein):

image

montehurd:
678 new files? Is there a sub-spec we could reference to trim?

bgerstle:
@montehurd not a bad idea to look into picking only the subspecs we need, but FWIW the total integration size of DTCoreText according to [CocoaPods](https://cocoapods.org/pods/DTCoreText) is just shy of a megabyte.  I've filed an [eng backlog task](https://phabricator.wikimedia.org/T107872) to look into using subspecs instead of the main pod.

bgerstle:
@montehurd also, one of the new pods is for testing only, so it doesn't impact the application's binary size

coreyfloyd:
@bgerstle FYI - Looks like we lost all of our conversation history because of the rebase. I think we probably want to go with merges once you create a pull request so we don't lose our comments.

bgerstle:
@coreyfloyd good point... will use merges going forward

coreyfloyd:
:+1: Looks good to me

@montehurd are you interested in giving this another once over?

montehurd:
Hmm from a clean install I just get white w no text in the summary area until I drag the view around a bit...

montehurd:
Also none of the links actually work - is that coming in a follow on?

montehurd:
One more - if I tap on a TOC item, then go back to the native card and tap on some text in the summary, when it loads the web view again it's not scrolled to the top - it's still scrolled to the previous section from the first TOC item tap.

montehurd:
Also need to double the whitespace around the summary text view (so the left margin for example is consistent with the TOC text left margin etc)

bgerstle:
There's no interactivity yet, that's a follow-on once the pop-up stuff is added in #33. @montehurd let me check on the other issues and L/R margins

montehurd:
I still think we should only show the first paragraph. Some lead sections are huge.

montehurd:
Ah cool.

coreyfloyd:
@montehurd lets have design get a handle on how big the lead sections are so they can come up with a solve. we need to show them the problem so they can come up with an idea.

montehurd:
@coreyfloyd yeah but we can help identify these problems - and we *really* should or this is going to take twice as long

bgerstle:
easy to tweak at this point, can we give it to design (and more people) to weigh in on the number of sections we show here and/or whether the design should change to accommodate more/less paragraphs?

bgerstle:
oh wait, image tests are failing, 1 sec

bgerstle:
ok should be good w/ 2f785ad

PR #32, opened 2015-08-04
Web transition WIP

coreyfloyd:
https://phabricator.wikimedia.org/T106113

The first 8 commits are can be ignored… they are part of #31 

The transition itself is still a work in process… lots of unfinished stuff. Some of the bigger things that need done:
- [ ] When transitioning to web view, the header animates without considering the navigation bar
- [ ] The cell and the web view header are styled differently
- [ ] If the web view is scrolled the dismiss animation is wrong
- [ ] If the article VC is scrolled, the present animation is wrong


bgerstle:
@coreyfloyd closing since it was opened prematurely. we can re-open when necessary

PR #34, opened 2015-08-04
Adding Search tab

coreyfloyd:
Moved search to tab bar

https://phabricator.wikimedia.org/T107761


montehurd:
There's whitespace at the top even when search is not showing...

montehurd:
Also, one concern is that search is the most primary functionality of the app, and now instead of always being a single tap away (i.e. tap on the search bar and type), it's two taps away (tap on tab button, tap on search bar and type). This is a pretty huge UX change imo.

coreyfloyd:
@montehurd thanks… forgot to push a change… forthcoming…

montehurd:
ah cool

montehurd:
Getting crash if I do a clean install and search for "d"...

coreyfloyd:
@montehurd  huh? weird… looking

coreyfloyd:
@montehurd got a stack trace? can't reproduce

PR #35, opened 2015-08-05
Use a nib for the native title overlay instead of padded label

montehurd:
- Nib now contains both a standard label and an edit button.
- Removed a view property from the model obj.
- Simplified some logic in the controller w/blockskit.


coreyfloyd:
Once overall comment… can we normalize the terminology of the objects? We have "Sections", "Headers", "Overlays" and "Titles" being mixed about.

My personal preference would be "Section Headers". Because that most closely matches the intent of the classes… to be headers for each section (like a table view). 

So "WMFSectionTitlesViewController" becomes "WMFSectionHeadersViewController" and "WMFTitleOverlay" becomes "WMFSectionHeader" and so on…

It doesn't have to be this exact phrasing, but we should have some consistency in the names of the classes.

bgerstle:
@montehurd looks good! one quick suggestion: we could probably remove the `wmf_` prefix on the `WMFEditSectionProtocol` methods, since only our classes are conforming to it.  In general, protocol methods shouldn't need prefixes unless you know most of the conformers will be OS framework classes.

bgerstle:
@montehurd :+1:

coreyfloyd:
:+1:  build wasn't finished so I didn't merge. @bgerstle can merge when it does.

bgerstle:
thanks for also putting the JS into their own files, forgot about that! :+1: travis says go, so I say go! 👷

PR #36, opened 2015-08-05
Make native links interactive

bgerstle:
Last bit of work for [T106115](https://phabricator.wikimedia.org/T106115)

### Design notes
Abstracting away & encapsulating link parsing and handling behind a couple protocols & categories:

##### `WMFArticleNavigationDelegate`
Protocol for handlers of link events originating from an article's content, whether from web or native (eventually). See the checklist for all the interactions.

##### `NSURL+WMFLinkParsing`
If you have link that you want to tell your `WMFArticleNavigationDelegate` about, this is your category. Uses `-[MWKTitle initWithURL:]` to create titles from itself and call the correct delegate method for the type of URL it is.

As a user, I can:
- [x] Tap links in the native article view
- [x] Tap citations, and see the "references popup"
- [x] Tap external links, and see the MWZero warning if necessary

###### Easter Eggs 🐇🐣
Added the host to the warning users see when leaving WikiZero.

### DIFF
--- a/MediaWikiKit/MediaWikiKit/MWKArticle.h
+++ b/MediaWikiKit/MediaWikiKit/MWKArticle.h
@@ -100,7 +100,7 @@ static const NSInteger kMWKArticleSectionNone = -1;
 ///
 
 /**
- * @return Summary of the receiver as an attributd string built from HTML.
+ * @return Summary of the receiver as an attributed string built from HTML.
  */
 - (NSAttributedString*)summaryHTML;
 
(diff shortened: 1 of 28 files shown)


bgerstle:
@coreyfloyd @montehurd coming along!

example gif

Parsing out references might be tricky, but should be do-able via XPath.

bgerstle:
Tested MWZero:

in native

in web

montehurd:
I don't think it's related, but did you notice web view links stopped working?

bgerstle:
@montehurd sorry, wrote a comment this morning but apparently didn't hit submit. you're right on both counts (unrelated, webview links now broken). I filed a ticket for it (top of backlog now).

bgerstle:
For now, just scrolling to the tapped reference in the webview.  **Note:** haven't done testing for the case where full article isn't downloaded when ref is tapped, though the current implementation should fetch and then try again.  Here's how it looks:

citation tap

PR #38, opened 2015-08-06
Made tap on native title scroll that section to top.

montehurd:
https://phabricator.wikimedia.org/T108250


montehurd:
I also rename a protocol for consistency.

bgerstle:
:+1:

PR #39, opened 2015-08-06
When no wikidata text present in read more, center title vertically.

montehurd:
https://phabricator.wikimedia.org/T106186


bgerstle:
:+1: posted screencap of the fix to the ticket, but including here as well:

image

PR #42, opened 2015-08-07
Fix search suggestion button suggested article string substitution.

montehurd:
https://phabricator.wikimedia.org/T107942

- Now substitutes the suggested title properly.
- Also now properly sizes the button if a given localization and
  suggested title causes the button text to wrap to more than one
  line.
- Multi-line button condition now also properly updates button
  height if device is rotated.
- Removed a bunch of crufty constraints from the search controller
  in interface builder.


montehurd:
https://phabricator.wikimedia.org/T107942

montehurd:
Ha! Apparently I'm not the first person to run into this issue, or the first person to come up with the solution I came up with:

http://dissociativecode.tumblr.com/post/107618619978/multiline-uibutton-and-autolayout

bgerstle:
@montehurd indeed, looks like you're right in that we'd need some kind of trickery to get `UIButton` to render multiple lines correctly.  There's one edge case I can show you, but it's relatively minor.  That said, I'm wondering if we should take a step back and consider using a `UILabel` w/ a tap recognizer instead.  `UIButton` is great for most use cases, and cases where we're fine w/ hard-coding the number of lines and truncating the rest, but it seems like if we anticipate needing a dynamic height based on the number of lines, we should use something else.

we can chat about it after you rebase post-standup

bgerstle:
or rather, not rebase, merge master into your branch ;-)

montehurd:
I tried UILabel, but it actually ends up being more complicated than the single constraint + category method on UIButton approach. 

Reason it's more complicated: UILabel has no easy way to add padding w/o subclassing. This means you'd have to use constraints for padding or *really* fiddle with attributed string paragraph styling. But adding constraint padding means you can't just hide the label by letting it auto-collapse when it's not showing any text, as it does with the UIButton approach above.

montehurd:
Note: iOS 8 adds "layoutMargins" to UIView, which would, I think, make the UILabel approach less complicated than the UIButton + category method approach.

bgerstle:
k, fine as-is for now, if this becomes a recurring problem we can revisit

PR #47, opened 2015-08-11
Show logo placeholder in header gallery

bgerstle:
Bug: [T106504](https://phabricator.wikimedia.org/T106504)

example

### Known issues
- Placeholder will "flicker" before the actual image is set
- Placeholder doesn't look that great in the list, we could blur/darken it but that's extra work


coreyfloyd:
:+1: looks good - just waiting for travis to finish

PR #49, opened 2015-08-11
Redo article transitions to work inside a `UINavigationController`

bgerstle:
Bug: [T106152](https://phabricator.wikimedia.org/T106152)

### Overview
This patch:
- [x] Uses navigationController pushes to traverse articles
- [x] embeds the webVC into the articleContainer alongside the articleVC
- [x] modifies list & popup transitions to make them work within a navigationController context
- [x] vends list & popup transitions as animators & interaction controllers to each navigationController in the tab bar via `WMFNavigationTransitionController`
- [x] Modifies list & popup transitions so that dismissal can start even if the user isn't scrolled to the top of the content

### Known issues
Can't start an interactive dismissal when webview is showing. This is due to the list & popup transitions only checking to the container's `articleViewController.tableView` when determining if a dismissal should begin.  The plan is to fix this after more work has been done to solidify transition & navigation UI/UX.

Since `WebViewController` was moved into the container, it no longer has it's data set by the `ArticleViewController` (after obtaining a mobileView response).  I propose we fix this by giving both of the container's children the same article fetcher which will make sure that all requests are idempotent. This will allow article & web controllers to determine when they need the data irrespective of each other while ensuring that any requests they make never download or serialize the same data twice.

### Amendments
- [x] add back the code to transition from web view to native and back
- [x] Re-add "fastswipe"
- [x] Move the interactive transition bool (if there are no issues)
- [x] make webVC fetch article data if needed
- [x] use custom transitions from search list


Wikipedia/UI-V5/WMFAppViewController.m
```diff
@@ -35,6 +55,8 @@ @interface WMFAppViewController ()<UITabBarControllerDelegate>
 
 @property (nonatomic, strong) SessionSingleton* session;
 
+@property (nonatomic, strong) WMFNavigationTransitionController* navigationTransitionController;
```
coreyfloyd: Whats the purpose of the navigation controller here? Is this the list transition?

bgerstle: this is actually something that's a bit hacky.. and dangerously stateful.  let's chat on IRC/hangout

coreyfloyd:
@bgerstle so just starting to get a handle here… so far this is looking like 3 separate changes:

- Pushing article containers VC in the nav stack. and vending transitions for the root nav controller
- Moving popups to be based on navigation transition 
- Refactoring the container VC architecture.

Additionally the web link stuff was fixed somewhere in here… by some combination of these changes (but not exactly clear how/where - just basically architecture changes and hooking things up it seems)

Are there other large fixes here that I missed? If so can you list them out?

If this sounds about right, it would be good to split this up into 3 pull requests so we can address each of these concerns individually. 

Also it would be nice to explain exactly what was broken about the web link stuff and how this fixes it (it seems it was mostly a matter of hooking things up through delegation, but maybe there is more to it)

We can talk more when I get back, but thought it would be good to get this to you know so you can start pulling it apart.

bgerstle:
> - Pushing article containers VC in the nav stack. and vending transitions for the root nav controller
> - Moving popups to be based on navigation transition 

These two can't really be separated as the custom transitions needed to be modified to work with a navigation controller .

> - Refactoring the container VC architecture.

I can look into separating this out, but at this point it would probably be easier to just revert if this setup isn't satisfactory.

> Additionally the web link stuff was fixed somewhere in here… by some combination of these changes (but not exactly clear how/where - just basically architecture changes and hooking things up it seems)

In a nutshell, the webview was delegating to the article, who tried to present another article modally, even though its view wasn't in the hiearchy:

```
Container:
  -> ArticleVC -x-> ArticleVC
    -> WebView
```

IMO fixing this by asking the article to ask the container to present another article is hacky.  Better to have webview delegate to the container (just as the articleVC now does) so the container can own both article<->web transitions and pushing new content onto the stack.

> Are there other large fixes here that I missed? If so can you list them out?

It's intermingled w/ the "making transitions work in a navigation controller," but the ability to start transitions w/o being at the top of the view.

> If this sounds about right, it would be good to split this up into 3 pull requests so we can address each of these concerns individually. 

Let's talk when you get back. I can either amend or revert the article container architecture (i.e. try putting back the internal navigation controller)

> Also it would be nice to explain exactly what was broken about the web link stuff and how this fixes it (it seems it was mostly a matter of hooking things up through delegation, but maybe there is more to it)

See above

Wikipedia/UI-V5/WMFAppViewController.m
```diff
@@ -214,28 +253,18 @@ - (void)runDataMigrationIfNeededWithCompletion:(dispatch_block_t)completion {
...
 
+- (BOOL)tabBarController:(UITabBarController*)tabBarController shouldSelectViewController:(UIViewController*)viewController {
+    // !!!: disable pop to root behavior
+    return tabBarController.selectedViewController != viewController;
+}
+
 - (void)tabBarController:(UITabBarController*)tabBarController didSelectViewController:(UIViewController*)viewController {
     [self wmf_hideKeyboard];
-
-    WMFAppTabType tab = [[tabBarController viewControllers] indexOfObject:viewController];
```
coreyfloyd: how do the tabs get configured with the saved pages, etc?

Wikipedia/UI-V5/WMFArticleContainerViewController.h
```diff
@@ -1,18 +1,22 @@
...
 @class MWKArticle;
 
-@interface WMFArticleContainerViewController : UIViewController
+NS_ASSUME_NONNULL_BEGIN
 
-+ (instancetype)articleContainerViewControllerWithDataStore:(MWKDataStore*)dataStore savedPages:(MWKSavedPageList*)savedPages;
+@interface WMFArticleContainerViewController : UIViewController
+    <WMFArticleContentController, WMFArticleListItemController>
 
-@property (nonatomic, strong) MWKArticle* article;
```
coreyfloyd: This should be tied to a single article, no? Why is that removed from here?

Wikipedia/UI-V5/WMFAppViewController.m
bgerstle: where they were before, when we [configure each tab's root view controller](https://github.com/wikimedia/wikipedia-ios/pull/49/files#diff-bf61ea9e1bae88ed23fb56e10275bea9L55) (you might need to expand the diff.

Wikipedia/UI-V5/WMFArticleContainerViewController.h
bgerstle: It was moved into a protocol: `WMFArticleContentController` which the container, article & webVC's all conform to

coreyfloyd:
@bgerstle looking good - leaving comments we already discussed in hangout for posterity:

- Need to add back the code to transition from web view to native and back
- Need to resolve merge conflicts
- After I will run the app and put it through its paces
- Monte should do the same

montehurd:
Seeing some funky business on rotation.

Wikipedia/UI-V5/WMFArticleListTransition.m
```diff
@@ -25,43 +23,22 @@ @interface WMFArticleListTransition ()<UIGestureRecognizerDelegate>
...
-    return self;
-}
-
-- (id<UIViewControllerAnimatedTransitioning>)animationControllerForDismissedController:(UIViewController*)dismissed {
-    return self;
-}
-
-- (id<UIViewControllerInteractiveTransitioning>)interactionControllerForPresentation:(id<UIViewControllerAnimatedTransitioning>)animator {
-    return nil;
+- (BOOL)isPresenting {
```
coreyfloyd: @bgerstle while merging master into my pull - i looked at this again and remembered why this was 2 different BOOLs and not just the opposite of each other.

So isPresenting was supposed to only YES during the transition - when it is done presenting, isPresenting is NO since the presentation animation is finished. After presenting, isPresented is YES. At this time isDismissing is also NO since we are not dismissing either. So both isDismissing and isPresenting can both be NO at the same time.

I believe the isPresented variable was doing what you are trying to accomplish here now - but this was a separate concern. to put it another way. isPresented is about current state, isPresenting/isDismissing is about changing states. So to lay it out:

isPresented: Are we currently presented (regardless if a transition is running)?
isPresenting: Are we running the presentation transition animation?
isDismissing: Are we running the dismiss transition animation?

PR #53, opened 2015-08-11
skip empty sections when calculating summaryHTML

bgerstle:
Need to consider sections like "<p></p>" as empty. See "en.m.wiktionary.org/stationary" as an example.


coreyfloyd:
:+1: Looks good

PR #54, opened 2015-08-11
Moved rotation js out into a "webViewRotationScroll.js" file.

montehurd:
Formerly had been crammed into the web view controller.

### DIFF
--- a/Wikipedia/View Controllers/WebView/WebViewController.m
+++ b/Wikipedia/View Controllers/WebView/WebViewController.m
@@ -1490,7 +1490,8 @@ typedef NS_ENUM (NSInteger, WMFWebViewAlertType) {
 #pragma mark Scroll to last section after rotate
 
 - (void)willRotateToInterfaceOrientation:(UIInterfaceOrientation)toInterfaceOrientation duration:(NSTimeInterval)duration {
-    [self.webView wmf_callJavascriptFunctionNamed:@"setPreRotationRelativeScrollOffset" withArguments:nil];
+    [[self.webView wmf_javascriptContext][@"setPreRotationRelativeScrollOffset"] callWithArguments:nil];
+
 
     [self tocHideWithDuration:@0.0f];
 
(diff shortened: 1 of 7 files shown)


montehurd:
This was a left-over hygiene commit from my hygiene work yesterday.

bgerstle:
looks OK, just a couple questions/comments. let me know what you think

Wikipedia/View Controllers/WebView/WebViewController.m
```diff
@@ -1488,17 +1490,8 @@ - (BOOL)isCurrentArticleSaved {
...
-                   @"    if (_topElement) {"
-                   @"        var rect = _topElement.getBoundingClientRect();"
-                   @"        return rect.top / rect.height;"
-                   @"    } else {"
-                   @"        return 0;"
-                   @"    }"
-                   @"})()";
-    float relativeScrollOffset = [[self.webView stringByEvaluatingJavaScriptFromString:js] floatValue];
-    self.relativeScrollOffsetBeforeRotate = relativeScrollOffset;
+    [[self.webView wmf_javascriptContext][@"setPreRotationRelativeScrollOffset"] callWithArguments:nil];
```
bgerstle: shouldn't we be calling this w/ the relative offset?

bgerstle: oh wait, n/m the function doesn't take any args does it. the name was confusing

montehurd: No it's what determines the offset. Turned out the js "_topElement" var was keeping state, so since it was already doing this I just made it also keep "_preRotationOffsetY" so we didn't have to do so in Obj-C. Not ideal...

montehurd: This whole way of dealing with rotation is a bit hacky, but at least this contains the hackiness mostly in the JS file for now.

bgerstle:
:+1:

PR #56, opened 2015-08-12
Moved onboarding out of web view controller.

montehurd:
https://phabricator.wikimedia.org/T107182

- Also made the nearby view controller not start updating its location/heading until it's actually being used. This fixes the issue with the "app wants permission to use location data" popup appearing too soon.

- Also fixes visual glitch in transition between "splash" globe image and the onboarding view controller globe image.

### DIFF
--- a/Wikipedia/UI-V5/WMFAppViewController.m
+++ b/Wikipedia/UI-V5/WMFAppViewController.m
@@ -16,6 +16,7 @@
 #import "NSString+WMFGlyphs.h"
 #import "OnboardingViewController.h"
 #import "UIViewController+WMFStoryboardUtilities.h"
+#import "NearbyViewController.h"
 
 typedef NS_ENUM (NSUInteger, WMFAppTabType) {
     WMFAppTabTypeHome = 0,
@@ -31,6 +32,7 @@ typedef NS_ENUM (NSUInteger, WMFAppTabType) {
 @property (nonatomic, strong) IBOutlet UIView* splashView;
 @property (nonatomic, strong) UITabBarController* rootTabBarController;
 
+@property (nonatomic, strong, readonly) NearbyViewController* nearbyViewController;
 @property (nonatomic, strong, readonly) WMFSearchViewController* searchViewController;
 @property (nonatomic, strong, readonly) WMFArticleListCollectionViewController* savedArticlesViewController;
 @property (nonatomic, strong, readonly) WMFArticleListCollectionViewController* recentArticlesViewController;
@@ -45,6 +47,7 @@ typedef NS_ENUM (NSUInteger, WMFAppTabType) {
 
 - (void)loadMainUI {
     [self configureTabController];
+    [self configureNearbyViewController];
     [self configureSearchViewController];
     [self configureSavedViewController];
     [self configureRecentViewController];
@@ -52,6 +55,10 @@ typedef NS_ENUM (NSUInteger, WMFAppTabType) {
 
 - (void)configureTabController {
     self.rootTabBarController.delegate = self;
+}
+
+- (void)configureNearbyViewController {
+    //[self.nearbyViewController doSomething];
 }
 
 - (void)configureSearchViewController {
@@ -128,6 +135,10 @@ typedef NS_ENUM (NSUInteger, WMFAppTabType) {
 
 - (MWKUserDataStore*)userDataStore {
     return self.session.userDataStore;
+}
+
+- (NearbyViewController*)nearbyViewController {
+    return (NearbyViewController*)[self rootViewControllerForTab:WMFAppTabTypeHome];
 }
 
 - (WMFSearchViewController*)searchViewController {
@@ -245,7 +256,7 @@ typedef NS_ENUM (NSUInteger, WMFAppTabType) {
     WMFAppTabType tab = [[tabBarController viewControllers] indexOfObject:viewController];
     switch (tab) {
         case WMFAppTabTypeHome: {
-            //TODO: configure Nearby
+            [self configureNearbyViewController];
         }
         break;
         case WMFAppTabTypeSearch: {
--- a/Wikipedia/UI-V5/WMFAppViewController.m
+++ b/Wikipedia/UI-V5/WMFAppViewController.m
@@ -164,7 +164,7 @@ typedef NS_ENUM (NSUInteger, WMFAppTabType) {
 
         [self runDataMigrationIfNeededWithCompletion:^{
             [self loadMainUI];
-            [self showOnboardingIfNeededWithCompletion:^(BOOL didShowOnboarding){
+            [self presentOnboardingIfNeededWithCompletion:^(BOOL didShowOnboarding){
                 [self hideSplashViewAnimated:!didShowOnboarding];
             }];
         }];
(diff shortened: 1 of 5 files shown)


Wikipedia/View Controllers/Nearby/NearbyViewController.m
```diff
@@ -122,12 +113,50 @@ - (BOOL)prefersStatusBarHidden {
...
-    [self.locationManager stopUpdatingLocation];
+- (void)viewWillAppear:(BOOL)animated {
+    [super viewWillAppear:animated];
+
+    self.skipStartUpdates = NO;
+    // Only begin updates if by the time the startUpdates selector below is performed
+    // self.skipStartUpdates is still NO - lets us add the nearby view controller
+    // somewhere and immediately cover it up without it starting to update locations
+    // and headings.
+    [self performSelector:@selector(startUpdates) withObject:nil afterDelay:0.5];
```
montehurd: This is a bit of a hack, but I kind of prefer this solution as it means you can still just add a nearby view controller and it *just works* without having to conditionally call some public "start working" method. And you can add a nearby view controller and cover it immediately without it throwing up a "app wants location data access" permissions popup over the top of the covering view controller's view.

montehurd:
Note: I haven't tested this with Data Migration yet!

Wikipedia/UI-V5/WMFAppViewController.m
```diff
@@ -142,14 +155,24 @@ - (WMFArticleListCollectionViewController*)recentArticlesViewController {
...
+
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        [self showSplashView];
+
+        [self runDataMigrationIfNeededWithCompletion:^{
+            [self loadMainUI];
+            if ([self shouldShowOnboarding]) {
+                [self showOnboardingWithCompletion:^{
+                    [self hideSplashViewAnimated:NO];
```
montehurd: Calling hideSplashViewAnimation *after* the onboarding is shown fixes the visual gitch that was stuttering the transition between the splash globe and the globe in the onboarding storyboard.

Wikipedia/UI-V5/WMFAppViewController.m
```diff
@@ -142,14 +155,24 @@ - (WMFArticleListCollectionViewController*)recentArticlesViewController {
...
+                }];
+            } else {
+                [self hideSplashViewAnimated:YES];
```
montehurd: In the case onboarding is *not* being shown, just do the animation (makes the globe expand and fade out).

Wikipedia/View Controllers/WebView/WebViewController.m
```diff
@@ -91,10 +91,6 @@ - (BOOL)prefersStatusBarHidden {
     return NO;
 }
```
montehurd: REMOVE ALL THE EVERYTHINGS!!!

Wikipedia/View Controllers/WebView/WebViewController.m
```diff
@@ -91,10 +91,6 @@ - (BOOL)prefersStatusBarHidden {
...
 
-- (BOOL)prefersTopNavigationHidden {
-    return [self shouldShowOnboarding];
-}
-
```
montehurd: Removing anything from the web vc makes me happy :)

montehurd:
Just tested with data migration - seems to work fine. But if you can please test with migration again.

Wikipedia/UI-V5/WMFAppViewController.m
coreyfloyd: To simplify the high level flow, I would take the approach of `runDataMigrationIfNeededWithCompletion:`method.

If you do `showOnboardingIfNeededWithCompletion` than the if/else statement becomes unnecessary here. It just makes the code more readable here (and pushes the if/else inside of the `showOnboardingIfNeededWithCompletion method`)

Wikipedia/View Controllers/Nearby/NearbyViewController.m
coreyfloyd: makes sense - at some point we will have some UI to allow users to opt in - just not yet

coreyfloyd:
@montehurd looking good… left one comment to simplify the high level logic, but otherwise i think we are good to go.

Wikipedia/View Controllers/Nearby/NearbyViewController.m
```diff
@@ -122,12 +113,50 @@ - (BOOL)prefersStatusBarHidden {
...
+}
+
+- (void)startUpdates {
+    if (!self.skipStartUpdates) {
+        static dispatch_once_t onceToken;
+        dispatch_once(&onceToken, ^{
+            // Needed by iOS 8.
+            SEL selector = NSSelectorFromString(@"requestWhenInUseAuthorization");
+            if ([self.locationManager respondsToSelector:selector]) {
+                NSInvocation* invocation =
```
bgerstle: just call the selector directly instead of using `NSInvocation`

Wikipedia/View Controllers/Nearby/NearbyViewController.m
bgerstle: I dislike this approach because it goes against two best practices:

- Keeping state to a minimum
- Avoiding delays if possible (i.e. race condition vulnerabilities)

This one of the things I was hoping to prevent by having data migration (and onboarding) short-circuit the application by preventing the main view controller from even being instantiated.  If we did that, we wouldn't need any of this logic, and everything can behave normally w/o worrying about some external state transition.  We might be able to work around it here, but there will be (if there aren't already) other things going on "behind the scenes" while we show on-boarding (or worse, data migration) that we'd be better off without.

Since this was an issue before this ticket, and this ticket was only about moving on-boarding out of webVC, we have the option to implement what I described above in a separate ticket.  Let me know if you'd prefer a separate ticket, and I can create one w/ acceptance criteria along these lines.

coreyfloyd: Oh - I didn't notice the perform selector delay with delay here. We really shouldn't be doing this. 

I'm not exactly clear on what this is circumventing, but my hunch here is that the issue is because we are not putting this logic in viewDidAppear with the rest of the loading logic. 

Moving this there should would solve this issue. The whole point of having the methods with completion call backs is so we can control exactly when events happen. If we need to use a delay, that means we aren't using the completion blocks correctly. 

We should definitely resolve this before merging. This should totally be solvable without making any architecture changes.

Wikipedia/View Controllers/Nearby/NearbyViewController.m
coreyfloyd: Yeah - using NSInvocation is almost never needed in day to day coding. It is a special case for dynamic work when you don't know the classes involved (maybe they don't even exist when the app is compiled). That should not be the case here.

coreyfloyd:
@montehurd sorry - I missed the stuff the @bgerstle found. I made notes above, but both should be rather quick fixes.

Wikipedia/View Controllers/Nearby/NearbyViewController.m
montehurd: Oops ya this was pre-existing code. Will update...

montehurd: Wow not sure where I got that old code for doing this... but wow... sure was silly :)

Wikipedia/UI-V5/WMFAppViewController.m
montehurd: Fixed!

Wikipedia/View Controllers/Nearby/NearbyViewController.m
montehurd: Removed this hackiness! :)

coreyfloyd:
:+1:

bgerstle:
:+1:

PR #61, opened 2015-08-13
Onboarding follow on

coreyfloyd:
I asked @montehurd to follow my completion block example without understanding that the on boarding does not call back when it is Complete. Instead it calls back when the presentation is complete, which is not useful for flow control.

Because of this we do not need a completion block to know when the on boarding is presented, we can just return a bool synchronously. This commit does just that.

This was my fault for not understanding the on boarding flow so correcting it here. If we do want to report back when onboarding is complete, we can re-introduce a completion block for this then.

### DIFF
--- a/Wikipedia/UI-V5/WMFAppViewController.m
+++ b/Wikipedia/UI-V5/WMFAppViewController.m
@@ -164,9 +164,8 @@ typedef NS_ENUM (NSUInteger, WMFAppTabType) {
 
         [self runDataMigrationIfNeededWithCompletion:^{
             [self loadMainUI];
-            [self presentOnboardingIfNeededWithCompletion:^(BOOL didShowOnboarding){
-                [self hideSplashViewAnimated:!didShowOnboarding];
-            }];
+            BOOL didShowOnboarding = [self presentOnboardingIfNeeded];
+            [self hideSplashViewAnimated:!didShowOnboarding];
         }];
     });
 }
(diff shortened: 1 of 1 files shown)


bgerstle:
@coreyfloyd i was talking to @montehurd about this as well, and I think the completion block is still useful.  Here's how I understand the flow currently:

1. App launch
2. Show splash
3. If we haven't shown on-boarding
  1. Present on-boarding w/ an animation (which does a seamless transition from the splash screen)
  2. After on-boarding is **presented** dismiss the splash screen _without animating_
  3. Go to (5)
4. Else, dismiss splash screen _with animation_
5. Main app UI is revealed

As such, having the completion block fire after the on-boarding is **done being presented**, we ensure that we don't hide the splash screen until it is out of view (i.e. obscured by on-boarding).  OTOH, when on-boarding is dismissed, the splash screen **should already be dismissed** so as to reveal the main app UI.

All of that said, the `completion` semantics are ambiguous, as it means one thing for migration (i.e. when migration has finished running) and on-boarding (i.e. when on-boarding is done being presented).  We either need to further disambiguate these terms and/or provide doc-comments describing the behavior.  In either case, I think a synchronous return is not appropriate for implementing the above execution flow (whereby the splash screen is dismissed after the on-boarding is presented).

coreyfloyd:
@bgerstle The application flow works now without a completion block and without any visual glitches.

Can you demonstrate a visual glitch or other bug that only happens when the completion block is removed?

bgerstle:
oh well, in that case. :+1:

PR #63, opened 2015-08-13
Make article equality check title, instead of just site

bgerstle:
Not checking article titles during equality checks was causing issues with partially-populated articles (e.g. filled with data from a search result).  Long-term, this would be a non-issue since we plan to use separate model objects for those responses, but it was good that it surfaced this oversight.


bgerstle:
@brion would you mind clarifying your rationale behind omitting titles from equality checks in `MWKArticle`? I can't think of any cases where considering articles with unequal titles would be desirable, but I'm not a battle-hardened MW veteran yet ;-).

brion:
Must've been an oversight, over reliance on unit tests in the coding. ;)

coreyfloyd:
:+1: looks good - just waiting on travis

coreyfloyd:
just fixed a project file conflict… merging without waiting for new tests to run (living on the edge)

PR #64, opened 2015-08-14
Style updates for native article card.

montehurd:
https://phabricator.wikimedia.org/T107803

- [x] adjust line spacing
- [x] adjust padding
- [x] move header text down
- [x] fix tests
- [x] ensure rtl text is still rtl
- [x] margin around title and description
- [x] collapse description (so there's no extra padding beneath title) if no description
- [x] title and description drop shadow
- [x] truncate title and description to max of 3 lines each
- [x] increased height of top native lead image to 250
- [x] increased height of cards (in stack) to 250 to match increased lead image height
- [x] hide summary header


WikipediaUnitTests/WMFArticleSummaryVisualTests.m
```diff
@@ -72,17 +78,15 @@ - (void)verifySummaryForFixtureData:(NSDictionary*)mobileViewJSON langCode:(NSSt
...
-
-    // attributed text cell need a superview else it won't layout its subviews
-    UIView* testView = [UIView new];
-    [testView addSubview:testCell];
-    testView.frame = (CGRect){
-        .origin = CGPointZero,
-        .size   = [testCell.attributedTextContextView suggestedFrameSizeToFitEntireStringConstraintedToWidth:320]
-    };
-    testCell.frame = testView.bounds;
+    WMFMinimalArticleContentCell* testCell = [self.articleVC.tableView dequeueReusableCellWithIdentifier:[WMFMinimalArticleContentCell wmf_nibName]];
```
bgerstle: ah, clever, didn't think about grabbing the VC's cell directly. this is a much more "accurate" test

Wikipedia/UI-V5/WMFMinimalArticleContentCell.m
```diff
@@ -2,65 +2,25 @@
...
-}
-
-@end
-
 
 @interface WMFMinimalArticleContentCell ()
-<DTAttributedTextContentViewDelegate>
 
-@property (nonatomic, strong) WMFLinkButtonFactory* buttonFactory;
+@property (nonatomic, strong) IBOutlet UITextView* textView;
```
bgerstle: not sure if this is a "best practice" or not, but I usually make subviews `weak` since they're retained by their superview

Wikipedia/UI-V5/WMFMinimalArticleContentCell.m
```diff
@@ -2,65 +2,25 @@
...
     if (self) {
-        self.attributedTextContextView.shouldDrawImages = NO;
-        self.attributedTextContextView.shouldDrawLinks  = YES;
-        self.hasFixedRowHeight                          = NO;
-        self.textDelegate                               = self.buttonFactory;
-        self.attributedTextContextView.edgeInsets       = UIEdgeInsetsMake(10.f, 10.f, 10.f, 10.f);
+        self.tintColor = [UIColor wmf_logoBlue];
     }
     return self;
 }
```
bgerstle: missing override of `prepareForReuse`, where we should empty `textView`.

WikipediaUnitTests/WMFArticleSummaryVisualTests.m
```diff
@@ -72,17 +78,15 @@ - (void)verifySummaryForFixtureData:(NSDictionary*)mobileViewJSON langCode:(NSSt
...
+
+    [testCell setAttributedText:self.article.summaryHTML];
+
+    CGSize preHeightAdjustmentSize = (CGSize){320, 100};
+
+    CGSize heightAdjustedSize = [testCell.contentView systemLayoutSizeFittingSize:preHeightAdjustmentSize withHorizontalFittingPriority:UILayoutPriorityRequired verticalFittingPriority:UILayoutPriorityFittingSizeLevel];
+
+    testCell.frame = (CGRect){{0, 0}, heightAdjustedSize};
```
bgerstle: FYI you can also use `CGPointZero` as shorthand for `{0,0}`

Wikipedia/UI-V5/WMFMinimalArticleContentCell.m
montehurd: Ah good point.

Wikipedia/UI-V5/WMFMinimalArticleContentCell.m
montehurd: Will do.

Wikipedia/Categories/NSAttributedString+WMFHTMLForSite.m
```diff
@@ -24,18 +25,27 @@ + (NSDictionary*)wmf_defaultHTMLOptionsForSite:(MWKSite*)site {
...
+    NSAttributedString* attrStr = [self initWithHTMLData:data
+                                                 options:[[self class] wmf_defaultHTMLOptionsForSite:site]
+                                      documentAttributes:nil];
+    attrStr = [attrStr wmf_attributedStringWithParagraphStylesAdjustments:^(NSMutableParagraphStyle* paragraphStyle){
+        /*
+           Needed because if you try adjust line spacing with DTDefaultLineHeightMultiplier
+           anything larger than 1.0 ends up adding a bunch of padding before the first
+           paragraph of text.
+         */
+        paragraphStyle.lineSpacing = 12;
```
bgerstle: is this necessary now that we're using `UITextView`?

WikipediaUnitTests/WMFArticleSummaryVisualTests.m
montehurd: I always forget that! :)

MediaWikiKit/MediaWikiKit/MWKArticle.m
```diff
@@ -570,9 +570,21 @@ - (NSAttributedString*)summaryHTML {
         if (!nonEmptyParagraphsWithSelectedChildren.count) {
             continue;
         }
-        NSString* lineSeparatedParagraphs = [nonEmptyParagraphsWithSelectedChildren componentsJoinedByString:@"<br/><br/>"];
-        NSData* xpathData                 = [lineSeparatedParagraphs dataUsingEncoding:NSUTF8StringEncoding];
-        return [[NSAttributedString alloc] initWithHTMLData:xpathData site:self.site];
+        NSString* lineSeparatedParagraphs = [nonEmptyParagraphsWithSelectedChildren componentsJoinedByString:@"</p><p>"];
```
bgerstle: this is and the next line could be replace by a map which wraps in paragraph in `<p></p>`. makes the intent clearer, IMHO. also, I'd do the wrapping with array joining instead of string formatting (which I'm guessing would be a hair faster): `[@["<p>", paragraphChildren, "</p>"] componentsSeparatedByString:@""]`.

Also, this would be much simpler written in Swift.  Just sayin :wink: 

``` swift
nonEmptyParagraphsWithSelectedChildren.map() { "<p>\($0)</p>" }
```

MediaWikiKit/MediaWikiKit/MWKArticle.m
```diff
@@ -570,9 +570,21 @@ - (NSAttributedString*)summaryHTML {
...
+        lineSeparatedParagraphs = [NSString stringWithFormat:@"<p>%@</p>", lineSeparatedParagraphs];
+        NSData* xpathData = [lineSeparatedParagraphs dataUsingEncoding:NSUTF8StringEncoding];
+
+        NSAttributedString* returnValue = [[NSAttributedString alloc] initWithHTMLData:xpathData site:self.site];
+
+        // Remove trailing line break. Prob move this to a category.
+        if (returnValue.length) {
```
bgerstle: there's already a method for this: `stringByTrimmingCharactersInSet:(NSCharacterSet*)` which returns:
> A new string made by removing from both ends of the receiver characters contained in set. If the receiver is composed entirely of characters from set, the empty string is returned.

bgerstle: also, shouldn't we do this on the string _before_ building the attributed string?

Wikipedia/Categories/NSAttributedString+WMFHTMLForSite.m
montehurd: Ya the comment on line 42 sums it up. This way we get an attributed string with no surprise padding which lets us control padding around the text via constraint constants in a predictable way.

Wikipedia/Categories/NSAttributedString+WMFModifyParagraphs.h
```diff
@@ -0,0 +1,20 @@
+//  Created by Monte Hurd on 8/17/15.
+//  Copyright (c) 2015 Wikimedia Foundation. Provided under MIT-style license; please copy and modify!
+
+#import <Foundation/Foundation.h>
+
+@interface NSAttributedString (WMFModifyParagraphs)
+
+/**
+ *  Sometimes you need to adjust all paragraph styles without destroying existing paragraph style properties.
```
bgerstle: comment styling nit-pick: doc comments should have 1 brief line describing the method's purpose, then if needed, a longer paragraph with details & usage examples. For example:

```
/**
 * A method that does a thing
 * 
 * Useful for doing stuff, with things. 
 *
 ... params & return comments, etc.
```

Wikipedia/Categories/NSAttributedString+WMFModifyParagraphs.m
```diff
@@ -0,0 +1,25 @@
...
+//  Copyright (c) 2015 Wikimedia Foundation. Provided under MIT-style license; please copy and modify!
+
+#import "NSAttributedString+WMFModifyParagraphs.h"
+
+@implementation NSAttributedString (WMFModifyParagraphs)
+
+- (NSAttributedString*)wmf_attributedStringWithParagraphStylesAdjustments:(void (^)(NSMutableParagraphStyle* existingParagraphStyle))block {
+    if (!block) {
+        return self;
+    } else {
```
bgerstle: style nit-pick: if you return early from a method, omit the `else` clause to remove unnecessary indentation

Wikipedia/Categories/NSAttributedString+WMFModifyParagraphs.m
```diff
@@ -0,0 +1,25 @@
...
+        NSMutableAttributedString* mutableCopy = self.mutableCopy;
+        [self enumerateAttribute:NSParagraphStyleAttributeName
```
bgerstle: I'm a bit wary of this approach, as we're always adding new constraints to a copy of the receiver, meaning you could get a buildup of attributes if this is called repeatedly.  it would be safer, i think, to recreate an "empty" attributed string w/ the receiver's `string`, then enumerate the receiver's attributes, passing them through a modifier block before setting on the new attributed string.  this way, you always have the same (or less) attributes than you started with, just with different values.

MediaWikiKit/MediaWikiKit/MWKArticle.m
montehurd: Oh ya, will do the map...

MediaWikiKit/MediaWikiKit/MWKArticle.m
montehurd: Will check...

Wikipedia/Categories/NSAttributedString+WMFModifyParagraphs.h
montehurd: Ah. K.

Wikipedia/Categories/NSAttributedString+WMFModifyParagraphs.m
montehurd: Haha oops yeah that's better.

Wikipedia/Categories/NSAttributedString+WMFModifyParagraphs.m
montehurd: I thought this too, but it looks because we're changing the paragraph for the exact same ranges, it actually replaces the old paragraph objs for those ranges. I confirmed by setting a breakpoint and looking at the string before and after. Want to double check? Or do you think I should add a test to confirm?

Wikipedia/Categories/NSAttributedString+WMFModifyParagraphs.h
coreyfloyd: @montehurd if you install the VVDocumenter Plugin via alcatra, typing "///" will create the code block for you with all the boiler plate filled out.

https://github.com/onevcat/vvdocumenter-xcode

montehurd: Already have it, just forgot it here.

MediaWikiKit/MediaWikiKit/MWKArticle.m
montehurd: Done.

MediaWikiKit/MediaWikiKit/MWKArticle.m
montehurd: So trim doesn't work on attributed strings, and it looks like the trailing line break is coming from DTCoreText. So I moved the trailing break stripping to NSAttributedString+WMFHTMLForSite.m

montehurd:
I especially like the scrolling idea if title/descrip combo is overly long. Really easy w/ UIScrollView I think.

montehurd:
For future reference, if we're targeting 8 and up we can use:
    - (CGSize)systemLayoutSizeFittingSize:(CGSize)targetSize withHorizontalFittingPriority:(UILayoutPriority)horizontalFittingPriority verticalFittingPriority:(UILayoutPriority)verticalFittingPriority NS_AVAILABLE_IOS(8_0);

Example:

    UILabel* label = [[UILabel alloc] init];
    label.lineBreakMode   = NSLineBreakByWordWrapping;
    label.numberOfLines   = 0;
    label.backgroundColor = [UIColor whiteColor];
    label.attributedText  = *some really long attr string*;

    CGSize preHeightAdjustmentSize = (CGSize){320, 100};

    CGSize heightAdjustedSize = [label systemLayoutSizeFittingSize:preHeightAdjustmentSize withHorizontalFittingPriority:UILayoutPriorityRequired verticalFittingPriority:UILayoutPriorityFittingSizeLevel];

    label.frame = (CGRect){CGPointZero, heightAdjustedSize};

montehurd:
Screenshots of title and description being truncated after 3 lines (from last commit):
screen shot 2015-08-19 at 4 30 40 pm
screen shot 2015-08-19 at 4 32 37 pm

montehurd:
@bgerstle hey just noticed the image on our "lopado" screenshots is black instead of the placeholder - is that a known issue?

bgerstle:
have you noticed that on any other pages? might be an anomaly w/ that page.

montehurd:
@bgerstle Hmm not sure - but I'll let you know if I see another one.

bgerstle:
@montehurd so, I just dropped a bunch of comments on your great start to standardizing `UILabel` testing. we don't need to do anything about it now, just wanted to jot my thoughts down while they were fresh. we can revisit as we add more tests & improve the API to make it more flexible.

bgerstle:
thanks for the screencaps @montehurd looks real nice! 😭 so beautiful!

bgerstle:
@montehurd looks great, i'm ready to merge! were you still planning to remove summary section header or was that a follow-on?

coreyfloyd:
@bgerstle yeah that needs to get done first… we can then put out a build with these changes for design.

coreyfloyd:
I'm good here… I'll leave @bgerstle to approve the feedback changes he gave.
:+1:

bgerstle:
:+1: muchas gracies señor @montehurd!

PR #65, opened 2015-08-14
[WIP] Demo to wire up DTCoreText NSAtrributedStrings to native views.

coreyfloyd:
So after the talk - I whipped this up in 20 minutes. 

This gets all the auto layout for free and has "native-looking" text rendering. It also supports all the links. And it is just a plain UITextView.

Like I said earlier… The view parts of DTCoreText were made before we could display NSAttributedStrings in standard UIControls. I think the HTML->AttributedString conversion stuff is really all we need for iOS 7 and up.


montehurd:
Whoa! No way!

montehurd:
I used this to modify my PR at https://github.com/wikimedia/wikipedia-ios/pull/64 - which I will update shortly.

PR #66, opened 2015-08-14
fix image request cancellation

bgerstle:
lots of pieces here to fix:

- forgot to add cancellable to our cancellables map in the image
  controller
- SDWebImage doesn't report cancellation! added as an option
- SDImageCache doesn't report errors! observe operation and reject
  promise if cancelled
- remove now-unused CancellablePromise & WrapperObject

And now it's tested! 👷

Tested by [searching "Claude Monet"](http://cl.ly/image/093R0c3e3S0g/Screen%20Recording%202015-08-14%20at%2006.58%20PM.gif) (almost all search results have images). The GIF is rather long, so can't be embedded here.


coreyfloyd:
@bgerstle getting failures… ran locally and it passed. Not sure if there is something we need to tell travis?

bgerstle:
Probably a race condition. Will look later

PR #67, opened 2015-08-14
Migrating existing data source to use SSDataSources

coreyfloyd:
This is prework for the home screen - just to make sure integration is not onerous.

Also this should resolve this bug:
https://phabricator.wikimedia.org/T109027


bgerstle:
:bug: crash when deleting an item from "Recent": https://www.irccloud.com/pastebin/r1RCzxxi/

bgerstle:
First encountered by:

1. Searching "Claude Monet"
2. Visiting a few results
3. Saving a few results
4. Restarting, searching for "Claude Monet" again
5. Visiting a couple new results
6. Go to "Saved"
7. Unsave something ✅
8. Go to "Recent"
9. Swipe to delete last item ❌ crash

Will try to reproduce more simply now.

bgerstle:
Minimal repro steps:

1. Clean install
2. Launch the app
3. Search for something
4. Tap first result
5. Go to "Recent"
6. Swipe to delete the only entry ❌ crash

Also tried searching & tapping first two results in succession, then deleting the second from "Recent," resulted in same crash.  Seems independent of what was searched or how many items are in the list.

coreyfloyd:
@bgerstle crash should be fixed

bgerstle:
yay no crash! :+1:. @montehurd will you have a look and merge if you don't find anything wrong?

PR #69, opened 2015-08-17
Fix image flickering issue caused by unnecessary dispatch & add cross-dissolve

bgerstle:
Bug: [T108053](https://phabricator.wikimedia.org/T108053)

First, I tried only animating the transition when the image wasn't retrieved from memory.  Still occasionally saw "flicker" when scrolling quickly over images that should've been in memory.  As a result, I ended up exposing `WMFImageController.cachedImageInMemoryWithURL(NSURL)` to get images from memory.  This allows the image to be set before returning from `collectionView:cellForItemAtIndexPath:`, preventing the cell from appearing image-less when the image is already in memory.

Check out [this sweet demo](https://s3.amazonaws.com/f.cl.ly/items/1g070k2w0l131g1M162b/Screen%20Recording%202015-08-17%20at%2004.22%20PM.gif).  Caveat: this actually demonstrates that the "wrong image" bug is still present. Fix for that to follow shortly. _Hint: it's because canceling image fetches doesn't also cancel face detection it has already started_. **UPDATE:** now fixed, see below.

#### `UIImageView+MWKImage`
Set an image on a `UIImageView` in one line of code (a la SDWebImage)!
``` objc
// from the article header gallery
[cell.imageView wmf_setImageWithFaceDetectionFromMetadata:self.images[indexPath.item]];
```

That one call is sugar on top of:
``` objc
/**
 *  Set the receiver's image to the source URL of the given `imageMetadata`.
 *
 *  Does any necessary fetching or processing to retrieve the image and detect/center faces. Will set the image
 *  synchronously if possible.  This method is both idempotent and cancelling in that setting the same `imageMetadata`
 *  repeatedly is a no-op after the first call, and setting a different `imageMetadata` or `nil` will prevent the
 *  previous metadata's image from being set.
 *
 *  @param imageMetadata Metadata with the `sourceURL` of the image you want to set.
 *  @param options       Control how the image is set by specifying a bitmask of `WMFImageOptions`.
 *  @param setImageBlock Called when the image is going to be set on the receiver. Default simply sets the image on the receiver.
 *  @param completion    Called after the image has been set (i.e. after any fetching, processing, and/or animations).
 *  @param failure       Called if the image could not be retrieved from the network.
 */
- (void)wmf_setImageFromMetadata:(MWKImage* __nullable)imageMetadata
                         options:(WMFImageOptions)options
                       withBlock:(WMFSetImageBlock __nullable)setImageBlock
                      completion:(dispatch_block_t __nullable)completion
                         onError:(void (^ __nullable )(NSError*))failure;
```

and here are the current options:

``` objc
typedef NS_ENUM (NSInteger, WMFImageOptions) {
    /**
     *  Run face detection (if needed) and center the face inside the receiver.
     */
    WMFImageOptionCenterFace = 1 << 0,

        /**
         *  Animate regardless of whether or not the image was retrieved from memory.
         *
         *  @warning Do not set this flag in combination with `WMFImageOptionNeverAnimate`.
         *
         *  @see WMFImageOptionCrossDissolve
         */
        WMFImageOptionAlwaysAnimate = 1 << 2,

        /**
         *  Do not animate when setting the new image.
         *
         *  @warning Do not set this flag in combination with `WMFImageOptionAlwaysAnimate`.
         *
         *  @see WMFImageOptionCrossDissolve
         */
        WMFImageOptionNeverAnimate = 1 << 3,
};
```

Which should satisfy all you image-setting needs.  Since it knows about `MWKImage` we can even add an `WMFImageOption` for fetching a low-res placeholder.  Additionally, my hope is that we can eventually encapsulate error handling (show an error image in here if the fetch fails) and loading UI (show a spinner if it's not loaded in X ms).


Wikipedia/Images/WMFImageDownload.swift
```diff
@@ -12,6 +12,7 @@ public enum ImageOrigin: String {
     case Network = "Network"
     case Disk = "Disk"
     case Memory = "Memory"
+    case None = "None"
```
coreyfloyd: Isn't case none nonsensical?
Are you passing an imageOrigin even if no image is returned? Shouldn't it just be an error with no origin?

Wikipedia/SDImageCache+PromiseKit.swift
```diff
@@ -20,23 +21,25 @@ extension SDImageCache {
...
-                                   keyPath: "isCancelled",
-                                   options: NSKeyValueObservingOptions.New)
-        { _, _, change in
-            let value = change[NSKeyValueChangeNewKey] as! NSNumber
-            if value.boolValue {
-                reject(NSError(domain: NSURLErrorDomain, code: NSURLErrorCancelled, userInfo: nil))
+        if let diskQueryOperation = diskQueryOperation {
+            // make sure promise is rejected if the operation is cancelled
+            self.KVOController.observe(diskQueryOperation,
+                keyPath: "isCancelled",
```
coreyfloyd: Didn't you have to update SDWebImage to call back when an operations cancelled? Why do you still need to observe isCancelled as well?

Wikipedia/SDImageCache+PromiseKit.swift
```diff
@@ -20,23 +21,25 @@ extension SDImageCache {
...
-        self.KVOController.observe(diskQueryOperation,
-                                   keyPath: "isFinished",
-                                   options: NSKeyValueObservingOptions.allZeros)
-        { [weak self] _, object, _ in
-            self?.KVOController.unobserve(object)
+            // tear down observation when operation finishes
+            self.KVOController.observe(diskQueryOperation,
+                keyPath: "isFinished",
+                options: NSKeyValueObservingOptions.allZeros)
+                { [weak self] _, object, _ in
```
coreyfloyd: Does cancelling also flip isFinished to YES as well? I forget.

Wikipedia/UI-V5/WMFArticleHeaderImageGalleryViewController.m
```diff
@@ -95,45 +98,141 @@ - (UICollectionViewCell*)collectionView:(UICollectionView*)collectionView
...
     @weakify(self);
     [[WMFImageController sharedInstance] fetchImageWithURL:[imageMetadata sourceURL]]
-    .then(^id (UIImage* image) {
+    .then(^id (WMFImageDownload* download) {
+        @strongify(self);
+        UIImage* image = download.image;
         if (!self) {
             return [NSError cancelledError];
+        }
+        BOOL shouldAnimate = ![download.origin isEqualToString:[WMFImageDownload imageOriginMemory]];
```
coreyfloyd: Can this BOOL be scoped to the where it is used?

Wikipedia/UI-V5/WMFArticleHeaderImageGalleryViewController.m
```diff
@@ -95,45 +98,141 @@ - (UICollectionViewCell*)collectionView:(UICollectionView*)collectionView
...
-            } else {
-                [self setImage:image centeringBounds:imageMetadata.firstFaceBounds forCellAtIndexPath:indexPath];
-                return nil;
-            }
+            DDLogVerbose(@"Setting image %@ after retrieving from %@", imageMetadata.sourceURL, download.origin);
+            [self setImage:image
+             inCellAtIndexPath:indexPath
+               centeringBounds:imageMetadata.firstFaceBounds
+                      animated:shouldAnimate];
+            return nil;
```
coreyfloyd: punt!

Wikipedia/UI-V5/WMFArticleHeaderImageGalleryViewController.m
```diff
@@ -95,45 +98,141 @@ - (UICollectionViewCell*)collectionView:(UICollectionView*)collectionView
...
         }
+    })
+    .catch(^(NSError* error) {
+        // TODO: show error in UI
```
coreyfloyd: punt!

Wikipedia/Images/WMFImageDownload.swift
bgerstle: Ah right, was using it at one point but then realized I didn't, but forgot to remove it. Will do.

Wikipedia/SDImageCache+PromiseKit.swift
bgerstle: The change I made to SDWebImage reports cancellations as errors _when using the `SDWebImageManager`_. This, however, is straight from the cache, which `WMFImageController` uses to async cache queries.  In order for this to report cancellation properly, I _have_ to monitor the operation.  We definitely have the option of factoring this into a category or something and using that approach instead of modifying SDWebImage, but I prefer to not rely on KVO on an `NSOperation` which is _really_ easy to get wrong in custom subclasses.

Wikipedia/SDImageCache+PromiseKit.swift
bgerstle: I know it does in this case, because otherwise we'd get a KVO exception due to the operation being deallocated while observers are attached to it.  I say this, because I ran into it :disappointed:.

Wikipedia/UI-V5/WMFArticleHeaderImageGalleryViewController.m
```diff
@@ -95,45 +98,141 @@ - (UICollectionViewCell*)collectionView:(UICollectionView*)collectionView
...
+ *  @param indexPath              The indexPath for the cell to update.
+ *  @param normalizedCenterBounds The bounds to center in the image view.
+ *  @param animated               Whether or not to animate the transition to `image`.
+ *
+ *  @see setImage:inCell:centeringBounds:animated:
+ */
+- (void)     setImage:(UIImage*)image
+    inCellAtIndexPath:(NSIndexPath*)indexPath
+      centeringBounds:(CGRect)normalizedCenterBounds
+             animated:(BOOL)animated {
```
coreyfloyd: Done really need 3 versions of this method - its kinda confusing. 

For instance, this one is only getting the cell and then calling through to the main method - so its not really saving any typing here. Why not just call these 2 lines where you need them instead of creating another similarly named method?

Wikipedia/UI-V5/WMFArticleHeaderImageGalleryViewController.m
bgerstle: 🏈

bgerstle: at least I added a catch statement! wasn't even there before :stuck_out_tongue:

Wikipedia/UI-V5/WMFArticleHeaderImageGalleryViewController.m
bgerstle: this is the lowest-common-scope where it's used. Would prefer not to re-check in both places, DRY and all that.

Wikipedia/UI-V5/WMFArticleHeaderImageGalleryViewController.m
```diff
@@ -95,45 +98,141 @@ - (UICollectionViewCell*)collectionView:(UICollectionView*)collectionView
         cell.imageView.contentMode = UIViewContentModeScaleAspectFill;
         return cell;
     }
+
     MWKImage* imageMetadata = self.images[indexPath.item];
+
+    // try to set cell's image from memory cache, otherwise fetch it and do face detection if needed
+    if (![self setCachedImageForCell:cell atIndexPath:indexPath metadata:imageMetadata]) {
+        [self fetchImage:imageMetadata forCellAtIndexPath:indexPath];
```
coreyfloyd: But `[self setCachedImageForCell:cell atIndexPath:indexPath metadata:imageMetadata]` could be returning NO because the Face detection didn't run - so in that case, you should run face detection, not fetch the image right?

bgerstle: Perhaps, but then what do you do if fact detection has run, but the image isn't in memory? Simpler to have one idempotent code path. This is one of the reasons I tried to avoid a separate fast path for in-memory, face-detected images, but it was shown that the optimization was worth it, so here we are :smile:.

Wikipedia/UI-V5/WMFArticleHeaderImageGalleryViewController.m
bgerstle: I guess the one that tries to retrieve the cell from the collection view is sort of unnecessary, but there definitely needs to be one with the cell, and one without.  Retrieving the cell from within `cellForItemAtIndexPath:` will miss stuff up—I tried.

Wikipedia/UI-V5/WMFArticleHeaderImageGalleryViewController.m
```diff
@@ -95,45 +98,141 @@ - (UICollectionViewCell*)collectionView:(UICollectionView*)collectionView
...
+/**
+ *  Attempt to populate the given `cell` with the image for `metadata` if it's stored in memory.
+ *
+ *  @param cell      The cell to populate.
+ *  @param indexPath The indexPath where the cell resides.
+ *  @param metadata  The metadata which specifies the image URL and face detection information.
+ *
+ *  @return `YES` if the image was set from memory, `NO` if a fetch or other processing needs to take place.
+ */
+- (BOOL)setCachedImageForCell:(WMFImageCollectionViewCell*)cell
```
coreyfloyd: Similar to: https://github.com/wikimedia/wikipedia-ios/pull/69/files#r37236899 

At first glance, it looks like you are running enough logic here that would necessitate adding another method - like checking if face detection ran.

But when I looked where this method is actually used, it seems that doing the returning `NO` when `didDetectFaces=NO` on line 176 here is causing us to conflate face detection with fetching the image: https://github.com/wikimedia/wikipedia-ios/pull/69/files#r37237100

So if that is removed, we are back to really just 2 method calls in here. 

All of this is to say, that it just seems you can rework 
`- (void)     setImage:(UIImage*)image inCellAtIndexPath:(NSIndexPath*)indexPath centeringBounds:(CGRect)normalizedCenterBounds animated:(BOOL)animated ` 
to encapsulate the caching as well and avoid this intertwining telescoping pattern that is here now.

bgerstle:
To reiterate our hangout convo, @coreyfloyd, I'll take a swing at wrapping up all this logic in a simpler category or object.  If it's too much to be considered in scope for this PR, I'll let you know.

coreyfloyd:
@bgerstle looking good - travis failed looks like the face detection test needs updated

PR #71, opened 2015-08-19
Home VC with Nearby Section 

coreyfloyd:
This pull request is for 2 tickets:
https://phabricator.wikimedia.org/T93327
https://phabricator.wikimedia.org/T104817

These are being done in tandem because it was hard for me to understand the needs of a section in the home screen without actually implementing a new section.

Before adding Nearby, I factored out logic in the Nearby VC into a few objects:
`WMFLocationManager`
`WMFLocationSearchFetcher` (with serializers)

I created a new cell, `WMFHomeNearbyCell`, as well.

Additionally, there are new model classes for Location Based Search Results.

For the Home screen itself, there is the `WMFHomeViewController`
This uses `SSDataSources` to power the collection view… specifically `SSSectionedDataSource`

Instead of putting all logic for each section in the Home VC, I took the approach of creating `WMFHomeSectionController` which is an abstract class for managing a specific section.

The concrete implementation of this is the `WMFNearbySectionController` which takes the `WMFLocationManager`, `WMFLocationSearchFetcher`, and the `SSSectionedDataSource` and uses those objects to manage the section.

What still needs done:
- [x] Present article VC on tap
- [x] Add logic for an "empty" section view
- [x] Move the collection view to pure auto layout (currently using fixed sizes because of some auto layout issue)
- [x] Move compass logic into new Nearby Cell

I'm going to call the above "enough" features for a single Pull request.

See comments below for other functionality coming in future PRs


MediaWikiKit/MediaWikiKit/MWKLocationSearchResult.m
```diff
@@ -0,0 +1,60 @@
...
+        NSDictionary* coords = [value firstObject];
+        NSNumber* distance = coords[@"dist"];
+        return distance;
+    }];
+}
+
+
++ (NSDictionary*)JSONKeyPathsByPropertyKey {
+    return @{
+             @"displayTitle": @"title",
```
bgerstle: 🚨`WMF_SAFE_KEYPATH` please

also, i thought about this earlier today.. need to file an eng backlog task to replace our macro with the original one from libextobjc.

Wikipedia/UI-V5/SSSectionedDataSource+WMFSectionConvenience.m
```diff
@@ -0,0 +1,52 @@
...
+//
+
+#import "SSSectionedDataSource+WMFSectionConvenience.h"
+
+@implementation SSSectionedDataSource (WMFSectionConvenience)
+
+- (NSIndexSet*)indexesOfItemsInSection:(NSInteger)section{
+    NSUInteger numberOfItems = [self numberOfItemsInSection:section];
+    if(numberOfItems == 0)
+        return nil;
```
bgerstle: curly brackets!

coreyfloyd: looks like my uncrustify plugin is down!

Wikipedia/UI-V5/SSSectionedDataSource+WMFSectionConvenience.m
```diff
@@ -0,0 +1,52 @@
...
+    return [NSIndexSet indexSetWithIndexesInRange:NSMakeRange(0, numberOfItems)];
+}
+
+- (NSArray*)indexPathsOfItemsInSection:(NSInteger)section{
+    return [[self indexesOfItemsInSection:section] bk_mapIndex:^id(NSUInteger index) {
+        return [NSIndexPath indexPathForItem:index inSection:section];
+    }];
+}
+
+- (void)replaceItemsWithItems:(NSArray *)items inSection:(NSInteger)section{
```
bgerstle: IMHO a better method name is `setItems:inSection`

Wikipedia/UI-V5/WMFHomeNearbyCell.m
```diff
@@ -0,0 +1,162 @@
+
+#import "WMFHomeNearbyCell.h"
+
+#import "Wikipedia-Swift.h"
+#import "PromiseKit.h"
+
+@interface WMFHomeNearbyCell ()
+
+@property (strong, nonatomic) IBOutlet UIImageView *imageView;
```
bgerstle: i mentioned this in @montehurd's PR, but I usually make my subviews `weak`. I'll flag it for discussion at this week's Code Review

Wikipedia/UI-V5/WMFHomeNearbyCell.m
```diff
@@ -0,0 +1,162 @@
...
+
+@end
+
+@implementation WMFHomeNearbyCell
+
+- (void)prepareForReuse{
+    [super prepareForReuse];
+    [[WMFImageController sharedInstance] cancelFetchForURL:self.imageURL];
+    self.imageView.image = [UIImage imageNamed:@"logo-placeholder-nearby.png"];
+    _imageURL = nil;
```
bgerstle: why aren't you using dot syntax here?

MediaWikiKit/MediaWikiKit/MWKLocationSearchResult.m
coreyfloyd: @bgerstle When you don't have an instance, the only way to use that macro is to alloc a new instance. Hesitant to to do that here… as this logic can be called frequently. In loops that are out of our control.

Additionally, using this macro in "business logic code" that makes sense… you are referencing key paths while trying to accomplish something else. So the macro is nice because they keep you fro tripping up on keypaths while you are working on other features and keep it in sync with refactors. 

But in this case, the entire point of the code is to define key value mappings - its not ancillary to the task - it is the task. I'm ok with knowing that if someone else or my future self changes properties on a Mantle object, they need to know to update the Key Value mappings. 

Even if someone gets it wrong or is somehow forgotten to update it, then Mantle will break and throw errors/exceptions - immediately when running the app. Additionally, Mapping responses is something we can test with Nocilla.

Wikipedia/UI-V5/WMFHomeNearbyCell.m
coreyfloyd: @bgerstle   Xcode's default/guidance for iBOutlets is strong for iOS projects under ARC. Some older documentation may say weak… but that was before iOS 6.

This Stackoverflow answer points to the new guidance as given at WWDC, along with Twitter conversations with Apple Engineers confirming the change and outdated docs: 
http://stackoverflow.com/a/31395938/48311

The relevant WWDC session: http://asciiwwdc.com/2015/sessions/407

Basically, Apple recommends that you use strong unless you have memory semantic issues that would cause you to do otherwise.

cc @montehurd

Wikipedia/UI-V5/WMFHomeNearbyCell.m
coreyfloyd: don't want the side effect. Just like an initializer.

bgerstle:
@coreyfloyd are you able to use the new `UIImageView+MWKImage` category, or are you dealing w/ "raw" URLs? also, do we want face detection on these?

coreyfloyd:
@bgerstle yeah - we don't have MWKImages… I one to use that category, and then realized all I had was a URL… maybe we can update that category to also take a plane URL 

Note: Nothing that comes from a search API will have an MWKImage. This includes Nearby, Read more (on articles and on the home screen), and of course search. So we probably should have an API that handles just URLs… because most of the lists won't have MWKImages.

coreyfloyd:
Sorry for junking up the diff with the pod… thats what I hate most about committing pods. Wish you could specify a folder to "ignore in diff" so it wouldn't show up.

montehurd:
Yay the compass view still scales! (I spent some time making sure it drew in resolution independent manner.)
img_1273

However, It looks like the cell height determination is perhaps not taking into the account the overall height of the cell once it had laid out all it's subviews, but rather it's maybe just cosidering the height of only the labels? See how increasing the size of the compass subview doesn't result in taller cell:
img_1274

montehurd:
@bgerstle @coreyfloyd visual test for compass view scaling would be awesome!

coreyfloyd:
Beyond this there are a couple things to do that are significant, but compartmentalized, and better to approach separately:
- [ ] General Styling of cells / headers
- [ ] Opening the source article of the read more section
- [ ] Opening up a "More like this" VC when taping the more footer
- [ ] A Regular Article Cell (Going to do this with the Read more section work since that is where it is needed)

coreyfloyd:
@montehurd the cell does NOT take into account the height of the compass for its height for calculating the cell size - this is intentional. It only takes into account the text. The text can expand, but the compass is size is fixed (unless you modify it programmatically as you did above).

bgerstle:
@coreyfloyd i can't comment on individual files due to diff size (we should talk about this tomorrow), but one minor thing:

``` objc
NSURL* url = useDeskTopURL ? [self.searchSite apiEndpoint] : [self.searchSite mobileApiEndpoint];
```

This is actually what `MWKSite` does under the hood, just expose `- (NSURL*)apiEndpoint:(BOOL)isMobile` on line 115 to DRY this up.

bgerstle:
@coreyfloyd :heart: SRP serializers & model objects 🌈

bgerstle:
@coreyfloyd i'm not crazy about the `isFetching` mechanism. at the very least, it should check the number of operations on its manager's operation queue instead of maintaining the state itself.  it seems like what you really want is cancellation and/or idempotent requests. we can chat about it tomorrow

bgerstle:
will have a look at the home VC & sectioned data source controller thingies tomorrow 🌄

Wikipedia/UI-V5/WMFHomeNearbyCell.m
```diff
@@ -0,0 +1,175 @@
...
+    UICollectionViewLayoutAttributes* preferredAttributes = [layoutAttributes copy];
+    CGFloat height                                        = MAX(120, self.titleLabel.intrinsicContentSize.height + WMFTextPadding + WMFTextPadding + WMFDistanceHeight + WMFTextPadding);
+    preferredAttributes.size = CGSizeMake(layoutAttributes.size.width, height);
+    return preferredAttributes;
+}
+
+- (void)setImageURL:(NSURL*)imageURL {
+    _imageURL = imageURL;
+    @weakify(self);
+    [[WMFImageController sharedInstance] fetchImageWithURL:imageURL]
```
bgerstle: cancel previous fetch first. also, might as well try to set from memory cache before fetching

Wikipedia/UI-V5/WMFHomeViewController.m
```diff
@@ -0,0 +1,267 @@
...
+    CGFloat width = self.view.bounds.size.width - self.collectionView.contentInset.left - self.collectionView.contentInset.right;
+    return CGSizeMake(width, 50.0);
+}
+
+#pragma mark - UICollectionViewDelegate
+
+- (void)collectionView:(UICollectionView*)collectionView didSelectItemAtIndexPath:(NSIndexPath*)indexPath {
+    id object = [self.dataSource itemAtIndexPath:indexPath];
+
+    //TODO: Casting for now - ned to make a protocol or something
```
bgerstle: FYI I don't' see a to prevent casting w/o XC7 ObjC "generics" or Swift

coreyfloyd: @bgerstle huh?

bgerstle: needing to cast the model object, fughedaboudit

PR #72, opened 2015-08-19
Fix some flaky unit tests

bgerstle:


WikipediaUnitTests/MWKImageFaceDetectionTests.m
```diff
@@ -96,7 +96,7 @@ - (void)testDetectingFacesInFacelessImage {
         .then(^(NSArray* faces) {
             [self.image setNormalizedFaceBoundsFromFeatures:faces inImage:facelessImage];
         });
-    } timeout:WMFDefaultExpectationTimeout WMFExpectFromHere];
+    } timeout:WMFDefaultExpectationTimeout * 2 WMFExpectFromHere];
```
coreyfloyd: should this be:
`WMFDefaultExpectationTimeout * 2 * WMFExpectFromHere`

?

coreyfloyd: Also - should we just make a longer constant?

coreyfloyd: Or i guess I don't understand what `WMFExpectFromHere` is

bgerstle: > should this be:
WMFDefaultExpectationTimeout * 2 * WMFExpectFromHere

Nope, `WMFExpectFromHere` is a macro that passes the current method & line number to the method which is used to generate an expectation description.

> Also - should we just make a longer constant?

Eventually, yes. I'd rather move towards something like:

``` objc
typedef NS_ENUM(NSUInteger, WMFTestTimeout) {
  WMFTestTimeoutShort = 0,
  WMFTestTimeoutDefault,
  WMFTestTimeoutMedium,
  WMFTestTimeoutLong
};
WMFTimeout(NSUInteger timeoutLength) {
  float constant = 0.1;
  if (getenv("TRAVIS") != NULL) {
    constant = 1;
  }
  return powf(2, timeoutLength) * 0.1;
}
```

coreyfloyd: Ok - that Macro is… confusing… especially when it is sitting next to math… and being named like a constant.

We should look to do something to make it more idiomatic and clear in the future.

PR #76, opened 2015-08-21
Related Section on Home VC

coreyfloyd:
Ready for review.

Now have a saved and recent items section in the feed. Not much of "new" this uses the architecture established for the home screen, just with some new sections.

Note: When tapping on an item in the list and showing the Article VC, the article VC hides the nav bar, which breaks navigation. This will be fixed as part of: https://phabricator.wikimedia.org/T109456

Also note, that "More like this" is not functional, this is coming in s follow on task:
https://phabricator.wikimedia.org/T110001

Also note, that images are not doing Face Detection… that will also come in a follow on:
https://phabricator.wikimedia.org/T110002


coreyfloyd:
adding a screen shot for @bgerstle 
screenshot 2015-08-21 17 25 43

coreyfloyd:
I guess I should have labeled that NSFW for @montehurd

montehurd:
If I tap on the Home icon after choosing one of the "because you read" entries, nothing changes...

bgerstle:
Some styling comments:

1. The text isn't aligned to the left of the card. The spec actually shows a small thumbnail here, are we omitting that for now?
2. I understand the white area below the lead image is supposed to have an excerpt from the page's summary, are we also omitting that for now?  Is that something we can even get from search/nearby?
3. Cells don't show any highlighted state while the user is pressing down, can we add that via cell selection style and/or overriding `setSelected:` in the cells to make the appropriate UI elements highlighted (e.g. image, label)

image

bgerstle:
another nitpick "Because **Y**ou **R**ead" isn't capitalized in the spec: "Because **y**ou **r**ead"

coreyfloyd:
@bgerstle Re: styling comments…

1. Done
2. Exactly - need to so some investigation there… if we can tweak search parameters or if we need to do a fetch for each article in the list. Prefer that we not have to do that - will be in a separate ticket.
3. Yeah - not in the design… will need a another ticket from design for cell selection.

PR #77, opened 2015-08-22
Made following links with fragments not underlap top native header.

montehurd:
https://phabricator.wikimedia.org/T108222


montehurd:
@bgerstle I was able to fix the intra-page link issue you identified! All I did is make such links use the same logic. They had formerly been manually setting the scroll position (in JS).

PR #80, opened 2015-08-24
As a user, I would like to see a rich overview of sections in the native card

bgerstle:
Phab: [T108042](https://phabricator.wikimedia.org/T108042)
- [x] Create a cell for each Top level section
- [x] Add title label for the section name (section depth 0)
- [x] Add labels/buttons subsections (section depth 1)
- [x] Do not include section that are greater than depth 1
- [x] Add arrow to the right of title
- [x] Tapping a (sub)section opens the web view to that section

*All image-related tasks & card styling will be done in a separate PRs for Mk 2 & 3*

---

From the "Beach Boys" page on EN Wiki
image

### Hierarchical sections
The biggest change here involves parsing a section hierarchy in order to show the root sections and their immediate descendants (aka "parent" or "top level" and "child" sections).  Check out `MWKSection` and `MWKSectionList` for details.

### Article tableview changes
After many experiments to have each top level section be a self-sizing cell, I ended up making parent & child sections cells within their own section, one for each cell.  For an article with M top level sections, the number of sections N will be M + 2:

- Table section 1
  - Cell 1: Article summary (i.e. article section 0)
- Table Section 2 (TOC starts)
  - Cell 1: Article Section 1
  - Cell 2: Article Section 1.1
  - Cell 3: Article Section 1.2
  - etc.
- Table Section 3
  - Cell: Article Section 2
  - ...
- Table Section N-1
  - Cell: Article Section M
- Table Section N
  - Cell: Read More result 1
  - Cell: Read More result 2
  - ...

#### Easter egg 🐰
While testing this, I started suffering more from a problem I noticed previously, where the table would [quickly jump from the TOC up to the summary](https://www.dropbox.com/s/dkmp4c2nh680e2s/one-size-fits-all-estimate.mov?dl=0) (watch the scroll indicator as well).  I figured this was due to the "one size fits all" `estimatedRowHeight` we were using, and quickly tried to fix the problem by [providing a per-cell-type estimate](https://github.com/wikimedia/wikipedia-ios/commit/694960566602c634fec46a623b36aaae00bc41f5) (i.e. summary, parent section, child section, & read more) and it [seems to work](https://www.dropbox.com/s/2rvjsr37mhrrxmz/per-row-estimate.mov?dl=0) 🎉

### Known issues
Both of these will be fixed when Mk 2 is implemented (card styling)
- [ ] headers are only pinned to the first top-level-section in the TOC. this should be fixed when we use non-sticky headers in the tableview and update spacing/styling
- [ ] similar to above, since the cells in the TOC sections can't have separators in order to appear seamless, the read more cells don't have a separator. this should be fixed when read more results are given the card UI as shown in the specs


MediaWikiKit/MediaWikiKit/MWKSectionList.m
```diff
@@ -140,4 +140,37 @@ - (NSArray*)entries {
     return self.sections;
 }
 
+#pragma mark - Hierarchy
+
+- (void)buildSectionHierarchy {
```
coreyfloyd: Can/Should this be invoked automatically when the sections are set? (Rather than invoking it externally by the article VC)

Wikipedia/UI-V5/WMFArticleSectionCell.m
```diff
@@ -6,37 +6,15 @@
...
-    UIView* selectionColor = [[UIView alloc] init];
-    selectionColor.backgroundColor = color;
-    self.selectedBackgroundView    = selectionColor;
-}
-
-- (void)setLevel:(NSNumber*)level {
-    [self applyIndentForLevel:level];
-}
-
-- (void)applyIndentForLevel:(NSNumber*)level {
```
coreyfloyd: FYI - when you re-indent… there are some properties on UITableViewCell you can use instead of constraints (if you prefer):

```
indentationLevel
indentationWidth
```

bgerstle: no reindentation necessary. not in the spec

bgerstle: but yes, I was wondering why we weren't using those as well

coreyfloyd: Ohhh… that was an old mockup, my bad

coreyfloyd:
@bgerstle looks good to me… only thing I have is the comment about the build sections logic.

MediaWikiKit/MediaWikiKit/MWKSectionList.m
bgerstle: I thought about that... but figured we only need it in one place right now, and can revisit if/when anywhere else needs hierarchical data about sections (YAGNI). I was reluctant to even make `children` a property on `MWKSection`, but this goes with the rest of MWK in WYSINAWYG (what you see is not always what you get)

bgerstle: To directly answer your question: it might be able to be called at the right point(s), but I'm not 100% sure it should be

bgerstle: addressed in ab1d0b3

PR #81, opened 2015-08-25
Added UIWebView category for JS-to-Xcode-console debug messages.

montehurd:
Adds 'console.log( )' global javascript namespace for easy logging
to the Xcode console.

Makes debugging much easier because you are no longer forced to
use Safari to inspect the app's UIWebview dom. You still can
if you need to, but for many small debugging tasks this is
just *much* less cumbersome.


Wikipedia/mw-bridge/CommunicationBridge.m
```diff
@@ -26,6 +27,7 @@ - (CommunicationBridge*)initWithWebView:(UIWebView*)targetWebView;
 
         __weak CommunicationBridge* weakSelf = self;
         [self addListener:@"DOMContentLoaded" withBlock:^(NSString* type, NSDictionary* payload) {
+            [weakSelf.webView wmf_enableJavascriptToXcodeConsoleLogging];
```
coreyfloyd: @montehurd Does this only log to Xcode, or also the standard console as well?
If so, we probably should only enable this for debug/beta builds.
If not, then we can leave it enabled all the time.

bgerstle: agree w/ corey about doing this in debug mode only.

also, why are we introducing a proprietary logging function? let's just replace `console.log` with a native version that reroutes to `DDLog`.

as an aside, i've been thinking we should also add a native `window.onerror` handler so we can (at least in debug mode) throw assertions on JS exceptions so we catch them faster.

bgerstle: for example:

``` objc
JSValue* console = [[self wmf_javascriptContext] globalObject][@"console"];
@weakify(self);
console[@"log"] = ^(NSString* message) {
    @strongify(self);
    [self wmf_logWebViewMessage:message];
};
```

could use this same approach to add `window.onerror` handler like I mentioned

Allows for all console log messages to show up:
```
(lldb) expr (void) [[[0x7ba30200 bridge] webView] stringByEvaluatingJavaScriptFromString:@"console.log('hello world!');"]
2015-08-25 09:49:30:867 Wikipedia Debug[main] -[UIWebView(WMFJavascriptToXcodeConsoleLogging) wmf_logWebViewMessage:]#L20 INFO:
hello world!
```

montehurd: @bgerstle nice! that's even better!!! :)

bgerstle:
:+1: waiting for travis

PR #88, opened 2015-08-26
Respect Search Site changes

coreyfloyd:
This affects Home and Search

The App View controller updates the Home and Search VCs if any changes to the search site are made. Other changes flow from there.

Also caught a bug with the language picker.


bgerstle:
:+1:

PR #90, opened 2015-08-27
Change MWKDataStore article cache behavior to prevent side effects

bgerstle:
Phab: [T110381](https://phabricator.wikimedia.org/T110381)

---

In a nutshell, we were putting "empty" articles in the data store's memory cache, which resulted in the article fetcher filling them with new data, when *then* resulted in view controllers getting false positives on equality checks which they were using to only set new data.  In other words:

- Create an "empty" article and put it in the cache
- Fetch the "full" article. populating the previously "empty" article w/ data
- Try to set the new "full" article in your view controller's in the completion block, but fail because it's the same object: `self.article == article` is true.

To fix it, I changed (and documented) the behaviors for the data store's article loading methods:

- `articleWithTitle:(MWKTitle*)`: high-level API, "just give me an article." if we have an **existing article** (see below), you'll get that. otherwise we'll create a new one
- `existingArticleWithTitle:(MWKTitle*)` similar pattern to the one employed by SDWebImage: give me an article from memory, or read from disk. if we get it from disk, we'll put it in memory so you'll hit the cache next time
- `articleFromDiskWithTitle:(MWKTitle*)` I think you see where this is going... 

On top of this, the fetcher now *always* creates new article objects and saves them.  The main benefit of this is that it doesn't touch the cache (multi-threading problems) or read data from disk (unnecessary).

Now that all of this is fixed, `WMFArticleViewController` can _actually_ rely on equality checks to determine if it got new data. Phew! 😌

In a separate patch, I'd like to look into whether or not we always end up with the new data we get from the server.  I have a hunch that old sections might be able to linger since we're not removing old data on a new download.

### DIFF
--- a/MediaWikiKit/MediaWikiKit/MWKDataStore.h
+++ b/MediaWikiKit/MediaWikiKit/MWKDataStore.h
@@ -91,9 +91,45 @@ extern NSString* const MWKArticleKey;
  */
 - (void)saveImageInfo:(NSArray*)imageInfo forArticle:(MWKArticle*)article;
 
-// Raw load methods
+///
+/// @name Article Load Methods
+///
+
+/**
+ *  Retrieves an existing article from the receiver.
+ *
+ *  This will check memory cache first, falling back to disk if necessary. If data is read from disk, it is inserted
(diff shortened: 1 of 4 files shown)


coreyfloyd:
:+1:

PR #94, opened 2015-08-28
More style tweaks

montehurd:
Vibha had a couple small style tweaks, but we only had time for one.

A unit test wasn't passing, so I fixed it and then wrapped up the fix in a visual testing convenience method.


bgerstle:
@montehurd the visual tests are breaking, seems like there's an unsatisfiable constraint

PR #96, opened 2015-08-28
Fix crash bug on home screen

coreyfloyd:
Found this while working on the popup.

Caused by out of bounds error when creating sections when resuming the app
(NSNotFound for section that was not yet added)

### DIFF
--- a/Wikipedia/UI-V5/WMFHomeViewController.m
+++ b/Wikipedia/UI-V5/WMFHomeViewController.m
@@ -87,7 +87,7 @@ NS_ASSUME_NONNULL_BEGIN
 }
 
 - (void)setSearchSite:(MWKSite* __nonnull)searchSite {
-    _searchSite                           = searchSite;
+    _searchSite                             = searchSite;
     self.nearbySectionController.searchSite = searchSite;
 }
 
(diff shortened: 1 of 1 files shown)


Wikipedia/UI-V5/WMFHomeViewController.m
```diff
@@ -427,6 +435,9 @@ - (void)controller:(id<WMFHomeSectionController>)controller didSetItems:(NSArray
 
 - (void)controller:(id<WMFHomeSectionController>)controller didAppendItems:(NSArray*)items {
     NSInteger section = [self indexForSectionController:controller];
+    if (section == NSNotFound) {
```
bgerstle: why would we get this callback when the section isn't found?

bgerstle: actually, can't we just set the controller's `delegate` to `nil` and prevent callbacks after it has been unloaded?

also, do you think now would be an OK time to start documenting this class? either that, or make the implementation a bit more expressive so the structure is clearer? hard to tell what should/shouldn't be happening in the various methods and what the guarantees are

coreyfloyd: @bgerstle this can happen during modifying the sections when the fetcher is working, but we are also potentially adding, removing sections. Since sections are added using the performBatchupdates method (which runs on a queue sometime in the future), its possible to get a callback when it is configured, but not quite loaded into the collection view yet - as the collection view is still processing the update.

As far as documenting… like I mentioned previously, the plan is to solidify the architecture when working on the ticket to actually implement the business logic for the sections. So thats coming soon. patience…

bgerstle: what about setting the delegate to `nil` until after it's loaded? also, i don't understand how loading a section is asynchronous. adding items to the collectionView is (since we're using `performBatchUpdates:completion:`), but isn't the `sections` property of the `SSSectionedDataSource` updated instantly?

coreyfloyd:
@bgerstle updated to not set the delegate until the load is actually performed.

bgerstle:
@coreyfloyd k, will wait for travis then merge. just running into this on my branch as well.... fun race conditions & whatnot

PR #97, opened 2015-08-28
Fix underlying issue of the bounds error

coreyfloyd:
Also found this while working on the pop up

Following up on:
https://phabricator.wikimedia.org/T110381

The collection view layout is attempting to make calculations using Key Value Coding when the height of the cells is 0
(This occurs when there is a section, but not cells in the section)

This was still happening before the section was loaded when launching in landscape.
Launching in landscape causes the view to rotate prematurely kicking off layout logic.

### DIFF
--- a/Wikipedia/UI-V5/WMFSelfSizingWaterfallCollectionViewLayout.m
+++ b/Wikipedia/UI-V5/WMFSelfSizingWaterfallCollectionViewLayout.m
@@ -1,0 +1,14 @@
+
+#import "WMFSelfSizingWaterfallCollectionViewLayout.h"
+
+@implementation WMFSelfSizingWaterfallCollectionViewLayout
+
+- (CGSize)collectionViewContentSize {
+    if ([self.collectionView numberOfSections] == 1 && [self.collectionView numberOfItemsInSection:0] == 0) {
+        return self.collectionView.bounds.size;
+    }
+
+    return [super collectionViewContentSize];
+}
+
+@end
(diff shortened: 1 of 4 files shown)


bgerstle:
Nice catch! Is this something we can push upstream as well?

Wikipedia/UI-V5/WMFSelfSizingWaterfallCollectionViewLayout.m
```diff
@@ -0,0 +1,14 @@
+
+#import "WMFSelfSizingWaterfallCollectionViewLayout.h"
+
+@implementation WMFSelfSizingWaterfallCollectionViewLayout
+
+- (CGSize)collectionViewContentSize {
+    if ([self.collectionView numberOfSections] == 1 && [self.collectionView numberOfItemsInSection:0] == 0) {
```
bgerstle: this _only_ happens when we have 1 section w/ 0 items? or does it happen when any section has 0 items?

coreyfloyd: @bgerstle Probably a little disingenuous with my subject line, this isn't the "underlying" cause… the underlying cause is the size calculation itself… which needs to be fixed. So that fix can be pushed up stream when I get around to it.

This fix here is the only way to really fix it by subclassing (without bringing it into the source). As far as I can tell, this is the only scenario it breaks on. 

But yeah… I'd like to fix it for realz and get it merged so we can just remove the subclass then.

bgerstle: So, does this mean it won't bite us again under different circumstances? If not, we can merge this and get around to upstreaming later.  OTOH, if we think this could be a problem, let's go ahead and make it a development pod and fix the underlying issue.

bgerstle:
Merging under the premise that we'll dig deeper (and upstream) later as needed

PR #98, opened 2015-08-28
Add extracts to related article feed items

bgerstle:
Phab: [T110578](https://phabricator.wikimedia.org/T110578)

- [x] fetch 4 lines of text
- [x] prevent TextExtracts ellipsis from showing up (by fetching 4.5 lines of text)

#### Review amendments
- [x] 1b8ca78 Move image-less stylesheet factory method into a category on the stylesheet class

---

## Approach
Add `extracts` as part of the related search API call, which returns a "subset of HTML", allowing us to show an attributed string in the preview cell.  When fetching related titles, `WMFRelatedSectionController` calculates the number of characters to fetch based on the maximum width of preview cells, approximate chars per line, and number of lines—adding in an extra half a line to force overflow and UILabel truncation.  See `-[WMFRelatedSectionController numberOfCharactersToFetch]` for details.

## Easter eggs
Fix some autolayout contraints that weren't satisfiable because we set `collectionView.estimatedItemSize = 150` even though the preview cell's image height was `160` :tired_face: 

### Networking architecture preview?!
Instead of copy/pasting/subclassing the `MWKSearchResponseSerializer`, I made a more dynamic version of it which can serialize any `MTLModel` subclass which is JSON serializable.  This provides a working proof of concept for how we might build the foundation of our fetching hierarchy.  Check out `WMFMantleJSONResponseSerializer` for details.

## Known issues
There are some cases where paragraphs are needlessly truncated.  After playing with this over the weekend, I've determined it's something weird with the attributed string returned by DTCoreText, since using the built-in `NSHTMLDocumentType` to generate the attributed string produces correct-looking output.  I'm going to file a bug to investigate this further.

---

## Screenshots

image


montehurd:
Couple small issues.

- Looks like the extract html has "..." so sometimes you get 2 "..." - one from extract text and one from label. If the api can be configured to not return "..." (assuming it's doing so) that would be ideal. This would avoid double "..." issue and things like the following where you have needless "...":

screen shot 2015-09-01 at 11 01 15 am

- There's extra padding on the bottom for some reason:

screen shot 2015-09-01 at 11 01 49 am

Compare that to the following (which lets us adjust padding purely through constraints so we don't have to fiddle with attr string)

screen shot 2015-09-01 at 11 01 49 am

bgerstle:
to reiterate, the extra spacing at the bottom is due to the way `UILabel` renders the string when truncation happens, not a blocking issue here, but we might want to re-investigate `NSHTMLDocumentType` for string generation and/or using a gradient instead of tail truncation for extracts.

Wikipedia/Categories/NSAttributedString+WMFHTMLForSite.m
```diff
@@ -11,13 +11,17 @@
...
 
 + (NSDictionary*)wmf_defaultHTMLOptionsForSite:(MWKSite*)site {
-    UIFont* defaultFont                = [UIFont preferredFontForTextStyle:UIFontTextStyleBody];
-    DTCSSStylesheet* defaultStyleSheet =
-        [[DTCSSStylesheet alloc] initWithStyleBlock:@"img { display: none } "];
+    UIFont* defaultFont = [UIFont wmf_htmlBodyFont];
+    static DTCSSStylesheet* defaultStyleSheet;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        defaultStyleSheet = [[DTCSSStylesheet alloc] initWithStyleBlock:@"img { display: none } "];
```
coreyfloyd: Would be nice if you moved this to a category in the same way you moved the `wmf_htmlBodyFont` to a category.

it's nice for encapsulation, and, honestly, I don't know what this does… so it would be nice to have it in a well named method that conveys it's purpose.

bgerstle: Good call, will do that now

Wikipedia/Networking/Serializers/WMFMantleJSONResponseSerializer.m
```diff
@@ -0,0 +1,111 @@
...
+@implementation WMFMantleJSONCollectionResponseSerializer
+
+- (id)responseObjectForResponse:(NSURLResponse*)response
+                           data:(NSData*)data
+                          error:(NSError* __autoreleasing*)error {
+    id value = [super responseObjectForResponse:response data:data error:error];
+    if ([value isKindOfClass:[NSArray class]]) {
+        return [MTLJSONAdapter modelsOfClass:self.modelClass fromJSONArray:value error:error];
+    } else if (value) {
+        // most MW API responses are indexed by page ID, just grab all the values in the dictionary
```
coreyfloyd: If we want to refactor the serializers to make them more reusable, we should be sure to make it configurable in the ways we need it to be. In other words, it should work without making assumptions about the data (explicit vs implicit).

In this case, it seems like we should be able to specify a mapping from a dictionary to an array. Something like this comes to mind (you probably have a better idea):

```
@property (nonatomic, copy) WMFCollectionTransformBlock transformBlock;
```

Wikipedia/UI-V5/WMFArticlePreviewCell.m
```diff
@@ -6,12 +6,15 @@
...
+#import "NSAttributedString+WMFHTMLForSite.h"
+#import "UIFont+WMFStyle.h"
 #import "UIButton+WMFButton.h"
 #import "WMFSaveButtonController.h"
 #import "SessionSingleton.h"
+#import "NSParagraphStyle+WMFParagraphStyles.h"
 
-static CGFloat const WMFTextPadding = 8.0;
-static CGFloat const WMFImageHeight = 160;
+CGFloat const WMFArticlePreviewCellTextPadding = 8.0;
```
coreyfloyd: Do you have a reference to demonstrate why removing `static` from file scoped constants is the preferred pattern? 

I ask because this goes against our style guide and we should update it if there is a technical reason to do this: https://www.mediawiki.org/wiki/Wikimedia_Apps/Team/iOS/ObjectiveCStyleGuide#Constants

bgerstle: They were exposed publicly, so I have to remove `static` in order for them to be `extern`.

MediaWikiKit/MediaWikiKit/MWKSearchResult.m
```diff
@@ -20,10 +20,10 @@ + (NSValueTransformer*)wikidataDescriptionJSONTransformer {
 
 + (NSDictionary*)JSONKeyPathsByPropertyKey {
     return @{
-               @"displayTitle": @"title",
-               @"articleID": @"pageid",
-               @"thumbnailURL": @"thumbnail.source",
-               @"wikidataDescription": @"terms.description"
+               WMF_SAFE_KEYPATH(MWKSearchResult.new, displayTitle): @"title",
```
coreyfloyd: When you made the suggestion of using `WMF_SAFE_KEYPATH` on my patch, I gave a reason about why I didn't think this was a good idea to use in in the context of Mantle methods:
https://github.com/wikimedia/wikipedia-ios/pull/71#discussion_r37452046

bgerstle: reiterating: `WMF_SAFE_KEYPATH` (like the [libextobjc macro I borrowed from](https://github.com/jspahrsummers/libextobjc/blob/45babac1a457b0f59404a366f26bac2b5b5e82b2/extobjc/EXTKeyPathCoding.h#L38)) doesn't actually run the code in the first argument (i.e. no objects are created).  this happens because the compiler can determine the code is unreachable, and optimize it away.

coreyfloyd:
:+1:

PR #101, opened 2015-08-30
Hooked up WMFArticlePreviewCell save button.

montehurd:
- Moved save button logic to WMFSaveButtonController.h

- Save buttons now working for:
   - Native article card
   - Search results
   - Recent
   - Saved pages
   - Related articles

https://phabricator.wikimedia.org/T110606

### DIFF
--- a/Wikipedia/UI-V5/WMFArticleContainerViewController.m
+++ b/Wikipedia/UI-V5/WMFArticleContainerViewController.m
@@ -9,6 +9,7 @@
 #import "WMFArticleViewController.h"
 #import "WebViewController.h"
 #import "UIViewController+WMFStoryboardUtilities.h"
+#import "WMFSaveButtonController.h"
 
 // Model
 #import "MWKDataStore.h"
@@ -37,7 +38,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (nonatomic, weak, readonly) UIViewController<WMFArticleContentController>* currentArticleController;
 
-@property (strong, nonatomic) UIBarButtonItem* saveButton;
+@property (nonatomic, strong) WMFSaveButtonController* saveButtonController;
 
 @end
 
@@ -78,6 +79,8 @@ NS_ASSUME_NONNULL_BEGIN
 
     _article = article;
 
+    self.saveButtonController.title = article.title;
+
     if (self.isViewLoaded) {
         self.articleViewController.article = article;
         self.webViewController.article     = article;
@@ -112,14 +115,7 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)viewDidLoad {
     [super viewDidLoad];
 
-    @weakify(self)
-    self.saveButton = [UIBarButtonItem wmf_buttonType:WMFButtonTypeBookmark handler:^(id sender){
-        @strongify(self)
-        [self toggleSave : self];
-    }];
-    self.navigationItem.rightBarButtonItem = self.saveButton;
-
-    [self observeSavedPages];
+    [self setupSaveButton];
 
     // Manually adjusting scrollview offsets to compensate for embedded navigation controller
     self.automaticallyAdjustsScrollViewInsets = NO;
@@ -141,8 +137,20 @@ NS_ASSUME_NONNULL_BEGIN
         self.articleViewController.article = self.article;
         self.webViewController.article     = self.article;
     }
+}
 
-    [self updateSavedButtonState];
+- (void)setupSaveButton {
+    @weakify(self)
+    self.navigationItem.rightBarButtonItem = [UIBarButtonItem wmf_buttonType:WMFButtonTypeBookmark handler:^(id sender){
+        @strongify(self)
+        if (![self.article isCached]) {
+            [self.articleViewController fetchArticle];
+        }
+    }];
+
+    self.saveButtonController =
+        [[WMFSaveButtonController alloc] initWithButton:(UIButton*)self.navigationItem.rightBarButtonItem.customView
+                                                  title:self.article.title];
 }
 
 - (void)viewDidLayoutSubviews {
@@ -265,45 +273,6 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (NSString*)analyticsName {
     return [self.articleViewController analyticsName];
-}
-
-#pragma mark - Saved Pages
-
-- (void)observeSavedPages {
-    [self.KVOControllerNonRetaining observe:self.savedPageList
-                                    keyPath:WMF_SAFE_KEYPATH(self.savedPageList, entries)
-                                    options:0
-                                      block:^(WMFArticleContainerViewController* observer, id object, NSDictionary* change) {
-        [observer updateSavedButtonState];
-    }];
-}
-
-- (void)unobserveSavedPages {
-    [self.KVOControllerNonRetaining unobserve:self.savedPageList keyPath:WMF_SAFE_KEYPATH(self.savedPageList, entries)];
-}
-
-- (IBAction)toggleSave:(id)sender {
-    if (![self.article isCached]) {
-        [self.articleViewController fetchArticle];
-    }
-
-    [self unobserveSavedPages];
-    [self.savedPageList toggleSavedPageForTitle:self.article.title];
-    [self.savedPageList save];
-    [self observeSavedPages];
-    [self updateSavedButtonState];
-}
-
-- (void)updateSavedButtonState {
-    ((UIButton*)self.saveButton.customView).selected = [self isSaved];
-}
-
-- (BOOL)isSaved {
-    return [self.savedPageList isSaved:self.article.title];
-}
-
-- (void)dealloc {
-    [self unobserveSavedPages];
 }
 
 @end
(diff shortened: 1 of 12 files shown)


Wikipedia/UI-V5/WMFSaveButtonController.m
```diff
@@ -0,0 +1,83 @@
...
+        self.button = button;
+        self.title  = title;
+        [self observeSavedPages];
+        [self updateSavedButtonState];
+    }
+    return self;
+}
+
+- (MWKSavedPageList*)savedPageList {
+    return [SessionSingleton sharedInstance].userDataStore.savedPageList;
```
montehurd: A central saved page list needs to be observed so changes are reflected anywhere this controller is used.

Wikipedia/UI-V5/WMFRecentPagesDataSource.m
```diff
@@ -36,6 +36,7 @@ - (nonnull instancetype)initWithRecentPagesList:(MWKHistoryList*)recentPages {
             cell.descriptionText       = article.entityDescription;
             cell.image                 = [article bestThumbnailImage];
             cell.summaryAttributedText = [article summaryHTMLWithoutLinks];
+            cell.title                 = article.title;
         };
```
montehurd: Since I have to pass title, I guess I could remove cell.titleText and get the article title from cell.title...

Wikipedia/UI-V5/WMFSaveButtonController.m
coreyfloyd: @montehurd Couple of notes…

1. Use injection whenever possible. Accessing singletons should be a pattern of last resort. If you have to access the `SessionSingleton` in a class… consider that a code smell. Try to pass anything required by a class on initialization. In this case, you should be passing the required`MWKSavedPageList` in the initializer. 

2. Re: "Central Saved page List" - currently we are using only one saved page list in the app. The one on the session singleton. This should be the one that is passed in. So you don't need to worry about observing the "correct one" - just observe the one that is passed in.

Wikipedia/UI-V5/WMFSaveButtonController.m
```diff
@@ -0,0 +1,83 @@
...
+    [self observeSavedPages];
+    [self updateSavedButtonState];
+}
+
+- (void)setTitle:(MWKTitle*)title {
+    _title = title;
+    [self updateSavedButtonState];
+}
+
+- (void)setButton:(UIButton*)button {
```
bgerstle: do we gain anything by keeping this private? seems like the setter was implemented to support reuse across multiple buttons

Wikipedia/UI-V5/WMFSaveButtonController.m
```diff
@@ -0,0 +1,83 @@
...
+- (MWKSavedPageList*)savedPageList {
+    return [SessionSingleton sharedInstance].userDataStore.savedPageList;
+}
+
+- (void)observeSavedPages {
+    [self.KVOControllerNonRetaining observe:self.savedPageList
+                                    keyPath:WMF_SAFE_KEYPATH(self.savedPageList, entries)
+                                    options:0
+                                      block:^(WMFSaveButtonController* observer, id object, NSDictionary* change) {
+        [observer updateSavedButtonState];
```
bgerstle: maybe a premature optimization, but it might save us some time if we checked the object that was added/removed to the list to see if matches our title, e.g.:

``` objc
switch ([changeDictionary[NSKeyValueChangeKindKey] intValue]) {
  case NSKeyValueChangeRemoval:
  case NSKeyValueChangeReplacement: {
    MWKSavedPageEntry* prevEntry = change[NSKeyValueChangeOldKey]
    if (prevEntry.title isEqual:self.title) { /* update state w/o checking entire list */ }
  }
}
```

Wikipedia/UI-V5/WMFSaveButtonController.m
```diff
@@ -0,0 +1,83 @@
...
+@implementation WMFSaveButtonController
+
+- (instancetype)initWithButton:(UIButton*)button
+                         title:(MWKTitle*)title {
+    self = [super init];
+    if (self) {
+        self.button = button;
+        self.title  = title;
+        [self observeSavedPages];
+        [self updateSavedButtonState];
```
bgerstle: Instead of manually calling `updateSavedButtonState` after every call to `observeSavedPages`, you can harden this "contract" by specifying the `NSKeyValueObservingOptionInitial` option, which invokes the KVO callback immediately after observation w/ the current value, thus causing you to update to the current state.

Either that, or add a call to `updateSavedButtonState` at the end of `observeSavedPages`.

Wikipedia/UI-V5/WMFArticleContainerViewController.m
```diff
@@ -141,8 +137,20 @@ - (void)viewDidLoad {
...
     }
+}
 
-    [self updateSavedButtonState];
+- (void)setupSaveButton {
+    @weakify(self)
+    self.navigationItem.rightBarButtonItem = [UIBarButtonItem wmf_buttonType:WMFButtonTypeBookmark handler:^(id sender){
+        @strongify(self)
+        if (![self.article isCached]) {
+            [self.articleViewController fetchArticle];
```
bgerstle: why are we fetching here?

montehurd: @bgerstle 
It was doing this previously (line 287) when save was tapped.

Wikipedia/UI-V5/WMFSaveButtonController.m
montehurd: @coreyfloyd @bgerstle 
Singleton bad yeah yeah. As a parameter though we may accidentally in the future pass it a different instance of saved pages list, forgetting that saved pages lists themselves don't observe the file system for changes, thus introducing a bug where these separate saved pages lists are out of sync.

That fragility had more of a *code smell* than pointing to a common saved pages list, hence my comment above.

Thoughts?

Wikipedia/UI-V5/WMFSaveButtonController.m
montehurd: Could, but presently it's not used in that way.

Wikipedia/UI-V5/WMFArticleContainerViewController.m
bgerstle: oh right to make sure we have it ready for offline. duh

Wikipedia/UI-V5/WMFSaveButtonController.m
bgerstle: k :+1:

Wikipedia/UI-V5/WMFSaveButtonController.m
montehurd: @bgerstle Oh good idea! I always forget about the *initial* option.

Wikipedia/UI-V5/WMFSaveButtonController.m
bgerstle: @montehurd **TL;DR;** I don't think that's sufficient reason to abandon DI. I think we should continue to pass (note, not create) saved page lists.  Synchronizing saved page lists is a separate issue which we can resolve separately if needed.

---

I also considered this vulnerability, but I consider it to be an edge case.  You could make the same argument for every other `MWKSavedPageList`-containing class in the app, but my opinion is that the architectural benefits of loose coupling outweigh the potential problems—which we have yet to encounter in practice AFAIK.*

From what I can see we have two problems:

1. How do we ensure saved pages is always up to date
2. How do we acquire a class's dependencies?

They are not orthogonal, but I consider the latter to be solved through the use of dependency injection.  The former, OTOH, is an unfortunate legacy issue which even a singleton doesn't solve.**  If we want, we can synchronize saved page lists rather easily:

- Post a notification from the data store (which is **the** way to save them\***)
- Observe that notification in saved page list
- Check if the list that was saved is equal to `self`
- If not, reset in-memory storage, which is lazily reloaded from disk on next access

\* Issues w/ the lists not updating is a symptom of poorly factored logic that relied on a single, multi-responsibility object to make all necessary updates, and should be fixed once we factor that logic out into a separate mechanism which doesn't rely on any one code path.
\** Creating and saving another saved page list could overwrite the data and knock everything else out of sync until next save.
\*** Although not the only way...

montehurd: @bgerstle good points!

bgerstle:
:+1: looks good! `WMFSaveButtonController` is a great candidate for unit testing as well.  we're not strictly requiring tests w/ new code, but it would be nice since we're relying on this to ensure save buttons work everywhere. however, since I'm in a meeting the rest of today (and potentially tomorrow), then I can't really offer much help.  

if you're down to tackle testing I say go for it, otherwise I can file an eng backlog ticket to add tests for this class.  whaddya say, @montehurd?

bgerstle:
if you do feel like adding tests, I would suggest using a real (temporary) saved page list, since testing things like  updating to the proper state after init will be tricky w/ a mock. see `MWKDataStore+TemporaryDataStore` for convenience in creating a data store in a random `tmp` directory (thus preventing side-effects between tests).

also, be sure to cover all the state transitions:

- initial observation
- toggle saved
- toggle unsaved
- correct button state after button is set
- correct button state after title is set
- toggle saved/unsaved **of a different title** results in no state change

PR #102, opened 2015-09-02
Fix for cell shadows extending down too far and not resizing properly.

montehurd:
https://phabricator.wikimedia.org/T110442

**Before:** *Note: shadows are bright red for easy comparison.*
 
- Shadow doesn't extend to bottom of cell:
screen shot 2015-09-01 at 5 59 36 pm
- Shadow wonky width after rotate:
screen shot 2015-09-01 at 5 59 46 pm
-Moar wonk after rotating back to portrait:
screen shot 2015-09-01 at 5 59 56 pm

**After:**
screen shot 2015-09-01 at 5 57 37 pm
screen shot 2015-09-01 at 5 57 49 pm
screen shot 2015-09-01 at 5 57 32 pm
^ I scrolled around on this last one - that's why it's not in same position as pre-rotation image.

### DIFF
--- a/Wikipedia/UI-V5/UIView+WMFShadow.m
+++ b/Wikipedia/UI-V5/UIView+WMFShadow.m
@@ -11,16 +11,11 @@
 @implementation UIView (WMFShadow)
 
 - (void)wmf_setupShadow {
-    self.clipsToBounds       = NO;
+    self.layer.masksToBounds = NO;
     self.layer.shadowColor   = [UIColor blackColor].CGColor;
-    self.layer.shadowOpacity = 0.15;
-    self.layer.shadowRadius  = 3.0;
+    self.layer.shadowOpacity = 0.2;
+    self.layer.shadowRadius  = 4.0;
     self.layer.shadowOffset  = CGSizeZero;
-    [self wmf_updateShadowPathBasedOnBounds];
-}
-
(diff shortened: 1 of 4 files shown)


Wikipedia/UI-V5/UIView+WMFShadow.m
```diff
@@ -11,16 +11,11 @@
 @implementation UIView (WMFShadow)
 
 - (void)wmf_setupShadow {
-    self.clipsToBounds       = NO;
+    self.layer.masksToBounds = NO;
```
bgerstle: is there a functional difference between setting the layer property and the `UIView` property? IIRC the `UIView` property is a proxy for `masksToBounds`, which is the case for many `UIView` properties.

bgerstle:
:+1: for fixing and :100: for before/after screenshots!  that said, we might have to revisit `shadowPath` if scrolling performance becomes a problem

Wikipedia/UI-V5/UIView+WMFShadow.m
montehurd: @bgerstle 
Not a proxy iirc as a view can have many layers. Switching this line to the layer's maskToBounds means we can now apply shadows to things like UITableView, which can look wonky if you turn off clipsToBounds.

PR #104, opened 2015-09-02
Update recent searches screen layout according to new mocks.

montehurd:
https://phabricator.wikimedia.org/T110724

Note: ended up diverging from mock a bit based on conversations
with Corey. For now the individual recent search terms are styled
more like small cards with their own shadow.

(reminder: this ticket doesn't address the search box itself)

**Before:**
screen shot 2015-09-02 at 2 18 56 pm

**After:**
screen shot 2015-09-02 at 2 14 21 pm

### DIFF
--- a/Wikipedia/View Controllers/RecentSearches/RecentSearchCell.m
+++ b/Wikipedia/View Controllers/RecentSearches/RecentSearchCell.m
@@ -2,59 +2,43 @@
 //  Copyright (c) 2014 Wikimedia Foundation. Provided under MIT-style license; please copy and modify!
 
 #import "RecentSearchCell.h"
-#import "PaddedLabel.h"
-#import "NSObject+ConstraintsScale.h"
-#import "Defines.h"
 #import "WikiGlyph_Chars.h"
 #import "UIFont+WMFStyle.h"
+#import "UIView+WMFShadow.h"
+#import "UIColor+WMFHexColor.h"
 
-#define FONT_SIZE (16.0f * MENUS_SCALE_MULTIPLIER)
-#define FONT_COLOR [UIColor grayColor]
-#define ICON_SIZE (20.0f * MENUS_SCALE_MULTIPLIER)
-#define ICON_GLYPH WIKIGLYPH_MAGNIFY_BOLD
-#define BORDER_COLOR [UIColor colorWithWhite:0.9 alpha:1.0]
-#define MAGNIFY_ICON_COLOR [UIColor colorWithWhite:0.8 alpha:1.0]
+static CGFloat const magnifyIconSize    = 28.f;
+static NSInteger const magnifyIconColor = 0x777777;
 
 @interface RecentSearchCell ()
 
-@property (weak, nonatomic) IBOutlet NSLayoutConstraint* topBorderHeightConstraint;
-@property (weak, nonatomic) IBOutlet UIView* topBorderView;
-@property (strong, nonatomic) NSAttributedString* magnifyIconString;
-@property (weak, nonatomic) IBOutlet PaddedLabel* iconLabel;
+@property (strong, nonatomic) IBOutlet UIView* shadowView;
+@property (strong, nonatomic) IBOutlet UILabel* iconLabel;
 
 @end
 
 @implementation RecentSearchCell
 
 - (void)awakeFromNib {
-    self.selectionStyle = UITableViewCellSelectionStyleNone;
-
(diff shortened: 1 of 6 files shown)


Wikipedia/View Controllers/RecentSearches/RecentSearchCell.m
```diff
@@ -2,59 +2,43 @@
...
-    return [[NSAttributedString alloc] initWithString:ICON_GLYPH
-                                           attributes:@{
-                NSFontAttributeName: [UIFont wmf_glyphFontOfSize:ICON_SIZE],
-                NSForegroundColorAttributeName: MAGNIFY_ICON_COLOR,
-                NSBaselineOffsetAttributeName: @1
-            }];
++ (NSAttributedString*)attributedIconString {
+    static dispatch_once_t once;
+    static NSAttributedString* sharedString;
+    dispatch_once(&once, ^{
```
bgerstle: this is fine, but in the future if we know something is UI related, it's enough to add an assertion that the call i made on the main thread* and skip the extra overhead of `dispatch_once`.

\* `NSParameterAssert([NSThread isMainThread])`

Wikipedia/View Controllers/RecentSearches/RecentSearchesViewController.m
```diff
@@ -103,8 +96,8 @@ - (void)saveTerm:(NSString*)term
          @"type": @(searchType)
      } atIndex:0];
 
-    if (self.tableDataArray.count > LIMIT) {
-        self.tableDataArray = [self.tableDataArray subarrayWithRange:NSMakeRange(0, LIMIT)].mutableCopy;
+    if (self.tableDataArray.count > recentSearchLimit) {
+        self.tableDataArray = [self.tableDataArray subarrayWithRange:NSMakeRange(0, recentSearchLimit)].mutableCopy;
```
bgerstle: use `NSArray wmf_safeSubarrayWithRange:` to avoid an `NSRangeException` when out of bounds.

Wikipedia/View Controllers/RecentSearches/RecentSearchCell.m
```diff
@@ -2,59 +2,43 @@
...
+        sharedString =
+            [[NSAttributedString alloc] initWithString:WIKIGLYPH_MAGNIFY_BOLD
+                                            attributes:@{
+                 NSFontAttributeName: [UIFont wmf_glyphFontOfSize:magnifyIconSize],
+                 NSForegroundColorAttributeName: [UIColor wmf_colorWithHex:magnifyIconColor alpha:1.0f],
```
bgerstle: i know it's trivial, but why are we converting a static value from hex? why not just store it as a float?  it would be one thing if we were converting serialized hex values into decimal floats, but if we're only using this for constant color values then IMO we should just do the math ourselves and store float constants.

montehurd: Single float is much less human readable.

Wikipedia/View Controllers/RecentSearches/RecentSearchCell.m
coreyfloyd: @bgerstle it seems like the dispatch once is just being used to make sure that the string is only generated once… as opposed to being concerned about the main thread. Am I misunderstanding?

@montehurd is this right?

Wikipedia/View Controllers/RecentSearches/RecentSearchesViewController.m
montehurd: Fixed!

coreyfloyd:
:+1:

Wikipedia/View Controllers/RecentSearches/RecentSearchCell.m
bgerstle: That might be true for some, but we're writing a native app, not CSS for a website.  Besides, IMHO RGB(A) is just as (if not more) readable than hex.  I would rather use something that most iOS devs would be familiar with (RGBA) than something most web devs would be familiar with.

Further, constants in code are not documentation.  If we need to capture how something looks or behaves, we look at logical/visual tests and/or designs.

Lastly, the only _human readable_ solution is to use `UIColor` class category methods for specific colors in the app. For example, `wmf_logoBlue` tells me exactly what it is.

bgerstle:
:+1:

PR #106, opened 2015-09-03
Search crash fix

bgerstle:
If we had to change constraints between search queries (e.g. if one had a suggestion and the other didn't), then we would trigger an animated layout pass, causing the collectionView to access it's **now deallocated** dataSource.  Step by step:

- Get new search results
- Start animating new search suggestion (or lack thereof)
- Before animation actually starts, set new dataSource, causing old one to be released
- Old dataSource is queried for numberOfSections
:boom: EXC_BAD_ACCESS

Actual bug repro steps (now fixed):
- Search for something w/o a suggestion (e.g. Rocky)
- Wait until results arrive
- Search for something w/ a suggestion (e.g. Appa)
- Quickly switch between the two, and you'll eventually hit the race condition & crash


montehurd:
Wow... that's some funky business!!! Good sleuthing! :+1:

PR #107, opened 2015-09-03
New search cell designs (T110725)

bgerstle:
Phab: [T110725](https://phabricator.wikimedia.org/T110725)

- [x] Custom cells
- [x] Text truncation rules (see `WMFSearchResultCellVisualTests`)
- [x] Highlight search query substring of title
- [x] Tested rotation handling manually
- [x] Changed search image width request parameter

---

#### `WMFSearchResultCell`
As the name implies, it's our new cell for search results, vended by `WMFSearchResults` (the pseudo-view-model for search).  It's a subclass of `WMFSaveableTitleCollectionViewCell` (see below) which provides a description of the search result and special title highlighting logic.  This cell also has a lot of "business logic" built into its constraints, for which I set up a comprehensive suite of visual tests.  See `WMFSearchResultCellVisualTests` for details & reference images.  Some caveats:

The description has to be moved down slightly when the title is 2 lines.  The alternative is to either change the text size and/or padding amounts, or have it be hidden entirely.  Preventing it from moving down w/ the current style results in the label being vertically squashed, which is why I need to add a height constraint, which only breaks _after_ the bottom constraints are broken (at which point the label collapses and the title takes up the entire cell).

The highlighting business logic wasn't clearly spec'd out anywhere, so I'm naively searching for a case-insensitive substring and applying bold to it.  Ideally the search API would tell us which parts of the title (if anything) matched or even return it as HTML (which we can handle now), instead of us having to duplicate the logic client-side.

#### `UIView+VisualTestingSizeUtils`
Extracted & DRY'd up "sizeToFit with auto layout" logic in `FBSnapshotTestCase+WMFConvenience` into separate categories for easier reuse & modular composition.  Used in the new `WMFSearchResultCellVisualTests`, which test the cell in isolation (w/o a VC or table view).

#### `WMFSaveableTitleCollectionViewCell`
Formerly known as `WMFShadowCell`, contains more shared logic extracted from `WMFArticlePreviewCell` (and DRY-ed up from `WMFHomeNearbyCell`).  Provides a home for the following shared functionalities across cells:

- Title label, with a customizable `updateTitleLabel` which subclasses can override to calculate & set plain or attributed strings
- Save button, for saving the `title` associated with the cell
- Image view, which can be populated with a plain URL or, if available, `MWKImage`. We might want to remove the `URL` functionality and prefer `MWKImage` for this hierarchy of cell classes
- Shading

See 69d3522, 76b0f02, and 0f104e0

#### `WMFSaveButtonController`
I followed through with our discussion on the save button controller an deleted all references to the `SessionSingleton`. This required a couple of changes:

- Adding a `setSavedPageList:` method to the `WMFHomeSectionController` protocol
- Implementing it as a property on all the section controllers
- Making all section controllers pass the saved page list to their cells

It's a bit more code, but now we're not implicitly relying on all view controllers being given the singleton's saved page list.  Instead, we're explicitly passing the root view controller's saved page list down to its children, and them to their children, and so on...

See 69d3522.

#### `WMFHomeSectionController`
In addition to modifications above, I added an optional method that allows sections to specify an `estimatedItemHeight`. This was necessary to prevent initially-unsatisfiable constraints in the search view when new cells were integrated. See 2fbf4d3.

#### `WMFSearchViewController`
Did some boyscout cleanups & styling stuff in 9537225 and 9196588.  See commit messages for TL;DR; and the diff for details.

#### `SearchResultFetcher` & `MWKArticle`
Changed search resolution back to `70 * mainScreen.scale` and removed legacy call (and implementation of) `loadThumbnailFromDisk`. See cddbf68 and 66d3777.

#### `WMFArticleListCollectionViewController`
In addition to refactors done in 69d3522, I made two changes:

- Setting `collectionView.dataSource` to `nil` when the `WMFArticleDataSource` property is set to `nil`, which prevents the `collectionView` from accessing a deallocated object
- Call `reloadData` manually to empty the collection view if no `dataSource` is set.

#### `FBSnapshotTestCase+WMFConvenience`
Refactored to use `UIView+VisualTestingSizeUtils` and DRY'd up the `FBSnapshotVerifyView` workaround logic into our own macro.

---

Example screencap:

image

Test GIF (see high-res video [here](https://www.dropbox.com/s/v9sjsdjo1eo5gg8/new%20search%20cell%20test%20run.mov?dl=0)):
new-search-cell-test-run


Wikipedia/UI-V5/WMFArticleListCollectionViewController.m
```diff
@@ -46,18 +46,33 @@ - (void)setDataSource:(SSArrayDataSource<WMFArticleListDataSource>* __nullable)d
...
 
+    [_dataSource setSavedPageList:self.savedPages];
+
     if ([self isViewLoaded]) {
-        _dataSource.collectionView = self.collectionView;
+        if (_dataSource) {
+            _dataSource.collectionView = self.collectionView;
+            if ([_dataSource respondsToSelector:@selector(estimatedItemHeight)]) {
+                [self flowLayout].estimatedItemHeight = _dataSource.estimatedItemHeight;
+            }
```
coreyfloyd: Any reason why we reload if the data source is nil, but not if it is non-nil?

Wikipedia/UI-V5/WMFArticleListDataSource.h
```diff
@@ -17,9 +18,13 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (MWKHistoryDiscoveryMethod)discoveryMethod;
 
+- (void)setSavedPageList:(MWKSavedPageList*)savedPageList;
```
coreyfloyd: Any reason for this to be mutable as opposed to initing with it?

Also, besides setting this property, the Article List VC really doesn't need to know that a data source has a saved page list, right? All the cells currently have a save button, but the Article VC doesn't know that. Maybe this points to this not really being the right place for this.

Wikipedia/UI-V5/WMFSaveableTitleCollectionViewCell.m
```diff
@@ -0,0 +1,115 @@
...
+- (void)setSaveButton:(UIButton*)saveButton {
+    self.saveButtonController.button = saveButton;
+}
+
+- (UIButton*)saveButton {
+    return self.saveButtonController.button;
+}
+
+- (void)setSavedPageList:(MWKSavedPageList*)savedPageList {
+    self.saveButtonController.savedPageList = savedPageList;
```
coreyfloyd: If this was wired at the view controller level, then the cell wouldn't need to know about the saved page list. 

Not saying the cell shouldn't/couldn't retain the reference to the savedButtonController for ease of use - just that cell needs to know about the saved page list only because it needs to set the property on the button controller.

Wikipedia/UI-V5/WMFSaveableTitleCollectionViewCell+Subclass.h
```diff
@@ -0,0 +1,48 @@
...
+//  WMFSaveableTitleCollectionViewCell+Subclass.h
+//  Wikipedia
+//
+//  Created by Brian Gerstle on 9/2/15.
+//  Copyright (c) 2015 Wikimedia Foundation. All rights reserved.
+//
+
+#import "WMFSaveableTitleCollectionViewCell.h"
+
+@interface WMFSaveableTitleCollectionViewCell ()
```
coreyfloyd: Should have a Category name here like `(Subclasses)`

In general class extensions are for private information to the class itself, rather than for public consumption via a separate header. The canonical example is UIGestureRecognizerSubclass.h.

Wikipedia/UI-V5/WMFSaveableTitleCollectionViewCell.h
```diff
@@ -0,0 +1,34 @@
...
+@class MWKSavedPageList;
+@class MWKImage;
+
+/**
+ *  Base class for cells which represent a title and have save button.
+ *
+ *  Will also be given the "card" style by applying a white background to its
+ *  @c contentView as well as a small shadow.
+ */
+@interface WMFSaveableTitleCollectionViewCell : SSBaseCollectionCell
```
coreyfloyd: Feels like the name should be a little more generic since this class includes 2 distinct pieces of functionality:

- Saving
- Shadow applying

Or, if you prefer to be more specific rather than generic, you may consider leaving "WMFShadowCell" as it's own subclass and having `WMFSaveableTitleCollectionViewCell` inherit from that instead.

Wikipedia/UI-V5/WMFArticleListDataSource.h
```diff
@@ -17,9 +18,13 @@ NS_ASSUME_NONNULL_BEGIN
...
+
 @optional
 - (void)deleteArticleAtIndexPath:(NSIndexPath*)indexPath;
 
+- (CGFloat)estimatedItemHeight;
```
coreyfloyd: Did you consider putting this on the Cell class instead? Seems more view related and tied directly to the cell (and delegate methods as opposed to datasource methods).

Wikipedia/UI-V5/WMFSearchViewController.m
```diff
@@ -75,17 +75,32 @@ - (void)updateSearchStateAndNotifyDelegate:(WMFSearchState)state {
...
+
+- (void)updateRecentSearchesVisibility:(BOOL)animated {
+    BOOL hideRecentSearches =
+        [self.searchBar.text length] > 0 || self.recentSearchesViewController.recentSearchesItemCount == 0;
+
+    /*
+       HAX: Need to show/hide superviews since recent & results are in the same container. should use UIViewController
+          containment/transition API instead in the future.
+     */
+    if ((self.recentSearchesViewController.view.superview.alpha == 0) == hideRecentSearches) {
```
coreyfloyd: Have a few questions / comments

1. The line it self pretty hard to read. We should favor readability over compactness. two equality comparisons makes this hard to grok.
2. At a minimum we can encapsulate this logic in a well named method to make it obvious what/why this is being compared.
3. Isn't self.recentSearchesViewController.view.superview just the container view?
4. If the answer is yes, then couldn't we can create an IBOutlet to hold the relevant view to make the intent more obvious?
5. Should we be capturing this state in an ivar (a `BOOL`) to be more explicit instead of relying on the alpha state which may be fragile?

Wikipedia/UI-V5/WMFSearchViewController.m
```diff
@@ -75,17 +75,32 @@ - (void)updateSearchStateAndNotifyDelegate:(WMFSearchState)state {
...
+        return;
     }
+
+    [UIView animateWithDuration:animated ? [CATransaction animationDuration] : 0.0
```
coreyfloyd: `[CATransaction animationDuration]` is non-specific in a case where we should probably be specific and specify the timing we want.

Wikipedia/UI-V5/WMFArticleListCollectionViewController.m
bgerstle: mainly because I went to let `SSDataSource` do its thing when the `collectionView` is set, which AFAICT means adding the items instead of reloading all the data (which also gives us a nicer transition).

Wikipedia/UI-V5/WMFArticleListDataSource.h
bgerstle: > Any reason for this to be mutable as opposed to initing with it?

Can't be passed as an initialization parameter because:

1. the data sources have their own designated initializers 
2. the article list doesn't create them

Maybe we can make `initWithSavedPageList:` the designated initializer of the [shared base class](https://phabricator.wikimedia.org/T111571), until then this was the easiest way.

We needed a way to propagate the saved page list down the object graph

Wikipedia/UI-V5/WMFSaveableTitleCollectionViewCell.m
bgerstle: The cell doesn't expose the `savedButtonController` because we don't want to confuse the API by having multiple places where a `title` can be set.  We need to set the `title` at the cell level so its label can be updated (unless we have the cell observe the `saveButtonController`, but that smells weird).

Wikipedia/UI-V5/WMFSaveableTitleCollectionViewCell+Subclass.h
bgerstle: Done

Wikipedia/UI-V5/WMFSaveableTitleCollectionViewCell.h
bgerstle: > Or, if you prefer to be more specific rather than generic, you may consider leaving "WMFShadowCell" as it's own subclass and having WMFSaveableTitleCollectionViewCell inherit from that instead.

Thought about that, but seems silly.  Even if we wanted the shadow to be optional, I would rather do it at runtime via configuration (`@property hasShadow`) or polymorphism (`+(BOOL)hasShadow`) over inheritance (picking a different base class).

bgerstle: I like the current name as it gives you some clue as to what's special about that class (i.e. it's a save-able cell).  However, given that (I think) we want all our cells to be save-able, we could rename this to something like `WMFBaseCollectionViewCell`.

coreyfloyd:
@bgerstle left some notes… One other general observation about the commit.

I definitely appreciate the insight into the pull request, but, looking at this from the vantage point of productivity, I'm not sure the depth of information in the description is required for this or most pull requests. 

We should strive for code to be self documenting… with tests, good architecture, code comments, and well named methods / variables. 

Information in the pull request description should serve to fill the gaps that weren't otherwise obvious - as opposed to (re)documenting the entire change set.

When we provide this level of detail in the pull request description, it comes at a cost of dev time (and essentially "disappears" from the record when the pull request is closed unlike the other methods of documentation mentioned above). 

I can definitely see where something more extensive is helpful to describe implementing a new complex architecture - I just think we should be prudent about when we provide this level of detail.

One thing great about pull requests is that they are a conversation… and if such an explanation is required for a pull request (which sometimes it can be), we can always ask for a more thorough breakdown of the work.

Wikipedia/UI-V5/WMFSearchViewController.m
bgerstle: addressed in f6d8b8f.

> 1. The line it self pretty hard to read. We should favor readability over compactness. two equality comparisons makes this hard to grok.

True, refactored the visibility check into a getter method.

> 2. At a minimum we can encapsulate this logic in a well named method to make it obvious what/why this is being compared.

Yep :point_up: 

> 3. Isn't self.recentSearchesViewController.view.superview just the container view?

Yep, didn't realize there were outlets for it.

> 4. If the answer is yes, then couldn't we can create an IBOutlet to hold the relevant view to make the intent more obvious?

Already existed :point_up:

> 5. Should we be capturing this state in an ivar (a `BOOL`) to be more explicit instead of relying on the alpha state which may be fragile?

turns out there are some issues w/ this.. looking into it now

bgerstle:
:-1: dealing w/ some issues w/ flickering between results & recent, stand by..

bgerstle:
ok should be good to go :checkered_flag:

PR #108, opened 2015-09-04
Article vc empty when loading

montehurd:
https://phabricator.wikimedia.org/T110631

- [x] table section headers only show once rows for respective sections are present
- [x] sections beneath minimal article summary only show once summary is present
- [x] show "loading" text until summary is fetched

### DIFF
--- a/Wikipedia/UI-V5/WMFArticleViewController.m
+++ b/Wikipedia/UI-V5/WMFArticleViewController.m
@@ -538,23 +538,12 @@ NS_ASSUME_NONNULL_BEGIN
 }
 
 - (CGFloat)tableView:(UITableView*)tableView heightForHeaderInSection:(NSInteger)section {
-    if ([self isLeadSection:section]
-        || ([self isTOCSection:section] && section != self.indexSetOfTOCSections.firstIndex)
-        || ![self tableView:tableView viewForHeaderInSection:section]
-        ) {
-        // omit headers for lead section & all but the first TOC section
-        return 0;
-    } else {
-        return UITableViewAutomaticDimension;
-    }
+    return [self shouldHideHeaderForSection:section] ? 0 : UITableViewAutomaticDimension;
 }
 
 - (UIView*)tableView:(UITableView*)tableView viewForHeaderInSection:(NSInteger)section {
     //TODO(5.0): localize these!
-    if (section == 0
-        || (section != 1 && section < tableView.numberOfSections - 1)
-        || ([self tableView:tableView numberOfRowsInSection:section] == 0)
-        ) {
+    if ([self shouldHideHeaderForSection:section]) {
         return nil;
     } else {
         WMFArticleSectionHeaderView* header =
@@ -566,6 +555,18 @@ NS_ASSUME_NONNULL_BEGIN
         }
         return header;
     }
+}
+
+- (BOOL)shouldHideHeaderForSection:(NSUInteger)section {
+    return [self isLeadSection:section] || [self isTOCSectionAfterFirst:section] || ![self hasRowsInSection:section];
+}
+
+- (BOOL)isTOCSectionAfterFirst:(NSUInteger)section {
+    return [self isTOCSection:section] && section != self.indexSetOfTOCSections.firstIndex;
+}
+
+- (BOOL)hasRowsInSection:(NSUInteger)section {
+    return [self.tableView numberOfRowsInSection:section] > 0;
 }
 
 #pragma mark - UITableViewDelegate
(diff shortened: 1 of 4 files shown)


Wikipedia/UI-V5/WMFArticleViewController.m
```diff
@@ -363,6 +369,37 @@ - (void)updateUIForMode:(WMFArticleControllerMode)mode animated:(BOOL)animated {
...
+}
+
+- (void)showLoadingMessageContainerView {
+    [self setupLoadingMessageContainerViewIfNecessary];
+    self.loadingMessageContainerView.hidden = NO;
+}
+
+- (void)tableView:(UITableView*)tableView willDisplayCell:(UITableViewCell*)cell forRowAtIndexPath:(NSIndexPath*)indexPath {
+    if (indexPath.section == 0) {
+        if (indexPath.row == 0) {
```
bgerstle: can you make this a single `if` statement w/ an `&&`? we're not in tokyo, so no need to rightward drift... 😬

Wikipedia/UI-V5/WMFArticleViewController.m
```diff
@@ -491,7 +541,9 @@ - (WMFArticleReadMoreCell*)readMoreCellAtIndexPath:(NSIndexPath*)indexPath {
 
 - (CGFloat)tableView:(UITableView*)tableView heightForHeaderInSection:(NSInteger)section {
     if ([self isLeadSection:section]
-        || ([self isTOCSection:section] && section != self.indexSetOfTOCSections.firstIndex)) {
+        || ([self isTOCSection:section] && section != self.indexSetOfTOCSections.firstIndex)
+        || ![self tableView:tableView viewForHeaderInSection:section]
```
bgerstle: i'm a bit baffled as to how this doesn't cause an infinite loop... don't you need to know the header height to get the header view?

bgerstle: also, this seems to only be calling `viewForHeaderInSection:` so it can indirectly run the conditional checks there. can we refactor this to use the conditional logic directly? it seems like all we really need to check is `numberOfRowsInSection == 0`, since the other two checks in `viewForHeaderInSection:` are the same as above. this is my bad, since I didn't refactor the `viewForHeaderInSection` conditionals to match these.

the way to fix all of this and DRY up the code is to refactor the 3 conditional checks into one method, and call it in both `heightForHeaderInSection` and `viewForHeaderInSection`. for example:

``` objc
 - (BOOL)shouldHideHeaderForSection:(NSUInteger)section {
  return [self isLeadSection:section] || [self isTOCSectionAfterFirst:section] || [self hasRowsInSection:section];
}

// took the liberty to factor this out too which makes the above even more expressive
- (BOOL)isTOCSectionAfterFirst:(NSUInteger)section {
  return [self isTOCSection:section] && section != self.indexSetOfTOCSections.firstIndex;
}

// same thing here, could even be a UITableView category
- (BOOL)hasRowsInSection:(NSUInteger)section {
  return [self.tableView numberOfRowsInSection:section] > 0;
}
```

Wikipedia/UI-V5/WMFArticleViewController.m
```diff
@@ -501,7 +553,10 @@ - (CGFloat)tableView:(UITableView*)tableView heightForHeaderInSection:(NSInteger
 
 - (UIView*)tableView:(UITableView*)tableView viewForHeaderInSection:(NSInteger)section {
     //TODO(5.0): localize these!
-    if (section == 0 || (section != 1 && section < tableView.numberOfSections - 1)) {
+    if (section == 0
```
bgerstle: see comment above about factoring these checks into a shared method

Wikipedia/UI-V5/WMFArticleViewController.m
montehurd: haha ok

Wikipedia/UI-V5/WMFArticleViewController.m
montehurd: will take a peek...

Wikipedia/UI-V5/WMFArticleViewController.m
montehurd: So I collapsed it to use && but it can't be single line 'cause it breaks the logic and doesn't work.

Wikipedia/UI-V5/WMFArticleViewController.m
montehurd: Done!

Wikipedia/UI-V5/WMFArticleViewController.m
montehurd: Fixed.

PR #109, opened 2015-09-08
Implement "more like this" view for related & nearby content (T110001)

bgerstle:
Phab task: [T110001](https://phabricator.wikimedia.org/T110001)

- [x] Tapping on the "More like this" button, should push a new VC
- [x] The new VC should be an extended list of the section
- [x] For Related sections, fetch 20 
- [x] For Nearby sections, fetch 20 
- [x] Use the same style cells and headers as the home VC
- [x] Make sure both views start w/ the same results

---

So this turned out to be a big refactor as well. In all the refactoring I verified/fixed:

- [x] Distance display still updates w/ correct locale
- [x] Compass bearing calculations
- [x] Compass orientation issues
- [x] "more like nearby" view doesn't update location while user is on a different tab


Wikipedia/UI-V5/WMFRelatedSectionController.m
```diff
@@ -107,6 +113,21 @@ - (void)configureCell:(UICollectionViewCell*)cell
     }
 }
 
+- (UIViewController*)moreViewController {
```
coreyfloyd: It seems all "more view controllers" should be instances of `WMFArticleListCollectionViewController`. 

In that case, returning a view controller here seems to be more flexible / generic than required. The only thing that should be different for each section is the data source - so it seems appropriate to return that instead. Then we can let the view controller can just inject the data source into the articleListCollectionVC and present it.

Wikipedia/UI-V5/WMFRelatedTitleListDataSource.h
```diff
@@ -0,0 +1,37 @@
...
+- (instancetype)initWithTitle:(MWKTitle*)title
+                    dataStore:(MWKDataStore*)dataStore
+                savedPageList:(MWKSavedPageList*)savedPageList
+    numberOfExtractCharacters:(NSUInteger)numberOfExtractCharacters;
+
+- (instancetype)initWithTitle:(MWKTitle*)title
+                    dataStore:(MWKDataStore*)dataStore
+                savedPageList:(MWKSavedPageList*)savedPageList
+    numberOfExtractCharacters:(NSUInteger)numberOfExtractCharacters
+                      fetcher:(WMFRelatedSearchFetcher*)fetcher NS_DESIGNATED_INITIALIZER;
```
coreyfloyd: Why the second method? If it knows how to create the fetcher… why also allow one to be passed in?

Wikipedia/UI-V5/WMFRelatedTitleListDataSource.m
```diff
@@ -0,0 +1,165 @@
...
+                                                          site:self.title.site];
+            NSError* error;
+            NSDictionary* resultJSON = [MTLJSONAdapter JSONDictionaryFromModel:relatedSearchResult error:&error];
+            if (!resultJSON) {
+                DDLogError(@"Unexpected error re-serializing search result %@. Error: %@",
+                           relatedSearchResult, error);
+                [mutableResults removeObject:relatedSearchResult];
+                return articles;
+            }
+            MWKArticle* article = [[MWKArticle alloc]
```
coreyfloyd: I think we want to move away from the article creation here as we talked about yesterday. Seems like this may be a good opportunity to refactor the articleDataSource to be a titleDataSource instead to prevent us from doing this here.

Wikipedia/UI-V5/WMFRelatedTitleListDataSource.h
bgerstle: For testing

Wikipedia/UI-V5/WMFRelatedTitleListDataSource.m
bgerstle: I'm thinking we should refactor this as a separate PR, as it fits in with the larger model object refactors that need to happen elsewhere, mainly:

- Convert search to use concrete model objects
- Create protocol(s) for "article placeholder" objects
- Break all tight couplings to `MWKArticle`—loosely couple all the things! *
- Implement placeholder handling logic in article container and/or article VC

There are a couple other concerns w.r.t. breaking couplings to `MWKArticle`; I'll list them in the refactoring ticket.

\* This is what would affect the code in question here.

Wikipedia/UI-V5/WMFRelatedTitleListDataSource.h
coreyfloyd: @bgerstle should be in a private header then

Wikipedia/UI-V5/WMFArticleListCollectionViewController.m
```diff
@@ -32,6 +32,10 @@ @interface WMFArticleListCollectionViewController ()
 @implementation WMFArticleListCollectionViewController
 @synthesize listTransition = _listTransition;
 
+- (instancetype)init {
```
coreyfloyd: Are we initing from code now? I guess we don't need a story board anymore?

MediaWikiKit/MediaWikiKit/MWKLocationSearchResult.h
```diff
@@ -2,10 +2,41 @@
...
+ *
+ *  These properties are set automatically by @c WMFNearbyViewModel in response to location & heading updates.
+ *
+ *  @see -[WMFNearbyViewModel autoUpdateResultAtIndex:]
+ */
+
+/**
+ *  Distance in meters from the user's current location.
+ */
+@property (nonatomic, assign) CLLocationDistance distanceFromUser;
```
coreyfloyd: Mutability is your enemy…  

Also this is conflating the user position (which is changing) with a model object that represents just a search result (which is static).

`distanceFromUser` and `bearingToLocation` really should be somewhere more appropriate - maybe another model object, or maybe cache these values in the view model. 

The nearby view model makes more sense as that is where user position and search results are actually used to calculate `distanceFromUser` and `bearingToLocation`.

Wikipedia/UI-V5/WMFHomeNearbyCell.h
```diff
@@ -2,11 +2,20 @@
 #import "WMFSaveableTitleCollectionViewCell.h"
 @import CoreLocation;
 
-@interface WMFHomeNearbyCell : WMFSaveableTitleCollectionViewCell
+@class MWKLocationSearchResult;
 
-@property (copy, nonatomic) NSString* descriptionText;
+@interface WMFHomeNearbyCell : WMFSaveableTitleCollectionViewCell
```
coreyfloyd: Maybe we should rename this to `WMFNearbyCell` since it is used outside of the home VC now?

Wikipedia/UI-V5/WMFHomeNearbyCell.h
```diff
@@ -2,11 +2,20 @@
...
+ *  Display the given search result.
+ *
+ *  Populates the receiver with data in the given result, in addition to binding to properties which are updated
+ *  automatically (i.e. @c distanceFromUser and @c bearingToLocation).
+ *
+ *  @warning Use this instead of @c -setTitle:.
+ *
+ *  @param locationSearchResult The result whose data will be displayed in the receiver.
+ */
+- (void)setLocationSearchResult:(MWKLocationSearchResult*)locationSearchResult withTitle:(MWKTitle*)title;
```
coreyfloyd: With all the other mutable properties on the search result, the title could easily be included as a mutable property as well so we only need to pass one object /endTroll

But seriously… maybe to solve the user position/bearing thing… we can create an object here that holds:

```
MWKLocationSearchResult result
MWKTitle title
CLLocationDegrees distance
CLLocationDistance bearing
```

And pass that into the cell… That object would have everything needed to update the cell.

Wikipedia/UI-V5/WMFHomeViewController.m
```diff
@@ -79,13 +78,21 @@ - (NSString*)title {
 
 #pragma mark - Accessors
 
+- (void)setSavedPages:(MWKSavedPageList* __nonnull)savedPages {
```
coreyfloyd: This line should be unnecessary because this object requires that properties are set before they are used See line 247:

```
 NSParameterAssert(self.dataStore);
 NSParameterAssert(self.searchSite);
 NSParameterAssert(self.recentPages);
 NSParameterAssert(self.savedPages);
```

Nothing should be created (lazy loading) until viewDidAppear. `configureDatasource` isn't called until then.

Wikipedia/UI-V5/WMFLocationManager.m
```diff
@@ -6,8 +6,6 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-static CLLocationDistance WMFMinimumDistanceBeforeUpdatingLocation = 1.0; //meters before we update location
-
 @interface WMFLocationManager ()<CLLocationManagerDelegate>
```
coreyfloyd: So looking at the history of this file I saw one commit about the dealloc method… so that seems good.

But the rest seems to be unexplained in the commit messages and mostly looks to be futzing around with the logic…  

However, I can clearly see that the behavior has been changed. 

Did you fix a bug? What was the bug? What's better now? Did you make tradeoffs?

Hesitant to merge changes in this file without explanations on how the behavior was changed and why.

Wikipedia/UI-V5/WMFNearbySectionController.m
```diff
@@ -79,8 +82,8 @@ - (NSString*)footerText {
 }
 
 - (NSArray*)items {
-    if ([self.nearbyResults.results count] > 0) {
-        return self.nearbyResults.results;
+    if ([self.viewModel.locationSearchResults.results count] > 0) {
```
coreyfloyd: The friends of my friends… are not my friends. 

To much touching here… really need to provide an API on the view model to get what you need.

Wikipedia/UI-V5/WMFNearbyViewModel.m
```diff
@@ -0,0 +1,193 @@
...
+        self.locationManager.delegate = self;
+    }
+    return self;
+}
+
+- (void)dealloc {
+    [self stopUpdates];
+}
+
+- (void)setSite:(MWKSite* __nonnull)site {
```
coreyfloyd: This isn't mutable, so we shouldn't need to override the setter, right?

coreyfloyd: Ahh never mind… it is mutable. I see now. Does this need to be mutable?

Wikipedia/UI-V5/WMFNearbyViewModel.m
```diff
@@ -0,0 +1,193 @@
...
+@interface WMFNearbyViewModel ()
+<WMFNearbyControllerDelegate>
+
+@property (nonatomic, strong) WMFLocationSearchFetcher* locationSearchFetcher;
+@property (nonatomic, strong) WMFLocationManager* locationManager;
+@property (nonatomic, assign) NSUInteger resultLimit;
+
+@property (nonatomic, strong, readwrite) WMFLocationSearchResults* locationSearchResults;
+
+@property (nonatomic, strong) NSMutableIndexSet* autoUpdatingIndexes;
```
coreyfloyd: What determines if a result is "auto-updated" or not. Does this mean some results don't get updates? 

Also `autoUpdatingIndexes ` seems very generic/dubious. Really only the bearing and distance are getting updated right? Or is there something else?

Wikipedia/UI-V5/WMFArticleListCollectionViewController.m
bgerstle: There hasn't yet been a need to instantiate a list VC programmatically.  The current setup seems to support both, since there's basically 0 configuration _of the list VC_ happening in the storyboard.

Wikipedia/UI-V5/WMFHomeNearbyCell.h
bgerstle: Yeah, good point

MediaWikiKit/MediaWikiKit/MWKLocationSearchResult.h
bgerstle: I had considered doing something along the lines of what you're suggesting, but nothing immediately came to mind.  I'll mull this over a bit more, but I'm not sure if the extra indirection is worth it for the "pureness" of a 100% immutable search result.  

> Also this is conflating the user position (which is changing) with a model object that represents just a search result (which is static).

What exactly do you think is confusing about the current properties?  I thought `distanceFromQueryCoordinates` and `distanceFromUser` were pretty distinct.

> The nearby view model makes more sense as that is where user position and search results are actually used to calculate `distanceFromUser` and `bearingToLocation`.

It also already has a reference to the search results, so IMHO it was intuitive to also manage updating their properties in response to events It was already responding to.

Wikipedia/UI-V5/WMFHomeNearbyCell.h
bgerstle: > With all the other mutable properties on the search result, the title could easily be included as a mutable property as well so we only need to pass one object /endTroll

I had considered denormalizing the title into the search result as a convenience, but there's no nice way to do it from the fetcher :trollface: 

Do we want to have `ViewModel` models, and rename the `WMFNearbyViewModel` to be a `WMFNearbyViewModelController`?

Wikipedia/UI-V5/WMFNearbyViewModel.m
```diff
@@ -0,0 +1,193 @@
...
+        /*
+           How often we search titles for a location is separate from how often the location updates.
+           See WMFLocationManager for distanceFilter settings.
+         */
+        DDLogVerbose(@"Not fetching nearby titles for %@ since it is too close to previously fetched location: %@.",
+                     location, self.locationSearchResults.location);
+        return;
+    }
+
+    NSAssert(![self.locationSearchResults.location wmf_isVeryCloseToLocation:location]
```
coreyfloyd: Is there logic prevent very close locations from being passed to this method? If so where? If not, shouldn't we also do the check here instead of just asserting?

Wikipedia/UI-V5/WMFHomeViewController.m
bgerstle: Maybe that wasn't the case when I started the patch, I'd have to look at the commit when this was added for info, but I'll try removing it and seeing what happens :pray:

Wikipedia/UI-V5/WMFLocationManager.m
bgerstle: There was some thrashing here, but the latest commit(s) should have comments explaining the restored legacy behavior:

- Update location every 1m (i.e. `distanceFilter = 1`)
- Re-fetch location search results every 500m (see check in `-[WMFNearbyViewModel fetchTitlesForLocation:]`)

bgerstle: To be clear, the behavior _should_ be identical to master (*cough* should be good for unit testing *cough*)

Wikipedia/UI-V5/WMFNearbySectionController.m
bgerstle: Bahh unnecessary indirection. `WMFLocationSearchResults` is immutable, and all you're asking me to do is write tedious proxy getter/setters when there's nothing to hide.

Wikipedia/UI-V5/WMFNearbyViewModel.m
```diff
@@ -0,0 +1,193 @@
...
+    // TODO: cancel fetch for previous location
+    @weakify(self);
+    [self.locationSearchFetcher fetchArticlesWithSite:self.site location:location resultLimit:self.resultLimit]
+    .then(^(WMFLocationSearchResults* locationSearchResults) {
+        @strongify(self);
+        if (![locationSearchResults.searchSite isEqualToSite:self.site]) {
+            DDLogWarn(@"Ignoring location search results from site %@ since it was changed to %@",
+                      locationSearchResults.searchSite, self.site);
+            return;
+        } else if (![locationSearchResults.location wmf_hasSameCoordinatesAsLocation:self.locationManager.lastLocation]) {
```
coreyfloyd: isn't this unnecessary?

Didn't line 136 accomplish the same thing above?

```
if ([self.locationSearchResults.location distanceFromLocation:location] < 500
              && [self.locationSearchResults.searchSite isEqualToSite:self.site]) {
```

Also don't you need to take the search site into account? If the search site is different than the `self.locationSearchResults.site` then can't the location be the same?

Wikipedia/UI-V5/WMFNearbyViewModel.m
bgerstle: It does, but I actually don't think we'd propagate search site changes here (since nobody has a direct reference to the article list controller that gets pushed)

Wikipedia/UI-V5/WMFNearbyViewModel.m
bgerstle: See the header for methods that start/stop auto updating.  I should've added comments, but it's basically a set of indexes which we're currently updating in response to location & heading changes. The API in the header is documented and provide some rationale.

Wikipedia/UI-V5/WMFNearbyViewModel.m
bgerstle: See my previous comment about the location/search-result update logic. The assertion is more of an extra safety net and expression of intent: we really shouldn't be fetching locations that are very close together.  It was also part of stuff I did while figuring out the aforementioned logic, but it should be better documented in the comment on line 139.

Wikipedia/UI-V5/WMFNearbyViewModel.m
bgerstle: ah ha, but consider the case when it was update _after_ the search was sent! this _should_ eventually be documented in tests, but the logic is effectively:

Fetch new results if:
- I don't have any results
- My site is updated one that's different than where I previously got results
  - Just thought of a couple edge cases on this actually...
- My location updated to one that's > 500m away

I have to check all 3 at the entry point, but also in the async callback to ensure I don't update the list if a new location (or site) is set before the results come back.

Wikipedia/UI-V5/WMFRelatedSectionController.m
bgerstle: just addressed

Wikipedia/UI-V5/WMFRelatedTitleListDataSource.h
bgerstle: @coreyfloyd IMO having designated & convenience initializers is pretty standard, and if we required separate headers for all designated initializers it would be a non-trivial amount of tedious work.  let's discuss on hangout/IRC

bgerstle: I guess I don't think there's anything dangerous about having the designated initializer publicly exposed.  The other option is to consolidate these into one method, and make the fetcher parameter optional (i.e. nullable), which is what I did in `WMFNearbyViewModel` for the location manager.  `WMFNearbyViewModel` however, has its own publicly exposed designated initializer.

Wikipedia/UI-V5/WMFNearbyViewModel.h
```diff
@@ -0,0 +1,130 @@
...
+ */
+- (void)nearbyViewModel:(WMFNearbyViewModel*)viewModel didUpdateResults:(WMFLocationSearchResults*)results;
+
+- (void)nearbyViewModel:(WMFNearbyViewModel*)viewModel didFailWithError:(NSError*)error;
+
+// TODO: separate callback for authorization prompts
+
+@end
+
+@interface WMFNearbyViewModel : NSObject
```
coreyfloyd: I'm not sure about calling this a view model - is it one? 

One responsibility of a view model should be to translate model data into a form that can be displayed on screen. But that is done elsewhere - mostly in the cell classes: that is where strings are created and formatted and where bearings and locations are turned into numbers.

Another way to put it is that view models should be able to act as a proxy for testing the UI… "given this model object, what is the string to be displayed on screen for the title label." At tis point only the cell could actually answer that question.

This seems more like a controller that is coordinating the fetching of models and some location logic to yield models (Search Results). 

Moving to MVVM may be a good idea… but we should be careful in how we name things as we transition to that type of architecture. Calling this a view model sets up expectations that this class doesn't quite hold up to. For now it makes more sense to keep with non-MVVM terminology until we can actually implement something that is a true view model.

Wikipedia/UI-V5/WMFRelatedSectionController.m
```diff
@@ -108,6 +122,21 @@ - (void)configureCell:(UICollectionViewCell*)cell
...
+    #warning Remove SessionSingleton
+    WMFRelatedTitleListDataSource* dataSource =
```
coreyfloyd: is the session singleton necessary? The related section controller already has access to the user data store right? And the user data store has the datastore as a property…

Wikipedia/UI-V5/WMFRelatedTitleListDataSource.h
coreyfloyd: Here it isn't about danger… its about exposing the smallest public API required. Which is one of our best practices.

https://www.mediawiki.org/wiki/Wikimedia_Apps/Team/iOS/BestPractices#What_should_be_public

This also goes back to what testing is for… which is to support development… methods for testing are just that… and should not be exposed to all users of a class.

Wikipedia/UI-V5/WMFRelatedSectionController.m
bgerstle: This will be gone in a second, once I push the commit that swaps `moreViewController` for `extendedListDataSource`.  However, the singleton is still necessary in the nearby data source which has to create its articles and doesn't (yet) have a way to get a data store passed to it.

Instead of adding extra properties to pass it down, I figured it will go away entirely once we remove the need to create articles.

MediaWikiKit/MediaWikiKit/MWKLocationSearchResult.h
coreyfloyd: > What exactly do you think is confusing about the current properties? I thought distanceFromQueryCoordinates and distanceFromUser were pretty distinct.

Those properties are not part of the search result… they come from other places… which without length comments would not be clear whose responsibility it is to set them or when they may be triggered. 

I'm in favor of ditching purity for pragmatism… but I think its a little more than that here… we have properties that are calculated from the ether and then just hung on this object…  

If the creator of the object doesn't have the information to populate the object… that is a bit of a code smell to me.

> It also already has a reference to the search results, so IMHO it was intuitive to also manage updating their properties in response to events It was already responding to.

Yeah… but this wasn't required… the cell was already getting these updates without corrupting the 
model object. And its not clear what the "win" was in refactoring these properties be here instead.

If we are going to dump time into refactoring… we may as well do it right.

Wikipedia/UI-V5/WMFLocationManager.m
coreyfloyd: And then the question is if behavior is identical… then why the diff?

bgerstle: Mostly moving stuff around

bgerstle: Oh, and removing duplicate state (making `lastHeading` and `lastLocation` actual proxy methods that return the underlying location manager's `heading` and `location`).

coreyfloyd: I can see one thing that changed… we are no longer caching the last location… which was intentional… because the current location in the CLLocation manager can be nilled out like when starting and stopping location updates. So now there is now time when the location can be nil in the between refreshes… when before that wasn't true.

bgerstle: The authorization stuff is mostly refactoring (i.e. moving logic out into its own methods instead of doing everything inline in `startMonitoring`)

Wikipedia/UI-V5/WMFNearbyViewModel.m
coreyfloyd: So what does adding the complexity of tracking specific indexes buy us?

It seems like its only needed because we moved the user distance and bearing calculations into this class instead of closer to the view where they are used.

Before… we didn't have to worry because we only updated the cells that are on screen… but now since the logic is deeper… we have to add complexity to track what should be updated… and the logic is "eager" instead of called on demand like before (just update this particular cell please)

So to me it just looks like some of the refactoring is making tradeoffs that I am not sure are better than before…

1. We put the user distance and bearing strings on the search results, making them mutable…
2. So we can set them in this object…
3. Which is too far away from the view to be smart about which objects it should be updating (doesn't know which cells are on screen)…
4. So we add the complexity of tracking indexes.

I'm not sure this is a good trade off. It seems like this class really wants to do some things, but isn't in the right part of the object graph to do them. If this object were not a view model… and were permitted to update the cells… you could avoid some of this complication.

Wikipedia/UI-V5/WMFNearbyViewModel.m
coreyfloyd: We should be deliberate in our assertions (really all code). 

If the condition isn't possible, then don't include it. If it is possible, then we should also handle this gracefully at run time (Another return method).

Wikipedia/UI-V5/WMFNearbyViewModel.m
coreyfloyd: I may not be understanding correctly… but I think i get it.

To me, this suggests you are letting multiple request through for the same location. I don't think that was the case before. This seems like a logic error. 

We should not initiate a second request for a location we are already fetching results for. Its not enough to just ensure we don't update the results… we should never have fetched in the first place.

Wikipedia/UI-V5/WMFNearbyViewModel.m
bgerstle: > If the condition isn't possible, then don't include it. 

Assertions are _meant_ for invariants. That said, in this case it's probably enough to let the conditional handle it and add a test which verifies the desired behavior.

Wikipedia/UI-V5/WMFNearbyViewModel.m
bgerstle: I see what you're saying now. That second conditional isn't supposed to be there.  I was thinking it would be like how we manage cancellation logic for `MWKImage`, but it's not applicable here since we don't have an accurate representation of the "last requested location"—though we could.

Wikipedia/UI-V5/WMFNearbySectionController.m
coreyfloyd: @bgerstle Yeah… i think you are right - covering them up necessarily fixing much.

This really about the code smell - inappropriate intimacy. 

Here I think its a clue about the abstraction not quite bing right. Why does this object need to reach that far down to get it info it needs to do its job? 

One thought may be that the NearbySectionController should defer *more* responsibility to the "view model". Like the nearby section controller should be even a thinner veneer over the view model. 

Another thought is that maybe the section controller should be implemented by a subclass of the view model or implemented by the view model itself.

Wikipedia/UI-V5/WMFLocationManager.m
```diff
@@ -68,14 +91,56 @@ - (void)startLocationUpdates {
...
+                                                           queue:nil
+                                                      usingBlock:^(NSNotification* note) {
+        @strongify(self);
+        [self updateHeadingOrientation];
+    }];
+    [self updateHeadingOrientation];
     [self.locationManager startUpdatingHeading];
 }
 
+- (void)updateHeadingOrientation {
```
coreyfloyd: Oh nice… does this prevent us from having to do the conversion manually later?

Wikipedia/UI-V5/WMFLocationSearchFetcher.h
```diff
@@ -1,16 +1,17 @@
...
 
-@property (nonatomic, assign) NSUInteger maximumNumberOfResults;
-
-- (AnyPromise*)fetchArticlesWithSite:(MWKSite*)site location:(CLLocation*)location;
-
-@property (nonatomic, assign, readonly) BOOL isFetching;
+- (AnyPromise*)fetchArticlesWithSite:(MWKSite*)site
+                            location:(CLLocation*)location
+                         resultLimit:(NSUInteger)resultLimit
+                         cancellable:(inout id<Cancellable> __nullable* __nullable)outCancellable;
```
coreyfloyd: holy shit… this is what you have to write to pass a pointer with nullability turned on?

bgerstle: Yes, but I think the ugliness comes from using an `id`. Might be able to sugar-coat it by using it a `typedef`.

coreyfloyd:
@bgerstle are you still going to do anything about the location manager (category or subclass instead of duplicating the delegate API) or did you decide to leave it how it is for now?

Wikipedia/UI-V5/WMFLocationManager.m
bgerstle: yep, removed that, so now compass always points where it's supposed to whether we're face down/up, portrait/landscape, etc.

bgerstle:
[punting :football:](https://phabricator.wikimedia.org/T112355) but could tackle it now. whatevs

PR #112, opened 2015-09-10
Migrate to Xcode 7 / Swift 2.0

coreyfloyd:
- [x] [set `osx_image` in `.travis.yml` to Xcode 7 GM](http://docs.travis-ci.com/user/languages/objective-c/#Xcode-7).
- [x] commit Frameworks built by Carthage to the repo 

---

Sooooo close!

Some notes:
Ran the Swift conversion that seems to have worked.

But, upgrading Cocoapods is a no go… after 0.38.2 there is a build error where Mantle and extobjc have a conflict because of how imports are performed. I searched/played around for a while to try and resolve this to no avail. It just seems like something changed in the setup of the Xcode projects / targets that cocoapods creates - I'm not exactly sure what it is.

The rest of this follows from there…

Because we can't upgrade Cocoapods, we can't use Cocoapods to integrate PromiseKit because PromiseKit is Swift and requires a newer version of Cocoapods (and the use of frameworks).

…hence Carthage. Carthage is now setup for integrating Promisekit… and seems to work.

However, it seems the Swift code can't see the PromiseKit headers. So we still have build errors. It's late but it feels like I am missing something simple here.

@bgerstle you want to take a peek at this when you get a chance? Maybe your Swift Compiler Fu can get this working.


bgerstle:
Reiterating: we should [update the `.travis.yml` to specify Xcode 7](http://docs.travis-ci.com/user/languages/objective-c/#Xcode-7).

coreyfloyd:
@bgerstle these is a class of problems - about 5 errors I am not sure how to address - can you take a look and see if you can figure it out?

screenshot 2015-09-12 10 45 14

bgerstle:
i'm on it :shipit:

bgerstle:
FYI, when the error panel doesn't provide enough info, jumping to the "Report navigator" [0] and expanding the message [1] often reveals more clues from the compiler that Xcode isn't bubbling up for you. [2]

0: 
image

1: 
image

2:
image

bgerstle:
getting closer... just getting an completely opaque seg fault compilation error in `WMFLegacyImageDataMigration`. can take a look tomorrow

bgerstle:
closer..... now hitting seg fault again w/ a couple test files. also need to merge master in

bgerstle:
FINALLY!!! :rocket: 
IT'S HAPPENING

PR #114, opened 2015-09-10
Fix bad recent searches logic

coreyfloyd:
Additionally moved over to using the actual recent searches list instead of the custom dictionary persistence that was used in the recent searches VC.


coreyfloyd:
@bgerstle pushed…

PR #117, opened 2015-09-16
Fix crashes in the layout - again… no really… again

coreyfloyd:
This time I properly forked the repo and fixed the issue at the source.
Updated the pods
removed custom subclass of layout


montehurd:
Gonna merge for now so I can submit my patch for corey's transition work on top of it.

PR #122, opened 2015-09-18
Search messages

montehurd:
https://phabricator.wikimedia.org/T104858

Temporarily show old-style alerts for search until we can restyle alerts.

**Show alert when no search results were found:**
screen shot 2015-09-18 at 1 33 07 pm
**Show alert relaying any error messages, such as connectivity problems:**
(this should also cause the *"Search is currently too busy.  Please try again later."* message from the phab ticket to be shown properly)
screen shot 2015-09-18 at 1 33 22 pm


bgerstle:
looks good! :+1:

PR #124, opened 2015-09-22
Hide tab bar when article is pushed onto nav stack

coreyfloyd:
nuf said


coreyfloyd:
screenshot 2015-09-22 13 18 45

PR #125, opened 2015-09-22
Add tool bar

coreyfloyd:
Adding the tool bar for the article VC and moved the save button into the tool bar

screenshot 2015-09-22 13 35 27

### DIFF
--- a/Wikipedia/UI-V5/WMFArticleContainerViewController.m
+++ b/Wikipedia/UI-V5/WMFArticleContainerViewController.m
@@ -163,7 +163,6 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (void)viewWillDisappear:(BOOL)animated{
     [super viewWillDisappear:animated];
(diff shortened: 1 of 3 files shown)


montehurd:
If you tap on links on a native card, the toolbar disappears and then animates up from the bottom each time, which looks a bit odd.

bgerstle:
:+1: I'm fine with this if it meets user requirements, but if we stick with this design I'd like to consider refactoring the responsibility into a `UINavigationController` subclass. @montehurd if you tested and looks good, +/-1 and merge at will.

The rationale being that a **declarative** API (e.g. `hidesBottomBarWhenPushed`) allows content view controllers to be simpler (only declaring their desired state), thus moving logic into a central navigation controller.  We can easily test that the toolbar and tabbar state are correct as a result of view controller transitions in our custom navigation controller.

bgerstle:
@montehurd did you want to do anything now that @coreyfloyd pushed another commit?

PR #126, opened 2015-09-22
Made MWKArticle isCached method not load section strings from disk.

montehurd:
Much faster.

https://phabricator.wikimedia.org/T111543

### DIFF
--- a/MediaWikiKit/MediaWikiKit/MWKDataStore.h
+++ b/MediaWikiKit/MediaWikiKit/MWKDataStore.h
@@ -154,4 +154,6 @@ extern NSString* const MWKArticleKey;
 
 - (NSError*)removeFolderAtBasePath;
 
+- (BOOL)htmlFileExistsForSectionId:(NSUInteger)sectionId title:(MWKTitle*)title;
+
 @end
--- a/MediaWikiKit/MediaWikiKit/MWKDataStore.h
+++ b/MediaWikiKit/MediaWikiKit/MWKDataStore.h
@@ -154,6 +154,6 @@ extern NSString* const MWKArticleKey;
 
 - (NSError*)removeFolderAtBasePath;
 
-- (BOOL)htmlFileExistsForSectionId:(NSUInteger)sectionId title:(MWKTitle*)title;
+- (BOOL)hasHTMLFileForSection:(MWKSection*)section;
 
 @end
--- a/MediaWikiKit/MediaWikiKit/MWKDataStore.m
+++ b/MediaWikiKit/MediaWikiKit/MWKDataStore.m
@@ -334,8 +334,8 @@ static NSString* const MWKImageInfoFilename = @"ImageInfo.plist";
     return html;
 }
 
-- (BOOL)htmlFileExistsForSectionId:(NSUInteger)sectionId title:(MWKTitle*)title {
-    NSString* path     = [self pathForSectionId:sectionId title:title];
+- (BOOL)hasHTMLFileForSection:(MWKSection*)section {
+    NSString* path     = [self pathForSectionId:section.sectionId title:section.title];
     NSString* filePath = [path stringByAppendingPathComponent:@"Section.html"];
     return [[NSFileManager defaultManager] fileExistsAtPath:filePath];
 }
--- a/MediaWikiKit/MediaWikiKit/MWKDataStore.m
+++ b/MediaWikiKit/MediaWikiKit/MWKDataStore.m
@@ -334,6 +334,12 @@ static NSString* const MWKImageInfoFilename = @"ImageInfo.plist";
     return html;
 }
 
+- (BOOL)htmlFileExistsForSectionId:(NSUInteger)sectionId title:(MWKTitle*)title {
+    NSString* path     = [self pathForSectionId:sectionId title:title];
+    NSString* filePath = [path stringByAppendingPathComponent:@"Section.html"];
+    return [[NSFileManager defaultManager] fileExistsAtPath:filePath];
+}
+
 - (MWKImage*)imageWithURL:(NSString*)url article:(MWKArticle*)article {
     if (url == nil) {
         return nil;
--- a/MediaWikiKit/MediaWikiKit/MWKSection.m
+++ b/MediaWikiKit/MediaWikiKit/MWKSection.m
@@ -107,7 +107,7 @@ NSString* const MWKSectionShareSnippetXPath = @"/html/body/p[not(.//span[@id='co
 }
 
 - (BOOL)hasTextData {
-    return [self.article.dataStore htmlFileExistsForSectionId:self.sectionId title:self.article.title];
+    return [self.article.dataStore hasHTMLFileForSection:self];
 }
 
 - (NSString*)text {
--- a/MediaWikiKit/MediaWikiKit/MWKSection.m
+++ b/MediaWikiKit/MediaWikiKit/MWKSection.m
@@ -106,6 +106,10 @@ NSString* const MWKSectionShareSnippetXPath = @"/html/body/p[not(.//span[@id='co
     }
 }
 
+- (BOOL)hasTextData {
+    return [self.article.dataStore htmlFileExistsForSectionId:self.sectionId title:self.article.title];
+}
+
 - (NSString*)text {
     if (_text == nil) {
         _text = [self.article.dataStore sectionTextWithId:self.sectionId article:self.article];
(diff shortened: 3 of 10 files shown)


WikipediaUnitTests/Utilities/XCTestCase+MWKFixtures.h
```diff
@@ -36,7 +36,7 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (MWKArticle*)articleWithMobileViewJSONFixture:(NSString*)fixtureName
                                       withTitle:(id)titleOrString
-                            insertIntoDataStore:(MWKDataStore*)dataStore;
+                                      dataStore:(MWKDataStore*)dataStore;
```
montehurd: Per Brian's suggestion. "Insert" wasn't entirely accurate because it doesn't call "[article save]".

MediaWikiKit/MediaWikiKit/MWKDataStore.m
```diff
@@ -334,6 +334,12 @@ - (NSString*)sectionTextWithId:(NSUInteger)sectionId article:(MWKArticle*)articl
     return html;
 }
 
+- (BOOL)htmlFileExistsForSectionId:(NSUInteger)sectionId title:(MWKTitle*)title {
+    NSString* path     = [self pathForSectionId:sectionId title:title];
```
montehurd: Note: we want this to return YES even if the "Section.html" file is empty - this is because some sections *do* have zero length text strings - such as sections which have immediate sub-sections. If we didn't do this articles having such sections would *never* appear to be cached.

montehurd: I added a test to cover this :)

MediaWikiKit/MediaWikiKit/MWKSection.m
```diff
@@ -106,6 +106,10 @@ - (MWKTitle*)sourceTitle {
     }
 }
 
+- (BOOL)hasTextData {
+    return [self.article.dataStore htmlFileExistsForSectionId:self.sectionId title:self.article.title];
+}
```
montehurd: Note: we want this to return YES even if the "Section.html" file is empty - this is because some sections *do* have zero length text strings - such as sections which have immediate sub-sections. If we didn't do this articles having such sections would *never* appear to be cached. I added a test to cover this :)

WikipediaUnitTests/Utilities/XCTestCase+MWKFixtures.h
bgerstle: > wasn't entirely accurate

In fact, it was dead wrong 😅

MediaWikiKit/MediaWikiKit/MWKSection.m
bgerstle: minor question, is it possible to simplify to `hasHTMLFileForSection:(MWKSection*)` and query the section's `id` and `title` to get the path to the file w/in the data store?

MediaWikiKit/MediaWikiKit/MWKDataStore.h
```diff
@@ -154,4 +154,6 @@ extern NSString* const MWKArticleKey;
 
 - (NSError*)removeFolderAtBasePath;
 
+- (BOOL)htmlFileExistsForSectionId:(NSUInteger)sectionId title:(MWKTitle*)title;
```
bgerstle: Cocoa comment: the convention for methods returning `BOOL` to have a verb prefix (e.g. `is`, `has`, `contains`). For example: `-[UIViewController isViewLoaded]`, `-[NSString hasPrefix:]`, and `-[NSArray containsObject:]`.

MediaWikiKit/MediaWikiKit/MWKSection.m
bgerstle: @montehurd thanks for the heads-up (for posterity, even though we talked about it over hangout), but this info should generally be evident in comments and/or tests.  The idea situation is for the code or tests to be as expressive and self-documenting as possible.

WikipediaUnitTests/MWKSectionHasTextDataTests.m
```diff
@@ -0,0 +1,67 @@
...
+- (MWKArticle*)getTestingArticle {
+    MWKTitle* title     = [MWKTitle titleWithString:@"Barack Obama" site:[MWKSite siteWithDomain:@"wikipedia.org" language:@"en"]];
+    MWKArticle* article = [self articleWithMobileViewJSONFixture:@"Obama" withTitle:title dataStore:self.session.dataStore];
+    [article save];
+    return article;
+}
+
+- (void)testArticleHasAtLeastOneSectionWithZeroLengthHTML {
+    // Ensure at least one section of article has zero length section html.
+    MWKArticle* article              = [self getTestingArticle];
```
bgerstle: It looks like you're testing that the test fixture has at least one non-empty section.  It's good that you thought to verify this, but having it as a standalone test is misleading.  It would be clearer if this were an assertion w/in a test that  requires the fixture to be non-empty.  e.g.

```objc
- (void)testSomethingExpectedWithCachedArticle {
    MWKArticle* myArticle = [...];
    NSAssert(myArticle.isCached, @"This test requires the fixture to have a non-empty section!");
    // exercise myArticle and verify functionality
}
```

WikipediaUnitTests/MWKSectionHasTextDataTests.m
```diff
@@ -0,0 +1,67 @@
...
+        if (section.text.length == 0) {
+            atLeastOneZeroLengthSection = YES;
+            break;
+        }
+    }
+    assertThat(@(atLeastOneZeroLengthSection), isTrue());
+}
+
+- (void)testHasTextDataMethodReturnsYESforZeroLengthSectionHTML {
+    // Ensure "[MWKSection hasTextData]" returns YES if section html isn't nil - even if it's a zero length string.
```
bgerstle: it's tough to come up w/ salient names for `XCTestCase` test methods, but I think it might be worth considering breaking this up into a few of different tests:

```objc
- (void)testArticleWithZeroSectionsIsNotCached;
- (void)testArticleWithEmptyStringSectionIsCached;
- (void)testArticleWithNonEmptyStringSectionIsCached;
```

Separating it out into different tests allows you to cover each scenario & code path separately, resulting in clearer tests w/ fewer assertions.  It might a bit of extra work to fabricate articles for these specific cases (since there's no designated initializer for an article that lets you inject it with all its properties), so do whatever's easiest for now.

WikipediaUnitTests/MWKSectionHasTextDataTests.m
```diff
@@ -0,0 +1,67 @@
...
+    MWKArticle* article = [self getTestingArticle];
+    for (MWKSection* section in article.sections) {
+        /*
+           Reminder: zero length strings are *valid* section text data!
```
bgerstle: Technically, this might belong in the documentation of `-[MWKSection text]`.  That is, `text` is a non-null string, which implies that an empty string is valid.  The caveat here is that what we're really considering (and there's no great place to document this yet) is that we consider any file written to the data store as a signal that we've downloaded **all** the data for that section.  I hope to eventually write up some high-level architecture documentation about all of this, but the data layer's documentation might have to wait until it's rewritten ;-).

WikipediaUnitTests/MWKSectionHasTextDataTests.m
montehurd: Oh man! That's what I did initially :) I see your point.

bgerstle:
:+1: looks great! 👷 just left a couple of suggestions.

WikipediaUnitTests/MWKSectionHasTextDataTests.m
montehurd: Fixed!

WikipediaUnitTests/MWKSectionHasTextDataTests.m
montehurd: Fixed!

MediaWikiKit/MediaWikiKit/MWKSection.m
montehurd: Oh good points!!! Fixing...

montehurd: Fixed!

MediaWikiKit/MediaWikiKit/MWKDataStore.h
montehurd: Fixed!

WikipediaUnitTests/MWKSectionHasTextDataTests.m
montehurd: I wasn't sure of the best way to handle this - I wanted to test the behavior of MWKSection's "hasTextData" method specifically because if it broke, the MWKArticle isCached would then also break...

montehurd:
@bgerstle Hey hold off on merging this... just ran performance tests and am actually seeing a slowdown... wft?

PR #127, opened 2015-09-23
Download articles as they are saved (T111831)

bgerstle:
Phab: [T111831](https://phabricator.wikimedia.org/T111831) [T113481](https://phabricator.wikimedia.org/T113481)

- [x] Download article when added to saved pages
- [x] Cancel download when removed from saved
- [x] Download images after article is saved
- [x] <strike>Skip memory cache when downloading images</strike> **not doing this anymore**

##### Amendments
- [x] Revert to conventional `SDWebImage` downloading & caching. Can run tests later if there appears to be a performance problem ddc3a52
- [x] Move legacy image data migration & saved articles fetcher into root VC (see **Where does `SavedArticlesFetcher` live?** below)
- [x] Remove legacy saved pages VC & make `SavedArticlesFetcher` immutable 4077959

---

No screenshots for this one, mostly technical.  But, there are tests!  And console logs! 👾  See `SavedArticleFetcherTests` for details.

#### Should we be downloading articles serially? 
**Proposal:** We're already limiting the number of concurrent operations at:
- The article level (which uses the default value based on the system)
- The image level (which SDWebImage defaults to 6)

<strike>
I have a feeling doing the all at once will lead to CPU/network starvation for large lists that aren't downloaded.  Ideally we'd have an algorithm to ensure we're saturating some percentage of available bandwidth, but doing it serially is probably the simplest/safest route.
</strike>

#### Where does `SavedArticlesFetcher` live?
**Proposal:** Put it in the app's root view controller, since its responsibilities include creating the "root object graph" and other "root" model responsibilities.  When we get around to background fetching, we can delegate that event to the rootVC.
<strike>
As mentioned in the commit, I put it in the `AppDelegate` for a couple reasons:

- It's where we do stuff on launch (image data migration, pywik, etc.)
- `AppDelegate` already gets called to start a background fetch. Other option is to respond to notifications.
- Don't like the idea of putting in a view controller, but we need to talk about the progress API, probably once we get final designs for the new saved pages tab. Either way, I'd prefer saved pages (or anything else w/ a save button) to not touch the fetcher.
</strike>

---

#### Next steps w/ Saved Pages
- Re -implementing/-integrating "delete all"
- Progress UI/UX
- Removing legacy code
- Checking WiFi vs. 4G availability
- Fetching in the background 

Plan to make tickets for all of these. Let me know if you think I missed anything.


Wikipedia/Networking/Fetchers/SavedArticlesFetcher.h
```diff
@@ -17,14 +23,14 @@
...
-@property (nonatomic, strong, readonly) MWKSavedPageList* savedPageList;
-@property (nonatomic, strong, readonly) MWKDataStore* dataStore;
-
-- (instancetype)initWithDataStore:(MWKDataStore*)dataStore;
+@property (nonatomic, strong) MWKSavedPageList* savedPageList;
 
 @property (nonatomic, weak) id<SavedArticlesFetcherDelegate> fetchFinishedDelegate;
 
-- (void)fetchSavedPageList:(MWKSavedPageList*)savedPageList;
+- (instancetype)initWithDataStore:(MWKDataStore*)dataStore;
```
coreyfloyd: These 2 initializers seem very different. 
We should create an initializer in the way that we intend on using it in the app. 

if you are creating extra initializers for testing, as I have mentioned before, they should be in private headers and not in the public API. 

We should always strive for the most obvious well understood public APIs in the classes we create - these 2 initializers in particular make it very non-obvious which is the right one to use.

Wikipedia/Networking/Fetchers/SavedArticlesFetcher.h
```diff
@@ -17,14 +23,14 @@
 + (SavedArticlesFetcher*)sharedInstance;
```
coreyfloyd: It we are creating an ivar to hold the reference to the long living saved fetcher (currently in the app delegate) - we should remove the shared access to it, since it is now duplicative.

Wikipedia/Networking/Fetchers/SavedArticlesFetcher.h
```diff
@@ -17,14 +23,14 @@
...
 + (void)                 setSharedInstance:(SavedArticlesFetcher*)fetcher;
 
-@property (nonatomic, strong, readonly) MWKSavedPageList* savedPageList;
-@property (nonatomic, strong, readonly) MWKDataStore* dataStore;
-
-- (instancetype)initWithDataStore:(MWKDataStore*)dataStore;
+@property (nonatomic, strong) MWKSavedPageList* savedPageList;
```
coreyfloyd: This never changes right? Shouldn't just be read only and be passed in on initialization?

Wikipedia/Networking/Fetchers/SavedArticlesFetcher.m
```diff
@@ -27,59 +24,188 @@ @implementation SavedArticlesFetcher
...
 }
 
-- (void)fetchSavedPageList:(MWKSavedPageList*)savedPageList {
-    self.savedPageList = savedPageList;
+- (instancetype)initWithDataStore:(MWKDataStore*)dataStore {
+    return [self initWithArticleFetcher:[[WMFArticleFetcher alloc] initWithDataStore:dataStore]
+                        imageController:[WMFImageController sharedInstance]];
+}
+
+- (void)setSavedPageList:(MWKSavedPageList*)savedPageList {
```
coreyfloyd: This should also get removed since we don't need a setter to do all this book keeping since the saved page list should never change. 

This back to Immutability - why would we make this mutable and add the additional book keeping when we know in practice we will never change this property.

Wikipedia/Networking/Fetchers/SavedArticlesFetcher.m
```diff
@@ -27,59 +24,188 @@ @implementation SavedArticlesFetcher
...
+- (void)fetchSavedPageList {
+    if (!self.savedPageList) {
+        return;
+    }
+
+    // build up initial state of current list
+    [self startFetching];
+
+    // observe subsequent changes
+    [self.KVOControllerNonRetaining observe:self.savedPageList
```
coreyfloyd: Why wouldn't we just always observe the saved page list - why only start observing it in the fetch method.

The method seems to be conflating 2 discrete pieces of work:
- Observing the saved page list, and responding to changes
- Downloading articles for items in the saved page list

This conflation is demonstrated by the name of the method, `fetchSavedPageList`, which gives no indication that observation logic is going on in here.

Wikipedia/Networking/Fetchers/SavedArticlesFetcher.m
```diff
@@ -27,59 +24,188 @@ @implementation SavedArticlesFetcher
...
+                                    keyPath:WMF_SAFE_KEYPATH(self.savedPageList, entries)
+                                    options:NSKeyValueObservingOptionOld | NSKeyValueObservingOptionNew
+                                     action:@selector(savedPageListDidChange:)];
+}
+
+- (void)startFetching {
```
coreyfloyd: `startFetching` should probably convey the fact that we are fetching all titles. Something like `fetchArticlesForAllSavedPages` would be more clear.

coreyfloyd: Or maybe: `fetchArticlesForUncachedSavedPages` would be even more clear

Wikipedia/Networking/Fetchers/SavedArticlesFetcher.m
```diff
@@ -27,59 +24,188 @@ @implementation SavedArticlesFetcher
...
+}
+
+- (void)savedPageListDidAddItems:(NSArray<MWKSavedPageEntry*>*)insertedEntries {
+    if (!insertedEntries.count) {
+        return;
+    }
+    [self fetchTitles:[insertedEntries valueForKey:WMF_SAFE_KEYPATH([MWKSavedPageEntry new], title)]];
+}
+
+- (void)mergeFetchesWithUpdatedSavedPageList {
```
coreyfloyd: Extra logic (and complexity) for dealing with a situation that in practice will never occur only because we exposed the savedPageList as readwrite in the public API.

Wikipedia/AppDelegate.m
```diff
@@ -60,11 +63,21 @@ - (WMFLegacyImageDataMigration*)imageMigration {
     return _imageMigration;
 }
 
+- (SavedArticlesFetcher*)savedArticlesFetcher {
```
coreyfloyd: I don't think this is a good pattern to start. A while ago we pulled a bunch of logic out of the app delegate and scoped the app delegates purpose to simply:

- Being the delegate of the app (implementing the callbacks from the UIApplication)
- Starting the App View Controller
- Starting Crash reporting/analytics - because we need to start it as soon as possible to catch crashes

Importantly the App Delegate is pretty devoid of business logic (aside from image migration, which actually needs to be moved to where the rest of the migration logic is - in the AppViewController)

Currently, the AppViewController is the place where the data store is inited and other list and model controllers are injected into other view controllers that need them. this basically in the App View Controller's job description - it is responsible for the App up and running by injecting the model controllers and models into the view controllers. 

Having said that, if you want to propose a new architecture for holding onto long living model controllers / business logic-y classes in a place different from the App View Controller, I'd be open to it and we should definitely discus at code review. In the mean time we should keep such logic in the same place and not start competing architectures.

Wikipedia/Images/WMFImageController.swift
```diff
@@ -124,14 +123,54 @@ public class WMFImageController : NSObject {
...
 
-    /// MARK: - Simple Fetching
+    /**
+    Use the shared session to download an image directly to the receiver's disk cache.
+
+    - parameter url: The URL to download the image from.
+
+    - returns: A promise which is resolved after the data has been successfully written to disk.
+    */
+    public func downloadImageDataInBackground(url: NSURL) -> Promise<Void> {
```
coreyfloyd: Some thoughts on the name here:
- It's not image data, its the image right?
- By background you mean low priority? Maybe we should say low priority - background just sounds like a background thread.
- The fact that this bypasses the memory cache seems really important.
- Other methods say `fetch`, not `download`

With this in mind something like `fetchImageWithLowPriorityAndCacheToDisk` or `fetchImageWithLowPriorityBypassingDiskCache` or `fetchImageWithLowPriority(url: NSURL cacheToDisk:BOOL cacheInMemory:BOOL)` seems more appropriate.

Wikipedia/Images/WMFImageController.swift
```diff
@@ -124,14 +123,54 @@ public class WMFImageController : NSObject {
...
+            return Promise()
+        }
+
+        let (promise, fulfill, reject) = Promise<NSURL>.pendingPromise()
+
+        /*
+          FIXME: Using sharedSession to punt on refactoring this to work with background sessions (which require using
+          a delegate).
+        */
+        let task = NSURLSession.sharedSession()
```
coreyfloyd: Why aren't we using the SDImageImageDownload here? We built all of our image download and caching logic on SDWebImage - we really shouldn't start circumventing it without good reason. 

For sure SDWebImage supports low priority downloads in the background. So it at least isn't obvious why we have deviated here.

Wikipedia/Images/WMFImageController.swift
coreyfloyd: Wondering why doesn't the method and `fetchImageWithURL` method call through to the same logic? The only thing that should be different is the priority and the cache destinations - which are all configurable using SDWebImage's Options instead of the defaults:

`SDWebImageContinueInBackground`
`SDWebImageCacheMemoryOnly`
`SDWebImageLowPriority`

bgerstle: I could investigate how much work it would be to add the capability of background, direct-to-disk downloads in `SDWebImage`, but my gut tells me it won't be easy and it doesn't get us extra affordances of using a background URL session (which has a lot more functionality than a simple `NSURLConnection` wrapped in a background task).

bgerstle: IMHO, the amount of work we'd have to do (or have done) to customize SDWebImage is approaching (if not past) the point of diminishing returns. The author even mentions that [you probably don't need it](https://github.com/rs/SDWebImage/wiki/How-is-SDWebImage-better-than-X%3F#since-afnetworking-provides-similar-functionality-for-uiimageview-is-sdwebimage-still-useful) if you're using a library like `AFNetworking` which provides in-memory caching.  That, coupled with the fact that we're using _very little_ of `SDWebImage` disk cache logic (i.e. not letting it cleanup, customizing the location), I'm starting to think we should contemplate using plain old `NSURLSession` (either directly or through something else) and dropping SDWebImage.

I'm also considering this for our networking stack, but I've yet to closely inspect what AFNetworking features we're actually benefiting from compared to the cost of add'l abstraction.

Ranting aside, this approach is:
- Only in the 10's of lines of code
- Optimized to avoid unnecessary image inflation & memory costs
- Debatably easier to move to using a background `NSURLSession`, which could let us reap add'l benefits for users (by making these & other tasks low-priority, discretionary tasks that can continue in the background **and launch the app to trigger background fetches**)

Wikipedia/Images/WMFImageController.swift
bgerstle: See my comment above, I don't think it would be trivial to customize SDWebImage to write directly to disk, but I'll to a quick peek and see what I can find.

Wikipedia/Networking/Fetchers/SavedArticlesFetcher.h
bgerstle: This, along with your other comments about setter methods and other things, are tied to the root cause of me not wanting to break _too_ much of the existing API, and thereby also having to fix or delete the legacy `SavedPagesViewController`. If we're all on the same page that it's going to be deleted eventually, I can make more drastic changes to the API which assume:

- The saved page list never changes*
- There's no `sharedInstance` of this class

\* This will probably be true for quite some time, but I hope we'll eventually do something crazy like support multiple logins & data stores

Wikipedia/Networking/Fetchers/SavedArticlesFetcher.m
bgerstle: Not true, this _should_ happen if someone calls `removeAllEntries` on the `savedPageList`, i.e. if we implement "delete all saved pages."  I put quickly wrote something that _should_ work, but didn't bother testing it until we've added the functionality.

coreyfloyd:
:+1:

PR #128, opened 2015-09-23
Add share to tool bar

coreyfloyd:
Most of this ended up being a refactor to get most of the sharing logic in the same place.

Basically all sharing logic is now in the WMFShareOptionsController (which is no longer a view controller, because it want being used as one anyways)

WMFShareOptionsController has also been refactored to no longer require a delegate.

The share funnel has also pretty much been moved inside of the share options controller - EXCEPT - there is some logging that can't be moved (it involves selection of text in the webview) - so that object is passed to the options controller on initialization.

(In case you were wondering, we can't move to using the UIAlertController - because you can't customize it to add images)

While I was in there:
I moved auto layout to masonry
Made the options menu fill the width of the screen
I fixed the display / dismiss animation to look native
Added the camera roll as an option for output (https://phabricator.wikimedia.org/T37764)
Did some minor clean up of the web view controller in regards to singleton references


bgerstle:
> Added the camera roll as an option for output (https://phabricator.wikimedia.org/T37764)

Does that save the share card, or the lead image?  Might not be quite in line with that that ticket (and OTRS tickets I've seen) are asking for, which is a long press on images to save.

bgerstle:
Not sure if I agree about having saved pages logging on the button controller.  For example, this wouldn't account for (eventual) deletion of saved pages en masse.  We could have a separate class which logs events as they happen to the list itself (similar to how I'm downloading articles in response to KVO changes in #127).

Wikipedia/View Controllers/ShareCard/WMFShareOptionsController.m
```diff
@@ -0,0 +1,346 @@
...
+    self.containerViewController = viewController;
+    self.originButtonItem = nil;
+    self.originView = view;
+    [self fetchImageThenShowShareCard];
+}
+
+#pragma mark - Asynchornous Fetch and Present
+
+- (void)fetchImageThenShowShareCard {
+    [[WMFImageController sharedInstance] fetchImageWithURL:[NSURL wmf_optionalURLWithString:self.article.imageURL]].then(^(WMFImageDownload* download){
```
bgerstle: need weak/strong-ify

Wikipedia/View Controllers/ShareCard/WMFShareOptionsController.m
```diff
@@ -0,0 +1,346 @@
...
+        [self showShareOptionsWithImage:download.image];
+    }).catch(^(NSError* error){
+        [self showShareOptionsWithImage:nil];
```
bgerstle: I guess this just shows an empty "card"?

Wikipedia/View Controllers/ShareCard/WMFShareOptionsController.m
```diff
@@ -0,0 +1,346 @@
...
+
+- (void)showShareOptionsWithImage:(nullable UIImage*)image {
+    
+    [self setupBackgroundView];
+
+    _shareImage = [self cardImageWithArticleImage:image];
+    
+    [self setupShareOptionsWithImage:_shareImage];
+    
+    [self presentShareOptionsView:self.shareOptions];
```
bgerstle: if we're only passing an ivar you can remove the argument

Wikipedia/View Controllers/ShareCard/WMFShareOptionsController.m
```diff
@@ -0,0 +1,346 @@
...
+    grayOverlay.backgroundColor = [[UIColor blackColor] colorWithAlphaComponent:0.42];
+    grayOverlay.alpha           = 0.0;
+    [containingView addSubview:grayOverlay];
+    self.grayOverlay = grayOverlay;
+    UITapGestureRecognizer* tapRecognizer = [[UITapGestureRecognizer alloc]
+                                             initWithTarget:self action:@selector(respondToDimAreaTapGesture:)];
+    [grayOverlay addGestureRecognizer:tapRecognizer];
+    grayOverlay.translatesAutoresizingMaskIntoConstraints = NO;
+    [grayOverlay mas_makeConstraints:^(MASConstraintMaker *make) {
+
```
bgerstle: extra whitespace! :scream:

Wikipedia/View Controllers/ShareCard/WMFShareOptionsController.m
```diff
@@ -0,0 +1,346 @@
...
+    shareOptionsView.cardImageView.image                           = _shareImage;
+    
+    [self.containerViewController.view addSubview:shareOptionsView];
+    self.shareOptions = shareOptionsView;
+}
+
+#pragma mark - Create Card Image
+
+- (UIImage*)cardImageWithArticleImage:(UIImage*)image{
+    WMFShareCardViewController* cardViewController =
```
bgerstle: is the end goal to eventually remove this, similar to how the view is being loaded directly from a nib above?

Wikipedia/View Controllers/ShareCard/WMFShareOptionsController.m
```diff
@@ -0,0 +1,346 @@
...
+        self.shareOptions = nil;
+    }];
+}
+
+
+#pragma mark - Tap Gestures
+
+- (void)respondToDimAreaTapGesture:(UITapGestureRecognizer*)recognizer {
+    [self.shareFunnel logAbandonedAfterSeeingShareAFact];
+    [[WMFImageController sharedInstance] cancelFetchForURL:[NSURL wmf_optionalURLWithString:self.article.imageURL]];
```
bgerstle: shouldn't we put this (and any other tear-down logic) into the `dismissShareOptions` method?

bgerstle: or better, in the cleanup method?

Wikipedia/View Controllers/ShareCard/WMFShareOptionsController.m
```diff
@@ -0,0 +1,346 @@
...
+                        permittedArrowDirections:UIPopoverArrowDirectionAny
+                                        animated:YES];
+        }
+    }
+    
+    [self cleanup];
+}
+
+- (void)popoverControllerDidDismissPopover:(UIPopoverController *)popoverController{
+    self.popover = nil;
```
bgerstle: should we be calling cleanup or something here?

bgerstle:
Just got a crash when trying to detect faces on the image. Unrecognized selector where something's trying to get the length of `NSValue` in `-[WMFShareCardImageContainer getPrimaryFocalRectFromCanonicalLeadImage]`.

bgerstle:
Yeah, `allNormalizedFaceBounds` is using `NSValue` now, where before it was an `NSString`. I guess this wasn't caught because share hasn't been hooked up, and opaque arrays prevent type checking from happening.  @coreyfloyd can you change the `allNormalizedFaceBounds` method to return `NSArray<NSValue>` and update any other usages that assume it's a string?

bgerstle:
This one spot might be the only place that I missed, sorry

Wikipedia/View Controllers/ShareCard/WMFShareOptionsController.m
coreyfloyd: Don't really have an end goal beyond just organizing the code

Wikipedia/View Controllers/ShareCard/WMFShareOptionsController.m
coreyfloyd: yeah - it does what it did before…which is a black card with text overlay

coreyfloyd:
@bgerstle In reference to the image camera roll thing… yeah - that was the wrong ticket - I'll have to find it - but there is one specifically for camera roll saving for these images somewhere - I know I triaged it into the product backlog.

Wikipedia/View Controllers/ShareCard/WMFShareOptionsController.m
coreyfloyd: @bgerstle no its cleaned up in the method above

bgerstle: oh, i think i got it. `cleanup` removes the share options view, which doesn't need to happen for a popover?

coreyfloyd:
> Not sure if I agree about having saved pages logging on the button controller. For example, this wouldn't account for (eventual) deletion of saved pages en masse. We could have a separate class which logs events as they happen to the list itself (similar to how I'm downloading articles in response to KVO changes in #127).

@bgerstle I see why you want to put logic in the same place in the model layer, but logging in general is about on UI interaction not about business logic in the model layer. As a practice, I keep logging close to the UI because if you put logging too deep in the stack because you start conflating business logic with user intent. 

In this case, the intent is to capture when a user intends to save a page, NOT about when we run saved page updates in the background. The business logic of downloading saved pages should be able to be started and stopped independently of logging a tap on a save button.

bgerstle:
@coreyfloyd I guess you're right, but this means every save button needs a save button controller—which might not be a bad thing, since the controller actually makes it _easier_ to have a save button in the first place.

bgerstle:
Minor thing, but @coreyfloyd can you resolve this warning in `WMFArticleContainerController`:

> /Users/bgerstle/Code/wikipedia-ios/Wikipedia/UI-V5/WMFArticleContainerViewController.m:56:1: Synthesized setter 'setShareFunnel:' for null_resettable property 'shareFunnel' does not handle nil

Wikipedia/View Controllers/ShareCard/WMFShareOptionsController.m
```diff
@@ -0,0 +1,349 @@
...
+    self.originButtonItem = nil;
+    self.originView = view;
+    [self fetchImageThenShowShareCard];
+}
+
+#pragma mark - Asynchornous Fetch and Present
+
+- (void)fetchImageThenShowShareCard {
+    @weakify(self);
+    [[WMFImageController sharedInstance] fetchImageWithURL:[NSURL wmf_optionalURLWithString:self.article.imageURL]].then(^(WMFImageDownload* download){
```
bgerstle: two comments about this:

1. Using `fetchImageWithURL:` unfortunately prevents this from showing immediately in the case that the image is already cached in memory
2. One of my gripes w/ how sharing is currently implemented is that we don't do anything until we have _all_ the data, which is exacerbated by 1.

Both of these result in a noticeable lag when the user taps the share button.  So, I'm going to file another ticket to fix this lag by displaying the share options immediately and lazily load the image using the `UIImageView+MWKImage` category.

Unrelated, but I'm also noticing some UX things that I'd like to bring up w/ design, namely:
- Bouncing the share button seems out of place. Nothing else in our UI does this, and I think it will seem out of place or distracting if we're presenting the share options immediately as I outlined above
- Adding a (red) cancel button to the options. All other OS sheets have one, and I expect that user testing would reveal users getting confused and just tapping the screen out of desperation to get out, instead of having a clear exit path.
Adding the cancel button is lower priority, but I think the removal of bouncing is has low enough cost:benefit ratio to do now if we agree it should be done.

bgerstle:
:+1: like the UX changes a lot! found an issue that also exists on 4.1.7 in iOS 9 which I'm filing now.

bgerstle:
:-1: actually, tried switching to a different site and the snippet is always empty: both when there is and isn't text selected.

also discovered our search site stuff isn't working. had to restart the app for the change to take affect.

coreyfloyd:
@bgerstle yeah agreed on the UX stuff… i thought of making a change and adding a button - but figured best to put that in another ticket - I think there needs to be more changes than just that to be honest.

As far as the not showing until fetching - totally agree there as well - I been thinking that really needs to go… we can show the image that we do have and then download a better one if we need it - AFTER we show the preview. Something else we can file (though it may also be informed by an UX changes)

As far as the bug - I'm looking into that - may be pre-existing if it has to do with the search site. So we may need to file a ticket to fix the search site stuff.

bgerstle:
@coreyfloyd I already filed a ticket, so if you find the issue we can consider this issue non-blocking

coreyfloyd:
@bgerstle found and fixed the issue. Mentioned this in Friday - but just to refresh - it wasn't the search language switch that was the problem - it was cached article vs un-cached article. 

The problem was that the article was being cached by the articleVC - and the articleContainerVC was never notified when the article was cached. Since the share button is now in the articleContainerVC, that meant it didn't have the cached article for sharing purposes. 

This pull was updated to move article fetching into the container.

bgerstle:
:-1: tapping share twice presents another share dialog.  shouldn't be able to interact w/ the toolbar while in share (or any other modal) dialog. we should probably hide the toolbar anyway, but that means we need to have a "didDismiss" callback to make it visible again.

bgerstle:
I can also go back while the share dialog is presented.  Maybe we should make this a true modal and avoid hacky workarounds?  I was looking at `UIPresentationController` for the new search UX, and it has a `shouldRemovePresentersView` property.  If I can get that working w/ search, maybe we can use it for share as well.  What do you think we should do in the meantime?

bgerstle:
confirm the issue i saw w/ empty cards is fixed though :white_check_mark:

bgerstle:
also, i found out why search site stuff is broken. will fix quickly on master (TL;DR; DYAC)

coreyfloyd:
@bgerstle fixed the button issue and moved dispatching to the main thread.

bgerstle:
noooooooo my fix for the flakey test didn't work :disappointed: wtf

coreyfloyd:
:stuck_out_tongue_winking_eye:

coreyfloyd:
@bgerstle re-running after merging master… :fearful:

PR #129, opened 2015-09-23
Comment flaky test

bgerstle:
Comment out the test that caused master to fail.  Will come back later and fix it, since it's mostly likely a race condition.  Also uncrustified some stuff.


coreyfloyd:
@bgerstle will re-add test with a better timeout

PR #133, opened 2015-09-28
use a protocol to ensure image requests never start during tests

bgerstle:
also fixes interesting crash on deinit due to attempting to make a weak reference


bgerstle:
@coreyfloyd @montehurd running the tests on this a few more times. only reran once to verify previous (false) fix, so will go for several this time around

PR #134, opened 2015-09-28
Push full article view (T113773)

bgerstle:
Phab: [T113773](https://phabricator.wikimedia.org/T113773) [T113737](https://phabricator.wikimedia.org/T113737)

- [x] Fix image display in header
- [x] Add edit pencil back to webview

---

Removes `WMFArticleViewController` from the VC hierarchy, showing `WebViewController` to the user instead on push.  As a result, I also:

- Refactored some of the toolbar logic to make the transitions more seamless
- Fixed `WebViewController` layout so it extends under the navigation bar w/ correct insets (yay translucence!)
- Added a refresh button to the toolbar (for debugging)
- By removing the embedded navigation controller in `WMFArticleContainerViewController`, I was also able to "fix" edge swiping (yay gestures!)
- Moved read more & article header gallery into `WMFArticleContainerViewController`—but, add them as headers and footers of the `WebViewController`.  This keeps the stuff around the article presentation separated from the article content itself.
- Re-added legacy functionality of collapsing header gallery when vertically compact (i.e. landscape on phone)
- Disable a bunch of `WebViewController` stuff that was causing scroll view offset jumps when it first appears
- Go back to legacy HTML sections (w/ edit pencil)—with the exception of showing the title in HTML instead of on top of the image

Check out the GIFs!

- [Edge swiping](https://cloud.githubusercontent.com/assets/444217/10177180/bc8009aa-66c6-11e5-8f3a-03c90b29ad82.gif)
- [Content size changes](https://cloud.githubusercontent.com/assets/444217/10177189/c9c76ff4-66c6-11e5-83ac-6227f89aa86c.gif)


bgerstle:
Edge swiping!

edge-swiping

bgerstle:
Content size changes!

content-size-handling

Wikipedia/UI-V5/WMFAppViewController.m
```diff
@@ -362,4 +360,14 @@ - (void)searchLanguageDidChangeWithNotification:(NSNotification*)note {
     [self configureSearchViewController];
 }
 
+#pragma mark - UINavigationControllerDelegate
+
+- (void)navigationController:(UINavigationController*)navigationController
```
coreyfloyd: This logic is a little hidden and not very obvious for the rest of the app (Its not clear who is hiding and unhiding the tool bar). Would be better that this is in a UINavigaitonController subclass where you can override push/pop.

Wikipedia/UI-V5/WMFAppViewController.m
```diff
@@ -362,4 +360,14 @@ - (void)searchLanguageDidChangeWithNotification:(NSNotification*)note {
...
+      willShowViewController:(UIViewController*)viewController
+                    animated:(BOOL)animated {
+    BOOL isToolbarEmpty = [viewController toolbarItems].count == 0;
+    [self.rootTabBarController wmf_setTabBarVisible:isToolbarEmpty animated:NO completion:nil];
```
coreyfloyd: Also the tab bar is now being hidden manually instead of using the API that was created for it - hidesBottomBarWhenPushed

Wikipedia/UI-V5/WMFAppViewController.m
bgerstle: I was thinking about doing a custom nav controller and adding our own `hidesBottomToolbarWhenPushed` property. Should I do that now?

Wikipedia/UI-V5/WMFArticleContainerViewController.m
```diff
@@ -8,17 +8,23 @@
...
 
 // Controller
-#import "WMFArticleFetcher.h"
 #import "WMFArticleViewController.h"
 #import "WebViewController.h"
 #import "UIViewController+WMFStoryboardUtilities.h"
 #import "WMFSaveButtonController.h"
 #import "WMFPreviewController.h"
 #import "WMFArticleContainerViewController_Transitioning.h"
+#import "WMFArticleHeaderImageGalleryViewController.h"
```
coreyfloyd: Its starting to feel a little crowded in here. Can we do some organization? if not some PONSOs, then maybe some categories. Just a lot of imports now.

Wikipedia/UI-V5/WMFAppViewController.m
bgerstle: @coreyfloyd on second thought i can just use the logic that's there now: show the toolbar if the view controller has items, otherwise hide it

Wikipedia/UI-V5/WMFArticleContainerViewController.m
```diff
@@ -262,29 +337,34 @@ - (void)viewWillTransitionToSize:(CGSize)size withTransitionCoordinator:(id<UIVi
...
 
 - (void)updateInsetsForArticleViewController {
-    CGFloat topInset = [self.navigationController.navigationBar frame].size.height
-                       + [[UIApplication sharedApplication] statusBarFrame].size.height;
-
-    UIEdgeInsets adjustedInsets = UIEdgeInsetsMake(topInset,
-                                                   0.0,
-                                                   self.tabBarController.tabBar.frame.size.height,
-                                                   0.0);
+    // TODO: remove
```
coreyfloyd: Its commented out, whats stopping us now?

Wikipedia/UI-V5/WMFArticleContainerViewController.m
bgerstle: Yeah, I thought of that too. There's a big legacy code cleanup coming which might help with this (removing a lot of leftover `WMFArticleViewController` stuff), but it would be nice to break out responsibilities a bit more.

Wikipedia/UI-V5/WMFAppViewController.m
bgerstle: I can try to add that back, but IIRC there were issues w/ it. Unfortunately I forgot to document them :disappointed:

Wikipedia/UI-V5/WMFArticleContainerViewController.m
```diff
@@ -262,29 +337,34 @@ - (void)viewWillTransitionToSize:(CGSize)size withTransitionCoordinator:(id<UIVi
...
 
-    self.articleViewController.tableView.contentInset          = adjustedInsets;
-    self.articleViewController.tableView.scrollIndicatorInsets = adjustedInsets;
+#pragma mark - Toolbar Actions
 
-    //adjust offset if we are at the top
-    if (self.articleViewController.tableView.contentOffset.y <= 0) {
-        self.articleViewController.tableView.contentOffset = CGPointMake(0, -topInset);
-    }
+- (void)didTapRefresh {
```
coreyfloyd: Unneeded method. We are using a block callback for the button, so no "action method" is needed - you should just call `[self fetchArticle]` directly from the block callback.

Wikipedia/UI-V5/WMFArticleContainerViewController.m
bgerstle: Was planning to do all the articleVC deletions in one pass to keep the diff manageable, but can remove this part now if you prefer.

Wikipedia/UI-V5/WMFArticleContainerViewController.m
```diff
@@ -407,24 +485,24 @@ - (void)webViewController:(WebViewController*)controller didTapOnLinkForTitle:(M
...
 
 - (void)navigationController:(UINavigationController*)navigationController willShowViewController:(UIViewController*)viewController animated:(BOOL)animated {
-    if (viewController == self.articleViewController) {
-        [self.navigationController setNavigationBarHidden:NO animated:NO];
-        [self.contentNavigationController setNavigationBarHidden:YES animated:NO];
-    } else {
-        [self.navigationController setNavigationBarHidden:YES animated:NO];
-        [self.contentNavigationController setNavigationBarHidden:NO animated:NO];
-    }
+//    if (viewController == self.articleViewController) {
```
coreyfloyd: Delete?

Wikipedia/UI-V5/WMFArticleContainerViewController.m
bgerstle: That said, the class still weighs in at a not-totally-unreasonable ~550 LOC (w/ plenty of multi-line assignments and things).  Compared to the OG `WebViewController` at over 1900 LOC :scream:

Wikipedia/UI-V5/WMFArticleListCollectionViewController.m
```diff
@@ -24,16 +25,28 @@
...
-    return [self initWithCollectionViewLayout:[SelfSizingWaterfallCollectionViewLayout new]];
++ (SelfSizingWaterfallCollectionViewLayout*)createLayout {
+    return [SelfSizingWaterfallCollectionViewLayout new];
+}
+
++ (Class)collectionViewClass {
+    return [UICollectionView class];
+}
+
++ (UICollectionView*)createCollectionView {
```
coreyfloyd: Why do this instead of using the purpose built initializer?

```
- (instancetype)init {
    return [self initWithCollectionViewLayout:[SelfSizingWaterfallCollectionViewLayout new]];
}
```

bgerstle: It's not a `UICollectionViewController` anymore

Wikipedia/UI-V5/WMFArticleListCollectionViewController.m
```diff
@@ -130,6 +143,21 @@ - (void)scrollToArticleIfOffscreen:(MWKArticle*)article animated:(BOOL)animated
 
 #pragma mark - UIViewController
 
+- (void)loadView {
```
coreyfloyd: Should be needed if you just use the initializer for passing in the collection view layout desired. 

You should only need this to use a custom UICollectionView subclass. `loadView` isn't required if all you need to do is configure the collection view.

coreyfloyd: Oh wait - are you using a custom collection view class??

coreyfloyd: Yep - you are - disregard… :stuck_out_tongue_closed_eyes:

bgerstle: see my other comment about this not being a `UICollectionViewController`, which AFAIK doesn't provide a way to create the `UICollectionView`—only allows custom layouts

Wikipedia/UI-V5/WMFAppViewController.m
bgerstle: I actually forgot, `hidesBottomBarWhenPushed` is being set on the article container... let me fiddle with this a bit and see if there's a clear reason why it works the way it does when setup this way and if there's a less-intrusive way to do it.

Wikipedia/View Controllers/WebView/WebViewController.m
```diff
@@ -1382,6 +1534,8 @@ - (void)didRotateFromInterfaceOrientation:(UIInterfaceOrientation)fromInterfaceO
 }
 
 - (void)scrollToElementOnScreenBeforeRotate {
+    // FIXME: rotating portrait/landscape repeatedly causes the webview to scroll down instead of maintaining the same position
+    return;
```
coreyfloyd: Do we just need to adjust finalScrollOffset by the header height or is it more complicated?

bgerstle: I haven't tried anything, couldn't tell you

bgerstle: Should I look into it now?

PR #137, opened 2015-09-30
Open external urls in modal window if pre ios 9.

montehurd:
https://phabricator.wikimedia.org/T104824

Used delegation. Reponder chain approach didn't quite do what we
needed.

Did not change the crash reporting url to open modally.


montehurd:
Needs rebasing, but I i ran into merge conflict here ( https://github.com/wikimedia/wikipedia-ios/commit/c40b18132a976ec3f0bf64b7777c5bb72598ee28#diff-6797d570310857eec6a38c955002bb24R251 ). That's the line that shows a warning if a zero-rated user is about to open a non-zero-rated link (we have do this or they might get charged by their carrier)... @bgerstle were you doing further refactoring on that singleton? I didn't just want to re-add a reference to it without checking first

montehurd:
On reflection, we'll need to show the "about to display non zero-rated content" alert for *all* of these external links...

montehurd:
@coreyfloyd @bgerstle do the zero-rated-warning bits as a follow on?

bgerstle:
@montehurd RE: session singleton stuff, the article presenter isn't used anymore, so you shouldn't need `SessionSingleton` in that file.  regarding zero warnings, you should be able to hook up the alert really easily inline using `UIAlertView+BlocksKit`, but i'm not 100% sure about what needs to happen. we can chat tomorrow when you get in if using `UIAlertView+BlocksKit` or something along those lines to open the URL after the user taps a button in an alert isn't what needs to happen

bgerstle:
@montehurd took a closer look RE: zero stuff. basically, you can invoke the `showWarningIfNeededBeforeOpeningURL:` from within your delegate callbacks and modify the `showWarningIfNeededBeforeOpeningURL:` implementation to call your conditional modal logic.

bgerstle:
@montehurd labeled as WIP until you're satisfied w/ Wikipedia Zero behavior

Wikipedia/Categories/UIViewController+WMFOpenExternalUrl.m
```diff
@@ -0,0 +1,51 @@
...
+#import "ZeroConfigState.h"
+#import "SessionSingleton.h"
+#import "UIAlertView+BlocksKit.h"
+
+@implementation UIViewController (WMFOpenExternalLinkDelegate)
+
+- (void)wmf_openExternalUrl:(NSURL*)url {
+    NSParameterAssert(url);
+
+    void (^ openExternalURL)(NSURL*) = ^void (NSURL* url) {
```
bgerstle: NBD, but breaking this blocks out into separate category methods might make the enclosing method easier to grok. IOW, the trade-off of capturing 1 argument (`url`) isn't worth the readability sacrifice of having all this inline blocks for things that could easily be instance methods w/ a URL argument.

montehurd: Done!

bgerstle:
:+1:

Wikipedia/View Controllers/Preview/PreviewLicenseView.m
```diff
@@ -132,13 +133,13 @@ - (UIActionSheet*)sheet {
...
-- (void)actionSheet:(UIActionSheet*)actionSheet clickedButtonAtIndex:(NSInteger)buttonIndex {
+- (void)actionSheet:(UIActionSheet*)actionSheet didDismissWithButtonIndex:(NSInteger)buttonIndex {
     switch (buttonIndex) {
         case BUTTON_TERMS:
-            [[UIApplication sharedApplication] openURL:[NSURL URLWithString:TERMS_LINK]];
+            [self.externalLinksOpenerDelegate wmf_openExternalUrl:[NSURL URLWithString:TERMS_LINK]];
             break;
         case BUTTON_LICENSE:
-            [[UIApplication sharedApplication] openURL:[NSURL URLWithString:LICENSE_LINK]];
+            [self.externalLinksOpenerDelegate wmf_openExternalUrl:[NSURL URLWithString:LICENSE_LINK]];
```
coreyfloyd: @montehurd  Do you still need the delegate? did the responder chain stuff not work out?

montehurd: No, it didn't :(

PR #138, opened 2015-09-30
New search UX

bgerstle:
Phab: [T113071](https://phabricator.wikimedia.org/T113071)

- [x] implement modal transition w/ blur
- [x] restyle search field
  - [x] "text field" appearance
  - [x] "X" button to close search text field (**note** the close button looks different than mocks, but will be fixed when get new icons)
- [x] add search button to... 
  - [x] home view
  - [x] article view
  - [x] saved
  - [x] recent
- [x] send "Search" log to pywik
- [x] add custom "X", W, Settings, & Search icons
- [x] style search results
- [x] make search field mimic navigation bar size class behavior

### Amendments
- [x] Remove empty `-[WMFSearchViewController initWithCoder:]`
- [x] Remove `wmf_presentSearchViewController` (YAGNI)
- [x] Remove `willAnimate` assertion in `WMFSearchViewController`
- [x] remove `presentArticle` macro

---

### Demos
- [Search UX demo](https://vid.me/kPpD)
- [Search bar UX](https://vid.me/DaMt)

### Caveats
Search result & recent search styling will be done in another PR. Please focus on architectural review of how search view controller is created, and UI/UX of search presentation and toolbars.

#### `WMFSearchPresentationDelegate`
Provides dependencies for creating & presenting a search controller and responding to its results.  The "informal protocol" leaves search result handling to the individual view controllers.  Not as DRY as I would like, but I think our object graph needs a little pruning & refactoring in general (see observations).

#### `UIViewController+WMFSearchButton`
Category which allows view controllers to create a search button (with a `WMFSearchPresentationDelegate`, often `self`), which handles creating & presenting a `WMFSearchViewController`.  As mentioned above, delegates still need to implement reacting to search result selection.  Currently this is repeated in Home, ArticleList, and ArticleContainer classes which all:

- Create an article container, using dependency injection
- Push it on the navigation stack
- Add it to history w/ the correct discovery method
- Save history list

This is kept DRY by way of...

#### `UIViewController+WMFArticlePresentation`
Category (and macros for convenience) to encapsulate business logic of presenting an article container.  The macros are my way of copying with not being able to use protocol extensions to reuse this logic, but works relatively well.  Not 100% sure this should be a macro vs. a C function, but there's not a big difference at this point, except that—if we ever get around to it—a Swift protocol extension could call the C function as a way to share the functionality across languages.  But I digress...

### Changed

`MWKUserDataStore` is now a persistent property of `MWKDataStore`. The idea is to facilitate dependency injection and prevent unexpected behavior due to re-instantiation of the `userDataStore` and its associated lists.

`WMFArticleListCollectionViewController` now has a selection delegate which, if set, silences its default behavior of pushing an article container.  This allows search to use a list controller which, through delegation, allows the search view controller's delegate to present the selected search result.


Wikipedia/UI-V5/UIViewController+WMFSearchButton.m
```diff
@@ -0,0 +1,51 @@
...
+    @weakify(delegate);
+    return [UIBarButtonItem wmf_buttonType:WMFButtonTypeMagnify handler:^(id sender) {
+        @strongify(self);
+        @strongify(delegate);
+        if (!delegate || !self) {
+            return;
+        }
+
+        MWKSite* searchSite;
+        if ([delegate respondsToSelector:@selector(searchSite)]) {
```
coreyfloyd: Why are we doing this?

Wikipedia/UI-V5/UIViewController+WMFSearchButton.m
```diff
@@ -0,0 +1,51 @@
...
+
+        WMFSearchViewController* searchVC =
+            [WMFSearchViewController searchViewControllerWithSite:searchSite
+                                                        dataStore:[delegate dataStore]];
+        searchVC.searchResultDelegate = delegate;
+        [self wmf_presentSearchViewController:searchVC];
+    }];
+}
+
+- (void)wmf_presentSearchViewController:(UIViewController*)searchViewController {
```
coreyfloyd: Just replacing one line - just present the view controller - 3 less lines of code and no confusing "special" way to present search - its presented the same way as any other VC

Wikipedia/Legacy Data Migration/WMFLegacyImageDataMigration.swift
```diff
@@ -23,20 +23,8 @@ public class WMFLegacyImageDataMigration : NSObject {
     /// Image controller where data will be migrated.
     let imageController: WMFImageController
 
-    /// List of saved pages which is saved when the tasks are finished processing.
```
coreyfloyd: Looks like a lot of refactoring in this class beyond the scope of the ticket. I didn't really see anything about the reasoning for the changes in the PR comments or the commit logs. 

I like removing code, but a little worried about these changes since they seem a little invasive. Do you have some rationale for these or is this something we can factor out to be reviewed separately?

Wikipedia/UI-V5/WMFArticleListCollectionViewController.m
```diff
@@ -198,43 +219,54 @@ - (void)viewWillTransitionToSize:(CGSize)size withTransitionCoordinator:(id <UIV
...
 
-- (void)collectionView:(UICollectionView*)collectionView didSelectItemAtIndexPath:(NSIndexPath*)indexPath {
-    self.selectedArticle = [self.dataSource articleForIndexPath:indexPath];
+- (void)showArticle:(MWKArticle*)article discoveryMethod:(MWKHistoryDiscoveryMethod)discoveryMethod {
+    if (self.delegate) {
+        [self.delegate didSelectArticle:article sender:self];
+        return;
+    }
 
     WMFArticleContainerViewController* container =
```
coreyfloyd: Is the below logic everywhere it needs to be (Mostly referring to the page logging)? 
Should we move article presentation into a category as well so it can be shared among all VCs? That way we don't have divergent presentation logic?

Wikipedia/Legacy Data Migration/WMFLegacyImageDataMigration.swift
bgerstle: this was mostly a result of making `userDataStore` a persistent property on `MWKDataStore`, which obviated the need for a separate saved page list in this class and a lot of the logic related to "transferring" the `didMigrateImageData` flags over to the "main" saved page list.

Wikipedia/UI-V5/UIViewController+WMFSearchButton.m
bgerstle: not all delegates have a search site, mainly `WMFArticleListCollectionViewController`.  we could either delegate to its underlying data source, or use the app-wide search site setting.

Wikipedia/UI-V5/WMFSearchViewController.m
```diff
@@ -41,19 +56,26 @@ - (void)setRecentSearchesHidden:(BOOL)hidingRecentSearches animated:(BOOL)animat
...
+    searchVC.modalPresentationStyle = UIModalPresentationOverFullScreen;
+    searchVC.modalTransitionStyle   = UIModalTransitionStyleCrossDissolve;
+    return searchVC;
 }
 
-- (void)setSavedPages:(MWKSavedPageList* __nonnull)savedPages {
-    _savedPages = savedPages;
-    // resultsListController might be nil if `prepareForSegue:sender:` hasn't been called
-    self.resultsListController.savedPages = savedPages;
+- (instancetype)initWithCoder:(NSCoder*)aDecoder {
```
coreyfloyd: empty implementation

Wikipedia/UI-V5/UIViewController+WMFSearchButton.m
bgerstle: True, we don't need to customize how it's presented now. I'll remove it

Wikipedia/UI-V5/WMFArticleListCollectionViewController.m
bgerstle: That's the part of this I don't like. I'd really like to DRY that up, but I'm not sure of the best way to do it.  We can brainstorm of hangout if you want.  The main issue is the who/where article container view controllers are configured & presented. So far the options I've come up with are:

- Use the `WMFSearchPresentationDelegate` accessors (which is a superset of those needed to create an article container) to create the article container, then present it
- Call `showViewController:sender:` and let the presenter handle configuration & presentation
- implement lazy getters for `WMFArticleContainerViewController` dependencies which grab the "main" instances of what it needs (data store, saved page list, etc.)

bgerstle: another implicit question: do we continue the trend of `WMFArticleContainerViewController` becoming the new home of `WebViewController` functionality, such as history list updates, logging, etc.?

Wikipedia/UI-V5/WMFSearchViewController.m
bgerstle: oops, forgot why i originally had that there, but must've forgot to remove it when it became unnecessary

Wikipedia/Legacy Data Migration/WMFLegacyImageDataMigration.swift
coreyfloyd: What about the queue what is the reasoning/effects of removing the background queue? Also there is new logic for re-dipatching to the main queue as well.

Wikipedia/UI-V5/UIViewController+WMFSearchButton.m
coreyfloyd: Sorry, my question is why do we have a search delegate at all - why are we not deferring to the app wide setting?

Wikipedia/Legacy Data Migration/WMFLegacyImageDataMigration.swift
bgerstle: right, since it's now using the main saved page list, i have to access it on the main queue, meaning there's no need for the internal queue or `thenOn` etc.

Wikipedia/UI-V5/WMFSearchViewController.m
```diff
@@ -99,59 +116,85 @@ - (void)setRecentSearchesHidden:(BOOL)hidingRecentSearches animated:(BOOL)animat
     _recentSearchesHidden = hidingRecentSearches;
 
     [UIView animateWithDuration:animated ? [CATransaction animationDuration] : 0.0
+                          delay:0 options:UIViewAnimationOptionBeginFromCurrentState
```
coreyfloyd: What did you need this for? Flakey animation?

Wikipedia/UI-V5/WMFSearchViewController.m
```diff
@@ -99,59 +116,85 @@ - (void)setRecentSearchesHidden:(BOOL)hidingRecentSearches animated:(BOOL)animat
...
+
+    BOOL willAnimate = [self.transitionCoordinator
+                        animateAlongsideTransition:^(id < UIViewControllerTransitionCoordinatorContext > _Nonnull context) {
+        [self.view layoutIfNeeded];
+        [self updateRecentSearchesVisibility:animated];
+        [self.searchField becomeFirstResponder];
+        [self setSeparatorViewHidden:NO animated:YES];
+    }
+                                        completion:nil];
+    NSParameterAssert(willAnimate);
```
coreyfloyd: What is this check here for? Does the transition animator sometimes not animate?

bgerstle: it will return `false` if the block is not added to the transition animation. for example, if someone moves this outside of `viewWillAppear:` or something else unexpected changes.

Wikipedia/UI-V5/WMFSearchViewController.m
bgerstle: Might not be necessary, will have to check

Wikipedia/UI-V5/WMFSearchViewController.m
coreyfloyd: Seems unnecessary then, unless we want to add that check everywhere we call `animateAlongsideTransition:` in the code base. But that seems excessive to me.

bgerstle: i mostly added it because this is the first time i've done this inside a view controller's `viewWillAppear:`. but, if this is the place to do this kind of things and it works, I think the assertion is unnecessary on the grounds that you wouldn't assert the result of this method when called inside something like `viewWillTransitionToSize:`

coreyfloyd:
@bgerstle left a comment on your latest. This is another reoccurring conversation I would like to resolve and avoid further reviews on. 

I know you favor macros frequently in your coding… but Macros:
- Are lower level abstraction that OOP
- Can be unsafe
- Can't be accessed in Swift

Our best practices specifically address the first 2, and we should probably update it for the last one as well since that is becoming more important. 

To me this points to Macros being used less frequently than they are now… especially in situations like this one where an OO approach is obvious and the Macro benefit is pretty low. 

We don't have a hard line on this subject, but to me simply reducing the number of parameters shouldn't enough of a reason to pull out a Macro and go against best practices.

As with the other issue, if you want to make changes to best practices, that is something we can discuss, but I'd like to close the book on this issue as well so we can reduce our code review discussion and time.

coreyfloyd:
:+1:

coreyfloyd:
@montehurd merge if you are good

PR #142, opened 2015-10-02
ToC update

coreyfloyd:
Added ToC

- New ToC VC - a table view.
- New cell for the table view
- Used presentation controller / transitioning API for presentation
- Wrote mostly in Swift
- Implemented some section list introspection
- Removed old ToC

Some style notes:
- The mock has no status bar… but we have a status bar. Animating it offscreen causes other issues for the main article view (tried that). So we need to discus with design if we really want to remove the status bar.
- The "scroll" web view to a section logic seems not great. It instantly scrolls - we should probably scroll with animation. Also `currentVisibleSection` seems to return the section above sometimes (seems like it has to do with the navigation bar) Need to get with @montehurd and maybe create a ticket to address.


MediaWikiKit/MediaWikiKit/MWKSection.h
```diff
@@ -76,47 +78,62 @@ extern NSString* const MWKSectionShareSnippetXPath;
...
+ */
+- (MWKSection*)rootSection;
+
+/**
  *  Sections that are descendants of the receiver.
  *
  *  @return An array of @c MWKSection objects, or @c nil if the hierarchy has not been built yet.
  */
-- (NSArray*)children;
+- (nullable NSArray*)children;
```
bgerstle: can haz type annotation? `NSArray<MWKSection*>*`

MediaWikiKit/MediaWikiKit/MWKSection.m
```diff
@@ -19,20 +21,21 @@ @interface MWKSection ()
...
 @property (readwrite, weak, nonatomic) MWKArticle* article;
 
-@property (readwrite, copy, nonatomic) NSNumber* toclevel;      // optional
-@property (readwrite, copy, nonatomic) NSNumber* level;         // optional; string in JSON, but seems to be number-safe?
-@property (readwrite, copy, nonatomic) NSString* line;          // optional; HTML
-@property (readwrite, copy, nonatomic) NSString* number;        // optional; can be "1.2.3"
-@property (readwrite, copy, nonatomic) NSString* index;         // optional; can be "T-3" for transcluded sections
-@property (readwrite, strong, nonatomic) MWKTitle* fromtitle; // optional
-@property (readwrite, copy, nonatomic) NSString* anchor;        // optional
+@property (readwrite, copy, nonatomic, nullable) NSNumber* toclevel;      // optional
```
bgerstle: can you please remove the `// optional` comments? maybe even move the other comments in to doc comments above the property?

bgerstle:
@coreyfloyd can you explain the changes to the code signing & prov profile settings?

Wikipedia/Categories/NSString+Extras.m
```diff
@@ -144,4 +142,17 @@ - (NSString*)wmf_trim {
     return [self stringByTrimmingCharactersInSet:[NSCharacterSet whitespaceCharacterSet]];
 }
 
+- (NSString*)wmf_substringUpToString:(NSString*)string{
+    return [[self componentsSeparatedByString:string] firstObject];
+}
+- (NSString*)wmf_substringAfterString:(NSString*)string{
+   NSArray* components = [self componentsSeparatedByString:string];
```
bgerstle: I don't think this would work for cases like `[@"foo1bar1baz" wmf_substringAfterString:@"1"]`. I would expect it to return `@"1bar1baz"` but seems like this would return `@"baz"`, is that intentional?  If not, I would consider using `rangeOfString:` and `substring(To|From)Index:`.

Wikipedia/Categories/UIView+WMFDefaultNib.m
```diff
@@ -1,17 +1,15 @@
...
-//
 
 #import "UIView+WMFDefaultNib.h"
+#import "NSString+Extras.h"
 
 @implementation UIView (WMFDefaultNib)
 
 + (NSString*)wmf_nibName {
-    return NSStringFromClass(self);
+    /* Swift has "Namespaced" class names that prepend the module
```
bgerstle: there's a ticket for Swift naming conventions, in which I propose we have `@objc` name-spaced names for classes that are used in objc. For example, `WMFImageController becomes `@objc(WMFImageController) ImageController`.  Doing this with cell classes would allow them to be referenced via prefix in ObjC and obviate these kinds of workarounds.  My only reservations with this are:

- Performance (very, very minor, but this is unnecessary in the vast majority of cases in the app)
- [Name mangling](http://stackoverflow.com/questions/24321773/how-can-i-demangle-a-swift-class-name-dynamically) in Swift could make this even more problematic
- Nested types (YAGNI, but possible down the line)

For example:
``` swift
@objc(WMFMyCell) public class MyCell : UITableViewCell {
}
```

results in:

```
(lldb) po NSStringFromClass([WMFMyCell class])
WMFMyCell
(lldb) po [WMFMyCell wmf_nibName]
WMFMyCell
```

Wikipedia/UI-V5/WMFArticleContainerViewController.m
```diff
@@ -217,6 +221,14 @@ - (WMFArticleHeaderImageGalleryViewController*)headerGallery {
     return _headerGallery;
 }
 
+
+- (WMFTableOfContentsViewController*)tableOfContentsViewController{
+    if(!_tableOfContentsViewController){
+        _tableOfContentsViewController = [[WMFTableOfContentsViewController alloc] initWithSectionList:self.article.sections delegate:self];
```
bgerstle: I might be missing it, but how do yo update the ToC if another article is set on the container (e.g. as the result of a fetch or another save)?

Wikipedia/UI-V5/WMFTableOfContentsAnimator.swift
```diff
@@ -0,0 +1,59 @@
+
+import UIKit
+
+class WMFTableOfContentsAnimator: NSObject, UIViewControllerAnimatedTransitioning {
+    
+    // MARK: - init
+    let isPresenting :Bool
```
bgerstle: any reason why we can't have two animator classes? something else apparently knows when to chose one over the other, and I can't see ay benefit to having both functions in the same class which could also be had w/ a shared base class

Wikipedia/UI-V5/WMFTableOfContentsCell.swift
```diff
@@ -0,0 +1,167 @@
...
+        self.selectedSectionIndicator.alpha = 0.0
+        self.sectionSelectionBackground.backgroundColor = UIColor.wmf_tableOfContentsSelectionBackgroundColor()
+        self.sectionSelectionBackground.alpha = 0.0
+        self.selectionStyle = .None
+    }
+    
+    // MARK: - Accessors
+    public var section: MWKSection?{
+        didSet(section) {
+            self.sectionTitle.text = WMFTableOfContentsCell.titleText(self.section)
```
bgerstle: if we're always passing `self.section`, these should just be instance methods/functions

Wikipedia/UI-V5/WMFTableOfContentsCell.swift
```diff
@@ -0,0 +1,167 @@
...
+            self.sectionTitle.textColor = WMFTableOfContentsCell.textColor(self.section)
+            self.leadSectionBorder.hidden = !WMFTableOfContentsCell.leadSectionBorderEnabled(self.section)
+            self.sectionBorder.hidden = !WMFTableOfContentsCell.topBorderEnabled(self.section)
+            self.indentationConstraint.constant = WMFTableOfContentsCell.indentationWidth(self.section)
+            self.setNeedsUpdateConstraints()
+        }
+    }
+    
+    public func setSectionSelected(selected: Bool, animated: Bool) {
+        UIView.animateWithDuration(animated ? 0.3 : 0.0) { () -> Void in
```
bgerstle: is the `() -> Void in` necessary here? in general we should try to let the compiler infer this for us, but there are cases that it needs help—just didn't think this would be one.

Wikipedia/UI-V5/WMFTableOfContentsCell.swift
```diff
@@ -0,0 +1,167 @@
...
+            if(selected){
+                self.sectionSelectionBackground.alpha = 1.0
```
bgerstle: ternary assignment would've also worked here

Wikipedia/UI-V5/WMFTableOfContentsCell.swift
```diff
@@ -0,0 +1,167 @@
...
+            }else{
+                self.sectionSelectionBackground.alpha = 0.0
+            }
+        }
+    }
+    
+    // MARK: - UITableVIewCell
+    public class func reuseIdentifier() -> String{
+        return self.wmf_nibName()
```
bgerstle: any reason why we aren't using `wmf_nibName` directly?

Wikipedia/UI-V5/WMFTableOfContentsCell.swift
```diff
@@ -0,0 +1,167 @@
...
+        self.sectionBorder.hidden = true
+        self.setSelected(false, animated: false)
+        self.sectionSelectionBackground.alpha = 0.0
+    }
+    
+    public override func setSelected(selected: Bool, animated: Bool) {
+        super.setSelected(selected, animated: animated)
+        if(selected){
+            //HACK: I don't know why I have to set the color here, but I do. Something is setting it to clear background color
+            self.selectedSectionIndicator.backgroundColor = UIColor.wmf_tableOfContentsSelectionIndicatorColor()
```
bgerstle: try changing the selection style to `none` or custom?

Wikipedia/Categories/NSString+Extras.m
coreyfloyd: `wmf_substringAFTERString` not `wmf_substringFROMString`

Wikipedia/Categories/UIView+WMFDefaultNib.m
coreyfloyd: Another issue is that it won't work with any Swift classes we don't define… so we can only use the category with classes we define in Swift or include in Objective-C - starts to get harry fast.

Wikipedia/UI-V5/WMFArticleContainerViewController.m
coreyfloyd: @bgerstle its cleaned up in the `setArticle:` method like all the other article dependent stuff.

Also been thinking about not caching this now though… and just recreating when needed.

Wikipedia/UI-V5/WMFTableOfContentsAnimator.swift
coreyfloyd: @bgerstle Old discussion… same rationale still stands: 
- It is pretty much convention for the entire internet to do these in the same class. 
- …which It makes sense… if not for code sharing (which can happen frequently), then for logic grouping (would you implement viewDidAppear and viewDidDisappear in 2 different classes?)

Just because this keeps coming up, I will dive a bit further so we can get it settled and avoid repeat discussions:
**Current Implementation**: the view controller figures out if it is presenting or not and sets a Bool on a class
**Your proposal**: Add a class. the view controller figures out if it is presenting or not, and vends a specific class.

In your proposal, the point of the decision hasn't changed at all: the responsibility of figuring out whether we are presenting or dismissing is still on the same view controller.

Additionally, no code was saved. The code isn't safer. It isn't more obvious. It doesn't have better performance. 

On the other side, it doubles the amount of animation classes, and goes against the convention the rest of the Cocoa world has settled on.

So for me suggesting this class be separated to prevent setting a single Bool on init seems like mostly bike shedding on something that doesn't actually benefit the code base.

If you want to discus further let me know… but to me it is pretty clear why we shouldn't go the way you are suggestion. Either way lets close the book on it.

Wikipedia/UI-V5/WMFTableOfContentsCell.swift
coreyfloyd: thats just the auto complete working there

Wikipedia/Categories/UIView+WMFDefaultNib.m
bgerstle: Man, harry is such a jerk.  :stuck_out_tongue: 

Yeah, if we ever use Swift reusable views in another module that doesn't define ObjC names... i guess this is fine then

Wikipedia/UI-V5/WMFArticleContainerViewController.m
bgerstle: I'm all about less state, but if it's easy enough to clean up I say leave it.

Wikipedia/UI-V5/WMFTableOfContentsCell.swift
coreyfloyd: I wanted to be specific about where the text came from - doing as an instance method gives the function access to all the state of the cell. Doing it this way scopes the state to the passed in section. I would have taken a similar approach in Obj-C. Additionally passing the parameter makes the methods testable if desired.

I also considered writing this code as an extension on MWKSection… but opted to just do it in the cell.

Wikipedia/UI-V5/WMFTableOfContentsCell.swift
coreyfloyd: Being explicit instead of implicit. This is the only reason i usually wrap a single line in another method… to translate to the language of the current domain. In this case, we want a reuseIdentifier, not a nib name.

coreyfloyd:
@bgerstle not sure how I committed providing changes… will back them out…

Wikipedia/Categories/NSString+Extras.m
```diff
@@ -144,4 +142,17 @@ - (NSString*)wmf_trim {
     return [self stringByTrimmingCharactersInSet:[NSCharacterSet whitespaceCharacterSet]];
 }
 
+- (NSString*)wmf_substringUpToString:(NSString*)string{
```
coreyfloyd: Actually I'm going to rename this to be consistent with the method below…

Wikipedia/Categories/UIView+WMFDefaultNib.m
coreyfloyd: lol

Wikipedia/UI-V5/WMFTableOfContentsCell.swift
coreyfloyd: Also I considered just making them functions outside of the implementation of the class…

Wikipedia/UI-V5/WMFTableOfContentsCell.swift
coreyfloyd: The selection style is none… 

Also this isn't a UIKit provided view, `selectedSectionIndicator` is a view that I added. So it doesn't pick up on any of the selection style stuff.

Wikipedia/UI-V5/WMFTableOfContentsAnimator.swift
bgerstle: K, consider the book closed

Wikipedia/UI-V5/WMFTableOfContentsCell.swift
bgerstle: We've been doing these kinds of transformations in the cell and sharing functionality via inheritance (see `setTitle(MWKTitle)` and `setImage(MWKImage)`. These kinds of transformations/logic are usually view-specific or inheritable, so unless it's going to be reused, abstraction isn't as much of a benefit.  I think it's fine as-is, but I think it would've been acceptably testable if the transformations were tested by setting a `MWKSection` and querying for expected values.

Wikipedia/UI-V5/WMFTableOfContentsCell.swift
bgerstle: Right, so should we provide a `UITableViewCell` category which has a default implementation of `wmf_nibName`? If it's Swift extension, I don't know what the best practices is regarding prefixes (probably omit it).

Wikipedia/UI-V5/WMFTableOfContentsCell.swift
bgerstle: D'oh missed that it was a custom view.

bgerstle:
:+1: LGTM. what say you, @montehurd? @coreyfloyd screencaps & GIFs would be greeaaaat

PR #146, opened 2015-10-06
Fix for constraint breaking at run time.

montehurd:
My restyling ticket kept doing unexpected things, so I thought I'd track down what was causing this breaking constraint to see if it had anything to do with the weirdness I was seeing. Checking that now, but figured it wouldn't hurt to submit this in the mean time...

### DIFF
--- a/Wikipedia/UI-V5/WMFHomeViewController.m
+++ b/Wikipedia/UI-V5/WMFHomeViewController.m
@@ -163,7 +163,7 @@ NS_ASSUME_NONNULL_BEGIN
 
     self.collectionView.dataSource = nil;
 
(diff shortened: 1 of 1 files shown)


bgerstle:
nice boyscouting! :+1: 

the phab ticket for this is [T113757](https://phabricator.wikimedia.org/T113757). @montehurd this should solve it most of the time, but when I last looked into this, i was still seeing it in the case where we displayed a card w/o a summary/description.

montehurd:
https://phabricator.wikimedia.org/T113757

montehurd:
@bgerstle checking...

PR #147, opened 2015-10-06
Stop reloading the home view when getting phone calls, etc…

coreyfloyd:
- Added a time for reloads - now set to 10 minutes
- Only automatically reload if it has been more than 10 minutes


bgerstle:
@coreyfloyd I'm pretty sure we should be using will enter foreground for this, not did become active. the timeout will address the issue of reloading immediately when active state changes, but we should still only do it after the user puts the app into the background.

coreyfloyd:
> @coreyfloyd I'm pretty sure we should be using will enter foreground for this, not did become active. the timeout will address the issue of reloading immediately when active state changes, but we should still only do it after the user puts the app into the background.

@bgerstle Not sure I follow the logic. Can you describe the behavior that is occurring now that is incorrect? How would your suggestion correct it?

PR #149, opened 2015-10-07
Toc scroll fix

montehurd:
https://phabricator.wikimedia.org/T114956

- Fix for TOC not always showing what you had just selected as the current section if you make selection from TOC then immediately re-open TOC.

- Made bottom of the section above the "current" section able to be scrolled a max of 50 pixels onscreen before that section is considered "current".


montehurd:
@coreyfloyd Fixed those issues!

PR #150, opened 2015-10-08
use xctest when running on travis

bgerstle:
switcing to xctool might have interfered w/ codecov


bgerstle:
@coreyfloyd 17% in one PR, not too shabby eh? :wink: but hooray for coverage being restored!

super duper

PR #152, opened 2015-10-08
rename WebViewController to WMFWebViewController

bgerstle:
Might as well!


coreyfloyd:
@bgerstle not yet… doing too much work in here

bgerstle:
@coreyfloyd that's fine actually, my stuff didn't end up touching this class anyway.

we'll meet another day... `WebViewController`...

PR #153, opened 2015-10-09
Card styling bits 2

montehurd:
**Summary of changes**
- Reduced vertical space between cells to 1px.
- Capitalize first character of wikidata descrips in recent/saved.
- Retrieve plain text extracts for cards with search data source.
- Clean summaries now extracted for when the article data store is the data source. i.e. used to get summaries for articles loaded when tapping links.
- Updated number of lines shown for title, description and summary.
- Summaries now usually show 4 good lines of text on landscape.
- Much closer matching of font sizes, padding and linespacing to mock - overlaid they're within a pixel or two.
- Collapsed padding above and below description if no description.
- No placeholder image shown if the article doesn't have a lead image.
- Updated placeholder image.

**Before screenshot**

*enwiki > goat* had quite a few issues related to html-to-nsattributed strings parsing. Note the double ellipsis and large padding above the save icon - this padding was partially from the nsattributed string generated from the html-to-nsattributed strings parser:
goat old

**After screenshots**

No more *goat* summary layout issues - note the padding, linespacing and font updates too:
screen shot 2015-10-10 at 7 29 29 pm

*Related* items fetcher now asks the search api for plain text extracts. Here such an extract is used as the article summary:
screen shot 2015-10-10 at 7 28 21 pm

*Saved Pages and Recent* items don't use the search api. We can now extract very nice summaries directly from data store section html:
screen shot 2015-10-10 at 7 41 47 pm

Example layout of articles without lead images - note the description padding has been collapsed on the article which doesn't have a description:
screen shot 2015-10-10 at 10 40 46 pm

Updated placeholder images:
screen shot 2015-10-12 at 2 28 03 pm
screen shot 2015-10-12 at 2 28 22 pm
screen shot 2015-10-12 at 3 40 31 pm
screen shot 2015-10-13 at 4 48 46 pm


MediaWikiKit/MediaWikiKit/MWKSection.h
```diff
@@ -159,6 +159,9 @@ extern NSString* const MWKSectionShareSnippetXPath;
  */
 - (BOOL)hasTextData;
 
+- (NSString *)summary;
++ (NSString*)cleanSummary:(NSString*)summary;
```
coreyfloyd: Shouldn't be in the public API

MediaWikiKit/MediaWikiKit/MWKSection.m
```diff
@@ -238,6 +239,63 @@ - (void)removeAllChildren {
...
+    
+    // Cleanups which need to happen before string is shortened.
+    output = [summary wmf_stringByRecursivelyRemovingParenthesizedContent];
+    output = [output wmf_stringByRemovingBracketedContent];
+    
+    // Now ok to shorten so remaining cleanups are faster.
+    output = [output wmf_safeSubstringToIndex:WMFNumberOfExtractCharacters];
+    
+    // Cleanups safe to do on shortened string.
+    return [[[[[[output stringByReplacingOccurrencesOfString:@"&amp;" withString:@"&"]
```
coreyfloyd: Logic on lines 291, 292, and 293 should be in NSString category method(s).

MediaWikiKit/MediaWikiKit/MWKArticle.m
```diff
@@ -474,119 +475,20 @@ - (NSArray*)citations {
...
-        NSArray* paragraphsWrappedWithParagraphTags =
-            [nonEmptyParagraphsWithSelectedChildren bk_map:^NSString*(NSString* thisParagraph) {
-            return [@[@"<p>", thisParagraph, @"</p>"] componentsJoinedByString : @""];
-        }];
-
-        NSData* xpathData = [[paragraphsWrappedWithParagraphTags componentsJoinedByString:@""] dataUsingEncoding:NSUTF8StringEncoding];
-
-        return [[NSAttributedString alloc] initWithHTMLData:xpathData site:self.site];
     }
+    _summary = nil;
```
coreyfloyd: why is this here?

Wikipedia/UI-V5/WMFArticlePreviewCell.m
```diff
@@ -24,6 +30,18 @@ - (void)prepareForReuse {
     self.summaryLabel.text = nil;
 }
 
+-(void)awakeFromNib {
+    [self rememberSettingsFromIB];
+}
+
+-(void)rememberSettingsFromIB {
+    NSRange r = NSMakeRange(0, 1);
+    self.summaryLabelAttributesFromIB = [self.summaryLabel.attributedText attributesAtIndex:0 effectiveRange:&r];
```
coreyfloyd: We should just do this in code. It's a bit tacky to configure this in IB and then grab it. If you really want to see what it looks like in IB, then we should implement the IBDesignable methods to make that happen.

Wikipedia/UI-V5/WMFArticlePreviewCell.m
```diff
@@ -24,6 +30,18 @@ - (void)prepareForReuse {
...
+    self.paddingAboveDescriptionFromIB = self.paddingConstraintAboveDescription.constant;
```
coreyfloyd: Instead of caching the content, did you try setting deactivating the constraint when you want to ignore it?

```
self.paddingConstraintAboveDescription.active = NO
```

Wikipedia/UI-V5/WMFArticlePreviewCell.m
```diff
@@ -24,6 +30,18 @@ - (void)prepareForReuse {
...
+    self.paddingBelowDescriptionFromIB = self.paddingConstraintBelowDescription.constant;
```
coreyfloyd: Same for this one

Wikipedia/UI-V5/WMFArticlePreviewCell.m
bgerstle: you could also do all of it in masonry, and write visual tests to verify the appearance of each state instead of relying on IB... just sayin' :trollface:

Wikipedia/UI-V5/WMFArticlePreviewCell.m
```diff
@@ -24,6 +30,18 @@ - (void)prepareForReuse {
...
+    self.heightOfImageFromIB = self.imageHeightConstraint.constant;
```
coreyfloyd: For this one, can you just set the vertical content hugging priority to be greater than the priority of the imageHeightConstraint? I would think it would just collapse to 0 in this case.

Wikipedia/UI-V5/WMFArticlePreviewCell.m
coreyfloyd: I'd lay off the test suggestions here. We are still running long on this ticket and just need to get it out the door. We can always do visual tests during our beta fix period.

bgerstle:
@montehurd this looks _really_ nice! aside from some minor nits, my only ask is that you create "expected output" tests for the other visual tests that you deleted. you already did it for the obama article, but we should at least have one w/ the Taj Mahal article as well.

bgerstle:
My only other question is: should we always show the `entityDescription` as first-letter-capitalized, and if so, how to manage that while preserving its raw form? i don't like putting the capitalization logic in multiple places, so I would vote for lazily computing & storing the result on `MWKArticle` while reading and writing the "raw" entity description from mobileview & to the data store respectively.

MediaWikiKit/MediaWikiKit/MWKArticle.m
montehurd: ah good catch!

MediaWikiKit/MediaWikiKit/MWKSection.h
montehurd: fixed

MediaWikiKit/MediaWikiKit/MWKSection.m
montehurd: done!

Wikipedia/UI-V5/WMFArticlePreviewCell.m
montehurd: @coreyfloyd active = no isn't what we want i think - we still want this constrain - we just don't want the padding.

bgerstle:
@montehurd not sure if you're already fixing this, but please change the placeholder in `WMFImageCollectionViewCell` so the article view also has the correct placeholder:

image

montehurd:
@bgerstle just fixed the lead image placeholder.

Wikipedia/UI-V5/WMFArticlePreviewCell.m
coreyfloyd: @montehurd What is the constraint doing without the padding? I would think nothing correct?

Wikipedia/UI-V5/WMFArticlePreviewCell.m
coreyfloyd: @montehurd Any movement on this one as well?

Wikipedia/UI-V5/WMFSaveableTitleCollectionViewCell.m
```diff
@@ -32,13 +41,13 @@ - (void)prepareForReuse {
...
     [self.imageView wmf_reset];
-    self.imageView.image = [UIImage imageNamed:[[self class] defaultImageName]];
+    [self configureImageViewWithPlaceholder];
 }
 
 #pragma mark - Accessors
 
-+ (NSString*)defaultImageName {
-    return @"lead-default";
+-(UIImage*)placeholderImage {
```
coreyfloyd: Is this different per instance? If not it probably should still be a class method.

Wikipedia/UI-V5/WMFArticlePreviewCell.m
coreyfloyd: Also this wasn't changed either - I think @bgerstle and I are in agreement here that we shouldn't be using IB to get the settings for Attributed Strings. If you want to discus we can, but as I mentioned above its a bit of a hack and I think as Brian as mentioned, it "hides" the configuration of the attributes where no one expects to look.

Wikipedia/UI-V5/WMFArticlePreviewCell.m
```diff
@@ -40,38 +59,55 @@ - (UICollectionViewLayoutAttributes*)preferredLayoutAttributesFittingAttributes:
 - (void)setDescriptionText:(NSString*)descriptionText {
     _descriptionText           = descriptionText;
     self.descriptionLabel.text = descriptionText;
+    if (!self.descriptionLabel.text || self.descriptionLabel.text.length == 0) {
```
coreyfloyd: Only the second check is necessary

Wikipedia/UI-V5/WMFArticlePreviewCell.m
montehurd: @coreyfloyd it's still constrains as configured, just without the padding.

Wikipedia/UI-V5/WMFArticlePreviewCell.m
montehurd: Fixed!

Wikipedia/UI-V5/WMFSaveableTitleCollectionViewCell.m
montehurd: Fixed.

Wikipedia/UI-V5/WMFArticlePreviewCell.m
montehurd: Fixed!

PR #155, opened 2015-10-09
bracket mistake fixed

negati-ve:


bgerstle:
nice catch @negati-ve! thanks for the PR

PR #156, opened 2015-10-09
Fix (and test) face centering in images T114685

bgerstle:
Phab: [T114685](https://phabricator.wikimedia.org/T114685)

---

Finally some tests for this behavior! Thanks again to POTUS for being a great test subject o> 🇺🇸


montehurd:
Seems to work nicely!

PR #159, opened 2015-10-09
(Re)Enable last scroll position restore for continue reading feature

coreyfloyd:
This enables articles shown on startup (restore) or articles shown from the continue reading section to restore their previous offset (using the legacy logic which used the discovery method to decide whether or not to scroll a user to the previous offset)

**IMPORTANT:** For evaluating this pull request, please look at each commit in order. They explain why changes were made along the way. Its probably best i walk you through.

TL;DR
Peeled an onion here. Basically, we needed to pass down the discovery method to the web view controller to get this working again. This involved changing some APIs, and from there I had decisions of whether to update old (soon to be deleted) code to work with the new APIs or delete the old code. I went with the latter. As noted above, I left a detailed story along the way to explain why some refactors were completed.


bgerstle:
this is a great first step to cleaning up webVC's responsibilities, thanks for rolling up your sleeves to get this done @coreyfloyd!  how much of the legacy functionality were you able to test after refactoring?  IIRC the edit button is broken on master, which probably blocks you from testing some of those flows. If needed, I can fix it real quick on master so you can merge it here, then we can do some quick smoke tests to make sure all the delegation is hooked up properly

coreyfloyd:
@bgerstle I wired up the editing so it works now

bgerstle:
ah, i see now that [you did](https://github.com/wikimedia/wikipedia-ios/pull/159/files#diff-ec4091178de488a1f0d40b5f8a31887aR576). thanks!

bgerstle:
@coreyfloyd Travis failed due to [missing import of `MWKHistoryEntry`](https://travis-ci.org/wikimedia/wikipedia-ios/builds/84975666#L1196)

coreyfloyd:
@bgerstle what the what! but it builds locally… hmmm

coreyfloyd:
oh its in a test…

coreyfloyd:
forgot to push the header… doh

bgerstle:
@coreyfloyd please merge master and @montehurd please review

montehurd:
**Functional regressions:**
- Links to sections no longer work - i.e. "/wiki/Family_of_Barack_Obama#Paternal_relations", which you can find and test on the "Community organizer and Harvard Law School" section of the Obama article - there's a "paternal relatives" link part way down the paragraph.

- Saved pages don't work - load and article, save it, browse away from it or quit app, disconnect internet, try to choose same page from saved pages.

- TOC broken in some cases - Barack Obama, pick "Introduction", then pick "Early life and career", which no longer scrolls to top of page properly.

**Visual regressions:**
- When articles load they're not scrolled to top, even when first time loaded.

- Animating scroll when making a selection from TOC is nice when you're picking a section close-ish to the current part of the article which is onscreen, but on long articles if you pick a section farther away from what's onscreen the animation looks really choppy - formerly the old style TOC did do this, but only because the web view was "zoomed out", which lessened the choppy appearance.

**Proposed solution:**
I say we merge this (tons of great stuff!), then I'd could take a stab at fixing everything except the *saved pages* not working - I think I have a sense for what's going on & how to reconnect things so everything plays nice with our js transforms - started poking around but it's getting late :)

montehurd:
I'll try to push a commit tonight if I can get it all wrangled in the next hour or so...

coreyfloyd:
@montehurd working through the issues…

> When articles load they're not scrolled to top, even when first time loaded.
Fixed

> TOC broken in some cases - Barack Obama, pick "Introduction", then pick "Early life and career", which no longer scrolls to top of page properly.
Hmmm, very weird… going to need to rely on you here for this - it seems like whenever you are on the introduction, selecting another section always scrolls it incorrectly. Maybe you have the fix for this one?

Have a fix up for the multiple setArticle calls… so we were calling this one more time then expected… but 2 times can occur normally (if we have an article, we set it, and then still fetch it with out regard to any cache time). So now it is called 1 or 2 times instead of 2 or 3 times.

> Saved pages don't work - load and article, save it, browse away from it or quit app, disconnect internet, try to choose same page from saved pages.

Saved pages works fine for me, so not sure what the issue is there for you (I just merged the latest from master - maybe that was the problem?).

> Animating scroll when making a selection from TOC is nice when you're picking a section close-ish to the current part of the article which is onscreen, but on long articles if you pick a section farther away from what's onscreen the animation looks really choppy - formerly the old style TOC did do this, but only because the web view was "zoomed out", which lessened the choppy appearance.

This can go to product/designers for testing - this restores functionality from the old app. Although it may look copy for long scrolls, we also know that scrolling without animation looks like a flicker as well, so it will be up to product to decide which is the lesser of 2 evils.

cc @bgerstle

bgerstle:
@coreyfloyd getting lots of missing nullability annotation warnings in WebVC. can you either add assume-nonnull, or comment out the `nullable` annotation on the `article` property?

Wikipedia/View Controllers/WebView/WebViewController.h
```diff
@@ -70,6 +48,8 @@
 
 @protocol WMFWebViewControllerDelegate <NSObject>
 
+- (void)webViewController:(WebViewController*)controller didLoadArticle:(MWKArticle*)article;
```
bgerstle: is there a particular reason why we're passing an article instead of a title?

coreyfloyd: The web view controller is set with an article, not a title.

bgerstle: ah right, because it's sole concern is about content and not navigation :+1:

bgerstle:
@coreyfloyd looks good to me! :+1:

coreyfloyd:
> right, just wanted to clarify this was intentional, and that you judged this (implicit) contract to not be worthy of an assertion. thinking about it some more, in this case since it's only related to tweaks I think it's fine to not assert here.

PR #158, opened 2015-10-09
Add Facebook tweaks for testing continue reading

coreyfloyd:
Shake it up!

PR #161, opened 2015-10-10
Add Tweaks, update Pods

bgerstle:
Based on #158, but:

- Integrates Tweaks via CocoaPods
- updates other Pods 
  - AFNetworking now has nullability annotations!
  - FBSnapshotTestCase reference img dir needed to be updated in build settings
- updated CocoaPods to 0.39.0
- wrapped all imports and usage of Piwik and Tweaks with preprocessor flag checks and updated `Podfile` to only link then for non-App-Store builds

Demo: https://vid.me/wpIh


bgerstle:
@coreyfloyd the tweaks issue was due to `use_frameworks!` which is [resolved](https://github.com/facebook/Tweaks/issues/86) in the latest version. fixed that and did a couple other things in this PR instead of pushing to yours.  I'm fine w/ this, but there is an issue w/ unsatisfiable constraints when we reload the sections. If you can fix it quickly that'd be great, otherwise I can file a ticket for a follow-on.

coreyfloyd:
@bgerstle added my window stuff - I'm good to go with this when we clean up the macro stuff

bgerstle:
@coreyfloyd should have already been cleaned up

bgerstle:
@coreyfloyd see 988e55c

PR #162, opened 2015-10-12
Fix inappropriate truncation of image filenames/extensions

mdholloway:
Bug: T104576

### DIFF
--- a/Wikipedia/mw-utils/WMFImageURLParsing.m
+++ b/Wikipedia/mw-utils/WMFImageURLParsing.m
@@ -6,7 +6,7 @@ static NSRegularExpression* WMFImageURLParsingRegex() {
     dispatch_once(&onceToken, ^{
         // TODO: try to read serialized regex from disk to prevent needless pattern compilation on next app run
         NSError* patternCompilationError;
-        imageNameFromURLRegex = [NSRegularExpression regularExpressionWithPattern:@"^\\d+px-([^.]*\\.[^.]*).*$"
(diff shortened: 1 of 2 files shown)


bgerstle:
thanks @mdholloway! would you mind adding some tests (or comments) to explain what this fixes? happy to pair with you on it tomorrow or tuesday

bgerstle:
or, if you're not up for writing ObjC/Swift, pointers to test cases or example strings & outputs would be nice so we can add the tests

mdholloway:
Will do!  I've been expanding my horizons with some iOS stuff this weekend; writing a test or two would be good practice.

mdholloway:
Any idea why I'd be getting an "'OCHamcrest/OCHamcrest.h' file not found" error when I try to run the tests?  It's definitely there in the project...

mdholloway:
I added a test but I figure I'll wait until I can get the tests running locally to verify it before adding it to the PR.  

But to describe what the patch does, the old regex was capturing all non-period characters after the (optional) size prefix, followed by a period and all non-period characters following that period until another period was encountered.  I think this was intended to truncate all file extensions beyond the first.  But this was being fooled by filenames containing periods into truncating too early.  So what I did is to add a check so that the expression captures:
1.  Up to any number of non-period characters, then
2.  Up to any number of instances of a pattern consisting of a period followed by four or more characters other than another period or the end of a line, then
3.  A period followed by any number of non-period characters (that will necessarily be three or fewer).

bgerstle:
@mdholloway great thanks! i have a fix on #161 for the false compiler errors, and I think I know what's causing the funny coverage data. going to re-run the build to confirm, but otherwise this is good to merge :+1:

mdholloway:
Sweet!

PR #163, opened 2015-10-12
Fix and remove preview

bgerstle:
After updating Mantle, we weren't able to successfully serialize article previews, which was preventing us from getting the full article payload.  So I...

- Fixed article preview serialization
- Realized we weren't doing anything w/ article preview data
- Removed `articlePreviewFetcher` from `WMFArticleContainerViewController`.

Maybe we'll use it as an optimization later, but at least we have something that works now.


bgerstle:
@coreyfloyd but that doesn't tell if it was successful or not:

> This property is 0.0 if the string doesn’t begin with a valid text representation of a floating-point number.

This way, we know for sure if we got something that isn't a number and can report it properly.

PR #165, opened 2015-10-13
clip header images to bounds

bgerstle:
prevents them from overlapping while scrolling:

### Before
image

### After
image


coreyfloyd:
:+1:

PR #167, opened 2015-10-14
Update Nearby title

coreyfloyd:


bgerstle:
@coreyfloyd please change these to use `main-menu-nearby` localized string

PR #172, opened 2015-10-15
Manually generate test crashes, plus minor fastlane & logging tweaks

bgerstle:
Phab: [T115384](https://phabricator.wikimedia.org/T115384)

---

#### :checkered_flag: Ready for merge! :checkered_flag: 
**[Verified!](https://rink.hockeyapp.net/manage/apps/206527/app_versions/93/crash_reasons/48758954)**

Added more logs & update fastlane as part of an attempt to fix changelogs & manual crash reporting.  However, it seems that [Hockey doesn't generate test crashes in AppStore](https://github.com/bitstadium/HockeySDK-iOS/blob/a63f18c6742841e574545693e05445f71a33927c/Classes/BITCrashManager.m#L730), and [TestFlight is currently also considered the App Store](https://github.com/bitstadium/HockeySDK-iOS/issues/212).  For now, we're going to use our own `generateTestCrash` method which will skip checks for the current environment, and rely on whether or not the button is shown to disable functionality in the App Store.

Regarding changelogs, it seems we can't really set "What To Test" for Alpha builds via fastlane, since it requires the `skip_submission` flag, which is meant for beta (external testing) only.  Will follow-up w/ KrauseFX to see if there's a way to do it, but low-priority.


bgerstle:
@coreyfloyd please review

PR #173, opened 2015-10-15
List fixes, and lots of tests

bgerstle:
Phab: [T104818](https://phabricator.wikimedia.org/T104818)

- [x] Fix recent search list
- [x] Prepend new saved pages to the top of the list
- [x] Migrate legacy saved page plist data by reversing order before import

---

So, in going to implement the (seemingly simple) feature to put the most recent saved page at the top of the list, I discovered that the strategy I wanted to use is currently being employed by `MWKRecentSearchList` :tada:

Unfortunately, there's a bug that reverses the list every time you save it:

1. [Initial state](https://cloud.githubusercontent.com/assets/444217/10521094/e5a2870a-733b-11e5-86e1-049018e13a72.png)
2. [Added a new search](https://cloud.githubusercontent.com/assets/444217/10521111/f6c20b82-733b-11e5-9b47-8e2fbea515b4.png)
3. [Restarted the app](https://cloud.githubusercontent.com/assets/444217/10521117/007e6800-733c-11e5-99b6-16ac493c501c.png) :scream:

So, I set out on a mission to make sure all our lists behave as we expect with a reasonable amount of effort. I think I've done that, here's how the cookie crumbled:

**TL;DR;** As part of getting 5.0 out the door, we _need_ to have some confidence in MediaWikiKit behavior.  I feel strongly that adding these tests was necessary, despite knowing that they'll be deleted once we go to CoreData.  Part of dealing with legacy code is testing it not only to make sure it works before you change it, but also to have a codified list of requirements for any proposed rewriting or re-architecting.

### `MWKRandom` 
`MWKRandom` is a protocol which I'm using to simplify creation of (model) objects during tests.  The rationale is that we usually don't care what the properties are, as long as they're valid and/or unique.  This makes tests much more terse & expressive, while allowing me to abstract away a lot of shared behaviors.

### Shared Behaviors
This is a concept incorporated into many popular BDD frameworks.[0][1]  I think we should use this as a turning point to evaluate whether we're starting to outgrow XCTest, and I've definitely gotten a better understanding of what using such a framework might provide as we begin to ramp up our testing practices.  Here, we're using plain-old XCTest w/ a little inheritance & abstraction to share `test*` methods across subclasses, which override a couple of methods to specify the class being exercised and the inputs to the test.  Namely, the `MWKList` (sub)class to test and the kind of entries to use.  The shared tests all rely on the generation of unique list entries to abstractly verify the addition and removal of new and existing entries.  As part of this, I was also able to remove some (now redundant) `MWKHistoryList` tests.

### Improved `MWKList` subclass API encapsulation
As part of this work, I wanted to ensure that all `MWKList` public APIs did not expose `insertEntry:atIndex:`, so as to ensure that `addEntry:` was the only way to add things to a list, enforcing any append/prepend contracts for that specialized list.

### Improved `MWKSavedPageList` migration encapsulation
While I was making that improvement, I also took a minute to encapsulate the `MWKSavedPageList` legacy image data migration flags.  Now the migration flag updating API is only exposed in the `MWKSavedPageList+ImageMigration` header, while testing APIs (for adding unmigrated entries) in the `MWKSavedPageList+ImageMigrationTesting`.  As for the flag itself, `MWKSavedPageEntry` only exposes the flag as readable, (needed to filter out all migrated entries), but writable within a separate category which is used by `MWKSavedPageList`.  

> MWK data layer rant: This is yet another reason why having dedicated "List" objects adds a _lot_ of incidental complexity

### Other minor things
Fixing an `MWKSection` test and taking an indulgent moment to sweeten the `XCTestCase+PromiseKit` ObjC API.  Still thinking about whether it's ready for a PR to the main repo as an add'l subspec for testing w/ promises.  Would appreciate feedback if you have any.

## Food for thought :cake:
This is also something we can look back on and decide if we want to look into property-based testing—of which I'm a big fan. [2][3]

0: http://blog.pivotal.io/labs/labs/drying-up-jasmine-specs-with-shared-behavior
1: http://artandlogic.com/2014/02/specta-shared-behavior/
2: https://github.com/codafi/swiftcheck
3: https://github.com/jeffh/Fox


MediaWikiKit/MediaWikiKit/MWKSavedPageList.m
```diff
@@ -35,14 +43,18 @@ - (instancetype)initWithDataStore:(MWKDataStore*)dataStore {
     return self;
 }
 
-#pragma mark - Entry Access
-
-- (MWKSavedPageEntry*)entryAtIndex:(NSUInteger)index {
-    return [super entryAtIndex:index];
+- (void)importEntries:(NSArray*)entries {
+    NSArray<MWKSavedPageEntry*>* validEntries = [entries bk_reject:^BOOL (MWKSavedPageEntry* entry) {
```
coreyfloyd: Any reason not to use the same valid entry pattern as the recent search list?

```
- (BOOL)isEntryValid:(MWKRecentSearchEntry*)entry {
```

MediaWikiKit/MediaWikiKit/MWKSavedPageList.m
```diff
@@ -35,14 +43,18 @@ - (instancetype)initWithDataStore:(MWKDataStore*)dataStore {
...
+        return entry.title.text.length == 0;
+    }];
+    NSArray<MWKSavedPageEntry*>* uniqueValidEntries = [[NSOrderedSet orderedSetWithArray:validEntries] array];
```
coreyfloyd: If NSOrderedSet encounters a duplicate - Which one does it keep? Do we care? Just curious if we know or if it even matters.

MediaWikiKit/MediaWikiKit/MWKSavedPageList.m
```diff
@@ -106,12 +117,39 @@ - (void)performSaveWithCompletion:(dispatch_block_t)completion error:(WMFErrorHa
...
-- (NSArray*)dataExport {
-    return [self.entries bk_map:^id (MWKSavedPageEntry* obj) {
-        return [obj dataExport];
+- (NSArray<NSDictionary*>*)exportedEntries {
+    return [self.entries bk_map:^NSDictionary*(MWKSavedPageEntry* entry) {
+        return [entry dataExport];
     }];
 }
 
+- (NSDictionary*)dataExport {
```
coreyfloyd: I can't believe all of this isn't implemented with NSCoding… :disappointed:

MediaWikiKit/MediaWikiKit/MWKRecentSearchList.m
```diff
@@ -37,22 +38,28 @@ - (instancetype)initWithDataStore:(MWKDataStore*)dataStore {
...
 
+- (void)importEntries:(NSArray*)entries {
+    [super importEntries:[entries bk_select:^BOOL (MWKRecentSearchEntry* entry) {
+        return [self isEntryValid:entry];
+    }]];
+}
+
 - (void)addEntry:(MWKRecentSearchEntry*)entry {
-    if (entry.searchTerm == nil) {
+    if (![self isEntryValid:entry]) {
```
coreyfloyd: Should this be a super class method? 

I guess this is related to the other comment… are all the lists validating data differently?

MediaWikiKit/MediaWikiKit/MWKSavedPageList.m
bgerstle: i know, i _almost_ did it for generic `MWKList` serialization. then i stepped back and remembered this is all going away

MediaWikiKit/MediaWikiKit/MWKSavedPageList.m
bgerstle: i'm pretty sure it would keep the one already in the list. it's probably an edge case and i'm not sure it matters which one we keep, or which one we'd _want_ to keep. probably whichever one had a newer date and/or not-unknown discovery method. i saw we can implement custom-dedupe logic if/when we get a bug report

MediaWikiKit/MediaWikiKit/MWKSavedPageList.m
coreyfloyd: Yeah i thought the same thing in my stuff now…

coreyfloyd:
@bgerstle I'm good here. Left comments about the validation logic, but really I don't think it needs to be consolidated for approval. Its something that seems like it might want to be on MWKList.

:+1:

coreyfloyd:
@montehurd feel free to merge if you are good after the tests are passing

MediaWikiKit/MediaWikiKit/MWKSavedPageList.m
bgerstle: got lazy / wasn't sure if/where the best place was for validation logic of list entries.  e.g. we should only need to validate on import and rely on nullability annotations & assertions to catch the rest. but prior experience tells us we should probably assert & have fallbacks anyway. OK if i file a ticket to harden all `MWKList` objects against bad data?

bgerstle:
> Its something that seems like it might want to be on MWKList.
Possibly, but it's tricky since the "bail if valid" behavior doesn't propagate through inheritance w/o some API trickery. can look into it as part of the follow-on ticket i mentioned for validation

bgerstle:
:scream: travis how could you?!

PR #175, opened 2015-10-15
Hide lead image if no article image or if mainpage.

montehurd:
The lead image (above web view article) was not collapsed when no image exists for article.

https://phabricator.wikimedia.org/T115417
https://phabricator.wikimedia.org/T115540

Before
soltan after
After
soltan before

Before
viral after
After
viral before

Before
theorem after
After
theorem before

Before
main afterAfter
main before


bgerstle:
:+1:

PR #176, opened 2015-10-16
Made ArticleListCollection VC refresh cells if cell article is saved.

montehurd:
Fixes this:
https://phabricator.wikimedia.org/T113738

**Also fixes following bug:**
-Go to recent.
-Tap first article card.
-Tap on a link in that article.
-Go back to recent.
-The card for the link-tap article will only be showing its title - not summary or image.

**Before:**
before

**After:**
after

### DIFF
--- a/Wikipedia/UI-V5/WMFArticleListCollectionViewController.h
+++ b/Wikipedia/UI-V5/WMFArticleListCollectionViewController.h
@@ -32,8 +32,6 @@ typedef NS_ENUM (NSUInteger, WMFArticleListMode) {
 
 - (SelfSizingWaterfallCollectionViewLayout*)flowLayout;
 
-- (void)refreshVisibleCells;
-
 - (void)scrollToArticle:(MWKArticle*)article animated:(BOOL)animated;
 
 - (void)scrollToArticleIfOffscreen:(MWKArticle*)article animated:(BOOL)animated;
--- a/Wikipedia/UI-V5/WMFArticleListCollectionViewController.m
+++ b/Wikipedia/UI-V5/WMFArticleListCollectionViewController.m
@@ -25,6 +25,7 @@
 #import "UIViewController+WMFArticlePresentation.h"
 
 #import "UIColor+WMFHexColor.h"
+#import <BlocksKit/BlocksKit.h>
 
 @interface WMFArticleListCollectionViewController ()
 <UICollectionViewDelegate, WMFSearchPresentationDelegate, WMFEditingCollectionViewLayoutDelegate>
@@ -54,6 +55,7 @@
 }
 
 - (void)commonInit {
+    [self observeArticleUpdates];
     self.navigationItem.rightBarButtonItem = [self wmf_searchBarButtonItemWithDelegate:self];
 }
 
@@ -126,10 +128,34 @@
     return [NSString stringWithFormat:@"%@ dataSourceClass: %@", self, [self.dataSource class]];
 }
 
-- (void)refreshVisibleCells {
-    [self.collectionView wmf_enumerateVisibleCellsUsingBlock:
-     ^(WMFArticlePreviewCell* cell, NSIndexPath* path, BOOL* _) {
+#pragma mark - Stay Fresh... yo
+
+- (void)observeArticleUpdates {
+    [[NSNotificationCenter defaultCenter] removeObserver:self name:MWKArticleSavedNotification object:nil];
+    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(articleUpdatedWithNotification:) name:MWKArticleSavedNotification object:nil];
+}
+
+- (void)unobserveArticleUpdates {
+    [[NSNotificationCenter defaultCenter] removeObserver:self name:MWKArticleSavedNotification object:nil];
+}
+
+- (void)articleUpdatedWithNotification:(NSNotification*)note {
+    MWKArticle* article = note.userInfo[MWKArticleKey];
+    [self refreshAnyVisibleCellsWhichAreShowingTitle:article.title];
+}
+
+- (void)refreshAnyVisibleCellsWhichAreShowingTitle:(MWKTitle*)title {
+    NSArray* indexPathsToRefresh = [[self.collectionView indexPathsForVisibleItems] bk_select:^BOOL (NSIndexPath* path) {
+        WMFArticlePreviewCell* cell = (WMFArticlePreviewCell*)[self.collectionView cellForItemAtIndexPath:path];
+        return (cell.title == title);
     }];
+    if (indexPathsToRefresh.count > 0) {
+        [self.collectionView reloadItemsAtIndexPaths:indexPathsToRefresh];
+    }
+}
+
+- (void)dealloc {
+    [self unobserveArticleUpdates];
 }
 
 #pragma mark - DataSource and Collection View Wiring
--- a/Wikipedia/UI-V5/WMFArticleListCollectionViewController.m
+++ b/Wikipedia/UI-V5/WMFArticleListCollectionViewController.m
@@ -144,13 +144,11 @@
 }
 
 - (void)refreshAnyVisibleCellsWhichAreShowingTitle:(MWKTitle*)title {
-    NSArray* indexPathsToRefresh = [[self.collectionView indexPathsForVisibleItems] bk_select:^BOOL (NSIndexPath* path) {
-        WMFArticlePreviewCell* cell = (WMFArticlePreviewCell*)[self.collectionView cellForItemAtIndexPath:path];
-        return (cell.title == title);
+    NSArray* indexPathsToRefresh = [[self.collectionView indexPathsForVisibleItems] bk_select:^BOOL (NSIndexPath* indexPath) {
+        MWKArticle* article = [self.dataSource articleForIndexPath:indexPath];
+        return [article.title isEqualToTitle:title];
     }];
-    if (indexPathsToRefresh.count > 0) {
-        [self.collectionView reloadItemsAtIndexPaths:indexPathsToRefresh];
-    }
+    [self.collectionView reloadItemsAtIndexPaths:indexPathsToRefresh];
 }
 
 - (void)dealloc {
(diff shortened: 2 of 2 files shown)


Wikipedia/UI-V5/WMFArticleListCollectionViewController.h
```diff
@@ -32,8 +32,6 @@ typedef NS_ENUM (NSUInteger, WMFArticleListMode) {
 
 - (SelfSizingWaterfallCollectionViewLayout*)flowLayout;
 
-- (void)refreshVisibleCells;
```
montehurd: This was not called by anything. Nor did its implementation do anything.

Wikipedia/UI-V5/WMFArticleListCollectionViewController.m
```diff
@@ -126,10 +128,34 @@ - (NSString*)debugDescription {
...
+
+- (void)articleUpdatedWithNotification:(NSNotification*)note {
+    MWKArticle* article = note.userInfo[MWKArticleKey];
+    [self refreshAnyVisibleCellsWhichAreShowingTitle:article.title];
+}
+
+- (void)refreshAnyVisibleCellsWhichAreShowingTitle:(MWKTitle*)title {
+    NSArray* indexPathsToRefresh = [[self.collectionView indexPathsForVisibleItems] bk_select:^BOOL (NSIndexPath* path) {
+        WMFArticlePreviewCell* cell = (WMFArticlePreviewCell*)[self.collectionView cellForItemAtIndexPath:path];
+        return (cell.title == title);
```
bgerstle: this would be ok in Swift, but in ObjC always use `isEqualToTitle:` (or `isEqual:` if a class-specific equality method isn't available). checking **identity** is _usually_ not what you want for model objects. i.e. we want to know if they're equivalent—even if they're different objects.

Wikipedia/UI-V5/WMFArticleListCollectionViewController.m
```diff
@@ -54,6 +55,7 @@ - (instancetype)initWithCoder:(NSCoder*)aDecoder {
 }
 
 - (void)commonInit {
+    [self observeArticleUpdates];
```
bgerstle: doing anything that touches views in `init` of a view controller is ill-advised. in this case, an article download elsewhere could have the unintended side effect of forcing the view to load prematurely.  we should at least wait until `viewDidLoad`. IMHO it would be more efficient to observe (and update) on `viewWillAppear` and un-observe in `viewDidDisappear`.  This means views which aren't shown are inert, and will **lazily** update when they're shown again.

Wikipedia/UI-V5/WMFArticleListCollectionViewController.m
```diff
@@ -126,10 +128,34 @@ - (NSString*)debugDescription {
     return [NSString stringWithFormat:@"%@ dataSourceClass: %@", self, [self.dataSource class]];
 }
 
-- (void)refreshVisibleCells {
-    [self.collectionView wmf_enumerateVisibleCellsUsingBlock:
-     ^(WMFArticlePreviewCell* cell, NSIndexPath* path, BOOL* _) {
+#pragma mark - Stay Fresh... yo
+
+- (void)observeArticleUpdates {
+    [[NSNotificationCenter defaultCenter] removeObserver:self name:MWKArticleSavedNotification object:nil];
```
bgerstle: 👌🏻 nice enforcing of balanced observation. however, might as well call `unobserveArticleUpdates` instead of repeating the impl up here.

Wikipedia/UI-V5/WMFArticleListCollectionViewController.m
bgerstle: note that if you agree with my rationale and decide to move observation in to `willAppear`/`didDisappear` that the unobservation in `dealloc` becomes unnecessary because 1) we only observe if the view appears and 2) disappear methods should always be called before dealloc if the view was shown.

Wikipedia/UI-V5/WMFArticleListCollectionViewController.m
```diff
@@ -126,10 +128,34 @@ - (NSString*)debugDescription {
...
+    [self refreshAnyVisibleCellsWhichAreShowingTitle:article.title];
+}
+
+- (void)refreshAnyVisibleCellsWhichAreShowingTitle:(MWKTitle*)title {
+    NSArray* indexPathsToRefresh = [[self.collectionView indexPathsForVisibleItems] bk_select:^BOOL (NSIndexPath* path) {
+        WMFArticlePreviewCell* cell = (WMFArticlePreviewCell*)[self.collectionView cellForItemAtIndexPath:path];
+        return (cell.title == title);
     }];
+    if (indexPathsToRefresh.count > 0) {
+        [self.collectionView reloadItemsAtIndexPaths:indexPathsToRefresh];
```
bgerstle: FYI: `UICollectionView` declares `indexPathsForVisibleItems` to be `nonnull`, so passing the result of your `bk_select:` directly to this method (w/o wrapping in an if) should be safe. also nice use of BlocksKit 👌🏻

Wikipedia/UI-V5/WMFArticleListCollectionViewController.m
```diff
@@ -126,10 +128,34 @@ - (NSString*)debugDescription {
...
+}
+
+- (void)articleUpdatedWithNotification:(NSNotification*)note {
+    MWKArticle* article = note.userInfo[MWKArticleKey];
+    [self refreshAnyVisibleCellsWhichAreShowingTitle:article.title];
+}
+
+- (void)refreshAnyVisibleCellsWhichAreShowingTitle:(MWKTitle*)title {
+    NSArray* indexPathsToRefresh = [[self.collectionView indexPathsForVisibleItems] bk_select:^BOOL (NSIndexPath* path) {
+        WMFArticlePreviewCell* cell = (WMFArticlePreviewCell*)[self.collectionView cellForItemAtIndexPath:path];
```
bgerstle: convention for retrieving model objects for an index path is to access them directly, or via any dataSource or other controller, as opposed to its corresponding cell. in general we should only try to access the information we need in a given method, and avoid views if possible. (hint: watch Andy's talk about the "value layer game" :wink:)

Wikipedia/UI-V5/WMFArticleListCollectionViewController.m
```diff
@@ -126,10 +128,34 @@ - (NSString*)debugDescription {
...
+        return (cell.title == title);
     }];
+    if (indexPathsToRefresh.count > 0) {
+        [self.collectionView reloadItemsAtIndexPaths:indexPathsToRefresh];
+    }
+}
+
+- (void)dealloc {
+    [self unobserveArticleUpdates];
```
bgerstle: see above comment about this being obviated by toggling observation in will/didAppear

bgerstle:
nice work @montehurd! just requested a couple amendments. i don't think i'll be available for a video hangout today, but you can call my cell if you want to talk about the observation stuff I mentioned. if the appear/disappear stuff seems like a premature optimization, i do feel strongly that we should at least wait until `viewDidLoad` to _start_ observing.  we can have a larger discussion about laziness and inertness of views later.

Wikipedia/UI-V5/WMFArticleListCollectionViewController.m
montehurd: @bgerstle doh! really good points - I should have caught this :)

Wikipedia/UI-V5/WMFArticleListCollectionViewController.m
montehurd: Will do!

Wikipedia/UI-V5/WMFArticleListCollectionViewController.m
montehurd: Ah! Makes sense!

Wikipedia/UI-V5/WMFArticleListCollectionViewController.m
montehurd: Man, I must have rushed through this :)

Wikipedia/UI-V5/WMFArticleListCollectionViewController.m
montehurd: Oh neat! I need to remember to check the new null notations for stuff like this!

Wikipedia/UI-V5/WMFArticleListCollectionViewController.m
montehurd: Great feedback! Amending...

Wikipedia/UI-V5/WMFArticleListCollectionViewController.m
montehurd: Hmm... doesn't seem to work :(

Wikipedia/UI-V5/WMFArticleListCollectionViewController.m
montehurd: Interesting... doesn't work with `viewWillAppear` / `didDisappear `. The update is happening when the view isn't visible, so I guess it makes sense. Moved it to `viewDidLoad` and it works though.

Wikipedia/UI-V5/WMFArticleListCollectionViewController.m
montehurd: Ended up moving the start observation method to `viewDidLoad`, so ended up needing this after all.

bgerstle: Did you observe _and_ update in willAppear? Should work

bgerstle: If you do DidLoad only, which is fine, make sure to remove I unobserve from
didDisappear

Wikipedia/UI-V5/WMFArticleListCollectionViewController.m
montehurd: Ya docs say it returns an empty array if no visible items.

montehurd:
@bgerstle Amended! Should be good to go. Did end up having to use `viewDidLoad` for observation start - see notes above.

bgerstle:
👍🏻

PR #177, opened 2015-10-16
WIP: Toc fixes

montehurd:
Fixes:
https://phabricator.wikimedia.org/T115145
https://phabricator.wikimedia.org/T115676

- changed top TOC item to use article title
- added "Read more" to bottom of TOC
- added "Read more" header above read more items at bottom of web view
- tapping "Read more" in TOC scrolls article to "Read more" header
- opening TOC when scrolled to "Read more" section makes "Read more" be selected when TOC appears


MediaWikiKit/MediaWikiKit/MWKArticle.h
```diff
@@ -151,4 +151,6 @@ static const NSInteger kMWKArticleSectionNone = -1;
 
 - (NSString*)summary;
 
+- (NSArray<MWKSection*>*)sectionsWithFakeReadMoreSection;
```
montehurd: Not sure if this is best way to handle this, but at least it's clear :)

Wikipedia/Categories/MWKSection+DisplayHtml.m
```diff
@@ -53,7 +52,7 @@ - (NSString*)getHeaderTag {
                 self.sectionId,
                 self.anchor,
                 self.line,
-                pencilAnchor,
+                [self.anchor isEqualToString:@"TOC_Read_More"] ? @"" : [self getEditPencilAnchor],
```
montehurd: Need to pull @"TOC_Read_More" out to a const string - used in one other place on this diff...

Wikipedia/View Controllers/WebView/WebViewController.m
```diff
@@ -341,17 +341,29 @@ - (CGFloat)clientBoundingRectVerticalOffset {
 }
 
 - (void)scrollToFragment:(NSString*)fragment {
```
montehurd: Ideas for the issue I mention in these temporary comments?

Wikipedia/View Controllers/WebView/WebViewController.m
```diff
@@ -341,17 +341,29 @@ - (CGFloat)clientBoundingRectVerticalOffset {
...
-    if (fragment.length == 0) {
-        // No section so scroll to top. (Used when "Introduction" is selected.)
-        [self.webView.scrollView scrollRectToVisible:CGRectMake(0, 1, 1, 1) animated:NO];
+// Temp fix! See note below about bug with animated scroll approach...
+    if (fragment) {
+        NSString* jumpToThis = [fragment copy];
+        [[self.webView wmf_javascriptContext][@"scrollToFragment"] performSelector:@selector(callWithArguments:) withObject:@[jumpToThis] afterDelay:0.1];
```
montehurd: Short term solution is just to use this, which works. We would loose the scrolling animation, but could address that later when we have time...

Wikipedia/View Controllers/WebView/WebViewController.m
```diff
@@ -341,17 +341,29 @@ - (CGFloat)clientBoundingRectVerticalOffset {
...
+        [self.webView.scrollView scrollRectToVisible:CGRectMake(0, 1, 1, 1) animated:NO];
+    }
+
+    return;
+
+// Bug with animated scrolling method below:
+// - scroll to bottom of read more items at bottom of article
+// - open TOC and select a section above "Read more"
+// - doesn't work
+// I think it's because clientBoundingRectVerticalOffset doesn't return anything if we've scrolled the web view completely up off the screen for some reason.
```
bgerstle: the issue doesn't appear to be w/ `clientBoudningRectVerticalOffset`, but that `getClientBoundingRect()` returns _very_ large negative numbers when invoked while the browser view is offscreen:

```
> document.getElementById("Birth_and_childhood").getBoundingClientRect()
< ClientRect = $11
bottom: -33552278
height: 43.1875
left: -33554416
right: -33554072
top: -33552322
width: 343
```
Where the scroll view's content size is only 49710, and `clientBoundingRectVerticalOffset` returns 0:

```
(lldb) p (CGFloat) [0x7fcf464570e0 clientBoundingRectVerticalOffset]
(CGFloat) $7 = 0
```

However, I would argue that this is an edge case, and would prefer if we scrolled when the rect is valid, otherwise falling back to setting `location.hash`.  In this case, "valid" would mean that the `elementOrigin` we calculate on line 363 is w/in the bounds of the scrollview's content.

montehurd:
@bgerstle "fake" section needs to be serialized to HTML so... 
- when you select "read more" from the TOC it will actually scroll to web view to the TOC section
- it is styled the same as the other section headers

montehurd:
@bgerstle oh one more - "fake" section needs to be serialized to HTML so...
- when you open the TOC when scrolled to the read more section the TOC can use the same mechanism it already uses to show the read more section is selected

bgerstle:
@montehurd ok, i understand now. thanks for elaborating.

coreyfloyd:
> @bgerstle "fake" section needs to be serialized to HTML so...
when you select "read more" from the TOC it will actually scroll to web view to the TOC section
it is styled the same as the other section headers

@montehurd While a fake section may be a method of accomplishing the scrolling and the styling, they are not required for doing this. There are other ways of doing this without "faking" in the model layer.

montehurd:
@coreyfloyd if we do that we'll have to add special handing for:
- scrolling the web view to the "read more" section if "read more" is selected from the TOC
- scrolling the TOC to "read more" if the TOC is opened when "read more" is scrolled to

montehurd:
@coreyfloyd i'd argue:
While a fake section may be a method of accomplishing the scrolling and the styling, they are not required for doing this. There are other **more complicated** ways of doing this without "faking" in the model layer.

PR #178, opened 2015-10-19
Fix toolbar not showing when pushing article view

bgerstle:
Phab: [T115649](https://phabricator.wikimedia.org/T115649), [T115913](https://phabricator.wikimedia.org/T115913)

---

`toolbarItems` wasn't set in time for the check that determines whether or not to show the toolbar.  Fixed now by moving `setupToolbar` to initialization time since we have (almost) everything we need to populate it.  When the article is set, we check if it's a main page, and if so, update the `toolbarItems` to remove the save & TOC buttons.  Also cleaned up a couple minor things.


montehurd:
@bgerstle @coreyfloyd you guys should be novelist ;)

PR #179, opened 2015-10-19
Remove TOC "Introduction" and use article title.

montehurd:
https://phabricator.wikimedia.org/T115145
Per Josh's comments in ^ : "Suggest we use the article title."


bgerstle:
:+1:

PR #180, opened 2015-10-20
Show previews from Home and all article lists

bgerstle:
:checkered_flag: Ready for review :checkered_flag:


montehurd:
@bgerstle everything seems to still work! :)

coreyfloyd:
@bgerstle Sucks that we have to use the ArticleVC to hold some state that it doesn't use just to keep a hold of the discovery method. Ok for now, there is probably something we can do later to make this nicer when we refactor to have a true "Article Reference" object

PR #182, opened 2015-10-20
Adding new home business logic

coreyfloyd:
https://phabricator.wikimedia.org/T115364
https://phabricator.wikimedia.org/T115371

This adds the new logic for determining order of sections and when to update.

In a nutshell:
- This adds the new section models for today and random
- Updates section ordering to use date sorting
- Adds logic for updating nearby only when the user moves moves more than 500 meters (by "update" I mean update the position in the list, not update the contents of the nearby section)
- Capped number of non-static section to 20 - that may be too many, we can tweak
- Adds some tweaks for testing

Notes:
- I renamed the classes to: `WMFHomeSectionSchema` and `WMFHomeSection`
- This pr does not include logic for restoring the scroll position, that will be in a separate pr following this one
- Though this does include the logic for today an random, it does not include sections for today and random, they are separate tickets/designs


coreyfloyd:
@bgerstle pushed up fixes for most of your comments. If there is anything you want to discus further let me know.

bgerstle:
:+1: code looks good. I think the schema creation logic should eventually be unit tested, though I'm not sure if now is the best time to do it. At the very least, do we have documented acceptance criteria for QA?

bgerstle:
Hit a crash when trying to run locally https://gist.github.com/bgerstle/c5488571f3452a5ae67f. It automatically directed me to beach boys, then I hit "back" and that's when it crashed.

bgerstle:
Not able to repro consistently.. :confused:

coreyfloyd:
@bgerstle The behavior as described in the ticket can be tested… kinda. Much of the functionality requires significant time delays, so there isn't a great practical way to QA those things.

PR #183, opened 2015-10-20
Add Read More as a TOC section

bgerstle:
Phab: [T115545](https://phabricator.wikimedia.org/T115545)

- [x] Refactor TOC to not use `MWKSection` directly
- [x] Add TOC item to represent Read More
- [x] Implement scrolling to read more
- [x] Implement automatic selection of read more
- [x] implement workaround for scrolling to an article section while viewing "Read More"

---

See [this demo](https://vid.me/qNTv) which demonstrates:

- Tapping to Read More from somewhere in article (top & in the middle)
- Scrolling up to a section from Read More and verifying proper highlight in TOC (External Links)
- Tapping on a section from the top of read more (scrolls)
- Tapping on a section from w/in read more w/ browser content hidden (no animation)


bgerstle:
@coreyfloyd @montehurd would appreciate a quick prelim review of the refactor (b35ac54) to ensure it's along the lines of what we discussed before moving forward w/ the remaining items in the above checklist.  for posterity, don't appear to be any visual regressions, and (thanks to using a protocol) the refactor wasn't too invasive (IMO)

image

Wikipedia/UI-V5/WMFTableOfContentsItem.swift
```diff
@@ -0,0 +1,63 @@
...
+    case None
+}
+
+// MARK: - TOC Item
+
+public protocol TableOfContentsItem {
+    var titleText: String { get }
+    var itemType: TableOfContentsItemType { get }
+    var borderType: TableOfContentsBorderType { get }
+    var indentationLevel: UInt { get }
```
coreyfloyd: Should we provide a defeat indentation level as well?

coreyfloyd:
Looks pretty good so far. Enum is a nice way to handle the states.

Wikipedia/UI-V5/WMFTableOfContentsItem.swift
bgerstle: Yeah, I'll make it 0.

bgerstle:
@coreyfloyd @montehurd if you get a chance, please have a look at 03340b3. it has some stuff related to Swift/ObjC interop I'm unsure about, and would like to hear your thoughts on how we handle these scenarios where we want to use Swift lang features but need to interact w/ ObjC classes.  

**TL;DR;** Do we want to pay the cost of some complexity in order to push the project towards Swift more aggressively, or sacrifice the ability to use Swift lang features in order to keep ObjC/Swift interop simpler?  

Though I've currently implemented the former, but I'm having second thoughts and would consider going towards the latter. That said, I'd have to play around w/ it a bit to see if it actually _does_ make the code simpler.

For the nitty gritty: now that the TOC delegate speaks in terms of `TableOfContentsItem` "objects", it can't be an `@objc` protocol since `TableOfContentsItem` has Swift Enum properties.  We could make the ObjC bridging less awkward here by sacrificing Swift Enums (and probably protocol extensions) by making `TableOfContentsItem` an `@objc` protocol, but that results in (somewhat awkward) glue code for `MWKSection` and any other "TOC Item" objects we want to create.

bgerstle:
All that said, I'm almost done w/ the actual functionality. Just need to incorporate read more into the "current section" logic (see stubbed out `visibleFooterViewIndex`)

Wikipedia/View Controllers/WebView/WebViewController.m
```diff
@@ -378,6 +414,10 @@ - (CGFloat)currentVerticalOffset {
     return self.webView.scrollView.contentOffset.y;
 }
 
+- (BOOL)isWebContentVisible {
+    return CGRectIntersectsRect(self.webView.scrollView.wmf_contentFrame, self.webView.wmf_browserView.frame);
```
bgerstle: add `contentInset.top`? fix bug where we try to scroll when webview is barely on screen

coreyfloyd:
:+1:

coreyfloyd:
Getting this merged with a single review so everyone can use in advance of our monday meeting.

PR #184, opened 2015-10-20
Fix for card image aspect ratio across devices.

montehurd:
Bug noted by Katie:
https://phabricator.wikimedia.org/T113001#1737188

Josh said give it a peek but timebox to before lunch.

Problem was the image height from IB was 16 by 9 ratio when shown only on the simulator device I used when testing, but not other devices - i.e. screens with different resolutions. 

Katie said use same height for landscape as we use for portrait.


bgerstle:
@montehurd is this only for cards, or should it apply to the header image gallery as well?

bgerstle:
either way, looks good :+1:

montehurd:
@bgerstle This one's just for cards. Katie had me increase the header image gallery a while ago... not sure if that one is using this same ratio. We can look at that later if they want though :)

bgerstle:
ok looks good to me then. @coreyfloyd will you review and/or merge?

PR #187, opened 2015-10-21
Add scroll position restore to home

coreyfloyd:
This pr saves the "current" section while updates are performed and attempts to restore the scroll position to that section afterwards. It works, but it isn't ideal.

This pr also improves the update logic to no longer require the dispatch after delay hack by using an operation queue to queue updates.


coreyfloyd:
Eh… I'm closing this its still shakey

PR #188, opened 2015-10-21
Fix queuing of updates to the home VC sections

coreyfloyd:
Uses an operation queue to make sure updates are queued properly.

This removes the need for the hack time delay.


bgerstle:
@coreyfloyd I have some questions about the need for queueing, specifically whether we really need a queue, or just a way to trigger another reload after the current one is done (i.e. preventing redundant reloads which are requested while one is taking place).  also would like to clarify w/ you on the need for a custom block operation, and the feasibility of using built-in Foundation classes to implement this logic.

coreyfloyd:
> @coreyfloyd I have some questions about the need for queueing, specifically whether we really need a queue, or just a way to trigger another reload after the current one is done (i.e. preventing redundant reloads which are requested while one is taking place). 

@bgerstle I think requiring a time delay to prevent crashes demonstrates that the performUpdates batching doesn't prevent collisions in update logic. Since queuing now removes the need for the delay *and* prevents crashing, it seems pretty obvious that the issue has been mitigated in a logically consistent way. 

Is there a specific reason that this solution feels unsatisfactory and you are looking for a better one?

> also would like to clarify w/ you on the need for a custom block operation, and the feasibility of using built-in Foundation classes to implement this logic.

Sorry, I thought the name of the operation made the purpose pretty clear - it is for executing blocks that run asynchronous code. Normal block operations require the code be synchronous since they flag themselves as complete as soon as the block finishes executing. Let me know if you need further clarification.

coreyfloyd:
> and the feasibility of using built-in Foundation classes to implement this logic.

@bgerstle note that **ALL** NSOperations except for NSBlockOperation require a custom subclass.

bgerstle:
@coreyfloyd sorry, I think I wasn't clear enough: we do need to prevent update "collisions", but it seems like we wouldn't need to build up a queue of updates: just call it again after the current one finishes to update to the current state. IOW, seems like queueing up the operations would update the collection view multiple times w/ the same state

RE: custom operations—i understand that custom operations require a subclass, but our case is just invoking a block. can't we use `NSBlockOperation` for that? wouldn't adding it to the main operation queue effectively "defer" the operation so it runs "asynchronously"?

bgerstle:
for posterity, we discussed on hangout and determine:

- Fine to accumulate queue items as updates come in until a problem is found w/ that approach
- Custom operation is needed to wait asynchronously for `performBatchUpdates` completion block to be called (otherwise we'd need to do spin lock stuff inside a regular operation on another thread, simpler to set `finished = YES` when completion is called)

@coreyfloyd but we can do it using promises! j/k :+1: @montehurd want to review and/or merge?

PR #189, opened 2015-10-22
Peek link previews

montehurd:
Web view link peeking now works!

- Peeks links to MWKTitles, but not references or other intra-page links:

img_1556

- Fix for crash on iOS 8 - wasn't checking min version before calling peek registration method.

- Fix peek interference by `editing long press recognizer`, which was making search results and saved/recent lists jitter and loose blurring of adjacent items during the initial press before the force had reached the "peek" threshold:

img_1553

- Hide the Safari "Add to reading list" menu on web view long-press on links and other elements.


montehurd:
@bgerstle At least the approach for getting the url of the link the touch is presently over seems to work well :) Ideas for the immediate dealloc?

Wikipedia/UI-V5/WMFArticleContainerViewController.m
```diff
@@ -557,6 +576,71 @@ - (void)sectionEditorFinishedEditing:(SectionEditorViewController*)sectionEditor
...
+
+
+
+#pragma mark - UIViewControllerPreviewingDelegate
+
+- (nullable UIViewController*)previewingContext:(id<UIViewControllerPreviewing>)previewingContext
+                      viewControllerForLocation:(CGPoint)location {
+
+//return
+//[[WMFArticleContainerViewController alloc] initWithArticleTitle:[[MWKTitle alloc] initWithURL:[NSURL URLWithString:@"http://en.wikipedia.org/wiki/Cat"]]
```
montehurd: If I do this it doesn't get dealloc'ed - although it only works if you press on a non-link... beginning to think there js anchor touch events are in the mix...

Wikipedia/UI-V5/WMFArticleContainerViewController.m
```diff
@@ -557,6 +576,71 @@ - (void)sectionEditorFinishedEditing:(SectionEditorViewController*)sectionEditor
...
+
+            
+            
+            MWKTitle* title = [[MWKTitle alloc] initWithURL:[NSURL URLWithString:[href toString]]];
+            //NSLog(@"title = %@", title);
+
+//return nil;
+
+//BUG! The WMFArticleContainerViewController below gets dealloc'ed right away...
+            return
```
montehurd: WHY U GET DEALLOC'ED?

Not owned by any nav controller at this point?

Wikipedia/UI-V5/WMFArticleContainerViewController.m
```diff
@@ -557,6 +576,71 @@ - (void)sectionEditorFinishedEditing:(SectionEditorViewController*)sectionEditor
...
+            [[WMFArticleContainerViewController alloc] initWithArticleTitle:title
+                                                                  dataStore:self.dataStore
+                                                            discoveryMethod:self.discoveryMethod];
+        }
```
montehurd: WHY U GET DEALLOC'ED?

Not owned by any nav controller at this point?

Wikipedia/UI-V5/WMFArticleContainerViewController.m
```diff
@@ -557,6 +576,71 @@ - (void)sectionEditorFinishedEditing:(SectionEditorViewController*)sectionEditor
...
+                      viewControllerForLocation:(CGPoint)location {
+
+//return
+//[[WMFArticleContainerViewController alloc] initWithArticleTitle:[[MWKTitle alloc] initWithURL:[NSURL URLWithString:@"http://en.wikipedia.org/wiki/Cat"]]
+//                                                      dataStore:self.dataStore
+//                                                discoveryMethod:self.discoveryMethod];
+
+
+    
+    JSValue *element = [[self.webViewController.webView wmf_javascriptContext][@"getElementFromPoint"] callWithArguments:@[@(location.x), @(location.y)]];
```
bgerstle: you're probably planning on doing this already, but everything from here down to line 626 should be refactored into WebVC, which could expose a higher-level method like `- (nullable NSDictionary*)webViewElementAtPoint:(CGPoint)`

Wikipedia/UI-V5/WMFArticleContainerViewController.m
bgerstle: doesn't need to be, the previewing context should own it from here...

Wikipedia/UI-V5/WMFArticleContainerViewController.m
```diff
@@ -340,8 +343,24 @@ - (void)viewDidLoad {
 
     self.article = [self.dataStore existingArticleWithTitle:self.articleTitle];
     [self fetchArticle];
+
+
+
+[self registerForPreviewingWithDelegate:self sourceView:[self.webViewController.webView wmf_browserView]];
```
bgerstle: ideally, we wouldn't need to expose the WebVC's `webView` (or the underlying browser view) publicly—in the interest of keeping WebVC's abstraction from leaking and keeping its responsibilities well encapsulated.  we could instead offer an API that allows users to `registerForBrowserPreviewsWithDelegate:(id<UIViewControllerPreviewing>)`, thereby keeping knowledge of the browser view tucked into the WebVC

Wikipedia/UI-V5/WMFArticleContainerViewController.m
bgerstle: weird, you think evaluating JS breaks something internally?

bgerstle: Try disabling all our JS stuff that does touch handling (i.e. return early from `www/listeners.js#maybeSendMessageForTarget`) to see if that makes a difference. If so, we might need to make some more invasive changes w/ how we intercept navigation interactions...

Wikipedia/UI-V5/WMFArticleContainerViewController.m
bgerstle: My guess is that the preview is being cancelled by something. Have a close look at the stack track from w/in `dealloc` to try to find out what's triggering it

www/less/misc.less
```diff
@@ -47,3 +47,8 @@ a:hover {
     margin-bottom: 5px;
 }
 
+// WARNING! TEMPORARY FOR TESTING FORCE TOUCH. DO NO MERGE!
+* {
```
bgerstle: hmmm, this could be a non-starter for this approach. if you're able to get the preview callbacks above to work, please re-enable this to see if you can still select text **and** peek on the page. if not, might need to go back to the drawing board

Wikipedia/UI-V5/WMFArticleContainerViewController.m
bgerstle: Oh never mind, I see you did this already by disabling `handleTouchEnded`. Unfortunately there's not much I can do to figure this out w/o a device. @coreyfloyd would you have time to play around with this?

Wikipedia/View Controllers/WebView/WebViewController.m
```diff
@@ -915,4 +918,31 @@ - (void)editHistoryButtonPushed {
...
+
+- (MWKTitle*)titleForHTMLElement:(JSValue*)element {
+    if ([[[element valueForProperty:@"tagName"] toString] isEqualToString:@"A"] && [element valueForProperty:@"href"]) {
+        NSURL* url = [NSURL URLWithString:[[element valueForProperty:@"href"] toString]];
+        // Don't peek intra-page fragment links - such as references.
+        if ([url wmf_isIntraPageFragment]) {
+            return nil;
+        }
+        return [[MWKTitle alloc] initWithURL:url];
+    }
```
montehurd: There's a bug - MWKTitle's `initWithURL` returns a non-nil MWKTitle for external URLs...

bgerstle: it's tricky, IMO creating a title w/ https://en.wikipedia.org/wiki/Cat is perfectly valid.  even "external links" can go to another wiki, right?  i don't think that `MWKTitle` should implement the "what kind of link this is" business logic, opting instead for maybe the WebVC to implement the following logic for **links on wiki pages**:

- Link is absolute (prefix of `(http(s):)//`)
  - domain contains `wikipedia.org`: create `MWKTitle` and ask delegate handle link tap for title
  - otherwise: open external URL
- Link is a relative path (prefix of `/`): create `MWKTitle` relative to the site where the link was found and ask delegate to handle link tap for title
- Link is a fragment (prefix of `#`): check for conventional prefixes (citation, section, etc.) and open accordingly

bgerstle:
Nice work @montehurd! The `sourceRect` stuff is a nice improvement too :smile:

Wikipedia/UI-V5/WMFArticleContainerViewController.m
```diff
@@ -557,6 +564,34 @@ - (void)sectionEditorFinishedEditing:(SectionEditorViewController*)sectionEditor
...
+    JSValue* element          = [self.webViewController htmlElementAtLocation:location];
+    MWKTitle* titleForElement = [self.webViewController titleForHTMLElement:element];
+    if (titleForElement) {
+        self.webViewController.isPeeking = YES;
+        previewingContext.sourceRect     = [self.webViewController rectForHTMLElement:element];
+        return [[WMFArticleContainerViewController alloc] initWithArticleTitle:titleForElement
+                                                                     dataStore:self.dataStore
+                                                               discoveryMethod:self.discoveryMethod];
+    }
+    return nil;
```
coreyfloyd: What about external links? Should we just be able to get the url back and peek that as well?

bgerstle: good point, we should still support that

montehurd: Follow on.

bgerstle:
:+1: @coreyfloyd please re-review and/or merge

PR #190, opened 2015-10-22
Prevent lead paragraph shifting from terminating prematurely

mdholloway:
This is a (slightly adapted) Javascript transform change we merged recently for the Android app that Elena mentioned you guys might also find useful.  It updates the lead paragraph shifting logic to create and move up a span consisting of the first qualifying P element encountered and any subsequent non-P elements until the next P is encountered.

Simply moving the first P element up, as we did before,  can result in elements appearing between the first paragraph as designated by P tags and other elements (such as an unnumbered list) that may be intended as part of the first display paragraph (see, e.g., "Planet" on enwiki).

Bug: T111958

### DIFF
--- a/Wikipedia/assets/bundle.js
+++ b/Wikipedia/assets/bundle.js
@@ -880,6 +880,64 @@ transformer.register( "hideRedlinks", function( content ) {
 },{"../transformer":8}],15:[function(require,module,exports){
 var transformer = require("../transformer");
 
+function findFirstGoodParagraphIn( nodes ) {
+    var minHeight = 40;
+    var firstGoodParagraphIndex;
+    var i;
+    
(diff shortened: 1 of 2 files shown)


mdholloway:
Yipes, not sure how the formatting in my comment got screwed up, sorry!

montehurd:
@mdholloway will test this out! thanks michael!

montehurd:
@bgerstle @coreyfloyd I'll pick this apart tomorrow :)

bgerstle:
@mdholloway i'm :-1: on behalf of @montehurd who mentioned that there were some issues on IRC. he should be reaching out to you once he gets into the office today

montehurd:
Chatted with Michael earlier - closing in favor of a DocumentFragment approach.

PR #194, opened 2015-10-26
Fix crash when interacting with toolbar while article loads

bgerstle:
Phab: [T116469](https://phabricator.wikimedia.org/T116469)

---

This disables interaction w/ article-sensitive toolbar itms until data is available (see screenshot below).  Doing so required we keep refs to toolbar items for easy access.  Similar logic can be employed when we implement [T112876](https://phabricator.wikimedia.org/T112876).

> Caveat: _technically_ there's race condition between when the DOM finishes being loaded and transformed and when we try to query it for user location.  The window of time should be relatively small on most newer devices, but ideally we'd find a way to make interactions requiring data from the DOM optimistically asynchronous.  That said, the app should currently throw an assertion if we couldn't get DOM information in debug, and do nothing in production.

### DIFF
--- a/Wikipedia/UI-V5/WMFArticleContainerViewController.m
+++ b/Wikipedia/UI-V5/WMFArticleContainerViewController.m
@@ -324,7 +324,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (UIBarButtonItem*)saveToolbarItem {
     if (!_saveToolbarItem) {
-        return [[UIBarButtonItem alloc] initWithCustomView:self.saveButtonController.button];
+        _saveToolbarItem = [[UIBarButtonItem alloc] initWithCustomView:self.saveButtonController.button];
     }
     return _saveToolbarItem;
 }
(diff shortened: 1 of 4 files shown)


Wikipedia/UI-V5/WMFArticleContainerViewController.m
```diff
@@ -291,17 +310,35 @@ - (UIBarButtonItem*)paddingToolbarItem {
...
+        _tableOfContentsToolbarItem.tintColor = [UIColor blackColor];
+        return _tableOfContentsToolbarItem;
+    }
+    return _tableOfContentsToolbarItem;
+}
+
 - (UIBarButtonItem*)saveToolbarItem {
-    return [[UIBarButtonItem alloc] initWithCustomView:self.saveButtonController.button];;
+    if (!_saveToolbarItem) {
+        return [[UIBarButtonItem alloc] initWithCustomView:self.saveButtonController.button];
```
coreyfloyd: do we have to assign this to the ivar?

bgerstle: no, the compiler does that for you... :confounded:

bgerstle: this is why we need `lazy var` in swift..

PR #198, opened 2015-10-26
Fix saved entry migration

bgerstle:
We were constantly attempting to re-migrate _all_ saved pages due to a bad schema check.  Pushed a (failing) test and then the fix.

### DIFF
--- a/MediaWikiKit/MediaWikiKit/MWKSavedPageEntry.m
+++ b/MediaWikiKit/MediaWikiKit/MWKSavedPageEntry.m
@@ -82,15 +82,17 @@ static NSString* const MWKSavedPageEntryDidMigrateImageDataKey = @"didMigrateIma
 }
 
 - (BOOL)isEqualToEntry:(MWKSavedPageEntry*)rhs {
-    return WMF_RHS_PROP_EQUAL(title, isEqualToTitle:);
+    return WMF_RHS_PROP_EQUAL(title, isEqualToTitle:)
+           && self.didMigrateImageData == rhs.didMigrateImageData;
 }
 
 - (NSUInteger)hash {
-    return self.title.hash;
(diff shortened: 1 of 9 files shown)


bgerstle:
@coreyfloyd @montehurd this is ready for review

MediaWikiKit/MediaWikiKit/MWKSavedPageEntry.m
```diff
@@ -82,15 +82,17 @@ - (BOOL)isEqual:(id)object {
 }
 
 - (BOOL)isEqualToEntry:(MWKSavedPageEntry*)rhs {
-    return WMF_RHS_PROP_EQUAL(title, isEqualToTitle:);
+    return WMF_RHS_PROP_EQUAL(title, isEqualToTitle:)
+           && self.didMigrateImageData == rhs.didMigrateImageData;
```
coreyfloyd: We covered this during review the other week… do we really want equality to fail for any other reason than the title being different? I thought we resolved this and agreed that we did not. 

Since you are changing it back, I am assuming you feel differently now? This feels a lot like we are repeating ourselves again during code review. 

This also appears to have little to do with the bug being addressed.

coreyfloyd:
:-1: 

Already discussed with @bgerstle he's going to revert the equality changes.

Basically the code base assumes that equality means just the title since this is the only property we care about. 

Changing this to include properties we don't care about may make it easier to write the test, but makes it harder to check equality in the code base itself. It also introduces the possibility of bugs since other code may depend on the previous behavior.

At a more best practices level, we should write equality methods not to be strict and include every property but instead to fit the domain of the problem. This is one reason why Apple does not automatically just implement the equality method for us - they expect the `isEqual` method to be domain specific.

bgerstle:
@coreyfloyd comments address, please re-review

PR #199, opened 2015-10-27
Improve tap area of save button in article toolbar

coreyfloyd:
Being a custom button was interfering with Apple's touch logic. Converted to a normal bar button item. Also updated the button controller to handle bar button items.

Additional pull request is coming for the save button on home screen feed items.

### DIFF
--- a/Wikipedia/UI-V5/WMFArticleContainerViewController.m
+++ b/Wikipedia/UI-V5/WMFArticleContainerViewController.m
@@ -117,8 +117,6 @@ NS_ASSUME_NONNULL_BEGIN
         self.discoveryMethod = discoveryMethod;
         [self observeArticleUpdates];
         self.hidesBottomBarWhenPushed = YES;
-        [self setupToolbar];
-        self.navigationItem.rightBarButtonItem = [self wmf_searchBarButtonItemWithDelegate:self];
     }
     return self;
 }
@@ -234,18 +232,6 @@ NS_ASSUME_NONNULL_BEGIN
     return _shareOptionsController;
 }
 
-- (WMFSaveButtonController*)saveButtonController {
-    if (!_saveButtonController) {
-        _saveButtonController = [[WMFSaveButtonController alloc] init];
-        UIButton* saveButton = [UIButton wmf_buttonType:WMFButtonTypeBookmark handler:nil];
-        [saveButton sizeToFit];
-        _saveButtonController.button        = saveButton;
-        _saveButtonController.savedPageList = self.savedPages;
-        _saveButtonController.title         = self.articleTitle;
-    }
-    return _saveButtonController;
-}
-
 #pragma mark - Notifications and Observations
 
 - (void)applicationWillResignActiveWithNotification:(NSNotification*)note {
@@ -279,6 +265,11 @@ NS_ASSUME_NONNULL_BEGIN
 }
 
 - (void)updateToolbarItemsIfNeeded {
+    
+    if (!self.saveButtonController) {
+        self.saveButtonController = [[WMFSaveButtonController alloc] initWithBarButtonItem:self.saveToolbarItem savedPageList:self.savedPages title:self.articleTitle];
+    }
+
     NSMutableArray<UIBarButtonItem*>* toolbarItems =
         [NSMutableArray arrayWithObjects:
          self.refreshToolbarItem, [self flexibleSpaceToolbarItem],
@@ -324,7 +315,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (UIBarButtonItem*)saveToolbarItem {
     if (!_saveToolbarItem) {
-        _saveToolbarItem = [[UIBarButtonItem alloc] initWithCustomView:self.saveButtonController.button];
+        _saveToolbarItem = [[UIBarButtonItem alloc] initWithImage:[UIImage imageNamed:@"save"] style:UIBarButtonItemStylePlain target:nil action:nil];
     }
     return _saveToolbarItem;
 }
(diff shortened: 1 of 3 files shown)


bgerstle:
good call, no need to use an embedded `UIButton` if we're using a plain ol' image

Wikipedia/UI-V5/WMFArticleContainerViewController.m
```diff
@@ -364,6 +355,9 @@ - (UIBarButtonItem*)shareToolbarItem {
 - (void)viewDidLoad {
     [super viewDidLoad];
 
+    [self setupToolbar];
```
bgerstle: :-1: unless I'm missing something, this still needs to happen in `setArticle:` in order to do the following when set either from data store in `viewDidLoad` or as a result of a fetch:

1. If `nil`, mark specific items requiring article data as disabled
2. Otherwise:
  1. Remove ToC if article is a main page
  2. Mark items requiring data as enabled—if we have data

Wikipedia/UI-V5/WMFArticleContainerViewController.m
```diff
@@ -279,6 +265,11 @@ - (void)setupToolbar {
 }
 
 - (void)updateToolbarItemsIfNeeded {
+    
+    if (!self.saveButtonController) {
```
bgerstle: when don't we keep this as a lazy property? if it's set manually, we can't declare it as `nonnull`, which makes reasoning about the `saveButtonController` availability and configuration much simpler, IMO

Wikipedia/UI-V5/WMFArticleContainerViewController.m
coreyfloyd: @bgerstle This still happens in set article, that was unaltered. I moved this from init to view did load.

bgerstle: ah sorry, misread the diff. still, it needed to be in `init` because that guarantees `toolbarItems` isn't empty, which allows the toolbar to be shown. unless you want to change how toolbar appearance/disappearance works, it needs to stay in `init, which is [what Apple recommends](https://developer.apple.com/library/prerelease/tvos/documentation/WindowsViews/Conceptual/ViewControllerCatalog/Chapters/NavigationControllers.html#//apple_ref/doc/uid/TP40011313-CH2-SW26).

bgerstle: another option is to manually call `setToolbarHidden:animated:` from within this class, and still rely on the current logic to hide it when the next VC comes in if `toolbarItems` is empty

Wikipedia/UI-V5/WMFArticleContainerViewController.m
coreyfloyd: @bgerstle because then you would have to call [self saveButtonController] just for the side effect on creating it. And that seemed worse.

bgerstle: i see what you mean. how about just creating it in `init` w/ the `articleTitle` and setting the save button on it in `viewDidLoad`? best of both worlds?

Wikipedia/UI-V5/WMFArticleContainerViewController.m
coreyfloyd: >  it needed to be in init because that guarantees toolbarItems isn't empty, which allows the toolbar to be shown.

As far as I know, toolbar visibility isn't determined by the toolbar items, it is determined by `setToolBarHidden`. Do you have documentation saying that the toolbar will ignore `toolBarHidden=YES` if `toolbarItems` are nil?

coreyfloyd: @bgerstle ^

coreyfloyd: @bgerstle FYI - From what I read on the link you posted, Apple does not really recommend setting tool bar items on init in that example, they happen to be walking the reader through some some sample code and only make mention of initialization in a figure:

> Listing 1-3 shows the code needed to specify the toolbar items programmatically. You would implement the method in your view controller and call it at initialization time.

And then go on to say that you are able to change toolbaritems after init:

> In addition to setting toolbar items during initialization, a view controller can also change its existing set of toolbar items dynamically using the setToolbarItems:animated: method…

Which we already know. This points to there being no technical reason to set toolbar items in init. If you know of one let me know. 

However, setting toolbar items in `viewDidLoad` is insanely common - I would venture to guess more common than setting in `init`. If that isn't allowed, then we and a lot of developers have a lot of code to go through and change.

Wikipedia/UI-V5/WMFArticleContainerViewController.m
coreyfloyd: bikeshedding? I'm working on other tickets and pressing to get the alpha out, so if it actually doesn't have an impact…

Wikipedia/UI-V5/WMFArticleContainerViewController.m
bgerstle: @coreyfloyd the `WMFAppViewController` [checks `toolbarItems` and hides/unhides it manually](https://github.com/wikimedia/wikipedia-ios/blob/master/Wikipedia/UI-V5/WMFAppViewController.m#L444).  if you want to get into refactoring that, it's fine, but I don't think it's necessary for this ticket if you keep initially toolbar setup in `init`, which is unlikely to have a negative impact IMO.

coreyfloyd: @bgerstle i haven't see a bug indicating the need for doing this in init - have you?

Wikipedia/UI-V5/WMFArticleContainerViewController.m
bgerstle: sure, i would argue there is an impact but will concede that the main thing we're missing here is a (UI) test to verify save buttons are hooked up properly and are functional.

Wikipedia/UI-V5/WMFArticleContainerViewController.m
bgerstle: @coreyfloyd i'm unable to reproduce either, since it now appears that `viewDidLoad` is being called _before_ the `WMFAppViewController` check.  however, it was how I [fixed the regression causing the toolbar to not be displayed](https://github.com/wikimedia/wikipedia-ios/pull/178/files#diff-6797d570310857eec6a38c955002bb24R107). so :+1: i guess?

PR #200, opened 2015-10-27
Add title to save button 

coreyfloyd:
(This addresses the tap target problem by adding the title)

### DIFF
--- a/Wikipedia/UI-V5/UIButton+WMFButton.m
+++ b/Wikipedia/UI-V5/UIButton+WMFButton.m
@@ -103,6 +103,8 @@ NS_ASSUME_NONNULL_BEGIN
         case WMFButtonTypeBookmarkMini:
             [self setImage:[UIImage imageNamed:@"save-mini"] forState:UIControlStateNormal];
             [self setImage:[UIImage imageNamed:@"save-filled-mini"] forState:UIControlStateSelected];
+            [self setTitle:@"Save for later" forState:UIControlStateNormal];
+            [self setTitle:@"Saved for later" forState:UIControlStateSelected];
             break;
         case WMFButtonTypeCloseMini:
             [self setImage:[UIImage imageNamed:@"close-mini"] forState:UIControlStateNormal];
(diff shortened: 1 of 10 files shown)


Wikipedia/UI-V5/UIButton+WMFButton.m
```diff
@@ -103,6 +103,8 @@ - (void)wmf_setButtonType:(WMFButtonType)type {
         case WMFButtonTypeBookmarkMini:
             [self setImage:[UIImage imageNamed:@"save-mini"] forState:UIControlStateNormal];
             [self setImage:[UIImage imageNamed:@"save-filled-mini"] forState:UIControlStateSelected];
+            [self setTitle:@"Save for later" forState:UIControlStateNormal];
+            [self setTitle:@"Saved for later" forState:UIControlStateSelected];
```
montehurd: Prob add these to the en localizations and the qqq so we start getting translations. Want to do this now or follow-on?

Wikipedia/UI-V5/WMFTitleInsetRespectingButton.m
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ *  UIButton does not take into acocunt its insets when calculating intrinisic content size
+ *  http://stackoverflow.com/a/17806333/48311
+ *
+ */
+- (CGSize)intrinsicContentSize {
+    CGSize s = [super intrinsicContentSize];
+
+    return CGSizeMake(s.width + self.titleEdgeInsets.left + self.titleEdgeInsets.right,
```
montehurd: I got a strange null frame crash once... maybe assert-check for that here?

Wikipedia/UI-V5/UIButton+WMFButton.m
coreyfloyd: @montehurd follow on - there are so many strings we need to add…

Wikipedia/UI-V5/WMFTitleInsetRespectingButton.m
coreyfloyd: Crash seems unrelated.

as far as I know, `[super intrinsicConstentSize]` always returns a value as does `self.titleEdgeInsets` Do you know a situation where those would return garbage?

PR #208, opened 2015-10-28
Add interactivity to the ToC

coreyfloyd:
Using the interactive transition APIs to implement.

Most of the code for this is in the animator (Sadly I was not able to factor this into separate objects)

Also removed some dead code…


bgerstle:
@coreyfloyd :+1:  code looks good, just want to do a quick smoke test

bgerstle:
Tested on my iPhone :+1:. Great overall, but there are some things I'd like to polish later:

* Animation speed more directly proportional to swipe velocity
* Add Tweak for "swipe to open" threshold

PR #209, opened 2015-10-28
Fix logic error preventing Toc scrolling on selection

coreyfloyd:


bgerstle:
:+1: will merge when travis ok's. @montehurd review before then, if you want

PR #210, opened 2015-10-28
Fix deletion

coreyfloyd:
Needed to increase the threshold for detecting a deletion pan in contrast to a scrolling pan
Also fading out the deleted item as it moves off screen so it is more obvious what is happening

### DIFF
--- a/Wikipedia/UI-V5/WMFEditingCollectionViewLayout.m
+++ b/Wikipedia/UI-V5/WMFEditingCollectionViewLayout.m
@@ -307,7 +307,7 @@ typedef NS_ENUM (NSInteger, TGLStackedViewControllerScrollDirection) {
 - (BOOL)gestureRecognizerShouldBegin:(UIGestureRecognizer*)gestureRecognizer {
     if ([gestureRecognizer isEqual:self.deletePanGesture]) {
         CGPoint velocity = [(UIPanGestureRecognizer*)gestureRecognizer velocityInView:self.collectionView];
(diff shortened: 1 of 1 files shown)


Wikipedia/UI-V5/WMFEditingCollectionViewLayout.m
```diff
@@ -371,6 +371,11 @@ - (IBAction)panWithGesture:(UIPanGestureRecognizer*)pan {
                 CGPoint currentCenter = self.movingCellCenter;
                 currentCenter.x               += translation.x;
                 self.movingSnapshotView.center = currentCenter;
+                CGFloat alpha = CGRectGetMaxX(self.movingSnapshotView.frame) / self.collectionView.bounds.size.width;
+                if(alpha > 1){
+                    alpha = 1 - (alpha - 1);
```
bgerstle: can you add some comments explaining what is going on here? it looks like you're "capping" the alpha at 1, but it's not clear why you're subtracting the "remainder" above 1 from 1.

bgerstle:
would like some comments on the alpha calculation, but otherwise :+1:

coreyfloyd:
@bgerstle this is working well for me with this new value.

bgerstle:
@coreyfloyd :+1:

coreyfloyd:
@bgerstle updated to a the more performant implementation for the indexPathForTitle. 

Obviously we need to dry all this up later.

PR #215, opened 2015-10-29
Fix order of search results

bgerstle:
Phab: [T117005](https://phabricator.wikimedia.org/T117005)

---

Deserializing & sorting by `index` property, which previous fetcher was doing manually.


WikipediaUnitTests/WMFSearchResultsSerializationTests.m
```diff
@@ -0,0 +1,48 @@
...
+#import <OCHamcrest/OCHamcrest.h>
+
+@interface WMFSearchResultsSerializationTests : XCTestCase
+
+@end
+
+@implementation WMFSearchResultsSerializationTests
+
+- (void)testSerializesPrefixResultsInOrderOfIndex {
+    NSString* fixtureName = @"BarackSearch";
```
montehurd: Awesome!

montehurd:
Opps we have a bug - we lost the "Redirected from:" text:

*Redirected results should have redirect info:*
screen_shot_2015-10-28_at_5 13 53_pm

*Presently, with this PR, results don't have the redirect info:*
screen shot 2015-10-29 at 12 04 20 pm

bgerstle:
Was that already a regression or caused by this PR specifically? F the
former, can we merge this anyway and fix the other as a follow on? Unless
you want to amend this to also fix "Redirected from:" (which IMO should be
a computer property)

montehurd:
Looks like it was cause by corey's pr. Just confirmed with Josh that he wants the redirected text - is in the spec :) So do as follow-on?

PR #216, opened 2015-10-30
Update recent search styling

montehurd:
**Mock divergence note:**
Punted on removing the separator beneath the search bar for now based on conversation with Josh. Nirzar wants a shadow-separator to appear on when you scroll search results, and for something similar to happen with recent search terms when you scroll them (shadow-separator appearing then beneath the *RECENTLY SEARCHED* header when recent search terms are scrolled), but those are both going to take a bit of hacking to do, which can happen as follow-ons, per Josh, if we have time later. Added notes to the phab ticket tagging both Josh and Nirzar about these follow-ons.

**Testing note:** 
If you are testing on the simulator, set its scale to 100% or sometimes the separator lines won't appear.

**Screenshot:** 
I searched for same terms as seen in the mock for easy comparison.
screen shot 2015-10-30 at 6 08 51 pm


Wikipedia/Categories/UIColor+WMFHexColor.m
```diff
@@ -14,12 +14,16 @@ + (UIColor*)wmf_colorWithHex:(NSInteger)hex
...
-    return [NSString stringWithFormat:@"#%02lX%02lX%02lX",
-            lroundf(components[0] * 255),
-            lroundf(components[1] * 255),
-            lroundf(components[2] * 255)];
+    CGFloat red;
+    CGFloat green;
+    CGFloat blue;
+    CGFloat alpha;
+    [self getRed:&red green:&green blue:&blue alpha:&alpha];
+    return [NSString stringWithFormat:@"%02x%02x%02x%02x",
```
montehurd: Wasn't working with non RGB color space UIColors. This method does.

Wikipedia/View Controllers/RecentSearches/RecentSearchesViewController.m
```diff
@@ -39,15 +34,16 @@ - (void)viewDidLoad {
...
 }
 
 - (void)setupTable {
-    self.table.separatorStyle = UITableViewCellSeparatorStyleNone;
     [self.table registerNib:[UINib nibWithNibName:@"RecentSearchCell" bundle:nil] forCellReuseIdentifier:@"RecentSearchCell"];
+
+    self.table.estimatedRowHeight = 52.f;
+    self.table.rowHeight          = UITableViewAutomaticDimension;
+
+    [self.table setBackgroundView:nil];
```
montehurd: Used grouped table layout to get, for free, separators above the first cell and below the last cell, but grouped layout adds a background which causes the translucency effect to break. Nil'ing out the background view fixes it.

Wikipedia/Categories/NSString+Extras.m
```diff
@@ -103,6 +103,15 @@ - (NSString*)wmf_stringByCapitalizingFirstCharacter {
     return self;
 }
 
+- (NSString*)wmf_stringByCapitalizingAllCharactersIfDeviceLocaleIsEnglish {
+    // Reminder: TWN has in the past rejected all-caps strings because there are complications
+    // with translation/meaning of all-caps in other languages. The recommendation
+    // was to submit strings to TWN with non-all-caps, and at display time force the string
+    // to all caps, but *only* for English.
+    NSString* deviceLanguageCode = [[NSLocale preferredLanguages] objectAtIndex:0];
+    return ([deviceLanguageCode isEqualToString:@"en"] || [[deviceLanguageCode substringToIndex:3] isEqualToString:@"en-"]) ? [self uppercaseString] : self;
```
coreyfloyd: Does  `-[NSString uppercaseStringWithLocale:]` take into account the issues with uppercasing?

coreyfloyd:
@montehurd looks good. Left one comment about if the built in upper case locale function can help out at all. If not fine, just thought it may solve the issue. Otherwise good to go.

Wikipedia/View Controllers/RecentSearches/RecentSearchesViewController.m
bgerstle: @montehurd thanks for the note, but it's better to put documentation or explanations with the source so it doesn't get lost.  in particular, this kind of comment (which is a workaround for built-in framework behavior) has been given `HAX:` prefix to signal to readers that this is code is intentionally hacky, so as to work around the described behavior.

montehurd: @bgerstle Makes sense. Will add that.

Wikipedia/Categories/NSString+Extras.m
montehurd: @coreyfloyd Good catch. From the limited info in the docs is seems it may... Let's try it!

Wikipedia/UI-V5/WMFSearchViewController.storyboard
```diff
@@ -218,28 +218,49 @@
...
+                                    </label>
+                                    <view contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="7SS-UN-pgW" userLabel="Trash Button Container">
+                                        <rect key="frame" x="546" y="22" width="54" height="34"/>
+                                        <animations/>
+                                        <color key="backgroundColor" white="0.0" alpha="0.0" colorSpace="calibratedWhite"/>
+                                        <constraints>
+                                            <constraint firstAttribute="width" constant="54" id="1mE-o0-SnE"/>
+                                        </constraints>
+                                        <userDefinedRuntimeAttributes>
+                                            <userDefinedRuntimeAttribute type="boolean" keyPath="keyPath" value="YES"/>
```
bgerstle: @montehurd please remove these, causing runtime exceptions (but not crashes, surprisingly):

```
2015-11-02 15:02:18.004 Wikipedia Debug[7709:2441108] Failed to set (keyPath) user defined inspected property on (UIView): [<UIView 0x13e7227f0> setValue:forUndefinedKey:]: this class is not key value coding-compliant for the key keyPath.
```

Wikipedia/Categories/NSString+Extras.m
```diff
@@ -103,6 +103,14 @@ - (NSString*)wmf_stringByCapitalizingFirstCharacter {
     return self;
 }
 
+- (NSString*)wmf_stringByCapitalizingAllCharactersIfDeviceLocaleIsEnglish {
```
coreyfloyd: Since you do this for all languages now, `…IfDeviceLocaleIsEnglish` is a bit of a lie now.
Instead of renaming, at this point, If it is just a single line of code, you don't really wrap it in a method anymore.

montehurd: @coreyfloyd oh god good catch.

Wikipedia/UI-V5/WMFSearchViewController.storyboard
montehurd: Fixed!

bgerstle:
:+1: :basketball: in Travis's court now

PR #219, opened 2015-10-30
Fix bug where ToC section was not highlighted if using the drag gesture

coreyfloyd:
converted the code to highlight the action to a delegate method.

### DIFF
--- a/Wikipedia/UI-V5/WMFArticleContainerViewController+TOC.swift
+++ b/Wikipedia/UI-V5/WMFArticleContainerViewController+TOC.swift
@@ -3,6 +3,20 @@ import Foundation
 import BlocksKit
 
 extension WMFArticleContainerViewController : WMFTableOfContentsViewControllerDelegate {
+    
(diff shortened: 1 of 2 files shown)


Wikipedia/UI-V5/WMFArticleContainerViewController+TOC.swift
```diff
@@ -3,6 +3,20 @@ import Foundation
 import BlocksKit
 
 extension WMFArticleContainerViewController : WMFTableOfContentsViewControllerDelegate {
+    
+    public func tableOfContentsControllerWillDisplay(controller: WMFTableOfContentsViewController){
```
bgerstle: :+1: but wondering if we should instead ask the delegate what the current item is, and let the ToC implement "selection on appear" instead of asking its delegate to do it. this way, delegates don't need to repeat implementations of "selection on appear."  nbd now since there's only 1 delegate

bgerstle:
code looks good but can't test atm. @montehurd :basketball: is in our court

montehurd:
Tested and seems good!

PR #221, opened 2015-11-02
Update missing localization strings. Part 1.

montehurd:
Part one of localization string updates! Part 2 coming Wednesday.

Feel free to merge this in the mean time. If not, that's ok too.

### DIFF
--- a/Wikipedia/UI-V5/UIButton+WMFButton.m
+++ b/Wikipedia/UI-V5/UIButton+WMFButton.m
@@ -100,8 +100,8 @@ NS_ASSUME_NONNULL_BEGIN
         case WMFButtonTypeBookmarkMini:
             [self setImage:[UIImage imageNamed:@"save-mini"] forState:UIControlStateNormal];
             [self setImage:[UIImage imageNamed:@"save-filled-mini"] forState:UIControlStateSelected];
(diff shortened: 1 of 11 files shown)


Wikipedia/UI-V5/UIButton+WMFButton.m
```diff
@@ -100,8 +100,8 @@ - (void)wmf_setButtonType:(WMFButtonType)type {
         case WMFButtonTypeBookmarkMini:
             [self setImage:[UIImage imageNamed:@"save-mini"] forState:UIControlStateNormal];
             [self setImage:[UIImage imageNamed:@"save-filled-mini"] forState:UIControlStateSelected];
-            [self setTitle:@"Save for later" forState:UIControlStateNormal];
-            [self setTitle:@"Saved for later" forState:UIControlStateSelected];
+            [self setTitle:MWLocalizedString(@"button-save-for-later", nil) forState:UIControlStateNormal];
+            [self setTitle:MWLocalizedString(@"button-saved-for-later", nil) forState:UIControlStateSelected];
```
bgerstle: this is ok, but in the future, can we make these kinds of keys more visually distinct? 1-letter differences can be very hard to spot

bgerstle: same w/ image names, etc.

PR #225, opened 2015-11-04
Update recent style

coreyfloyd:
https://phabricator.wikimedia.org/T114967

Implemented our "first table view" as discussed on this ticket. Auto layout in this case is much simpler. I would estimate it would take half a day to a full day to convert our remaining collection code over to table views.

screen shot 2015-11-03 at 7 00 45 pm


bgerstle:
@coreyfloyd any idea how some of those titles got wrapping `i` tags?

bgerstle:
looks like the article list cell visual tests are failing, might need to record new reference images

bgerstle:
:+1: lgtm!  as we discussed on hangout, it's looking like we'll probably use this as the foundation to remove all the collection view workarounds which make them look like grouped tables—better to just use grouped tables.  historically we wanted collection views so we could have custom transitions, but current app behavior is more closely matching grouped tables in that:

- Cells are laid out vertically in a single column 
- Headers & footers scroll w/ the content (no floating on top)
- Table views handle auto-sizing content more easily than collection views (see all `preferredLayoutAttributes` code)

Will file a tickets for subsequent refactors later.

bgerstle:
@coreyfloyd just testing on iOS 8 when I noticed that cells don't have consistent heights:

image

is that intentional?

bgerstle:
@coreyfloyd also, I can see the history list update when I:

- Select "Recent" tab
- Tap Search
- Search for something and select a result

Search dismisses and then I see the history list update before the article view is pushed. Maybe an artifact of current search flow which is probably going to change anyway?

coreyfloyd:
@bgerstle did you update to the latest? i don't see that - and from the code, it looks like it would be impossible for that to happen.

bgerstle:
@coreyfloyd sorry, the search thing i described isn't happening after i pulled the latest—mea culpa. 

everything else seems OK, do you want to bring up the cell heights in design review and address in a follow-on if needed?

coreyfloyd:
@bgerstle missed that…

So its not actually a height constraint its that the wikidata description was limited to a single line. I made that change in the late commit.

bgerstle:
:+1: cool. @montehurd can you review and/or merge?

PR #226, opened 2015-11-04
Fix initial search results not having thumbnails

bgerstle:
Phab: [T117012](https://phabricator.wikimedia.org/T117012)

---

## Diagnosis
The issue was caused by a couple of things.  On our end, we were setting the search results list data source twice, _even if we didn't fetch more full-text search results_.  This caused the collection view to render the same cells twice, implicitly cancelling and re-fetching the cells' images.

I think SDWebImage is also partially to blame, as cancelling & refetching _should_ be fine.  You can see logs for the second fetch get emitted, but no callback ever occurs for the second fetch.  I think this is caused by the internal callback queueing mechanism in `SDWebImageManager` which is supposed to prevent redundant requests, but seems to also accidentally cancel fetches that are executed too closely together.

## Cure
Anyway, I solved the issue by only using one `WMFSearchResults` object for each query.  In case we fetch more full-text results, we append them to the first set of results.  I'm not crazy about the implications for mutability, but it's either this or calculating collection view updates from a diff of two immutable objects.  

As an aside, this "merge" API exists on _all `MTLModel`subclasses_, which we'll need to be careful with... Great power and great responsibility and all that...

### DIFF
--- a/MediaWikiKit/MediaWikiKit/UIImageView+WMFImageFetchingInternal.m
+++ b/MediaWikiKit/MediaWikiKit/UIImageView+WMFImageFetchingInternal.m
@@ -184,6 +184,8 @@ static const char* const WMFImageControllerAssociationKey = "WMFImageController"
 
 - (void)wmf_cancelImageDownload {
     [self.wmf_imageController cancelFetchForURL:[self wmf_imageURLToFetch]];
+    self.wmf_imageURL      = nil;
+    self.wmf_imageMetadata = nil;
 }
 
 @end
--- a/Wikipedia.xcodeproj/project.pbxproj
+++ b/Wikipedia.xcodeproj/project.pbxproj
@@ -301,6 +301,7 @@
 		BC23759E1AB8928600B0BAA8 /* WMFDateFormatterTests.m in Sources */ = {isa = PBXBuildFile; fileRef = BC23759D1AB8928600B0BAA8 /* WMFDateFormatterTests.m */; };
 		BC2375C11ABB14CC00B0BAA8 /* WMFArticleImageInjectionTests.m in Sources */ = {isa = PBXBuildFile; fileRef = BC2375C01ABB14CC00B0BAA8 /* WMFArticleImageInjectionTests.m */; };
 		BC2CBB8E1AA10F400079A313 /* UIView+WMFFrameUtils.m in Sources */ = {isa = PBXBuildFile; fileRef = BC2CBB8D1AA10F400079A313 /* UIView+WMFFrameUtils.m */; };
+		BC2F420E1BEB9BBC0033E185 /* WMFImageControllerTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = BC12F2451B7A8EEB0042DB3C /* WMFImageControllerTests.swift */; };
 		BC3047C21B45E65400D7DF1A /* SDWebImageManager+WMFCacheRemoval.m in Sources */ = {isa = PBXBuildFile; fileRef = BC3047C11B45E65400D7DF1A /* SDWebImageManager+WMFCacheRemoval.m */; };
 		BC3059B41B69416E00AA5703 /* MWKImageFaceDetectionTests.m in Sources */ = {isa = PBXBuildFile; fileRef = BC3059B31B69416E00AA5703 /* MWKImageFaceDetectionTests.m */; };
 		BC3059B91B6948C800AA5703 /* XCTestCase+PromiseKit.m in Sources */ = {isa = PBXBuildFile; fileRef = BC3059B81B6948C800AA5703 /* XCTestCase+PromiseKit.m */; };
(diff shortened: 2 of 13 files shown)


montehurd:
Confirmed that this fixes the issue :+1:

coreyfloyd:
@bgerstle SDWebImage is pretty rock solid and widely used, so it seemed weird that something as basic as canceling would be buggy. 

Taking a look at our code the thing that stuck out to me the most is that we bounce our cancellations through an additional cancellation queue before sending the actually cancelling the request on SDWebImage.

Since calling cancel must travel through 2 queues to be precessed, but fetching only needs to travel through 1, it makes sense that a second fetch can actually start before the cancellation. It also makes sense that SDWebImage would not exhibit this behavior on its own because cancellation using SDWebImage APIs don't have this additional delay.

I ran the app looking at this and put 2 break points in

On inside the dispatchAsync:

```
public func cancelFetchForURL(url: NSURL?) {
        if let url = url {
            weak var wself = self;
            dispatch_async(self.cancellingQueue) {
                let sself = wself
                if let cancellable = sself?.cancellables.objectForKey(url.absoluteString) as? Cancellable {
                    sself?.cancellables.removeObjectForKey(url.absoluteString)
                    cancellable.cancel()
                }
            }
        }
    }
```

And one inside the image fetch:

```
func promisedImageWithURL(URL: NSURL, options: SDWebImageOptions) -> (Cancellable, Promise<WMFImageDownload>) {
        let (promise, fulfill, reject) = Promise<WMFImageDownload>.pendingPromise()

        let promiseCompatibleOptions: SDWebImageOptions = [options, SDWebImageOptions.ReportCancellationAsError]

        let webImageOperation = self.downloadImageWithURL(URL, options: promiseCompatibleOptions, progress: nil)
        { img, err, cacheType, finished, imageURL in
                if finished && err == nil {
                    fulfill(WMFImageDownload(url: URL, image: img, origin: asImageOrigin(cacheType).rawValue))
                } else {
                    reject(err)
                }
        }

        return (SDWebImageOperationWrapper(operation: webImageOperation), promise)
    }

```

I definitely noticed requests being cancelled after being requested. Further investigation would be needed to confirm. But it looks like if we just process cancellations immediately instead of bouncing them through the queue, cancelation would work as expected. Or as an alternative, make all fetch request pass through the same queue so its impossible for them to "jump the line".

bgerstle:
@coreyfloyd I would have to write a test to double check, but fetching and cancelling requests via `WMFImageController` should be FIFO operations since both are funneled through the same queue (see [`addCancellableForURL`](https://github.com/wikimedia/wikipedia-ios/blob/feature/fix-image-placeholder/T117012/Wikipedia/Images/WMFImageController.swift#L324)).

`WMFImageController` keeps track of `Cancellable` objects so as to allow callers to cancel requests by URL.  Another option would be to return the cancellable as part of the API, and let callers cancel directly.  This would remove the need for any queueing, but like I said, I don't think that's the issue.

coreyfloyd:
> fetching and cancelling requests via WMFImageController should be FIFO operations since both are funneled through the same queue (see addCancellableForURL).

```
 public func fetchImageWithURL(
                    url: NSURL?,
                    options: SDWebImageOptions = SDWebImageOptions()) -> Promise<WMFImageDownload> {
        // HAX: make sure all image requests have a scheme (MW api sometimes omits one)
        return checkForValidURL(url) { url in
            let (cancellable, promise) =
            imageManager.promisedImageWithURL(url.wmf_urlByPrependingSchemeIfSchemeless(), options: options)
            addCancellableForURL(cancellable, url: url)
            return applyDebugTransformIfEnabled(promise)
        }
    }
```

@bgerstle `addCancellable` is called after `imageManager.promisedImageWithURL`, so it appears the fetch itself does not go though the cancellable queue.

coreyfloyd:
Also note that `addCancellable` returns immediately - the work inside of `addCancellable` is performed asynchronously, so even if that line was first, it wouldn't make a difference. you would need to do something like pass a completion block - like `addCancellableWithCompletion:` - to ensure that the fetch did not start until after the cancellable was added.

bgerstle:
i think i see what you're saying: the second request would be queued in SDWebImage before the first was cancelled.

> you would need to do something like pass a completion block - like addCancellableWithCompletion:

not possible, since we only acquire the cancellable after the fetch is started.

I'll add a test for this to verify tomorrow.  My current thinking is to either make cancellation synchronous (thereby guaranteeing that any outstanding requests are cancelled before your next call to `fetch...`) or get rid of the queues by returning the cancellable to the caller

coreyfloyd:
> i think i see what you're saying: the second request would be queued in SDWebImage before the first was cancelled.

Yeah - that's what it looks like.

coreyfloyd:
@bgerstle I know you were heading out, but just to comment without holding you up:

>4:55 PM <coreyfloyd> bgerstle_afk: quick question - did we back out the other changes on that PR- do we need to still work around with the search stuff?
4:57 PM <bgerstle_afk> coreyfloyd: didn't back them out, no
4:57 PM <bgerstle_afk> figured it's still a weird behavior that we're better off without

I am assuming the weird behavior you are referring to is the double reload of the search? Or was it something else?

The big question I had was: Does making search results mutable do anything for us now? If it doesn't, do we want to back that out? Otherwise I am good.

If you think that is still needed or doesn't matter, I will merge - just let me know.

bgerstle:
> Does making search results mutable do anything for us now? If it doesn't,
do we want to back that out?

The main thing it gets us is preventing the extra reload of cells that are
already onscreen. The only time this would happen now is when we get full
text results back.

That said, I think having it mutable now also sets us up nicely for
continuing/paginating the results. Then again, so would core data 😉

No strong feelings, but if you think backing out is warranted, I can back
it out later or you can amend. Can also be done as a separate PR.

coreyfloyd:
no I'm good - just wanted to check

montehurd:
Something went funky here with how we're appending full text search results. I added notes to the phab ticket with screenshots and stuck it back in "ready for dev".

PR #228, opened 2015-11-04
Second round of localization updates.

montehurd:
Various updates to localized strings related bits.

Second round of updates for: https://phabricator.wikimedia.org/T105532

### DIFF
--- a/Wikipedia/Categories/NSLocale+WMFExtras.swift
+++ b/Wikipedia/Categories/NSLocale+WMFExtras.swift
@@ -1,0 +1,13 @@
+//  Created by Monte Hurd on 10/4/15.
+//  Copyright (c) 2015 Wikimedia Foundation. Provided under MIT-style license; please copy and modify!
+
+import Foundation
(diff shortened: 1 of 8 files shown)


Wikipedia/Categories/NSLocale+WMFExtras.swift
```diff
@@ -0,0 +1,13 @@
...
+//  Copyright (c) 2015 Wikimedia Foundation. Provided under MIT-style license; please copy and modify!
+
+import Foundation
+
+extension NSLocale {
+    class func wmf_isCurrentLocaleEnglish() -> Bool {
+        guard let langCode = NSLocale.currentLocale().objectForKey(NSLocaleLanguageCode) as? String else {
+            return false
+        }
+        return (langCode == "en" || langCode.hasPrefix("en-")) ? true : false;
```
bgerstle: FYI ternary statement is unnecessary here: the result of the expression is already boolean

PR #229, opened 2015-11-04
Capitalize 1st letter of wikidata descrip in continue reading card.

montehurd:


montehurd:
We're capitalizing the first letter every else the descrip is shown - we may wan't to DRY this up system-wide, but did quick fix in the mean time.

PR #231, opened 2015-11-05
Fix recent previewing

bgerstle:
We were using `UICollectionView` index path stuff instead of `UITableView`.  Yay weak typing!  See screenshot below for fix (note recent tab is selected in blurry background):

screen shot 2015-11-05 at 2 26 45 pm

### DIFF
--- a/Wikipedia/UI-V5/WMFArticleListTableViewController.m
+++ b/Wikipedia/UI-V5/WMFArticleListTableViewController.m
@@ -216,12 +216,12 @@
 
 - (nullable UIViewController*)previewingContext:(id<UIViewControllerPreviewing>)previewingContext
                       viewControllerForLocation:(CGPoint)location {
-    NSIndexPath* previewIndexPath = [(UITableView*)previewingContext.sourceView indexPathForRowAtPoint:location];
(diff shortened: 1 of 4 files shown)


Wikipedia/UI-V5/WMFArticleListTableViewController.m
```diff
@@ -216,12 +216,12 @@ - (void)didCommitToPreviewedArticleViewController:(WMFArticleContainerViewContro
 
 - (nullable UIViewController*)previewingContext:(id<UIViewControllerPreviewing>)previewingContext
                       viewControllerForLocation:(CGPoint)location {
-    NSIndexPath* previewIndexPath = [(UICollectionView*)previewingContext.sourceView indexPathForItemAtPoint:location];
+    NSIndexPath* previewIndexPath = [(UITableView*)previewingContext.sourceView indexPathForRowAtPoint:location];
```
coreyfloyd: why don't we just call `[self indexPathForRowAtPoint:location]`?

coreyfloyd: same for the collectionView code.

PR #232, opened 2015-11-05
Fix lead image placeholder

bgerstle:
Somehow forgot to use the same placeholder stuff in header image gallery.  I don't think there's much more DRY-ing that can be done for this sort of thing.

image


coreyfloyd:
:+1: whenever travis finishes…

PR #235, opened 2015-11-09
Add today's featured article to feed (on EN Wiki)

bgerstle:
Phab: [T105424](https://phabricator.wikimedia.org/T105424)

---

Adds a new fetcher which gets the content of the [Template:TFA_title page](https://en.wikipedia.org/wiki/Template:TFA_title), then runs an API query to get page extract, image, etc.

"Main Page" home section on EN:
image

on non-EN (FR in this case):
image


Wikipedia/Categories/AFHTTPRequestOperationManager+WMFDesktopRetry.h
```diff
@@ -38,4 +42,19 @@
...
+ *  Send a @c GET request to the given site, falling back to desktop URL if the mobile URL fails.
+ *
+ *  @param site The site to send the request to, using its API endpoints. First mobile, then desktop.
+ *
+ *  @return A promise which will be resolved with the successful response of the request, or rejected with any errors
+ *          that occur.
+ *
+ *  @see wmf_GETWithMobileURLString:desktopURLString:parameters:retry:success:failure:
+ */
+- (AnyPromise*)wmf_GETWithSite:(MWKSite*)site
```
coreyfloyd: @bgerstle If we pass out an outOperation, we also need to provide a retry block as in the method above. The reason for passing the retry block is because if a retry occurs, then the operation you received initially cannot be used for cancellation.

The other option is not passing back an outOperation if cancellation is not needed (making this more of a convenience method).

Wikipedia/Networking/Fetchers/WMFENFeaturedTitleFetcher.h
```diff
@@ -0,0 +1,21 @@
...
+//  Copyright © 2015 Wikimedia Foundation. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+@class AnyPromise;
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface WMFENFeaturedTitleFetcher : NSObject
```
coreyfloyd: `English` - spell it out, especially if its confusingly jammed up against the prefix.

Wikipedia/Networking/Fetchers/WMFENFeaturedTitleFetcher.h
```diff
@@ -0,0 +1,21 @@
...
+
+- (AnyPromise*)featuredArticlePreviewForDate:(nullable NSDate*)date;
```
coreyfloyd: `fetch` - for conformity with our other fetch methods

Wikipedia/Networking/Fetchers/WMFENFeaturedTitleFetcher.m
```diff
@@ -0,0 +1,175 @@
...
+@implementation WMFENFeaturedTitleRequestSerializer
+
++ (NSDateFormatter*)featuredArticleDateFormatter {
+    static NSDateFormatter* feedItemDateFormatter;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        feedItemDateFormatter = [[NSDateFormatter alloc] init];
+        feedItemDateFormatter.dateFormat = @"MMMM d, YYYY";
+        // feed format uses US dates
+        feedItemDateFormatter.locale = [NSLocale localeWithLocaleIdentifier:@"en-US"];
```
coreyfloyd: @bgerstle I don't think `en-US` is a locale? This is probably evaluating to nil. `en_US` is a locale. Maybe it accepts the hyphen format too? Not sure, but would check.


If it is evaluating to nil, then do we even need to set the locale? 

Also any reason not to use the currentLocal? Even if they are getting content in english, chrome is usually still in the users device language.

Wikipedia/UI-V5/WMFMainPageSectionController.m
```diff
@@ -88,24 +128,59 @@ - (void)configureCell:(UICollectionViewCell*)cell
...
+- (void)fetchData {
+    if (self.dataPromise) {
+        DDLogInfo(@"Fetch is already pending, skipping redundant call.");
         return;
     }
+
     @weakify(self);
-    [self.fetcher fetchSiteInfoForSite:self.site]
-    .then(^(id result){
+    self.dataPromise = ([self fetchFeaturedTitleIfAvailable] ? : [self fetchSiteInfo]).catch(^(NSError* error){
```
coreyfloyd: where is the `then` block?

Wikipedia/UI-V5/WMFMainPageSectionController.m
```diff
@@ -88,24 +128,59 @@ - (void)configureCell:(UICollectionViewCell*)cell
...
+    return [self.siteInfoFetcher fetchSiteInfoForSite:self.site].then(^(MWKSiteInfo* siteInfo){
+        @strongify(self);
+        self.dataPromise = nil;
+
+        self.siteInfo = siteInfo;
+        [self.delegate controller:self didSetItems:self.items];
+    });
+}
+
+- (nullable AnyPromise*)fetchFeaturedTitleIfAvailable {
```
coreyfloyd: Oh here… can we just put this in line so its more readable? I'm not really sure whats going on to be honest. The ternary slipped in there really complicates this.

The whole point of the promise API is to make a readable chain of asynchronous work  "do, then, catch, then, etc…" splitting up this logic in different places makes the code difficult to read and reason about. 

if you really feel the need to wrap logic in separate methods, then wrap up the "internals" so it still reads well:

```
[self doThingOne].then(^(id obj){
    return [self doThingTwo];
}).then(^(id obj)){
    return [self doThingThree];
}).catch{
    [self doSomethingWithError];
});
```

Wikipedia/mw-utils/NSString+WMFPageUtilities.h
```diff
@@ -49,4 +49,11 @@ extern NSString* const WMFCitationFragmentSubstring;
  */
 - (NSString*)wmf_normalizedPageTitle;
 
+/**
+ *  Process a normalized title into one which can be used as an API request parameter.
+ *
+ *  @return The receiver, with spaces replaced by underscores and added percent escapes.
+ */
+- (NSString*)wmf_denormalizedPageTitle;
```
coreyfloyd: Normalized seems like the wrong thing here - aren't these really `URLEscapedTitle` and  `URLUnescapedTitle`?

bgerstle: so, the idea was to try to sync our code to the MW domain concept of a "normalized title," as seen when you send a request for a title that's not normalized.  for example, [sending a `mobileview` request for "Barack_Obama"](https://en.wikipedia.org/w/api.php?action=mobileview&page=Barack_Obama) has `"normalizedtitle": "Barack Obama"`.

That said, there are only two things our app _really_ cares about:

* Having a "canonical" title that we use when writing stuff to disk (i.e. don't end up w/ redundant "Barack Obama" and "Barack_Obama" folders)
* Adding percent escapes to an un-escaped title before sending to server

I think this means I can do away w/ "denormalizing" in this case.  However, we might want to chose one (normalized or not) when writing to disk, and enforce it—if that's not being done already.

Wikipedia/UI-V5/WMFMainPageSectionController.m
bgerstle: @coreyfloyd I had it the way you're suggesting initially, but then it was more complicated because I had an (unnecessary) fallback layer. now that it doesn't fall back to site info if the TFA preview fails, the chain should be straightforward.  i did kind of like the ternary though ;-)

Wikipedia/Categories/AFHTTPRequestOperationManager+WMFDesktopRetry.h
bgerstle: Yeah, I think I'll go w/ option 2 for now. YAGNI!

bgerstle:
:-1: need to fix some stuff to allow featured article to update. stand by

bgerstle:
fixed the issue in dbe8e45:

image

Wikipedia/UI-V5/WMFMainPageSectionController.m
coreyfloyd: I'm sure I'm not able to fully understand the complications you are describing from the comments. If you want to discus options we can and figure something out. But as it is now its pretty difficult to grok - we shod be able to find something to make it more readable.

bgerstle:
@coreyfloyd i think your comments should all be addressed. could you please re-review?

Wikipedia/Networking/Fetchers/WMFEnglishFeaturedTitleFetcher.m
```diff
@@ -0,0 +1,174 @@
...
+@implementation WMFEnglishFeaturedTitleRequestSerializer
+
++ (NSDateFormatter*)featuredArticleDateFormatter {
+    static NSDateFormatter* feedItemDateFormatter;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        feedItemDateFormatter = [[NSDateFormatter alloc] init];
+        feedItemDateFormatter.dateFormat = @"MMMM d, YYYY";
+        // feed format uses US dates—specifically month names
+        feedItemDateFormatter.locale = [NSLocale localeWithLocaleIdentifier:@"en-US"];
```
coreyfloyd: @bgerstle I left this but it got lost in an outdated diff…

I don't think en-US is a locale? This is probably evaluating to nil. en_US is a locale. Maybe it accepts the hyphen format too? Not sure, but would check.

If it is evaluating to nil, then do we even need to set the locale?

Also any reason not to use the currentLocale? Even if they are getting content in english, chrome is usually still in the users device language.

bgerstle: You're right, Apple docs say to use an underscore, but looks like a hyphen also works:

```
(lldb) po [[NSLocale localeWithLocaleIdentifier:@"en_US"] displayNameForKey:NSLocaleIdentifier value:@"en-US"]
English (United States)

(lldb) po [[NSLocale localeWithLocaleIdentifier:@"en-US"] displayNameForKey:NSLocaleIdentifier value:@"en-US"]
English (United States)
```

> Also any reason not to use the currentLocale? 

We don't want current local because it has to be english (keep in mind this is for Today's Featured Article, which is on EN wiki).

coreyfloyd: I would ask/confirm the product and/or designers on this.

For instance, if the search site is english, but the device language is in french, then the "Continue Reading" "Random", Nearby", and "Because You read" section headers will all be in french, but this date will be the only header not localized to the device language and instead be in english.

bgerstle: reiterating from IRC: this is for API URL construction, not display, so we don't need to worry about localizing strings from this date formatter. can add comments to make this clearer

PR #236, opened 2015-11-10
Make iOS 8 use pre-SanFrancisco article font.

montehurd:
iOS 8 Before fix:
screen shot 2015-11-09 at 4 42 05 pm

iOS 8 After fix:
screen shot 2015-11-09 at 4 41 33 pm


bgerstle:
:+1:

PR #237, opened 2015-11-10
Fix for potential high cpu usage on image widening.

montehurd:
Android guys identified and fixed this. Recommended we do same just
in case:

https://gerrit.wikimedia.org/r/#/c/247876/1/www/js/transforms/widenImages.js

### DIFF
--- a/Wikipedia/assets/bundle.js
+++ b/Wikipedia/assets/bundle.js
@@ -1056,6 +1056,7 @@ function widenImage(image) {
 
 function maybeWidenImage() {
     var image = this;
+    image.removeEventListener('load', maybeWidenImage, false);
     if (shouldWidenImage(image)) {
         widenImage(image);
     }
--- a/www/js/transforms/widenImages.js
+++ b/www/js/transforms/widenImages.js
@@ -91,6 +91,7 @@ function widenImage(image) {
 
 function maybeWidenImage() {
     var image = this;
+    image.removeEventListener('load', maybeWidenImage, false);
     if (shouldWidenImage(image)) {
         widenImage(image);
     }


coreyfloyd:
@montehurd: did you verify the same thing was happening on iOS? Also is there any downside to removing the listener?

bgerstle:
Agree w/ @coreyfloyd. At the very least, you could try adding console logs and viewing them in the safari debugger (or Xcode) to see how our various `load` event listeners behave.  We might be running into this with other stuff as well.  Also, keep in mind that Android only saw this on a very specific device (and maybe OS?) combo.  It's not clear to me where this is a bug fix or a browser-bug workaround.

montehurd:
@coreyfloyd @bgerstle I tested it for 5 minutes or so to see if removing the listener had any adverse effects. Tried all sorts of things (rotate, reload, collapse/expand tables etc), but it all seems to be ok. 

On Android iirc they had a hard time reproducing this and it took a while to finally find devices which exhibited the issue.

coreyfloyd:
> @coreyfloyd @bgerstle I tested it for 5 minutes or so to see if removing the listener had any adverse effects. Tried all sorts of things (rotate, reload, collapse/expand tables etc), but it all seems to be ok.

@montehurd That is good. In addition to testing it, what about from a logic perspective? Meaning: Why were we not removing the listener before? Was it intentional? Did keeping it around serve a purpose?

> On Android iirc they had a hard time reproducing this and it took a while to finally find devices which exhibited the issue.

As @bgerstle mentioned iOS is a bit different - we have fewer devices with known hardware configurations and a different browser. So it may just not be an issue - especially if its a hardware bug.

It may be helpful to know if the exploding in memory usage that the Android team saw was expected based on the logic ("oh we are never releasing listeners, of course the memory is growing!") or was it a bug in the browser/device/OS ("oh, keeping these listeners around should not be causing this memory to leak, but it is, so lets remove them.")

bgerstle:
@coreyfloyd it was a CPU issue, not memory. supposedly due to an infinite loop in the old implementation.  not clear why it was recursing, all I can think of is that somehow the `widenImages` call resulted (somehow) in another `load` event, which then caused the callback to fire, which then caused another `load` event, ad inifitum

montehurd:
@coreyfloyd 
`Why were we not removing the listener before? Was it intentional? Did keeping it around serve a purpose?`
^ Oh I wrote the widening code - I don't think it needs to happen more than once, hence the fix.

My suspicion is that on Android it was an issue with particularly slow/old devices - no proof though.

montehurd:
Ah! As @bgerstle points out it was identified to be a recursion issue on Android. @coreyfloyd I should have stated that. Sorry.

coreyfloyd:
If we think the code being removed was unnecessary anyways then this is fine with me. But since we haven't actually observed the issue, its also fine with me to leave as is. However you guys want to resolve it… :neutral_face:

PR #240, opened 2015-11-10
Fix bug where header frame is hosed when rotating

coreyfloyd:
https://phabricator.wikimedia.org/T117913

Fixed this bug but discovered another:
https://phabricator.wikimedia.org/T118300

Wan't able to fix it after numerous attempts so filed it and moved on.

### DIFF
--- a/Wikipedia/UI-V5/WMFTableOfContentsViewController.swift
+++ b/Wikipedia/UI-V5/WMFTableOfContentsViewController.swift
@@ -95,19 +95,23 @@ public class WMFTableOfContentsViewController: UITableViewController, WMFTableOf
         }
     }
 
+    // MARK: - Header
+    func forceUpdateHeaderFrame(){
+        //See reason for fix here: http://stackoverflow.com/questions/16471846/is-it-possible-to-use-autolayout-with-uitableviews-tableheaderview
+        self.tableView.tableHeaderView!.setNeedsLayout()
+        self.tableView.tableHeaderView!.layoutIfNeeded()
+        let headerHeight = self.tableView.tableHeaderView!.systemLayoutSizeFittingSize(UILayoutFittingCompressedSize).height
+        var headerFrame = self.tableView.tableHeaderView!.frame;
+        headerFrame.size.height = headerHeight
+        self.tableView.tableHeaderView!.frame = headerFrame;
+        self.tableView.tableHeaderView = self.tableView.tableHeaderView
+    }
+
     // MARK: - UIViewController
     public override func viewDidLoad() {
         super.viewDidLoad()
         let header = WMFTableOfContentsHeader.wmf_viewFromClassNib()
-        
-        var headerString = localizedStringForKeyFallingBackOnEnglish("table-of-contents-heading")
-        
-        if(NSLocale.wmf_isCurrentLocaleEnglish()){
-            headerString = headerString.uppercaseStringWithLocale(NSLocale.currentLocale())
-        }
-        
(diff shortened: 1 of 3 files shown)


Wikipedia/UI-V5/WMFTableOfContentsViewController.swift
```diff
@@ -129,7 +134,15 @@ public class WMFTableOfContentsViewController: UITableViewController, WMFTableOf
         super.viewDidDisappear(animated)
         deselectAllRows()
     }
-
+    
+    public override func viewWillTransitionToSize(size: CGSize, withTransitionCoordinator coordinator: UIViewControllerTransitionCoordinator) {
+        
+        coordinator.animateAlongsideTransition({ (context) -> Void in
```
bgerstle: hmm this reads a bit weird in swift. I'm guessing the second closure (which is indented strangely) is the completion block? maybe we should use the labels in this instance, as the "implicit trailing closure as last parameter" feature works best for functions with only one parameter (or parameters that are more clearly labeled).

Wikipedia/UI-V5/WMFTableOfContentsViewController.swift
```diff
@@ -95,19 +95,23 @@ public class WMFTableOfContentsViewController: UITableViewController, WMFTableOf
         }
     }
 
+    // MARK: - Header
+    func forceUpdateHeaderFrame(){
+        //See reason for fix here: http://stackoverflow.com/questions/16471846/is-it-possible-to-use-autolayout-with-uitableviews-tableheaderview
+        self.tableView.tableHeaderView!.setNeedsLayout()
```
bgerstle: crash operators everywhere!

Wikipedia/UI-V5/WMFTableOfContentsViewController.swift
```diff
@@ -95,19 +95,23 @@ public class WMFTableOfContentsViewController: UITableViewController, WMFTableOf
...
+        self.tableView.tableHeaderView!.layoutIfNeeded()
+        let headerHeight = self.tableView.tableHeaderView!.systemLayoutSizeFittingSize(UILayoutFittingCompressedSize).height
+        var headerFrame = self.tableView.tableHeaderView!.frame;
+        headerFrame.size.height = headerHeight
+        self.tableView.tableHeaderView!.frame = headerFrame;
+        self.tableView.tableHeaderView = self.tableView.tableHeaderView
```
bgerstle: I read the SO post, and see what you're doing, but still...
wat

Wikipedia/UI-V5/WMFTableOfContentsViewController.swift
coreyfloyd: @bgerstle this is just how Xcode autocompletes the blocks in the methods. I didn't really argue with it - all the closure stuff in swift reads weird to me.

Wikipedia/UI-V5/WMFTableOfContentsViewController.swift
coreyfloyd: @bgerstle yeah - thats how I feel about it too… basically you need to manually update the header view to get the sizing right and then reset the tableHeaderView property to make the table view respect the new size. 

From my poking around and SO spelunking this seems to be a known problem, and this answer was accepted a few times on variations of this question. 

I think we have seen this before - that headers and footers don't necessarily play nice with auto layout.

bgerstle:
@coreyfloyd lgtm :+1: what say you, @montehurd?

PR #243, opened 2015-11-11
On main pages gray out TOC button instead of hiding it.

montehurd:
https://phabricator.wikimedia.org/T118222

### DIFF
--- a/Wikipedia/UI-V5/WMFArticleContainerViewController.m
+++ b/Wikipedia/UI-V5/WMFArticleContainerViewController.m
@@ -302,7 +302,7 @@ NS_ASSUME_NONNULL_BEGIN
     self.tableOfContentsToolbarItem.enabled = self.article != nil;
     self.shareToolbarItem.enabled           = self.article != nil;
     self.languagesToolbarItem.enabled       = self.article.languagecount > 1;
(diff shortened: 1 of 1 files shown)


montehurd:
Main page with disabled TOC button:
screen shot 2015-11-10 at 4 04 13 pm

For reference, a non-main page:
screen shot 2015-11-10 at 4 04 02 pm

Wikipedia/UI-V5/WMFArticleContainerViewController.m
```diff
@@ -304,6 +302,7 @@ - (void)updateToolbarItemEnabledState {
     self.tableOfContentsToolbarItem.enabled = self.article != nil;
     self.shareToolbarItem.enabled           = self.article != nil;
     self.languagesToolbarItem.enabled       = self.article.languagecount > 1;
+    self.tableOfContentsToolbarItem.enabled = !self.article.isMain;
```
bgerstle: minor, can you combine this w/ the statement on line 302? (i.e. `enabled = self.article != nil && !self.articleisMain`). otherwise, the button will be accidentally enabled when the article is `nil`.

coreyfloyd:
Good with me after @bgerstle's comment is addressed :+1:

PR #244, opened 2015-11-11
Fix gallery crash

bgerstle:
Phab: [T118047](https://phabricator.wikimedia.org/T118047)

---

In a nutshell, the bug was caused by:

1. Creating header gallery w/ the article's current image list
2. Webview fetches images, which are then added to global current article 😱 as a side effect 👻
3. When image gallery is presented, using the (now altered) article image list

Instead of working around the race condition & side effects, I chose to eliminate them.  As an add'l bit of boyscouting, I cleaned up and added tests for our (somewhat outdated and IMO questionable) image parsing logic.  Some highlights:

- No more C functions (you're welcome @coreyfloyd!)
- Test that we exclude small images
- Test that we parse out both `src` and `srcset` elements
- Test that we get the (no non-zero!) expected number of images when parsing the Obama article

As an added, bonus this also fixes images sneaking past our size threshold (you're welcome, @montehurd!).  You can verify this by seeing the add'l icons in both galleries on master, and _not_ seeing them in this patch.


MediaWikiKit/MediaWikiKit/MWKArticle.h
```diff
@@ -99,6 +99,25 @@ static const NSInteger kMWKArticleSectionNone = -1;
 - (MWKImage*)importImageURL:(NSString*)url sectionId:(int)sectionId;
 
 /**
+ *  Add the given URL to the receiver's image list, as well as to the corresponding section's image list.
+ *
+ *  @param sourceURL The URL string to add.
+ *  @param sectionId The section whose image list the URL should also be added to.
+ *  @param skipIfPresent @c YES if the image should only be added if the list doesn't contain it.
+ */
+- (void)updateImageListsWithSourceURL:(NSString*)sourceURL inSection:(int)sectionId skipIfPresent:(BOOL)skipIfPresent;
```
coreyfloyd: Seems this should be "add" instead of "update" - I suspected thats what it did, but had to read the comments to be sure what this method was doing.

coreyfloyd:
left one comment… if you think the naming is sufficient, then it is ok with me. I am otherwise good :+1:

MediaWikiKit/MediaWikiKit/MWKArticle.h
bgerstle: yeah, i wrote a `HAX` comment about it, but the data layer is very error-prone when it comes to duplicating or accumulating data.  i'll change it to say `add` to be more explicit about its typical usage, but anytime you're in `MWK` land you need to be careful of this sort of thing 😰

coreyfloyd:
@montehurd I am good here - go ahead and checkout and merge.

PR #245, opened 2015-11-12
Fix full-text search regression

bgerstle:
Phab: [T117012](https://phabricator.wikimedia.org/T117012)

---

Regression caused by some refactoring & bug fixing.  We were relying on KVO updates that weren't actually happening.  Now they happen!  Check out the tests, but if you want [actual proof](https://www.youtube.com/watch?v=m0c38Wtdvz0)

The GIF is a bit fast, but IRL the first prefix result renders, followed by the subsequent full-text results.
claude monet search


coreyfloyd:
:+1:

PR #253, opened 2015-11-13
Fix selection of recent titles from before today. Plus add Quick & Nimble!

bgerstle:
Phab: [T118291](https://phabricator.wikimedia.org/T118291)

---

Why just fix something, when you can add tests? Why just add tests, when you can finally introduce BDD frameworks into the project?!  Curious what you guys think of the added test's readability.  The hang-up I ran into was syntax coloring & auto-complete breaking (i.e. clang indexer errors).

**Before merging, please checkout, build, & run the tests to see if you get the same errors.**


montehurd:
Tests passed.

bgerstle:
except not? ... hokay

montehurd:
:+1: It works!

bgerstle:
dern right it does 😤 if this keeps cropping up i'll have to see what's going on. maybe another timeout-related issue

coreyfloyd:
yeah -  much prefer the BDD syntax - been waiting to get something like this in here - its like Hamcrest goes out of its way to be unreadable.

PR #255, opened 2015-11-13
table view conversion

coreyfloyd:
Converts the following to table views: 
- home view
- "more like lists" to table views.
- Read more at the bottom of articles

Additionally this removes all collection view code.


coreyfloyd:
FYI - waiting until this is merged to remove the self sizing waterfall pod - that adds a lot of clutter to the diff.

coreyfloyd:
And of course, I pushed the pod update accidentally… doh

montehurd:
Couple small things:
- the gray background on the nearby images was per Nirzar. Prob want to restore that.
- i'm still seeing no placeholder in nearby cells, but only some of the time, I think it's difference between cell being created vs dequeue'd/reused (placeholder needs to be set in both cases):

screen shot 2015-11-16 at 3 03 15 pm

- delete dragging works in feed, but probably shouldn't:

screen shot 2015-11-16 at 3 20 59 pm
screen shot 2015-11-16 at 3 21 11 pm

montehurd:
Found one more issue which may be unrelated to this pr:
- Peek on the article title links in feed "Because you read... " header shows peek of that title in mobile web. Either disable peeks in these headers or fix the peeks?

PR #257, opened 2015-11-17
Adding Placeholder cells to home VC

coreyfloyd:
Adding 2 placeholder cells - one for previews, on for nearby. 

Also some refactoring to unbundle main pages from featured articles so that business logic can be placed within the schema manager.

Awaiting artwork form Nirzar, but much of the logic are done.

Note: I explored some ways to share the placeholder logic and expose it as a first class citizen in the homesectioncontroller API, but some refactors need to take place before that really makes sense  (mainly separating out datasource logic from cell updating logic).


coreyfloyd:
Screen caps:
screen shot 2015-11-18 at 12 17 31 pm
screen shot 2015-11-18 at 12 17 28 pm

coreyfloyd:
@bgerstle @montehurd this is ready to go.

Wikipedia/Categories/UIColor+WMFStyle.m
```diff
@@ -54,6 +54,16 @@ + (instancetype)wmf_lightGrayColor {
     return c;
 }
 
++ (instancetype)wmf_placeholderLightGrayColor {
+    static UIColor* c = nil;
+
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        c = [UIColor colorWithWhite:0.975 alpha:1];
```
montehurd: Maybe dry this and wmf_placeholderImageBackgroundColor together.

coreyfloyd: I would but they have different values.

montehurd: 1.5% diff is fairly slight. No worry if you want to keep it though.

coreyfloyd: yeah - it has to match the image that was given - so I used the value from that (this color is what makes the button the same gray as the placeholders fake lines of text)

Wikipedia/UI-V5/WMFHomeSectionSchema.m
```diff
@@ -122,7 +128,7 @@ - (void)update:(BOOL)force {
     //Check Tweak
     if (!FBTweakValue(@"Home", @"General", @"Always update on launch", NO)) {
         //Check force flag and minimum relaod time
-        if (!force && [[NSDate date] timeIntervalSinceDate:self.lastUpdatedAt] < WMFHomeMinimumAutomaticReloadTime) {
+        if (!force && self.lastUpdatedAt && [[NSDate date] timeIntervalSinceDate:self.lastUpdatedAt] < WMFHomeMinimumAutomaticReloadTime) {
```
bgerstle: maybe a product question, but should we make sure "featured" is refreshed when it's a new day?

Wikipedia/UI-V5/WMFMainPageSectionController.m
```diff
@@ -92,96 +73,64 @@ - (UIImage*)headerIcon {
...
-        WMFArticlePreviewTableViewCell* previewCell = (WMFArticlePreviewTableViewCell*)cell;
-        previewCell.titleText       = self.featuredArticlePreview.displayTitle;
-        previewCell.descriptionText = self.featuredArticlePreview.wikidataDescription;
-        previewCell.snippetText     = self.featuredArticlePreview.extract;
-        [previewCell setImageURL:self.featuredArticlePreview.thumbnailURL];
-        [previewCell setSaveableTitle:[self titleForItemAtIndex:indexPath.row] savedPageList:self.savedPageList];
+        if (self.siteInfo) {
+            mainPageCell.mainPageTitle.text = self.siteInfo.mainPageTitleText;
+        } else {
+            mainPageCell.mainPageTitle.text = @"Loading…";
```
bgerstle: is there a localized string for this? if not, can you create one?

montehurd: Ya let's start adding these as we go.

Wikipedia/UI-V5/WMFHomeSectionSchema.m
bgerstle: this might be out of scope for this PR. we can have a separate discussion w/ Josh since scroll position restoration is tied up in this as well

Wikipedia/UI-V5/WMFHomeSectionSchema.m
```diff
@@ -251,7 +259,28 @@ - (WMFHomeSection*)mainPageSection {
     return main;
 }
 
-- (WMFHomeSection*)continueReadingSection {
+- (nullable WMFHomeSection*)featuredArticleSection {
+    //Never show if we are not in english
+    if (![self.site.language isEqualToString:@"en"] && ![self.site.language isEqualToString:@"en-US"]) {
```
montehurd: I think maybe check for "en" and prefix of "en-" - so we account for en-UK... I hear they speak English there ;)

Wikipedia/UI-V5/WMFHomeSectionSchema.m
coreyfloyd: featured is refreshed if the date is new. 

This is the update the entire feed logic - which is different.

Wikipedia/UI-V5/WMFMainPageSectionController.m
coreyfloyd: sorry… placeholder… replacing

Wikipedia/UI-V5/WMFHomeSectionSchema.m
coreyfloyd: yeah but featured article is a feature that is only enabled for en us, not en uk

bgerstle: @montehurd i guess in this case we only want en.wikipedia.org, so checking for "en" should be better. there is no "en-us.wikipedia.org"

bgerstle: @coreyfloyd so, probably better to check lang == "en" and domain "wikipedia.org" (just in case)

montehurd: @bgerstle ah yeah i was confusing wiki site langs with iOS lang codes...

bgerstle:
:+1: looks good! there a couple things I'd like to revisit when we have time:

- elegance of section sorting behaviors (see comment on `-[WMFHomeSection compare:]`
- how we handle placeholder items (using dummy values could lead to unexpected crashes)

PR #258, opened 2015-11-18
Article language tweaks for Article "Read More"

montehurd:
Hebrew version of "food" article:

**Before:**
screen shot 2015-11-17 at 4 36 41 pm

**After:**
screen shot 2015-11-17 at 4 03 06 pm


Wikipedia/UI-V5/TableOfContentsReadMoreItem.swift
```diff
@@ -13,7 +13,15 @@ public protocol TableOfContentsFooterItem : TableOfContentsItem {
 }
 
 public class TableOfContentsReadMoreItem : NSObject, TableOfContentsFooterItem {
-    public let titleText: String = localizedStringForKeyFallingBackOnEnglish("article-read-more-title")
+    let site:MWKSite
+    init(site: MWKSite) {
+        self.site = site
```
coreyfloyd: is calling `super.init()` optional in swift?

Wikipedia/UI-V5/WMFTableOfContentsViewController.swift
```diff
@@ -111,6 +113,7 @@ public class WMFTableOfContentsViewController: UITableViewController, WMFTableOf
     public override func viewDidLoad() {
         super.viewDidLoad()
         let header = WMFTableOfContentsHeader.wmf_viewFromClassNib()
+        header.articleSite = delegate?.tableOfContentsArticleSite()
```
coreyfloyd: we may want to assert that the delegate is set in `viewDidLoad` now since we require it for the header.

Wikipedia/UI-V5/TableOfContentsReadMoreItem.swift
montehurd: Fixed. Interestingly in Swift instance vars need to be set before super init... http://stackoverflow.com/a/24150540

coreyfloyd: yeah - anything that is a `let` needs to be set before you call init - since that is a constant (can't change later). for instance you can see the delegate ivar in the doc view controller is just a `var`. so it doesn't need to be set.

coreyfloyd:
:+1:

bgerstle:
:+1: restarted the build, which looks to be failing due to timeout issues.

PR #260, opened 2015-11-18
Fix for appending to empty search results & full-text search regression

bgerstle:
Phab: [T118921](https://phabricator.wikimedia.org/T118921)

---

Caused a crash due to `nil` throwing a wrench in KVO connections.  Also re-fixed full-text appending which wasn't hooked up properly.  Will add an integration test for all of this soon...


coreyfloyd:
@bgerstle I think there is some weirdness in the KVO code… don't want to to type it all out, but can discuss on hangout when you get back.

bgerstle:
@coreyfloyd please re-review when build finishes

montehurd:
IT WORKZ!!!

bgerstle:
TRAVIS Y U TIMEOUT SO MUCH :trollface:

PR #265, opened 2015-11-19
Add search language selection

coreyfloyd:
une image vaut mieux que mille mots

screen shot 2015-11-19 at 2 19 11 pm

### DIFF
--- a/Wikipedia/Categories/UIImage+WMFStyle.m
+++ b/Wikipedia/Categories/UIImage+WMFStyle.m
@@ -14,4 +14,15 @@
     return img;
 }
 
++ (instancetype)wmf_imageFromColor:(UIColor*)color {
+    CGRect rect = CGRectMake(0, 0, 1, 1);
+    UIGraphicsBeginImageContext(rect.size);
+    CGContextRef context = UIGraphicsGetCurrentContext();
+    CGContextSetFillColorWithColor(context, [color CGColor]);
+    CGContextFillRect(context, rect);
+    UIImage* image = UIGraphicsGetImageFromCurrentImageContext();
+    UIGraphicsEndImageContext();
+    return image;
+}
+
 @end
--- a/Wikipedia/UI-V5/WMFArticleContainerViewController.m
+++ b/Wikipedia/UI-V5/WMFArticleContainerViewController.m
@@ -132,8 +132,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (void)showLanguagePicker {
     LanguagesViewController* languagesVC = [LanguagesViewController wmf_initialViewControllerFromClassStoryboard];
-    languagesVC.downloadLanguagesForCurrentArticle = YES;
-    languagesVC.languageSelectionDelegate          = self;
+    languagesVC.articleTitle              = self.articleTitle;
+    languagesVC.languageSelectionDelegate = self;
     [self presentViewController:[[UINavigationController alloc] initWithRootViewController:languagesVC] animated:YES completion:nil];
 }
 
--- a/Wikipedia/UI-V5/WMFSearchViewController.m
+++ b/Wikipedia/UI-V5/WMFSearchViewController.m
@@ -174,7 +174,7 @@ static NSUInteger const kWMFMinResultsBeforeAutoFullTextSearch = 12;
     [self.searchField setPlaceholder:MWLocalizedString(@"search-field-placeholder-text", nil)];
 }
 
-- (void)configureLangaugeButtons{
+- (void)configureLanguageButtons{
     [self.languageButtons enumerateObjectsUsingBlock:^(UIButton*  _Nonnull obj, NSUInteger idx, BOOL * _Nonnull stop) {
         obj.tintColor = [UIColor wmf_blueTintColor];
     }];
@@ -197,7 +197,7 @@ static NSUInteger const kWMFMinResultsBeforeAutoFullTextSearch = 12;
     [super viewDidLoad];
 
     [self configureSearchField];
-    [self configureLangaugeButtons];
+    [self configureLanguageButtons];
     
     // move search field offscreen, preparing for transition in viewWillAppear
     self.searchFieldTop.constant = -self.searchFieldHeight.constant;
--- a/Wikipedia/UI-V5/WMFSearchViewController.m
+++ b/Wikipedia/UI-V5/WMFSearchViewController.m
@@ -474,6 +474,7 @@ static NSUInteger const kWMFMinResultsBeforeAutoFullTextSearch = 12;
 
 - (void)selectLanguageForButton:(UIButton*)button{
     NSUInteger index = [self.languageButtons indexOfObject:button];
+    NSAssert(index != NSNotFound, @"language button not found for language!");
     if(index != NSNotFound){
         MWKLanguageLink* lang = self.searchLanguages[index];
         [self selectLanguageForSite:lang.site];
--- a/Wikipedia/UI-V5/WMFSearchViewController.m
+++ b/Wikipedia/UI-V5/WMFSearchViewController.m
@@ -499,17 +499,10 @@ static NSUInteger const kWMFMinResultsBeforeAutoFullTextSearch = 12;
     [self searchForSearchTerm:self.searchField.text];
 }
 
-- (IBAction)setLanguageOne:(id)sender {
+- (IBAction)setLanguageWithSender:(id)sender {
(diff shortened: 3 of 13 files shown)


bgerstle:
Je ne parle pas français

bgerstle:
@coreyfloyd are we persisting the results between languages? that might be kinda cool

bgerstle:
we would get that "for free" if we used the URL cache for the JSON, at least

Wikipedia/Categories/UIImage+WMFStyle.m
```diff
@@ -14,4 +14,15 @@ + (UIImage*)wmf_placeholderImage {
     return img;
 }
 
++ (instancetype)wmf_imageFromColor:(UIColor*)color {
+    CGRect rect = CGRectMake(0, 0, 1, 1);
```
bgerstle: i'm guessing `backgroundColor` didn't cut it?

bgerstle: if this needs to stay, can we cache these? no need to recreate the same image a bunch of times, and this kind of stuff can be expensive.

bgerstle: if we don't want to do a shared cache object, we could at least wrap something domain-specific in a dispatch once (much like the other color categories). e.g. `+ (UIImage*)wmf_languageSelectionBackgroundImage`

Wikipedia/UI-V5/WMFArticleContainerViewController.m
```diff
@@ -132,8 +132,8 @@ - (instancetype)initWithArticleTitle:(MWKTitle*)title
 
 - (void)showLanguagePicker {
     LanguagesViewController* languagesVC = [LanguagesViewController wmf_initialViewControllerFromClassStoryboard];
-    languagesVC.downloadLanguagesForCurrentArticle = YES;
-    languagesVC.languageSelectionDelegate          = self;
+    languagesVC.articleTitle              = self.articleTitle;
```
bgerstle: you could be a good boyscout by drying up essentially this whole method into a `LanguagesViewController` class method

coreyfloyd:
no we aren't persisting - we can always optimize later.

Wikipedia/Categories/UIImage+WMFStyle.m
coreyfloyd: > i'm guessing backgroundColor didn't cut it?

Produces rounded corners.

> if this needs to stay, can we cache these? no need to recreate the same image a bunch of times, and this kind of stuff can be expensive.

Not sure why we need to cache these. If an object needs to generate a lot of images form colors at some point then that would make sense.  Seems like YAGNI / pre-optimization. 
> If we don't want to do a shared cache object, we could at least wrap something domain-specific in a dispatch once (much like the other color categories). e.g. + (UIImage*)wmf_languageSelectionBackgroundImage

Pre-Optimization. Colors are a shared pallet. This isn't a shared pallet of images. This is why we have viewDidLoad - to do something once for a specific view.

Wikipedia/UI-V5/WMFArticleContainerViewController.m
coreyfloyd: Already boycotted a few things… I feel good enough. Lets get a beta out instead of refactoring.

Wikipedia/Categories/UIImage+WMFStyle.m
bgerstle: ah, i was thinking we were using this on some scrolling/reused view. if it's only done once for all 3 buttons at `viewDidLoad` it should be OK for now

Wikipedia/UI-V5/WMFSearchViewController.m
```diff
@@ -391,6 +434,52 @@ - (void)saveLastSearch {
...
+            }];
+        }
+    }];
+    NSString* query = self.searchField.text;
+    [self searchForSearchTerm:query];
+}
+
+- (void)selectLanguageForButton:(UIButton*)button{
+    NSUInteger index = [self.languageButtons indexOfObject:button];
+    if(index != NSNotFound){
```
bgerstle: this should never be false, can you add an assertion?

Wikipedia/UI-V5/WMFSearchViewController.m
```diff
@@ -410,6 +499,23 @@ - (IBAction)searchForSuggestion:(id)sender {
...
 
+- (IBAction)setLanguageOne:(id)sender {
+    [self selectLanguageForButton:sender];
+}
+
+- (IBAction)setLanguageTwo:(id)sender {
+    [self selectLanguageForButton:sender];
+}
+
+- (IBAction)setLanguageThree:(id)sender {
```
bgerstle: can we get rid of these extra IBActions? seems like they're not doing anything

coreyfloyd: yes they do

coreyfloyd: or do you want them to all call the same action? i can do that… usually don't though…

bgerstle: IIUC each method is doing the same thing: call `selectLanguageForButton:` with the `sender`, which seems to me like you could just use `selectLanguageForButton:` as the IBAction

bgerstle:
@coreyfloyd code looks good! :+1: @montehurd do you want to review? i'll clone the source & test in a bit

montehurd:
The feed's random and featured article switch to whatever search language is chosen. We may want a follow-on to decide if this is desired...

montehurd:
Found an issue which may be tricky to solve.

If user's top 3 langs have long names the buttons all blend together:

screen shot 2015-11-19 at 2 27 25 pm

bgerstle:
I agree, that doesn't seem desirable. But then again, that is the legacy
functionality

bgerstle:
Do we make it scrollable, or truncate?

bgerstle:
Maybe we want separate sites for the feed and search?

coreyfloyd:
> If user's top 3 langs have long names the buttons all blend together:

@montehurd yeah - looked at that just now - not easy to solve because its a tool bar with embedded buttons and doesn't really understand constraint relationships like that. Aggravating. 

Really though - thats an extreme edge case. We should file a bug - just don't think we need to worry about that to get this done for beta. We can polish it up after next week.

bgerstle:
I say we talk to deciding, and if it's acceptable truncate for now and
revisit how to solve this case

montehurd:
I pinged Nirzar so he can figure out how he wants to handle these, but my recommendation was going with this for now and doing changes as follow-ons.

Scrolling through the complete list of languages it doesn't look like long lang names is really much of an edge case though - lots of them have long or even compound scripts in their names. I updated the phab ticket with some screenshots to follow-on designs can handle those cases as well.

coreyfloyd:
> Scrolling through the complete list of languages, it doesn't look like long lang names is really much of an edge case though - lots of them have long or even compound scripts in their names. I updated the phab ticket with some screenshots to follow-on designs can handle those cases as well.

@montehurd Yeah - There are lots of long languages… thats not the edge case. Any one could read a single language that is long. The edge case is that someone has 2 or 3 long language names as their favorites. Thats easy for us to do as a test, but unlikely for someone to actually have in practice.

> I say we talk to deciding, and if it's acceptable truncate for now and
revisit how to solve this case

@bgerstle Truncation is even tricky here… The buttons get as big as they "need to be" - they don't know how much room they have. So you would have to trim to an arbitrary number of characters or something.

But like @montehurd said, I'm inclined to iterate on this. Its something that can be worked out post beta… I probably have to do some weird stuff to get the tool bar to stop being a dick.

coreyfloyd:
I'll going to have another look at the sizing in the mean time… to see if i can get a quick follow up, but just not promising so far.

montehurd:
@coreyfloyd 

> Thats easy for us to do as a test, but unlikely for someone to actually have in practice.

Hmm... I'd wager it's likely there exist millions of people in certain geographic regions who happen to speak more than one regional/related language which have a long-ish names in the list.

montehurd:
But I say go with this for now and let Nirzar and Josh decide how to polish.

bgerstle:
@coreyfloyd the truncation is better, but will need to follow-up w/ nirzar that if we do want to truncate, we need to figure out how we want to do it (maybe giving priority to selected lang):

image

PR #268, opened 2015-11-20
Fix placeholder tap crash

coreyfloyd:


Wikipedia/UI-V5/WMFHomeViewController.m
```diff
@@ -522,7 +522,10 @@ - (BOOL)tableView:(UITableView*)tableView shouldHighlightRowAtIndexPath:(NSIndex
 
 - (void)tableView:(UITableView*)tableView didSelectRowAtIndexPath:(NSIndexPath*)indexPath {
     id<WMFHomeSectionController> controller = [self sectionControllerForSectionAtIndex:indexPath.section];
-    MWKTitle* title                         = [controller titleForItemAtIndex:indexPath.row];
+    if ([controller respondsToSelector:@selector(shouldSelectItemAtIndex:)] && ![controller shouldSelectItemAtIndex:indexPath.item]) {
+        return;
```
bgerstle: how is this possible? i thought you couldn't select items period unless the section had data

montehurd: Ya this works, but... why?

coreyfloyd: Returning `NO` from `- (BOOL)tableView:(UITableView*)tableView shouldHighlightRowAtIndexPath:(NSIndexPath*)indexPath` only disables the selection highlight. It does not prevent the selection itself. 

After doing that, you still need to ignore the selection in `- (void)tableView:(UITableView*)tableView didSelectRowAtIndexPath:(NSIndexPath*)indexPath`

coreyfloyd: @bgerstle @montehurd ^

bgerstle: right, `UITableViewDelegate` has `should_Highlight_` but `UICollectionViewDelegate` has `should_Select_`. :+1:

bgerstle: at this point, should we call `deselect`? otherwise i think tableview might still have it in `selectedCells`?

coreyfloyd: it doesn't matter since the highlight doesn't show. and we don't have multiple selection so the next selection will always overwrite it. And since it is a table view controller and `clearsSelectionOnViewWillAppear == YES` selection is always purged then as well.

PR #269, opened 2015-11-20
Minor visual design addition

nirzar:
Added a border around uiimageview using interface builder

Phab card:
https://phabricator.wikimedia.org/T119243


montehurd:
@bgerstle ah good point.

montehurd:
@bgerstle hey do you want to maybe push a commit drying up `CHROME_OUTLINE_WIDTH (1.0f / [UIScreen mainScreen].scale)` and any other occurrences of `1.0f / [UIScreen mainScreen].scale`?

nirzar:
@montehurd @bgerstle will it crash on non-retina devices? if the line does not show up on ipad2 which is the only non retina device with iOS 8 available, we should be good. I think this is an enhancement, if it was a feature i would be worried.

bgerstle:
@nirzar no crash, good point that we can circle back and handle edge cases later if needed

PR #271, opened 2015-11-20
Show recent search header only if there are recent search terms

montehurd:
**Header is visible when there are recent search terms:**
screen shot 2015-11-20 at 3 14 02 pm

**Header is NOT visible when there are NO recent search terms:**
screen shot 2015-11-20 at 3 14 08 pm


bgerstle:
:+1:

PR #272, opened 2015-11-23
Add Picture of the Day to feed

bgerstle:
- [x] Add fetch to get image info for Common's current POTD
- [x] Display in feed w/ title
- [x] Add gradient below title
- [x] Localize extmetadata from image info (i.e. image description) [0]
- [x] Add image thumb resolution as fetch parameter (shouldn't to show 1280-wide image the feed)
- [x] don't show caption (and gradient) on top of placeholder (see `WMFPicOfTheDayTableViewCellVisualTests` and corresponding reference images)

0: POTD on 11/25 w/ device lang set to Spanish:
image


bgerstle:
@nirzar what do you think so far:

image

Note:

- Header isn't what you spec'd
- Not sure about spacing/positioning

coreyfloyd:
@bgerstle all looking good so far - i think the only thing I have is the placeholder stuff. But otherwise I'm good here.

montehurd:
One for @nirzar - it looks like the line width of the POTD icon is inconsistent with the other icons in the feed:

screen shot 2015-11-25 at 12 51 55 pm

^ also, if we're overlaying text over the image again (really?!) we're back to having all the same annoying text contrast issues with light/white background images. Did we really forget about this again?

screen shot 2015-11-25 at 12 52 00 pm

screen shot 2015-11-25 at 12 52 19 pm

bgerstle:
> ^ also, if we're overlaying text over the image again (really?!) we're back to having all the same annoying text contrast issues with light/white background images. Did we really forget about this again?

There's a gradient, but you can't see it in this picture. This is how we're currently overlaying text in the image gallery—though I believe the gradient parameters are slightly different here.

montehurd:
@bgerstle I prefer consistency with the other items actually onscreen than consistency with items from a different interface.

montehurd:
But we can always tweak later.

montehurd:
It would have been so much easier to just format it like other feed article cards with "Picture of the day" as the title and the description as the description, keeping the image height consistent with other cards.

montehurd:
@nirzar another thing to consider with the increased height of the image is it's going to take up most of the screen on small screen devices.

bgerstle:
> It would have been so much easier to just format it like other feed article cards with "Picture of the day" as the title and the description as the description, keeping the image height consistent with other cards.

All of this feedback is good, but is better suited to Phabricator, specifically [the mock for this feature](https://phabricator.wikimedia.org/T119150). I mentioned some of these points to @nirzar but didn't hear back in time.

That said, there is another point that I could've essentially re-used the preview card, but filled it w/ data for POTD. I made the judgement call that I should try to get closer to 90% of the mock, and it wasn't that hard to do.  Like you said yourself, there's always the chance we can tweak this later if we have time.

Wikipedia/Networking/Fetchers/MWKImageInfoFetcher.m
```diff
@@ -46,49 +47,88 @@ - (instancetype)initWithDelegate:(id<FetchFinishedDelegate>)delegate
...
                                           success:(void (^)(NSArray*))success
                                           failure:(void (^)(NSError*))failure {
-    NSParameterAssert([imageTitles count]);
-    NSAssert([imageTitles count] <= 50, @"Only 50 titles can be queried at a time.");
+    // This is currently only used for fullscreen image gallery.  If more dynamic image sizes are desired, expose
+    // the parameter in the API.
+    // 1280 is a well-populated image width in back-end cache that gives good-enough quality on most iOS devices
+    return [self fetchInfoForTitles:imageTitles
+                           fromSite:site
+                     thumbnailWidth:1280
```
montehurd: This seems far too large. On a 3g connection that's gonna take a while :)

bgerstle: this is the size we've been fetching for the fullscreen gallery. the gallery will first show the image we've already retrieved for the image that's in the article, and then goes to fetch the higher-res one in the background.

bgerstle: to be clear: not used for the feed (see comment). also, I believe you were the one who recommended I use 1280 originally, as I used to fetch the original file (which could be much, _much_ larger!)

montehurd: Ah didn't realize it wasn't for the feed. Thanks.

PR #274, opened 2015-11-23
Conditionally append read more

coreyfloyd:
If no read more results are returned, not read more is not appended to the bottom of the article or the ToC.


montehurd:
Confirmed that if you have no interwebz and try to load saved page, the read more bits are gone :)

PR #275, opened 2015-11-23
Preserve search results during a session

coreyfloyd:
- Keep search VC as a static ivar
- Remove it when the app backgrounds
- Remove it if we get a memory warning
- Don't restore if the search site has changed

### DIFF
--- a/Wikipedia/UI-V5/UIViewController+WMFSearchButton.m
+++ b/Wikipedia/UI-V5/UIViewController+WMFSearchButton.m
@@ -25,11 +25,15 @@ static WMFSearchViewController * _sharedSearchViewController = nil;
 }
 
 + (void)wmfSearchButton_applicationDidEnterBackgroundWithNotification:(NSNotification*)note {
-    _sharedSearchViewController = nil;
+    if (!_sharedSearchViewController.view.window) {
+        _sharedSearchViewController = nil;
+    }
 }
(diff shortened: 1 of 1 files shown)


Wikipedia/UI-V5/UIViewController+WMFSearchButton.m
```diff
@@ -11,11 +11,27 @@
...
 
+static WMFSearchViewController * _sharedSearchViewController = nil;
+
++ (void)load {
+    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(wmfSearchButton_applicationDidEnterBackgroundWithNotification:) name:UIApplicationDidEnterBackgroundNotification object:nil];
+    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(wmfSearchButton_applicationDidReceiveMemoryWarningWithNotification:) name:UIApplicationDidReceiveMemoryWarningNotification object:nil];
+}
+
++ (void)wmfSearchButton_applicationDidEnterBackgroundWithNotification:(NSNotification*)note {
+    _sharedSearchViewController = nil;
```
bgerstle: should we only `nil` it out if it's not in the window? e.g. if the user goes to background while on search? (same for memory warning)

montehurd:
This is great!

PR #276, opened 2015-11-23
enforce code coverage via fastlane

bgerstle:
(don't depend on Xcode)


bgerstle:
w00t, it's back! @coreyfloyd @montehurd can haz merge plz

PR #277, opened 2015-11-23
Only show read more for recent items that have been significantly read.

coreyfloyd:
Adding "significantly viewed" property to MWKHistory to signify when a user as read an article.

- Significant viewed property is toggled by the article VC after 30 seconds (tweakable)
- Home Schema ignores history items that are not significantly viewed

### DIFF
--- a/MediaWikiKit/MediaWikiKit/MWKHistoryList.m
+++ b/MediaWikiKit/MediaWikiKit/MWKHistoryList.m
@@ -70,6 +70,16 @@ NSString* const MWKHistoryListDidUpdateNotification = @"MWKHistoryListDidUpdateN
     }];
 }
 
+- (void)setSignificantlyViewedOnPageInHistoryWithTitle:(MWKTitle*)title {
+    if ([title.text length] == 0) {
+        return;
+    }
+    [self updateEntryWithListIndex:title update:^BOOL (MWKHistoryEntry* __nullable entry) {
+        entry.titleWasSignificantlyViewed = YES;
+        return YES;
+    }];
+}
+
 - (void)removeEntry:(MWKListEntry)entry {
     [super removeEntry:entry];
     [[NSNotificationCenter defaultCenter] postNotificationName:MWKHistoryListDidUpdateNotification object:self];
--- a/MediaWikiKit/MediaWikiKit/MWKHistoryList.m
+++ b/MediaWikiKit/MediaWikiKit/MWKHistoryList.m
@@ -75,6 +75,9 @@ NSString* const MWKHistoryListDidUpdateNotification = @"MWKHistoryListDidUpdateN
         return;
     }
     [self updateEntryWithListIndex:title update:^BOOL (MWKHistoryEntry* __nullable entry) {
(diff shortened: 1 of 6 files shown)


coreyfloyd:
@bgerstle this is failing because of the search fetcher tests again

bgerstle:
Just to be sure, this has the latest from master? If so, disable that class
in the scheme again. So bizarre that they're failing, won't reenable until
I have a confirmed fix for the issue (I.e. can reliably repro).

bgerstle:
@coreyfloyd i confirmed this had my other CI fixes, so I created #279. Once that's merged, you can merge master and the tests will re-run.

MediaWikiKit/MediaWikiKit/MWKHistoryList.m
```diff
@@ -70,6 +70,16 @@ - (void)setPageScrollPosition:(CGFloat)scrollposition onPageInHistoryWithTitle:(
     }];
 }
 
+- (void)setSignificantlyViewedOnPageInHistoryWithTitle:(MWKTitle*)title {
+    if ([title.text length] == 0) {
+        return;
+    }
+    [self updateEntryWithListIndex:title update:^BOOL (MWKHistoryEntry* __nullable entry) {
+        entry.titleWasSignificantlyViewed = YES;
```
bgerstle: if you return `NO` here, it won't re-save the list, e.g. if the entry was already marked as significantly viewed.

Wikipedia/UI-V5/WMFArticleContainerViewController.m
```diff
@@ -452,6 +459,27 @@ - (CGFloat)totalProgressWithArticleFetcherProgress:(CGFloat)progress {
...
+#pragma mark - Significantly Viewed Timer
+
+- (void)startSignificantlyViewedTimer {
+    if(self.significantlyViewedTimer){
+        return;
+    }
+    MWKHistoryList* historyList = self.dataStore.userDataStore.historyList;
+    MWKHistoryEntry* entry      = [historyList entryForTitle:self.articleTitle];
+    if (!entry.titleWasSignificantlyViewed) {
+        self.significantlyViewedTimer = [NSTimer scheduledTimerWithTimeInterval:FBTweakValue(@"Home", @"Related items", @"Required viewing time", 30.0) target:self selector:@selector(significantlyViewedTimerFired:) userInfo:nil repeats:NO];
```
bgerstle: missing cancel/resume on `viewDidDisappear`/`viewDidAppear`, which means this will always be marked as significant since `NSTimer` retains its target—which also means our VCs are living for however long the timeout is.

bgerstle:
:-1: please address [my comment](https://github.com/wikimedia/wikipedia-ios/pull/277#discussion_r45800818) about not marking articles you've navigated away from as significantly read.  you can verify the problem by setting a breakpoint in the callback and seeing it fire after popping the article VC or tapping a link to another article (or any other navigation that makes its view disappear).

coreyfloyd:
@bgerstle building…

PR #278, opened 2015-11-24
Fix for iOS8 cell layout bugs.

montehurd:
See phab ticket for repro steps and more screenshots:
https://phabricator.wikimedia.org/T119437

**Before**
screen shot 2015-11-23 at 2 45 24 pm
screen shot 2015-11-23 at 3 51 52 pm

**After**
screen shot 2015-11-23 at 4 03 05 pm
screen shot 2015-11-23 at 4 03 22 pm

### DIFF
--- a/Wikipedia/Categories/UITableViewCell+WMFLayout.m
+++ b/Wikipedia/Categories/UITableViewCell+WMFLayout.m
@@ -1,0 +1,15 @@
+//  Created by Monte Hurd on 11/23/15.
+//  Copyright (c) 2015 Wikimedia Foundation. Provided under MIT-style license; please copy and modify!
+
+#import "UITableViewCell+WMFLayout.h"
+#import "NSProcessInfo+WMFOperatingSystemVersionChecks.h"
+
+@implementation UITableViewCell (WMFLayout)
+
+- (void)wmf_layoutIfNeededIfOperatingSystemVersionLessThan9_0_0 {
+    if ([[NSProcessInfo processInfo] wmf_isOperatingSystemVersionLessThan9_0_0]) {
+        [self layoutIfNeeded];
+    }
+}
+
+@end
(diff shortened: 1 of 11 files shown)


montehurd:
Hey I threw a WIP tag on this - open to suggestions for better approach...

montehurd:
@bgerstle hmm on my local machine tests fail on the MWKHistoryListPerformanceTests measure block... any ideas?

Wikipedia/Categories/UITableViewCell+WMFLayout.m
```diff
@@ -0,0 +1,15 @@
...
+//  Copyright (c) 2015 Wikimedia Foundation. Provided under MIT-style license; please copy and modify!
+
+#import "UITableViewCell+WMFLayout.h"
+#import "NSProcessInfo+WMFOperatingSystemVersionChecks.h"
+
+@implementation UITableViewCell (WMFLayout)
+
+- (void)wmf_layoutIfNeededIfOperatingSystemVersionLessThan9_0_0 {
+    if ([[NSProcessInfo processInfo] wmf_isOperatingSystemVersionLessThan9_0_0]) {
+        [self layoutIfNeeded];
```
coreyfloyd: Can't we just call `layoutIfNeeded` always? Or to ask another way, what is the downside of calling `layoutIfNeeded` in iOS 9?

montehurd: Hmm I wasn't sure about downside - they've changed something (subtle?) about how constraints are taken into account between 8 and 9, I think, given that these bugs don't happen in 9. I thought scoping the layoutIfNeeded to 8 would be super-safe... I'm fine with removing the version check if you guys are. One nice thing about leaving it in, though - we would be we'd be notified that it's ok to remove these when we drop 8... wmf_isOperatingSystemVersionLessThan9_0_0 has a `WMF_DEPRECATED_WHEN_DEPLOY_AT_LEAST_9` check.

bgerstle:
:+1: i'm fine with this as-is. this is just more evidence that variable-height cells in iOS 8 and below is just hacky. [0] let's keep these hacks explicit and flagged w/ deprecation-triggering macros so we can wipe the slate clean when we drop iOS 8.

0: http://stackoverflow.com/questions/18746929/using-auto-layout-in-uitableview-for-dynamic-cell-layouts-variable-row-heights/18746930#18746930

PR #280, opened 2015-11-25
Create wikimedia project

taefinal54:


bgerstle:
Hi @taefinal54! Is there something specific you're hoping to accomplish?  Let us know if we can help, otherwise this PR doesn't seem to add anything as-is.

PR #281, opened 2015-11-25
Request multiple random articles so we can show best one.

montehurd:
Best one being non-disambig, with longest extract and hopefully
an image.

Needed to request multiple items because the api doesn't presently
let you request non-disambig random items. Filed ticket with
Discovery team for this.

Need similar fix for "read more like" items...


bgerstle:
TIL Travis fails when you merge before tests start running. Disregard this error, as the commit only added comments

PR #282, opened 2015-11-26
Fix for article language button crash.

montehurd:


montehurd:
@bgerstle feel free to tweak and/or merge! :)

bgerstle:
@montehurd nice job! this will fix the issue, but this also looks like a great opportunity for adding tests if you want to give that a go.  There are basically two groups of responsibilities in here that need to be tested:

- Managing persistent language preferences (all the `NSUserDefaults` stuff)
  - Not duplicating languages when they're saved (e.g. if a language was in `[NSLocale preferredLanguages]`) 
  - inserting (or moving) new selections to the front of the list
  - exclusion of "unsupported" languages
- Filtering & partitioning languages
  - putting languages that were previously selected in the `filteredPreferredLanguages` property
  - putting everything else in the `filteredOtherLanguages` property
  - ensuring both are filtered when `languageFilter` is changed

These tests would have very easily caught this crash by detecting the unexpected `NSNull` objects we were adding to `filteredPreferredLanguages`.  To be fair, we should've also manually tested this functionality before merging the search language patch since it touched this class—especially given the lack of unit tests.

bgerstle:
@montehurd i lied, turns out `MWKLanguageLinkControllerTests` already exists. can you add an additional test there?

bgerstle:
this "edge" case would be that the language links we set on the control doesn't intersect with `preferredLanguages`.  currently those tests all use the static lang data, so you might need to inject some language links as part of the test

coreyfloyd:
@bgerstle @montehurd I have to change the API of the language link controller for my patch so probably not worth adding tests here now. We probably also want to prioritize the items in the ready for dev column over additional tests as we are already going to miss our beta deadline of Tuesday.

coreyfloyd:
:+1:

bgerstle:
@coreyfloyd spent a few minutes merging master & adding regression tests, both for the failing case and for sanity checks on filtering as well

bgerstle:
simulator stall? re-building...

bgerstle:
I hope this doesn't become another :snowflake: ...

bgerstle:
w00t @montehurd @coreyfloyd have at it

coreyfloyd:
:+1:

PR #283, opened 2015-11-30
Fix progress not updating until all network activity is finished.

coreyfloyd:
Needed to use dispatch queues to make this work. See comment in code for explanation.

### DIFF
--- a/Wikipedia/UI-V5/WMFArticleContainerViewController.m
+++ b/Wikipedia/UI-V5/WMFArticleContainerViewController.m
@@ -448,7 +448,14 @@ NS_ASSUME_NONNULL_BEGIN
 }
 
 - (void)updateProgress:(CGFloat)progress animated:(BOOL)animated {
-    [self.progressView setProgress:progress animated:animated];
+    /**
+     *  Progress updates must be dispatched using GCD in order for them to show up during network activity.
+     *  This is because AFNetworking dispatches progress updates using GCD. And in order for the UI updates to be interweaved with the networking updates,
+     *  they must be queued as well. If not, then progress updates are deferred until after the disaptch queue is empty.
+     */
+    dispatchOnMainQueue(^{
+        [self.progressView setProgress:progress animated:animated];
+    });
 }
 
 /**


bgerstle:
Tried this on WiFi NLC enabled and clean install, the progress indicator remained at 80% until the article _finally_ finished rendering in the webview.[0]  I'm pretty sure the payload didn't download instantaneously, maybe there's something else going on?  Also, re-dispatching to the main queue doesn't make sense to me, given that the entire app would be unresponsive if the queue was "blocked" by networking.

0: https://vid.me/BMnB

coreyfloyd:
@bgerstle

> Tried this on WiFi NLC enabled and clean install, the progress indicator remained at 80% until the article finally finished rendering in the webview.[0] I'm pretty sure the payload didn't download instantaneously, maybe there's something else going on? 

Two things:

1. Barack Obama is about as intense of an article to render as you can pick, so its going to take the longest to render. In general, the larger the article, the more complex the HTML, the more images -> the larger the proportion of rendering time to download time.

2. Wifi is also about the fastest you can make network conditioner run (unless you are running gigabit ethernet through your lightening port). In general article downloads are pretty quick over wifi, even Barack Obama probably downloaded < 1/2 second. If you want to really observe progress updates you probably should try a slower setting (Edge/3G).

So I'm not surprised by the example: The Barack Obama article on Wifi maximizes rendering time and minimizes download time.

TL;DR
Really the core problem - accurate progress - requires us to move WKWebView and to let the web view request the download (likely intercepted by a URLProtocol so we can cache the structured data) and let the web view update the progress using the official API.

> Also, re-dispatching to the main queue doesn't make sense to me, given that the entire app would be unresponsive if the queue was "blocked" by networking.

Note: I never use the word "blocked" in the explanation. Updates aren't blocked, but the progress view won't re-draw without re-dispatching the update progress calls. You can test by simply running the app without my changes (adding break points and enabling the network conditioner if you want to see the blow-by-blow).

TL;DR 
Basically we are working around some implementation details of Cocoa: run loop modes, run loop source processing, UIKit drawing, networking threading, and dispatch queues along with how AFNetworking uses those APIs and chooses to propagate its updates. 

If a block is waiting in the main dispatch queue and a draw call is made to a UIKit component, which is processed first? 
Are dispatch queues processed as a run loop source? If so how are they are prioritized? If not, are they processed before or after run loop sources?
Does it depend on which run loop mode you are in? 
When are UIKit controls updated? Is it immediately? Is it the same run loop? Is it the next run loop? Do any use GCD internally? Does it depend on if there drawing is implemented using Core Animation Layers or Draw Rect? Does it matter if the updates are being animated?

How dispatch queues interact with the old threading models and run loop modes isn't really mentioned anywhere in official documentation. When UIKit controls update is mentioned a bit, but its not definitive from the documentation. Others have tried to reverse engineer it or postulate about it: 
http://lapcatsoftware.com/articles/dispatch-queues-and-run-loop-modes.html
http://bou.io/RunRunLoopRun.html

Practically, devs often do something similar to what I did if drawing updates seem to get postponed - it seems to happen more often with progress views, but also occurs when updating labels and other controls. The fix isn't always straight forward, usually its using a dispatch on main queue or a perform selector on main queue, possibly in a specific run loop mode. 

You can see lots of convoluted answers basically suggesting similar things when searching for problems like: 
https://www.google.com/search?client=safari&rls=en&q=uiprogressview+update+in+loop&ie=UTF-8&oe=UTF-8#q=uiprogressview+not+updating

I'm sure there are other solutions, and most are likely working around the same issues in a different way.

bgerstle:
@coreyfloyd I posted the video w/o NLC (your quote was from a older "revision" of my comment) because I had tried with NLC and noticed the same behavior in both cases:

- [with dispatch async on 3G](https://vid.me/MtyO)
- [without dispatch async on 3G](https://vid.me/CE1G)

I chose the Obama article because it's one I thought of off the top of my head w/ a large payload. Is there another article or NLC setting I can use to verify the bug & fix?

coreyfloyd:
@bgerstle hmmm… very odd - I just checked master too - and it seems to work for me now as well… though I swear I confirmed the issue when @nirzar reported the issue. 

At this point I can't say whether it is just a trick of the eye when you have a fast internet connection or if it is a real but intermittent problem (maybe depending on the speed that the packets are delivered or something else).

i can check with @nirzar to have him verify and see what he wants to do. If it is just an trick of the eye… meaning that downloads are so fast that it fills up to 80% very quickly… maybe we should make it only fill to 50% and just fake the last 50% on a timer.

bgerstle:
Also, I agree about the real fix for accurate progress is moving to `WKWebView`, which we're thinking about doing anyway, so we don't need to spend a lot of time evaluating workarounds for this.  I'm just trying to understand what the problem was, as I couldn't really reproduce it.

bgerstle:
@coreyfloyd if it works (mostly?) on master or is intermittent, and we both agree the real fix for this (and potentially other issues) is to switch to `WKWebView`, what do you think we should do w/ this change?

I have two other theories (which are not mutually exclusive):

We're setting the progress to 0.0 "completing" the download at 80%:

> 2015-12-01 12:33:44:717 Wikipedia Debug[main] -[WMFArticleContainerViewController updateProgress:animated:]#L451 DEBUG:
Progress: 0.592000
2015-12-01 12:33:44:719 Wikipedia Debug[main] -[WMFArticleContainerViewController updateProgress:animated:]#L451 DEBUG:
Progress: 0.000000
2015-12-01 12:33:44:719 Wikipedia Debug[main] -[WMFArticleContainerViewController updateProgress:animated:]#L451 DEBUG:
Progress: 0.800000

Not sure where that 0 is coming from or if it could pose a problem, but seems unintentional.

My other theory is that progress updates are cancelling each other when they happen in quick succession.  This isn't something we can easily test for w/ the `UIProgressView` API, but my hunch is that if it had a completion block (like `UIView`) API's, the `finished` parameter would be `NO`, since the animated property was changed before the animation completed.

Could be a combination of both these things, or potentially something related to how run loops and dispatch queues interoperate.  Do you think run loop & dispatch queue interop (if that's the underlying issue) could affect anything else?

coreyfloyd:
@bgerstle have you produced that errant "0.00" progress more than once?

bgerstle:
Going through and reading some of the sources you cited, I found:

http://lapcatsoftware.com/articles/dispatch-queues-and-run-loop-modes.html

Says that blocks invoked on the main queue run in the "common" modes, which is always the case unless there's a special UIKit mode (e.g. event tracking). If you wanted a custom mode to be run all the time you need to add it to the common modes via `CFRunLoopAddCommonMode`.

The RunRunLoopRun article is a great one, which also points out that dispatch queues (at least the main one) are piped directly into the run loop.  There could be something in there but it's hard to tell w/o a specific set of repro steps for a runloop-related issue.

Regarding the SO posts, the two problems I found with upvoted answers [0][1] are basically (AFAICT) "always call `setProgress:` on the main thread," wherein the common mistake is to call `UIKit` w/in your completion block which happens to be executing on a background thread.  AFNetworking calls everything on the main thread for us, which is the main use case for most completion blocks.  In particular, I'm not seeing anything that references re-dispatching as a workaround.  Was there a specific answer you were referencing?

0: http://stackoverflow.com/questions/11226509/uiprogressview-not-updating
1: http://stackoverflow.com/questions/8803189/setprogress-is-no-longer-updating-uiprogressview-since-ios-5

bgerstle:
let me investigate the "0.0 before 80%" thing..

bgerstle:
@coreyfloyd appears to be intermittent w/ a warm cache but new VC (i.e. not refreshing in place). could be that we're loading the article data from cache, then setting progress to 80%, then downloading the latest version (setting progress back to 0). some kind of race condition between those two?

bgerstle:
@coreyfloyd you can test by going to/from an article and recent pages (i.e. always pop the VC off the stack to start w/ a fresh `UIProgressView`)

bgerstle:
I just got it on a "cold" article load too (remove Obama from History, restart the app to clear object cache, and load Obama again via Search):

> 2015-12-01 13:01:29:864 Wikipedia Debug[main] -[WMFArticleContainerViewController updateProgress:animated:]#L453 WARN:
Unexpected progress regression from 0.496000 to 0.000000

coreyfloyd:
Closing for now in favor of fixing the 0 progress issue first and then revisiting design considerations with @nirzar

PR #285, opened 2015-11-30
Home --> Explore

nirzar:
+  removing the translucent tabbar

https://phabricator.wikimedia.org/T115049


bgerstle:
@coreyfloyd i had brought up similar concerns in design review. I suggest you talk to him about it in detail (not via PR comments) but (@nirzar correct me if I'm wrong) the gist was: translucency interferes w/ readability (e.g. Twitter and Facebook both use opaque bars last I checked).

However, I do agree if we're going to do this it should be for all bars, not just the tab bar.

bgerstle:
@coreyfloyd @nirzar spent a minute addressing comments. if there aren't any issues blocking this design (which you two can sort out separately) this is good to go IMO :+1: 

image

nirzar:
@bgerstle Yes, this looks good! that was one of the things i mentioned in the design blueprint.
translucent bars implies certain material. plastic, glass etc. and we are setting a paper and organic direction in the visual design. sorry to have this conversation here. let's talk separately.

PR #289, opened 2015-12-01

niedzielski:


bgerstle:
:+1: lgtm will merge when Travis re-confirms

bgerstle:
@montehurd did you want to give this another look?

bgerstle:
@niedzielski also, nice work on an 800+ LOC contribution to an iOS project :stuck_out_tongue_winking_eye:

bgerstle:
> No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

noooo!!

montehurd:
@bgerstle lol

montehurd:
@niedzielski hey just a side-note, but in case you closed/reopened to restart travis, you can also restart it by tapping "details" link, then there's a round restart button near right side of the details page.

niedzielski:
Looks like everyone's builds are hanging on `Touching WikipediaUnitTests.xctest`.

niedzielski:
@montehurd thanks! Travis wanted me to sign in with GitHub so I just did it the old fashioned way :)

montehurd:
@niedzielski so weird!

montehurd:
@niedzielski ah makes sense :)

montehurd:
Maybe related? https://github.com/travis-ci/travis-ci/issues/4906

bgerstle:
@montehurd if you scroll to the bottom of that issue, you'll see i've referenced it in some of the commits I've already applied here to work around simulator timeouts.  I don't believe this is related to simulator timeouts, but I'm not sure what else it could be.

montehurd:
IT PASSED!

bgerstle:
i'm tempted to restart it, just to see...

bgerstle:
> The log length has exceeded the limit of 4 MB (this usually means that the test suite is raising the same exception over and over).

montehurd:
Interesting... it passed but not the first time. Had to restart it once.

bgerstle:
@montehurd IMO that means there's no difference. I tried something similar to your strategy with 9ae19125. Removing pre-build starting of sim might re-introduce simulator timeouts as well.

montehurd:
@bgerstle ya i'm grasping at straws at bit :)

bgerstle:
yeah, unfortunately we're not able to get into a box and run stuff manually, and this particular failure means we can't even get error logs—build is silent and just aborted altogether.  i think the last thing to try it stuff w/ the project & groups to see if something's messed up (e.g. that last commit you mentioned).  i've noticed that creating new files in groups doesn't automatically point to the "Code" folder... not sure if that could be related.

montehurd:
Piping to /dev/null still shows failing test error messages though... not sure why. I think once we get this figured out that would be a bit more useful than the mega console spam dump.

montehurd:
Oh yeah! That last commit I mentioned may have screwed something up...

montehurd:
It's so weird though - same xcodebuild command works fine locally...

bgerstle:
Bash redirects are fun: http://tldp.org/HOWTO/Bash-Prog-Intro-HOWTO-3.html

`foo > bar` redirects `stdout` to whatever `bar` is, which can either be a regular file or "special" file like `stderr`.  error output still shows up because you're not redirecting `stderr`.

bgerstle:
@montehurd yeah, definitely something specific to travis, or only exposed on travis. our Jenkins box has been OK

montehurd:
@bgerstle I think I'll toss the ball back to you here :)

montehurd:
Hmm...

    Could not find minitest-5.8.2 in any of the sources
    Run `bundle install` to install missing gems.

bgerstle:
hm, that's a new one, probably just a temporary issue w/ RubyGems. happens rarely

niedzielski:
I was under the impression this was happening for other patches. Is this unique to my PR?

bgerstle:
@niedzielski not unique, no. we'll split this out to save you from the noise

bgerstle:
it does seem particularly bad for this one though (other PRs have been less problematic for some reason)

bgerstle:
nuked it down to the localization updates, will merge just those so we can put it behind us

PR #295, opened 2015-12-02
Make certain feed headers respond to taps.

montehurd:
- [x] "Continue reading..." (tap should load the article)
- [x] "Featured article" (tap should load the article)
- [x] "Today on Wikipedia" (tap should load the article)
- [x] "Random" (tap should load the article - but should not prevent refresh from working)
- [x] "Nearby" (tap should load full Nearby list)
- [x] "Today's picture" (tap should load the picture)
- [x] Make header taps use same highlighting - (Josh said it's ok for now if the cell beneath the header highlights)
- [x] Bypass this tap handling for other header types
- [x] "Because you read..." (no longer make only the link tappable - make tapping *anywhere* on the "Because you read..." header load the article, the link can still be highlighted, it just won't handle taps)


montehurd:
@bgerstle maybe overlaps a tad with your gallery bits (making "Today's picture" header tap-able).

coreyfloyd:
Headers with links in them already open the article. This is done by the attributed string. Also is there a ticket for this?

montehurd:
Headers with links in them already open the article. This is done by the attributed strings.
This doesn't affect those. 

    Also is there a ticket for this?
Ya! See branch name.

coreyfloyd:
@montehurd I didn't see the ticket because it's not in scope for beta. Ticket comment from Josh confirms this:
> Yeah, this will add a bit of inconsistency in the feed item behavior but I've also heard/experienced the expectation that header seems like it should be tappable. Its fine to leave in Ready for Dev, but lets leave the beta tag off this for now.

Please be sure you are only pulling work that is tagged for beta - especially since we are trying to not introduce any more scope this week before we release the beta.

> Headers with links in them already open the article. This is done by the attributed strings.
>> This doesn't affect those.

If we are changing headers to be completely tappable instead of relying on a link, then we should be removing the url link opening from the attributed string. Basically, if we want to refactor/build on this functionality, we should remove/refactor the old functionality.

montehurd:
@coreyfloyd I appreciate your concern, but I asked Josh if it was ok to knock this one out.

> If we are changing headers to be completely tappable instead of relying on a link, then we should be removing the url link opening from the attributed string.

Oh, we are not changing headers which have links to not be tappable. The link behavior, per Josh, is to stay the same for now. The ticket describes which headers were to change.

coreyfloyd:
> @coreyfloyd I appreciate your concern, but I asked Josh if it was ok to knock this one out.

Ok - but the ticket still says otherwise - Josh clearly stated not to tag it for beta. If you had a conversation with Josh after he made that comment, then the ticket should be tagged with beta so its clear to everyone its in scope for the beta.

coreyfloyd:
>> If we are changing headers to be completely tappable instead of relying on a link, then we should be removing the url link opening from the attributed string.

> Oh, we are not changing headers which have links to not be tappable. The link behavior, per Josh, is to stay the same for now. The ticket describes which headers were to change.

Its fine that we still want links - but the tap action should be consistent. If you do this, tapping the links will still works as Josh mentioned, since the link is within the header. This way the related headers get the same behavior as the other headers… and have a larger tap area which is good thing.

Basically this is a maintenance issue - if we want to support taps on headers, we should support it in a single consistent way.

bgerstle:
keep in mind, we talked about removing links since we don't want to support peeking from headers

montehurd:
Ya that's part of the reason why this ticket didn't change the "because you read" headers.

bgerstle:
@montehurd to be clear, i was thinking we _should_ change "Because you read" since we should be using a button instead of a link to prevent peek-ability

montehurd:
Ah, I see. I was thinking that was a separate bit, so separate ticket, but I guess I'm in the minority there :) No worries! It should be quick.

montehurd:
I'll wait until your PoTD bits are merged too so I can hook up the "Today's picture" header tap.

montehurd:
@coreyfloyd I updated the checklist above with a new "Because you read.." item at the bottom - can you peek and see if it looks ok?

coreyfloyd:
@montehurd FYI - your git configuration seems messed up - your wikimedia email is not being associated with your commits. You should check your git config and make sure your email is in there. You may need to configure it for just the wikipedia repo. 

https://help.github.com/articles/setting-your-email-in-git/

Wikipedia/Code/WMFHomeViewController.m
```diff
@@ -472,6 +466,37 @@ - (void)didTapFooterInSection:(NSUInteger)section {
...
+        case WMFHomeSectionTypeMainPage:
+        case WMFHomeSectionTypeFeaturedArticle:
+        case WMFHomeSectionTypePictureOfTheDay:
+        case WMFHomeSectionTypeRandom:
+            [self selectFirstRowInSection:section];
+            break;
+        case WMFHomeSectionTypeNearby:
+            [self didTapFooterInSection:section];
+            break;
+        case WMFHomeSectionTypeSaved: {
```
bgerstle: why are we only doing this for saved, and not history?

Wikipedia/Code/WMFHomeViewController.m
```diff
@@ -472,6 +466,37 @@ - (void)didTapFooterInSection:(NSUInteger)section {
...
+            WMFRelatedSectionController* controller = (WMFRelatedSectionController*)[self sectionControllerForSectionAtIndex:section];
+            [self wmf_pushArticleViewControllerWithTitle:controller.title
+                                         discoveryMethod:MWKHistoryDiscoveryMethodLink
+                                               dataStore:self.dataStore];
+            break;
```
bgerstle: turns out we don't have a best practice on this, but related to my comment above, we should (in general) avoid `default` cases in our `switch` statements, unless it's actually the intention to have a global fallback.  in this case, we should remove it and force that all cases (now and future) are listed in the `switch`.

Wikipedia/Code/WMFHomeViewController.m
```diff
@@ -472,6 +466,37 @@ - (void)didTapFooterInSection:(NSUInteger)section {
...
+        }
+        default:
```
bgerstle: turns out we don't have a best practice on this, but related to my comment above, we should (in general) avoid `default` cases in our `switch` statements, unless it's actually the intention to have a global fallback.  in this case, we should remove it and force that all cases (now and future) are listed in the `switch`.

bgerstle:
Please add the case for history sections as I mentioned in the diff.  Also, I can't really peek headers or see the selection state because pressing for a long time results in the text selection loupe appearing.  We should remove attributed strings in headers and/or swap `UILabel` for `UITextView`.

From a code standpoint, having a centralized switch statement to differentiate behaviors is a bit of a code smell to me.  This sort of pattern in particularly lends itself well to protocols or other form of polymorphism  (e.g. inheritance) to specialize the behaviors.  Refactoring this now might not be easy due to the current home section controller architecture (given that HomeVC implements a lot of the actual tap handling), so we can address in a separate ticket/discussion.

Wikipedia/Code/WMFHomeViewController.m
montehurd: The feed doesn't show history items.

Wikipedia/Code/WMFHomeViewController.m
montehurd: See above.

Wikipedia/Code/WMFHomeViewController.m
montehurd: My bad. Fixed.

bgerstle:
:+1: travis seems to be hanging atm, but the tests pass locally :white_check_mark:

bgerstle:
@coreyfloyd anything to add?

PR #296, opened 2015-12-02
Increase random refresh button touch area.

montehurd:


montehurd:
**Before:**
screen shot 2015-12-02 at 3 57 19 pm

**After:**
screen shot 2015-12-02 at 3 56 49 pm

PR #297, opened 2015-12-03
Adding welcome on boarding 

coreyfloyd:
Adds welcome screens

This includes: 
- a container welcome VC that transitions between each welcome page
- a welcome page listing new items
- a welcome page for picking languages that allows re-ordering and deletion
- a welcome page for enabling analytics/crash reporting

For preferred language reordering and forthcoming deletion, the language classes were refactored:
- LanguagelinkController now manages OS languages and preferred languages
- An MWKLanguageFilter was extracted to handle filtering
- An MWKTitleLanguageController was extracted to handle languages for specific articles.
- The Language picker was refactored to allow the hiding of preferred languages

Other small refactors song the way.


montehurd:
I think WMFBoringNavigationTransition files are in the root dir...

Wikipedia/Localizations/en.lproj/Localizable.strings
```diff
@@ -332,3 +332,28 @@
 "home-nearby-footer" = "More from nearby your location";
 "home-nearby-nothing" = "Nothing nearby";
 "home-nearby-check-again" = "Check again";
+
+"welcome-whats-new-title" = "Welcome To Wikipedia";
+"welcome-whats-new-whats-new" = "WHAT'S NEW";
```
montehurd: Have gotten push-back from TWN in past for using all caps - they said this made translation to certain languages trickier and to use normal casing. If we need all caps we can then upper case just before display for English only.

montehurd:
Guessing we'll want to limit this to portrait only? In landscape stuff gets funky:

screen shot 2015-12-02 at 6 35 16 pm

screen shot 2015-12-02 at 6 35 57 pm

Wikipedia/Localizations/en.lproj/Localizable.strings
coreyfloyd: @montehurd Improved the code comment to explain why. Lets see if that works for them. This way we get upper case capitalizations in translations if they are appropriate for a language. If we still get blow back we can go back to doing it in code.

montehurd: Sounds good!

coreyfloyd:
:-1:  

I'm getting a crash in 8.4 when set stack view deallocates… doh

coreyfloyd:
:+1:  Ok bug fixed in the pod. All should be well.

coreyfloyd:
Unfortunately this comes with pod updates… probably best to review commit by commit

bgerstle:
@coreyfloyd @montehurd some thoughts on usage of BlocksKit. this area of the code was recently the origin of an insta-crash due to `bk_map` yielding an `NSNull` object.  To cope with ObjC's weak type system (IOW, this isn't a problem in Swift), we might want to put our own measures in place. We can either:

- Do it ourselves
- Fork [this PR](https://github.com/zwaldowski/BlocksKit/pull/288) to make it merge-able, and potentially use it in our project

In either case, I'm thinking we should have a `strictMap` which asserts on `nil` values returned by the `bk_map` block parameter, and drops them "in production." IOW, behaves like `wmf_mapAndRejectNil:` but raises an assertion if `nil`.

As a stopgap, we can add a `strictMap` category on `NSArray` only (instead of trying to support all the "collection" types from the get go) and look into updating BlocksKit later.

bgerstle:
As an aside, we could also update BlocksKit to use lightweight ObjC generics.

montehurd:
Looks like this will need to be on a scroll view for small screens and/or long localized strings:

screen shot 2015-12-07 at 2 14 10 pm

bgerstle:
@montehurd sorry, it's hard to tell what the problem is from that screenshot, is anything actually cut off on 4s?

montehurd:
@bgerstle added reference screenshot.

bgerstle:
Oh, so that's not a strikethrough you added... got it. Shouldn't be hard to fix that

montehurd:
>Shouldn't be hard to fix that

I'm just as concerned about, say, the German version of "Updated design focused on..." string being, say, 6 lines...

coreyfloyd:
Ahh like like the button height priority needs to be set higher

coreyfloyd:
In case it wasn't obvious, it's not a strike through its the button border being squished.

montehurd:
One more (maybe same cause?)

screen shot 2015-12-07 at 2 53 10 pm

bgerstle:
> I'm just as concerned about, say, the German version of "Updated design focused on..." string being, say, 6 lines...

Yeah, this is especially true if the user cranks up dynamic type (if/when we support it).  I say we go with this for now (given that it won't affect most beta users and we don't have any copy for these views in other languages).  In the meantime, we can get design to think about how to handle longer localizations & dynamic type.

montehurd:
@bgerstle agreed. before launch we should an audit where we modify MWLocalizedString to double up the string it returns. then we can just look at every interface for un-graceful handling of loooong localizations.

montehurd:
@coreyfloyd i was testing swipe to delete on the language list and it seems to only let me swipe one of the 3 entries i have (english, test wiki and hebrew)...

bgerstle:
@coreyfloyd when adding a new language, and the list is long, we should scroll to the bottom, otherwise you can't see the "add languages" button:

image



@montehurd you can only delete languages that aren't part of your OS-selected languages (i.e. `[NSLocale preferredLanguages]`

bgerstle:
scroll would also allow this to support landscape orientations, if we wanted to

bgerstle:
:+1:

coreyfloyd:
@montehurd it doesnt let you delete your OS languages. currently there is no indicator to show if a language is deletable or not

montehurd:
@coreyfloyd gotcha thx!

bgerstle:
tests passed on my machine :+1:

PR #298, opened 2015-12-03
Minor build & project tweaks

bgerstle:
- update-qqq will be run outside of build pipeline (maybe as a jenkins job w/ TWN sync one day)
- reduce project down to 6 warnings!!... by wrapping deprecation warnings & TODOs! (y/w @coreyfloyd)
  - when `WMF_SHOW_TECH_DEBT_WARNINGS` is enabled, we get deprecation & TODOs like:
image
  - otherwise... silence


montehurd:
Interesting about the location setting and file creation - I think Corey ran in to that.

montehurd:
We've moved lots of things around lately - will prob want to test that grunt and web scripts still do their thing...

bgerstle:
@montehurd still works! (2e32e98) makes sense, given that we didn't move the assets or www folders

bgerstle:
if any web stuff crops up we can revert the commit, but figured might as well update now

coreyfloyd:
:+1:

PR #299, opened 2015-12-04
Fix nearby cell updates

bgerstle:
Phab: [T120018](https://phabricator.wikimedia.org/T120018)

---

See testing notes in Phab task.

### DIFF
--- a/Wikipedia/Code/WMFHomeSectionSchema.m
+++ b/Wikipedia/Code/WMFHomeSectionSchema.m
@@ -43,6 +43,12 @@ static NSString* const WMFHomeSectionsFileExtension = @"plist";
 
 @implementation WMFHomeSectionSchema
 
+- (NSString*)description {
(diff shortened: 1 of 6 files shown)


coreyfloyd:
We probably should turn on and off when cells are off screen though and when the home view is off screen as well.

bgerstle:
> We probably should turn on and off when cells are off screen though and when the home view is off screen as well.

Already implemented. The HomeVC stops the location manager. The list view also does this automatically by calling `stopUpdates` on its dynamic data source.

bgerstle:
I don't have time this second to investigate when the `stopUpdates` call was added in the completion block, but we should add some tests to verify nearby update behavior. I'm updating the spec doc now.

bgerstle:
Ah, the intent was probably to make the _list_ "static." I'll see what I can do here.

coreyfloyd:
:+1: 

@montehurd you want to review and merge?

PR #300, opened 2015-12-04
Add loading indicator and blur to random card when refreshing it.

montehurd:


Wikipedia/Code/WMFArticlePreviewTableViewCell.m
```diff
@@ -45,6 +60,28 @@ - (void)awakeFromNib {
...
+- (void)setupBlurViewAndLoadingIndicator {
+    UIBlurEffect* blurEffect = [[UIBlurEffect alloc] init];
+    self.blurView = [[UIVisualEffectView alloc] initWithEffect:blurEffect];
+    [self.contentView addSubview:self.blurView];
+    [self.contentView bringSubviewToFront:self.blurView];
+    [self.blurView mas_makeConstraints:^(MASConstraintMaker* make) {
+        make.edges.equalTo(self.contentView);
+    }];
+
+    self.activityIndicator = [[UIActivityIndicatorView alloc] init];
```
montehurd: I tried adding the activity indicator in IB but there were weird auto layout issues. It was completely breaking the cell layout... not sure why.

Wikipedia/Code/WMFRandomSectionController.m
```diff
@@ -112,6 +112,9 @@ - (void)getNewRandomArticle {
     if (self.fetcher.isFetching) {
         return;
     }
+
+    [self.delegate controllerWillGetItems:self];
```
coreyfloyd: you can just call self.delegate
`- (void)controller:(id<WMFHomeSectionController>)controller didSetItems:(NSArray*)items`

You do't really need the extra method.

Then on line 103 add:

```
if(self.isFetching){
//add blur
}else{
//remove blur
}
```

coreyfloyd: this is because didSetItems causes the configureCell to fire. 

If the animations don't look good let me know - we can work it out.

Wikipedia/Code/WMFArticlePreviewTableViewCell.m
```diff
@@ -30,11 +34,22 @@ @implementation WMFArticlePreviewTableViewCell
 
 #pragma mark - Setup
 
+- (void)blurAndShowLoadingIndicator {
+    self.blurView.hidden = NO;
```
montehurd: Should I maybe only add the blur and loading indicator views just before showing them and remove them after hiding them so we don't have extra views on every cell? Only used by random atm..

Wikipedia/Code/WMFArticlePreviewTableViewCell.m
```diff
@@ -30,11 +34,22 @@ @implementation WMFArticlePreviewTableViewCell
...
+    [self.activityIndicator startAnimating];
+}
+
+- (void)unblurAndHideLoadingIndicator {
+    self.blurView.hidden = YES;
+    [self.activityIndicator stopAnimating];
```
bgerstle: tip: can also show/hide a container view instead of toggling the visibility of a group of subviews. in this case: the spinner is always spinning when shown, and it always needs to be shown on top of the blur. can just group blur & spinner in a container which we show/hide (or even constrain) as needed.

Wikipedia/Code/WMFArticlePreviewTableViewCell.m
```diff
@@ -44,6 +59,28 @@ - (void)awakeFromNib {
...
+
+    [self setupBlurViewAndLoadingIndicator];
+    [self unblurAndHideLoadingIndicator];
+}
+
+- (void)setupBlurViewAndLoadingIndicator {
+    UIBlurEffect* blurEffect = [[UIBlurEffect alloc] init];
+    self.blurView = [[UIVisualEffectView alloc] initWithEffect:blurEffect];
+    [self.contentView addSubview:self.blurView];
+    [self.contentView bringSubviewToFront:self.blurView];
```
bgerstle: is this necessary? adding a subview automatically puts it in the front.

Wikipedia/Code/WMFArticlePreviewTableViewCell.m
```diff
@@ -44,6 +59,28 @@ - (void)awakeFromNib {
...
+    [self.blurView mas_makeConstraints:^(MASConstraintMaker* make) {
+        make.edges.equalTo(self.contentView);
+    }];
+
+    self.activityIndicator = [[UIActivityIndicatorView alloc] init];
+    [self.activityIndicator setTranslatesAutoresizingMaskIntoConstraints:NO];
```
bgerstle: this is unnecessary, masonry already sets this to `NO` when you call any of its make/remake/update methods.

Wikipedia/Code/WMFArticlePreviewTableViewCell.m
montehurd: The blur view *is* that container.

Wikipedia/Code/WMFArticlePreviewTableViewCell.m
montehurd: Ah good catch.

Wikipedia/Code/WMFArticlePreviewTableViewCell.m
```diff
@@ -30,11 +34,22 @@ @implementation WMFArticlePreviewTableViewCell
 
 #pragma mark - Setup
 
+- (void)blurAndShowLoadingIndicator {
```
bgerstle: API tip: these kinds of methods which are "mirror images" of each other and change what is effectively the same state can be refactored into a single property. for example: `@property (nonatomic, assign) BOOL loading`. this will simplify the `if` statement in the cell configuration block to: `cell.loading = self.isFetching`.

Wikipedia/Code/WMFArticlePreviewTableViewCell.m
bgerstle: right, just got to that part :smile: you could probably omit the `start`/`stopAnimating` and let the activity indicator always animate when shown.

Wikipedia/Code/WMFArticlePreviewTableViewCell.m
montehurd: Agreed. I've been a little shy lately about making things properties if they're not just super simple value assignments, but in this case it seems ok.

Wikipedia/Code/WMFArticlePreviewTableViewCell.m
montehurd: http://i.imgur.com/ZDoiyUW.gifv

Wikipedia/Code/WMFArticlePreviewTableViewCell.m
```diff
@@ -44,6 +59,28 @@ - (void)awakeFromNib {
...
+    [self.contentView addSubview:self.blurView];
+    [self.contentView bringSubviewToFront:self.blurView];
+    [self.blurView mas_makeConstraints:^(MASConstraintMaker* make) {
+        make.edges.equalTo(self.contentView);
+    }];
+
+    self.activityIndicator = [[UIActivityIndicatorView alloc] init];
+    [self.activityIndicator setTranslatesAutoresizingMaskIntoConstraints:NO];
+    self.activityIndicator.color = [UIColor blackColor];
+    [self.blurView addSubview:self.activityIndicator];
```
bgerstle: this is interesting... the docs state "Add subviews to the `contentView` and not to `UIVisualEffectView` directly."  this seems to work though? :confounded:

coreyfloyd: Should be added to the content view for sure

Wikipedia/Code/WMFArticlePreviewTableViewCell.m
coreyfloyd: Hmm… did you just center it vertically and horizontally? Activity indicators have a specified size, so trying to do anything that makes them bigger/smaller will break.

Wikipedia/Code/WMFArticlePreviewTableViewCell.m
montehurd: opps! easy fix...

Wikipedia/Code/WMFArticlePreviewTableViewCell.m
montehurd: yes. it was really weird.

montehurd: i can try again...

Wikipedia/Code/WMFArticlePreviewTableViewCell.m
coreyfloyd: setting `hidesWhenStopped` causes it to hide when stopped. Not sure if the logic works the other way around - it may not redrawing because its subview is hidden - but it may be running some other logic if Apple isn't checking the hidden property.

Wikipedia/Code/WMFArticlePreviewTableViewCell.m
coreyfloyd: you don't have too… was just curious if you constrained the hight or width, or leading or trailing, etc… anything that would have stretched it.

Wikipedia/Code/WMFArticlePreviewTableViewCell.m
bgerstle: @montehurd what i've done in these cases is just expose a setter method (no `@property`), since "getting" the value doesn't really make sense (nor is it usually necessary).

Wikipedia/Code/WMFArticlePreviewTableViewCell.m
bgerstle: @montehurd what i've done in these cases is just expose a setter method (no `@property`), since "getting" the value doesn't really make sense (nor is it usually necessary).

bgerstle:
@montehurd overall looks good! only two amendments i'd request:

- refactor the `blur`/`unblur` into a single method which takes a single `BOOL` argument for the "blur/loading" state
- to a cross dissolve (or animate the alpha) when showing the blur/loading view
- **extra credit** defer the loading state for 250ms. I've [already done this in the gallery](https://github.com/wikimedia/wikipedia-ios/blob/master/Wikipedia/Code/WMFImageGalleryCollectionViewCell.m#L234). you can refactor it into a utility if you want, then reuse it here.  this will prevent the loading view from "blinking" on a fast connection.

bgerstle:
sorry, make that three amendments (for @coreyfloyd)

- add loading indicator to visual effect view's `contentView`

Wikipedia/Code/WMFArticlePreviewTableViewCell.m
montehurd: Removed.

Wikipedia/Code/WMFArticlePreviewTableViewCell.m
montehurd: Fixed.

Wikipedia/Code/WMFArticlePreviewTableViewCell.m
montehurd: Done!

Wikipedia/Code/WMFArticlePreviewTableViewCell.m
montehurd: Fixed.

coreyfloyd:
:+1:

bgerstle:
IMO the transition is a bit quirky right now because it's not animated (the `WMFHomeSectionController` abstraction/indirection makes this tricky) and the entire feed shifts when the height of the cell changes. we can look into refining this in a separate PR.

PR #302, opened 2015-12-07

niedzielski:


montehurd:
These updates look great! qqq looks good - nothing disappeared.

PR #303, opened 2015-12-07
Re-add login to settings screen

montehurd:
https://phabricator.wikimedia.org/T115754

- [x] re-add "login" to settings menu when user not logged in (login-related item was formerly only shown in settings menu if user was logged in)
- [x] after login interface is dismissed (on successful login or if user cancels) the settings interface is re-shown
- [x] when logged in the settings menu item changes to "Log out *username*"
- [x] change "Log in" menu item text to "Log in to Wikipedia"
- [x] test logging in from settings
- [x] test account creation from settings (the login interface allows the user to tap "create account" if they don't already have one)

screen shot 2015-12-07 at 12 28 21 pm
screen shot 2015-12-07 at 12 29 33 pm

### DIFF
--- a/Wikipedia/Code/WMFSettingsViewController.m
+++ b/Wikipedia/Code/WMFSettingsViewController.m
@@ -464,11 +464,6 @@ static SecondaryMenuRowIndex const WMFDebugSections[WMFDebugSectionCount] = {
             [self deleteRowWithTag:WMFDebugSections[i]];
         }
     }
-
-    NSString* userName = [SessionSingleton sharedInstance].keychainCredentials.userName;
-    if (!userName) {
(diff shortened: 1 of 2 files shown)


bgerstle:
lgtm :+1:

coreyfloyd:
is this supposed to be buried so low? Seems like it shod be the first thing - or very close to the top. That might not even make it above the fold on an iPhone 4.

bgerstle:
@coreyfloyd that's where it is in 4.1.7

coreyfloyd:
@bgerstle that was log out - not log in.

montehurd:
@coreyfloyd per Josh - he wants to do a ticket to re-organize the settings buttons post-beta.

PR #304, opened 2015-12-08
Fix POTD error handling

bgerstle:
Phab: [T120687](https://phabricator.wikimedia.org/T120687)

Oversight, meant to force date formatter used w/ POTD to be EN-only.  Added tests, including a scheme just for running the app & tests in Arabic.  New `.travis.yml` should test all our usual tests in English, and the selected tests again in Arabic, in parallel.

### DIFF
--- a/Wikipedia/Code/NSDateFormatter+WMFExtensions.h
+++ b/Wikipedia/Code/NSDateFormatter+WMFExtensions.h
@@ -39,6 +39,6 @@
 
 + (instancetype)wmf_mediumDateFormatterWithoutTime;
 
-+ (instancetype)wmf_hyphenatedYearMonthDayFormatter;
++ (instancetype)wmf_englishHyphenatedYearMonthDayFormatter;
 
 @end
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
 language: objective-c
 osx_image: xcode7.1
+env:
+  matrix:
+    - FL_ARGS="verify"
+    - FL_ARGS="verify scheme:WikipediaArabic"
 
 install:
   - make travis-get-deps
@@ -11,10 +15,13 @@ script:
   - echo $SIMULATOR_ID
   - open -b com.apple.iphonesimulator --args -CurrentDeviceUDID $SIMULATOR_ID
   - set -o pipefail
-  - bundle exec fastlane verify
+  - bundle exec fastlane $FL_ARGS
 
 after_success:
   - bash <(curl -s https://codecov.io/bash)
+
+after_failure:
+  - tail -n 500 ./build/xcodebuild.og
 
 branches:
   only:
--- a/Wikipedia.xcodeproj/xcshareddata/xcschemes/WikipediaArabic.xcscheme
+++ b/Wikipedia.xcodeproj/xcshareddata/xcschemes/WikipediaArabic.xcscheme
@@ -1,0 +1,361 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Scheme
+   LastUpgradeVersion = "0710"
+   version = "1.3">
+   <BuildAction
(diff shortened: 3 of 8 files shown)


bgerstle:
@coreyfloyd @montehurd we could also use this setup now to run visual tests both in EN and Arabic (i.e. LTR & RTL)

montehurd:
@bgerstle that's so cool!

Wikipedia/Code/NSDateFormatter+WMFExtensions.h
```diff
@@ -39,6 +39,6 @@
 
 + (instancetype)wmf_mediumDateFormatterWithoutTime;
 
-+ (instancetype)wmf_hyphenatedYearMonthDayFormatter;
++ (instancetype)wmf_englishHyphenatedYearMonthDayFormatter;
```
coreyfloyd: Are we doing anything for other languages?

bgerstle: not sure what you mean, this date formatter is meant to convert dates into english format (e.g. `2015-12-8`) regardless of the current device language.

bgerstle: AFAIK we don't need a generic hyphenated date formatter

coreyfloyd: Ahh, so we are using english locale for the date formatter regardless of the user's locale?

bgerstle: yup, you can see the implementation, but the tests also verify this behavior (now that we have EN and AR language schemes)

coreyfloyd: @bgerstle yeah - I'm not sure what setting the locale to english actually does if you have specified the format that you want.

bgerstle: it forces the strings to be english. the default is to use the current locale, which can result in output that doesn't use the [numeric system](https://en.wikipedia.org/wiki/Decimal) that's required in this case (e.g. Arabic instead of English)

bgerstle:
:-1: now that I think of it, let me double check how this handles different calendars too

bgerstle:
Seems to work :checkered_flag::

image

coreyfloyd:
What is "BE"?

bgerstle:
"Buddhist Era," which is an epoch in the [Buddhist calendar](https://en.wikipedia.org/wiki/Buddhist_calendar)

Wikipedia/Code/NSDateFormatter+WMFExtensions.h
coreyfloyd: @bgerstle i was looking up the locale stuff as well - it seems to have the most effect on the preset styles (long medium short) - as it formats them specifically with each locale. 

With custom formats, that all it does is basically what you said: Just change characters/numbers. (Also it seems the formatter can fail because not all things work in all locales)

Makes me wonder if for non-english languages if we should use the built in styles because they are formatted in a way the user is guaranteed to understand. Rather than forcing an english convention.

bgerstle: @coreyfloyd sorry if there was some confusion about this again, but this formatter is only used to create page titles for network requests—i.e. never seen by the user.  I can separate these kinds of formatters out into a different category to clarify this.

bgerstle: or, I can not use a date formatter at all, and instead use `NSDateComponents` to pull out the year, month, and day.

coreyfloyd: Oh - my bad. What is the picture of then? Where is that date being formatted?

PR #306, opened 2015-12-09
Add multiple featured sections and preserve featured section site

coreyfloyd:
What's on the tin.

### DIFF
--- a/Wikipedia/Code/WMFFeaturedArticleSectionController.h
+++ b/Wikipedia/Code/WMFFeaturedArticleSectionController.h
@@ -8,8 +8,9 @@ NS_ASSUME_NONNULL_BEGIN
 @interface WMFFeaturedArticleSectionController : NSObject<WMFArticleHomeSectionController>
 
 @property (nonatomic, strong, readonly) MWKSite* site;
+@property (nonatomic, strong, readonly) NSDate* date;
 
-- (instancetype)initWithSite:(MWKSite*)site savedPageList:(MWKSavedPageList*)savedPageList;
+- (instancetype)initWithSite:(MWKSite*)site date:(NSDate*)date savedPageList:(MWKSavedPageList*)savedPageList;
 
 @end
 
--- a/Wikipedia/Code/WMFFeaturedArticleSectionController.m
+++ b/Wikipedia/Code/WMFFeaturedArticleSectionController.m
@@ -16,7 +16,7 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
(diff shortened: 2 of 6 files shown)


bgerstle:
for featured article, the `site` parameter isn't really necessary since it should always be "en" anyway. can we either remove it and hard-code to "en" or add an assertion?

bgerstle:
it was an oversight to accept a `site` param in the initializer to begin with

coreyfloyd:
I don't want to start assuming that the featured article comes form en. Would rather just encode it now. Better to keep the site encoded with the data that it represents than let assumptions about en bubble up all the way tot he view controller.

Accepting the site isn't really an oversite… that is what it should be - we are in a temporary situation where english has special things.

bgerstle:
I realize it's not ideal, but IMO the trade-off of making the class (and bugs like these) less error-prone vs. future-proofing (potentially quite far into the future) is a good one.  At the very least, we should add an assertion so we can catch it early, and remove the assertion when it's no longer needed.

coreyfloyd:
@bgerstle Assertion isn't needed - its encoded now - problem solved. And there is an assertion in the section itself. Also the bug is fixed. 

I'm not sure what part of the code you are referring to as bug prone. Allowing the assumptions to hang around is what caused the bug in the first place and is likely to cause future bugs… encoding the site removes the ambiguity.

If you want to discus further let me know… but its done, it works, its more explicit that what you are suggesting and has no apparent disadvantages compared to your proposed solution. This mostly sounds like bike shedding.

bgerstle:
I do want to discuss further, but it doesn't have to block the PR. Code is good :+1:

bgerstle:
@montehurd :basketball:'s in your court

montehurd:
We have some repeats:

feed

Repro:
- load feed
- stop app
- change calendar date
- start app
- repeat

Appears to show an additional featured item each time I change the date, but it makes all of the featured items show the article for the current date instead of the article from the date during which the respective featured item was fetched.

montehurd:
We have some repeats. gif below shows what happened when I used the repro steps below with 3 different dates to simulate using the app for 3 consecutive days:

feed

Repro steps:
- start app
- load feed
- stop app
- change OS X calendar date
- repeat

Appears to show an additional featured item each time I change the date, but it makes all of the featured items show the article for the current date instead of the article from the date during which the respective featured item was fetched.

montehurd:
:+1: @bgerstle - wanna merge once travis is done?

PR #307, opened 2015-12-09
Banners

coreyfloyd:
Covers:
https://phabricator.wikimedia.org/T97786
https://phabricator.wikimedia.org/T118350

Adding TSMessages pod for the banner logic.
This is managed by a WMFAlertManager. 

- [x] Need real assets
- [x] Fine tune the styling


Pods/HexColors/Classes/HexColors.m
```diff
@@ -0,0 +1,119 @@
...
+    }
+    
+    return hexString;
+}
+
++ (unsigned)hexValueToUnsigned:(NSString *)hexValue
+{
+    unsigned value = 0;
+    
+    NSScanner *hexValueScanner = [NSScanner scannerWithString:hexValue];
```
montehurd: :hankey:

bgerstle:
:-1: banner should always cover everything, but doesn't cover search:

image


when dismissed:

image

Also, it should display the localized version of "no search results."  Doesn't seem to have any icons either.  I'm seeing a bunch of logs which might be related (even after clean build & install):

> 2015-12-09 22:00:47.630 Wikipedia Debug[31945:7477677] CUICatalog: Invalid asset name supplied: (null)
2015-12-09 22:00:47.630 Wikipedia Debug[31945:7477677] CUICatalog: Invalid asset name supplied: (null)
2015-12-09 22:00:47.630 Wikipedia Debug[31945:7477677] CUICatalog: Invalid asset name supplied: (null)
2015-12-09 22:00:47.630 Wikipedia Debug[31945:7477677] CUICatalog: Invalid asset name supplied: (null)

bgerstle:
Also, some ideas for the alert API:

### `WMFAlertManager`
Doesn't seem to do much at the moment, what was the rationale behind it?  If there aren't any immediate plans to add more logic to this class, I'd suggest we remove the extra code/class by moving the call to `TSMessages` into an instance method on `WMFAlert`. For example, `WMFAlert.show(duration:viewController:?)`

### `WMFAlert`
:warning: Disclaimer: These are architectural code concerns.  I'm fine filing this as a "tech debt" task which we address after the beta is released.  That said, read on if you were interested in ways to improve the API and would consider making changes now. :warning:

This is another instance a pattern which I think leads to less maintainable code: one class that branches its behavior based on the (immutable) value of a single property, and stores its "real" data in various optional properties (also see `WMFHomeSection`).  Breaking this up into class hierarchy* would:

- Clarify which properties are available at any given time (instead of making everything optional and having to do `if let` or worse `!`)
- Improve readability by separating all the custom behaviors into their own implementations (e.g. all the `messageTitle` logic would be functions in each specialized alert class)

Making these changes can still preserve the following positive characteristics of the current API:

- Sharing "defaults" (e.g via the base class)
- Simple API (e.g. factory methods: `WMFAlert.alertWithError:` `WMFAlert.notificationWithMessage:`)

* I suggest class hierarchy even though we're in Swift, because this API is likely to get heavy use from ObjC, so interoperability is a chief concern.

coreyfloyd:
@bgerstle yeah - before I even read this I refactored it - didn't like it much myself. Current changes probably address most of your concerns - take a look and we can jump on hangout.

bgerstle:
:heart: will have a look in a bit :eyeglasses:

bgerstle:
@coreyfloyd liking this new architecture much better! :100: The one, minor point I would make is that it still seems like the `WMFAlertManager` class isn't necessary, since most of this API/logic could be implemented as a category on `TSMessage` by:

- Refactoring the `WMFAlertManager` methods into our own class category methods on `TSMessage`
- Adding our style file in a `class func load()` method of the category (similar to how we're [currently configuring CocoaLumberjack](https://github.com/wikimedia/wikipedia-ios/blob/master/Wikipedia/Code/DDLog+WMFLogger.m#L15))

This way, all our "business logic" on top of TSMessage wouldn't require a separate class or any state on top of what TSMessage already provides.  Let me know if you want to hop on hangout to discuss more or leave as-is.

bgerstle:
also, more of a semantic thing: but the "alert" terminology seems a little weird to me when used for things that aren't bad (e.g. `showSuccessAlert`). I'm guessing this is why TSMessage uses "notification" terminology.  What do you think?

bgerstle:
Yay, fixed!

image

But, shouldn't it be yellow/red and have the !/X graphic?

bgerstle:
This one shows the correct graphic:

image

montehurd:
I'm no longer getting any alert message when I try to log in with a bad password.

Username: 
    acct_creation_test_023
Good pwd:
    test123
Bad pwd:
    anything else :)

montehurd:
That fixed the bad pwd login msg! :)

Found one more:

I tried to create acct which already exists and it just stuck on the "Saving..." message and never shows the failure response back from the server...

screen shot 2015-12-10 at 12 46 23 pm

montehurd:
Ah the message was beneath the "Saving..." message! So I think the default behavior should maybe be replacing previous alerts? Then we could make it not replace as needed?

montehurd:
Either way is fine, but we'll have to do deeper audit of all alerts if we're wanting to switch to non-replacement as the default.

montehurd:
Same issue in edit workflow...

screen shot 2015-12-10 at 12 57 08 pm

bgerstle:
small regression: languages picker should dismiss the "loading" banner after article languages are downloaded

montehurd:
That seems to have fixed most of the issues.

I'm just seeing one issue at the end of the acct creation sequence.

At the very end of the gif below the "Logging in" message shouldn't have been hanging around...

untitled-2

bgerstle:
also, there shouldn't be "X" buttons on the "saving" or "logging in" banners

montehurd:
@bgerstle in case they get stuck I'd not mind so much if the x button stayed for these...

montehurd:
@bgerstle oh i see there's the x button on the left...

montehurd:
Oh looks like we're not flipping in RTL...

screen shot 2015-12-10 at 2 18 30 pm

Not using leading/trailing for constraints and natural for alignment?

bgerstle:
looks like all the frames are [laid out manually](https://github.com/KrauseFx/TSMessages/blob/master/Pod/Classes/TSMessageView.m#L278) :grimacing: and I guess autoresizing masks don't do leading/trailing by default

coreyfloyd:
@bgerstle @montehurd yeah - looks like no RTL on this - Since its frames, I'm going to have to redo it in auto layout. All the better anyways. Going to file that as a bug and do it as a follow though. That will take me a while.

@montehurd about to look into your weirdness again in a minute.

montehurd:
@coreyfloyd auto layout follow-up sounds good

montehurd:
@coreyfloyd if you test acct creation I've been changing to test wiki first, then naming them all like so:

username:
acct_creation_test_xxx (where xxx is a 3 digit number)

pwd:
test123

I've used up most numbers between 001 and 120 or so.
I kept them namespaced with consistent pwds in case we need to clean up after ourselves.

montehurd:
Seems to have fixed all the login/acct creation issues. Tested from settings menu and editing workflow.

montehurd:
One time after tapping the edit pencil the loading message didn't go away even after the wikitext was loaded, but I wasn't able to repro.

montehurd:
I did notice something that may be out of scope for this ticket:

- on a slow network (i used network link conditioner) there is no indication that anything is happening between the time it type a letter in the search box and the time search results are retrieved.

coreyfloyd:
@montehurd yeah slow networks don't have any feedback. We don't have anything in scope for when something takes a long time. Something you could bring up to nirzar. But yeah out of scope here.

montehurd:
@coreyfloyd cool

PR #308, opened 2015-12-09
Fix issue causing nearby to be empty after clean install or after feed reloads

bgerstle:
Phab: [T120505](https://phabricator.wikimedia.org/T120505)

---

- [x] Nearby section is populated with titles after clean install where user allows access
- [x] Nearby section is populated w/ same items after new sections are added on top of it (and feed reloads)

Turns out we had re-entrant code that was requesting access to the user's multiple times which, when removed, allowed the location manager to return valid locations.

This one's not easy to unit test, given all the various `WMFLocationManager` instances being used.  Open to ideas though.


coreyfloyd:
:+1:

PR #309, opened 2015-12-09
Fix crash when tapping on placeholder cells

bgerstle:
Phab: [T120473](https://phabricator.wikimedia.org/T120473)

See commit message.  Also a minor boyscout of adding the error delegate callback to the random section.


coreyfloyd:
:+1:

PR #310, opened 2015-12-09
Language picker updates

montehurd:
https://phabricator.wikimedia.org/T119150
https://phabricator.wikimedia.org/T111555

- [x] Add localized language name to language picker cell.
Note the following: 

At the moment, when we get a list of *article* languages (i.e. when you want to see all the languages the Obama article is available in) we are also returned the localized name for each language, with some having English fallbacks. 

When we're showing the full language list but *not* for a given article (i.e. when you're picking the search language) we don't presently get this list from the server, but we may want to in the future. In the mean time, for this use case, the localized language names are retrieved from iOS, which can provide *most* of them for given language codes - i.e. "en" -> "English". We fallback to English when iOS can't provide a localized language name. We also don't ask iOS for language names if the language code is compound - i.e. if the code has a "-" in it. This is because iOS will, for example, report back "Chinese" for every language code starting with "zh-" regardless of what comes after the "-". This is a further reason to either refactor later to ping the server for these, or bundle a full matrix mapping all localized lang names to all wiki lang codes.

- [x] Add language code to language picker cell.
- [x] Make localized language name be first thing in lang cell. (Per Moriel and Josh)
- [x] If the localized language code turns out to be the *exact* same thing as the non-localized lang name, don't show the non-localized lang name. Looks silly to have same string twice.
- [x] Make language filter work on lang codes too.
- [x] Sort by lang codes - same as desktop wikipedia. (Per Moriel and Josh)

**English Screenshots**

Before 1:
before 1

After 1:
after 1

Before 2:
before 2

After 2:
after 2

**Hebrew Screenshots**

Before 1:
before 1

After 1:
after 1

Before 2:
before 2

After 2:
after 2

Before 3:
before 3

After 3:
after 3

Before 4:
before 4

After 4:
after 4

### DIFF
--- a/Wikipedia/Code/LanguageCell.h
+++ b/Wikipedia/Code/LanguageCell.h
@@ -4,13 +4,11 @@
 /// Table cell designed for displaying a language autonym & a page title in that language.
 @interface LanguageCell : UITableViewCell
 
-/// Large, left-aligned label.
-@property (weak, nonatomic) IBOutlet UILabel* languageLabel;
+@property (strong, nonatomic) NSString* localizedLanguageName;
+@property (strong, nonatomic) NSString* articleTitle;
+@property (strong, nonatomic) NSString* languageName;
+@property (strong, nonatomic) NSString* languageCode;
 
-/// Smaller, left-aligned label beneath @c languageLabel.
-@property (weak, nonatomic) IBOutlet UILabel* titleLabel;
-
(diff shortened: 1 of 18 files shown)


bgerstle:
lgtm :+1:

coreyfloyd:
Aside from my comment - this is all looking very solid. Nice work.

coreyfloyd:
:+1:

PR #314, opened 2015-12-11
Make search language and sendUsageReports settings idempotent

bgerstle:
Phab: [T120434](https://phabricator.wikimedia.org/T120434)

---

Fixes were relatively simple, but went ahead and added some tests to make sure this sort of thing doesn't happen again.

- [x] Feed no longer reloads needlessly when search is dismissed
- [x] Feed reloads when search site is changed via search language picker
- [x] Feed & search are affected by search site changes from settings


montehurd:
Haven't been able to determine reproduction steps, but twice when switching languages and doing searches it got stuck where I'd see search results, but tapping on them did nothing... not sure if it's related this pr...

montehurd:
:+1:

PR #318, opened 2015-12-12
only search using the searchSite user preference

bgerstle:
Phab: [T121184](https://phabricator.wikimedia.org/T121184)

---

- [x] changing site from settings works
  - [x] reloads feed
  - [x] sets selected site when search is presented
- [x] reading an article from a site other than the current search site does not change search site (e.g. EN featured article while search site is ES)

removes the searchSite delegate method & any implementations (it was optional anyway).


montehurd:
It works! :+1:

PR #319, opened 2015-12-14
App icon force touch

montehurd:
Grabbed this as a fun weekend ticket. (only finished the interface bits - feel free to push commits for the last 2 checklist items)

- [x] add items to app icon force touch
- [x] show *continue reading* only if article title is available
- [x] show article title beneath *continue reading*
- [x] use localized strings
- [x] icons!
- [x] hook up *pops* to load respective interfaces

Testing:
- [x] ensure selecting "Search Wikipedia" works if search interface had been visible when the app was last closed
- [x] ensure selecting "Search Wikipedia" shows the keyboard so user can immediately search
- [x] ensure "Continue reading" appears conditionally (i.e. not if user was last looking at feed)
- [x] ensure selecting "Random article" actually requests a new random article each time
- [x] ensure app icon items work if app minimized when on "Welcome" screen
- [x] ensure the "Continue reading" subtitle actually updates to title of the article last read 
- [ ] test with data migration

untitled mov

Note: iOS moves the shortcut icons to the left or right of the text to try to keep the shortcut icons lined up with the actual app icon. Move the app icon to the right side of the screen and you'll see the shortcut icons move to the right of their text as well. iOS will also reverse order of appearance of items (Search will be on the bottom) if the app icon is placed in the bottom half of the screen.


montehurd:
Nearby and random icon screenshot:
screen shot 2015-12-14 at 9 51 18 am

montehurd:
:-1: Getting crash... hmm...

nirzar:
@montehurd for nearby articles, can we put the user into the VC for "more nearby" which basically shows more than 3 nearby articles?

Wikipedia/Code/AppDelegate.m
```diff
@@ -52,15 +51,52 @@ - (BOOL)application:(UIApplication*)application didFinishLaunchingWithOptions:(N
...
+    
     NSLog(@"%@", [[[NSFileManager defaultManager] URLsForDirectory:NSDocumentDirectory inDomains:NSUserDomainMask] lastObject]);
 
     return YES;
 }
 
+- (void)updateDynamicIconShortcutItems {
+    self.shortcutItemSelectedAtLaunch = nil;
+
+    if (![[UIApplication sharedApplication] respondsToSelector:@selector(shortcutItems)]) return;
```
bgerstle: wrap in curly braces please

Wikipedia/Code/AppDelegate.m
```diff
@@ -52,15 +51,52 @@ - (BOOL)application:(UIApplication*)application didFinishLaunchingWithOptions:(N
...
+    
+    UIApplicationShortcutItem* (^makeShortcut)(NSString*, NSString*, NSString*, NSString*) = ^(NSString* type, NSString* title, NSString* subtitle, NSString* icon) {
```
bgerstle: would you consider moving this into a category—as it's basically a custom/convenience factory/initializer for `UIApplicationShortcutItem`? doesn't seem to gain anything from being a block (AFAICT) and only makes this method longer.

Wikipedia/Code/AppDelegate.m
```diff
@@ -52,15 +51,52 @@ - (BOOL)application:(UIApplication*)application didFinishLaunchingWithOptions:(N
...
+
+-(void)application:(UIApplication *)application performActionForShortcutItem:(UIApplicationShortcutItem *)shortcutItem completionHandler:(void (^)(BOOL))completionHandler {
+    self.shortcutItemSelectedAtLaunch = shortcutItem;
+}
+
 - (void)applicationWillResignActive:(UIApplication*)application {
     // Sent when the application is about to move from active to inactive state. This can occur for certain types of temporary interruptions (such as an incoming phone call or SMS message) or when the user quits the application and it begins the transition to the background state.
     // Use this method to pause ongoing tasks, disable timers, and throttle down OpenGL ES frame rates. Games should use this method to pause the game.
     [[NSUserDefaults standardUserDefaults] wmf_setAppResignActiveDate:[NSDate date]];
+    [self updateDynamicIconShortcutItems];
```
bgerstle: move this to background, no need to do this when app resigns active (i.e. phone call, notification, etc.)

Wikipedia/Code/AppDelegate.m
```diff
@@ -80,6 +116,7 @@ - (void)applicationDidBecomeActive:(UIApplication*)application {
 
 - (void)applicationWillTerminate:(UIApplication*)application {
     // Called when the application is about to terminate. Save data if appropriate. See also applicationDidEnterBackground:.
+    [self updateDynamicIconShortcutItems];
```
bgerstle: would need to check this, but might be redundant if we already call it when going to background

Wikipedia/Code/WMFAppViewController.m
```diff
@@ -209,14 +221,72 @@ - (void)pauseApp {
     [self performHousekeepingIfNecessary];
 }
 
+-(void)handleSelectionOfShortcutItemIfNecessaryThen:(void (^)(BOOL shortcutWasHandled))completion{
+    UIApplicationShortcutItem *shortcutItemSelectedAtLaunch = [((AppDelegate*)[UIApplication sharedApplication].delegate) shortcutItemSelectedAtLaunch];
```
bgerstle: this should be injected, not accessed via the shared  `UIApplication` delegate.

Wikipedia/Code/WMFAppViewController.m
```diff
@@ -209,14 +221,72 @@ - (void)pauseApp {
...
+    if (shortcutItemSelectedAtLaunch) {
+        
+        // Main UI may *not* have been loaded if the user...
```
bgerstle: IMO, this means the shortcut items should be empty until the user goes through FTUE, obviating this logic. IOW, i don't think we want the user to circumvent FTUE via 3D touch shortcuts

Wikipedia/Code/WMFAppViewController.m
```diff
@@ -209,14 +221,72 @@ - (void)pauseApp {
...
+            handleSelection();
+        }
+    }else{
+        completion(NO);
+    }
+}
+
+-(void)popToHomeAndScrollToSectionWithIdentifier:(NSString*)identifier performingHeaderButtonAction:(BOOL)performHeaderButtonAction{
+    [[self navigationControllerForTab:WMFAppTabTypeHome] popToRootViewControllerAnimated:NO];
+    dispatchOnMainQueueAfterDelayInSeconds(0.0, ^{ // Prevent crash when device not connected to Xcode.
```
bgerstle: this deserves a `HAX` prefix for sure.  sounds very strange...

Wikipedia/Code/WMFRandomSectionController.h
```diff
@@ -5,6 +5,8 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+static NSString* const WMFRandomSectionIdentifier = @"WMFRandomSectionIdentifier";
```
bgerstle: typically public constants are `extern` so as to encapsulate their raw value (and remove temptation to use it directly).

Wikipedia/Code/WMFArticleContainerViewController.m
```diff
@@ -556,13 +556,17 @@ - (void)viewWillAppear:(BOOL)animated {
     }
 }
 
+- (void)viewDidAppear:(BOOL)animated {
+    [super viewDidAppear:animated];
+    [[NSUserDefaults standardUserDefaults] wmf_setOpenArticleTitle:self.articleTitle];
```
bgerstle: did you check if this cause problems w/ peeking "erasing" the previous open article?

bgerstle:
overall looks good @montehurd! left some minor comments, but the main points I'd like to talk about more are:

- how we handle opening the chosen shortcut item (i.e. try to remove app-delegate access from AppVC, cleaning up launch/shortcut cooperation)
- how shortcut items are (or aren't) shown when FTUE is still in progress

Wikipedia/Code/AppDelegate.m
montehurd: Uncrustify ENGAGE!

Wikipedia/Code/AppDelegate.m
montehurd: Done!

Wikipedia/Code/WMFRandomSectionController.h
montehurd: Done!

Wikipedia/Code/AppDelegate.m
montehurd: Fixed!

Wikipedia/Code/AppDelegate.m
montehurd: Oh yeah that's redundant. Good catch.

Wikipedia/Code/WMFAppViewController.m
montehurd: Fixed!

bgerstle:
@montehurd is this still a WIP?

Wikipedia/Code/AppDelegate.m
```diff
@@ -52,11 +52,41 @@ - (BOOL)application:(UIApplication*)application didFinishLaunchingWithOptions:(N
...
+        [shortcutItems addObject:[UIApplicationShortcutItem wmf_shortcutItemOfType:WMFIconShortcutTypeContinueReading]];
+    }
+
+    [shortcutItems addObject:[UIApplicationShortcutItem wmf_shortcutItemOfType:WMFIconShortcutTypeSearch]];
+
+    [UIApplication sharedApplication].shortcutItems = shortcutItems;
+}
+
+- (void)application:(UIApplication*)application performActionForShortcutItem:(UIApplicationShortcutItem*)shortcutItem completionHandler:(void (^)(BOOL))completionHandler {
+    self.appViewController.shortcutItemSelectedAtLaunch = shortcutItem;
```
bgerstle: so, while things seem to work, we're ignoring the `completionHandler` parameter, which is supposed to say whether or not the action was successful . 

i have no idea what the consequences of never calling this could be, so I'd rather be on the safe side and honor the API contract by calling the block.  also, the docs state:

> If you find that your app was indeed launched using a quick action, perform the requested quick action within the launch method and return a value of NO from that method. When you return a value of NO, the system does not call the application:performActionForShortcutItem:completionHandler: method.

However, our `performActionForShortcutItem:` isn't being called twice, even though we're returning `YES` from `didFinishLaunching:`.  Not going to block PR since it's functional, but I'd like to talk about this more tomorrow, since I'm now a bit confused :confounded:

Wikipedia/Code/WMFAppViewController.m
```diff
@@ -207,14 +220,74 @@ - (void)pauseApp {
     [self performHousekeepingIfNecessary];
 }
 
+- (void)handleSelectionOfShortcutItemIfNecessaryThen:(void (^)(BOOL shortcutWasHandled))completion {
+    UIApplicationShortcutItem* shortcutItemSelectedAtLaunch = self.shortcutItemSelectedAtLaunch;
+    if (shortcutItemSelectedAtLaunch) {
+        // Main UI may *not* have been loaded if the user...
+        //      - minimized the app when viewing a "Welcome" screen
+        //      - then selected a 3d touch app icon shortcut menu item
```
coreyfloyd: If the user hasn't finished viewing the welcome screen, should there be any actions in the short cut menu?
Users probably shouldn't be able to use the shortcut menu to "skip" the welcome screen.

Wikipedia/Code/WMFNearbySectionController.h
```diff
@@ -5,6 +5,8 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+static NSString* const WMFNearbySectionIdentifier = @"WMFNearbySectionIdentifier";
```
coreyfloyd: should be `extern` with the string defined in the implementation file (instead of a static exposed in the header)

Wikipedia/Code/WMFAppViewController.m
bgerstle: they won't see the shortcut items since they need to be added at runtime.
However, they could minimize the app during on-boarding and *then* try to
use a shortcut item. side note: users can also circumvent on-boarding by
force-quitting and relaunching, but that's not really related to this.

On Tue, Jan 12, 2016 at 1:39 PM, Corey Floyd <notifications@github.com>
wrote:

> In Wikipedia/Code/WMFAppViewController.m
> <https://github.com/wikimedia/wikipedia-ios/pull/319#discussion_r49494519>
> :
>
> > @@ -207,14 +220,74 @@ - (void)pauseApp {
> >      [self performHousekeepingIfNecessary];
> >  }
> >
> > +- (void)handleSelectionOfShortcutItemIfNecessaryThen:(void (^)(BOOL shortcutWasHandled))completion {
> > +    UIApplicationShortcutItem* shortcutItemSelectedAtLaunch = self.shortcutItemSelectedAtLaunch;
> > +    if (shortcutItemSelectedAtLaunch) {
> > +        // Main UI may *not* have been loaded if the user...
> > +        //      - minimized the app when viewing a "Welcome" screen
> > +        //      - then selected a 3d touch app icon shortcut menu item
>
> If the user hasn't finished viewing the welcome screen, should there be
> any actions in the short cut menu?
> Users probably shouldn't be able to use the shortcut menu to "skip" the
> welcome screen.
>
> —
> <https://github.com/wikimedia/wikipedia-ios/pull/319/files#r49494519>.
>



-- 
IRC: bgerstle

Wikipedia/Code/WMFNearbySectionController.h
montehurd: Fixed!

Wikipedia/Code/WMFAppViewController.m
coreyfloyd: I guess what I am saying is that should we shouldn't try to build the UI if it doesn't exist - that means we haven't completed the on boarding. So is the UI doesnt exist - that should be a logic error. So maybe what we should have is an assertion here to guarantee the UI has been loaded?

Wikipedia/Code/WMFRandomSectionController.h
coreyfloyd: @montehurd looks like this one (in the random section controller) still needs to be converted to an extern

coreyfloyd:
@montehurd implemented the todos for random and nearby - should be good to go on that front now - not sure what other cleanup you need to do before it is ready to review.

Wikipedia/Code/WMFAppViewController.m
```diff
@@ -182,6 +199,8 @@ - (void)resumeApp {
             [[WMFAlertManager sharedInstance] showAlert:@"You have been notified" sticky:NO dismissPreviousAlerts:NO tapCallBack:NULL];
         }
     }
+
+    [self handleSelectionOfShortcutItemIfNecessary];
```
coreyfloyd: Doesn't this line need to be before line 177: 

`if ([self shouldShowHomeScreenOnLaunch]) {`

I don't think we `shouldShowHomeScreenOnLaunch` or `shouldShowLastReadArticleOnLaunch` if we are going to respect the app shortcut.

So seems like we should start this if/else chain with:

`if(self.shortcutItemSelectedAtLaunch){`

Wikipedia/Code/WMFAppViewController.m
```diff
@@ -189,14 +208,56 @@ - (void)pauseApp {
...
+            } else if ([shortcutItemSelectedAtLaunch.type isEqualToString:WMFIconShortcutTypeRandom]) {
+                [self.tabBarController setSelectedIndex:WMFAppTabTypeHome];
+                [[self navigationControllerForTab:WMFAppTabTypeHome] popToRootViewControllerAnimated:NO];
+                [self showRandomArticleAnimated:YES];
+            } else if ([shortcutItemSelectedAtLaunch.type isEqualToString:WMFIconShortcutTypeNearby]) {
+                [self.tabBarController setSelectedIndex:WMFAppTabTypeHome];
+                [[self navigationControllerForTab:WMFAppTabTypeHome] popToRootViewControllerAnimated:NO];
+                [self showNearbyListAnimated:YES];
+            } else if ([shortcutItemSelectedAtLaunch.type isEqualToString:WMFIconShortcutTypeContinueReading]) {
+                // Fall through to normal "Continue reading..." handling.
```
coreyfloyd: Not sure if we should rely on the other code path - better to just do it here. (as I noted above we probably want to short circuit the other code path to make sure that only perform the short cut action requested by the user)

Wikipedia/Code/WMFAppViewController.m
montehurd: @coreyfloyd iirc we need shouldShowLastReadArticleOnLaunch 'cause we just fall through and let  that logic handle showing the "continue reading" app icon menu item article...

coreyfloyd: @montehurd see my note below…

Wikipedia/Code/WMFAppViewController.m
montehurd: Ya that's probably the way to go. Do you want to tweak those bits?

coreyfloyd: @montehurd also note that if it has been more than 24 hours, the article won't actually be pushed (it will just be the first card on the home screen)

montehurd:
oh cool. lemme fix that...

PR #321, opened 2015-12-14
Be sure to reset nearby error when results are delivered

coreyfloyd:
https://phabricator.wikimedia.org/T121432


bgerstle:
@coreyfloyd IMO we should completely reset the nearby section so placeholders are shown while we fetch nearby articles. not sure how we should test this though

coreyfloyd:
@bgerstle yeah - i can bring the placeholders back while the search is happening - good idea

PR #325, opened 2015-12-14
Implement fetching of gallery metadata & images for offline use

bgerstle:
Phab: [T118350](https://phabricator.wikimedia.org/T118350)

---

No more offline banner in the image gallery—when everything finishes saving!  

> [Demo video](https://vid.me/XiS7)

## Some caveats

- There's no easy to way cancel image info requests at the moment, so if they start, we don't stop them
  - This can be addressed later w/ some more invasive refactoring
- We're fetching all the image info individually instead of in batches
  - Wouldn't be hard to batch (as we're doing in the normal gallery case), but could be considered premature optimization (the article gallery case itself is somewhat of a premature optimization 😅)

Did some Xcode reorganizing while I was implementing stuff.  Check out the "Images" group in the Xcode project for details.  One major idea I want your feedback on: do model objects all belong together, or in their respective domains?  I thought it made sense to put `MWKImage` in the "Images" group since it's sort of the entry point for a lot of image functionality.  Also in the top level of the folder is `UIImageView+WMFImageFetching`, which is our main API for image downloading.

## Food for thought
Testing & extending this class made me yearn for some refactoring. Let's talk about it at a future code review meeting.  I'll add my thoughts to the ether pad.


bgerstle:
@coreyfloyd added error reporting & changed success callbacks to wait for article image & gallery data downloads. can haz re-review?

coreyfloyd:
@bgerstle all looks good! :+1:

coreyfloyd:
@bgerstle forgot to mention as far as organization… 

I think organizing by domain is good. As far as putting models in with views… I can take it or leave it. As long as we are consistent I can't think of a reason that putting saved page model objects under the same folder as saved page views (in subfolders of course) would be problematic. (Even if we want to extract models into a separate library later - that still doesn't matter that much)

bgerstle:
cool, thanks @coreyfloyd

bgerstle:
@montehurd what say ye?

PR #330, opened 2015-12-15
Adding delete all for saved pages

coreyfloyd:
(This also adds deletion to recent pages as well per the discussion at planning meeting)

Added button
Extended the data source API
Propagate KVO messages to the view controller

### DIFF
--- a/Wikipedia/Code/WMFRecentPagesDataSource.m
+++ b/Wikipedia/Code/WMFRecentPagesDataSource.m
@@ -59,6 +59,26 @@ NS_ASSUME_NONNULL_BEGIN
             [dataSource removeItemAtIndexPath:indexPath];
             [[NSNotificationCenter defaultCenter] addObserver:dataSource selector:@selector(rebuildSections) name:MWKHistoryListDidUpdateNotification object:recentPages];
         };
+
+        [self.KVOController observe:self.recentPages keyPath:WMF_SAFE_KEYPATH(self.recentPages, entries) options:NSKeyValueObservingOptionInitial | NSKeyValueObservingOptionPrior block:^(WMFRecentPagesDataSource* observer, MWKHistoryList* object, NSDictionary* change) {
+            BOOL isPrior = [change[NSKeyValueChangeNotificationIsPriorKey] boolValue];
+            NSKeyValueChange changeKind = [change[NSKeyValueChangeKindKey] unsignedIntegerValue];
+            NSIndexSet* indexes = change[NSKeyValueChangeIndexesKey];
+
+            if (isPrior) {
+                if (changeKind == NSKeyValueChangeSetting) {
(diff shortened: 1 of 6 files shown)


Wikipedia/Code/WMFTitleListDataSource.h
```diff
@@ -24,6 +24,16 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (CGFloat)estimatedItemHeight;
 
+- (BOOL)showsDeleteAllButton;
```
bgerstle: I would prefer these deletion methods be grouped under their own protocol. this would strictly enforce that implementing `showsDeleteAllButton` requires "delete-able" data sources implement the others.

Wikipedia/Code/WMFRecentPagesDataSource.m
```diff
@@ -59,6 +59,26 @@ - (nonnull instancetype)initWithRecentPagesList:(MWKHistoryList*)recentPages sav
             [dataSource removeItemAtIndexPath:indexPath];
             [[NSNotificationCenter defaultCenter] addObserver:dataSource selector:@selector(rebuildSections) name:MWKHistoryListDidUpdateNotification object:recentPages];
         };
+
+        [self.KVOController observe:self.recentPages keyPath:WMF_SAFE_KEYPATH(self.recentPages, entries) options:NSKeyValueObservingOptionInitial | NSKeyValueObservingOptionPrior block:^(WMFRecentPagesDataSource* observer, MWKHistoryList* object, NSDictionary* change) {
+            BOOL isPrior = [change[NSKeyValueChangeNotificationIsPriorKey] boolValue];
+            NSKeyValueChange changeKind = [change[NSKeyValueChangeKindKey] unsignedIntegerValue];
+            NSIndexSet* indexes = change[NSKeyValueChangeIndexesKey];
+
+            if (isPrior) {
```
bgerstle: would be nice to not have to forward these KVO calls, but if we can't think of a way now it's fine.

bgerstle:
:+1: but would send some extra :heart: if you can figure out a way to avoid KVO forwarding or make the deletion part of data source API more explicit

Wikipedia/Code/WMFRecentPagesDataSource.m
coreyfloyd: Yeah - I can't really. Since "items" isn't a real property in the data source, there wasn't a way to support observations. Of course this all goes away with Core Data (and removal of the lists classes), so I didn't like it - but wasn't too concerned.

Wikipedia/Code/WMFTitleListDataSource.h
coreyfloyd: Yeah good idea - let me do that real quick…

bgerstle:
:+1:

PR #334, opened 2015-12-17
Add event logging for home view

coreyfloyd:
Did some cleanup and improved general logging as well:

- Moved all piwik logic to a category
- Generalized page tracking to take a "source" - so we know where the view came from (home, search, saved, recent)
- Added events for saving (also takes a source)
- Added events for home impression tracking (also takes a source)
- Re-added some preview tracking that was removed when we changed the preview logic


bgerstle:
:sparkles: :100: :sparkles:

PR #337, opened 2015-12-18
Fix continue reading

coreyfloyd:
Just need to use the open article title from user defaults


bgerstle:
image

bgerstle:
@coreyfloyd only problem is, I can't test w/ the tweak (feed never shows continue reading since I have to pop back to the feed to see the last open article, which removes the setting, which hides continue reading).  should we update the "continue reading tweak" to always go to home/explore root on launch?

coreyfloyd:
@bgerstle added a tweak to turn off the showing the last open article

bgerstle:
image

bgerstle:
:+1:

PR #338, opened 2015-12-18
Make sure we highlight cells when touching down in the ToC

coreyfloyd:
Because we use some custom selection logic - its a bit complicated
we have to apply the highlights manually when getting delegate callbacks
also have to handle the deselection when a user scrolls after touching down (which uitableview uses to cancel the existing highlight)

### DIFF
--- a/Wikipedia/Code/WMFTableOfContentsCell.swift
+++ b/Wikipedia/Code/WMFTableOfContentsCell.swift
@@ -59,6 +59,12 @@ public class WMFTableOfContentsCell: UITableViewCell {
     }
     
     public func setSectionSelected(selected: Bool, animated: Bool) {
+        if(selected && self.sectionSelectionBackground.alpha > 0){
+            return;
+        }
+        if(!selected && self.sectionSelectionBackground.alpha == 0){
+            return;
+        }
         UIView.animateWithDuration(animated ? 0.3 : 0.0) {
             if (selected) {
                 self.sectionSelectionBackground.alpha = 1.0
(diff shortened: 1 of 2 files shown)


Wikipedia/Code/WMFTableOfContentsCell.swift
```diff
@@ -91,6 +97,16 @@ public class WMFTableOfContentsCell: UITableViewCell {
             selectedSectionIndicator.alpha = 0.0
         }
     }
+    
+//    public override func setHighlighted(highlighted: Bool, animated: Bool) {
```
bgerstle: delete?

bgerstle:
@coreyfloyd i don't know if there's a dedicated spec/mock for this, but personally, I would expect the individual subsection (cell) to highlight during touch down, not the entire section:

screen shot 2015-12-18 at 3 34 44 pm

bgerstle:
:-1: bug when selection is cancelled by dragging the finger away before releasing:

screen shot 2015-12-18 at 3 34 21 pm

bgerstle:
aside from the bug, looks good. would be great to get visual tests for this in at some point

coreyfloyd:
@bgerstle i don't think there is really a spec for the highlighting - but I made the change - I think this is better since you know which cell you are actually touching.

bgerstle:
and so we define the specs by implementing them :stuck_out_tongue_winking_eye:

bgerstle:
:+1:

PR #339, opened 2015-12-18
Fix feed jumping & remove table view update queueing

bgerstle:
Phab: [T120474](https://phabricator.wikimedia.org/T120474)

---

Look ma, [no jumps](https://vid.me/LUHQ)!

### Important stuff

- `UITableView+WMFLockedCell` suppresses table view animations and repositions the desired cell to its previous position _in the window_
- Removing all the `reloadSections:` & `reloadRows:` calls, since
  - They were causing all kinds of weird stuff to fly around (headers, cells, etc.)
  - They were causing weird crashes from the `UITableView` implementation
    - We can investigate this further when re-architecting/writing the feed

### Plus a couple boyscouts:

- Add & silence some logs
- Fixing a bug when popping VC at the same time "read more" response comes back (could also be a cancellation race condition, didn't investigate further)


coreyfloyd:
:+1:

PR #341, opened 2015-12-18
Don't show errors for 0 nearby results.

coreyfloyd:
We get this error when we are in a weird language and nothing is nearby. Not really an error.
Also, we need to stop trying to update when we get an error.

### DIFF
--- a/Wikipedia/Code/WMFNearbySectionController.m
+++ b/Wikipedia/Code/WMFNearbySectionController.m
@@ -169,7 +169,15 @@ static NSString* const WMFNearbySectionIdentifier = @"WMFNearbySectionIdentifier
 - (void)nearbyViewModel:(WMFNearbyViewModel*)viewModel didFailWithError:(NSError*)error {
     self.nearbyError = error;
     [self.delegate controller:self didSetItems:self.items];
-    [self.delegate controller:self didFailToUpdateWithError:error];
+    
+    //This means there were 0 results - not neccesarily a "real" error.
+    //Only inform the delegate if we get a real error.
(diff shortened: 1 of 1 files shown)


Wikipedia/Code/WMFNearbySectionController.m
```diff
@@ -169,7 +169,15 @@ - (BOOL)hasResults {
...
-    [self.delegate controller:self didFailToUpdateWithError:error];
+    
+    //This means there were 0 results - not neccesarily a "real" error.
+    //Only inform the delegate if we get a real error.
+    if (!([error.domain isEqualToString:MTLJSONAdapterErrorDomain] && error.code == MTLJSONAdapterErrorInvalidJSONDictionary)){
+        [self.delegate controller:self didFailToUpdateWithError:error];
+    }
+
+    //Don't try to update after we get an error.
+    [self.viewModel stopUpdates];
```
bgerstle: not sure stopping updates is a good idea because:

- we might get valid (or non-empty) results w/ a new location
- we might be able to determine the location if it's initially unknown (which is a specific error)
- if we have results already, we want to continue updating distance & heading

coreyfloyd: I mean if we don't we just keep attempting fetches every few seconds - thats why the banner shows up

bgerstle: i see, didn't think of that.. the user can manually retry w/ "Check Again" button on error view, which is triggered by internal `nearbyError` property :+1:

PR #343, opened 2015-12-20
Mention build requires imagemagick and ghostscript

dusek:
With ImageMagick and without ghostscript, e.g., the build never finishes
and infinitely generates the following error message when Xcode runs
some build script:

```
convert: unable to read font `(null)' @ error/annotate.c/RenderFreetype/1152.
```


bgerstle:
Thanks for reporting this, @dusek! This is actually a bug, since the build should succeed even if either or both of those executables aren't missing.  I think it's better if we fixed this directly instead of asking developers to install these tools, so I'd rather not add instructions to install them to the README.

bgerstle:
Confirmed that this is definitely an issue. Uninstalling ghostscript and running a clean build results in:

image

bgerstle:
closing in favor of #344

PR #345, opened 2015-12-21
Deploy alphas & betas using production app ID

bgerstle:
See latest build in TestFlight/iTC (5.0.0.570).

All Alphas and Betas will be deployed using the `org.wikimedia.wikipedia` bundle identifier.  This obviates the need for special bundle ID suffixes, and a separate "Alpha" configuration (since it was equivalent to Beta except for the `.tfalpha` bundle ID suffix).


coreyfloyd:
@bgerstle has a conflict

bgerstle:
finally :relieved: @coreyfloyd please re-review

PR #348, opened 2015-12-21
Fix various warnings

coreyfloyd:
Mostly about not sending non conforming object to event logging.
But also some nullability stuff, etc…


montehurd:
Seems to have reduced the warnings I see. @coreyfloyd are you no longer seeing any warning? I see a couple still...

coreyfloyd:
@montehurd thee are still a few

Wikipedia/Code/MWKList.h
```diff
@@ -40,9 +40,9 @@ typedef id<NSCopying, NSObject> MWKListIndex;
 
 - (EntryType)entryAtIndex:(NSUInteger)index;
 
-- (EntryType __nullable)entryForListIndex:(IndexType)listIndex;
+- (EntryType __nullable)entryForListIndex:(MWKListIndex)listIndex;
```
bgerstle: this effectively removes the generic aspect of the `MWKList` API, meaning all the entry-related APIs won't auto-complete to the type parameters for the specific subclass.  AFAICT, the warning here is in `MWKHistoryList.addEntry`, which should be using `entry.title`, not `entry.listIndex`.

Wikipedia/Code/WMFArticleContainerViewController.m
```diff
@@ -228,7 +228,7 @@ - (WMFArticleListTableViewController*)readMoreListViewController {
     return _readMoreListViewController;
 }
 
-- (WMFArticleFetcher*)articleFetcher {
+- (nullable WMFArticleFetcher*)articleFetcher {
```
bgerstle: why is this `nullable` if we always return nonnull?

coreyfloyd:
@bgerstle updated

bgerstle:
@coreyfloyd :+1: thanks! left a minor comment on the "append read more" section stuff in 352ab9f. what do you think about adding a guard & assertion there?

bgerstle:
:+1:

PR #349, opened 2015-12-21
Don't show footer menu item for langs if none.

montehurd:
Hide the `Read in X languages` button if there are no other language versions of article.

**Before:**
screen shot 2015-12-21 at 12 15 01 pm

**After:**
screen shot 2015-12-21 at 12 29 21 pm


coreyfloyd:
:+1:

PR #351, opened 2015-12-22
Fixed regression with space beneath article wikidata description.

montehurd:
I accidentally added space beneath the wikidata description when I added padding to bottom of article to match side padding.

**Regression:**
before

**This fix:**
after


montehurd:
@bgerstle Unrelated to this pr, but I also noticed that face detection doesn't seem to be called for the lead image here...

montehurd:
@could just be the image quality, but I seem to recall the face detection being surprisingly good even with poor quality images.

PR #352, opened 2015-12-22
Update to welcome search languages text.

montehurd:
https://phabricator.wikimedia.org/T120770

screen shot 2015-12-22 at 11 29 47 am

screen shot 2015-12-22 at 11 30 09 am


nirzar:
Can we use titles and description for alerts. the UI alert here uses entire dialog as a title. (bold text)

bgerstle:
Agree w/ @nirzar, we should be doing this on the first welcome view as well

bgerstle:
:+1: looks great @montehurd

PR #354, opened 2015-12-22
Face detection fix

coreyfloyd:
Trade off time!

Both the accuracy and the min feature size were too low to work on the Princess Helena picture.

The fix is using the defaults:
The default feature size is 0.1 of the size of the picture
The default accuracy is not defined - but it is higher than low accuracy and lower than high accuracy. It is documented as "good enough for most purposes". 

@bgerstle were there any performance reasons necessitating these higher performance values? If not, we should probably go with the system defaults.


bgerstle:
@coreyfloyd i might've picked the lower-accuracy ones just to optimize for time to screen, but couldn't remember and unfortunately didn't write a comment

bgerstle:
@coreyfloyd please re-run visual tests locally and check the diff to see if it's acceptable. if it is, please re-record the snapshots for failing tests

Wikipedia/Code/CIDetector+WMFFaceDetection.m
```diff
@@ -22,8 +22,8 @@ + (instancetype)wmf_sharedLowAccuracyBackgroundFaceDetector {
         defaultFaceDetector = [CIDetector detectorOfType:CIDetectorTypeFace
                                                  context:[CIContext wmf_sharedBackgroundContext]
                                                  options:@{
-                                   CIDetectorAccuracy: CIDetectorAccuracyLow,
-                                   CIDetectorMinFeatureSize: @(0.15)
+//                                   CIDetectorAccuracy: CIDetectorAccuracyLow,
```
bgerstle: can we pass `nil` instead?

coreyfloyd: @bgerstle we don't pass anything if we want to get the defaults - why do you want to pass nil?

coreyfloyd: oh you mean nil for the dictionary? Oh yeah… duh

bgerstle:
:+1:

PR #358, opened 2015-12-23
Fix missing featured article & POTD on clean install

bgerstle:
TFA & POTD section weren't part of the "starting" schema nor added during `reset`. Remedied this and also refactored the "add featured article section if supported" logic into one place (as this change required it being in multiple places). Ended up moving it into featured section creation as we should always handle cases where creating a featured section for a given site isn't supported.


coreyfloyd:
:+1:

PR #359, opened 2015-12-23
Added "About this article" header above footer menu.

montehurd:
https://phabricator.wikimedia.org/T122109

screen shot 2015-12-23 at 12 52 02 am

### DIFF
--- a/Wikipedia/Code/WMFArticleFooterMenuViewController.m
+++ b/Wikipedia/Code/WMFArticleFooterMenuViewController.m
@@ -39,6 +39,16 @@
     [self.tableView deselectRowAtIndexPath:[self.tableView indexPathForSelectedRow] animated:YES];
 }
 
+- (CGFloat)tableView:(UITableView *)tableView heightForHeaderInSection:(NSInteger)section {
+    // HAX: collapses space between grouped table sections
+    return 0.00001;
+}
+
+- (CGFloat)tableView:(UITableView *)tableView heightForFooterInSection:(NSInteger)section {
+    // HAX: collapses space between grouped table sections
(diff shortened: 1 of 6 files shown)


Wikipedia/Code/WMFArticleFooterMenuViewController.m
```diff
@@ -39,6 +39,16 @@ - (void)viewWillAppear:(BOOL)animated {
     [self.tableView deselectRowAtIndexPath:[self.tableView indexPathForSelectedRow] animated:YES];
 }
 
+- (CGFloat)tableView:(UITableView *)tableView heightForHeaderInSection:(NSInteger)section {
```
bgerstle: why can't we use a plain table instead? will "about this article" ever have multiple sections?

montehurd: @bgerstle 
With grouped tables we don't have to have any special logic to have separator on top of the first cell and on bottom of the last cell.

bgerstle: ah, gotchya. i knew we had discussed this before, but forgot. would this also work if we set `sectionHeaderHeight` and `sectionFooterHeight` properties in `viewDidLoad`?  might also be worth DRYing this up into a category and explaining why we do this in doc comments.

montehurd: > would this also work if we set sectionHeaderHeight and sectionFooterHeight properties in viewDidLoad

No unfortunately that doesn't seem to work...

screen shot 2015-12-23 at 11 45 54 am

PR #360, opened 2015-12-23
Fix wonky TOC header view rotation behavior.

montehurd:
TOC header autolayout now works fine during rotation without any special handling :)

----

Note: this has a run-time crash in iOS 8 related to a known issue with swift initializers and iOS 8 UITableViewController: 
http://stackoverflow.com/a/30719434

Ideas for how to proceed? The easiest suggested fix is to use a nib:
http://stackoverflow.com/a/29853391

### DIFF
--- a/Wikipedia/Code/WMFTableOfContentsViewController.swift
+++ b/Wikipedia/Code/WMFTableOfContentsViewController.swift
@@ -1,6 +1,6 @@
 
 import UIKit
-
+import Masonry
 
 public protocol WMFTableOfContentsViewControllerDelegate : AnyObject {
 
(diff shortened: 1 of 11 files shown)


Wikipedia/Code/WMFTableOfContentsViewController.swift
```diff
@@ -44,7 +44,10 @@ public class WMFTableOfContentsViewController: UITableViewController, WMFTableOf
         self.items = items
         self.delegate = delegate
         tableOfContentsFunnel = ToCInteractionFunnel()
-        super.init(nibName: nil, bundle: nil)
+        super.init(style:.Grouped)
```
montehurd: This is the iOS crash cause...

bgerstle:
@montehurd let me know how that works, I went w/ the SO answerer's suggestion of taking `UITableViewController` out of the equation instead of using a nib. one question I have for you about the changes I made: was there a specific reason why you used a background view for the table view, and does my change break something you did?

coreyfloyd:
Hmm… this is all pretty weird. 

I guess it works now - I don't think it really matters which approach is taken. As long as it works. Either a nib or forgoing using UITableVIewController - one thing to keep in mind is that UITableViewController deselects any selected cells in viewWillAppear - I don't think that matters here because we reselect everything before the ToC is displayed.

bgerstle:
> Hmm… this is all pretty weird.
Yeah, the Swift bug in iOS 8 is unfortunate, but we have to deal with it :disappointed:. From what you said, it sounds like you agree w/ my rationale that we weren't taking advantage of much `UITableViewController` functionality anyway.  I just went ahead and opted for the nib-less approach :wink:.

montehurd:
Cool thanks @bgerstle! Works perfectly :) :+1: 

> was there a specific reason why you used a background view for the table view

Oh, ya that was just for setting the background color.

PR #361, opened 2015-12-23
Removing main page from related

coreyfloyd:
Needed to re-add the code for getting the main page json from disk and using it to check if a title is a main page


montehurd:
Tested feed and seems fine :+1:

PR #363, opened 2015-12-23
Show localized language names in welcome screen and use article language for article footer menu localized strings.

montehurd:
https://phabricator.wikimedia.org/T122352
https://phabricator.wikimedia.org/T122353

**Welcome screen:**
screen shot 2015-12-23 at 1 13 05 pm

**Article footer:**
We don't have translations back for these yet. I faked out the strings in Hebrew and it worked, but I forgot to take a screen shot.


bgerstle:
:+1:

PR #364, opened 2015-12-23
Fix infinite loop when offline & add reachability handling

bgerstle:
Phab:
- [T122349](https://phabricator.wikimedia.org/T122349)
- [T121908](https://phabricator.wikimedia.org/T121908)
- [T121924](https://phabricator.wikimedia.org/T121924)

---

Look ma, [no internet connection!](https://cloud.githubusercontent.com/assets/444217/11985941/e4dfed8e-a997-11e5-8e8b-05017a66afa0.gif) (Full video [here](https://vid.me/7Cun))

Lazy fetching got too fancy by trying to automatically re-fetch sections the next time they were scrolled past if they didn't have any items (e.g. if it encountered an error). What I didn't think of, was the case where the user was offline, so the section would always error, then `reloadData`, then `fetchIfNeeded`, then error, etc. etc.  Now the Home VC handles this gracefully by not reloading sections for which it previously encountered an error, and resetting error states when the sections are reloaded.

While I was in there, I went ahead and added a few lines to implement reloading all the sections when the network comes back online. Should resolve the Phab tickets mentioned above and make offline UX a bit better for the Explore tab.


coreyfloyd:
Nice! :+1:

PR #365, opened 2015-12-26
Nearby cell styling update

nirzar:
nearby cell update 

1. the dashed border should be grey and the needle should be black so as it not interfere visually

2. the label background creates a better separation between very similar type sizes and colors used for wikidata descriptor


nirzar:
Here's a comparison. 
pr-preview

I talked to KH and we came to conclusion that the label background is important and the needle color also need more emphasize, based on that, here is a PR. The earlier visual design change was not reviewed by any designers. Sorry about back and forth.

PR #366, opened 2015-12-26
Updated two mini icons to match the stroke width

nirzar:
POTD and Recent stroke width did not match nearby.


nirzar:
@montehurd you mentioned that the POTD circle stroke width did not match the nearby mini stroke width. fixed it. also recent was also 1px lesser. fixed that too

PR #367, opened 2015-12-28

bgerstle:


montehurd:
Unfortunately our test for catching "1$" seems to be failing and some translator keeps doing it :(

montehurd:
Maybe ping Amir?

PR #368, opened 2015-12-29
Added small divider between wikidata descrip and first section of article.

montehurd:
Per @nirzar 

- [x] appears only atop first section
- [x] single pixel line width on retina
- [x] moves to right side in RTL langs

untitled mov

screen shot 2015-12-28 at 4 36 40 pm


www/less/misc.less
```diff
@@ -53,6 +53,29 @@ a{
     padding-bottom:0px;
 }
 
+// Horizontal divider between title/description and lead section text
+#content_block_0_hr {
+    border:1px solid #999;
```
nirzar: @montehurd hey, we need the hairline border here. it's important that it's 0.5px on retina screens. can we try  transform: scaleY(0.5); 

more info here http://atirip.com/2013/09/22/yes-we-can-do-fraction-of-a-pixel/

Also, i checked the color on my phone screen again and i think we should change it to #CCC 
Final screenshot

image

www/less/misc.less
```diff
@@ -53,6 +53,29 @@ a{
...
+    margin-top: 12px;
+    margin-bottom: 16px;
+    width: 60px;
+    margin-left:0px;
+}
+
+@media (-webkit-min-device-pixel-ratio: 2) {
+    #content_block_0_hr {
+        border-width:0.5px;
```
montehurd: @nirzar It already does the hairline here if the device is retina.

www/less/misc.less
montehurd: This needs to be 1 by default for non-retina screens. Retina screens use 0.5 based on my other comment.

montehurd: transform:scaleY is not necessary. Since we're iOS8 and above we can use media selectors `@media (-webkit-min-device-pixel-ratio: 2) {` to adjust the border width to 0.5 for retina screens.

PR #369, opened 2015-12-29
Reorganize Xcode groups & delete more dead code

bgerstle:
Phab: [T116615](https://phabricator.wikimedia.org/T116615)

--- 

Tried to do this in stages, but the end result is 3 top-level "Code" subgroups:

- **Common**: All "common" code which is shared across features.  This includes code related to networking, images, model layer, and utilities.
- **Features**: Contains a group for each feature domain in the app, which contains subgroups for its specialized components and sub-features.
- **Legacy**: as the name implies, legacy code that is on its way out and so not worth integrating into other groups

Some notable aspects of the group "schema":

- Not grouping all models into one group. This is an extension of the rationale behind not having all ViewControllers, Views, etc. in one group.  Keep stuff near where it's used so it's easy to understand what's part of a specific feature implementation or what's more general-purpose
- Large "Common" group which encompasses Networking, Style, etc. which is intended to help "guide" people to where they need to go (i.e. feature vs. common domains) and only go into the "Common" group when working at "lower level" (i.e. core abstractions & reusable components).

In my travels, I found code which was obviously not being used, and so deleted it as a follow-on to the last "dead code removal" PR.


montehurd:
:+1:

PR #370, opened 2015-12-30
Fix constraint breakage on iOS 6 plus on explore PoTD.

montehurd:


bgerstle:
:+1: fixed! but... I don't understand why the constraints weren't satisfiable to begin with:

>     "<NSLayoutConstraint:0x7fabf0ee02a0 UIImageView:0x7fabf5c0ccf0.height == 345>",
    "<NSLayoutConstraint:0x7fabf0ee6180 UIImageView:0x7fabf5c0ccf0.top == UITableViewCellContentView:0x7fabf5c01d70.top>",
    "<NSLayoutConstraint:0x7fabf0e70360 UITableViewCellContentView:0x7fabf5c01d70.bottom == UIImageView:0x7fabf5c0ccf0.bottom>",
    "<NSLayoutConstraint:0x7fabf0ed3b60 UITableViewCellContentView:0x7fabf5c01d70.height == 345>"

IOW, they have equal heights, and the image view is set to fill the content view.  @montehurd did you repro this on device or simulator?  I only ask because the intent w/ that cell is to always be 345 tall, and having the height constraint be lower priority results in the possibility of it being less.  The fact that this works on all other devices and the unsatisfiable constraint message is nonsensical indicates to me that this is either a bug in the simulator (to be verified by repro'ing on device) or auto layout in iOS 9 on iPhone 6+.

bgerstle:
to be clear, I'm not seeing any shortened POTD cells as a result of the change, but if the intent is to "require" the cell be 345 tall, IMO it would be ideal if the constraint reflected that explicitly instead of relying on `UITableView` and auto layout to somehow end up at 345

bgerstle:
confirmed that this is an issue w/ 6+ rounding by logging frames on 6+ after image is laid out in table view:

> (lldb) p (CGRect) [0x7fe21bca2810 frame]
(CGRect) $118 = (origin = (x = 0, y = 0), size = (width = 414, height = 345.00001017252606))
(lldb) p (CGRect) [0x7fe21bce56a0 frame]
(CGRect) $119 = (origin = (x = 0, y = 0), size = (width = 414, height = 345))

PR #374, opened 2016-01-12
Create concrete list view subclasses

coreyfloyd:
This patch creates a concrete VC for each type of list that is displayed.

This was done to separate the view concerns from the data sources and to give more control over the presentation of the different list VCs (for instance adding pull to refresh or enabling a delete all button).

A new VC was created for the following lists:
- Saved
- History
- Search
- Nearby
- Related Titles
- Article Read More

Note:
The nearby view model / distance bearing provider architecture is making separation of view concerns a bit harder in that situation. This should be tagged for future clean up refactoring.

Other note:
At this point I'm not sure we should be subclassing the data sources. We have a "coupled class hierarchy" between data sources and list view controllers. This smells a bit to me.

I don't think we are getting a lot of code re-use here. Basically we could move the code in the data sources into the view controllers and not really duplicate ourselves. We can/should consider doing this before merging.

### DIFF
--- a/Wikipedia.xcodeproj/project.pbxproj
+++ b/Wikipedia.xcodeproj/project.pbxproj
@@ -10,6 +10,7 @@
 		041EFC371996A1F800B2CB28 /* MapKit.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 041EFC361996A1F800B2CB28 /* MapKit.framework */; };
 		0493C2D419526A0100EBB973 /* WikiFont-Glyphs.ttf in Resources */ = {isa = PBXBuildFile; fileRef = 0493C2D319526A0100EBB973 /* WikiFont-Glyphs.ttf */; };
 		04D34DB21863D39000610A87 /* libxml2.dylib in Frameworks */ = {isa = PBXBuildFile; fileRef = 04D34DB11863D39000610A87 /* libxml2.dylib */; };
+		0E0361741C4485FA00FD9642 /* WMFSelfSizingArticleListTableViewController.m in Sources */ = {isa = PBXBuildFile; fileRef = 0E0361731C4485FA00FD9642 /* WMFSelfSizingArticleListTableViewController.m */; };
 		0E26B0721C0F4F550004D687 /* WMFAppViewController.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 0E26B0701C0F4F550004D687 /* WMFAppViewController.storyboard */; };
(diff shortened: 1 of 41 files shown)


bgerstle:
nice work, @coreyfloyd! some comments about architecture and a bug I found, but here's my $0.02 on the questions you raised in the PR desc:

> The nearby view model / distance bearing provider architecture is making separation of view concerns a bit harder in that situation. This should be tagged for future clean up refactoring.

I would've thought the view model would make this easier, as it doesn't know anything about the view. Did you have any trouble refactoring this?  Seems like it came together relatively easily and as I would've expected.  That said, there is a _tiny_ bit of duplication in the cell configuration (list VC uses the data source cell config block and the section controller does it conditionally based on the type of cell), but nothing to shake a stick at.

> At this point I'm not sure we should be subclassing the data sources. We have a "coupled class hierarchy" between data sources and list view controllers. This smells a bit to me.

> I don't think we are getting a lot of code re-use here. Basically we could move the code in the data sources into the view controllers and not really duplicate ourselves. We can/should consider doing this before merging.

Right, this came up in the gallery/POTD refactor as well, but in that case, the VC's had _very, very_ minor differences (literally, just a different data source), so it was worth consolidating.  In this case, I think it's beneficial to keep them separate now that they're already separate, instead of consolidating logic into the VC.  I'm really liking how small the new VC's are, we should keep it that way :+1:.  You can almost call them SRP-compliant—if you squint and sorta forget all the stuff the base class does.  Lastly, I'd offer that this is just composition at work, and I don't think we should consider these kinds of couplings a code smell unless the VC's they're coupled to have trivial differences (like I mentioned earlier).

bgerstle:
> At this point I'm not sure we should be subclassing the data sources

**TL;DR;** Probably a good idea to refactor shared functionality, but this is orthogonal to the "VC<->dataSource coupling" issue

They could be refactored into a more generic base class which takes a fetcher and populates its items. This could be DRY'd up really nicely in Swift, but I'm not sure the interop cost would be worth it.  Could be done in ObjC, which puts us in a similar situation as the `MWKList` hierarchy—base class that does core functionality, subclasses for list-specific behaviors.  IMO that works nicely, and could be done while preserving the "couplings" you mentioned above.

montehurd:
:+1:

coreyfloyd:
@bgerstle should be good to go

bgerstle:
for posterity: :+1:

PR #375, opened 2016-01-13
Fix save button RTL compliance

bgerstle:
Phab: [T121681](https://phabricator.wikimedia.org/T121681)

---

See reference images for the end result, and `WMFTextualSaveButton` header doc comments for rationale (capturing our previous discussion).


bgerstle:
@montehurd updated to use Hebrew text. please let me know before this is merged so i can update Jenkins to use the new scheme name

coreyfloyd:
42 files - holy shit!

Wikipedia/Code/UIColor+WMFStyle.h
```diff
@@ -67,4 +67,20 @@
...
+///
+/// @name Derivative Colors
+///
+
+/**
+ *  @return A dimmed copy of the receiver.
+ *
+ *  @see -wmf_colorByScalingComponents:
+ */
+- (instancetype)wmf_colorByApplyingDim;
```
coreyfloyd: maybe a bit more explanation here - maybe some domain knowledge of why we would do this - and maybe rename the method to capture that info.

Wikipedia/Code/WMFArticlePreviewTableViewCell.m
```diff
@@ -7,11 +7,12 @@
 #import "UITableViewCell+SelectedBackground.h"
 #import <Masonry/Masonry.h>
 #import "UITableViewCell+WMFEdgeToEdgeSeparator.h"
+#import "WMFTextualSaveButton.h"
 
 @interface WMFArticlePreviewTableViewCell ()
 
 @property (strong, nonatomic) IBOutlet UILabel* snippetLabel;
-@property (strong, nonatomic) IBOutlet UIButton* saveButton;
+@property (weak, nonatomic) IBOutlet WMFTextualSaveButton* saveButton;
```
coreyfloyd: strong

Wikipedia/Code/WMFSaveButtonController.m
```diff
@@ -15,7 +15,7 @@ @interface WMFSaveButtonController ()
 
 @implementation WMFSaveButtonController
 
-- (instancetype)initWithButton:(UIButton*)button
+- (instancetype)initWithButton:(UIControl*)button
```
coreyfloyd: `initWithControl:` unless you want to subclass `UIButton` instead.

Wikipedia/Code/WMFTextualSaveButton.m
```diff
@@ -0,0 +1,132 @@
...
+    self.saveTextLabel.tintAdjustmentMode     = UIViewTintAdjustmentModeAutomatic;
+}
+
+/**
+ *  Add any views which weren't set to the receiver's IBOutlet's in interface builder.
+ *
+ *  This allows the component to be created programmatically, dropped into IB using default views, and customizing one
+ *  or more subviews.
+ */
+- (void)addMissingViews {
```
coreyfloyd: If we are allowing configuration of the subviews in IB, then we should implement IBDesignable.

coreyfloyd: Also this class is being very specific and very generic at the same time:

Its very specific about only being a save button (but really it could be just a `WMFImageTextButton`)
But very general in what its subviews are (Hey drop in some random views to customize me!)

We should go one way or the other… either make a generic configurable button OR make the exact save button that we need with very little configuration possible. YAGNI would say the latter, but I would have no issues going with the former since its a very general problem with RTL support across multiple OSs.

Wikipedia/Code/main.m
```diff
@@ -5,14 +5,51 @@
...
+#if DEBUG
     // disable app when unit testing to allow tests to run in isolation (w/o side effects)
     BOOL const isUnitTesting = NSClassFromString(@"XCTestCase") != nil;
+#endif
     @autoreleasepool {
         return UIApplicationMain(argc,
                                  argv,
                                  nil,
-                                 isUnitTesting ? nil : NSStringFromClass([AppDelegate class]));
+                                 #if DEBUG
```
coreyfloyd: Repeating yourself (and making it harder to read) - just define the class in the `#if DEBUG` block above.

WikipediaUnitTests/Code/UIApplication+VisualTestUtils.h
```diff
@@ -0,0 +1,17 @@
...
+//  Copyright © 2016 Wikimedia Foundation. All rights reserved.
+//
+
+#import <UIKit/UIKit.h>
+
+@interface UIApplication (VisualTestUtils)
+
+- (NSString*)wmf_systemVersionAndWritingDirection;
+
+- (NSString*)wmf_userInterfaceLayoutDirectionAsString;
```
coreyfloyd: no reason to expose if not used externally

WikipediaUnitTests/Code/UIView+VisualTestSizingUtils.m
```diff
@@ -17,6 +17,7 @@ - (CGRect)wmf_sizeThatFitsScreenWidth {
 }
 
 - (CGRect)wmf_sizeThatFitsScreenWidthForOrientation:(UIInterfaceOrientation)orientation {
+    WMF_TECH_DEBT_TODO(use nativeBounds of mainScreen)
```
coreyfloyd: why not do this now… just `[UIScreen mainScreen].bounds`…

bgerstle: because some of the tests are meant to happen across multiple lines, which isn't true anymore if the view is wider. it's just more work that's not worth doing atm

WikipediaUnitTests/Code/UIApplication+VisualTestUtils.h
bgerstle: thought i did, will expose

Wikipedia/Code/main.m
bgerstle: i'll make this more straightforward w/ a variable instead of breaking up the ternary statement. is there something else about `DEBUG` usage you had issue with?

Wikipedia/Code/WMFSaveButtonController.m
bgerstle: sorry, got lazy—will rename

Wikipedia/Code/WMFArticlePreviewTableViewCell.m
bgerstle: sorry, missed that—it was the IB default

bgerstle:
@coreyfloyd RE: 42 files—i updated FBSnapshotTestCase

Wikipedia/Code/UIColor+WMFStyle.h
bgerstle: IMO that's backwards. this method says exactly what it does—it's up to the application to decide when it would be used. so far, it's only used to dim a color when a label is highlighted

Wikipedia/Code/WMFTextualSaveButton.m
bgerstle: Yeah, YAGNI's probably the safer bet at this point.  This is meant to be a standard save button (i.e. effectively hard coding the label text & icon image), so customizing doesn't really make sense.

Regarding IBDesignable, I'm tempted to implement it anyway since it's already being used in IB—even if its outlets aren't assigned to anything.

bgerstle: Sorry, to be explicit: the logic in this class should still be refactored into a more generic component (which it subclasses) should the need arise.

WikipediaUnitTests/Code/UIView+VisualTestSizingUtils.m
coreyfloyd: Not sure I follow - we have other visual tests that rely on the width of the device they are run on. Why does this particular suite need to be 320 pt wide?

Also if we are actually not resizing based on the screen width, then this method should be renamed - We shouldn't have to look at a method's implementation to know that it is lying to me.

Wikipedia/Code/main.m
coreyfloyd: just the 2 things: 
1. repeating the `#if ` twice within a couple lines.
2. Using `#if` in the middle of a ternary statement in the middle of a method is hard to read

Wikipedia/Code/WMFTextualSaveButton.m
coreyfloyd: > Regarding IBDesignable, I'm tempted to implement it anyway since it's already being used in IB—even if its outlets aren't assigned to anything.

Yeah, IBDesignable could still be nice because you get the preview in IB. I would kill the outlets though for YAGNI - since you can't customize the subviews any longer.

> Sorry, to be explicit: the logic in this class should still be refactored into a more generic component (which it subclasses) should the need arise.

Yeah - we can always extract if needed - wouldn't be too hard.

coreyfloyd:
> @coreyfloyd RE: 42 files—i updated FBSnapshotTestCase and uncrustified stuff. starting to wonder if we should make that part of the CI so other we're not polluting each other's diffs w/ uncrustification

Yeah - TBH I am leaning towards getting rid of having the pods committed to the repo.

It has mostly felt like a burden - the primary reason (for me) was to make sure that a pod can't get pulled out from under us - but the reality is that is very unlikely.

If we really want to guarantee we always have the source for our pods, we could:
1. Fork all pods to wikimedia so we have them
2. Create a repo specifically for pods - and commit them there. Then include that repo as a pod in the main project

bgerstle:
> Yeah - TBH I am leaning towards getting rid of having the pods committed to the repo.

@coreyfloyd I probably shouldn't have updated FBSnapshotTestCase as part of this PR, but I originally though the add'l diff of just updating that one Pod would be minimal.  No need to get into pros/cons of committing the Pods folder here, but happy to discuss further if you feel strongly about it.

bgerstle:
@coreyfloyd should be ready for re-review

WikipediaUnitTests/Code/UIApplication+VisualTestUtils.h
bgerstle: misunderstood, this is now encapsulated

bgerstle:
@coreyfloyd @montehurd `WMFTextualSaveButton` is now IBDesignable:

image

montehurd:
Whoa! Slick!

PR #377, opened 2016-01-13
Disable system share item in contextual text selection menu

coreyfloyd:
Disables the system share menu by returning NO for being able to perform the system share action.


bgerstle:
:+1: hopefully this doesn't get flagged in app store review though..

PR #378, opened 2016-01-13
Article footer menu RTL

montehurd:
https://phabricator.wikimedia.org/T122365

ios 8

ios 9


Wikipedia/Code/WMFArticleListTableViewCell.m
```diff
@@ -52,6 +52,8 @@ - (void)awakeFromNib {
     [super awakeFromNib];
     [self configureImageViewWithPlaceholder];
     [self wmf_makeCellDividerBeEdgeToEdge];
+    self.titleLabel.textAlignment       = NSTextAlignmentNatural;
```
bgerstle: nbd, but any reason why you didn't do this in the nib?

bgerstle:
:+1:

Wikipedia/Code/WMFArticleListTableViewCell.m
montehurd: I tried but it didn't seem to work... was going to try to track down why but I forgot.

Wikipedia/Code/WMFArticleFooterMenuDataSource.m
```diff
@@ -2,7 +2,7 @@
 #import "WMFArticleFooterMenuItem.h"
 #import "MWKArticle.h"
 #import "NSDate+Utilities.h"
-#import "WMFArticleFooterMenuCell.h"
+#import "WMFArticleListTableViewCell.h"
```
coreyfloyd: Why are we removing the footer cell and replacing with a cell that is made for articles?

I don't think we want the layout of articles tied to the layout of footer items.

bgerstle: It's basically at `WMF` "default cell" w/ an image & text view that's RTL compliant at this point.  I'd consider renaming rather than make a separate cell that does the same thing.

PR #380, opened 2016-01-14
Add pull to refresh

coreyfloyd:
Had to clean up some of the loading logic as well to make the refresh look right.


Wikipedia/Code/WMFExploreViewController.m
```diff
@@ -357,23 +363,58 @@ - (WMFFeaturedArticleSectionController*)featuredArticleSectionControllerForSchem
...
-    [self wmf_hideEmptyView];
     BOOL forceUpdate = self.sectionLoadErrors.count > 0;
-    self.sectionLoadErrors = [NSMutableDictionary dictionary];
-    [self.schemaManager update:forceUpdate];
+    [self updateSectionSchemaForce:forceUpdate];
+}
+
+- (void)updateSectionSchemaForce:(BOOL)force {
+    [self.refreshControl beginRefreshing];
+    self.refreshControl.attributedTitle = [[NSAttributedString alloc] initWithString:@"Updating…"];
```
montehurd: Need localized string for this I think.

montehurd:
Looking good! 

Just noticed a couple small things...

- If I pull to refresh more than once it seems like it's repeatedly refreshing instead of cancelling previously in-flight refreshes (or discarding new refresh request). On a fast connection a user would have to pull to refresh repeatedly quickly to see this issue (as I do), but on a slower connections this would be increasingly problematic.

- Maybe add slight delay when hiding the refresh spinner after a refresh is triggered? (Josh commented on this as well.)

- Perhaps unrelated, but the feed is sometimes jumping to the top of the screen now - I think my sim location is set to "City Drive" or some such, but regardless, if a user is trying to use the feed in the car and it jumps to the top it's they are going to be all :(

Wikipedia/Code/WMFExploreViewController.m
coreyfloyd: missed that one… thanks

montehurd:
One thing I didn't catch yesterday - "Last updated: never" isn't quite right, because the first time the feed loads it has technically been "updated". So maybe we could update the "last updated date" when the feed first loads on its own (w/o refreshing via pull) too?

coreyfloyd:
@montehurd. Not going to fuss over the copy here. It wasn't in the spec so will just put it out and let product decide if we need to update later.

montehurd:
Oh, I'm just saying update the date when the feed loads w/o pulling so the pull message is accurate. Should be easy.

montehurd:
Can still have the "never" message if the date isn't set in the case the feed never actually updated by either mechanism.

coreyfloyd:
> One thing I didn't catch yesterday - "Last updated: never" isn't quite right, because the first time the feed loads it has technically been "updated". 

> I'm just saying update the date when the feed loads w/o pulling so the pull message is accurate. Should be easy.

@montehurd yeah - its actually not updated yet - it just has the default stuff in there. when the user pulls or the user relaunches the app after 24h hours, it updates and sets a date.

PR #385, opened 2016-01-15
Fix crash when location permissions are reported

coreyfloyd:
https://phabricator.wikimedia.org/T123760

We weren't handling all the cases - more specifically we were asserting on the unhanded cases.


montehurd:
In the case that the user had denied the app location permissions, can we show the following messages (shown by 4.x in same case)?

"nearby-location-updates-denied" = "This app does not have permission to receive location updates.";

"nearby-location-updates-enable" = "To enable location updates open the Settings App and select the following:";

"nearby-location-updates-settings-menu" = "Privacy > Location Services > Wikipedia";

Although this is prob best done as separate ticket...

PR #390, opened 2016-01-20
Hiding sections

coreyfloyd:
New feature centers on the blacklist object. The blacklist serializes titles that should be excluded from the explore feed.

Aside from that it is all pretty standard, just KVOing on the black list and updating the feed when new titles are added.

I'd like to add some tests for the black list - I'll do a follow on tomorrow - just wanted to get this up for the beta.


bgerstle:
:+1: @coreyfloyd some technical follow-ons for this that we discussed:

- [ ] refactoring black list into an `MWKList` subclass (test coverage + DRY-ness)
  - [ ] Remove add'l mutation methods (e.g. `removeBlacklistItem:`) in favor of `MWKList` API (`removeEntry:`), which also removes `@synchronization` locking
  - [ ] Ensure shared blacklist is always mutated on main thread
    - [ ] _Optional:_ override `MWKList` methods to assert it's being called on the main thread

@montehurd can you please have a look if you haven't already?

montehurd:
Checking...

montehurd:
I hid one, then scrolled down a bit, and got a crash at "WMFRelatedSectionController configureCell:withObject:inTableView:atIndexPath:"... hmm...

montehurd:
Trace: 
https://gist.github.com/montehurd/2209ddb64ae4f2992a80

montehurd:
Was able to reproduce a second time. Odd because I had already scrolled past the nearby section when I hid an item, yet the crash didn't happen until I then scrolled down a little bit more, at which time it crashed and I saw this:

screen shot 2016-01-20 at 12 34 37 pm

coreyfloyd:
@bgerstle when you get in can you run this on your device and/or simulator to let me know if you are getting the crash that monte was getting. FWIW he isn't getting it any more. This crash he is seeing is the message being sent to NSNumber (in the relatedItemController). So I/we have seen it before - but may not specific to this pr.

bgerstle:
> I hid one, then scrolled down a bit, and got a crash at "WMFRelatedSectionController configureCell:withObject:inTableView:atIndexPath:"

Not seeing this on iPhone 6S

PR #392, opened 2016-01-20
Fix sharing for articles w/o images

bgerstle:
Phab: [T123662](https://phabricator.wikimedia.org/T123662)

---

`nil` URL results in cancelled image promise, meaning then/catch blocks aren't invoked:

image


bgerstle:
@coreyfloyd addressed by making invalid/empty URLs a "regular" error, which uncaught, log to the console:

> 2016-01-20 17:11:56:222 Wikipedia Debug[main] __56-[WMFShareOptionsController fetchImageThenShowShareCard]_block_invoke#L99 INFO:
Ignoring share card image error: Error Domain=WMFImageControllerErrorDomain Code=1 "(null)"

PR #393, opened 2016-01-20
Typography update for H2 and H3

nirzar:
H2 and H3 are in need of a dire update on constrast and spacing. this patch tries to solve all the issues. check PR comments for screenshots


nirzar:
Issues with the current version
artboard 1 copy

After the patch
artboard 1

artboard 1 copy 2

coreyfloyd:
@nirzar changes are looking good - i think removing the redundant pencils looks cleaner as long as you can still edit using the pencil above (as @montehurd was asking)

nirzar:
@coreyfloyd yeah i checked. you can still edit content of H3 with with H2 pencils

montehurd:
Will kick the tires on these in the morning...

coreyfloyd:
:+1:

montehurd:
Checking...

www/less/misc.less
```diff
@@ -82,3 +82,26 @@ a{
 * {
     -webkit-touch-callout: none !important;
 }
+
+//updating typography for h2 and h3
+.content h2 {
```
montehurd: Probably want to scope these more tightly to h2's which are definitely section headers... i.e. `h2.section_heading` instead of `.content h2`. Same below and for h3's.

www/less/misc.less
```diff
@@ -82,3 +82,26 @@ a{
...
+    font-weight: 700;
+    padding: 15px 0 0!important;
+}
+
+.content h3 a.edit_section_button {
+    display: none !important;
+}
+
+h2 a.edit_section_button,
+h3 a.edit_section_button {
```
montehurd: Maybe no need for 'h3 a.edit_section_button' here because those are being hidden on line 100?

www/less/misc.less
```diff
@@ -82,3 +82,26 @@ a{
 * {
     -webkit-touch-callout: none !important;
 }
+
+//updating typography for h2 and h3
+.content h2 {
+    padding: 20px 0 10px;
+    border-bottom: 1px dotted #eee;
```
montehurd: The dots are causing funky times for some section heading if the section heading text extends almost to the right edge of the screen. See `enwiki > History of Apple Inc > 1969-1984: Jobs and Wozniak` and other sections on that article.

If you want to just remove the dots for now until we can iron that out, I'd be fine merging the other changes if you want to do a quick commit to fix the other issues I noted.

montehurd:
@nirzar I made those tiny scoping tweaks I mentioned. Could you pull this down and double-check everything still looks ok?

nirzar:
@montehurd i think this is good to go :+1:  thanks

PR #396, opened 2016-01-21
Add accessibility to CompassView and NearbyTableViewCell

hhanke:
Treat the whole NearbyTableViewCell as one accessibility element and add distance and direction descriptions to its label.

### DIFF
--- a/Wikipedia/Code/WMFCompassView.m
+++ b/Wikipedia/Code/WMFCompassView.m
@@ -223,4 +223,17 @@ static CGFloat const WMFCompassOppositeLineWidth = 2.0;
     }
 }
 
+#pragma mark - Accessibility
+
+- (BOOL)isAccessibilityElement {
+    return YES;
+}
+
+- (NSString *)accessibilityLabel {
+    NSInteger clockDirection = WMFRadiansToClock(self.angleRadians);
+    NSString* label = MWLocalizedString(@"compass-direction", nil);
+    label = [label stringByReplacingOccurrencesOfString:@"$1" withString:[NSString localizedStringWithFormat:@"%@", @(clockDirection)]];
+    return label;
+}
+
 @end
--- a/Wikipedia/Code/WMFNearbyArticleTableViewCell.m
+++ b/Wikipedia/Code/WMFNearbyArticleTableViewCell.m
@@ -201,4 +201,20 @@
     [self.articleImageView wmf_setImageWithMetadata:image detectFaces:YES];
 }
 
+#pragma mark - Accessibility
+
+- (BOOL)isAccessibilityElement {
+    return YES;
+}
+
+- (NSString*)accessibilityLabel {
+    NSString *titleAndDescription;
+    if (self.descriptionText) {
+        titleAndDescription = [NSString stringWithFormat:@"%@, %@", self.titleText, self.descriptionText];
+    } else {
+        titleAndDescription = self.titleText;
+    }
+    return [NSString stringWithFormat: @"%@, %@ %@", titleAndDescription, self.distanceLabel.text, self.compassView.accessibilityLabel];
+}
+
 @end
--- a/Wikipedia/Code/WMFMath.h
+++ b/Wikipedia/Code/WMFMath.h
@@ -54,6 +54,20 @@ extern double RoundWithPrecision(double (* rounder)(double), double x, unsigned 
  */
 extern double WMFFlooredPercentage(double x) __attribute__((const)) __attribute__((pure));
 
+/**
+ *  Convert angle in radians into hours on a clockface.
+ *
+ *  If we imagine a clockface instead of a unit circle (the hour 12 is aligned with 0 radians),
+ *  then we set the clock pointer to the given the degree, this function returns the number
+ *  of the nearest hour.
+ *
+ *  @param radians Angle in radians
+ *
+ *  @return NSInteger representing the corresponding reading on the clockface.
+ */
+
+NSInteger WMFRadiansToClock(double radians);
+
 ///
 /// @name Geometry
 ///
--- a/Wikipedia/Code/WMFMath.m
+++ b/Wikipedia/Code/WMFMath.m
@@ -10,3 +10,12 @@ double WMFFlooredPercentage(double x) {
     return WMFRoundedPercentage(floor, x, 100);
 }
 
+NSInteger WMFRadiansToClock(double radians) {
+    double unitRadians = fmod(radians, 2*M_PI);
+    double positiveRadians = unitRadians >= 0 ? unitRadians : unitRadians + 2*M_PI;
+    NSInteger clockDirection = lroundf(positiveRadians / (M_PI / 6));
+    if (clockDirection == 0) {
+        clockDirection = 12;
+    }
+    return clockDirection;
+}
--- a/Wikipedia/Localizations/en.lproj/Localizable.strings
+++ b/Wikipedia/Localizations/en.lproj/Localizable.strings
@@ -403,3 +403,6 @@
 
 "close-button-accessibility-label" = "Close";
 "back-button-accessibility-label" = "Back";
+
+"compass-direction" = "at $1 o'clock";
+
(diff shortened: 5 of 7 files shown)


Wikipedia/Code/WMFCompassView.m
```diff
@@ -223,4 +223,21 @@ - (void)drawCompassLinesInContext:(CGContextRef)ctx
...
 }
 
+#pragma mark - Accessibility
+
+- (BOOL)isAccessibilityElement {
+    return YES;
+}
+
+- (NSString *)accessibilityLabel {
+    double positive_angle = self.angleRadians >= 0 ? self.angleRadians : self.angleRadians + 2*3.14;
```
bgerstle: please use math constants, e.g. `2_M_PI` for 2 * pi.

Wikipedia/Code/WMFCompassView.m
```diff
@@ -223,4 +223,21 @@ - (void)drawCompassLinesInContext:(CGContextRef)ctx
...
+    NSInteger clock_direction = lroundf(positive_angle / (3.14 / 6));
+    if (clock_direction == 0) {
+        clock_direction = 12;
+    }
+    NSString* label = MWLocalizedString(@"compass-direction", nil);
+    label = [label stringByReplacingOccurrencesOfString:@"$1" withString:[NSString localizedStringWithFormat:@"%@", @(clock_direction)]];
```
bgerstle: i think the `localizedStringWithFormat` is unnecessary here, since all the values are "**N** o'clock" and don't vary based on units of measurement or 12/24-hour time.

bgerstle: also, do you have any sense for whether the "o'clock" convention is used across languages?  for example, how would you convey directionality in Hebrew, German, or Spanish?

coreyfloyd: yeah this may be a time where we want to use the base units of degrees which is universal. Though not as nice in english clock positions, it should be understood by everyone.

Wikipedia/Code/WMFNearbyArticleTableViewCell.m
```diff
@@ -197,4 +197,20 @@ - (void)setImage:(MWKImage*)image {
...
+}
+
+- (NSString*)accessibilityLabel {
+    NSString *titleAndDescription;
+    if (self.descriptionText) {
+        titleAndDescription = [NSString stringWithFormat:@"%@, %@", self.titleText, self.descriptionText];
+    } else {
+        titleAndDescription = self.titleText;
+    }
+    return [NSString stringWithFormat: @"%@, %@ %@", titleAndDescription, self.distanceLabel.text, self.compassView.accessibilityLabel];
```
bgerstle: maybe a minor point, but shouldn't the accessibility label for the distance say the full unit name, i.e. meters/feet instead of m/ft?

Wikipedia/Code/WMFCompassView.m
```diff
@@ -223,4 +223,21 @@ - (void)drawCompassLinesInContext:(CGContextRef)ctx
...
 
+#pragma mark - Accessibility
+
+- (BOOL)isAccessibilityElement {
+    return YES;
+}
+
+- (NSString *)accessibilityLabel {
+    double positive_angle = self.angleRadians >= 0 ? self.angleRadians : self.angleRadians + 2*3.14;
+    NSInteger clock_direction = lroundf(positive_angle / (3.14 / 6));
```
bgerstle: would you mind writing unit tests for these calculations? e.g. that some example angles return the expected "clock" orientation

coreyfloyd: see my not above about degrees before investing time in a unit test.

Wikipedia/Code/WMFCompassView.m
bgerstle: i did a bit of research by turning on VoiceOver and using the built-in Compass app. It uses degrees for all languages I tested (English, Spanish, & Chinese), plus the initials the cardinal direction, e.g. 145 degrees, SE (it says "S-E" not "Southeast").  So I'd suggest:

- Converting radians to degrees (using `RADIANS_TO_DEGREES`)
- Optionally, adding cardinal direction (N, NE, E, SE, S, etc.)
- Localizing the label using the format: "$1 degrees" and, optionally, combining that w/ another localized string for the direction initials (i.e. `MWLocalizedString(@"west-initial", nil)` would be "W" in English, and "O" in Spanish)

bgerstle: Side note, something interesting about our "compass" view is that it doesn't even tell you where North is. In fact, the "arrow" of our compass points in the direction of the object, not North (as is the convention, AFAIK).

bgerstle: @Hhanke @coreyfloyd for now, let's just use degrees and we can follow-on the cardinal direction later if needed

hhanke: Regarding the clock direction system, this is widely used among blind people also outside of English speaking countries. I can definitely confirm that for my country, the Czech Republic. 

It is being used for description of direction in orientation (e.g. when you describe to a blind person where objects are around him), in navigation (e.g. the Blindsquare iOS app) or e.g. to describe the position of items on someones plate while eating.

But Brian is right, we need to check how this is handled in more remote countries like China. I get a feeling that the clockface is very widely known. I will try to check it and I will write another comment here.

I think at least in countries where it is routinely used among blind people, it is much superior to giving degrees in a situation like this. Degrees carry too much detail and are complicated to imagine. I do not know how about you, but I cannot intuitively (without some thinking) realize what a direction like like 210 degrees, 72 degrees or 315 degrees actually mean around me.

The Compass app is not a good example. In that case, the aim is to give an exact direction and that is the principal purpose of the app. In the Wikipedia app, if I understand correctly, the main purpose is giving the users easy access to articles containing information about their surroundings. The information about distance and direction in this later case serve another purpose and should be easily understandable.

So maybe we will have to differentiate based on the current language settings and maybe default to degrees where a better system is not available.

For the same reason, I also think that the cardinal direction is not really necessary, although it could be a nice addition and might also be relevant in some cases.

Regarding the first comment, isn't localizedStringWithFormat necessary for languages which use other characters for numerals, e.g. Japanese?

Thanks for the other comments as well!

bgerstle: > Regarding the first comment, isn't localizedStringWithFormat necessary for languages which use other characters for numerals, e.g. Japanese?

Oh, didn't consider that... we might need to do that in some other places as well.  For example, using the Swift REPL to print a localized string arabic numeral(?):

```
 8> NSString(format: "%@", locale:NSLocale(localeIdentifier:"ar"), NSNumber(int: 1))
$R6: NSString = "١"
  9> NSString(format: "%d", locale:NSLocale(localeIdentifier:"ar"), 1)
$R7: NSString = "١"
```

bgerstle:
@hhanke i've sent an email to some mailing lists asking for advice on the issue. in the meantime can we go with whatever is simplest to implement?  i agree with you that degrees and cardinal directions aren't ideal for this use case, so maybe we can go w/ clock faces while we look into the issue further?

hhanke:
Thank you Brian! I have also sent some questions to people whom I think might know this. Should we then wait a few days to see what will be the feedback?

bgerstle:
@hhanke we could, but IIRC there are other worthwhile changes in here that don't need to be delayed.  even if the compass itself isn't accessible yet (or says degrees, or something) it will be an improvement.  either way, I'd prefer to not leave this open while the issue is pending (we can have a ticket on the backlog)

hhanke:
@bgerstle OK, I will prepare it. We know that for people from some countries, the clockface is right. If it turns out that we need to do more from speakers from other countries, we can always improve it.

bgerstle:
> we can always improve it

that's the idea 😊

hhanke:
I have made the conversion from radians to degrees more robust, moved it to WMFMath and provided a test for it. I have also used the M_PI constant and fixed the naming of variables. It would be nicer to use "20 meters" instead of "20 m", but presently, it has the same effect as VoiceOver expands measurements shortcuts during TTS anyway. So it does in fact read "20 meters" to the user. Therefore, I didn't change this part and maybe it falls into the category of what can be improved later on :-)

bgerstle:
👍🏻

coreyfloyd:
looking good - thanks @hhanke!

PR #397, opened 2016-01-21
Add accessibility labels to some elements which were missing them.

hhanke:
The elements are:
- Title of Explore view
- Close button in References view
- Back button from editor
- Table of Contents button in Article view
- Icon for rows in Settings view (this was removed from accessibility elements as it seems to serve no independent purpose but illustration)

Also handle traits and isAccessibilityElement correctly for these elements.

### DIFF
--- a/Wikipedia/Code/WMFExploreViewController.m
+++ b/Wikipedia/Code/WMFExploreViewController.m
@@ -93,6 +93,9 @@ NS_ASSUME_NONNULL_BEGIN
     self = [super initWithCoder:aDecoder];
     if (self) {
         self.navigationItem.titleView          = [[UIImageView alloc] initWithImage:[UIImage imageNamed:@"W"]];
+        self.navigationItem.titleView.isAccessibilityElement = YES;
+        self.navigationItem.titleView.accessibilityLabel = MWLocalizedString(@"home-accessibility-label", nil);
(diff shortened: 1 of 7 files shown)


Wikipedia/Code/WMFExploreViewController.m
```diff
@@ -92,6 +92,9 @@ - (nullable instancetype)initWithCoder:(NSCoder*)aDecoder {
     self = [super initWithCoder:aDecoder];
     if (self) {
         self.navigationItem.titleView          = [[UIImageView alloc] initWithImage:[UIImage imageNamed:@"W"]];
+        self.navigationItem.titleView.isAccessibilityElement = YES;
```
bgerstle: should this also be done on the tab bar item?

hhanke: I don't think so, the tab bar items are accessible already. The title is not, because it is an `UIImageView`.

bgerstle: Ah you're right, `UITabBarItem` does that automatically

bgerstle:
@hhanke please resolve merge conflicts so i can merge this :-)

hhanke:
I have rebased it and it should be OK now :-)

PR #398, opened 2016-01-21
Add and correct accessibility traits for Search and Settings

hhanke:
Accessibility trait SearchField is added to the search field in the Search view and accessibility traits are fixed/added in the Search view.


bgerstle:
:+1: do you also plan to add accessibility ID's, labels, and titles? if so, please consult our docs on [working with localized strings](https://github.com/wikimedia/wikipedia-ios/blob/master/docs/localization.md).

PR #401, opened 2016-01-21
fix overflow & random blue colors, remove redundant blue color

bgerstle:
New, matching colors: 

image

image


bgerstle:
ping @montehurd @nirzar

PR #402, opened 2016-01-21
convert blacklist mwklist subclass

coreyfloyd:
They just had so much in common!


Wikipedia/Code/WMFRelatedSectionBlackList.m
```diff
@@ -1,10 +1,18 @@
...
 #import "WMFRelatedSectionBlackList.h"
+#import "MWKList+Subclass.h"
 
 static NSString* const WMFRelatedSectionBlackListFileName      = @"WMFRelatedSectionBlackList";
 static NSString* const WMFRelatedSectionBlackListFileExtension = @"plist";
 
+@implementation MWKTitle (MWKListObject)
+
+- (id <NSCopying, NSObject>)listIndex {
+    return self;
```
bgerstle: can this be moved into a separate category? wouldn't think to find this here

bgerstle: also surprised this wasn't done already for the other lists

Wikipedia/Code/WMFRelatedSectionBlackList.m
```diff
@@ -51,70 +60,20 @@ + (instancetype)loadFromDisk {
 
 - (void)addBlackListTitle:(MWKTitle*)title {
```
bgerstle: now this class has two mutation APIs: one thread-safe and one not.  it would be simpler and less error-prone if `addEntry:`, `removeEntry:`, etc. were overridden to be thread safe, instead of creating a thread safe API on top of it.  you could even go so far as to make a `MWKAtomicList` subclass which does this, and have `WMFBlacklist` subclass that.

bgerstle: that said, I thought we had talked about removing the locking mechanisms since this class is only going to be mutated on the main thread anyway. i'm ambivalent about whether the locking stays or goes, but either way I think it's better if the API surface area is kept to a minimum

WikipediaUnitTests/Code/WMFBlacklistTests.m
```diff
@@ -0,0 +1,53 @@
...
+- (void)tearDown {
+    WMFRelatedSectionBlackList* bl = [[WMFRelatedSectionBlackList alloc] init];
+    [bl removeAllEntries];
+    [bl save];
+    [super tearDown];
+}
+
+- (void)testPersistsToDisk {
+    PushExpectation();
+    MWKTitle* title                = [[MWKTitle alloc] initWithString:@"some-title" site:[MWKSite siteWithCurrentLocale]];
```
bgerstle: `+[MWKTitle random]` is provided to reduce this kind of boilerplate

WikipediaUnitTests/Code/WMFBlacklistTests.m
```diff
@@ -0,0 +1,53 @@
...
+    WMFRelatedSectionBlackList* bl = [[WMFRelatedSectionBlackList alloc] init];
+    [bl addEntry:title];
+    [bl save].then(^(){
```
bgerstle: `XCTestCase+PromiseKit.h` is provided to simplify async testing w/ Promises—no need to push/pop/wait for expectations.

WikipediaUnitTests/Code/WMFBlacklistTests.m
```diff
@@ -0,0 +1,53 @@
...
+        [self popExpectationAfter:nil];
+    }).catch(^(NSError* error){
+        XCTFail(@"Error callback erroneously called with error %@", error);
+    });
+    WaitForExpectations();
+
+    bl = [WMFRelatedSectionBlackList loadFromDisk];
+    MWKTitle* first = [[bl entries] firstObject];
+
+    XCTAssertTrue([title isEqual:first],
```
bgerstle: use a matcher, obviates the need for a hand-written description. doesn't matter if you use Hamcrest or Nimble

Wikipedia/Code/WMFRelatedSectionBlackList.m
```diff
@@ -26,22 +34,23 @@ + (instancetype)sharedBlackList {
...
 }
 
-+ (void)saveToDisk:(WMFRelatedSectionBlackList*)blackList {
-    if (![NSKeyedArchiver archiveRootObject:blackList toFile:[[self fileURL] path]]) {
-        //TODO: not sure what to do with an error here
-        DDLogError(@"Failed to save sections to disk!");
+- (void)performSaveWithCompletion:(dispatch_block_t)completion error:(WMFErrorHandler)errorHandler {
+    @synchronized(self) {
+        if (![NSKeyedArchiver archiveRootObject:self toFile:[[[self class] fileURL] path]]) {
+            //TODO: not sure what to do with an error here
```
bgerstle: that's up to the `save` caller to decide, nothing _TODO_ here

bgerstle:
@coreyfloyd also, please add a subclass of `MWKListSharedTests` which exercises shared functionality of `MWKBlacklist` (see `MWKHistoryListSharedTests` for an example).

Wikipedia/Code/WMFRelatedSectionBlackList.m
coreyfloyd: I didn't want to put it into a separate category since it isn't used anywhere else. It only pertains to MWKList and doesn't need to be explored to the outside world.Its a common to only define categories in the implementation they are used.

Wikipedia/Code/WMFRelatedSectionBlackList.m
coreyfloyd: I moved the code into `addEntry:`, et al. But as far as the locking itself - I wanted to leave it in there because it is inconsequential. Since this class only gets one thing added on occasion, the locking is fairly cheap.

coreyfloyd:
> @coreyfloyd also, please add a subclass of MWKListSharedTests which exercises shared functionality of MWKBlacklist (see MWKHistoryListSharedTests for an example).

I didn't add anything to MWKListSharedTests intentionally. Mentioned this last time MWKListSharedTests came up in testing: We need refactor this to make it more understandable to both Xcode and to other developers.

Just like we prefer composition over subclassing in our app code base - so that functionality doesn't get hidden and the architecture isn't brittle -  we should do the same with the test code.

Aside from being hard to grok for me, Xcode has trouble navigating it as well and has trouble showing you were the test broke and re-running tests from the Xcode UI doesn't work either.

Subclassing can be ok if we want to do something like WMFAsyncTestCase (shared functionality), but putting assertions themselves in super classes I think is the key issue here. Another way to say it is that shred code should be like a library (call if you need it) as opposed to a framework (your code is called when the framework chooses)

WikipediaUnitTests/Code/WMFBlacklistTests.m
coreyfloyd: So… hamcrest is terrible as I have said before - that why i don't use that in general - and you can't add quick/nimble without ruining code completion and showing errors (since you have to use @import to get it to work). So I choose to use the built in stuff instead.

Wikipedia/Code/WMFRelatedSectionBlackList.m
bgerstle: :ok:

Wikipedia/Code/WMFRelatedSectionBlackList.m
bgerstle: :ok:

WikipediaUnitTests/Code/WMFBlacklistTests.m
bgerstle: > without ruining code completion and showing errors (since you have to use @import to get it to work)

I've run into this issue as well. I filed a radar for it, but the workaround is to disable "Gather code coverage" in the scheme's "Test" action options and then build for testing.  this should regenerate the missing indexes and remove the errors (which are actually indexing errors, not compilation).

bgerstle:
RE: `MWKListSharedTests`, I knew there would be some issues w/ XCTest, but I guess I underestimated the usability issues.  That, or I figured it couldn't be worse that XCTest already is, since I regularly have issues running specific tests even w/ plain `XCTestCase` subclasses.  

However, at this point the amount of work it would take to refactor the assertions out of the base class is probably about the same it would take to port it to use Quick's "shared examples"—which are already being used in the gallery data source tests.  IMO Quick's shared examples are the best way to do this sort of thing, but I understand if you don't want to port it just yet.  Leaving the blacklist uncovered wouldn't be the end of the world as `MWKList` is already covered, but I would much prefer having those behaviors covered for the blacklist as well.

coreyfloyd:
> @coreyfloyd RE: MWKListSharedTests, I knew there would be some issues w/ XCTest, but I guess I underestimated the usability issues. That, or I figured it couldn't be worse that XCTest already is, since I regularly have issues running specific tests even w/ plain XCTestCase subclasses.

> However, at this point the amount of work it would take to refactor the assertions out of the base class is probably about the same it would take to port it to use Quick's "shared examples"—which are already being used in the gallery data source tests. IMO Quick's shared examples are the best way to do this sort of thing, but I understand if you don't want to port it just yet. Leaving the blacklist uncovered wouldn't be the end of the world as MWKList is already covered, but I would much prefer having those behaviors covered for the blacklist as well.

@bgerstle Just to document our hangout conversation - we both agree that the above is the way forward (Quick) - so we need to adda a engineering ticket maybe to refactor these tests (and flip the setting on Xcode to make Quick/Nimble play nice with Xcode)? And as far as coverage now - since the black list doesn't really add any functionality (just some domain specific API), I am satisfied that is well covered with the other base list tests.

montehurd:
Deferring to you guys on this one :+1:

PR #404, opened 2016-01-21
TOC on left for RTL layout direction.

montehurd:
When layout direction is RTL:
- [x] TOC appears on left side of screen.
- [x] Progressive drag from left edge.
- [x] When TOC is onscreen drag toward left edge to dismiss.
- [x] When TOC is onscreen tap on article view to dismiss.
- [x] Dynamically switch side TOC appears on based on device interface layout direction
- [x] Ensure iOS 8 doesn't explode.

----

**RTL** *New!*
toc rtl mov

----

**LTR** *Still works!*
toc ltr mov

### DIFF
--- a/Wikipedia.xcodeproj/xcshareddata/xcschemes/Wikipedia.xcscheme
+++ b/Wikipedia.xcodeproj/xcshareddata/xcschemes/Wikipedia.xcscheme
@@ -92,8 +92,7 @@
       ignoresPersistentStateOnLaunch = "NO"
       debugDocumentVersioning = "YES"
       debugServiceExtension = "internal"
-      allowLocationSimulation = "YES"
-      language = "he">
+      allowLocationSimulation = "YES">
       <BuildableProductRunnable
          runnableDebuggingMode = "0">
          <BuildableReference
--- a/Wikipedia.xcodeproj/xcshareddata/xcschemes/Wikipedia.xcscheme
+++ b/Wikipedia.xcodeproj/xcshareddata/xcschemes/Wikipedia.xcscheme
@@ -92,7 +92,8 @@
       ignoresPersistentStateOnLaunch = "NO"
       debugDocumentVersioning = "YES"
       debugServiceExtension = "internal"
-      allowLocationSimulation = "YES">
+      allowLocationSimulation = "YES"
+      language = "he">
       <BuildableProductRunnable
          runnableDebuggingMode = "0">
          <BuildableReference
--- a/Wikipedia/Code/WMFTableOfContentsAnimator.swift
+++ b/Wikipedia/Code/WMFTableOfContentsAnimator.swift
@@ -185,8 +185,8 @@ public class WMFTableOfContentsAnimator: UIPercentDrivenInteractiveTransition, U
             self.presentingViewController?.presentViewController(self.presentedViewController!, animated: true, completion: nil)
         case (.Changed):
             let position = gesture.locationInView(gesture.view);
-            let distance = WikipediaAppUtils.isDeviceLayoutDirectionRTL() ? position.x : CGRectGetMaxX(gesture.view!.bounds) - position.x
-            let transitionProgress = distance / CGRectGetMaxX(gesture.view!.bounds)
+            let distanceFromSide = WikipediaAppUtils.isDeviceLayoutDirectionRTL() ? position.x : CGRectGetMaxX(gesture.view!.bounds) - position.x
+            let transitionProgress = distanceFromSide / CGRectGetMaxX(gesture.view!.bounds)
             self.updateInteractiveTransition(transitionProgress)
         case (.Ended):
             self.isInteractive = false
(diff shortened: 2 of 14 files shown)


Wikipedia.xcodeproj/xcshareddata/xcschemes/Wikipedia.xcscheme
```diff
@@ -92,7 +92,8 @@
       ignoresPersistentStateOnLaunch = "NO"
       debugDocumentVersioning = "YES"
       debugServiceExtension = "internal"
-      allowLocationSimulation = "YES">
+      allowLocationSimulation = "YES"
```
bgerstle: please revert this. it might be easier to use the dedicated RTL scheme to avoid issues w/ making the regular scheme RTL

montehurd: This is just temporary.

montehurd: The next commit will make the flipping happen automatically and this line will be removed. I just left it in case you wanted to check this commit out for testing so it would work as-is.

montehurd:
:-1:  

Just found an iOS 8 edge-drag bug. Will fix either tonight or in the morning.

montehurd:
Fixed! Thumbs up on own ticket? :+1:

coreyfloyd:
:+1:  looking good

Wikipedia/Code/WMFTableOfContentsAnimator.swift
```diff
@@ -35,6 +35,13 @@ public class WMFTableOfContentsAnimator: UIPercentDrivenInteractiveTransition, U
     
     private(set) public var isInteractive: Bool
     
+    static func shouldMoveTOCToLeft() -> Bool {
+        return WikipediaAppUtils.isDeviceLayoutDirectionRTL() && !NSProcessInfo.processInfo().wmf_isOperatingSystemVersionLessThan9_0_0()
+    }
+    
+    private func rtlMultiplier() -> CGFloat {
```
bgerstle: you can make this a `var` and omit all the parens

bgerstle: or, make both this and `shouldMoveTOCToLeft` as extensions on `UIApplication`:

``` swift
extension UIApplication {
  var wmf_tocShouldBeOnLeft: Bool {
    get { return self.wmf_isRTL && NSProcessInfo.processInfo().wmf_isOperatingSystemMajorVersionAtLeast(9) }
  }

  var wmf_tocRTLMultiplier: Int {
    get {
      return self.wmf_isRTL ? 1.0 : 1.0
    }
  }

  var wmf_isRTL: Bool {
    return self.userInterfaceLayoutDirection == UIUserInterfaceLayoutDirectionRTL
  }
}
```

Wikipedia/Code/WMFTableOfContentsAnimator.swift
```diff
@@ -182,16 +189,16 @@ public class WMFTableOfContentsAnimator: UIPercentDrivenInteractiveTransition, U
...
             self.updateInteractiveTransition(transitionProgress)
         case (.Ended):
             self.isInteractive = false
-            let velocityRequiredToPresent = -CGRectGetWidth(gesture.view!.bounds)
+            let velocityRequiredToPresent = -CGRectGetWidth(gesture.view!.bounds) * rtlMultiplier()
             let velocityRequiredToDismiss = CGRectGetWidth(gesture.view!.bounds)
             
             let velocityX = gesture.velocityInView(gesture.view).x
-            if velocityX > velocityRequiredToDismiss{
+            if velocityX * rtlMultiplier() > velocityRequiredToDismiss{
```
bgerstle: can you double check this? if the velocity is -800 and the threshold is -500, the threshold _should_ be exceeded, but -800 is **less than** -500, not greater than

bgerstle: IOW the spec should be:

if I'm swiping towards the **presenting edge** and the velocity if above X, cancel the transition

Wikipedia/Code/WMFTableOfContentsAnimator.swift
montehurd: Done!

Wikipedia/Code/WMFTableOfContentsAnimator.swift
montehurd: Tested and it appears correct. This case is only fallen into if you (LTR) begin dragging the TOC in from the right edge, then quickly flick it *back toward* the right edge, or if RTL begin dragging the TOC in from the left edge then quickly flick it back toward the left edge. The existing code handles both cases in the same logical way as the old code did for LTR.

coreyfloyd:
I'm still good - will let @bgerstle check in and merge it

PR #405, opened 2016-01-22
Fix Explore scrolling issues

bgerstle:
Phab: [T124424](https://phabricator.wikimedia.org/T124424), [T123248](https://phabricator.wikimedia.org/T123248)

---

Before: https://vid.me/aB0l
After: https://vid.me/mXTd

## Problem

Our Explore `UITableView` was estimating all rows at the same height, resulting in a lot of extra room at the bottom of the table before content was loaded and actual row heights were calculated.  In addition, we weren't able to calculate row heights quickly because the CPU was overloaded by redundantly rendering the same content to the screen when offscreen section data were downloaded.

## Solution

Made each section controller dictate its estimated row height.  This makes the estimated height of the entire table much more accurate, so there's less jumpiness after content loads.  However, this didn't appear to be enough to fully resolve the issue, as noted in the second part of the problem.  Even though the esitmates were more accurate, calculating the actual row heights & rendering the currently visible cells was delayed due to CPU cost of downloading section content which wasn't even going to be displayed.  To address this, the lazy fetching is now deferred by 250ms and then only triggered if the section is still visible.  This is similar to the approach we take with rate-limiting search queries:

- When a section is about to be come visible (`tableView:willDisplayCell:forRowAtIndexPath:`) we defer a call to `fetchSectionIfShowing:` for the section (if it's a fetching section)
  - If the section is scrolled away (`tableView:didEndDisplayingCell:forRowAtIndexPath:`), we cancel the deferred call to `fetchSectionIfShowing:` for that section
- When the deferred `fetchSectionIfShowing:` call fires **and** the section is still visible, that section is sent a `fetchDataIfNeeded` message

This prevents sections from fetching their data when the user scrolls quickly past them (in either direction), while still allowing them to be fetched almost instantly if the user remains in that section.

### DIFF
--- a/Wikipedia/Code/WMFAppViewController.m
+++ b/Wikipedia/Code/WMFAppViewController.m
@@ -134,10 +134,7 @@ static dispatch_once_t launchToken;
 }
 
 - (void)configureExploreViewController {
-    self.exploreViewController.searchSite  = [self.session searchSite];
-    self.exploreViewController.dataStore   = self.session.dataStore;
-    self.exploreViewController.savedPages  = self.session.userDataStore.savedPageList;
-    self.exploreViewController.recentPages = self.session.userDataStore.historyList;
+    [self.exploreViewController setSearchSite:[self.session searchSite] dataStore:self.dataStore];
 }
 
 - (void)configureArticleListController:(WMFArticleListTableViewController*)controller {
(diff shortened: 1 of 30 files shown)


coreyfloyd:
Whhheeeeew doggie! :100: :fire:

bgerstle:
@coreyfloyd @montehurd i was able to demonstrate the bug & fix manually, but I couldn't come up with a profile trace that easily showed the quantitative change in performance.  Neither the CPU nor FPS are significantly different, and unfortunately the AutoLayout instrument doesn't support iOS.  IMHO the fix is self evident and extra effort to do more profiling (e.g. comparing Time Profiler samples to see change in time spent) isn't that worth it.

coreyfloyd:
@bgerstle yeah I think you are good. You would have to do some filtering on the calls to see a difference - which can be a bit tedious. Like you said, the fix is self evident.

montehurd:
@bgerstle Very nice work! :+1:

PR #407, opened 2016-01-22
Add info button to modal image gallery

bgerstle:
Phab: [T100335](https://phabricator.wikimedia.org/T100335), [T124491](https://phabricator.wikimedia.org/T124491)

---

Adds an info button in the top right which opens an external URL to the currently-shown image's file page (shown here in iOS 8):

gallery-info-demo


bgerstle:
@nzr, I used the built-in `UIButton` type which looks identical to the mock. As a result, I've removed the "info" image from the app until we need an actual `UIImage` of that for something—which is unlikely.

Wikipedia/Code/ErrorWorkaround.swift
```diff
@@ -0,0 +1,27 @@
+//
+//  ErrorWorkaround.swift
```
coreyfloyd: Is there a better name for this, or some comments? I'm not sure what is being worked around.

Also It looks like you are extending `WMFImageController` - is this where that extension should be? We haven't really discussed swift style - so idk.

Wikipedia/Code/WMFModalImageGalleryViewController.m
```diff
@@ -478,6 +525,13 @@ - (void)updateImageForCell:(WMFImageGalleryCollectionViewCell*)cell
         [self setPlaceholderImage:download.image ofInfo:infoForImage forCellAtIndexPath:indexPath];
     }]
     .catch(^(NSError* error) {
+        @strongify(self);
+        BOOL const isInvalidURL =
```
coreyfloyd: So we are frequently attempting to load invalid urls here? What's the cause of that?

Wikipedia/Code/ErrorWorkaround.swift
bgerstle: I'll move the comments i wrote in `WMFImageController` here, but it's basically something weird going on with casting swift `ErrorType` objects to `NSError`.  this is only extending `WMFImageController` because it's "ground zero" for this issue (a Swift class with custom errors that we're referencing in ObjC).

Wikipedia/Code/WMFModalImageGalleryViewController.m
bgerstle: will double check on this, could be cases where we couldn't fetch the high-res URL (or metadata) for an image due to filename parsing issues

bgerstle: @coreyfloyd ah, i know why: we use the same code path for setting gallery placeholders regardless of whether the image info is ready, which results in the image info being `nil`, which results in the high-res being `nil`, which results in the banner (invalid/empty URL).

Wikipedia/Code/WMFPageCollectionViewController.m
```diff
@@ -19,19 +19,18 @@ - (void)primitiveSetCurrentPage:(NSUInteger)page;
...
-        NSAssert(currentPageAttributes,
-                 @"Layout attributes for current page were nil because %@ was called too early!",
-                 NSStringFromSelector(_cmd));
-        [self.collectionView setContentOffset:currentPageAttributes.frame.origin animated:animated];
-    }
+    NSParameterAssert(self.isViewLoaded);
+    // check page bounds here in case it was set before view was loaded
+    NSParameterAssert(self.currentPage < [self.collectionView.dataSource collectionView:self.collectionView numberOfItemsInSection:0]);
+
+    // can't use scrollToItem because it doesn't handle post-rotation scrolling well on iOS 6
```
montehurd: Can remove this comment now?

montehurd:
Tested on iOS 8.4 - seems fine.

coreyfloyd:
:+1:

Wikipedia/Code/WMFPageCollectionViewController.m
bgerstle: we're still using `setContentOffset:animated:`, but can investigate using `scrollToItem:animated:` later. i'll add a tech debt warning

PR #410, opened 2016-01-24
Fix T124593: Implement accessibility escape

dusek:
This is implemented in:
- **Settings** (alternative to Close button)
- **Search** (alternative to Close button)
- **Languages** (alternative to Close button)
- **Image viewer** (alternative to Close button)
- **Table of Contents** (currently not possible in any other way)
- **Section Editor** (alternative to Back button; truth to be told,
                      Section Editor cannot be entered with VoiceOver
                      and Switch Control currently; but that is for
                      another bug)

Tested with VoiceOver and Switch Control on iOS 9.2.1 (13D15).

Fixes [T124593](https://phabricator.wikimedia.org/T124593)

Patch provided by [A11Y LTD.](http://accessibility.expert)

### DIFF
--- a/Wikipedia/Code/WMFSearchViewController.m
+++ b/Wikipedia/Code/WMFSearchViewController.m
@@ -278,9 +278,18 @@ static NSUInteger const kWMFMinResultsBeforeAutoFullTextSearch = 12;
 
 #pragma mark - Dismissal
 
-- (IBAction)didTapCloseButton:(id)sender {
+- (void)dismiss {
     [self.searchField resignFirstResponder];
     [self dismissViewControllerAnimated:YES completion:nil];
+}
+
+- (IBAction)didTapCloseButton:(id)sender {
+    [self dismiss];
+}
+
+- (BOOL)accessibilityPerformEscape {
+    [self dismiss];
+    return YES;
 }
 
 #pragma mark - UITextFieldDelegate
--- a/Wikipedia/Code/LanguagesViewController.m
+++ b/Wikipedia/Code/LanguagesViewController.m
@@ -304,4 +304,11 @@ static NSString* const LangaugesSectionFooterReuseIdentifier = @"LanguagesSectio
     [searchBar resignFirstResponder];
 }
 
+#pragma mark - UIAccessibilityAction
+
+- (BOOL)accessibilityPerformEscape {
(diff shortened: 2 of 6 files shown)


Wikipedia/Code/WMFSearchViewController.m
```diff
@@ -278,9 +278,18 @@ - (void)setSeparatorViewHidden:(BOOL)hidden animated:(BOOL)animated {
 
 #pragma mark - Dismissal
 
-- (IBAction)didTapCloseButton:(id)sender {
+- (BOOL)didRequestDismiss {
```
coreyfloyd: "did" usually indicates a delegate method. This is just an internal command - so simply `- (void)dismiss` would be an appropriate name.

Wikipedia/Code/WMFSearchViewController.m
```diff
@@ -278,9 +278,18 @@ - (void)setSeparatorViewHidden:(BOOL)hidden animated:(BOOL)animated {
...
     [self.searchField resignFirstResponder];
     [self dismissViewControllerAnimated:YES completion:nil];
+    return YES; // we did dismiss the controller
```
coreyfloyd: Since this is always YES and this is only used for the `accessibilityPerformEscape` method it would be better to make this void and just move it below:

```
- (BOOL)accessibilityPerformEscape {
    [self dismiss];
    return YES;
  }
```

Wikipedia/Code/WMFArticleContainerViewController+TOC.swift
```diff
@@ -35,8 +35,9 @@ extension WMFArticleContainerViewController : WMFTableOfContentsViewControllerDe
         }
     }
 
-    public func tableOfContentsControllerDidCancel(controller: WMFTableOfContentsViewController) {
+    public func tableOfContentsControllerDidCancel(controller: WMFTableOfContentsViewController) -> Bool {
```
coreyfloyd: If we are always going to return true - then no need to return a Bool.

dusek:
@coreyfloyd I addressed your comments, and used the opportunity to rebase on latest master; force pushed, please review.

coreyfloyd:
:+1:

PR #414, opened 2016-01-25
Fix T124643: Makes TOC, ShareOptions and References modal for accessibility

hhanke:
Makes sure that the text of the article and other irrelevant elements are hidden from VoiceOver and Switch Control when user invokes the Table of Contents, the ShareOptions dialog or the References view, therefore fixing [T124643](https://phabricator.wikimedia.org/T124643) .

Tested on iPhone 5 and iPad Mini (iOS 9.2.1).

Patch submitted by [A11Y LTD.](http://a11y.ltd.uk)


bgerstle:
@hhanke does this branch have the latest from `wikimedia:master`? the test that failed is (should be) disabled when run on Travis. you should see the `+testInvocations` method override in `SavedArticlesFetcherTests`

hhanke:
Rebased on master and now it passes.

hhanke:
@coreyfloyd Thanks for your comments. It seems we have lost them with a rebase now, but I have worked them in.

bgerstle:
@hhanke @coreyfloyd glad we're addressing this, but i find it a little odd that we don't get this "for free" w/ the TOC (which uses a modal presentation).  would be nice to investigate how custom modal transitions affect accessibility so we can ensure any future modal views don't need to copy/paste this logic

Wikipedia/Code/WebViewController.storyboard
```diff
@@ -42,6 +42,9 @@
                                 <constraints>
                                     <constraint firstAttribute="height" constant="200" id="Z6r-VJ-hik"/>
                                 </constraints>
+                                <userDefinedRuntimeAttributes>
```
bgerstle: is this necessary? the webview is contained w/in the `WMFArticleContainerViewController`

bgerstle:
OTOH, references & share are installed "manually" into the view hierarchy, and so I'm not surprised we need custom handling for them

bgerstle:
After giving it some more thought, I suppose it's expected that a view controller using `UIModalPresentationStyleCustom` and/or a custom `UIPresentationController` has to implement of this itself.  That said, I think it would be simpler to call `hideArticleViewForAccessibility` (or more generically `togglePresentingViewControllerAccessibility`) from `WMFTableOfContentsPresentationController`.

Also, the `WMFTableOfContentsPresentationController` has a custom `backgroundView` which obscures the entire screen.  @hhanke is there a way we can set accessibility properties on this background view to indicate that it obscures everything within its bounds, e.g. `UIAccessibilityTraitNotEnabled`?  I believe we're also using similar tricks to allow for tapping the article view to dismiss share & reference panels.

bgerstle:
@hhanke, I spoke w/ @coreyfloyd and we're fine with merging this as is :+1:. However, if you're up for it, we would like to explore the following ideas to make this a bit more elegant (in order of preference):

#### Use a custom background view
Both share & TOC use overlays to block the content beneath them: `WMFShareOptions.grayOverlay` and `WMFTableOfContentsPresentationController.backgroundView` respectively. If these views handled the accessibility logic for us, it could remove the need to manually set accessibility flags in multiple places.

This is the most compatible w/ the current architecture, but also has a clear path forward for refactoring into a shared presentation controller class which we can use for all such modal presentations that preserve the presenting view hierarchy.

#### Override presentation & dismissal callbacks in `WMFTableOfContentsPresentationController`
If the background view approach doesn't work, we can at least move this logic in the presentation controller, which can modify accessibility of its `presentingViewController`.  This logic will be easy to share across the app as we refactor existing modal views to use the modern presentation APIs.

Let us know if you're interested in exploring either of those options.  If not, I'm ready to :shipit: :grinning:

coreyfloyd:
> @coreyfloyd Thanks for your comments. It seems we have lost them with a rebase now, but I have worked them in.

No problem - This came up for the team a few months ago so we made a best practice to not rebase our branches after creating a pull request so we don't lose our comment history (we just merge instead). It would be great if you can do the same - thanks again for the contributions!

hhanke:
Yes, I will have a look at it :-)

hhanke:
@bgerstle, @coreyfloyd, I have first looked into the custom background view approach. The background view makes the elements in the presenting view non-interactive using its alpha property. This makes the background view opaque but also it disables interactivity on the elements it covers.

The fact that the alpha property doesn't also disable the elements for VoiceOver might seem as a bug at first. However, it is important to realize that the ability to be seen with VoiceOver (i.e. be selected with its cursor) does not equal interactivity. Also non interactive elements need to be able to be selected with VoiceOver because they might convey information and a blind user would not be able to perceive them otherwise. The content below the overlay is visible to people with sight and therefore it cannot be automatically determined by iOS whether it should be visible or hidden to VoiceOver.

I think it is therefore up to the developers of the application to judge in each particular situation whether the content obscured by the overlay does still convey some important information or whether it is used merely as a "graphical decoration". And even if it does convey information, there is the question whether the information is valuable enough to justify the possible difficulties in navigation for a blind person or whether it is non-essential and hiding it will improve user experience.

In the TOC situation, there is no useful information below the overlay, so the answer is easy.

I tried to find a way how to make the overlay block the accessibility of elements inside its bounds, but I didn't find a way and I do not think it is possible in any standard way.

I have therefore implemented the WMFTableOfContentsPresentationController approach and you were right that it simplifies the implementation, thanks for your help! I didn't rebase, so it is in another commit.

bgerstle:
@hhanke :100: thanks for doing such a thorough investigation and articulating the results so clearly!  i just left a minor comment about the TOC method you refactored, otherwise this looks good to me :+1:

coreyfloyd:
:+1: looks good to me as well… will let @bgerstle handle the final comment/merge

hhanke:
> is this sufficient? IIRC you were also disabling the navigation bar & toolbar

Yes, I believe so. I tested it and this disables the navigation bar and the toolbar as well. I didn't investigate it, I just assumed this must be because we are disabling a view that is higher up in the view hierarchy.

PR #417, opened 2016-01-26
Update css and fix regression with edit preview not using preview.html

montehurd:
https://phabricator.wikimedia.org/T102149

No more funky "EDIT" link in preview:

**BEFORE**
screen shot 2016-01-25 at 5 25 54 pm

**AFTER**
screen shot 2016-01-25 at 5 22 31 pm


bgerstle:
@montehurd maybe out scope here, but preview.html & index.html are almost identical... would it be worth DRY-ing them up into a template? seems like a quick win to consolidate these two nearly identical files

montehurd:
@bgerstle not a bad idea! Prob do as a follow-on?

bgerstle:
@montehurd follow-in is OK. i'll file an eng backlog ticket so we don't forget

PR #418, opened 2016-01-26
make tabBarController unavailable, use rootTabBarController

bgerstle:
Phab: [T124738](https://phabricator.wikimedia.org/T124738)

tabBarController is `nil` (since this VC contains it), so we should always use rootTabBarController. this caused issues where the app wouldn't show "Explore" tab when nearby/random shortcuts were selected, and possibly other issues


montehurd:
Seems to work! :+1: I was hoping it would also address https://phabricator.wikimedia.org/T124381, but it seems that issue remains.

PR #419, opened 2016-01-26
Fix gallery info button visibility & image URL parsing

bgerstle:
Phab: [T124726](https://phabricator.wikimedia.org/T124726)

---

This PR fixes two issues:

- Only showing info button when image's info is available and it has a file page URL
- Correctly parse image URLs which have multiple dot-separated components (see added test in [`WMFImageURLParsingTests`](https://github.com/wikimedia/wikipedia-ios/pull/419/files#diff-a128dce327b7690819fa2406bb2da2c5R46))

[Swaging](https://en.wikipedia.org/wiki/Swaging) only really has one image (the "lead image"), but the gallery had 3 images: the lead image (blacksmith's tools), in-article thumbnail (blacksmith's tools), and wiktionary "logo."*  The thumbnail for blacksmith's tools (second image) was being parsed incorrectly (ending with the ".org" substring of the URL instead of capturing the entire filename), which resulted in no image info, which resulted in the crash.  Not only should the info button have been hidden, but we also had image info for that image and it was identical to the previous image in the gallery.

With this change, image URLs similar to the blacksmith thumbnail will successfully have their filenames parsed and the gallery should only show the info button when we have the data necessary for the action.  Cases when we really don't have data are more difficult to find in production, so I'd like to add unit tests to confirm this functionality (at some point).

\* After the fix the gallery has two images because filtering out the Wikitionary image is out of scope since it's due to a separate issue

### DIFF
--- a/Wikipedia/Code/WMFImageURLParsing.m
+++ b/Wikipedia/Code/WMFImageURLParsing.m
@@ -1,4 +1,5 @@
 #import "WMFImageURLParsing.h"
+#import "NSString+WMFExtras.h"
 
 static NSRegularExpression* WMFImageURLParsingRegex() {
     static NSRegularExpression* imageNameFromURLRegex = nil;
@@ -6,7 +7,7 @@ static NSRegularExpression* WMFImageURLParsingRegex() {
     dispatch_once(&onceToken, ^{
         // TODO: try to read serialized regex from disk to prevent needless pattern compilation on next app run
         NSError* patternCompilationError;
-        imageNameFromURLRegex = [NSRegularExpression regularExpressionWithPattern:@"^\\d+px-([^.]*(\\.[^.$]{4,})*\\.[^.]*).*$"
+        imageNameFromURLRegex = [NSRegularExpression regularExpressionWithPattern:@"^\\d+px-(.*)"
                                                                           options:0
                                                                             error:&patternCompilationError];
         NSCParameterAssert(!patternCompilationError);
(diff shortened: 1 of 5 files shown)


Wikipedia/Code/WMFImageURLParsing.m
```diff
@@ -6,7 +7,7 @@
     dispatch_once(&onceToken, ^{
         // TODO: try to read serialized regex from disk to prevent needless pattern compilation on next app run
         NSError* patternCompilationError;
-        imageNameFromURLRegex = [NSRegularExpression regularExpressionWithPattern:@"^\\d+px-([^.]*(\\.[^.$]{4,})*\\.[^.]*).*$"
+        imageNameFromURLRegex = [NSRegularExpression regularExpressionWithPattern:@"^\\d+px-(.*)"
```
bgerstle: @mdholloway @niedzielski @dbrant check out this change to image filename parsing regex & business logic (see below). might want to adopt in android app as well. This image URL was discovered on the [Swaging](https://en.wikipedia.org/wiki/Swaging) article in EN Wiki.

bgerstle: You can also easily [write a test case](https://github.com/wikimedia/wikipedia-ios/pull/419/files#diff-a128dce327b7690819fa2406bb2da2c5R46) in the Android app to see if your parsing code has this issue.

niedzielski: @bgerstle I really appreciate the heads up! I think this is working? The lead image opens and shares fine.

bgerstle: @niedzielski in our case, the lead image was fine, it was the one after it that was problematic. See the example URL in the test case I linked to

coreyfloyd:
:+1:

PR #421, opened 2016-01-26
Add tests for & fix Welcome language view UI errors

bgerstle:
Phab: [T124219](https://phabricator.wikimedia.org/T124219)

---

At first, I noticed we were using `CGRectGetMinX` instead of `CGRectGetMinY`.  This should have been caught in code review or pre-merge QA, but wasn't.  I initially fixed the subtle (but important) typo* and was going to submit as-is, but then decided to add visual tests.  This led me to discover that the views were (had been?) broken for RTL languages, which has now been addressed (see the added LTR & RTL snapshots in iOS 8 & 9).

\* This issue was instigated by our [best practice to use CoreGraphics getter functions over struct fields](https://www.mediawiki.org/wiki/Wikimedia_Apps/Team/iOS/ObjectiveCStyleGuide#CGRect_Functions), which I've disagreed with in the past under the rationale that the "dot syntax" is much more readable.  As a corollary, this is also an example of why readability & expressivity are important.


montehurd:
Y u no Travis?

bgerstle:
🎤⬇️🚶🏻

PR #426, opened 2016-01-27
Fixing explore tableview bugs

coreyfloyd:
https://phabricator.wikimedia.org/T124209
https://phabricator.wikimedia.org/T124210

To fix these bugs, this patch refactors the explore architecture to better encapsulate where "truth" lies:
- Remove SSDatasources
- Use WMFExploreSectionSchema.sections as the truth for section counts
- Use the WMFSectionController.items as the truth for row counts
- Use KVO to propagate changes and trigger tableview updates

Additionally, shared functionality of section controllers has been refactored into a shared base class: WMFBaseSectionController:
- Manages a homogeneous array of items
- KVO compliant
- Supports placeholders
- Supports empty view cells
- Fetch business logic

Some other minor refactors:
- reworking the API of WMFExploreSectionController
- Move section controller management to a purpose built WMFExploreSectionControllerCache

Lastly, a few fixes to section schema logic:
- Only refresh random section according to spec (every 7 days)
- Check nearby sections for non nil location property before returning them


bgerstle:
> Merging #426 into master will increase coverage by +1.83% as of 17e2a8a

without even adding any tests?

are you a wizard

coreyfloyd:
@bgerstle @montehurd this should be good to go now

montehurd:
Yay WMFExploreSectionController protocol! Just ran into this run-time error issue yesterday. Now the compiler yells at me right away :100:

coreyfloyd:
> @coreyfloyd this makes me a little uneasy. 

@bgerstle thats not terribly hard to do :wink: 

> keeping this isolated from the other feed constants makes it easy to forget to increase it after raising other limits (or if other limits are added). this number should never be less than the maximum capacity of the feed,

Yeah - I see this as temporary - I just want to address this more holistically after modifying the feed to be push instead of pull - and actually flesh out the real rules - rather than just exploring a bunch of weird constants.

>  otherwise we could get into a situation where we're constantly purging section controllers when the feed is still below capacity.

Maybe - but thats an way out there edge case (The feed push work slips a long time + a developer increases the other numbers significantly and forget to increase this limit + lots of people have lots of things in the feed + lots of people scroll all the way down to the bottom of the feed and then all the way back up the feed multiple times in a row for an unknown reason causing things to possibly be purged)

bgerstle:
@coreyfloyd can you please respond to:

- https://github.com/wikimedia/wikipedia-ios/commit/665720c508e84027864d1df281d7afe83bc4f358#commitcomment-15773978
- https://github.com/wikimedia/wikipedia-ios/commit/b0b9254163c5b7156023799a4aae6d9a3860447f#commitcomment-15774165
- https://github.com/wikimedia/wikipedia-ios/commit/3f007a473d2d42585d1206bfd7e2cf497e6b7278#commitcomment-15774244

otherwise, looks good :+1:

montehurd:
"continue reading" may be broken...

Wikipedia/Code/WMFContinueReadingSectionController.m
```diff
@@ -27,13 +27,12 @@ @interface WMFContinueReadingSectionController ()
...
 
 @implementation WMFContinueReadingSectionController
-@synthesize delegate = _delegate;
 
 - (instancetype)initWithArticleTitle:(MWKTitle*)title
                            dataStore:(MWKDataStore*)dataStore {
     NSParameterAssert(title);
     NSParameterAssert(dataStore);
-    self = [super init];
+    self = [super initWithItems:@[self.title]];
```
montehurd: I think this is causing the continue reading crash. Should be `@[title]` not `@[self.title]`.

coreyfloyd: good catch - fixed!

bgerstle:
@coreyfloyd thanks for responses, will re-test

bgerstle:
@coreyfloyd just got a crash after dismissing POTD and attempting to scroll:

> 2016-02-01 14:09:01.724 Wikipedia Debug[581:84461] *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[WMFRandomSectionController willDisplaySection]: unrecognized selector sent to instance 0x147953600'

guessing random was underneath POTD. investigating now

bgerstle:
@coreyfloyd confirmed, happens every time i dismiss the POTD gallery and Explore attempts to re-fetch visible sections:

```
2016-02-01 14:11:07.079 Wikipedia Debug[588:85481] -[WMFRandomSectionController willDisplaySection]: unrecognized selector sent to instance 0x12615bfc0
(lldb) po [[NSThread currentThread] callStackSymbols]
error: Execution was interrupted, reason: Attempted to dereference an invalid ObjC Object or send it an unrecognized selector.
The process has been returned to the state before expression evaluation.
(lldb) po [NSThread callStackSymbols]
<_NSCallStackArray 0x126181d80>(
0   ???                                 0x0000000103e9cd7c 0x0 + 4360621436,
1   Wikipedia Debug                     0x0000000100196f30 main + 0,
2   CoreFoundation                      0x00000001810093e8 <redacted> + 408,
3   CoreFoundation                      0x0000000180f0d68c _CF_forwarding_prep_0 + 92,
4   Wikipedia Debug                     0x00000001001bd3d8 __42-[WMFExploreViewController viewDidAppear:]_block_invoke + 136,
5   CoreFoundation                      0x0000000180efc370 <redacted> + 132,
6   CoreFoundation                      0x0000000180efc208 <redacted> + 212,
7   Wikipedia Debug                     0x00000001001bd260 -[WMFExploreViewController viewDidAppear:] + 1516,
)
... redacted ...
```

montehurd:
Just tried to hide a suggestion and nothing happened... perhaps related?

bgerstle:
also not seeing section disappear after attempting to hide via overflow menu

Wikipedia/Code/WMFRelatedSectionController.m
```diff
@@ -121,57 +99,74 @@ - (NSAttributedString*)headerText {
...
+        self.searchResults = nil;
+        return error;
+    });
+}
+
+#pragma mark - WMFHeaderMenuProviding
+
+- (UIActionSheet*)menuActionSheet {
+    UIActionSheet* sheet = [[UIActionSheet alloc] bk_initWithTitle:nil];
+    [sheet bk_addButtonWithTitle:MWLocalizedString(@"home-hide-suggestion-prompt", nil) handler:^{
```
bgerstle: @coreyfloyd @nirzar this should be a destructive button (`bk_setDestructiveButtonWithTitle`)

bgerstle:
crash is fixed, just need to address blacklist regression

coreyfloyd:
@bgerstle @montehurd addressed the black list comments (no working and text is red)

montehurd:
Confirmed blacklist fix! :)

PR #427, opened 2016-01-27
Calling reload method causes the scroll. Instead just update in place

coreyfloyd:


bgerstle:
@coreyfloyd nice fix :+1: 

some food for thought: this logic was only necessary in saved/recent, where the user could potentially navigate to those tabs before the article metadata was downloaded.  IMO this is an edge case, and thus shouldn't be an inherited behavior.  what do you think about moving this logic out of the base class?

coreyfloyd:
@bgerstle Maybe - I'm not too worried about it now because it works for all lists without negative effects. Really core data removes the need for this logic entirely so I'm inclined to leave as is until we remove it instead of refactoring it now.

bgerstle:
@coreyfloyd there's a bug now in saved caused by this PR. sorry i didn't catch this before merging:

- Enable NLC (I used 3G)
- Save a page on Explore or "More like" list
- Immediately navigate to "Saved" tab
- Item shows in "Saved" w/ just title
  - Expected: item shows description & image placeholder once metadata is downloaded, eventually downloading image
  - Actual: item shows w/ title & truncated description w/o image:

screen shot 2016-01-27 at 4 03 29 pm

I suspect it's due to the cell needing to be completely reloaded (and laid out by the table view), which doesn't occur when you configure in place

PR #433, opened 2016-01-30
Tabbar Visual Design Update

nirzar:
1. Added shadowImage for UItabbar
2. Added selected and delselected images separately to change the gray tint of unselected icons

Before
simulator screen shot jan 26 2016 3 02 27 pm

After
simulator screen shot jan 29 2016 4 26 30 pm

Checked with acessibility for button shapes. no problems there because of this customization.


wikipedia/Images.xcassets/tabbar-shadow.imageset/Contents.json
```diff
@@ -0,0 +1,23 @@
+{
```
bgerstle: @nirzar can we make this into a PDF?

bgerstle:
:+1: but would be nice to have consistency w/ as many images as PDFs as possible

PR #437, opened 2016-02-01
Add cancel button to sharing dialog

bgerstle:
Phab: [T124818](https://phabricator.wikimedia.org/T124818)

---

cancel-share

But, doesn't work so well in landscape, especially on 4S:

image

Not sure if we care too much about landscape handling, since the user can just rotate back to portrait to see the full card.  OTOH, we might eventually [revert to using the built-in share dialog anyway](https://phabricator.wikimedia.org/T107074) (w/ our own extension)


nirzar:
@bgerstle is it possible to fix the action sheet coming above the article toolbar. default action sheet comes on top of everything. also grouping is between main actions and destrictive action. 

img_9318

coreyfloyd:
:+1:

bgerstle:
@nirzar possible, but unfortunately out of scope for this ticket

PR #439, opened 2016-02-01
Move VoiceOver cursor to place in article on TOC item selection

hhanke:
This tries to move the VoiceOver cursor to the correct place in article
when the user selects an item in the Table of Contents.

However, the handling of VoiceOver cursor receival seems to be buggy in
UIWebView and therefore this commit is only a partial solution.  It
just implements a way that resolves the situation in the majority of
cases.

Please see [T125354](https://phabricator.wikimedia.org/T125354) for more details.


montehurd:
Hmm seeing a bug iOS 8.4.

If you make a TOC selection, then scroll, it's jumping back to the top of the section. It seems to happen only if you scroll within the one second delay before accessibilityCursorToSection fires. Confirmed it stops happening if I turn off accessibilityCursorToSection and happens later, and confirmed it is the cause because if I turn the delay up, to say 5 seconds, make a TOC selection, then scroll, 5 seconds later it jumps... 

Could we maybe turn the delay down to narrow the window in which this could happen?

Wikipedia/Code/WebViewController.m
```diff
@@ -409,6 +417,11 @@ - (void)scrollToSection:(MWKSection*)section animated:(BOOL)animated {
     [self scrollToFragment:section.anchor animated:animated];
 }
 
+- (void)accessibilityCursorToSection:(MWKSection*)section {
+    NSString *js = [NSString stringWithFormat:@"focus_element = document.getElementById('%@'); other_element = document.body;  other_element.setAttribute('tabindex', 0); focus_element.setAttribute('tabindex', 0); other_element.focus(); focus_element.focus();", section.anchor];
```
bgerstle: @montehurd we can refactor this into a JS script which we invoke via the JS context. I wouldn't worry about breaking into multiple lines.

@hhanke we also have some JS functions which are included in the page.  the process for modifying those is documented in `docs/web-dev.md` if you're feeling brave. If not, we can refactor it after this is merged

Wikipedia/Code/WMFArticleContainerViewController+TOC.swift
```diff
@@ -19,19 +19,29 @@ extension WMFArticleContainerViewController : WMFTableOfContentsViewControllerDe
...
                                           didSelectItem item: TableOfContentsItem) {
-
+        var dismissVCCompletionHandler: (() -> Void)?
         if let section = item as? MWKSection {
             // HAX: webview has issues scrolling when browser view is out of bounds, disable animation if needed
             self.webViewController.scrollToSection(section, animated: self.webViewController.isWebContentVisible)
+            dismissVCCompletionHandler = {
+                // HAX: This is terrible, but iOS events not under our control would steal our focus if we didn't wait long enough here and due to problems in UIWebView, we cannot work around it either.
+                dispatchOnMainQueueAfterDelayInSeconds(1) {
+                    self.webViewController.accessibilityCursorToSection(section)
```
bgerstle: @montehurd @hhanke the default animation duration is 250ms, can we something closer to that, like 300-400ms?

Wikipedia/Code/WebViewController.m
montehurd: @bgerstle Ya I figure that would be a good follow-on for when we have time to make a sweep and do so for all inline JS strings :)

Wikipedia/Code/WMFArticleContainerViewController+TOC.swift
montehurd: @bgerstle I was thinking around a third of a second for the same reason. If @hhanke agrees I think it's good.

hhanke: @montehurd @bgerstle I will try it, but I think unfortunately we can't go much under the 1 second as we would get the VO cursor stolen from us, as I described on the Phabricator :-/ However, I think we can work around the bug that @montehurd spotted (thank you!) by only invoking this when we know that an assistive technology is running. We can detect that. I think this is currently only relevant to VoiceOver, but I must give it a little more thinking. It will be a little less universal, but this solution is a hack anyway. That way it won't affect users not using assistive technologies at all.

bgerstle: Only using delay w/ accessibility w/ active sounds reasonable

On Tue, Feb 2, 2016 at 2:41 PM, hhanke <notifications@github.com> wrote:

> In Wikipedia/Code/WMFArticleContainerViewController+TOC.swift
> <https://github.com/wikimedia/wikipedia-ios/pull/439#discussion_r51623145>
> :
>
> >          if let section = item as? MWKSection {
> >              // HAX: webview has issues scrolling when browser view is out of bounds, disable animation if needed
> >              self.webViewController.scrollToSection(section, animated: self.webViewController.isWebContentVisible)
> > +            dismissVCCompletionHandler = {
> > +                // HAX: This is terrible, but iOS events not under our control would steal our focus if we didn't wait long enough here and due to problems in UIWebView, we cannot work around it either.
> > +                dispatchOnMainQueueAfterDelayInSeconds(1) {
> > +                    self.webViewController.accessibilityCursorToSection(section)
>
> @montehurd <https://github.com/montehurd> @bgerstle
> <https://github.com/bgerstle> I will try it, but I think unfortunately we
> can't go much under the 1 second as we would get the VO cursor stolen from
> us, as I described on the Phabricator :-/ However, I think we can work
> around the bug that @montehurd <https://github.com/montehurd> spotted
> (thank you!) by only invoking this when we know that an assistive
> technology is running. We can detect that. I think this is currently only
> relevant to VoiceOver, but I must give it a little more thinking. It will
> be a little less universal, but this solution is a hack anyway. That way it
> won't affect users not using assistive technologies at all.
>
> —
> <https://github.com/wikimedia/wikipedia-ios/pull/439/files#r51623145>.
>



-- 
IRC: bgerstle

hhanke:
@bgerstle @montehurd I have added another commit which makes sure the javascript focus() is only invoked when VoiceOver is running, so now this should not affect other users.

I have experimented with delays below 1 second but confirmed that it makes the cursor transition less reliable.

I have also tried to move the Javascript to ```www/js```. I hope I did it correctly, please doublecheck ;-)

(One thing I encountered when following the guide in docs/web-dev.md is that ```nodenv install``` didn't want to understand the ```v0.10``` in .node-version. Therefore, I had to change it to ``` 0.10.41``` (without the ```v``` and specifying subversion). I hope it is the correct version.)

hhanke:
Btw. I also realized there might be some more places where we need to call "accessibilityCursorToFragment" as there are more places that call scrollToFragment but I did not know exactly what situations in the app these are and I realized that because of the timing issue, this might need more thinking and testing. So I thought it is best to just make sure the situation with the TOC will improve now and think of the other cases later.

bgerstle:
:+1: lgtm. @montehurd can you please re-review?

hhanke:
I just noticed this does not seem right now. As if there isn't the original commit.  Let me check it first.

hhanke:
@bgerstle @montehurd  I had to rebase this, but didn't force push here so that we do not loose your comments. Please see [PR456](https://github.com/wikimedia/wikipedia-ios/pull/456). Still not sure what happened, but probably my mistake. I am sorry for the inconvenience. I have re-tested the new PR to make sure it works OK.

PR #440, opened 2016-02-02
Shadow on ArticleView UIToolbar

nirzar:
Following the tabbar, article toolbar follows same aesthetics

**The reason for a white-patch.pdf**
So there is a either a OS bug or our app bug when you set transculency to NO on UIToolbar. there is  no other way to make the toolbar #FFFFFF 

@montehurd and I spent an hour digging into the bug but it was getting too complicated. Instead we can use a pdf as a background image and solve that bug later. 

**About the bug** 
if you set translucent = NO then there is a unknown (very random, around 100px) gap between webview and toolbar.


Wikipedia/Code/WMFArticleContainerViewController.m
```diff
@@ -301,6 +301,13 @@ - (void)unobserveArticleUpdates {
 #pragma mark - Toolbar Setup
 
 - (void)setupToolbar {
+    
+    [self.navigationController.toolbar setBackgroundImage:[UIImage imageNamed:@"white-patch"] forToolbarPosition:UIBarPositionBottom barMetrics:UIBarMetricsDefault];
```
bgerstle: please add `HAX:` comments w/ brief version of what you wrote in PR so we can get the same context in the future

bgerstle:
@nirzar @montehurd interesting, i know there are some weird behaviors w/ translucency & whether or not content extends under the bar. we can spend more time trying to remove the `HAX` later if needed

nirzar:
@bgerstle will do

nirzar:
@bgerstle added the comment

PR #441, opened 2016-02-02
Move tweaks to settings menu.

montehurd:
- No more shake to show tweaks menu.
- Added "Tweaks" entry to settings under "Debug".
- Removed some old un-used animation code, but refrained from other boyscouting on the Settings vc for now.

shake mov

### DIFF
--- a/Wikipedia/Code/AppDelegate.m
+++ b/Wikipedia/Code/AppDelegate.m
@@ -8,8 +8,6 @@
 #import "WMFAppViewController.h"
 #import "UIApplicationShortcutItem+WMFShortcutItem.h"
 
-@import Tweaks;
-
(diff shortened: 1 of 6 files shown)


bgerstle:
looks good! :+1: what would you think about changing the "Tweaks" to something more intuitive like "Developer Settings"?

bgerstle:
@montehurd did you also test that the button is hidden in release builds?

montehurd:
@bgerstle oh i like "Developer settings" :)

montehurd:
Confirmed it's hidden on release builds:

screen shot 2016-02-02 at 11 26 02 am

bgerstle:
:+1: will merge once travis passes (again)

PR #443, opened 2016-02-03
Add modal browser

coreyfloyd:
https://phabricator.wikimedia.org/T124802

There was some upfront refactoring to make this cleaner:
1. Removing MWKTitles from the Piwik Tracker API 
We were sending titles to Piwik though we weren't really tracking them. This complicated some APIs by requiring extra parameters in some methods.
2. Remove MWKHistoryDiscoveryMethod
This was only being used for restoring scroll position - so it really wasn't needed anymore. This had a similar affect of complicating APIs
3. Rename WMFArticleContainerViewController -> WMFArticleViewController. 
This was an artifact of the old architecture and will be misleading in the new browser architecture.

The real work is WMFArticleBrowserViewController. The browser VC now controls the navigation bar and tool bar for articles. It also maintains the stack for the back forward navigation of the new design.


montehurd:
Looks way cool! Couple small notes for first testing pass:
- getting assert on link tap 
- white screen on gallery dismiss
- "W" should be white
- nice if "W" tap scrolled to top of article

coreyfloyd:
@montehurd on the color of the W - FYI - its not white in the mocks…

montehurd:
Oh man, I'm *really* liking how this works!

montehurd:
On issue - if I'm reading an article, then search and select a result, it's pushing *another* chrome VC on the stack rather than pushing the search selection article into the chrome VC from the article we had been reading.

montehurd:
Seeing crashes on device when trying 3d peek on article links.

montehurd:
I haven't tested the app icon shortcuts yet - especially the "Continue reading..." one. May want to check that.

coreyfloyd:
@montehurd so I think I am about good on your feedback…

> One issue - if I'm reading an article, then search and select a result, it's pushing another chrome VC on the stack rather than pushing the search selection article into the chrome VC from the article we had been reading.

fixed

> nice if "W" tap scrolled to top of article (out of scope)

yeah out of scope

> since refresh button is gone use pull to refresh? (out of scope?)

ditto - that is in a separate ticket somewhere

> Seeing crashes on device when trying 3d peek on article links.

On master - not this patch

> getting assert on link tap

fixed

> white screen on gallery dismiss

fixed


I'm working sole on the tweak now - let me know if you are good otherwise.

montehurd:
One tiny layout issue. On my device, when I'm reading an article and scroll, I'm seeing the article show through a 1 or 2 pixel horizontal line *beneath* the bottom toolbar. Do you see this too?

montehurd:
Oh and I just tested app icon "Continue reading" 3d touch shortcut. Is broken - took me directly to feed instead of article.

coreyfloyd:
@montehurd 

> Oh and I just tested app icon "Continue reading" 3d touch shortcut. Is broken - took me directly to feed instead of article.

Fixed

> One tiny layout issue. On my device, when I'm reading an article and scroll, I'm seeing the article show through a 1 or 2 pixel horizontal line beneath the bottom toolbar. Do you see this too?

Can't reproduce

montehurd:
Found a bug related to the `Fix article presentation from search` issue - the same things still happens if you select an item from the "Read more" items at the bottom of the article.

montehurd:
Oh it's also happening if you select an article from the "Similar pages" list from the "About this article" footer menu.

montehurd:
Merging even though we're seeing some test failure so we can get this out for testing. Will double back with Brian on Monday.

PR #446, opened 2016-02-05
Fix footer menu cell constraint breakage.

montehurd:
Fixed constraint breakage and updated heights to reflect Nirzar's comment below.


nirzar:
The height for 2 line cell should 60px and the height for single title cell should 44px

montehurd:
Done!

PR #449, opened 2016-02-05
Don't show tiny images in gallery.

montehurd:
2 parts:

- Make the gallery data source ignore tiny images.
- Make article javascript ignore taps on tiny images.

Issue:
- Not all MWKImage records have size info. Such sizes are usually determined when we parse the article HTML, but not all image urls have size info to be parsed, so these end up with no height or width that we can check.

Workaround:
- If size isn't available in MWKImage record, ask SDImageCache.


montehurd:
Will look at the test failures tomorrow...

Wikipedia/Code/WMFArticleImageGalleryDataSource.m
```diff
@@ -26,7 +27,29 @@ @implementation WMFArticleImageGalleryDataSource
...
+        // Keep image if it's the article image even if we can't determine its size - we always want to show article image as first "lead" image by gallery.
+        if ([article.image isEqualToImage:image]) {
+            return YES;
+        }
+        
+        if (!image.width || !image.height) {
+            // HAX: if this image MWKImage record doesn't have a width and height value (because it wasn't determined by parsing it
+            // from the article HTML's image url) see if the cache can tell us the size.
+            SDImageCache* myCache = [SDImageCache wmf_appSupportCacheWithNamespace:@"default"];
+            UIImage* img = [myCache imageFromDiskCacheForKey:image.sourceURLString];
```
coreyfloyd: We should expose this method on WMFImageController so we don't need to work with the SDImageCache directly.

montehurd: Done!

Wikipedia/Code/WMFArticleImageGalleryDataSource.m
```diff
@@ -26,7 +27,28 @@ @implementation WMFArticleImageGalleryDataSource
...
+        // always want to show article image as first "lead" image by gallery.
+        if ([article.image isEqualToImage:image]) {
+            return YES;
+        }
+        
+        if (!image.width || !image.height) {
+            // HAX: if this image MWKImage record doesn't have a width and height value (because
+            // it wasn't determined by parsing it from the article HTML's image url) see if the
+            // cache can tell us the size.
+            UIImage* img = [[WMFImageController wmf_appImageCache] imageFromDiskCacheForKey:image.sourceURLString];
```
bgerstle: FYI, this could be very expensive (reading the entire image from disk to get the size). you're doing this synchronously, on the main thread. are there any known cases where images w/o given widths/heights are actually above the threshold _and_ worth including in the gallery?

bgerstle: put another way, what happens if these images aren't included?

montehurd: @bgerstle 

Yes there are such cases. 

Any time an image size is requested which exceeds the width or height canonical image the image tag html won't have a width or height attribute - so we can't catch it at parse time. For instance, this image on "enwiki > color wheel":

screen shot 2016-02-05 at 4 08 39 pm

Without this code that ^ image disappears from the gallery. 

Thankfully imageFromDiskCacheForKey first tries imageFromMemoryCacheForKey.

montehurd: This was a know issue pre 5.0. See MWKImage's `updateWithData:` method. It accounted for such cases slipping between the cracks of the html parse:

    if (!self.width || !self.height) {
        UIImage* img = [UIImage imageWithData:data];
        self.width  = [NSNumber numberWithInt:img.size.width];
        self.height = [NSNumber numberWithInt:img.size.height];
    }

This formerly got invoked when we intercepted the WebView's retrieval of image assets. That only happened *after* we parsed the html, of course, so only occasionally did this happen, but the result was that MWKImage pretty much always had width and height we could check. We *could* do the same somehow with SDImageCache at the time it saves an image - we could see if our MWKImage record has a size and if not we could update it on a background thread.

bgerstle: @montehurd got it, thanks for detailed explanation. Ideally we could workaround this in a way that wouldn't Impact UX directly, but if it doesn't caused obvious freezing/hanging/bloating we can address later. For example, doing this processing in background to get list of gallery images asynchronously. 

👍🏻

montehurd: @bgerstle One thing I'm trying real quick - checking uniqueSmallestVariants instead of largest - the smallest should still tell us if it's big enough for gallery inclusion, and then I could bk_map those to the largest variant. That way any image inflation wouldn't be happening to huge images...

montehurd:
Didn't get to the tests yet :(

Wikipedia/Code/MWKImage.m
```diff
@@ -278,4 +278,58 @@ - (BOOL)isLeadImage {
...
+}
+
+- (MWKImage *)largestNonCanonicalVariant {
+    NSString *largestNonCanonicalVariantURL = [[[self.article.images imageSizeVariants:self.sourceURLString] bk_select:^BOOL (NSString* url) {
+        return [MWKImage fileSizePrefix:url] != NSNotFound;
+    }] lastObject];
+    return [self.article imageWithURL:largestNonCanonicalVariantURL];
+}
+
+-(UIImage*)imageFromAppImageCache {
```
bgerstle: this sets a bad precedent of breaking the `WMFImageController` abstraction, which for all other things intentional encapsulates the cache. instead of querying the `MWKImage` for its data, ask the `WMFImageController` for the image from its cache. for example: `WMFImageController.sharedInstance().cachedImageWithURL(self.sourceURLString)`.  having this as an instance method on `MWKImage` is acceptable IMO as a convenience, but it should still query the image controller, not the cache itself.

montehurd: @bgerstle the existing `cachedImageWithURL` returns 'Promise<WMFImageDownload>', which is not what I need here...

bgerstle:
:+1:

PR #450, opened 2016-02-05
Search View Visual Design Fixes

nirzar:
Bunch of small fixes i have been meaning to do

- Bigger search bar - enough padding around
- BUG FIX: the underline button now shows the underline below the title, it used to do that only when you change the language
- removed the blur, most of the time blur is interfering with what we want the user to focus on
- recently search header now has base color so the cells pop up
- BUG FIX: more button when pressed used to be whole dark gray. now has better tap highlight
- Close button aligns with the search icon

new-search


Wikipedia/Code/WMFSearchViewController.m
```diff
@@ -434,7 +440,7 @@ - (void)updateSearchSuggestion:(NSString*)searchSuggestion {
 }
 
 - (CGFloat)searchFieldHeightForCurrentTraitCollection {
-    return self.traitCollection.verticalSizeClass == UIUserInterfaceSizeClassCompact ? 44 : 64;
+    return self.traitCollection.verticalSizeClass == UIUserInterfaceSizeClassCompact ? 44 : 72;
```
coreyfloyd: @nirzar 3 comments:

1. You only changed the vertical size, not the size when in horizontal and the status bar disappears - is that intentional?

2. This is now the only top bar in the whole app that is taller than the iOS standard 64 - so it looks out of place. Why do we need this bar to be bigger? 

3. We now have less vertical space for the results to show above the keyboard - not so bad on the iPhone 6+ but its getting cramped on 4 inch screens.

Wikipedia/Code/WMFSearchViewController.m
```diff
@@ -501,6 +507,9 @@ - (void)updateLanguageButtonsToPreferredLanguages {
 - (void)resizeLanguageButtonsIfNeeded {
     [self.languageButtons enumerateObjectsUsingBlock:^(UIButton* _Nonnull obj, NSUInteger idx, BOOL* _Nonnull stop) {
         [obj sizeToFit];
+        CGRect f = obj.frame;
+        f.size.height = 30.0f;
+        obj.frame = f;
```
coreyfloyd: This height constraint should be placed in the same place the height is currently set - that is probably in the storyboard. But it might be in the button subclass.

nirzar:
>This is now the only top bar in the whole app that is taller than the iOS standard 64 - so it looks out of place. Why do we need this bar to be bigger?

This was never meant as a titlebar or topbar. think of it as a textfield. 
img_9344

From day one i wanted "search wikipedia" to be bigger and nicer to type in. For twitter ^ tweeting is important so their tweet field is a richer to type in than their search fields.

This is what a search field looks like. has a 13px font size and a standard ios style. 
img_9344

I wanted to get away from it to create a better place to type your wikipedia queries.

Now the question of why bigger hight. when the font size is bigger, it needs breathing room around it. the richness comes from whitespace. i have added 4px on top and 4px on bottom to get this. this text was cramped otherwise. 

>We now have less vertical space for the results to show above the keyboard - not so bad on the iPhone 6+ but its getting cramped on 4 inch screens.

4px on top and bottom is hardly capturing a lot of real estate. I checked this on 5 and 6 and both do not take up take up even space for 1 search result unit. 

> You only changed the vertical size, not the size when in horizontal and the status bar disappears - is that intentional?

for all the above reasons I would love to, but space is a very serious issue in landscape than portrait. maybe we can reduce the font size on landscape to get the same whitespace that is required.

Wikipedia/Code/WMFSearchViewController.m
nirzar: [obj sizeToFit];
^ This changes the height from storyboard height. that's why the bug. first time you see these buttons, the height of button is 44px. (so the bottom line is sticking bottom of the toolbar) and when you change the language, height of the button is 30px so the line has a gap. we want to keep the height 30px. maybe this is a wrong way to do it. I will consult @montehurd on how to do this right. i saw it done like this in the same file so i copy-pasted the code.

coreyfloyd:
> This was never meant as a titlebar or topbar. think of it as a textfield. 

That makes sense as a goal - but the actual implementation is a white rectangle pinned to the top of the screen. With another scope bar pinned to the bottom of it and the content of the screen pinned to that.

On top of that, every other screen has a white or black bar at the top of the screen.

You might not want it to be a bar - but it visually thats what it is. If you want it to be a floating textfield it really needs different treatment - it needs to not look like a bar.

> From day one i wanted "search wikipedia" to be bigger and nicer to type in.

Nicer is better - but this isn't really a "why". Bigger isn't always better - we need to be qualified. Why does it need to be bigger - what are you visually trying to fix? How does larger text benefit the user? Why is it important that text be bigger in this context than other contexts?

> For twitter ^ tweeting is important so their tweet field is a richer to type in than their search fields.

Agreed - but the analogy is a little stretched - its not a search field - its text view that is for composing multiline message. It has to be bigger. Additionally, Twitter does not need to display any content below the tweet - the tween IS the content.

Also notice that twitter actually is using a bar for the profile image and the "X", the bottom border is just invisible. They didn't try to make the bar bigger to fit their needs.



@nirzar Sorry for the long responses but the gist for me is this:
If we want to make our search unique and not use a bar - I'm all for it, but we need to make it not look like a bar.
If we don't want to change visuals enough to differentiate it more from a standard bar, then we should use standard bar heights.

nirzar:
image

bgerstle:
@nirzar please merge master into your branch so Travis can re-run the tests w/ my fixes

bgerstle:
@nirzar not sure which "master" you merged in, but now the tests are passing 😜 however, it looks like you and @coreyfloyd still had some beef to chew on, so I'll let him pass judgement upon his return ✌🏻

nirzar:
@bgerstle Corey and I had a heart-to-heart and shared our feelings. there were tears but we are all good and resolved all the issues. 

----- unrelated issues ------
:-1:  found a bug / fixing it

bgerstle:
> @bgerstle Corey and I had a heart-to-heart and shared our feelings. there were tears but we are all good and resolved all the issues.

That's so beautiful! 😭

PR #453, opened 2016-02-08
Fix unit tests

bgerstle:
`UIViewController_WMFSearchButtonTests` were causing other tests to fail by throwing exceptions after the search button tests have already run.[0][1][2]  This was previously handled by using expectations w/ predicates to "block" on search presentation & dismissal, ensuring proper transitions & cleanup during & in-between tests.

0: https://travis-ci.org/wikimedia/wikipedia-ios/jobs/107341266#L4316
1: https://travis-ci.org/wikimedia/wikipedia-ios/jobs/107126593#L4295
2: https://travis-ci.org/wikimedia/wikipedia-ios/jobs/107126593#L4492


bgerstle:
> Merging #453 into master will increase coverage by +23.53% as of ea29083

As Mr. Connery would say, not too shabby

bgerstle:
@montehurd that's two in a row, should I go for three?

PR #454, opened 2016-02-08
Run uncrustify & static analyzer on pull requests

bgerstle:
We have a code standard, I think we should enforce it. Otherwise, master gets into a wonky state and the next time one of the maintainers goes to push a change, they end up polluting their diff with uncrustification of sources their PR didn't make any meaningful changes to.

As an additional bonus, thanks to the work done by @barbarar, I've upgraded our compiler warning settings to fail compilation on all but a specific subset of warnings (see commit msg for the reason why nullability-completeness is permitted).  I've also added static analysis to our build matrix, since we might as well prevent add'l static analysis warnings as well as compiler warnings.

All in all, keep master clean & green ✅


bgerstle:
needs to be rerun once #453 is merged

bgerstle:
hm, having some issues finding RubyGems... might need to update some of our dependencies soon

bgerstle:
> Could not find minitest-5.8.4 in any of the sources

😡

bgerstle:
@montehurd can haz :ship: plz

PR #456, opened 2016-02-08
Move VoiceOver cursor to place in article on TOC item selection (rebased)

hhanke:
This is a rebased version of [PR439](https://github.com/wikimedia/wikipedia-ios/pull/439)

It tries to move the VoiceOver cursor to the correct place in article
when the user selects an item in the Table of Contents.

However, the handling of VoiceOver cursor receival seems to be buggy in
UIWebView and therefore this commit is only a partial solution.  It
just implements a way that resolves the situation in the majority of
cases.

Please see [T125354](https://phabricator.wikimedia.org/T125354) for more details.


bgerstle:
re-approving :+1: @montehurd please re-review

montehurd:
Cool! I'll re-test in a few.

PR #457, opened 2016-02-09
Restyle Settings!

montehurd:
**Settings interface using standard grouped table bits!**
- [x] Setup rough scaffold for using SSSectionedDataSource (ss data source, cell model obj, cell obj and xib etc)
- [x] Make cell obj capable of displaying according to mock (icon, title, switch, disclosure icon, disclosure "EN" lang code text, etc) and able to accommodate long localized strings
- [x] Use the actual icons from the mocks
- [x] Use icon colors from mocks
- [x] Use localized strings
- [x] Update localized strings to match mocks
- [x] Hook up taps to load view controllers or external URLs
- [x] Hook up switch taps to do their things
- [x] Configure special functionality for cells/sections such login, search lang, and the debug section - i.e. make sure the old behavior works or works according to new mock 
- [x] Add "Support Wikimedia" option and hook it up!
- [x] Delete any cruft which is now completely orphaned as a result of this refactor
- [x] Use cell spacing from mocks
- [x] Use header spacing from mocks
- [x] Check with iOS 8 and adjust if needed
- [x] Figure out tap delay issue
- [x] Add visual tests for settings cells

untitled mov


montehurd:
@bgerstle still getting delay on login tap - it's weird!

montehurd:
@coreyfloyd this is ready for some reviewz... still not sure why tapping login has a delay. Ideas?

nirzar:
@montehurd can we change the string "Rate app" to "Rate the app" and "Support Wikimedia" to "Support Wikipedia"

montehurd:
@nirzar yes

montehurd:
@bgerstle @coreyfloyd Should be good to go! The weird tap delay is fixed!

montehurd:
@bgerstle @coreyfloyd Good to go?

montehurd:
untitled mov

montehurd:
untitled mov

Wikipedia/Code/WMFSettingsViewController.m
```diff
@@ -1,634 +1,296 @@
...
+        cell.title          = menuItem.title;
+        cell.iconColor      = menuItem.iconColor;
+        cell.iconName       = menuItem.iconName;
+        cell.disclosureType = menuItem.disclosureType;
+        cell.disclosureText = menuItem.disclosureText;
+        
+        cell.selectionStyle = (menuItem.disclosureType == WMFSettingsMenuItemDisclosureType_Switch) ? UITableViewCellSelectionStyleNone : UITableViewCellSelectionStyleDefault;
+        
+        @strongify(self)
+        [cell.disclosureSwitch setOn:[self getSwitchOnValueForMenuItemType:menuItem.type]];
```
coreyfloyd: Its better if this piece of information should be inside the WMFSettingsItem that way each cell is driven solely by the values stored in the model object.

coreyfloyd:
:+1:

bgerstle:
:+1:

PR #459, opened 2016-02-09
Fix article restore logic

coreyfloyd:
I wasn't able to reproduce the bug here:
https://phabricator.wikimedia.org/T124381

This is quite possible due to the recent navigation refactor…

However I did notice that the logic for restoring articles wasn't quite right since we started processing the shortcuts. So I did fix that here.

@montehurd can you test this to see if indeed you can still cause the crash?

### DIFF
--- a/Wikipedia/Code/NSUserDefaults+WMFExtensions.swift
+++ b/Wikipedia/Code/NSUserDefaults+WMFExtensions.swift
@@ -26,8 +26,12 @@ extension NSUserDefaults {
         return self.wmf_dateForKey(WMFAppBecomeActiveDateKey)
     }
     
-    public func wmf_setAppBecomeActiveDate(date: NSDate) {
-        self.setObject(date, forKey: WMFAppBecomeActiveDateKey)
+    public func wmf_setAppBecomeActiveDate(date: NSDate?) {
+        if let date = date {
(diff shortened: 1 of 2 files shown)


montehurd:
@coreyfloyd confirmed that I can no longer reproduce this on master. Testing the PR next...

montehurd:
I noticed something a bit weird - if you hit the home button when reading an article, then select "continue" from the app icon shortcut, it looks like it's doubling up:

untitled mov

Could we not do the animation?

bgerstle:
@montehurd might be related to https://phabricator.wikimedia.org/T126300

coreyfloyd:
@montehurd fixed

bgerstle:
code looks good :+1:

bgerstle:
also confirmed this fixes [T126300](https://phabricator.wikimedia.org/T126300)

coreyfloyd:
@bgerstle @montehurd also should note the solution that I employed is not great because of the dual navigation logic. Once we throw that out we can consolidate this and remove the need for interrogating the app structure to figure this out.

bgerstle:
@coreyfloyd is there some sort of comment you can add to remind us to address that when we make a decision about navigation?

PR #460, opened 2016-02-09
Boyscout - statuts bar fix

coreyfloyd:
This is from the new aviation code last week.

Fix toc not respecting the previous status bar style (caused the status bar to "disappear" when article is not in browser mode)


bgerstle:
> from the new aviation code

You mean navigation, right? Although, it would be sweet to have a flight-sim easter egg :-P

bgerstle:
:+1:

PR #461, opened 2016-02-09
Fix delete crash 

coreyfloyd:
Delete animation was being done twice.
This was unhanded in iOS 8, but apparently iOS 9 is ok with it?


montehurd:
Hmm not seeing the crash w/iOS 8 sim on master, but the fix works fine too and looks right. :+1:

coreyfloyd:
@montehurd i had to test on device to see the crash

montehurd:
@coreyfloyd ah cool

PR #462, opened 2016-02-10
Fix bug where explore would not reload on network connect

coreyfloyd:
Fixes:
https://phabricator.wikimedia.org/T126302

Also fixing this one as well:
https://phabricator.wikimedia.org/T121924

A confluence of issues fixed by the following:

- Be sure to reload table on reconnect
- All sections need to have an empty cell - this is necessary because if a section doesn't have a cell on screen, then it isn't "visible"
- Always report 1 row per section (even if empty, see above)

Also some minor refactors

### DIFF
--- a/Wikipedia/Localizations/qqq.lproj/Localizable.strings
+++ b/Wikipedia/Localizations/qqq.lproj/Localizable.strings
@@ -372,4 +372,4 @@
 "explore-random-article-sub-heading" = "Subtext beneath the 'Random article' header with name of language Wikipedia";
 "explore-potd-heading" = "Text for 'Picture of the day' header";
 "home-empty-section" = "Text shown when a section of the home screen has no results";
-"home-empty-section-check-again" = "Button title to relaod a section when it has no results";
+"home-empty-section-check-again" = "Button title to reload a section when it has no results";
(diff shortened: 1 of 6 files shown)


Wikipedia/Localizations/qqq.lproj/Localizable.strings
```diff
@@ -371,3 +371,5 @@
 "explore-random-article-heading" = "Text for 'Random article' header";
 "explore-random-article-sub-heading" = "Subtext beneath the 'Random article' header with name of language Wikipedia";
 "explore-potd-heading" = "Text for 'Picture of the day' header";
+"home-empty-section" = "Text shown when a section of the home screen has no results";
+"home-empty-section-check-again" = "Button title to relaod a section when it has no results";
```
montehurd: Typo

coreyfloyd:
@bgerstle @montehurd should be good to go

bgerstle:
:+1: lgtm. lots of changes to offline stuff. will need more regression testing in QA

montehurd:
Offline reading seems solid still :+1:

montehurd:
Hmm travis seems stuck on this one...

PR #464, opened 2016-02-10
Fix crash due to citation being parsed into title for disambiguations

coreyfloyd:
- Ignoring text that contains "cite_note"

https://rink.hockeyapp.net/manage/apps/152725/app_versions/42/crash_reasons/101251146

### DIFF
--- a/Wikipedia/Code/MWKSection.m
+++ b/Wikipedia/Code/MWKSection.m
@@ -280,7 +280,7 @@ static NSString* const WMFSectionDisambiguationTitlesXPathSelector = @"//div[@cl
 - (nullable NSArray<MWKTitle*>*)disambiguationTitles {
     NSArray* textNodes = [self elementsInTextMatchingXPath:WMFSectionDisambiguationTitlesXPathSelector];
     return [textNodes wmf_mapAndRejectNil:^id (TFHppleElement* node) {
-        if (node.text.length == 0 || [node.text containsString:@"redlink=1"]) {
+        if (node.text.length == 0 || [node.text containsString:@"redlink=1"] || [node.text containsString:@"cite_note"]) {
             return nil;
         }
         return [[MWKTitle alloc] initWithInternalLink:node.text site:self.site];
--- a/Wikipedia/Code/MWKSection.m
+++ b/Wikipedia/Code/MWKSection.m
@@ -280,7 +280,7 @@ static NSString* const WMFSectionDisambiguationTitlesXPathSelector = @"//div[@cl
 - (nullable NSArray<MWKTitle*>*)disambiguationTitles {
     NSArray* textNodes = [self elementsInTextMatchingXPath:WMFSectionDisambiguationTitlesXPathSelector];
     return [textNodes wmf_mapAndRejectNil:^id (TFHppleElement* node) {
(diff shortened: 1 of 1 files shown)


Wikipedia/Code/MWKSection.m
```diff
@@ -280,7 +280,7 @@ - (NSString*)summary {
 - (nullable NSArray<MWKTitle*>*)disambiguationTitles {
     NSArray* textNodes = [self elementsInTextMatchingXPath:WMFSectionDisambiguationTitlesXPathSelector];
     return [textNodes wmf_mapAndRejectNil:^id (TFHppleElement* node) {
-        if (node.text.length == 0 || [node.text containsString:@"redlink=1"]) {
+        if (node.text.length == 0 || [node.text containsString:@"redlink=1"] || [node.text containsString:@"cite_note"]) {
```
bgerstle: not clear why we're combining XPath & manual string matching.  @coreyfloyd could you try consolidating this into the XPath, e.g. `//div[@class='hatnote']//a/[@href and not(@id = 'cite_note') and not(boolean(@redlink))]` @montehurd any thoughts?

bgerstle: also, we should really have some tests for this. even simple "expected output for this sample string" tests

montehurd: @bgerstle would probably advise against changing this for now. Maybe look at post 5.0?

bgerstle: sure, but if we had tests, it wouldn't be a "scary thing" that we couldn't change. e.g. the image URL parsing regex was recently changed, and had several regression tests to verify it still covered specific cases

montehurd: @bgerstle for sure!

bgerstle: also: `#cite_note` is just one instance of a string w/o the "internal link prefix" (as the assertion says `MWKTitle` is expecting).  maybe it would be safer to just filter out all nodes w/o the internal link prefix (i.e. `/wiki/...`)?

bgerstle: in XPath parlance: `starts-with(@href, '/wiki/')`

montehurd: @bgerstle only advising waiting per our discussion at standup about being super careful about not taking on more work pre 5.0.

bgerstle:
@coreyfloyd sorry got a bit into the weeds on that comment. in an ideal world, we'd do something we knew was performant and elegant (e.g. an optimized, single XPath query), but lacking a test harness for that, this is fine.  however, i do think it's worth filtering out strings that don't start with `WMFInternalLinkPrefix` (`/wiki`) instead of looking for the specific case of `cite_note`

coreyfloyd:
@bgerstle not sure I follow the end result of the conversation - so should we just change it to?

```
if(![node.text contains:WMFInternalLinkPrefix]){
    return nil;
}
```

bgerstle:
@coreyfloyd :+1: but we might need to leave the redlink thing in there too. so just replace your `cite_note` check for what you wrote above

montehurd:
@coreyfloyd @bgerstle is ^ ok?

coreyfloyd:
@montehurd looks good to me - @bgerstle ?

bgerstle:
:+1:

PR #465, opened 2016-02-10
Fix article not refreshing when editing

coreyfloyd:
This also fixes the article not refreshing if the reload button is pressed

We never refreshed the article if we had an existing article - the is not always desired.

https://phabricator.wikimedia.org/T126064

### DIFF
--- a/Wikipedia/Code/WMFArticleViewController.m
+++ b/Wikipedia/Code/WMFArticleViewController.m
@@ -703,7 +703,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (void)fetchReadMore {
     @weakify(self);
-    [self.readMoreListViewController fetch].then(^(id readMoreResults) {
+    [self.readMoreListViewController fetchIfNeeded].then(^(id readMoreResults) {
         @strongify(self);
         [self updateArticleFootersIfNeeded];
     })
--- a/Wikipedia/Code/WMFArticleViewController.m
+++ b/Wikipedia/Code/WMFArticleViewController.m
@@ -144,10 +144,6 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)setArticle:(nullable MWKArticle*)article {
     NSAssert(self.isViewLoaded, @"Expecting article to only be set after the view loads.");
 
-    if (_article == article) {
-        return;
-    }
-
     _footerMenuViewController      = nil;
     _tableOfContentsViewController = nil;
     _shareFunnel                   = nil;
@@ -163,9 +159,6 @@ NS_ASSUME_NONNULL_BEGIN
     // always update toolbar
     [self updateToolbar];
 
-    // always update footers
-    [self updateArticleFootersIfNeeded];
-
     if (self.article) {
         [self startSignificantlyViewedTimer];
         [self wmf_hideEmptyView];
@@ -175,6 +168,10 @@ NS_ASSUME_NONNULL_BEGIN
             [self fetchReadMore];
         }
     }
+    
+    // always update footers
+    [self updateArticleFootersIfNeeded];
+
 }
 
 - (MWKHistoryList*)recentPages {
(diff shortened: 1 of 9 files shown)


Wikipedia/Code/WMFArticleViewController.m
```diff
@@ -850,7 +858,7 @@ - (void)showProtectedDialog {
 
 - (void)sectionEditorFinishedEditing:(SectionEditorViewController*)sectionEditorViewController {
     [self dismissViewControllerAnimated:YES completion:NULL];
-    [self fetchArticleIfNeeded];
+    [self fetchArticleForce:YES];
```
bgerstle: nit: pick which layer of this "abstraction" should be used: either the high level methods (i.e. `fetch` or `fetchIfNeeded`) or the underlying `fetch:(BOOL)force`. to be extra explicit, you can prefix the "internal" method w/ an underscore: `_fetchArticle:(BOOL)`

bgerstle:
:+1: with a nit. checking out now to verify

Wikipedia/Code/WMFArticleViewController.m
```diff
@@ -641,9 +641,9 @@ - (void)saveWebViewScrollOffset {
...
 #pragma mark - Article Fetching
 
-- (void)fetchArticleIfNeeded {
+- (void)fetchArticleForce:(BOOL)force {
     NSAssert(self.isViewLoaded, @"Should only fetch article when view is loaded so we can update its state.");
-    if (self.article) {
+    if (!force && self.article) {
         return;
     }
     [self showProgressViewAnimated:YES];
```
bgerstle: :-1: when refreshing an article that's up to date (should be no change), the view flickers and the progress gets stuck at 80% (i.e. 100% download). IMO, for the case where we already have an article we should wait until the revision check finishes before updating the UI. if the article is out of date, show progress UI as usual, otherwise do nothing (for now). ideally, we'd also show some kind of loading indicator (e.g. network activity indicator) and confirm to the user that their article is up to date if revision is current.

screen shot 2016-02-10 at 3 00 24 pm

coreyfloyd:
@bgerstle that should fix the issue you saw

montehurd:
:-1: 

Fixed the refresh issue, but introduced a weird issue not on master. After edit there's "READ MORE" label overlaying the upper left corner of the image...

untitled mov

coreyfloyd:
@montehurd lets file a separate bug for that - something weird must be going on in the web view footers when we reload the article. (I know the footers weren't really designed to handle being reloaded - so we should address that specifically and it should take care of this visual glitch)

montehurd:
Hmm since this introduces the bug I think we should figure it out.

**Simplified repro steps:**

*With "A" navigation mode*

- Tap refresh

untitled mov


*With "B" navigation mode:*

- Tap edit pencil
- Dismiss edit window

untitled 2 mov

Wikipedia/Code/WMFArticleViewController.m
```diff
@@ -175,6 +168,10 @@ - (void)setArticle:(nullable MWKArticle*)article {
             [self fetchReadMore];
         }
     }
+    
+    // always update footers
+    [self updateArticleFootersIfNeeded];
```
montehurd: I think this is redundant? fetchReadMore makes this happen already...

coreyfloyd:
@montehurd it doesn't introduce a bug - its existed  - we just haven't been able to refresh a page in the UI in a while. I'm fine with this work getting prioritized soon, but its totally unrelated to this ticket - so it should be tracked separately,

montehurd:
@coreyfloyd Just confirmed the bug I pointed out does not exist on master, neither does it manifest on master through either of the two reproduction steps I outlined:

untitled mov

montehurd:
:+1: seems solid!

bgerstle:
@montehurd @coreyfloyd as discussed... couldn't help myself and fixed the regression caused by the workaround. cleaning up some logic in the process. sorry to draw this out, but please review.  here are some GIFs w/ my changes (**note that I added a tweak which lets me see the full timestamp of the article's "last modified" date**):

updated footer fix demo mov
visual glitch fixed demo mov

montehurd:
Testing...

bgerstle:
@montehurd should be fixed now, thanks for catching

regression fix mov

montehurd:
testing...

montehurd:
IT'S WORKING!!!

montehurd:
:+1:

PR #467, opened 2016-02-10
Fix for article not showing empty view

coreyfloyd:
Also Boyscout - reattempt fetch when network comes back online

### DIFF
--- a/Wikipedia/Code/WMFArticleViewController.m
+++ b/Wikipedia/Code/WMFArticleViewController.m
@@ -682,12 +682,19 @@ NS_ASSUME_NONNULL_BEGIN
                                                       sticky:NO
                                        dismissPreviousAlerts:NO
                                                  tapCallBack:NULL];
+
+            if ([error wmf_isNetworkConnectionError]) {
+                @weakify(self);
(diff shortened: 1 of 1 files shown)


bgerstle:
verified iPad 3 iOS 8.3 :+1: 

- Launch app
- Enable airplane mode
- Try to navigate to article from Explore
- Article view is pushed, showing proper error view & warning banner :white_check_mark:

bgerstle:
@coreyfloyd minor thing: when fetching article after connection comes back online, there's no progress feedback

coreyfloyd:
@bgerstle looked at it and not sure why that is happening - lets put a pin in it and file a ticket to investigate the issue.

montehurd:
Hmm seeing some funky times... not sure if related:

If I turn off wifi, then tap on an article it looks good, but if I go back and try to scroll the feed weird things happen:

untitled mov

Also noticed if I start the app from a clean install w/o wifi, tap through the welcome screens to the feed, and then enable wifi, and then try to refresh, it's just... stuck:
untitled mov

coreyfloyd:
@montehurd not related - and fixed in another pr #462

PR #468, opened 2016-02-10
Enabling delete all for saved pages and history

coreyfloyd:
https://phabricator.wikimedia.org/T115136

Already implemented just updating to fix design and re-enable


bgerstle:
So, technically this PR does what it's supposed to... but...

:-1: 🚨 i'm pretty sure we should be deleting article data when they're deleted from history:

- Clear all (or delete individual item) from recent
- The deleted articles' folders should be removed (check Documents/Data/sites/...)

montehurd:
:-1: 

Missed a couple strings from the mocks...

screen shot 2016-02-10 at 7 24 25 pm

...instead of: 
screen shot 2016-02-10 at 7 25 13 pm

coreyfloyd:
@montehurd updated strings - good catch

PR #469, opened 2016-02-10
Fix Explore assertion when showing last open article while offline

bgerstle:
And related issue hiding error view when browser/article is dismissed.


bgerstle:
continue-reading-offline-demo

coreyfloyd:
:+1:

PR #470, opened 2016-02-11
T123626: Pressing the share-button before content loads and again after opens 2 dialoges

barbaramartina:
Fix for ticket [T123626:](https://phabricator.wikimedia.org/T123626):

> 1.     Navigate [through blue links] to an article.
> 2.     Before the content loads, but after the image has loaded, press the share button.
> 3.     When the content has loaded press the share button again (which still is active, even though the share-dialoge is visable)


coreyfloyd:
@barbaramartina could we use the this to disable the button in the method:
`updateToolbarItemEnabledState`?

barbaramartina:
@coreyfloyd I see the shareItem in the toolbar is updated at three points, when the view is load, an article has loaded or when a view controller is shown in the current navigation controller.
I was analyzing if it make sense to use  the new isActive property I added to disable the share item in the toolbar, but since the toolbar is not visible when the share dialog is opened I think there is no need to disable the button, because anyway it can not be touched nor seen.
If we would like to disable the share option in the toolbar, even when it's not visible, we would need to call updateToolbarItemEnabledState two more times, when the shareOptionsController presents the dialog and when the dialog is closed. For that we would need to add a callback, KVO, or delegate to know when the shareOptionsController has dismissed its view... I think that would add more dependencies between the shareOptionsController and WMFArticleBrowserViewController, and given the share button is not touchable nor visible while sharing the article I think is better not to use this isActive property for disabling the button.

bgerstle:
> but since the toolbar is not visible when the share dialog is opened

How is it possible to open the share dialog when the toolbar isn't visible?

barbaramartina:
@bgerstle  the bar is visible, you click the share button, the dialog open, the bar is behind the dialog, the share toolbar item is not visible.

bgerstle:
@barbaramartina right, so to summarize: leave enabled/disabled logic as it was, simply do nothing when the item was tapped and share options are already presented?

also, according to the bug description, you shouldn't really be able to share something until the content is loaded, so that might need to be updated.

barbaramartina:
@bgerstle oka I will add this logic too, to prevent sharing until the whole content is loaded. I have not interpreted that this was a bug, I thought it was possible to share something after the image was loaded, before the complete content.
I will update this PR next week.

bgerstle:
> I thought it was possible to share something after the image was loaded, before the complete content.

I don't think it should be supported, since we intend to either extract the lead paragraph or use text selected by the user—neither of which is possible until the article content has been retrieved.

barbaramartina:
I will check the canShare logic to make sure it's correctly implemented. I've also found some code that seems to be dead / unused, I would like you to check this video when possible https://www.dropbox.com/s/ubx7aau9p5mxr0z/old-sharing-code.mov?dl=0

barbaramartina:
@coreyfloyd  / @bgerstle  I have added this commit: 
https://github.com/barbaramartina/wikipedia-ios/commit/fd1fa02016698a52b9d39c0d81e8e7579477fdce
To fix the following:
- Do not share the article until there is text to share (I checked if there is a share snippet available)
- Do not show duplicate share dialog is one is already active in class WMFArticleViewController

bgerstle:
@barbaramartina this PR seems to contain a lot of older commits. can you please rebase so we can focus on your changes? hard to parse everything out of the history

coreyfloyd:
@bgerstle I'm fine with if you are. Not sure why all those extra commits are showing up , but @barbaramartina already identified the new commit and I'm satisfied.

coreyfloyd:
GitHub is supposed to at least make the diff readable. Don't know why t isn't diffing with master properly

PR #388, opened 2016-01-17
Trending articles!

montehurd:
Weekend fun ticket.
Implemented the fetching and list displaying logic - see image below.
If we just use a regular feed "group" with a "*More trending articles*" footer the remaining work is pretty easy.

----

To see the list of currently trending articles (WIP) trigger a memory warning in the simulator (`cmd-shift-m`).

----

TODO:
- [ ] Chat with Kevin (I think) to see if we can request smaller title count and to confirm how often the list is updated. Also ask about stability / whether we can send it lots of requests etc.
- [ ] Change titleExclusions string for "Main_page" to use proper main page name for site lang.
- [ ] Add to feed, showing first 3(?) with footer tap to call up full list of 50(?) or so.
- [ ] Remove "didReceiveMemoryWarning" testing stub.
- [ ] Localize some strings.

----

untitled mov

PR #471, opened 2016-02-12
Top read articles!

bgerstle:
#Phab: [T124716](https://phabricator.wikimedia.org/T124716), 

---

- [x] Get previews of (somewhat-filtered) top-read articles into feed
- [x] Remove extract from preview request
- [x] Test across multiple days
- [x] Incorporate title blacklist
- [x] Get preview namespace and filter out anything but "Main"
- [x] Add "more top read articles" :sparkles: 

image


montehurd:
Didn't see title and title prefix exclusions on your checklist.

I built these exclusion lists by setting the day to yesterday, looking at the results, seeing what ones didn't make sense (i.e. xhamster porn, "-", file pages for images, etc), then going back to the day before that and looking again, and repeated this for iirc 30 days or so:

https://github.com/wikimedia/wikipedia-ios/pull/388/files#diff-242c970b3f03da0b80d49c34a5477b41R43

montehurd:
@bgerstle 
> can haz more top read articles plz kthxbai

lol

Wikipedia/Code/MWKSearchResult.m
```diff
@@ -90,11 +96,8 @@ + (NSValueTransformer*)isDisambiguationJSONTransformer {
         }
         // HAX: occasionally the search api doesn't report back "disambiguation" page term ( T121288 ),
         // so double-check wiki data description for "disambiguation page" string.
-        NSArray* description = value[@"terms.description"];
-        if (description.count && [description.firstObject containsString:@"disambiguation page"]) {
-            return @YES;
-        }
-        return @NO;
+        NSArray* descriptions = value[@"terms.description"];
+        return @(descriptions.count && [descriptions.firstObject containsString:@"disambiguation page"]);
```
montehurd: Oops I should have done that :)

Wikipedia/Code/NSNumber+MWKTitleNamespace.h
```diff
@@ -0,0 +1,54 @@
...
+//
+
+#import <Foundation/Foundation.h>
+
+/**
+ *  List of all built-in MediaWiki namespaces.
+ *
+ *  see https://www.mediawiki.org/wiki/Manual:Namespace#Built-in_namespaces
+ */
+typedef NS_ENUM (NSInteger, MWKTitleNamespace) {
```
montehurd: Woohoo!

bgerstle:
@nirzar updated copy for section header to include language:

image

nirzar:
@bgerstle can we remove the "on" before the date. it reads oddly, Top read ON english wikipedia ON date the date can be on its own. 

and the more can be "See all top read articles"

The separator above the more is missing?

bgerstle:
>  can we remove the "on" before the date. it

@nirzar sorry, josh actually requested "for $date". would that be better or should i revert to having the date by itself?

> The separator above the more is missing?

Screenshot artifact due to less-than-natural-size scaling of simulator:

image

> and the more can be "See all top read articles"

It could, but we're not showing "all" of them.

montehurd:
@bgerstle 
> It could, but we're not showing "all" of them.

I don't think excluding the top 5 from the full list of top articles makes sense. It's actually confusing :(

bgerstle:
@nirzar @jtminor i just realized: all the other "daily" sections are formatted "Tuesday, February 16". should "most read" use the same format?

image

image

Main Page too, but I'm not including it here since it's sort of phabricated

bgerstle:
> I don't think excluding the top 5 from the full list of top articles makes sense

@montehurd Not sure what you mean, the top 5 aren't excluded. I only meant that we're not showing the full list of several hundred as returned by the pageviews API. Currently, we should the first 5, then tap "More" to reveal the longer list of 30 or so (depending on filtering, etc.)

montehurd:
@bgerstle ah gotcha

montehurd:
@bgerstle but to your point, "all" doesn't have to mean it's the same number the api returns. That's an arbitrary number.

bgerstle:
@montehurd i guess "all" is actually more accurate in this case because that's "all" of them that we're showing.

bgerstle:
@nirzar @jtminor:

image

nirzar:
@bgerstle the separator above 4th top read article is missing 
.
.
.
just kidding :P

bgerstle:
@nirzar :rage1:

coreyfloyd:
:+1:

bgerstle:
@montehurd fixed issue w/ thumbnail resolutions

PR #472, opened 2016-02-12
Perform house keeping each time the app closes

coreyfloyd:


Wikipedia/Code/WMFAppViewController.m
```diff
@@ -530,17 +530,9 @@ - (void)showNearbyListAnimated:(BOOL)animated {
...
-    if (daysSinceLastHouseKeeping > 1) {
-        //NSLog(@"Performing housekeeping...");
-        MWKDataHousekeeping* dataHouseKeeping = [[MWKDataHousekeeping alloc] init];
-        [dataHouseKeeping performHouseKeeping];
-        [[NSUserDefaults standardUserDefaults] setObject:[NSDate date] forKey:@"LastHousekeepingDate"];
-        [[NSUserDefaults standardUserDefaults] synchronize];
-    }
+- (void)performHousekeeping {
+    MWKDataHousekeeping* dataHouseKeeping = [[MWKDataHousekeeping alloc] init];
+    [dataHouseKeeping performHouseKeeping];
```
bgerstle: so `MWKDataHouskeeping` isn't really a class. it's just a function. as a boyscout, can we at least extract `performHouseKeeping` into a method on `MWKDataStore`? this way it's just: `[self.dataStore performHouseKeeping]` and we can even test it!

bgerstle: i dunno, feel free to disregard this, that class just irks me

bgerstle:
code looks good, just need to kick the tires. stepping out for a bit.

montehurd:
Seems to work fine. I reduced max history entries to 5 and minimized the app and saw the documents folders disappear for the articles being pruned.

PR #474, opened 2016-02-14
Remove MCC-MNC header enrichment

mdholloway:
This info was originally used to track changes in Zero partner
IP blocks, but is no longer used, so we're removing it.


montehurd:
Oh cool!!! Thanks Michael!!!

coreyfloyd:
thank god

coreyfloyd:
:+1:

PR #475, opened 2016-02-15
T117474: Status bar white background

nirzar:
Note: feel free to close this PR if the method used here is shabby
- Added a background view with bottom border as a subview of presentation view
- hiding the view on landscape mode
- removing the view on completion
- changed the background view tint from white to black

Bug: https://phabricator.wikimedia.org/T117474


nirzar:
simulator screen shot feb 15 2016 3 59 18 pm
simulator screen shot feb 15 2016 4 00 09 pm

montehurd:
@nirzar - oh test on iPad sim if you haven't already

nirzar:
@montehurd I did... the only problem is on iPad landscape I am hiding the view but iPad landscape doesn't hide the status bar like iPhone. 

so I wanted to use isStatusBarHidden as a condition but its not giving consisting output for that method.

coreyfloyd:
@nirzar it looks like the white background isn't under the full status bar - for that to happen this needs to be done in the presentation controller.

nirzar:
@coreyfloyd try now

coreyfloyd:
:+1:  @bgerstle @montehurd one of you guys want to check this out?

Wikipedia/Code/WMFTableOfContentsPresentationController.swift
```diff
@@ -55,11 +60,21 @@ public class WMFTableOfContentsPresentationController: UIPresentationController
...
+        self.statusBarBg.addSubview(statusBarBgBottomBorder)
+        self.backgroundView.addSubview(self.statusBarBg)
+
+
         // Hide the presenting view controller for accessibility
         self.togglePresentingViewControllerAccessibility(false)
 
         //Add shadow to the presented view
         self.presentedView()?.layer.shadowOpacity = 0.5
+        self.presentedView()?.layer.shadowOffset = CGSize.init(width: 3, height: 5)
```
bgerstle: remove `.init`, it's not necessary

Wikipedia/Code/WMFTableOfContentsPresentationController.swift
```diff
@@ -55,11 +60,21 @@ public class WMFTableOfContentsPresentationController: UIPresentationController
         self.containerView!.addSubview(self.backgroundView)
         self.containerView!.addSubview(self.presentedView()!)
         
+        // Add a status bar background so the TOC scroll below it
+        self.statusBarBg = UIView(frame: CGRectMake(self.containerView!.frame.origin.x, 0, self.containerView!.frame.size.width, UIApplication.sharedApplication().statusBarFrame.size.height))
```
bgerstle: use `CGRect()` initializer instead of `CGRectMake`. checkout [this blog post](https://medium.com/swift-programming/swift-cgrect-cgsize-cgpoint-5f4196da9cf8?utm_campaign=This%2BWeek%2Bin%2BSwift&utm_medium=web&utm_source=This_Week_in_Swift_73#.wc1vodie0) for pointers on how to CGGeometry in swift

bgerstle:
maybe not a blocker, but we should use `verticalSizeClass` property of the current size class to toggle the status bar background, which will prevent it from being hidden in `.Regular` vertical situations, like iPad landscape:

image

Wikipedia/Code/WMFTableOfContentsPresentationController.swift
```diff
@@ -30,6 +30,11 @@ public class WMFTableOfContentsPresentationController: UIPresentationController
         return view
         }()
 
+    // MARK: - Status Bar Background
+
+    var statusBarBg: UIView = UIView()
+
```
bgerstle: make this a `lazy var` like `backgroundView` instead of initializing here and then re-setting on L64. you can add it as a subview during `backgroundView`'s (lazy) initialization

Wikipedia/Code/WMFTableOfContentsPresentationController.swift
```diff
@@ -107,9 +128,32 @@ public class WMFTableOfContentsPresentationController: UIPresentationController
     
     override public func viewWillTransitionToSize(size: CGSize, withTransitionCoordinator transitionCoordinator: UIViewControllerTransitionCoordinator) {
         super.viewWillTransitionToSize(size, withTransitionCoordinator: transitionCoordinator)
-        
+
         transitionCoordinator.animateAlongsideTransition({(context: UIViewControllerTransitionCoordinatorContext!) -> Void in
             self.backgroundView.frame = self.containerView!.bounds
             }, completion:nil)
+
+        if(UIDeviceOrientationIsLandscape(UIDevice.currentDevice().orientation))
```
bgerstle: :-1: do this logic in `willTransitionToTraitCollection` and check `newCollection.verticalSizeClass` for either `.Regular` or `.Compact`

bgerstle: also FYI, device orientation != application/view orientation. the device can be in any orientation, but the application (or current view controller) can be in another based on many different variables.

bgerstle:
@nirzar left some swift comments & a comment about using trait collection instead of device orientation to adhere to best practices & fix an iPad bug

nirzar:
@bgerstle this is great, let me fix these. i was also worried about device orientation but did not know a better solution. thanks

bgerstle:
@nirzar also, are you sure it should take up the entire top of the screen? what about something like Google Sheets (i.e. only having status bar background fill the ToC's width):

sheets screenshot

nirzar:
@bgerstle i had done that earlier but it was creating some cruft with AT&T showing up halfway. their drawers are narrow.

nirzar:
@bgerstle over to you!

coreyfloyd:
@bgerstle this should be good to go

bgerstle:
:-1: iPad is still messed up :(

image

bgerstle:
the height looks normal though, which is nice

coreyfloyd:
doh

montehurd:
@nirzar hey just one thing - if you actually select items on the TOC the scrolling animation on the web view looks pretty wonky now 'cause its all scrolling by while blurred. It's especially odd if you'd been at the top of an article and you choose a TOC section way down and between the top and the chosen section there are lots of images... can you take a peek and see if this is ok? I kind of think if we're going to blur so heavily we should maybe consider turning off the scroll animation when you make a TOC selection?

nirzar:
@montehurd i actually liked that blurring out and scrolling animation together? did you see it on phone? its much smoother

coreyfloyd:
@nirzar i like the animation as well.

nirzar:
@coreyfloyd @bgerstle @montehurd shall we merge?

bgerstle:
LGTM

PR #476, opened 2016-02-16
T125462: Geocode the location name

nirzar:
**NOTE: Feel free to close this PR if the code is shabby**

- [x] Use apple geocode rated api to get name and city name of the area
- [ ] if we hit the rate limit, use the first item's name as subtitle
- [ ] handle NULL case


nirzar:
simulator screen shot feb 15 2016 8 40 54 pm

coreyfloyd:
@nirzar we have a some work to do here for caching and serializing of this info that needs to happen in other layers - so its not a straightforward ticket.

nirzar:
@coreyfloyd I suspected that, I will close this :+1:

PR #477, opened 2016-02-16
Update existing analytics to new spec

coreyfloyd:
https://phabricator.wikimedia.org/T126766

For a list of what we now support you can look at the main epic:
https://phabricator.wikimedia.org/T123679

The gist of this pr is that existing events are now switched to the context / content type / action schema for events as laid out in the spec.


Wikipedia/Code/WMFArticleListTableViewController.m
```diff
@@ -207,10 +206,6 @@ - (void)viewWillAppear:(BOOL)animated {
 
 - (void)viewDidAppear:(BOOL)animated {
```
montehurd: Remove viewDidAppear now that it's empty?

Wikipedia/Code/WMFAnalyticsLogging.h
```diff
@@ -2,7 +2,19 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-@protocol WMFAnalyticsLogging <NSObject>
+@protocol WMFAnalyticsContextProviding <NSObject>
```
montehurd: Awesome!

montehurd:
@bgerstle I'm good if you are :+1:

Wikipedia/Code/WMFAppViewController.m
```diff
@@ -301,7 +301,7 @@ - (UINavigationController*)navigationControllerForTab:(WMFAppTabType)tab {
     return (UINavigationController*)[self.rootTabBarController viewControllers][tab];
 }
 
-- (UIViewController<WMFAnalyticsLogging>*)rootViewControllerForTab:(WMFAppTabType)tab {
+- (UIViewController<WMFAnalyticsViewNameProviding>*)rootViewControllerForTab:(WMFAppTabType)tab {
     return [[[self navigationControllerForTab:tab] viewControllers] firstObject];
```
bgerstle: might want to add an assertion for `WMFAnalyticsViewNameProviding` conformance, otherwise VC's that don't conform will (silently?) fail to log

bgerstle: or blow up due to unrecognized selector, so maybe assertion isn't needed, unless we want to make that of mistake a no-op in production

Wikipedia/Code/WMFSaveButtonController.h
```diff
@@ -24,6 +24,7 @@
 /**
  *  Set to provide a source for logging saved pages
  */
-@property (weak, nonatomic) id<WMFAnalyticsLogging> analyticsSource;
+@property (weak, nonatomic) id<WMFAnalyticsContextProviding> analyticsContext;
```
bgerstle: what happens if a save occurs and either of this is `nil`? since they're `weak`, it's possible, though in practice the controller will always be deallocated with its view.  also wondering if these should be part of the designated initializer so as to enforce the API contract that all saves have a context and/or content type

Wikipedia/Code/WMFSearchResultsTableViewController.m
```diff
@@ -73,4 +73,9 @@ - (WMFEmptyViewType)emptyViewType {
     return WMFEmptyViewTypeNone; //Is controlled by the search VC
 }
 
+- (NSString*)analyticsContext {
+    return @"Search";
```
bgerstle: what's the difference between this and the `analyticsContext` provided by `WMFSearchViewController`? if they're actually meant to indicate the same context, they should be using a global constant.

Wikipedia/Code/WMFSaveButtonController.h
coreyfloyd: @bgerstle can't be in the initializer because these are recycled with cells.

bgerstle: right, but cells are always reused w/in the same view, so the context won't change during its lifetime.

bgerstle: oh i see your point, it's more important that the context/etc. are all set during configuration time, not at cell creation. just like saved page list :ok_hand:

coreyfloyd:
@bgerstle updated the patch - should be good now

bgerstle:
:+1:

PR #479, opened 2016-02-16
Capitalize first letter of continue reading descrip.

montehurd:
Before: (used tweak to always show continue reading)
screen shot 2016-02-16 at 2 03 00 pm

After:
screen shot 2016-02-16 at 2 03 31 pm


bgerstle:
:+1: but for the record, this is why I keep saying we need to DRY it up.

montehurd:
@bgerstle agreed. It's kinda silly it's persisted as an issue this long :)

PR #481, opened 2016-02-17
Make the article browser NOT the default tweak

coreyfloyd:


bgerstle:
:+1: will merge when travis reports back

PR #482, opened 2016-02-17
Fix Xcode warnings in release builds

bgerstle:
For some reason, certain warnings are only showing up in release builds.  I've fixed those, along with some others.  I also went ahead and (as follow-up from our meeting last week) removed uncrustify as a PR blocker, but made static analysis & compiler warnings PR blockers.

We might still need to hot-fix compiler warning/errors introduced to master since I can't seem to get them to trigger in both debug & release builds.  Might file a radar for this issue.


Wikipedia/Code/WMFExploreSectionSchema.m
```diff
@@ -400,7 +400,7 @@ - (nullable WMFExploreSection*)existingNearbySection {
 
     WMFExploreSection* latestMostReadSection = [self newMostReadSectionWithLatestPopulatedDate];
 
-    BOOL containsLatestSectionEquivalent = [mostReadSections bk_match:^BOOL (WMFExploreSection* mostReadSection) {
+    BOOL containsLatestSectionEquivalent = [mostReadSections bk_any:^BOOL (WMFExploreSection* mostReadSection) {
```
montehurd: ooh good catch

Wikipedia/Code/WebViewController.m
```diff
@@ -687,8 +687,11 @@ - (void)updateReferencesHeightAndBottomConstraints {
 }
 
 - (CGFloat)getRefsPanelHeight {
-    // TODO: this code could be changed to use size classes instead of orientation which is deprecated from iOS 8
+    WMF_TECH_DEBT_WARN(use size classes instead of interfaceOrientation)
+#pragma clang diagnostic push
```
montehurd: what is this wizardry?

Wikipedia/Code/WMFExploreSectionSchema.m
bgerstle: yeah, this is the warning I'd like to trigger in debug mode :confounded: don't know why it's only showing up in release. oh well, at least we caught it

Wikipedia/Code/WebViewController.m
coreyfloyd: @montehurd clang push/pop allows you to enable/disable warnings over a range of code

coreyfloyd:
:+1:

Wikipedia/Code/WebViewController.m
bgerstle: @montehurd remember, with great power, comes great responsibility.

oh ben...

Specifically, I'm using it here because I'm specifically :football: punting fixing this until later—but the rest of the app is clear of deprecation warnings. So, I'm special-casing this LOC so the rest of the app can be protected against deprecated method usage.

montehurd: Oh coolness! lol points for the gif :)

PR #483, opened 2016-02-17
Add funnel for logging retention

coreyfloyd:
https://phabricator.wikimedia.org/T126693

### DIFF
--- a/Wikipedia/Code/NSUserDefaults+WMFExtensions.swift
+++ b/Wikipedia/Code/NSUserDefaults+WMFExtensions.swift
@@ -79,5 +79,35 @@ extension NSUserDefaults {
             return false
         }
     }
+    
+    public func wmf_setAppInstallDateIfNil(date: NSDate) {
+        let previous = self.wmf_appInstallDate()
+        
+        if previous == nil {
+            self.setObject(date, forKey: "AppInstallDate")
+            self.synchronize()
+        }
+    }
+    
+    public func wmf_appInstallDate() -> NSDate? {
+        if let date = self.objectForKey("AppInstallDate") as? NSDate {
+            return date
+        }else{
+            return nil
+        }
+    }
+
+    public func wmf_setDateLastDailyLoggingStatsSent(date: NSDate) {
+        self.setObject(date, forKey: "DailyLoggingStatsDate")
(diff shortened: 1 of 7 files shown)


Wikipedia/Code/NSUserDefaults+WMFExtensions.swift
```diff
@@ -79,5 +79,35 @@ extension NSUserDefaults {
...
+        let previous = self.wmf_appInstallDate()
+        
+        if previous == nil {
+            self.setObject(date, forKey: "AppInstallDate")
+            self.synchronize()
+        }
+    }
+    
+    public func wmf_appInstallDate() -> NSDate? {
+        if let date = self.objectForKey("AppInstallDate") as? NSDate {
```
bgerstle: FYI: you can return the result of the optional cast directly

Wikipedia/Code/NSUserDefaults+WMFExtensions.swift
```diff
@@ -79,5 +79,35 @@ extension NSUserDefaults {
...
+        }
+    }
+
+    public func wmf_setDateLastDailyLoggingStatsSent(date: NSDate) {
+        self.setObject(date, forKey: "DailyLoggingStatsDate")
+        self.synchronize()
+    }
+
+    public func wmf_dateLastDailyLoggingStatsSent() -> NSDate? {
+        if let date = self.objectForKey("DailyLoggingStatsDate") as? NSDate {
```
bgerstle: same here

Wikipedia/Code/NSUserDefaults+WMFExtensions.swift
coreyfloyd: @bgerstle I did that at first and it complained - it wanted me to wrap everything in a force unwrap

bgerstle: strange, seems to work in REPL 😕

``` swift
extension NSUserDefaults {
    func foo() -> NSDate? {
        return objectForKey("Foo") as? NSDate
    }
}
NSUserDefaults.standardUserDefaults().foo()
$R1: NSDate? = nil
```

PR #487, opened 2016-02-19
Make sure changing the app project site updates the feed

coreyfloyd:
https://phabricator.wikimedia.org/T127235


bgerstle:
:+1: fixes the issue. it's unclear how much work it would take to "re-test" the search VC, but i'll consider it punted on for now

bgerstle:
oh and i almost forgot, one step closer to destroying `SessionSingleton`!!!!

PR #489, opened 2016-02-19
don't allow previews of cells which you cannot select (fixes crash)

coreyfloyd:
https://phabricator.wikimedia.org/T127258


montehurd:
Seems legit :+1:

bgerstle:
ship it!

PR #490, opened 2016-02-19
Updating some random links to https scheme

josve05a:
Just playing around a little


coreyfloyd:
@josve05a we don't directly modify anything in the pods directory. Everything in there is managed by the cocoa pods dependency manager.

josve05a:
Ah :relieved:

PR #491, opened 2016-02-19
Fix progress bar not hiding by informing delegate when an error is en…

coreyfloyd:
…countered

https://phabricator.wikimedia.org/T127237


bgerstle:
commit msg too long

bgerstle:
:+1: confirmed:

image

PR #492, opened 2016-02-19
T120474: Feed jumpiness

bgerstle:
Phab: [T120474](https://phabricator.wikimedia.org/T120474)

---

Current scroll behavior on master when scrolling doing as sections are added: https://vid.me/MVlM
New behavior: https://vid.me/h1Om

The issue was due to a couple things:

- We updated the placeholder image & nib, but not the actual `estimatedRowHeight` value
- The way we were setting items in section controllers was causing extra KVO callbacks, which caused the table to reload each. time. a. cell. was. added.  by the end, our table be all like :rage4: 

So, we just call `self.fetchError = oops` and `self.fetchedItems = yayNewStuff`, and the setters handle the internal state transitions.

Plus, now we have some tests to verify expected KVO notifications! (Hopefully they pass in Travis :pray:, if not, I'll make them Jenkins-only).

### DIFF
--- a/Wikipedia/Code/WMFBaseExploreSectionController.m
+++ b/Wikipedia/Code/WMFBaseExploreSectionController.m
@@ -67,13 +67,13 @@ static NSString* const WMFExploreSectionControllerException = @"WMFExploreSectio
              NSStringFromProtocol(@protocol(WMFExploreSectionController)),
              [self class]);
     return [NSString stringWithFormat:@"%@ identifier = %@",
-            [super description], [(id<WMFExploreSectionController>)self sectionIdentifier]];
+            [super description], [(id < WMFExploreSectionController >)self sectionIdentifier]];
 }
 
 #pragma mark - WMFBaseExploreSectionController
 
 - (BOOL)containsPlaceholders {
-    return [self.items bk_all:^BOOL(id obj) {
+    return [self.items bk_all:^BOOL (id obj) {
         return [obj isKindOfClass:[NSNumber class]];
     }];
 }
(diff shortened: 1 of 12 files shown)


Wikipedia/Code/WMFBaseExploreSectionController.m
```diff
@@ -204,61 +216,39 @@ - (AnyPromise*)fetchDataIgnoreError:(BOOL)ignoreError ignoreCurrentItems:(BOOL)i
...
-    [self setItemsToPlaceholders];
+    [self setItemsToPlaceholdersOrEmpty];
 }
 
 #pragma mark - Utility
 
-- (void)setItemsToPlaceholders {
+- (void)setItemsToPlaceholdersOrEmpty {
     if (![self shouldShowPlaceholderCell]) {
+        self.items = @[];
```
coreyfloyd: @bgerstle can you somehow indicate why you are doing this. Its not clear why we are setting this to an empty array.

bgerstle: so, `items` is `nonnull`, and if the section doesn't support placeholders, it should be an empty array. I could either to do it explicitly here, or implicitly by falling through to the logic below and assuming the non-placeholder section has `numberOfPlaceholderCells == 0`.  I'll add a comment explaining the rationale

coreyfloyd:
:+1:

coreyfloyd:
@bgerstle you're my hero

montehurd:
@bgerstle @coreyfloyd seems to have fixed the refresh jump! :)

montehurd:
:+1:

bgerstle:
@coreyfloyd anyone can look cool

PR #493, opened 2016-02-19
fix app rating on iOS 8

coreyfloyd:
we need to avoid the internal browser for the app store url

https://phabricator.wikimedia.org/T126839

### DIFF
--- a/Wikipedia/Code/WMFSettingsViewController.m
+++ b/Wikipedia/Code/WMFSettingsViewController.m
@@ -158,7 +158,7 @@ static NSString* const WMFSettingsURLSupport = @"https://donate.wikimedia.org/?u
             [self wmf_openExternalUrl:[NSURL URLWithString:WMFSettingsURLZeroFAQ]];
             break;
         case WMFSettingsMenuItemType_RateApp:
-            [[UIApplication sharedApplication] openURL:[NSURL URLWithString:WMFSettingsURLRate]]; //always exit the app
+            [self wmf_openExternalUrl:[NSURL URLWithString:WMFSettingsURLRate] useSafari:YES];
             break;
         case WMFSettingsMenuItemType_SendFeedback:
             [self wmf_openExternalUrl:[self emailURL]];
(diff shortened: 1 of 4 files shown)


Wikipedia/Code/WMFSettingsViewController.m
```diff
@@ -158,7 +158,7 @@ - (void)tableView:(UITableView*)tableView didSelectRowAtIndexPath:(NSIndexPath*)
             [self wmf_openExternalUrl:[NSURL URLWithString:WMFSettingsURLZeroFAQ]];
             break;
         case WMFSettingsMenuItemType_RateApp:
-            [self wmf_openExternalUrl:[NSURL URLWithString:WMFSettingsURLRate]];
+            [[UIApplication sharedApplication] openURL:[NSURL URLWithString:WMFSettingsURLRate]]; //always exit the app
```
bgerstle: don't we need to have in-between logic for wikipedia zero?

bgerstle: sorry but, technically yes we do. so i would recommend modifying `wmf_openExternalURL` by adding another `BOOL` parameter which tells the logic whether it should open in app or Safari

bgerstle: e.g. 

``` objc
- (void)wmf_openExternalURL:(NSURL*)url {
  [self wmf_openExternalURL:url inSafari:NO];
}

- (void)wmf_openExternalURL:(NSURL*)url inSafari:(BOOL)inSafari {
  // blah blah blah, previous logic, check `inSafari` at critical point (passing to internal methods if needed) and calling either `openURL:` or our own in-app stuff
}
```

bgerstle:
:-1: should probably use wikipedia zero alert handling for the case when a user is on a W0 network so we can warn them before leaving (as is our policy)

coreyfloyd:
@bgerstle ^

bgerstle:
code looks good, just need to remember how to test Wikipedia 0 handling. To be continued...

bgerstle:
tested wikipedia zero (with our old "faking" code) :+1: 

image

montehurd:
:+1:

PR #494, opened 2016-02-19
Make language bar only show preferred languages

coreyfloyd:
Also did some work converting the bar to a stack view.

May need to adjust the spacing distributions of the buttons before we merge this. @nirzar we should discus.

screen shot 2016-02-19 at 6 54 38 pm

### DIFF
--- a/Wikipedia/Code/WMFSearchViewController.m
+++ b/Wikipedia/Code/WMFSearchViewController.m
@@ -59,7 +59,7 @@ static NSUInteger const kWMFMinResultsBeforeAutoFullTextSearch = 12;
 @property (strong, nonatomic) IBOutlet UIButton* languageTwoButton;
 @property (strong, nonatomic) IBOutlet UIButton* languageThreeButton;
 @property (strong, nonatomic) IBOutlet UIButton* otherLanguagesButton;
-@property (strong, nonatomic)IBOutletCollection(UIButton) NSArray * languageButtons;
+@property (strong, nonatomic) IBOutletCollection(UIButton) NSArray *languageButtons;
 
 @property (nonatomic, strong) WMFSearchFetcher* fetcher;
 
@@ -541,9 +541,10 @@ static NSUInteger const kWMFMinResultsBeforeAutoFullTextSearch = 12;
     [self.languageButtons enumerateObjectsUsingBlock:^(UIButton*  _Nonnull obj, NSUInteger idx, BOOL * _Nonnull stop) {
         if(idx >= [self.searchLanguages count]){
             obj.enabled = NO;
-            [obj setTitle:nil forState:UIControlStateNormal];
+            obj.hidden = YES;
         }else{
             obj.enabled = YES;
+            obj.hidden = NO;
         }
     }];
     [self resizeLanguageButtonsIfNeeded];
--- a/Wikipedia/Code/WMFSearchViewController.m
+++ b/Wikipedia/Code/WMFSearchViewController.m
@@ -523,22 +523,28 @@ static NSUInteger const kWMFMinResultsBeforeAutoFullTextSearch = 12;
 
 #pragma mark - Languages
 
-- (NSArray*)allLanguagesFromController {
-    NSMutableArray* lang = [NSMutableArray array];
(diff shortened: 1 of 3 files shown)


Wikipedia/Code/WMFSearchViewController.m
```diff
@@ -523,22 +523,29 @@ - (void)saveLastSearch {
...
-    NSArray* languages = [self allLanguagesFromController];
+    NSArray* languages = [MWKLanguageLinkController sharedInstance].preferredLanguages;
     self.searchLanguages = [languages wmf_arrayByTrimmingToLength:3];
 }
 
 - (void)updateLanguageButtonsToPreferredLanguages {
     [self updateLanguages];
-    [self.languageButtons enumerateObjectsUsingBlock:^(UIButton* _Nonnull obj, NSUInteger idx, BOOL* _Nonnull stop) {
-        [obj setTitle:[(MWKLanguageLink*)self.searchLanguages[idx] localizedName]  forState:UIControlStateNormal];
+    [self.searchLanguages enumerateObjectsUsingBlock:^(MWKLanguageLink*  _Nonnull obj, NSUInteger idx, BOOL * _Nonnull stop) {
```
bgerstle: seems like you're enumerating twice, with each enumeration referencing the other array. could these enumerations be consolidated into one for simplicity? :

``` objc
// all we care about is configuring the buttons, so enumerate those:
[self.languageButtons enumerateObjecstsUsingBlock:^(UIButton* obj, NSUInteger idx, BOOL* _) {
  if (idx < self.searchLanguages.count) { 
    obj.hidden = NO;
    [obj setTitle:self.searchLanguages[idx].localizedName];    
  } else {
    // hide buttons which are out of bounds
    obj.hidden = YES 
  }
}];
```

I also omitted setting `enabled` in my example since hidden buttons should be disabled automatically.

bgerstle:
Seems like the buttons are not being allowed to fill to the natural size of their text:

image

bgerstle:
@coreyfloyd it seems like `.FillProportionally` is what we want for the inner stack view (holding the language buttons) but there doesn't seem to be a good setting for the outer stack view that gets the "More" button & other language buttons to play nicely.  It the outer stack view necessary, or could the "More" button but constrained to the trailing edge while the (currently "inner") stack view fills the rest of the space?

bgerstle:
seems to work in RTL:

image

coreyfloyd:
@bgerstle i asked @nirzar and he wanted left aligned. I'm not thrilled either but its ok - I think we need to update the design the buttons for the left alignment to look better…

coreyfloyd:
@bgerstle as far as them being "equal" size (which means that there can be some truncation), I think that is right… @nirzar can you comment on this?

bgerstle:
>  i asked nirzar and he wanted left aligned. I'm not thrilled either but its ok - I think we need to update the design the buttons for the left alignment to look better…

right, but this doesn't look "left-aligned" to me. IMO that would be like:

```
+---------+       +----------------+         +------+
| English | < n > | Simple English | < ... > | More |
+---------+       +----------------+         +------+
```

where the ellipsis is empty space and "More" is aligned to trailing edge.

coreyfloyd:
> right, but this doesn't look "left-aligned" to me. IMO that would be like:

@bgerstle left aligned WITH sizes and spacing are equal…

coreyfloyd:
@bgerstle also keep in mind that we have another ticket to remove names and go to only abbreviations… which will all be just a few letters… so this isn't a permanent problem.

Also the current language bar also truncates the names, so this isn't a regression either.

bgerstle:
@coreyfloyd i'm not saying it's a regression. truncation itself is not the problem, since we want truncation in cases where we need to fit all 4 bars on the screen and any titles are too long to fit the width. However, in the case I showed above, the "Simple English" button should be wider (to fit the full title) and the space between buttons should then be smaller.  

Ultimately it's up to @nirzar, but I'm fine w/ this as-is :+1:

bgerstle:
@nirzar you had an hour :stuck_out_tongue:  we can adjust distribution/alignment of this in subsequent iterations

coreyfloyd:
@bgerstle i have a follow up anyways after discussing persisting the language bar search site as well

PR #495, opened 2016-02-20
Add NSUserActivity and Spotlight Indexing Support

coreyfloyd:
https://phabricator.wikimedia.org/T127612

This supports creating user activities for the main screens of the app and then opening this content when a user taps on the item in Siri Results. 

Additionally this indexes a users saved pages for look up in spotlight as well.

Results in spotlight look like so:
screenshot 2016-02-20 15 31 58


coreyfloyd:
This is good to go… already spoke with Josh about implementing this on the weekend, but just want to go over the final implementation with him before we put it in.

montehurd:
Oh awesome!!!

coreyfloyd:
@bgerstle the tests keep timing out on on of the 8.4 builds, any ideas?

coreyfloyd:
@bgerstle @montehurd this is good to be reviewed and merged (per Josh)

coreyfloyd:
@bgerstle actually I forgot  - need to update the plist and provisioning profile with domains again to have this work.

coreyfloyd:
@bgerstle @montehurd this is ready to go - added the domains.

Wikipedia/Code/MWKSite.m
```diff
@@ -37,7 +37,7 @@ - (instancetype)initWithLanguage:(NSString*)language {
 }
 
 - (MWKSite* __nullable)initWithURL:(NSURL* __nonnull)url {
-    NSArray* hostComponents = [url.host componentsSeparatedByString:@"."];
+    NSMutableArray* hostComponents = [[url.host componentsSeparatedByString:@"."] mutableCopy];
```
bgerstle: why are we stripping the "m" compoment

coreyfloyd:
@bgerstle pushed up some fixes

Wikipedia/Code/MWKSite.m
coreyfloyd: answered on hangout, but for posterity:

The code below this method doesn't work with mobile URLs and was the source of a bug where mobile URLs were never parsed properly.

bgerstle:
thanks @coreyfloyd, none of my other comments were really blocking, and this can be refined/cleaned up over subsequent iterations.

PR #496, opened 2016-02-21
Share and share alike

coreyfloyd:
https://phabricator.wikimedia.org/T107074

Revamp of sharing across the app…

This was a bit messy but I did my best to implement this following some basic principles: Predictable, Contextual, Simple

In essence we have 3 contexts for sharing:

## 1. Tap Share on an Article
Easy and the biggest use case. In this case we want to share the page url and the title of the article. This also enables adding to the reading list and opening in Safari. 

Of note, we are not sharing the selected text, if any. Rationale: for one no other apps connect selected text with the tool bar button. This includes safari and medium. Sharing selected text is solely done from the copy/paste menu. This makes sense since the button and text are separated in space. Secondly, this greatly simplifies the process of sharing by being more contextual. 

## 2. Select text and Share
This actually has 2 sub-cases
### a. Share the selected text
This is enabled via the standard UIWebview text selection menu item "Share…"
### b. Share-A-Fact™
This is our custom handler… we add a menu item (and name it "Share-A-Fact" to differentiate it) The user is presented with a preview, then after confirming the standard share sheet is shown. This is essentially the pre-existing behavior except the share as text option is removed as it is taken care of in case 2.a.

## 3. Tap Share on an Image
This is new! A sharing button is added to the image gallery to allow users to share the specific image thy are looking at. When possible, the info page URL is also provide for licensing information.

## Other Share A Fact Notes
Overall this flow seems pretty natural. But definitely affects Share-A-Fact in a couple ways.

1. One fall out of this flow is that Share-A-Fact only ever uses text that was selected by the user. In the current production app, it is possible to make a share card with automatically extracted text, which when
you think about it is not a good idea. Share a fact is about composing an image and quote - auto filling in the text is antithetical to this.

2. Share-A-Fact is now only available via the text selection menu. This seems to be best as it is the most contextual place to put it, but text selection is a relatively rare action. 

At first I tried to make a share a fact activity for the share sheet, but this was clunky. Doing this the workflow was:
1. Tap Share
2. Show Share Sheet
3. Tap Share A Fact activity
4. Show Share A Fact Preview
5. Tap Share As Image
6. Show Share Sheet
7. Select the sharing service

This big problem here is that the Share Sheet was shown twice. That was just awkward, which lead me to the current implementation.

Some things we need to do to get this in the app:
- [x] Agree on share a fact implementation
- [x] Add properly sized share image
- [x] Discus image sharing / attribution

NOTE: This pr also this fixes a bug where the photo owner button in the gallery was untappable. 

Some previews…

Standard Article Sharing
---
screenshot 2016-02-21 00 48 58

Gallery Overlay (share button up top)
---
screenshot 2016-02-21 00 48 58 2

Gallery Share Sheet
---
screenshot 2016-02-21 00 48 59

Text Selection Menu
---
screenshot 2016-02-21 00 49 01

Share A Fact Preview
---
screenshot 2016-02-21 00 49 04

Share A Fact Share Sheet
---
screenshot 2016-02-21 00 52 30


nirzar:
@coreyfloyd hey, there seems to be a lot of design considerations here. is this somewhere on Phab where we can discuss?

coreyfloyd:
@nirzar I inserted the phab card at the top. Yeah - this needs some discussion -  I did what seems to make the most sense in each context.  But there are some things to to hash out for sure.

montehurd:
oh man, this is great!

coreyfloyd:
@nirzar ^ updated with the updated suggestions from the meeting: 
screenshot 2016-02-24 16 27 31
screenshot 2016-02-24 16 27 27

coreyfloyd:
@nirzar can you serve up a share icon? (we can't use the built in one here - it can only be used in toolbars)

bgerstle:
image

bgerstle:
:+1: looks good! now we just need an "Open in Wikipedia" extension to go from safari to the app :wink:

jtminor:
Okay, still not a fan of last minute risk here, but I'm over it.

I pulled this and couple small details I noticed:
 - Please label highlighted text option as "Share-a-fact"
 - There seems to be inconsistency in formatting of text strings. When "sharing as text" (highlight->share->share as text) to email, I don't see the wprov=sfti1 parameter. I'm not sure how this variable is being used, but I think we should be consistent about any tracking of inbounds from shares we produce. Suggest we add this param to any share string.

bgerstle:
@montehurd ball's in your court :basketball:

montehurd:
One last small issue.

**Email from toolbar** (has string regression - see next image)
img_1718

**Email from sharing by share a fact**
img_1719

coreyfloyd:
@montehurd no italics… just text now

montehurd:
To be clear, the regression is...

`https://en.wikipedia.org/wiki/Halo_wars?wprov=sfti1`
`<i>Halo wars</i>`

Which should be this...

`"Halo wars" on @Wikipedia:`
`https://en.wikipedia.org/wiki/Halo_wars?wprov=sfti1`

(html is stripped and `"#title#" on @Wikipedia:` comes on line before url)

Confirmed with Josh.

coreyfloyd:
@montehurd ahh - ok - read that backwards…

montehurd:
Woohoo! :+1:

bgerstle:
:shipit:

PR #498, opened 2016-02-21
Add multiple picture of the day sections

coreyfloyd:
Couldn't find a ticket for this, but know we have discussed it… so I made one:
https://phabricator.wikimedia.org/T127662

In case you were wondering… Tasha is out of town for the week. Too much free time.


montehurd:
Ya I was wondering :) Glad you found some fun ones!

PR #500, opened 2016-02-21
Long pending icon fixes

nirzar:
The icons have gone through a lot because of different times we added them. now that the scope is finalized, i have refined the icons on single grid. 

- Pixel pushing the mini icons
- Figured out wrong names in two places
- Updated trending icon
- Quick actions icon are pixel fitted


nirzar:
image
Quick action icons are all line icons with widths matched
------------------------------------
Earlier mini icons were
Random - filled
featured - filled
nearby - line
potd - line

Now all of them are filled because they need some weight on the side of the header.
The styles match.
---------
The main page of wikipedia was using featured icon. now it uses the wikipedia layout icon
---------
The color code for custom grey was wrong. 
---------

PR #502, opened 2016-02-22
Fix app shortcut showing black screen of death or doing nothing

coreyfloyd:
When we moved the "app resume logic" to work on the didenterforeground notification, that inadvertently caused the "processshortcutitem" logic to run before the shortcut was ever delivered to the app.
This caused unpredictable behavior.

In the end, this is much cleaner and gets the shortcut processing out of the main resume logic - and removed 2 ivars…


bgerstle:
:-1: only partially fixes the issue. shortcuts work while the app is running, but nothing happens when a shortcut is used to **launch** the app.

coreyfloyd:
@bgerstle doh

bgerstle:
:+1: works! @coreyfloyd verified data migration, i also checked (among other things):

- Opening an article
- Quitting the app
- Launching via "Search" shortcut item
- :white_check_mark: App launched to search view
- :white_check_mark: Dismissing search revealed the last open article

PR #503, opened 2016-02-22
T124820: Table of contents close button

nirzar:
- [x] Add close button and tap action
- [x] Align it on portrait and landscape on iPhone and iPad
- [x] add accessibility label and hints

### DIFF
--- a/Wikipedia/Code/WMFTableOfContentsPresentationController.swift
+++ b/Wikipedia/Code/WMFTableOfContentsPresentationController.swift
@@ -21,10 +21,11 @@ public class WMFTableOfContentsPresentationController: UIPresentationController 
     public var minimumVisibleBackgroundWidth: CGFloat = 60.0
     public var maximumTableOfContentsWidth: CGFloat = 300.0
     public var closeButtonPadding: CGFloat = 10.0
+    public var statusBarEstimatedHeight: CGFloat = 20.0
     
     // MARK: - Views
     lazy var statusBarBackground: UIView = {
-        let view = UIView(frame: CGRect(x: CGRectGetMinX(self.containerView!.bounds), y: CGRectGetMinY(self.containerView!.bounds), width: CGRectGetWidth(self.containerView!.bounds), height: 20.0))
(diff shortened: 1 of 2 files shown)


nirzar:
simulator screen shot feb 22 2016 1 35 34 pm

coreyfloyd:
@nirzar why not upper left? Its not idiomatic for the platform to have a close button in the bottom left.

nirzar:
@coreyfloyd your finger is already at bottom right now as you pressed the toc on bottom; if not used the swipe gesture but gesture person uses gestures to dismiss too. _sounds like a twisted logic. _

But I don't have strong opinion on this.

coreyfloyd:
@nirzar yeah - i think the thumb reach would be applicable if we didn't have the gesture. Since we have the gesture though, it serves basically the same purpose - makes it easy to close the ToC without having to awkwardly reach the upper left.

When Apple started making screens that were too big for people to reach the upper left, they introduced the swipe back gesture - what they didn't do was move the "back" button to the bottom left. I think they felt that the upper left back button was sacred… because that was such a strong well understood behavior (upper left always gets you back to where you started).

hhanke:
@nirzar @coreyfloyd I think from the accessibility standpoint of view, top left corner would be better than bottom left corner. This is where users will be expecting it and with people with disabilities, you can never assume that they will see the whole screen at once (e.g. blind VoiceOver users only "see" (hear) the current item and nothing else, Zoom users only see one section of the screen etc.) Finally, I think that even for healthy users it is easier to notice at the more usual place.

I also see the rationale for placing it where you put it, but maybe that is solved by the gesture (or even tap outside) as @coreyfloyd suggests.

In any case, it will be a great improvement! Thank you very much!

bgerstle:
sorry @nirzar, looks like you're outnumbered. I side with @coreyfloyd and @hhanke as well. please move to top left.

nirzar:
moved to top!

coreyfloyd:
:+1:

PR #504, opened 2016-02-22
Pull to refresh is here

nirzar:
- Added refresh to scrollview of webview
- Hooked up original refresh action
- Removed the refresh from toolbar

NOTE: refresh button was never meant for article toolbar. this is almost a regression issue over the app store version.


Wikipedia/Code/WMFArticleViewController.m
```diff
@@ -705,6 +710,7 @@ - (void)fetchArticleForce:(BOOL)force {
 
 - (void)fetchArticle {
     [self fetchArticleForce:YES];
+    [_pullToRefresh endRefreshing];
```
bgerstle: shouldn't refresh stay visible until article finishes being downloaded?

Wikipedia/Code/WMFArticleViewController.h
```diff
@@ -32,6 +32,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (nonatomic, strong, readonly) MWKTitle* articleTitle;
 @property (nonatomic, strong, readonly) MWKDataStore* dataStore;
+@property (nonatomic, strong, readonly) UIRefreshControl* pullToRefresh;
```
bgerstle: need to remove `refreshToolbarItem` getter method as well

Wikipedia/Code/WMFArticleViewController.m
nirzar: When the refresh is ended the progress bar takes over and hence there is no need for showing the  activity indicator.

coreyfloyd:
:+1:

PR #505, opened 2016-02-22
Refactor Wikipedia Zero testing into tweak

bgerstle:
Implemented as part of testing "Rate the App" fixes (#493), and liked it so much I figured we should keep it.  The category only exists because using the same Tweak in two places is poorly documented and I can't find an example, so I decided to do it manually.

### DIFF
--- a/Wikipedia/Code/FBTweak+WikipediaZero.m
+++ b/Wikipedia/Code/FBTweak+WikipediaZero.m
@@ -1,0 +1,36 @@
+//
+//  FBTweak+WikipediaZero.m
+//  Wikipedia
+//
+//  Created by Brian Gerstle on 2/22/16.
+//  Copyright © 2016 Wikimedia Foundation. All rights reserved.
+//
+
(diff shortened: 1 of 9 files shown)


coreyfloyd:
Nice 💪💪

Wikipedia/Code/FBTweak+WikipediaZero.m
```diff
@@ -0,0 +1,36 @@
...
+//  Wikipedia
+//
+//  Created by Brian Gerstle on 2/22/16.
+//  Copyright © 2016 Wikimedia Foundation. All rights reserved.
+//
+
+#import "FBTweak+WikipediaZero.h"
+#import <Tweaks/FBTweakInline.h>
+
+static NSString* const WMFWikipediaZeroHeaderTweakIdentifier = @"org.wikimedia.wikipedia.tweaks.networking.zero.fakezeorheaders";
```
montehurd: @bgerstle typo in `fakezeorheaders`

PR #507, opened 2016-02-23
Restyle welcome screens

montehurd:
- [x] Update strings
- [x] Update images/icons
- [x] Restyle all 3 welcome screens
- [x] Ensure layout on small screens and iPads isn't explody - i.e. doesn't stretch to full width on iPads, doesn't clip top or bottom of content on 4s, etc.
- [x] Fix crashing bug on existing welcome lang picker (see https://github.com/wikimedia/wikipedia-ios/commit/1e5bb13f0d4b0c5f6ab605eafbc40d84c87c4ce7) - prob need a test case for TSG for this.
- [x] Use WMF green for switches system-wide (and use for a couple Settings icons)
- [x] Add and hook up back button from mocks at top of second 2 screens
- [x] Add "Thank you" flourish when opt-ing in to send usage reports
- [x] <strike>Auto-hide welcome screens after opt-in, per the mock video</strike> **Changed the 'Send usage reports' message instead**
- [x] Double-check for RTL issues (identical to `UINavigationController` behavior)
  - [x] iOS 9 RTL
    - [x] Push starts on left (trailing)
    - [x] Pop starts on right (leading)
    - [x] Back button pinned to right (leading)
  - [x] iOS 8 RTL
    - [x] Back button pinned to left (not flipped for RTL, mirror
- [x] DRY up a couple colors which are presently in the storyboard - move to UIColor category
- [x] Use built-in navigation bar for back button & better transitions
- [x] Enable swipe to go back
- [x] Use light status bar on first panel, dark on the rest

Note: animations (will be follow-on ticket) 

**iPhone 6**
1 mov

**iPad**
2 mov

**iPhone 4s**
0 mov


coreyfloyd:
@montehurd looking good

Wikipedia/Code/WMFWelcomeLanguageViewController.m
```diff
@@ -22,38 +22,47 @@ - (void)viewDidLoad {
...
-    }
+    self.subTitleLabel.text = MWLocalizedString(@"welcome-languages-sub-title", nil);
+
+    [self.moreLanguagesButton setTitle:MWLocalizedString(@"welcome-languages-add-button", nil)
+                              forState:UIControlStateNormal];
+    
+    [self.nextStepButton setTitle:[MWLocalizedString(@"welcome-languages-continue-button", nil) uppercaseStringWithLocale:[NSLocale currentLocale]]
+                         forState:UIControlStateNormal];
+    
+    self.nextStepButton.backgroundColor = [UIColor wmf_colorWithHex:0xE8F3FE alpha:1.0];
```
bgerstle: something to keep in mind: all the "next step" buttons share some common elements. maybe once this starts to solidify, you can go back and refactor the common elements into color/font style categories and maybe a `UIButton` factory method.

montehurd: Oh yeah that's one of the things in my checklist - DRY'ing up some colors and such.

Wikipedia/Code/WMFWelcomeAnalyticsViewController.m
```diff
@@ -33,20 +34,42 @@ - (void)viewDidLoad {
...
 }
 
-- (IBAction)toggleAnalytics:(id)sender {
-    if ([(UISwitch*)sender isOn]) {
+- (IBAction)toggleAnalytics:(UISwitch*)sender {
+    if ([sender isOn]) {
         [[BITHockeyManager sharedHockeyManager] crashManager].crashManagerStatus = BITCrashManagerStatusAutoSend;
         [[NSUserDefaults standardUserDefaults] wmf_setSendUsageReports:YES];
+        
+        dispatchOnMainQueueAfterDelayInSeconds(1.4, ^{
```
coreyfloyd: Are we auto pushing to the next VC when someone toggles a switch?

montehurd: It's in the mock video (dismissing the welcome screens a second or so after activating the switch), but we could instead just show the "thank you" text for a second then go back to having the button say "continue", and then the user would have to tap it to dismiss the welcome screens.

bgerstle: is it possible to dismiss before the block fires? if so, we should use `@weakify` here to ensure that popping the VC & its view deallocates it, and prevents this from being called twice.

Wikipedia/Code/WMFWelcomeAnalyticsViewController.m
```diff
@@ -33,20 +34,42 @@ - (void)viewDidLoad {
...
     [self wmf_openExternalUrl:[NSURL URLWithString:URL_PRIVACY_POLICY]];
 }
 
+-(void)updateNextStepButtonStyleForUsageReportsIsOn:(BOOL)isOn {
+    NSString* buttonTitle = isOn ? [MWLocalizedString(@"welcome-volunteer-thanks-button", nil) stringByReplacingOccurrencesOfString:@"$1" withString:@"😀"] : MWLocalizedString(@"welcome-volunteer-continue-button", nil);
+
+    [self.nextStepButton setTitle:[buttonTitle uppercaseStringWithLocale:[NSLocale currentLocale]]
+                         forState:UIControlStateNormal];
+    
+    self.nextStepButton.backgroundColor = isOn ? [UIColor wmf_welcomeThanksButtonBackgroundColor] : [UIColor wmf_welcomeNextButtonBackgroundColor];
```
bgerstle: nice use of ternary statements :-)

Wikipedia/Code/WMFWelcomeAnalyticsViewController.m
bgerstle: @nirzar @montehurd i'm not so sure we should be auto-dismissing from a switch interaction. can you think of any examples where Apple does this? HUI guidelines generally discourage doing things on the user's behalf like this.

Wikipedia/Images.xcassets/welcome-bg.imageset/Contents.json
```diff
@@ -0,0 +1,21 @@
...
+      "idiom" : "universal",
+      "scale" : "1x"
+    },
+    {
+      "idiom" : "universal",
+      "scale" : "2x"
+    },
+    {
+      "idiom" : "universal",
+      "filename" : "welcome-bg@3x.jpg",
```
bgerstle: no PDF? missing 1x & 2x images?

Wikipedia/Code/WMFWelcomeLanguageViewController.m
```diff
@@ -106,4 +122,8 @@ - (void)languagesController:(LanguagesViewController*)controller didSelectLangua
     [controller dismissViewControllerAnimated:YES completion:NULL];
 }
 
+-(IBAction)prepareForUnwindFromVolunteer:(UIStoryboardSegue *)segue {
+    // Needed so we can set up unwind segue from storyboard.
```
bgerstle: not seeing any setup happening here. can this be removed?

bgerstle:
:-1: RTL transitions are broken.  iOS 9 should be identical to `UINavigationController` (push happens left-to-right). 

on iOS 8, we shouldn't flip the chevron position or direction:

image

also (IMHO) this is more of a design critique (@nirzar), but the chevron should be in the same spot it would be if there were a navigation bar. that's where users would intuitively reach to "go back."  also, it's not strictly a requirement, but we have a "back button" and transitions essentially identical to `UINavigationController`, but no swipe navigation—which IMO users would expect (if it looks like a duck, quacks like a duck...).

I normally wouldn't bring up design critique in code review, but here I'm doing it here to make the point: all of the above would "just work" if we used a `UINavigationController` instead of storyboards & custom segues.  So, we can either fix & re-implement all of this logic, or refactor it to use a navigation controller.  IMO the second is the cheapest option.

bgerstle:
@nirzar @montehurd minor design/implementation suggestion, what do you think of different status bar styles per welcome view?  for example, the first should be "light content" (white status bar text on top of dark background), and the others should be "default" (dark status bar text on top of light background).

for example:

Explore view with "light content" status bar
image

Languages view with "default" status bar
image

Compared to current "Explore" view:
image

bgerstle:
I went and peeked at the storyboard, and it turns out we _are_ using a `UINavigationController`, but we're using custom animators for the transitions.  @coreyfloyd I'm not seeing any docs or comments explaining why those were used in the first place, and I suppose we never got around to "fixing" welcome RTL compliance as part of the [RTL epic](https://phabricator.wikimedia.org/T116989), but seems like now's the time to do it.

Wikipedia/Code/WMFWelcomeAnalyticsViewController.m
coreyfloyd: @nirzar @montehurd agree with @bgerstle here. Users just don't expect a view to disappear because the flip a switch. 

If we really want it to disappear on interaction then there show be 2 buttons - and tapping yes/no would then dismiss.

bgerstle:
other than all that, looks REALLY NICE!!!! :sparkles::art::sparkles:

coreyfloyd:
>  @coreyfloyd I'm not seeing any docs or comments explaining why those were used in the first place, and I suppose we never got around to "fixing" welcome RTL compliance as part of the RTL epic, but seems like now's the time to do it.

That custom navigator is just for making the transition not look like the new view is not covering the old view during the transition (it's meant to look more like a page view controller than a navigation animation)

I agree that we probably should be showing the navigation bar here - just with a transparent background. That would solve these RTL issues… and should be relatively quick.

bgerstle:
>
> I agree that we probably should be showing the navigation bar here - just
> with a transparent background. That would solve these RTL issues… and
> should be relatively quick.


Except that we need to also animate from a different starting point for RTL
vs. LTR on iOS 9.

Another minor (potentially premature) point, we could probably optimize
initial load time greatly by only creating the first welcome view at
startup, instead of deserializing all the views at once from the
storyboard.  I'd have to measure startup time on older devices to confirm,
but experience suggests this should be a non-trivial speed improvement.
Depending on how the cookie crumbles, this might be easier to do, but
something to keep in the back of our minds.

On Wed, Feb 24, 2016 at 11:17 AM, Corey Floyd <notifications@github.com>
wrote:

> @coreyfloyd <https://github.com/coreyfloyd> I'm not seeing any docs or
> comments explaining why those were used in the first place, and I suppose
> we never got around to "fixing" welcome RTL compliance as part of the RTL
> epic, but seems like now's the time to do it.
>
> That custom navigator is just for making the transition not look like the
> new view is not covering the old view during the transition (it's meant to
> look more like a page view controller than a navigation animation)
>
> I agree that we probably should be showing the navigation bar here - just
> with a transparent background. That would solve these RTL issues… and
> should be relatively quick.
>
> —
> <https://github.com/wikimedia/wikipedia-ios/pull/507#issuecomment-188329071>
> .
>



-- 
IRC: bgerstle

bgerstle:
>
> instead of deserializing all the views at once from the storyboard.  I'd
> have to measure startup time on older devices to confirm


in fairness, that assumes the storyboard is "dumb" about deserializing VC's
which aren't the initial view controller. disregard my premature
optimization comment. this can be investigated if we get reports of poor
startup performance when presenting FTUE

On Wed, Feb 24, 2016 at 11:21 AM, Brian Gerstle <bgerstle@wikimedia.org>
wrote:

> I agree that we probably should be showing the navigation bar here - just
>> with a transparent background. That would solve these RTL issues… and
>> should be relatively quick.
>
>
> Except that we need to also animate from a different starting point for
> RTL vs. LTR on iOS 9.
>
> Another minor (potentially premature) point, we could probably optimize
> initial load time greatly by only creating the first welcome view at
> startup, instead of deserializing all the views at once from the
> storyboard.  I'd have to measure startup time on older devices to confirm,
> but experience suggests this should be a non-trivial speed improvement.
> Depending on how the cookie crumbles, this might be easier to do, but
> something to keep in the back of our minds.
>
> On Wed, Feb 24, 2016 at 11:17 AM, Corey Floyd <notifications@github.com>
> wrote:
>
>> @coreyfloyd <https://github.com/coreyfloyd> I'm not seeing any docs or
>> comments explaining why those were used in the first place, and I suppose
>> we never got around to "fixing" welcome RTL compliance as part of the RTL
>> epic, but seems like now's the time to do it.
>>
>> That custom navigator is just for making the transition not look like the
>> new view is not covering the old view during the transition (it's meant to
>> look more like a page view controller than a navigation animation)
>>
>> I agree that we probably should be showing the navigation bar here - just
>> with a transparent background. That would solve these RTL issues… and
>> should be relatively quick.
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/wikimedia/wikipedia-ios/pull/507#issuecomment-188329071>
>> .
>>
>
>
>
> --
> EN Wikipedia user page: https://en.wikipedia.org/wiki/User:Brian.gerstle
> IRC: bgerstle
>



-- 
IRC: bgerstle

coreyfloyd:
> in fairness, that assumes the storyboard is "dumb" about deserializing VC's
which aren't the initial view controller. disregard my premature
optimization comment. this can be investigated if we get reports of poor
startup performance when presenting FTUE

@bgerstle Yeah agreed… this has pre-optimization written all over it.

> Except that we need to also animate from a different starting point for RTL
vs. LTR on iOS 9.

Yeah thats also pretty easy - its just checking one more bool to decide the direction - the animations themselves don't change.

bgerstle:
> it's meant to look more like a page view controller than a navigation animation

yeah, it's either going to be a `UIPageViewController` w/ a toolbar and custom view controller delegate logic for RTL support, or a `UINavigationController` with custom animation w/ RTL support.  now having gotten a grasp on all the variables, it does seem like the latter would require less code.

@coreyfloyd @montehurd should we merge this once the auto-dismiss issue is resolved and do RTL support as a follow-on? this (accidentally) ended up happening for TOC close button, so i wouldn't be against it, as long as we make it clear in Phab

coreyfloyd:
@bgerstle rtl nav transition is done ^

bgerstle:
@coreyfloyd :-1: sorry, not complete:

- pop needs to also reverse direction in RTL on iOS 9
- transitions & back button shouldn't be flipped in iOS 8 (unless we want it to be inconsistent w/ every other navigation controller in the app)

bgerstle:
another reason to use a navigation bar for the "back" button, is that you get a new chevron animating in for each view, as opposed to the nicer navigation bar transitions that come with `UINavigationController`

bgerstle:
example RTL navigation transition in iOS 9:

iOS 9 RTL

coreyfloyd:
@bgerstle  Fixed the pop and only flipping in iOS 9

I'll let @montehurd handle anything else RTL - just wanted to make sure the transition was ok to use. So he can switch to a nav bar.

bgerstle:
@coreyfloyd @montehurd transitions look good :ok_hand:, just need to fix back button and, if we agree, use a navigation bar to enable swipe navigation, fix back button position, and have cooler transitions

Also, I'm adding a small change to make it easier for devs to test welcome view w/o reinstalling the app. just modify the xcode scheme and add `WMFShowWelcomeView` environment variable and set to `YES`.

nirzar:
yeah the statusbar can be light content for first screen. the white background is reducing impact of the image.

bgerstle:
@montehurd updated PR checklist w/ blocking amendments. i'll file a subtask for status bar follow-on

bgerstle:
@montehurd also updated "auto-hide" checklist item and changed to pending

Wikipedia/Code/WMFWelcomeLanguageViewController.m
montehurd: Had to add a method with this signature for the storyboard to be able to see this as a segue unwind destination:

untitled mov

Per: https://spin.atomicobject.com/2014/10/25/ios-unwind-segues/

There may be other way to do this, but I think this is the standard way?

montehurd:
Note: gifs have been updated to latest.

coreyfloyd:
:+1:

PR #512, opened 2016-02-23
Fix nearby not showing up on clean install

bgerstle:
As part of this, I wanted to take some reasonable precautions to ensure I didn't introduce any unexpected behaviors w/ the feed initial state—so I added tests!  There are quite a few pending ones still, but I think this is a good start. 

As an added bonus, this also fixes the "Most Read" section being in the wrong place in the feed.


coreyfloyd:
:+1:

PR #515, opened 2016-02-24
Language bar follow on

coreyfloyd:
app language is always first

last language selection is persisted

fix up of layout of language bar

### DIFF
--- a/Wikipedia/Code/MWKLanguageLinkController.h
+++ b/Wikipedia/Code/MWKLanguageLinkController.h
@@ -62,6 +62,9 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (BOOL)languageIsOSLanguage:(MWKLanguageLink*)language;
 
+
+- (nullable MWKLanguageLink*)languageForSite:(MWKSite*)site;
+
 @end
 
 NS_ASSUME_NONNULL_END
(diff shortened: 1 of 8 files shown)


bgerstle:
:-1: Wasn't sure if your intent was to fix the truncation issue, but IMO this is a regression compared to how it used to behave (truncating Simple English instead of English).

image

Also, selecting different languages "expands" the English button, but the underline doesn't seem to match:

image

bgerstle:
you might find it helpful to add a couple visual tests for proper layout & compression/truncation behavior. for example:

- populating language bar with 2 languages ensures they expand to their intrinsic size, leaving white space until "More" button which is pinned on trailing edge
- populating language bar with 4 (max?) languages
  - all equal lengths, fitting inside bar
  - 1st is longer, and should be truncated
  - 2nd is longer
  - etc.

coreyfloyd:
@bgerstle ugh… the layout keeps changing… hmm… investigating.

coreyfloyd:
@bgerstle fixed… Now it never truncates the first language. And truncates the other buttons as needed.

I'm still not sure this will be what we want but should be good enough for now… as far as tests… :football: until after release - I have a lot of other work to do before Friday

bgerstle:
@coreyfloyd like the new overflow button too!

bgerstle:
oh wait, @coreyfloyd the more button was actually truncated:

image

and expanded in landscape:

image

coreyfloyd:
LOLOL

bgerstle:
ruh roh, if i select a _really_ long language, i can't even select new ones or see all the other "preferred" languages:

image

until I re-display search:

image

bgerstle:
removed old comments since they were false test results. I've pulled the latest and it works! :+1: sorry for the mixup

PR #517, opened 2016-02-24
install nodejs through travis

paladox:
This update allows npm to be run but also moves npm to parent directory.


bgerstle:
@paladox could you explain the rationale behind these changes?  We don't currently need travis to install or run npm  because Grunt output is committed to the repo. This allows us to checkout the repo and build w/o installing or running anything.

paladox:
I'm just matching changes I'm doing for the android version in gerrit.

bgerstle:
@paladox I see, looks like you're mainly trying to update our npm dependencies, can you please confirm?  If that's the case, I'd prefer to close this for now, as we're about to cut a build for a major release and I don't want to introduce any unnecessary risk.

paladox:
Ok. This patch should not cause any problems for anyone. And yes this patch also updates some packages but should not change anything or cause any problems.

bgerstle:
@paladox sorry, but we really can't take the chance right now unless there's a specific issue you're fixing by updating the dependencies. thanks again for the thought, but I'm going to close this since the build isn't passing and we don't want to update dependencies right now.

PR #518, opened 2016-02-24
Update .travis.yml

paladox:


bgerstle:
@paladox we do need to update to iOS 9.3 & Xcode 7.3, but we shouldn't run iOS 9.2 builds in addition to 9.3.  Also, there will be more changes necessary (like re-recording all snapshot tests) when we update to iOS 9.3.

.travis.yml
```diff
@@ -1,6 +1,9 @@
 language: objective-c
-osx_image: xcode7.2
+osx_image: xcode7.3
 env:
+  - IOS_VER="9.3" FL_ARGS="verify"
+  - IOS_VER="9.3" FL_ARGS="analyze"
+  - IOS_VER="9.3" FL_ARGS="verify scheme:WikipediaRTL"
   - IOS_VER="9.2" FL_ARGS="verify"
```
bgerstle: these should be removed in favor of 9.3

bgerstle:
looks like there are some nullability warnings causing compilation errors. as we're not actively using the Xcode 7.3 beta, I'm going to close this for now. the team can reopen once we decide to bump the required Xcode version to 7.3 and start developing & testing on iOS 9.3.

PR #520, opened 2016-02-25
Fix deadlocks in SDWebImage

bgerstle:
Deadlock was due to unnecessarily performing cancellation logic on internal thread, which was deadlocked due to the `cancel` method having already required the lock it needs to do its work.

See example of reproducing the deadlock by [scrolling in search on master](https://vid.me/GRzC) vs. [scrolling in search with these fixes](https://vid.me/pfJe).


coreyfloyd:
:+1:

montehurd:
SHIPIT!

PR #525, opened 2016-02-25
Fix frozen compasses

bgerstle:
Use magnetic heading if true heading isn't available.  Also, only report "accurate" heading values.  This is all per [Apple's recommendations](https://developer.apple.com/library/ios/documentation/UserExperience/Conceptual/LocationAwarenessPG/GettingHeadings/GettingHeadings.html#//apple_ref/doc/uid/TP40009497-CH5-SW56) for handling heading updates.

### DIFF
--- a/Wikipedia/Code/CLLocation+WMFBearing.m
+++ b/Wikipedia/Code/CLLocation+WMFBearing.m
@@ -25,7 +25,14 @@
 
 - (CLLocationDegrees)wmf_bearingToLocation:(CLLocation*)location
                          forCurrentHeading:(CLHeading*)currentHeading {
-    return [self wmf_bearingToLocation:location] - currentHeading.trueHeading;
+    CLLocationDegrees bearing = [self wmf_bearingToLocation:location];
+
+    // use true heading if available, otherwise fall back to magnetic heading
+    if (currentHeading.trueHeading >= 0) {
+        return bearing - currentHeading.trueHeading;
+    } else {
+        return bearing - currentHeading.magneticHeading;
+    }
 }
 
 @end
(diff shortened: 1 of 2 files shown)


coreyfloyd:
brrrrrr… :snowman: 

compass why so cold?

montehurd:
:-1: Not fixed :(

Worked for a minute. Then I locked and unlocked the phone and all compasses arrows froze.

Also distances don't seem to be updating at all.

montehurd:
Ok some after some walk-about time here's what I'm seeing:

If I walk around with the full list of nearby items, the compasses work, but distance only updates if I lock the phone and unlock it.

If I walk around with the explore nearby onscreen the compass works, but distance doesn't update, and if I lock the phone and unlock it the compass arrows all freeze and the distances all go to -3.

bgerstle:
As I said, I think there are two issues here. I suggest we merge this if it
fixes compasses on @nirzar and @montehurd's devices. @montehurd it would be
best if you could debug the remaining issues on your device, as I do think
this is device-specific. We can pair on it tomorrow as well.

On Thursday, February 25, 2016, montehurd <notifications@github.com> wrote:

> Ok some after some walk-about time here's what I'm seeing:
>
> If I walk around with the full list of nearby items, the compasses work,
> but distance only updates if I lock the phone and unlock it.
>
> If I walk around with the explore nearby onscreen the compass works, but
> distance doesn't update, and if I lock the phone and unlock it the compass
> arrows all freeze and the distances all go to -3.
>
> —
> <https://github.com/wikimedia/wikipedia-ios/pull/525#issuecomment-189037560>
> .
>


-- 
IRC: bgerstle

montehurd:
@bgerstle sounds good!

PR #526, opened 2016-02-26
Enable piwik on production

coreyfloyd:
Keys for piwik and hockey are now injected by the CI server


coreyfloyd:
@bgerstle ok - good to go now

PR #529, opened 2016-02-26

niedzielski:
Sorry for the lateness of this patch! Meetings ran long today in light of recent events :/


coreyfloyd:
@niedzielski you are fired

PR #530, opened 2016-02-26
Fix nearby section not updating when shown & distance update resolution

bgerstle:
Phab:
- https://phabricator.wikimedia.org/T128410
- https://phabricator.wikimedia.org/T128388

---

Sorry about the [mega commit](https://github.com/wikimedia/wikipedia-ios/commit/2c774f4ce32e199a9e2e86cbefde071a274714e6), but rebasing & dissecting it now will probably hard w/ the merge commits I've had to do since.  Here's the gist:

- The coup de grace: location accuracy was always 1km, making distance updates so infrequent that we thought it was never updating. This was exacerbated & confounded by the compass also not updating for unrelated reasons, but the fix was straightforward (see location accuracy commit)
- And some minor stuff:
  - Bunch of stuff in ExploreVC & section controller cache
    - Section controllers were being implicitly retained in various blocks in exploreVC, causing them to outlive cache invalidation
  - Schema was being updated even though nothing changed (in `updateContinueReading`)
  - Minor thing: forcing new heading updates after the device orientation changes
  - Table view being reloaded when offscreen sections items arrived

And probably more I'm forgetting right now. Will update later this weekend or first thing Monday.


montehurd:
hmm when i started it there was no nearby in the feed... i guess that's one way to make the problem go away ;)

bgerstle:
😖 i've seen where the nearby section is "invalid" (compass 0 degrees & distance negative) but no entirely missing. that's a new one

montehurd:
ok did clean install and nearby shows up

arrows no longer freeze after device lock and unlock, but distances only updated if i lock and unlock - tested by walking around

Wikipedia/Code/WMFExploreSectionController.h
```diff
@@ -139,12 +139,19 @@ NS_ASSUME_NONNULL_BEGIN
 @optional
 
 /**
- *  Called when a section is about to be displayed
+ *  Called when a section is about to be displayed.
+ *
+ *  @warning 
+ *  This will be called once for each cell that is displayed. It is the controller's responsibility to make sure this
```
coreyfloyd: @bgerstle did you end up needing to do this rather than being explicit?

Wikipedia/Code/WMFExploreSectionControllerCache.m
```diff
@@ -43,22 +50,80 @@ - (instancetype)initWithDataStore:(MWKDataStore*)dataStore {
         self.dataStore                              = dataStore;
         self.sectionControllersBySection            = [[NSCache alloc] init];
         self.sectionControllersBySection.countLimit = [WMFExploreSection totalMaxNumberOfSections];
-        self.sectionsBySectionController            = [NSMapTable mapTableWithKeyOptions:NSMapTableWeakMemory | NSMapTableObjectPointerPersonality
-                                                                            valueOptions:NSMapTableWeakMemory];
+
+        /*
+         Controllers (keys in the map) are retained by the NSCache and can be evicted at any time. On the other hand,
+         sections (values in the map) are "value types" which should be owned by the receiver.  In other words, sections
+         are created as needed, and not retained for any particular reason, and it would be unintentional for a value in
```
coreyfloyd: @bgerstle can you explain why we would want this to retain the sections? This seems counter-intuitive. 

1. The cache is supposed to retain the controllers (and sections using its internal memory rules)
2. The map table only exists as a convenience - its only meant to replace the more verbose reverse look up code that previously existed.
3. I can't think of a reason that we would want the map table to retain a section if it has been released everywhere else.

Wikipedia/Code/WMFExploreSectionControllerCache.m
```diff
@@ -43,22 +50,80 @@ - (instancetype)initWithDataStore:(MWKDataStore*)dataStore {
...
+         this map to disappear because it was dereferenced externally.
+         */
+        self.sectionsBySectionController            =
+            [NSMapTable mapTableWithKeyOptions:NSMapTableWeakMemory | NSMapTableObjectPointerPersonality
+                                  valueOptions:NSMapTableStrongMemory];
     }
     return self;
 }
 
+#if DEBUG
```
coreyfloyd: Duplication. Already doing `#if DEBUG` above in the macro.

Wikipedia/Code/WMFExploreViewController.m
```diff
@@ -157,6 +163,18 @@ - (BOOL)rowAtIndexPathIsOnlyRowVisibleInSection:(NSIndexPath*)indexPath {
...
 
+/**
+ *  Check whether or not a section is going in or out of view.
+ *
+ *  @param indexPath The index path of the row which will or did end displaying.
+ *
+ *  @return @c YES if that section isn't displaying or the given row is the only one visible in its section, otherwise @c NO.
+ */
+- (BOOL)isVisibilityTransitioningForRowIndexPath:(NSIndexPath*)indexPath {
+    return ![self isDisplayingCellsForSection:indexPath.section]
```
coreyfloyd: Returning `YES` when `![self isDisplayingCellsForSection:indexPath.section]` doesn't seem right. Just because a section is off screen shouldn't mean that it is "transitioning"  - it just means it is off screen. Or am I missing something here?

Wikipedia/Code/WMFExploreViewController.m
```diff
@@ -438,7 +484,7 @@ - (nullable UIView*)tableView:(UITableView*)tableView viewForFooterInSection:(NS
 - (void)tableView:(UITableView*)tableView willDisplayCell:(UITableViewCell*)cell forRowAtIndexPath:(NSIndexPath*)indexPath {
     id<WMFExploreSectionController> controller = [self sectionControllerForSectionAtIndex:indexPath.section];
 
-    if ([controller respondsToSelector:@selector(willDisplaySection)] && (![self isDisplayingCellsForSectionController:controller] || [self rowAtIndexPathIsOnlyRowVisibleInSection:indexPath])) {
```
coreyfloyd: @bgerstle i guess this goes with my other comment - why are we removing the check here? 

That would seem to suggestion that the call to `sendWillDisplayToVisibleSectionControllers` when the app returns to the foreground isn't working.

Wikipedia/Code/WMFLocationManager.h
```diff
@@ -17,8 +17,43 @@ NS_ASSUME_NONNULL_BEGIN
...
 @property (nonatomic, strong, readonly) CLHeading* heading;
 
+/**
+ *  Whether or not the receiver is listening for updates to location & heading.
+ *
+ *  @note
+ *  CLLocationmanager doesn't always immediately listen to the request for stopping location updates
+ *  We use this to ignore events after a stop has been requested
+ */
+@property (nonatomic, assign, readonly, getter=isUpdating) BOOL updating;
```
coreyfloyd: Why do we need to expose this to the outside world?

bgerstle: didn't end up needing to in the end. i can remove it

Wikipedia/Code/WMFExploreViewController.m
bgerstle: this method encapsulates the logic we're using in `didEndDisplayingCell` (and previously in `willDisplayCell`). the idea is that the visibility is "transitioning" if either:

1. the section isn't visible (yet, either scrolling on/off screen)
2. the specified row is the only one visible (first row to appear)

Wikipedia/Code/WMFExploreSectionControllerCache.m
bgerstle: sort of, this removes the actual implementation. the macro above ensures it only tries to call it in debug configurations. otherwise the methods would remain even though they're never called.

Wikipedia/Code/WMFExploreSectionControllerCache.m
bgerstle: > I can't think of a reason that we would want the map table to retain a section if it has been released everywhere else.

This is mostly pedantic, since the cache retains (note, **not** copies) the sections as its keys, but I should be able to retrieve a controller for a _copy_ of the section of the section I previously associated it with. IOW, it shouldn't matter that it's a copy of the _original_ key as long as they're equivalent.  On the other hand, I can't think of any cases—pedantic or practical—that I would want sections to be released in the reverse lookup map.  As long as the controller is in the cache, I should be able to perform a reverse lookup of the section it's associated with, **regardless of whether that specific section object was dereferenced**.

Put in Swift terms, think of controllers as classes & sections as structs.  You can't dereference a struct because you can't reference it in the first place.  They're "value types," meaning they only represent a specific value, like a number, not a specific object in memory.

Wikipedia/Code/WMFExploreSectionController.h
bgerstle: sorry, i don't understand the question

Wikipedia/Code/WMFExploreViewController.m
bgerstle: > That would seem to suggestion that the call to sendWillDisplayToVisibleSectionControllers when the app returns to the foreground isn't working.

I'll go back and double check this, but IIRC the specific case was first appearance of nearby section:

- Start the app, `viewDidAppear` on explore, only visible sections sent `willDisplay` (but nearby isn't visible yet)
- Scroll down to nearby
- Placeholder cells come on screen (willDisplay)
- Nearby results are fetched, table is reloaded (didEndDisplaying)
- "real" nearby cells are displayed (willDisplay), **but more than one is already on screen**
- nearby is "broken" (compass at 0deg, distance -3), since `willDisplaySection` is never called

This isn't a problem for other sections, because they're always sent `fetchIfNeeded`.  Now we're doing the same thing for "will display," which fixes the above issue (since it doesn't matter that more than 1 are displaying, we call `willDisplaySection` anyway)

Wikipedia/Code/WMFExploreSectionControllerCache.m
coreyfloyd: Yeah the first is sufficient. There is no need for both removing the calls and the implementations. 

Similar to how piwik and hockey app are both available even if not enabled in all configurations. 

With this duplication, if someone wants to make a change (say to enable it on ad hoc builds), they have to make the same change it in 2 places.

Wikipedia/Code/WMFExploreSectionControllerCache.m
coreyfloyd: >This is mostly pedantic, 

Not a good start :)

> since the cache retains (note, not copies) the sections as its keys, but I should be able to retrieve a controller for a copy of the section of the section I previously associated it with. 

Not following. Retrieving a controller for a copy of a section is done by the cache, not by this object. Did you say this backwards?

> IOW, it shouldn't matter that it's a copy of the original key as long as they're equivalent. 


This seems like a separate issue from retaining. This has to do with the object personality setting, no?


> On the other hand, I can't think of any cases—pedantic or practical—that I would want sections to be released in the reverse lookup map. 

I'm looking at this from the other side. I can't think of a reason that the reverse map should ever retain anything since it is just a convienence and not a cache itself. It should not increment retain counts for anything, that is the responsibility of the cache. It has no need to hold onto sections any longer than they are retained by other objects in the app. 


> As long as the controller is in the cache, I should be able to perform a reverse lookup of the section it's associated with, regardless of whether that specific section object was dereferenced.

Correct - which means as long as the object is held alive by the cache, you can look it up. So why would this object also need to bump the retain count?

> Put in Swift terms, think of controllers as classes & sections as structs. You can't dereference a struct because you can't reference it in the first place. They're "value types," meaning they only represent a specific value, like a number, not a specific object in memory.
  
Not sure this analogy flies or at least I don't follow the implication, we aren't in swift and nothing is a struct.   


If my questions/comments above don't make sense we may need to hangout.

Wikipedia/Code/WMFExploreViewController.m
coreyfloyd: @bgestle true but the name is a lie. 

This is only if the method is called within the scope of didEndDispalyingCell. Called in any other context it may or may not be true.

Wikipedia/Code/WMFExploreSectionControllerCache.m
bgerstle: Basically, comes down to this:

> I'm looking at this from the other side. I can't think of a reason that the reverse map should ever retain anything **since it is just a convienence and not a cache itself.**

I wasn't really taking that to heart.  I'm against "weakly" referencing sections on principle, but in this case it doesn't matter since the weak map is just another "view" of the cache its associated with.  Put another way, the weak map only exists because it's not possible to do reverse lookup of `NSCache` since it doesn't expose its keys.

I'll change this on or before Monday.

---

> Not following. Retrieving a controller for a copy of a section is done by the cache, not by this object. Did you say this backwards?

Right, retrieving a section by its corresponding controller should always be possible even if nothing is referencing it.

> This seems like a separate issue from retaining. This has to do with the object personality setting, no?

Right, the **controllers** are compared using object personality, _not_ the sections.  Meaning, any two equivalent sections can be treated as if they're the same object, since they're immutable.  This is what I mean when I say sections are struct-like.

> we aren't in swift and nothing is a struct.

No, but "value types" are just a concept.  Swift structs are one implementation, doesn't mean we can't have value types in ObjC.

bgerstle: > I wasn't really taking that to heart. I'm against "weakly" referencing sections on principle, but in this case it doesn't matter since the weak map is just another "view" of the cache its associated with.

On second thought, it almost makes more sense to subclass `NSCache` and encapsulate this "reverse lookup" logic. At least for me, that would make this a less contentious.

Wikipedia/Code/WMFExploreViewController.m
bgerstle: prior uses notwithstanding, I think it is true in both contexts, which I tried to explain in the above conditions.  let me try to restate more explicitly:

- section isn't listed in `indexPathsForVisibleCells`
  - `willDisplayCell`: we're about to display it
  - `didEndDisplayingCell`: the section scrolled out of view
- `indexPath.row` is the only one visible in its section
  - `willDisplayCell`: this will be the first row to appear in its section
  - `didEndDisplayingCell`: this was the last cell to scroll away in its section

having said all that, since this method is only used one place i can give it a name more specific to the context in which its currently used.

bgerstle:
@coreyfloyd let me try to recap the things I need to change, and the things we might need to discuss more on monday:

- Amendments
  - [ ] make `WMFLocationManager.updating` private
  - [ ] make "reverse lookup" map weak:weak again, and clarify docs to state it's just a "view" of the `NSCache`, which is source of truth for lifecycle management rules
  - [ ] remove conditional compilation of cache consistency methods (conditional invocation via `WMFVerifyCacheConsistency` macro is sufficient)
- Further discussion/investigation
  - [ ] repeated calls to `willDisplaySection`, aka `sendWillDisplaySection` sufficiency
  - [ ] naming/usage of `isVisibilityTransitioning`

coreyfloyd:
@bgerstle that seems like a good recap of the discussions.

> I'll go back and double check this, but IIRC the specific case was first appearance of nearby section:

> Start the app, viewDidAppear on explore, only visible sections sent willDisplay (but nearby isn't visible yet)
> Scroll down to nearby
> Placeholder cells come on screen (willDisplay)
> Nearby results are fetched, table is reloaded (didEndDisplaying)
> "real" nearby cells are displayed (willDisplay), but more than one is already on screen
> nearby is "broken" (compass at 0deg, distance -3), since willDisplaySection is never called
> This isn't a problem for other sections, because they're always sent fetchIfNeeded. Now we're doing the same thing for "will display," which fixes the above issue (since it doesn't matter that more than 1 are displaying, we call willDisplaySection anyway)

I thought this logic was for returning from the background, not because of the scenario. I guess what you are saying is that because willDisplay is called when placeholders are on screen, the real cells are never updated? Does that sound right?

So would another way of saying this is that the section controller doesn't know whether it is on screen or not? (Do we need to store this state? Probably a bad idea)

Would this problem also be solved if we turned on the updates in the configure cell method? (You might have mentioned this before)

montehurd:
eqmllaz

bgerstle:
> - Further discussion/investigation
>  - [x] repeated calls to `willDisplaySection`, aka `sendWillDisplaySection` sufficiency
>  - [x] naming/usage of `isVisibilityTransitioning`

@coreyfloyd not sure if you'll have a chance to re-review this before it gets merged (trying to get stuff in for TSG), but here's the important things that changed since we last spoke:

- Went back to `willDisplaySection` behavior that's on master. I can't reproduce the issue, so I'm guessing it wasn't the case I tried to describe, and was due to something else.
- @montehurd found [another issue](https://phabricator.wikimedia.org/T128410) which [I fixed](https://github.com/wikimedia/wikipedia-ios/commit/34f433a793619af561e5040c09510c52b8592c67), basically controllers were being wiped out when going to background, leading `shoudlHighlightCell` to return `NO` so you couldn't interact (and the views were no showing data we didn't have), so I just reload the entire table :grimacing: 
- I **didn't** end up changing the `isVisibilityTransitioning` method because i couldn't think of a better name

we can talk about any/all of this when you're back.

PR #532, opened 2016-02-27
Fix (silent) errors when Explore schema saves fail & save copies async

bgerstle:
And make them LOUD again! 💣

TL;DR; the way we were creating file URL's was confusing Foundation, and we kept getting "No such 'file'" errors, because it thought the URL we were saving to was a directory.  Now fixed via super explicitness.

As an add'l boyscout, I changed the asynchronous save implementation to save **copy** instead of the original.  See the commit message for extra details about why the change was made, but suffice it to say that passing **mutable** objects between threads is bad m'kay.


montehurd:
Not seeing the save crashes any more! Yay! :+1:

PR #535, opened 2016-02-27
Fix crash when removing nearby

coreyfloyd:
https://phabricator.wikimedia.org/T128229

When the nearby section is removed, but before the table is updated, we can be in a state where the table delegate methods are still requesting cells/sections/etc… with the old section indexes
This makes sure we don't try to access a section that is beyond the bounds of the section array.

To be more specific - this method:

```
- (void)tableView:(UITableView*)tableView didEndDisplayingCell:(UITableViewCell*)cell forRowAtIndexPath:(NSIndexPath*)indexPath
```

…fires when we reload the table. And if we are scrolled down to the end of the table, then the last section will now be out of bounds if we removed a section. 

So this really is only a situation that exists during the reload process, and only because we try to make a call to a section that is going away in that delegate method.

### DIFF
--- a/Wikipedia/Code/WMFExploreViewController.m
+++ b/Wikipedia/Code/WMFExploreViewController.m
@@ -611,7 +611,7 @@ NS_ASSUME_NONNULL_BEGIN
 }
 
 - (nullable id<WMFExploreSectionController>)sectionControllerForSectionAtIndex:(NSInteger)index {
-    if (index >= self.schemaManager.sections.count) {
+    if(index >= self.schemaManager.sections.count){
         return nil;
     }
     WMFExploreSection* section = self.schemaManager.sections[index];
--- a/Wikipedia/Code/WMFExploreViewController.m
+++ b/Wikipedia/Code/WMFExploreViewController.m
@@ -611,6 +611,9 @@ NS_ASSUME_NONNULL_BEGIN
 }
 
 - (nullable id<WMFExploreSectionController>)sectionControllerForSectionAtIndex:(NSInteger)index {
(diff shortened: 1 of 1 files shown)


bgerstle:
> When the nearby section is removed, but before the table is updated, we can be in a state where the table delegate methods are still requesting cells/sections/etc

How is this possible?  What's triggering those methods?  It sounds to me like controllers whose sections are deleted are still trying to update the table.  If so, this would probably be better addressed by destroying a section controller when its section is explicitly removed from the schema.  This would prevent results from arriving after the section was removed, blocking these sorts of "posthumous" updates.  Also, I would've though our "reverse lookup" early return would prevent controllers whose sections have been removed from altering the table?

> And if we are scrolled down to the end of the table, then the last section will now be out of bounds if we removed a section.

Along the lines of what I said above, it seems to me the problem is that the section _doesn't exist_, and in this case just happens to be out of bounds.  If a related or nearby section is removed with an index that's still in bounds after the schema is updated, we could end up with other weird side effects.

bgerstle:
We discussed this on the phone, and I now realize why the state is inconsistent: `didEndDisplaying` is invoked on reload for all visible cells, then `willDisplay` is invoked for all cells coming back on screen.  The end result is that you can get a mix of unbalanced `didEnd`/`willDisplay` calls when the rows don't match up before/after update.

This means we can potentially run into unintended consequences where various controllers' requests to fetch are cancelled, or have their updates stopped.  I have some ideas about how to to prevent controllers from being affected when cells they don't manage go offscreen, but probably low priority at this point.

PR #542, opened 2016-02-29
Update string and string key from "project" to "language".

montehurd:
https://phabricator.wikimedia.org/T128252

Was confusing, reported by TWN, because project refers to things
like Wiktionary.


bgerstle:
Yes, but it's really a language either (e.g. "simple"). But I agree project can be confusing as well.

PR #544, opened 2016-02-29
Fix account creation crash.

montehurd:
https://phabricator.wikimedia.org/T128412

Was crashing on a masonry constraint assert from the alert pod when fading out the alert. The alert is still being hidden without this line. 

Hmm... since this was an assert not sure if what I fixed was the cause...

### DIFF
--- a/Wikipedia/Code/AccountCreationViewController.m
+++ b/Wikipedia/Code/AccountCreationViewController.m
@@ -258,8 +258,6 @@
 - (void)viewWillDisappear:(BOOL)animated {
     [self enableProgressiveButton:NO];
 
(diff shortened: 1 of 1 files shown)


bgerstle:
@montehurd the bug description said this was only on a particular device, which makes me wonder if other alerts could cause issues. just so happens I have an iPad 3, I can do some testing on it later

coreyfloyd:
> Was crashing on a masonry constraint assert from the alert pod when fading out the alert. The alert it still being hidden without this line.

How is the alert being dismissed now that this line is deleted? 

Also left a note on the main ticket about a stack trace… I assume you are seeing this constraint break in the debugger? Or have you confirmed it in hockey?

PR #545, opened 2016-02-29
Fix set article crash

coreyfloyd:
Was tripping an assertion - this makes sense we were observing the article on init, so of course there is the possibility of getting an update before the view is on screen.

Now we only observe article after the article is set, nice and neat.


bgerstle:
I wish I knew how to repro the scenario, but this seems like it would fix it :+1:

PR #547, opened 2016-03-01
Welcome screen detailed animations.

montehurd:
Note: playground file and contents are redundant. Will figure out what to do with it tomorrow.

untitled mov


montehurd:
@nirzar check it out

AnimationPlayground.playground/Contents.swift
```diff
@@ -0,0 +1,437 @@
...
+liveView.backgroundColor = UIColor.whiteColor()
+XCPlaygroundPage.currentPage.liveView = liveView
+
+
+
+// Express position/widths etc in relation to the view's width (i.e. "lineWidth = w * 0.02" instead of "lineWidth = 3") so all drawing scales proportionally with the animationView's size
+
+
+
+struct WMF_Welcome_Plus {
```
bgerstle: we haven't explicitly decided on a swift coding style yet, but we should at least try to follow Apple & community conventions:

- No prefixes
- UpperCase classes, camelCase vars, fields, & methods

AnimationPlayground.playground/Contents.swift
```diff
@@ -0,0 +1,437 @@
...
+    var frame: CGRect
+    var color: UIColor
+    var cornerRadius: CGFloat
+    var opacity: Double
+    var initialTransform: CATransform3D
+    var finalTransform: CATransform3D
+}
+
+extension UIBezierPath {
+    class func wmf_circlePathWithRadius(radius: CGFloat) -> UIBezierPath {
```
bgerstle: `class func` is one way, convenience initializer would've also worked:

``` swift
UIBezierPath(circlePathWithRadius radius: CGFloat) {
  self.init(arcCenter: ... )
}
```

AnimationPlayground.playground/Contents.swift
```diff
@@ -0,0 +1,437 @@
...
+extension CAShapeLayer {
+    func wmf_applyCommonSettingsForSize(size: CGSize, position: CGPoint) -> Void {
+        self.lineCap = "round"
+        self.opacity = 0.0
+        self.lineWidth = size.width * 0.014
+        self.strokeEnd = 1.0
+        self.position = CGPointMake(size.width * position.x, size.height * position.y)
+        self.zPosition = 100
+    }
+    class func wmf_addLine(line: WMF_Welcome_Line, toView: UIView) -> CAShapeLayer {
```
bgerstle: why is this a static function on `CAShapeLayer`? IMO would make more sense to extend `UIView` to `addWelcomeLine(line: WelcomeLine)`

AnimationPlayground.playground/Contents.swift
```diff
@@ -0,0 +1,437 @@
...
+        anim.fillMode = kCAFillModeForwards
+        anim.removedOnCompletion = false
+        anim.toValue = opacity
+        anim.timingFunction = CAMediaTimingFunction(name: kCAMediaTimingFunctionEaseInEaseOut)
+        return anim;
+    }
+}
+
+extension UIView{
+    func wmf_addCircles(circles: WMF_Welcome_Circle...) {
```
bgerstle: nice use of variadics

bgerstle: but... all of these methods are copy/pasted. c'mon man, use your swift!

``` swift
extension UIView {
  func addWelcomeElement<E: WelcomeElement where E: CAShapeLayerConvertible>(element: E) {
    // refactor all element-specific (shape?) layer creation here, including animations
    let shapeLayer = CAShapeLayer(e)
    // apply common settings and scale unit-rect frame to receiver's frame?
    shapeLayer.applyCommonSettingsForSize(frame.size, position: element.position)
    layer.addSublayer(shapeLayer)
}
```

bgerstle: anyway, not sure if we have to time to DRY up all of this, but would be a fun project for you to work on

bgerstle:
again, nice work! very clever to use unit rects so everything scales across screen sizes :100:

coreyfloyd:
@nirzar noticed this on the earlier welcome screen pr, but didn't get a chance to comment…

But these animations are really, really slow. 

They shouldn't be super quick, but animations typically are never longer than 0.75 s and usually 0.5 or less… I have personally never seen any appearance animations (outside of a game) go longer than 1 second. 

We are currently at 1.3 seconds. Its pretty painful and users will get impatient seeing this 3 times in a row.

Going slow can make the app feel sluggish and look like we are showing off. Even Apple got caught doing this and dialed back in iOS 7 after many complaints from designers and developers: 
https://marco.org/2013/08/27/along-for-the-ride (this is just one post of many…)

We should clip these down to a reasonable duration with this pr.

montehurd:
@coreyfloyd we haven't tweaked the animation timings/duration yet. Later this morning :)

montehurd:
Hence the WIP

coreyfloyd:
@montehurd going to :+1:  this so we can get the build out - but have some comments to address for tomorrow.

You'll start to see some comments coming in…

Wikipedia/Code/WelcomeAnimationExtensions.swift
```diff
@@ -0,0 +1,753 @@
+import Foundation
```
coreyfloyd: This file is really big - and contains some distinct extensions / objects - so we should break this up

montehurd:
@coreyfloyd Ya Brian and I are going to pair up if we have time before he leaves to DRY things up.

Wikipedia/Code/WelcomeAnimationExtensions.swift
```diff
@@ -0,0 +1,753 @@
...
+}
+
+struct WelcomeBar {
+    var frame: CGRect
+    var color: UIColor
+    var cornerRadius: CGFloat
+    var animation: WelcomeAnimation
+}
+
+struct WelcomeAnimation {
```
coreyfloyd: This struct is essentially just the animation parameters… so we can just create an animation with these values instead. 

Structs are nice, but if you find yourself doing a 1-1 mapping between a class's properties, its best just to just use the class.

Wikipedia/Code/WelcomeAnimationExtensions.swift
montehurd: For sure. Hugeness is artifact of proofing in playground and lack of time :(

Wikipedia/Code/WelcomeAnimationExtensions.swift
```diff
@@ -0,0 +1,753 @@
...
+    }
+    class func wmf_plusPathWithWidth(width: CGFloat) -> UIBezierPath {
+        let path = UIBezierPath.wmf_horizontalLinePathWithWidth(width)
+        path.appendPath(UIBezierPath.wmf_verticalLinePathWithHeight(width))
+        return path
+    }
+}
+
+extension CAShapeLayer {
+    func wmf_applyCommonSettingsForSize(size: CGSize, position: CGPoint, initialOpacity: Double) -> Void {
```
coreyfloyd: "applyCommonSettings" is pretty nondescript

Wikipedia/Code/WelcomeAnimationExtensions.swift
montehurd: Ah good point.

Wikipedia/Code/WelcomeAnimationExtensions.swift
```diff
@@ -0,0 +1,753 @@
...
+        self.lineCap = "round"
+        self.opacity = Float(initialOpacity)
+        self.lineWidth = size.width * 0.014
+        self.strokeEnd = 1.0
+        self.position = CGPointMake(size.width * position.x, size.height * position.y)
+        self.zPosition = 100
+    }
+    class func wmf_addLine(line: WelcomeLine, toView: UIView) -> CAShapeLayer {
```
coreyfloyd: This seems to be more than adding a line. Also its returning a shape layer… so this seems more like a constructor `shapeLayerWithLine…` or something to that affect

Wikipedia/Code/WelcomeAnimationExtensions.swift
```diff
@@ -0,0 +1,753 @@
...
+        let shape = CAShapeLayer()
+        shape.path = UIBezierPath.wmf_horizontalLinePathWithWidth(toView.frame.size.width * line.width).CGPath
+        shape.strokeColor = UIColor.blackColor().CGColor
+        shape.transform = line.animation.initialTransform
+        shape.wmf_applyCommonSettingsForSize(toView.frame.size, position: line.position, initialOpacity: line.animation.initialOpacity)
+        toView.layer.addSublayer(shape)
+        return shape;
+    }
+    class func wmf_addDashedCircle(circle: WelcomeCircle, toView: UIView) -> CAShapeLayer {
```
coreyfloyd: same here and for the rest of these class constructors.

Wikipedia/Code/WelcomeAnimationExtensions.swift
```diff
@@ -0,0 +1,753 @@
...
+        shape.path = UIBezierPath.wmf_circlePathWithRadius(toView.frame.size.width * circle.radius).CGPath
+        shape.fillColor = UIColor.clearColor().CGColor
+        shape.strokeColor = UIColor.blackColor().CGColor
+        shape.lineDashPattern = [toView.frame.size.width * 0.029, toView.frame.size.width * 0.047]
+        shape.transform = circle.animation.initialTransform
+        shape.wmf_applyCommonSettingsForSize(toView.frame.size, position: circle.position, initialOpacity: circle.animation.initialOpacity)
+        toView.layer.addSublayer(shape)
+        return shape;
+    }
+    class func wmf_addCircle(circle: WelcomeCircle, toView: UIView) -> CAShapeLayer {
```
coreyfloyd: Also passing the "owner" of the shape layer is a bit of the inversion. You should return the shape layer so it can be added to a view.

Wikipedia/Code/WelcomeAnimationExtensions.swift
```diff
@@ -0,0 +1,753 @@
...
+            (toView.frame.size.width * bar.frame.origin.x),
+            (toView.frame.size.height * bar.frame.origin.y)
+        )
+        toView.layer.addSublayer(shape)
+        return shape;
+    }
+}
+
+extension CALayer{
+    func wmf_addWelcomeAnimation(animation: WelcomeAnimation) {
```
coreyfloyd: not very descriptive. Try to use names that describes what is happening - as opposed to `welcome…`

Wikipedia/Code/WelcomeAnimationExtensions.swift
montehurd: Good points. Brian said something similar.

Wikipedia/Code/WelcomeAnimationExtensions.swift
```diff
@@ -0,0 +1,753 @@
...
+        self.transform = animation.initialTransform;
+        self.addAnimation(CABasicAnimation.wmf_animationToXF(animation.finalTransform, delay: animation.delay, duration: animation.duration), forKey: nil)
+        
+        self.opacity = Float(animation.initialOpacity);
+        self.addAnimation(CABasicAnimation.wmf_animationToOpacity(animation.finalOpacity, delay: animation.delay, duration: animation.duration), forKey: nil)
+    }
+}
+
+extension CABasicAnimation{
+    class func wmf_animationToXF(transform: CATransform3D, delay: Double, duration: Double) -> CABasicAnimation {
```
coreyfloyd: Not sure what XF is… remember spell it out.

Also like above, you are not adding an animation… you are building one and returning it. (a constructor)

Wikipedia/Code/WelcomeAnimationExtensions.swift
montehurd: Did so to simplify unit calculation (so it's resolution independent). Can do this in a less implicit way though for sure!

Wikipedia/Code/WelcomeAnimationExtensions.swift
montehurd: Transform. My bad :)

Wikipedia/Code/WelcomeAnimationExtensions.swift
```diff
@@ -0,0 +1,753 @@
...
+        anim.toValue = opacity
+        anim.timingFunction = CAMediaTimingFunction(name: kCAMediaTimingFunctionEaseInEaseOut)
+        return anim;
+    }
+}
+
+extension UIView{
+    func wmf_addCircles(circles: WelcomeCircle...) {
+        for circle: WelcomeCircle in circles {
+            let circleShape = circle.isDashed ? CAShapeLayer.wmf_addDashedCircle(circle, toView: self) : CAShapeLayer.wmf_addCircle(circle, toView:self)
```
coreyfloyd: A bit inside out… you made objects to encapsulate the logic for a circle, but then here you need to ask the circle a question and then decide how to draw it.

The circle should know how to draw itself. 

https://pragprog.com/articles/tell-dont-ask

montehurd: Good point!

Wikipedia/Code/WelcomeAnimationExtensions.swift
```diff
@@ -0,0 +1,753 @@
...
+        anim.beginTime = CACurrentMediaTime() + delay
+        anim.fillMode = kCAFillModeForwards
+        anim.removedOnCompletion = false
+        anim.toValue = opacity
+        anim.timingFunction = CAMediaTimingFunction(name: kCAMediaTimingFunctionEaseInEaseOut)
+        return anim;
+    }
+}
+
+extension UIView{
```
coreyfloyd: Should every view know how to draw welcome circles, pluses, etc?

Wikipedia/Code/WelcomeAnimationExtensions.swift
```diff
@@ -0,0 +1,753 @@
...
+            imageView.image = UIImage(named: image.name)
+            imageView.contentMode = UIViewContentMode.ScaleAspectFit;
+            imageView.layer.zPosition = image.zPosition
+            imageView.layer.wmf_addWelcomeAnimation(image.animation);
+            self.addSubview(imageView)
+            return imageView
+        })
+    }
+    
+    func wmf_configureForIntroAnimation(){
```
coreyfloyd: same thing here… this seems like very specific for one view in the app (not just one class, but only one instance)

Wikipedia/Code/WelcomeAnimationExtensions.swift
```diff
@@ -0,0 +1,753 @@
...
+        CATransaction.begin()
+        
+        let finalDashedCircleOpacity = 0.15
+        let finalCircleOpacity = 0.09
+        let finalPlusOpacity = 0.15
+        let finalLinesOpacity = 0.15
+        let scaleZeroXF = CATransform3DMakeScale(0, 0, 1)
```
coreyfloyd: scaleZeroXF - same question - what is an XF?

Wikipedia/Code/WelcomeAnimationExtensions.swift
```diff
@@ -0,0 +1,753 @@
...
+        let rightXF = CATransform3DMakeTranslation(50, 0, 0)
```
coreyfloyd: Lots os objects / structs / methods above for creating all the items, but we still need additional configuration numbers here.

All of this should be encapsulated. Basically we are mixing levels of abstraction.

At the highest level, you should only interact with fully configured objects. In pseudo code it might look something like this:

```
view = someCircle()
animation = someAnimation()
self.view.addSubview(view)
view.startAnimation(animation)

coreyfloyd:
@montehurd ok - left the basic comments… we can discus more tomorrow.

PR #548, opened 2016-03-01
Fix crash in sharing on iPad

coreyfloyd:
https://phabricator.wikimedia.org/T128430

in iOS 8 and above you need to configure the popover presentation controller

Additionally, removed some nullable attributes that were not needed.


bgerstle:
:+1: looks good, checked on my iPad 3 running iOS 8 (toolbar & share-a-fact)

PR #555, opened 2016-03-01
Fix bugs refreshing explore when returning from airplane mode

coreyfloyd:
https://phabricator.wikimedia.org/T128427

In addition to the crash, I found and fixed several other issues when coming back from airplane mode:
1.The no results view flickers over and over again
2. (After fixing 1) No results appear for items which should have results after airplane mode is disabled 
3. Taping "Try again" on the no results cell does not reload the section
4. (After fixing 3) Taping “try again”, can reload the wrong section
5. (After Fixing 1) Pulling to refresh does not reload sections with errors

Unfortunately, I found another issue which I will be correcting in a separate ticket:
After fixing all of the above, upon reconnect, the first few requests fail due to SSL errors. (However the user can manually refresh using the try again button or by pulling to refresh)


bgerstle:
:+1: looks good, nice catches

PR #557, opened 2016-03-01
Fix auto-layout crash from alert dismissal

bgerstle:
See the new [commit in our TSMessages fork](https://github.com/wikimedia/TSMessages/commit/ec0aa9be29ab79083e4957ac500f3af2f622605d), but TL;DR; don't try to constrain a view to a superview which doesn't exist.  
login spam demo mov


coreyfloyd:
nice! :+1:

PR #560, opened 2016-03-02
Fix Potd crash

coreyfloyd:
inserting nil - doh!

https://phabricator.wikimedia.org/T128554


montehurd:
SHIPIT!

bgerstle:
@coreyfloyd won't this cause the feed to contain empty sections in this case? it might be better to catch this in the response serializer/fetcher and treat as an error so we can display the proper empty/error view in the feed and retry at the appropriate time.

bgerstle:
also not seeing this crash on master, maybe resolved on the back-end somehow?

bgerstle:
are you _sure_ the response was `nil`? i just ran into a case where the section itself was `nil` (in this case , was main page because it had been deallocated)

coreyfloyd:
@bgerstle the crash log says it is nil. 

Regardless, we shouldn't be putting a possible nil value into an array without checking it.

As far as treating it as an error - yeah we can do that…

bgerstle:
:+1:

PR #561, opened 2016-03-02
Make sure language buttons stay left aligned

coreyfloyd:


bgerstle:
:+1:

bgerstle:
@montehurd can you double check this?

montehurd:
Seeing ellipses issue...
untitled mov

montehurd:
Still seeing it. Repro steps:

untitled mov

coreyfloyd:
screenshot 2016-03-02 15 10 05


@montehurd Not sure why you are still seeing it - I am not. Regardless, at this point you can file a ticket if its still an issue. Shouldn't be a blocker. Especially since it isn't all the time.

montehurd:
@coreyfloyd can you try the exact repro steps in my gif? (Same langs in same order) Curious if it's just my simulator.

coreyfloyd:
@montehurd yeah I did - not sure why its being fussy with you. I also tried clicking about 30 other ones and never get that issue anymore. 

It was pretty clear why you saw the issue before I made the fix… but I am at a loss at why you would see it now. You might want to clear out some stuff or try other devices. But otherwise like I said we should file a ticket for 5.0.1.

montehurd:
:+1: ya further tweaks can be follow-ons.

PR #562, opened 2016-03-02
Always monitor location when nearby cell appears

bgerstle:
Phab: [T128608](https://phabricator.wikimedia.org/T128608)

See testing notes in Phab ticket :point_up:

---

This was originally part of the original fix (#530), but I removed it when I couldn't reproduce the issue.  I've now found additional conditions which make the issue more reproducible, but still not perfectly consistent.  With this change, I can't get compasses to freeze after several tries.

@coreyfloyd we had discussed issues w/ sections "never" fetching to do cancellation of fetches when the last cell is ends displaying, but I can't reproduce the issue and don't have concrete evidence that it's actually a problem. I'll continue trying and if I run into again and I have logs to show the sections never would've been fetched, I'll file an issue.

### DIFF
--- a/Wikipedia/Code/WMFExploreSectionController.h
+++ b/Wikipedia/Code/WMFExploreSectionController.h
@@ -140,6 +140,12 @@ NS_ASSUME_NONNULL_BEGIN
 
 /**
  *  Called when a section is about to be displayed.
+ *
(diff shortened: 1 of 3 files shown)


coreyfloyd:
@bgerstle ok - as long as you can't reproduce that makes sense. 

Otherwise this looks good.

I might consider converting `willDisplaySection` to `willDisplayCellAtIndexPath` at some point in the future. That way it maps better to the table API.

:+1:

PR #565, opened 2016-03-02
Welcome tweaks

montehurd:
Per Nirzar
- [x] adjust telescope animation to not start so low and to finish a little higher
- [x] slightly increase fade-up duration

### DIFF
--- a/Wikipedia/Code/WelcomeAnimationExtensions.swift
+++ b/Wikipedia/Code/WelcomeAnimationExtensions.swift
@@ -207,6 +207,10 @@ extension UIView{
             return imageView
         })
     }
+
+    func wmf_radiansFromDegrees(degrees: CGFloat) -> CGFloat{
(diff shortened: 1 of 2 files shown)


bgerstle:
:+1:

Wikipedia/Code/WelcomeAnimationExtensions.swift
```diff
@@ -207,6 +207,10 @@ extension UIView{
             return imageView
         })
     }
+
+    func wmf_radiansFromDegrees(degrees: CGFloat) -> CGFloat{
```
bgerstle: duplication, make a C function in WMFGeometry so you can call it from Swift. we should get rid of the macro anyway

montehurd: Ya if it's ok I'll do this when I address the other Welcome CR comments.

PR #568, opened 2016-03-02
Don't show non-HTTP URLs in safari VC

bgerstle:
Would have refactored & clarified that if statement, but erred on the side of minimal changes for now.


coreyfloyd:
:+1:

montehurd:
:+1:

PR #571, opened 2016-03-03
prevent nil obj in placeholder array crash, add lang/site fallbacks

bgerstle:
not sure how site was nil or not contained in language links, but we should at least prevent crashes in production

see: https://rink.hockeyapp.net/manage/apps/152731/app_versions/35/crash_reasons/113233708


coreyfloyd:
:+1:

PR #575, opened 2016-03-05
Cleanup of welcome animation bits

montehurd:
- [x] custom UIView objects for each of the 3 welcome animations
- [x] better division between when the animation views are setup and when their animations are kicked off
- [x] file for extensions 
- [x] file for custom welcome CAShapeLayers
- [x] better encapsulation
- [x] animations still scale to any resolution
- [x] added small tweak to telescope animation per Nirzar (slightly over-tilts then tilts back down a bit)

untitled mov


montehurd:
@nirzar ^ added the over-tilt and correction :)

Wikipedia/Code/WelcomeAnalyticsAnimationView.swift
```diff
@@ -0,0 +1,239 @@
...
+        
+        self.solidCircle = WelcomeCircleShapeLayer(
+            unitRadius: 0.235,
+            unitOrigin: CGPointMake(0.654, 0.41),
+            referenceSize: self.frame.size,
+            isDashed: false,
+            transform: scaleZeroTransform,
+            opacity:0.0
+        )
+        self.layer.addSublayer(self.solidCircle!)
```
coreyfloyd: We should try to think of ways to get rid of the `!` when we can. Just because that is dangerous in swift land. 

One idea is to not make `self.solidCircle` an optional - you could do that by making it lazy so it always returns a value.

coreyfloyd: (this applies to all the layers / views)

Wikipedia/Code/WelcomeAnimationExtensions.swift
```diff
@@ -1,155 +1,54 @@
...
-    var animation: WelcomeAnimation
-}
-
-struct WelcomeBar {
-    var frame: CGRect
-    var color: UIColor
-    var cornerRadius: CGFloat
-    var animation: WelcomeAnimation
+extension CGFloat {
+    func wmf_denormalizeUsingReference (reference: CGFloat) -> CGFloat {
```
coreyfloyd: I think I know whats going on here… but not quite sure. Just hit me up after you get in.

coreyfloyd:
@montehurd this all looks really good.

The only thing I really would like to address is better handling of optionals (or eliminating them if possible) - but otherwise this is great.

montehurd:
Closing to re-push to `develop` branch.

PR #576, opened 2016-03-06
Only process coverage for `Wikipedia*` reports

stevepeak:
> CC @bgerstle @coreyfloyd 

Thank you for helping out with this! :+1:


coreyfloyd:
@stevepeak awesome thanks again!

PR #577, opened 2016-03-06
Fix for "Unbalanced calls to begin/end appearance transitions" warning.

montehurd:
The warning is appearing when the FTUX is shown.

One of the comments [here](http://stackoverflow.com/a/7886348) suggest kicking things off in viewDidAppear... Maybe that's a better fix?


coreyfloyd:
@montehurd hmmmm interesting - yeah viewdidload appear is probably the ticket. We moved some stuff around in here a few weeks a go to simplify the flow - I don't think we noticed the unbalanced calls

montehurd:
@coreyfloyd do you have a handle on this / want to take it from here?

PR #583, opened 2016-03-08
TWN sync with some new tests. Adds Colognian localization.

montehurd:
- [x] added test for `{{...}}` substitution in TWN strings (for plurals/gender/etc syntax, which are presently unsupported in the iOS app)
- [x] added tests for html in TWN strings (app is native)
- [x] removed brackets from `cs.lproj/Localizable.strings` and `pl.lproj/Localizable.strings` so new test passes
- [x] removed html and brackets from `ksh.lproj/Localizable.strings` so new tests pass
- [x] added Colognian localization to the project so the tests actually run against `ksh.lproj/Localizable.strings` (and Colognian users can actually see the translations!)
- [x] confirmed Arabic `$1` substitution string is working (per @niedzielski)


montehurd:
@niedzielski heya!

I added some tests to catch html and plural brackets and did some manual cleanup (mostly to the Colognian "ksh" file) so the new tests pass.

How do you notify TWN when a translation set has issues?

niedzielski:
@montehurd, sweet! Hm, I think there's a way to contact a specific translator but I don't know it. Maybe @berndsi does? If not, let's ping Siebrand or Niklas for help.

berndsi:
@montehurd Check out https://www.mediawiki.org/wiki/Translation_of_app_string_resources#Fixing_translations_on_TWN_site. If there are still questions let me know.

PR #584, opened 2016-03-08
Cleanup of welcome animation bits

montehurd:
- [x] custom UIView objects for each of the 3 welcome animations
- [x] better division between when the animation views are setup and when their animations are kicked off
- [x] file for extensions 
- [x] file for custom welcome CAShapeLayers
- [x] better encapsulation
- [x] animations still scale to any resolution
- [x] added small tweak to telescope animation per Nirzar (slightly over-tilts then tilts back down a bit)

untitled mov


montehurd:
@coreyfloyd I closed the one on `master` - this one's on `develop`.

PR #593, opened 2016-03-10
Use display title, if available, in web view.

montehurd:
https://phabricator.wikimedia.org/T65971

**Before**
screen shot 2016-03-10 at 9 38 53 am

**After**
screen shot 2016-03-10 at 9 37 48 am

**Before**
screen shot 2016-03-10 at 9 38 46 am

**After**
screen shot 2016-03-10 at 9 38 19 am


coreyfloyd:
NICE!

PR #594, opened 2016-03-10
Fix crash for main page

coreyfloyd:
This was caused by the same bug we fixed before… either self is gone or the object is nil -  we missed this one.

https://phabricator.wikimedia.org/T129494

### DIFF
--- a/Wikipedia/Code/WMFMainPageSectionController.m
+++ b/Wikipedia/Code/WMFMainPageSectionController.m
@@ -129,8 +129,11 @@ static NSString* const WMFMainPageSectionIdentifier = @"WMFMainPageSectionIdenti
     @weakify(self);
     return [self.siteInfoFetcher fetchSiteInfoForSite:self.site].then(^(MWKSiteInfo* data) {
         @strongify(self);
+        if (!self || !data.mainPageTitle) {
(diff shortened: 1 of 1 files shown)


Wikipedia/Code/WMFMainPageSectionController.m
```diff
@@ -129,8 +129,11 @@ - (AnyPromise*)fetchData {
     @weakify(self);
     return [self.siteInfoFetcher fetchSiteInfoForSite:self.site].then(^(MWKSiteInfo* data) {
         @strongify(self);
+        if (!self || !data.mainPageTitle) {
+            return (id)[AnyPromise promiseWithValue:[NSError cancelledError]];
+        }
         self.siteInfo = data;
-        return [self.titleSearchFetcher fetchArticlePreviewResultsForTitles:@[self.siteInfo.mainPageTitle] site:self.site];
+        return (id)[self.titleSearchFetcher fetchArticlePreviewResultsForTitles:@[self.siteInfo.mainPageTitle] site:self.site];
```
montehurd: Are `self.siteInfo.mainPageTitle` and `data.mainPageTitle` always the same?

coreyfloyd: @montehurd see line 135 - we set `self.siteInfo = data`

PR #604, opened 2016-03-11
Hide special pages in Top read article list.

montehurd:
https://phabricator.wikimedia.org/T127890

**Before:**
screen shot 2016-03-10 at 5 26 07 pm

**After:**
screen shot 2016-03-10 at 5 23 34 pm


coreyfloyd:
also nice!

PR #612, opened 2016-03-12
Text size adjustment

coreyfloyd:
Put this together on my flight. Needs an icon and some discussion but not to far from shippable.

text size


LeoNatan:
You should be using iOS's dynamic type for this.

PR #614, opened 2016-03-14
Fix for date field crash.

montehurd:
https://phabricator.wikimedia.org/T129912

https://rink.hockeyapp.net/manage/apps/152731/app_versions/40/crash_reasons/114070711


coreyfloyd:
@montehurd do you know why the field is missing and are their any repercussions to just setting an arbitrary date?

Also should we just make the date optional? Is this just for showing the last modified date in the UI?

montehurd:
@coreyfloyd lemme check a bit more...

coreyfloyd:
👌

coreyfloyd:
@montehurd couple questions:

1. What about displaying something in the last updated field of the article footer when the date is nil?
2. Is the hash succeeding when the date is nil?

montehurd:
@coreyfloyd 
- hash seems fine 'cause it when it's nil the bitwise XOR sees it as 0 and thus the bitwise XOR doesn't change anything 
- checking the footer now...

montehurd:
@coreyfloyd hmm yeah if the date is nil the footer says it was last edited 9 billion trillion zillion days ago... lemme fix that :)

montehurd:
In rare case where the last modified date will be nil, per josh just show last modified `0 days ago` for now.

**Before:**
screen shot 2016-03-23 at 4 49 12 pm

**After:**
screen shot 2016-03-23 at 4 49 36 pm

coreyfloyd:
:+1: will merge when done

PR #618, opened 2016-03-20
Visual History

coreyfloyd:
@jtminor told me about this component last week that @nirzar had showed him a while back.

It was rather easy to integrate, so I thought I would put it up for us to look at.

Looks compelling to me.

history


montehurd:
now that's just cool!

PR #619, opened 2016-03-24
Fix "read more" flakiness.

montehurd:
https://phabricator.wikimedia.org/T130794
(took a stab at this as an "end-of-day" bite-sized ticket)

Seems to fix the bug. Unsure about exact cause...

**Before:** (on second view article seemed to always be missing the read more items)
before

**After:** (confirmed the read more results *are* being retrieved, so thought it was some synchronicity issue - this PR's added delay fixing the bug seems to confirm this?)
after mov


coreyfloyd:
@montehurd is it possible to check the to see if the web view isn't loaded yet? Is that possibly the issues?
Maybe you could wait to call `fetchReadMoreIfNeeded` until after the article is loaded in the web view?

coreyfloyd:
:+1:

PR #620, opened 2016-03-24
Fixing crash in image protocol again

coreyfloyd:
https://phabricator.wikimedia.org/T129522

Ok - well the first fix was a guess… and it was wrong.

This time, I did a bit more digging. 

This fix has 2 changes:
1. **Process images on main queue** - This shouldn't matter, but its one of the few places we turn processing over to a background queue. And I don't see it any benefit - There is no slow down in the UI or any noticeable performance loss. In fact because of change 2, we might be even loading articles faster…
2. **Update AFNetworking and move to NSURLSession APIs** - This was hinted at by the threads in the ticket. Some people with a similar crash were only seeing it in NSURLConnection APIs. So we are moving to the NSURLSession APIs to remove that as a potential issue. As a side benifit… articles appear to be loading faster for me… but this is anecdotal. Would like to hear what others see.

Because we are updating the underlying networking of the app, this has some risk. I may suggest to Josh that we run a regression early to see if TSG comes up with any problems.


montehurd:
I tested feed, search, login, editing, acct creation etc... all worked fine.

There's one place we could do @weakify / @strongify (MWKTitleLanguageController.m around line 49) but we can do as follow-on.

PR #621, opened 2016-03-24
Fix warnings after upgrading Xcode 7.3

coreyfloyd:
Whats on the tin. Lets see if we need to update travis as well…


coreyfloyd:
@montehurd this needs merged into development first and it will fix Travis for all the other patches.

PR #623, opened 2016-03-28
Fix universal links not opening on launch

coreyfloyd:
Was caused by a race condition… see commit messages and code comments for more.

Also fixes last read not restoring consistently.


montehurd:
Looks like the "Search" shortcut isn't working on cold start...

PR #629, opened 2016-03-29
Fix cold start app icon search shortcut.

montehurd:
https://phabricator.wikimedia.org/T131137

Note to self: Corey said to try the same approach he used with universal links fix (something similar to the "unprocessedUserActivity" iirc) - will try that tomorrow morning instead of this delay :)

@coreyfloyd ^ feel free to push a commit in the morning if you want - otherwise I'll follow up :)


coreyfloyd:
Fixed here: #631

PR #630, opened 2016-03-29
handle null case for article history delta

ndis1:
[Ticket: T127373 article history item null delta](https://phabricator.wikimedia.org/T127373)
article history delta is calculated between two items
we can't know the delta of the last item unless it is the first item ever created
this bumps the number of items requested and doesn't show the final item received
unless it is the first in the article history

If someone could point me to the preferred pattern for adding tests to these fetchers I'd be happy to add some or update existing.

### DIFF
--- a/Wikipedia/Code/PageHistoryFetcher.m
+++ b/Wikipedia/Code/PageHistoryFetcher.m
@@ -73,7 +73,7 @@
         @"action": @"query",
         @"prop": @"revisions",
         @"rvprop": @"ids|timestamp|user|size|parsedcomment",
-        @"rvlimit": @50,
+        @"rvlimit": @51,
         @"rvdir": @"older",
         @"titles": title.text,
         @"continue": @"",
@@ -124,6 +124,7 @@
 
             [self calculateCharacterDeltasForRevisions:revisionsByDaySorted
                                        fromParentSizes:parentSizes];
+            [self pruneIncompleteRevisionFromRevisions:revisionsByDaySorted];
 
             output = revisionsByDaySorted;
         }
@@ -148,6 +149,14 @@
                 }
             }
         }
+    }
+}
+
+- (void)pruneIncompleteRevisionFromRevisions:(NSMutableArray*)revisions {
+    //Unless the last item is the oldest, remove it since we don't know its characterDelta
+    NSNumber *parentId = revisions.lastObject[@"parentId"];
+    if (parentId == nil || parentId.integerValue != 0) {
+        [revisions removeLastObject];
     }
 }
 
--- a/Wikipedia/Code/PageHistoryFetcher.m
+++ b/Wikipedia/Code/PageHistoryFetcher.m
@@ -153,10 +153,16 @@
 }
 
 - (void)pruneIncompleteRevisionFromRevisions:(NSMutableArray*)revisions {
-    //Unless the last item is the oldest, remove it since we don't know its characterDelta
-    NSNumber *parentId = revisions.lastObject[@"parentId"];
-    if (parentId == nil || parentId.integerValue != 0) {
-        [revisions removeLastObject];
+    NSDictionary *lastDayRevisionsMeta = revisions.lastObject;
+    NSMutableArray *lastDayRevisions = lastDayRevisionsMeta[@"revisions"];
+    NSNumber *lastRevisionParentId = lastDayRevisions.lastObject[@"parentid"];
+
+    if (lastRevisionParentId == nil || lastRevisionParentId.integerValue != 0) {
+        if (lastDayRevisions.count <= 1) {
+            [revisions removeLastObject];
+        } else {
+            [lastDayRevisions removeLastObject];
+        }
     }
 }
 


Wikipedia/Code/PageHistoryFetcher.m
```diff
@@ -73,7 +73,7 @@ - (NSDictionary*)getParamsForTitle:(MWKTitle*)title {
         @"action": @"query",
         @"prop": @"revisions",
         @"rvprop": @"ids|timestamp|user|size|parsedcomment",
-        @"rvlimit": @50,
+        @"rvlimit": @51,
```
coreyfloyd: @ndis1 can you explain why this is now 51 instead of 50 and how this fixes the issue?

coreyfloyd:
@ndis1 hey I get the pruning part. I'm not sure how the number of revisions plays into it.

Also does this satisfy the comment on the ticket:

> Acceptance criteria:
> Given that I am viewing the article history
> When the server returns (null) or empty values for change counts
> Then I should see "0" in the article history view

I didn't see any code explicitly setting the number in the Article History UI to "0" if the delta can't be calculated. Or does this change now make that impossible now?

coreyfloyd:
As far as testing…

best thing is to open quickly "fetchertests" and you will see some examples. 
We are not religious about the patterns for unit tests (just patterns in the app code itself). 

We use a couple different libraries, there is straight XCTest, you will see some OCHamcrest (personally I hate this), and some Quick & Nimble. We also have visual tests using FBSnapshotTestCase. Use what makes sense to you.

ndis1:
@coreyfloyd Since we're throwing out the last item, we need to get one more item in order to match the 50 items the app is currently showing. If we throw out the last item but still only get 50 items, we will be showing one fewer item. My thinking was it would be nice to keep that feature exactly as it was from the user's point of view, also I thought that a change in the number of items shown would be confusing for QA because the last item they are looking to inspect would have disappeared.

The requirement around showing zero is, I think, based on an incomplete understanding of the bug. The bug was not due to an isolated data point being missing, but rather to a result of the way we are calculating the diffs on the client side, as opposed to receiving them in the response. We could add logic to show `0` in the case that we are missing one of the fields we are using to calculate the diff, but that change would be defensive, rather than addressing a bug that has been noticed. I'm happy to add that logic, and default values for any other fields that may be missing, if it is seen as being valueable

coreyfloyd:
@ndis1 that makes sense to me.

Hey do you have Phabricator account? So I can assign the ticket to you?
https://www.mediawiki.org/wiki/Phabricator/Help#Creating_your_account

@montehurd looks good to me :+1:  - you want to check out as well?

coreyfloyd:
@ndis1 also can you merge the latest from develop into your branch?

montehurd:
I'll test it - overall sounds legit for now, although I think the long-term solution would be to use query-continue and request the previous 50 changes when the user scrolls to the bottom of the first set of 50 changes (I think we're just requesting the 50 or so most recent changes at the moment).

coreyfloyd:
@montehurd sounds good - we should file a ticket for that if we haven't already. Thanks @ndis1 we will get this in when tests finish

montehurd:
:-1: 

I found a small bug - if there are fewer than 50 revisions this PR chops off the initial article creation revision. Should be fairly easy to fix I think.

Revisions from *"enwiki > Bumba airport"*:

**Before:** (see last item in image)
screen shot 2016-04-01 at 12 54 00 pm

**After:** (last item from above is missing)
screen shot 2016-04-01 at 12 53 15 pm

coreyfloyd:
@montehurd thats a "feature" not a bug from @ndis1 description. He is removing the last item so the calculation does not fail

coreyfloyd:
At least that is how I understood it… @ndis1 is that correct?

ndis1:
@coreyfloyd this looks like a bug. The first item ever is a special case because its diff is the length of the article, so there is no dependency to calculate it. I tried to put in logic to cover this case `(parentId == nil || parentId.integerValue != 0)` but I guess there's something else needed. I'll take another look

montehurd:
@coreyfloyd the last item doesn't need to be removed if we have fewer than 50 revisions. If fewer than 50 there is no null case (see first screenshot). I think quick fix would be just not removing last item if revision count < 50.

coreyfloyd:
Sounds like you 2 got a handle on it. @montehurd good catch.

ndis1:
@montehurd  updated with bug fix. The last item is not removed if it is the first item. Also added handling for the case when the last day has multiple revisions

I guess I should put this on the ticket as well - QA rundown -

there are 3 known cases
1. all revisions present/fewer than 52 revisions downloaded. In this case parentid of last revision is 0. do not remove any revisions// EXAMPLE: "enwiki > Bumba airport"
2. only 51 of n revisions have been downloaded. The earliest day we have revisions for has multiple revisions. in this case, remove the earliest revision on the earliest day, but leave the day in the table with the rest of its revisions// EXAMPLE: "enwiki > Gregor MacGregor"
3  only 51 of n revisions have been downloaded. The earliest day we have revisions for has only one revision. in this case, remove the day from the revisions days array// EXAMPLE: "enwiki > Nicole, Erica and Jaclyn Dahm"

montehurd:
@ndis1 works great thanks! I tested it with a newly created page too and it looks good in that case as well. :+1:

PR #636, opened 2016-03-31
Fix broken footer license link.

montehurd:
@josve05a

### DIFF
--- a/Wikipedia/Code/WebViewController.m
+++ b/Wikipedia/Code/WebViewController.m
@@ -39,8 +39,8 @@ typedef NS_ENUM (NSInteger, WMFWebViewAlertType) {
     WMFWebViewAlertZeroInterstitial
 };
 
-NSString* const WMFLicenseTitleOnENWiki =
-    @"creativecommons:by-sa/3.0";
+NSString* const WMFCCBySALicenseURL =
+    @"https://creativecommons.org/licenses/by-sa/3.0/";
 
 @interface WebViewController () <ReferencesVCDelegate>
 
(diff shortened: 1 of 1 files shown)


Wikipedia/Code/WebViewController.m
```diff
@@ -290,8 +290,7 @@ - (WMFArticleFooterView*)footerLicenseView {
         @weakify(self);
         [_footerLicenseView.showLicenseButton bk_addEventHandler:^(id sender) {
             @strongify(self);
-            MWKSite* site = [[MWKSite alloc] initWithDomain:WMFDefaultSiteDomain language:@"en"];
-            [self.delegate webViewController:self didTapOnLinkForTitle:[site titleWithString:WMFLicenseTitleOnENWiki]];
```
josve05a: Just a though,
'@"by-sa/3.0";'
and
`MWKSite* site = [[MWKSite alloc] initWithDomain:WMFDefaultSiteDomain language:@"creativecommons:"];
[self.delegate webViewController:self didTapOnLinkForTitle:[site titleWithString:WMFLicenseTitleOnENWiki]];`
would not work then?

montehurd: @josve05a I tried it. No it doesn't work. Out of curiosity, what was the problem with the former page this was linking to?

josve05a: It's an internal page on enwp, which is not translated in to multiple languages (officially). It's true that en.wp (desktop) links to it, but mobile does not, and I think the app should use same standard and wordings as mobile.

montehurd: ah gotcha.

PR #638, opened 2016-04-01
No rotation for old men

coreyfloyd:
err… explore and the tab bar UI. Rotating the UI for Articles and the Image gallery only.

Animation when doing this is a little… meh. But thats just how it works. To doing anything "better" would take a large investment. Probably better to spend that time making better layouts for rotation.


coreyfloyd:
@montehurd spoke with josh about this implementation (only rotating for articles and image gallery) - he is onboard.

montehurd:
:-1: crash on clean install on iOS 8.4 after the onboarding screens are dismissed :(

coreyfloyd:
Wtf. iOS 8!!!!!!! You are the bane of my existence.

coreyfloyd:
@montehurd fixed the bug also added in special handling for iPad (we don't want to restrict rotation on iPad).

montehurd:
Hmm still getting a crash - but now with iOS 9 when the last onboarding screen is dismissed:

> Terminating app due to uncaught exception 'UIApplicationInvalidInterfaceOrientation', reason: 'preferredInterfaceOrientationForPresentation must return a supported interface orientation!'

PR #640, opened 2016-04-02
Multilanguage icon state

ndis1:
[T131081 language icon disabled despite multiple languages](https://phabricator.wikimedia.org/T131081)

There are 4 small parts to this PR. 

1. Article language count is zero indexed. Whether an article has multiple languages was being checked in two places, `WMFArticleFooterMenuDataSource` and `WMFArticleViewController`. `WMFArticleViewController` was saying there are multiple languages if the language count is greater than one, which is an off by one error. Part one fixes that off by one error

2. Because that logic exists in two places, part 2 centralizes the logic as a property on `MWKArticle`

3. Part 3 addresses a different related issue I noticed. On the Main Page, the app is currently showing many languages for the main page feature, even though no other languages are available. This is because the `languagecount` field works differently for the Main Page articles. Part 3 here adds logic to say that if the article title is main, there is only one language.

4 copy update [T131080](https://phabricator.wikimedia.org/T131080)

EDIT: The build on this was failing even though tests were passing locally. I took a look at the `travis.yml` file and noticed `- IOS_VER="9.2" FL_ARGS="verify"`

After bumping that to `9.3` tests are passing. I suspect this is the reason https://github.com/wikimedia/wikipedia-ios/pull/639 failed as well. I'm guessing this sets an environment variable fastlane picks up for some checks?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: objective-c
 osx_image: xcode7.3
 env:
-  - IOS_VER="9.2" FL_ARGS="verify"
+  - IOS_VER="9.3" FL_ARGS="verify"
     #  - IOS_VER="9.2" FL_ARGS="analyze"
     #  - IOS_VER="9.2" FL_ARGS="verify scheme:WikipediaRTL"
     #  - IOS_VER="8.4" FL_ARGS="verify sim_os:${IOS_VER}"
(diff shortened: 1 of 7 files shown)


coreyfloyd:
@ndis1 thanks for fixing the testing flag. I missed that during the "Great Conversion" to Xcode 7.3/iOS 9.3 last week

coreyfloyd:
@ndis1 looks good. 👍  Thanks for tracking down this bug. 

@montehurd want to give this a pass as well?

coreyfloyd:
@montehurd don't worry about updating this branch before merging - we should be fine.

montehurd:
:+1: Very nice!!!

PR #641, opened 2016-04-04
History paging

ndis1:
[Ticket : T70055 paging for edit history](https://phabricator.wikimedia.org/T70055)

This PR adds paging to the edit history page. There are a couple new models using Mantle. The PageHistoryFetcher has been rewritten in swift using the promisekit pattern. I hope I hit the important points of how that's being used here.

A couple TODOs:
1. Tests (I'll try and add early this week)
2. Right now if the call succeeds but json/mantle parsing fails, this shows the user no items. I'm guessing we want to show a generic error message in that case?


coreyfloyd:
@ndis1 this is so awesome. Will get to reviewing this likely tomorrow. @montehurd take a look whenever  you have time - I think this will be a huge improvement.

@ndis1 I also assigned this task to you in Phabricator. Thanks!

Wikipedia/Code/PageHistoryViewController.m
```diff
@@ -191,4 +194,23 @@ - (void)didReceiveMemoryWarning {
     // Dispose of any resources that can be recreated.
 }
 
+- (BOOL)shouldLoadNewData {
+    
+    CGFloat maxY = self.tableView.contentOffset.y + self.tableView.frame.size.height;
+    BOOL shouldLoad = NO;
+    if (!self.loadInProgress && maxY >= self.tableView.contentSize.height) {
```
montehurd: Nitpick: consider moving the `loadInProgress` check to the first line of the method so it can bail out before calculating `maxY`.

Wikipedia/Code/PageHistoryViewController.m
```diff
@@ -81,7 +81,7 @@ - (void)viewDidLoad {
 - (void)getPageHistoryData {
     self.isLoadingData = YES;
     @weakify(self);
-    [self.pageHistoryFetcher fetchRevisionInfoForTitle:self.article.title].then(^(NSMutableArray* items){
+    [self.pageHistoryFetcher fetchRevisionInfoForTitle:self.article.title].then(^(NSArray<NSArray<WMFRevision*>*>* items){
```
montehurd: I'm so glad this can be done in Obj-C now :)

montehurd:
Super slick! Very nice all around. @coreyfloyd I'll leave the deeper review & merge to you - nothing big jumped out at me. I tested it quite a bit and the only thing I noticed was a tiny jitter when new batch is fetched, but I don't think that should block us - this is waaaay better than the old crufty code :+1:

PageHistoryFetcher.swift
```diff
@@ -0,0 +1,129 @@
...
+
+
+class PageHistoryFetcher: NSObject {
+    private let operationManager: AFHTTPSessionManager = {
+        let manager = AFHTTPSessionManager.wmf_createDefaultManager()
+        manager.responseSerializer = WMFApiJsonResponseSerializer()
+        return manager
+    }()
+
+    func fetchRevisionInfo(title: MWKTitle) -> AnyPromise {
```
coreyfloyd: can you mark as public so its obvious that this is the public API?

WMFPageHistorySection.h
```diff
@@ -0,0 +1,17 @@
...
+//  Created by Nick DiStefano on 4/3/16.
+//  Copyright © 2016 Wikimedia Foundation. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+@class WMFPageHistoryRevision;
+
+@interface WMFPageHistorySection : NSObject
+
+@property (nonatomic, copy, readwrite) NSString* _Nullable sectionTitle;
```
coreyfloyd: Can we make these properties readonly and pass them on init so the class is immutable?

coreyfloyd:
@ndis1 This is pretty good good stuff. Left a few small comments, but really on point.

I have a couple larger feedback items - mostly due to the fact that the history view controller / fetcher is one of our legacy classes - so hasn't been updated with our new patterns. If you are up for it, there are a few refactors that would be nice bring it line with our newer classes.

If not thats cool too, we will come back to it eventually as we refactor these classes.

##### 1. Make the fetcher stateless. 
Our new fetchers don't retain anything that has to do with the last fetch. The client is responsible for keeping any state. This would mean removing the batching/paging ivars. The pattern we use for this is instead of returning an array of the results, return a model object like: 

```
class HistoryFetchResults
    var sections
    var continueKey
    var batchComplete
```

Then the client can pass the continue key on the next call.

For a concrete example see the `WMFSearchResults`

##### 2. Use AFNetworking serializers for both request and response serialization. 
This keeps this logic out of the main fetcher class and encapsulates that logic a bit better. Sometimes this isn't entirely possible because our API isn't very restful.

`WMFSearchFetcher` is a pretty good example. (`WMFArticleFetcher` is a good example of where this breaks down because of the API - though I have a solution for this now)


----

If you can at least get my 2 small in line comments that is enough for now. For the big items, let me know if you are for 1 or both of them. If not, like I said, no sweat, this is a big improvement either way.


Thanks again for the help!

💪💪💯

WMFPageHistorySection.h
ndis1: @coreyfloyd sectionTitle can be immutable. The way the items are constructed now the object needs to allow a new item to be added. The parsing could be reconsidered to allow this to be immutable, but that might be more trouble than it's worth. I'll take another look at that later on. 

not 100% related, but I think there is room for improvement in the way dictionaries are constructed in swift in general. I like python's init with array of tuples. https://github.com/apple/swift-evolution/pull/125 is a proposal along those lines. might be worth rolling a version of that if there is a bunch more dictionary manipulation to be done

ndis1:
@coreyfloyd I addressed the smaller updates. I'll take a crack at the larger ones.

WMFPageHistorySection.h
ndis1: @coreyfloyd 

on second thought I see a simple way to make this immutable and keep the current parsing logic. It would involve a good deal of object creation which I'm inclined to shy away from when parsing large data sets, but since there are only 50 items I guess it's ok

ndis1:
@coreyfloyd made fetcher stateless and made afn serializers.
The tricky bit here is holding onto the last revision from the previous call. It would be nice if we could manipulate the paging offset so we wouldn't have to hold onto that item. The api doc says not to manipulate the continue fields though.

coreyfloyd:
👍 looks good to me. @montehurd ok for me to merge?

PR #643, opened 2016-04-05
Fixing nil single frame gifs

coreyfloyd:
https://phabricator.wikimedia.org/T129967

This fixes by incorporating the change from joein3d in our new fork of the AnimatedGIFImageSerialization repo
We had to fork so we could accept Joe's change.

Check https://github.com/wikimedia/AnimatedGIFImageSerialization for details on the fix.

@joein3d can you verify this looks good to you?


montehurd:
Hmm I think we may have mis-identified the issue...

For testing I referenced a single frame gif...
https://commons.wikimedia.org/wiki/File:Name.gif


...on a test wiki article:
> https://test.m.wikipedia.org/wiki/Bird

I then switched the app search language to `test` and loaded the `bird` article and set a breakpoint inside the method. I visually confirmed the UIImage passed to the method was the image from the article and that it had a nil `images` property (i.e. it's a single frame and falls into the PR's single frame handling). I then switched to the develop branch and the old code hit the breakpoint too and the existing code also handled and displayed the image correctly.

So I could be wrong or course, but I suspect single frame handling may not be the issue.

During testing I did notice the gif processing hangs the app significantly when loading the english `phase velocity` article, which has a bunch of gifs. I wonder if the issue is related to processing large gifs with tons of frames?

joein3d:
The title of the issue is a bit misleading - the AnimatedGIFImageSerialization library returns nil, but the images are still rendered correctly in the app due to the fallback to PNG:```    } else if ([mimeType hasSuffix:@"gif"]) {
        NSData* data = [self wmf_gifRepresentation];
        if (data) {
            if (*outMimeType) {
                *outMimeType = @"image/gif";
            }
            return data;
        }
    }

    DDLogWarn(@"Unknown Image Type %@, falling back on PNG", mimeType);
    *outMimeType = @"image/png";
    return [self wmf_pngRepresentation];
}```

montehurd:
@joein3d Oh cool! I'll check that out...

joein3d:
@montehurd your theory about the slow loading process for `phase velocity` is correct:
screen shot 2016-04-06 at 9 08 54 pm
most of the CPU time during the article load is spent in AnimatedGIFImageSerialization. On top of that, it's happening on the main thread. I can look into offloading this onto a background queue, but probably should be written up as a separate issue.

montehurd:
@coreyfloyd I just got it working on a background thread. The speed jitter totally went away.

montehurd:
@coreyfloyd but yeah the main thread issue is separate issue I think.

coreyfloyd:
👍

montehurd:
@joein3d 

> The title of the issue is a bit misleading - the AnimatedGIFImageSerialization library returns nil, but the images are still rendered correctly in the app due to the fallback to PNG

Yup. Spot on. :+1:

PR #646, opened 2016-04-05
Fix build by re-adding localization file

coreyfloyd:
Had to re-add the localization file as it broke our builds.

@siebrand can you confirm that this change does not break your bot?


siebrand:
This should indeed be fine. It's a pattern I see in other files, too.

coreyfloyd:
@siebrand thanks for looking!

coreyfloyd:
@montehurd I'm just merging this… nothing to inspect just fixing our builds.

PR #648, opened 2016-04-07
Move image protocol processing to background thread.

montehurd:
Noticed `enwiki > phase velocity` article, which has a couple animate gifs, was hanging severely on load.

Investigated and was a bit surprised to see our image processing was happening on the main thread.

I suspect this main thread thrashing could have something to do with the odd image protocol crash reports we've seen.

*I know you shouldn't do anything from a background thread which could cause already onscreen control elements to trigger a layout pass, but iirc processing "headless" UIImages is fine.*

- Moves image protocol image processing to background thread.
- Improves error handling a bit.

`enwiki > phase velocity` no longer hangs! :)


montehurd:
Was a bit of a surprise to see our image processing was happening on the main thread. 

I know you shouldn't do anything from a background thread which could cause already onscreen control elements to trigger a layout pass, but iirc processing "headless" UIImages is fine.

montehurd:
Even more convinces this could be behind the odd image protocol errors we were seeing...

Wikipedia/Code/WMFArticleImageProtocol.m
```diff
@@ -55,37 +71,37 @@ - (void)startLoading {
...
-        NSData* data       = [image wmf_dataRepresentationForMimeType:mimeType serializedMimeType:&mimeType];
-        
-        NSURLResponse* response =
-        [[NSURLResponse alloc] initWithURL:self.request.URL
-                                  MIMEType:mimeType
-                     expectedContentLength:data.length
-                          textEncodingName:nil];
-        
-        return (response && data) ? @[response, data]: nil;
+        return [self getResponseAndDataForDownload:download];
```
montehurd: All image processing now cleanly on background thread.

Wikipedia/Code/WMFArticleImageProtocol.m
```diff
@@ -55,37 +71,37 @@ - (void)startLoading {
...
+    .then(^id(WMFArticleImageProtocolResponseAndData* responseAndData) {
         @strongify(self);
         if(!self){
-            return;
+            return nil;
         }
 
-        if (!responseAndDataArray || (responseAndDataArray.count != 2)) {
-            return;
+        if (!responseAndData.response) {
```
montehurd: If there was any problem making a response or processing the image now returns errors so the catch block will fire. Formerly if anything went wrong in either of these parts "URLProtocolDidFinishLoading" was still invoked, which indicated success, when in fact the loading had failed, which is what the catch block correctly reports.

Wikipedia/Code/WMFArticleImageProtocol.m
```diff
@@ -55,12 +71,44 @@ - (void)startLoading {
...
-        [self respondWithDataFromDownload:download];
+        return [self getResponseAndDataForDownload:download];
+    })
+    .then(^id(WMFArticleImageProtocolResponseAndData* responseAndData) {
+        @strongify(self);
+        if(!self){
+            return nil;
+        }
+
+        if (!responseAndData.response) {
```
montehurd: If there were any problems making a response or processing the image (on the background thread) we now return errors so the catch block will fire. Formerly if anything went wrong in either of these parts "URLProtocolDidFinishLoading" was still invoked, which indicated success, when in fact the loading had failed, which is what the catch block correctly reports.

I very much suspect by not correctly reporting failure there were protocol bits behind the scenes which were then hanging around too long any time there were issues retrieving images.

Wikipedia/Code/WMFArticleImageProtocol.m
```diff
@@ -55,12 +71,44 @@ - (void)startLoading {
...
     [[WMFImageController sharedInstance] fetchImageWithURL:self.request.URL]
-    .then(^(WMFImageDownload* download) {
+    .thenInBackground(^WMFArticleImageProtocolResponseAndData*(WMFImageDownload* download) {
         @strongify(self);
         if(!self){
-            return;
+            return nil;
         }
-        [self respondWithDataFromDownload:download];
+        return [self getResponseAndDataForDownload:download];
```
montehurd: All image processing now cleanly on background thread.

Wikipedia/Code/WMFArticleImageProtocol.m
```diff
@@ -23,6 +23,22 @@
 
 NSString* const WMFArticleImageSectionImageRetrievedNotification = @"WMFSectionImageRetrieved";
 static NSString* const WMFArticleImageProtocolHost               = @"upload.wikimedia.org";
+static NSString* const WMFArticleImageProtocolResponseAndDataErrorDomain = @"WMFArticleImageProtocolResponseAndDataErrorDomain";
+
+@interface WMFArticleImageProtocolResponseAndData: NSObject
```
montehurd: Containment for the parts we want to hand off to a background thread for processing.

joein3d:
This looks great! Definitely improves the loading experience. Only thing that might be worth exploring (if there are any memory use issues for articles with a bunch of these heavy GIF conversions happening) is using a dispatch_semaphore or NSOperationQueue to control the number of concurrent conversions that are happening.

Long term I would propose a solution that doesn't use SDWebImage under the hood. If I'm understanding correctly - SDWebImage grabs the image data (or an image object) from the memory cache, disk cache, or from a network request. Then, if it's data and not an image object, it converts it to a UIImage, which wikipedia-ios converts back to data to send to the web view. If the cache solution could return the original data (for example, with NSURLSession's built in cache), the conversion to animated gif data could be removed from wikipedia-ios entirely.

Wikipedia/Code/WMFArticleImageProtocol.m
coreyfloyd: Got it… so now you are making sure we respond with an error if there is not image to display.

Wikipedia/Code/WMFArticleImageProtocol.m
```diff
@@ -55,12 +71,44 @@ - (void)startLoading {
...
+        }
+
+        if (!responseAndData.response) {
+            return [[NSError alloc] initWithDomain:WMFArticleImageProtocolResponseAndDataErrorDomain
+                                              code:0
+                                          userInfo:@{
+                                                     NSLocalizedDescriptionKey: @"Response for image not found",
+                                                     NSURLErrorKey: self.request.URL
+                                                     }];
+        }else if (!responseAndData.data) {
```
coreyfloyd: You can remove this check and just use the one below… `responseAndData.data.length=0` will catch this as well.

Wikipedia/Code/WMFArticleImageProtocol.m
```diff
@@ -23,6 +23,22 @@
 
 NSString* const WMFArticleImageSectionImageRetrievedNotification = @"WMFSectionImageRetrieved";
 static NSString* const WMFArticleImageProtocolHost               = @"upload.wikimedia.org";
+static NSString* const WMFArticleImageProtocolResponseAndDataErrorDomain = @"WMFArticleImageProtocolResponseAndDataErrorDomain";
```
coreyfloyd: `WMFArticleImageProtocolErrorDomain` sound be enough - domains can usually just be scoped to the class

coreyfloyd:
@montehurd looks good - nice work. 💪 

I left some minor comments but nothing crazy. 


@joein3d we know about what is happening with SDWebImage, but it hasn't seemed to noticeably affect the performance. But it for sure is something to keep an eye on.

I do think it is worth investigating allowing the web view to use the NSURLCache through - that is something I wanted to look into anyways.

coreyfloyd:
> I know you shouldn't do anything from a background thread which could cause already onscreen control elements to trigger a layout pass, but iirc processing "headless" UIImages is fine.

UIImages are fine to handle on a background thread. They are a UIKit class, but they are not tied to an on screen component. So you can always go to a background thread if need be. Just make sure you need to touch a view, you go back to the main thread before doing so.

That doesn't apply here because you are responding using NSURLProtocol methods, which is a Foundation class.

coreyfloyd:
@montehurd I am pre- 👍 ing pending your upcoming changes.

montehurd:
>  If the cache solution could return the original data (for example, with NSURLSession's built in cache), the conversion to animated gif data could be removed from wikipedia-ios entirely.

Oh, these images are also presented in native parts of the app so I think we still need this handling for those parts.

joein3d:
@montehurd yeah, for the memory cache, the data would have to be stored along with the UIImage objects, which is less than ideal. It seems like @coreyfloyd 's suggestion of using the built in caching would be better.

joein3d:
As a test, I disabled WMFArticleImageProtocol and did a clean install of the app on the iOS sim. I loaded the phase velocity article, quit the app, disabled my internet connection, and re-ran the app. The images are all there. So there is caching happening within UIWebView - are there places in the app where more control is needed over the cache - for example are articles saved for reading later supposed to be guaranteed to be available offline?

coreyfloyd:
> are there places in the app where more control is needed over the cache - for example are articles saved for reading later supposed to be guaranteed to be available offline?

Yeah - for saved pages we need to have the images available offline - guaranteed. My thought was that if we used the built in cache, we could pull the images out of the NSURLCache when a user "saves" an article" and put them in our persistent cache. That way they will never be cleaned up.

coreyfloyd:
@joein3d also that means we would need a NSURLCache subclass still - to route requests for images on saved pages into our persistent cache.

coreyfloyd:
@montehurd @joein3d just to put this to bed - I'm going to file a ticket for the investigating NSURLCache usage so we can just let UIWebView do "its thing" for caching but with a way for us to guarantee some images are persistent.

montehurd:
@joein3d oh for sure the UIWebView does some of its own caching, both in-memory and persistent. (iirc a couple versions ago, maybe iOS 5 or 6, I seem to recall even tracking down the sqlite store it was using)

Anyways, at that time I did something similar to the steps you outline (save a bunch of articles, turn off our cache interception, go offline, restart app). 

The thing I found is if you browse to enough articles offline (at the time iirc it was ~6 or 8 medium sized articles) it would fairly quickly get into a state where the memory cache was purged and it wasn't able to get the resources from the persistent store, so you do eventually see missing images. That's where our protocol interception comes in to save the day.

But for sure things may have changed since then and we may have to re-evaluate - or maybe there are new WKWebview bits we could try... :)

montehurd:
@coreyfloyd ^

PR #649, opened 2016-04-07
Add support for custom Zero exit interstitial messages

mdholloway:
Needed for Wikipedia Zero program compliance.

I'm opening this PR early for comment but will add commits with the implementation.

Bug: https://phabricator.wikimedia.org/T119128


mdholloway:
For some reason I haven't yet gotten the otherButton in the exit dialog to follow the partnerInfoUrl when it's present.  Instead the button does nothing.

coreyfloyd:
@mdholloway I think the Api you want to use to add buttons is:


- (NSInteger)bk_addButtonWithTitle:(NSString *)title handler:(void (^)(void))block

mdholloway:
Thanks @coreyfloyd.  Turns out I also needed to create a new NSString* with the url from self.zeroMessage.partnerInfoUrl to feed to the handler before openUrl would work.  

This needs more work, though.  I just realized I changed the wrong dialog.

mdholloway:
OK, this now changes the correct dialog (the one for external URLs)!

coreyfloyd:
@mdholloway can you update your branch with the latest from develop?

Also you don't have to use a separate fork - just make a branch and push to this repo - makes it easier to keep your pr in sync with develop.

coreyfloyd:
👍 @montehurd ?

montehurd:
@coreyfloyd testing...

montehurd:
@mdholloway heya just a quick confirmation - I faked out an wmf_openExternalUrl warning by just changing the new fields to their keys - does the screenshot below look good (assuming actual messages will be shown instead of the keys)?

screen shot 2016-04-08 at 1 41 36 pm

mdholloway:
Yep, looks as intended!

PR #650, opened 2016-04-07
(Text) Size Matters

coreyfloyd:
Adding text size slider / popup

The text setting implementation was based on @montehurd's spike work.

Not much else to say, used some Pods… couldn't use the PopoverViewController because that won't display as a popup on iPhones.

You can play with the step values in the Developer settings (Tweaks) menu under "Article"

text size


coreyfloyd:
@montehurd I am making a separate ticket for auditing this functionality for accessibility. I want to get the implementation out for the designers/product to evaluate.

reetssydney:
Hey @coreyfloyd, just updated the sizes for the small and large T to 1x size.
[T-large.pdf](https://github.com/wikimedia/wikipedia-ios/files/208864/T-large.pdf)
[T-small.pdf](https://github.com/wikimedia/wikipedia-ios/files/208865/T-small.pdf)

montehurd:
@coreyfloyd crashes on iOS 8. Gonna keep looking over code...

montehurd:
@coreyfloyd crash on iOS 9.3 after peeking and popping a couple web view links.

coreyfloyd:
> @coreyfloyd crash on iOS 9.3 after peeking and popping a couple web view links.

@montehurd  Can you attach a stack trace? I can't repro.

I am working on the iOS 8 crash…

coreyfloyd:
> @coreyfloyd crash on iOS 9.3 after peeking and popping a couple web view links.

Finally able to repro: So not related to this patch:

```
2016-04-07 19:04:08.506 Wikipedia Debug[7584:2723377] *** Assertion failure in -[WMFArticleViewController initWithArticleTitle:dataStore:], /Users/cfloyd/Dropbox/Development/wikipedia-ios/Wikipedia/Code/WMFArticleViewController.m:133
2016-04-07 19:04:08.514 Wikipedia Debug[7584:2723377] *** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'Invalid parameter not satisfying: title'
libc++abi.dylib: terminating with uncaught exception of type NSException
```

It says the article VC isn't being instantiated with a title. So something with peeks and pops… filling a bug ticket.

(Note: since this is an assertion, you won't see a crash in production)

TL;DR relevant method:
```
- (instancetype)initWithArticleTitle:(MWKTitle*)title
                           dataStore:(MWKDataStore*)dataStore {
    NSParameterAssert(title);
    NSParameterAssert(dataStore);

    self = [super init];
    if (self) {
        self.articleTitle             = title;
        self.dataStore                = dataStore;
        self.hidesBottomBarWhenPushed = YES;
    }
    return self;
}
```

coreyfloyd:
@montehurd also fixed the iOS 8 bug

coreyfloyd:
@montehurd FYI https://phabricator.wikimedia.org/T132117

Wikipedia/assets/index.html
```diff
@@ -6,7 +6,11 @@
     <meta name="viewport" id="viewport" content="width=device-width, user-scalable=no">
     <link href="styles.css" rel="stylesheet" type="text/css"></link>
     <link href="styleoverrides.css" rel="stylesheet" type="text/css"></link>
-    
+    <style type="text/css">
```
montehurd: I'd like to revisit having the text size set via strings substitution when we have more time to visit how transforms are handled. But in the mean time this is fine with one exception:

- as things are presently structured the editing interface's `preview.html` is processed by the same `stringWithFormat`, so for it to continue working  `preview.html` will need `%@` to be added to it as well. It can just be surrounded by html comments - something like: `<!-- %@ (needed for consistency with index.html) -->`. Will also have to come before the `%@` already in `preview.html`.

montehurd:
Sorry found one more crasher:

- Clean install
- Tap on article from feed
- Click on link in article
- When new article loads increase its text size by one increment
- Now tap on another link
- Keep repeating the last 2 steps a few more times (tap link, when it loads make the text size one step bigger)
- Then tap back button all the way back to the feed
- Crashy crash

reetssydney:
hi @coreyfloyd - looks good except for a few tweaks:

- for some reason the arms of the capital T in all the icons "TT" icon are rendering thicker; so I've resupplied the pdfs with slight tweaks which should resolve.
[resize-icon.pdf](https://github.com/wikimedia/wikipedia-ios/files/210162/resize-icon.pdf)
[T-large.pdf](https://github.com/wikimedia/wikipedia-ios/files/210190/T-large.pdf)
[T-small.pdf](https://github.com/wikimedia/wikipedia-ios/files/210191/T-small.pdf)

- the vertical bar of the small and large t icons should align with the min and max tick marks respectively. Can we fix this by extending the track's length by 12px (6px on either side?)
image

- when an article loads, the TT text-size icon (and the Save for later icon) are actually shown in the #3366CC blue colour already, instead of the transitioning from pale grey to blue as expected like the language, share and TOC icons on the toolbar.
icon-fadeto 3366cc
 
- As for layout, the one thing I have noticed is that the line-height on the section headings is looser than expected, which is that it should be the same line-height: 1.3 as the web (if not tighter). See example comparison below.
image
Is this something we can fiddle with as well?

PR #651, opened 2016-04-08
Making text slider Accessible… yo

coreyfloyd:
https://phabricator.wikimedia.org/T132059

- Updated our slider pod to be accessible.
- Using standard popup (didn't know I could do that and have it still look like a pop up on the iPhone!)

Below is a gif using voice over to navigate our slider… 💥 
🎤 ⬇️ 🚶 

access


coreyfloyd:
Note the gif does not contain the updated artwork

Wikipedia/Images.xcassets/T-large.imageset/Contents.json
```diff
@@ -2,7 +2,7 @@
   "images" : [
     {
       "idiom" : "universal",
-      "filename" : "T-large.pdf"
+      "filename" : "T-large-2.pdf"
```
montehurd: Hmm... I think we'll want to delete the `T-large` and `T-small` from the assets catalog, then re-add them so it doesn't do this weird `-2` business.

PR #652, opened 2016-04-09
Fix request "user agent" headers. Cover with tests.

montehurd:
https://phabricator.wikimedia.org/T131824

- request `user agent` headers now properly begin with `WikipediaApp`, as needed for analytics
- added tests for `mobileview` request header user agent, gzip encoding, and opt-in uuid

### DIFF
--- a/Wikipedia/Code/WMFArticleFetcher.h
+++ b/Wikipedia/Code/WMFArticleFetcher.h
@@ -23,8 +23,6 @@ extern NSString* const WMFArticleFetcherErrorCachedFallbackArticleKey;
 - (void)cancelFetchForPageTitle:(MWKTitle*)pageTitle;
 - (void)cancelAllFetches;
 
-@property (nonatomic, strong) AFHTTPSessionManager* operationManager;
-
 @end
 
 @interface WMFArticleFetcher : WMFArticleBaseFetcher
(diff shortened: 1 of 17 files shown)


montehurd:
@coreyfloyd I was wrong about the user agent being set all other places correctly. *Some* were if they happened to use "wmf_createDefaultManager" and didn't override the manager's request serializer. This PR should fix that :)

Wikipedia/Code/WMFArticleFetcher.h
```diff
@@ -22,6 +23,8 @@ extern NSString* const WMFArticleFetcherErrorCachedFallbackArticleKey;
 - (void)cancelFetchForPageTitle:(MWKTitle*)pageTitle;
 - (void)cancelAllFetches;
 
+@property (nonatomic, strong) AFHTTPSessionManager* operationManager;
```
coreyfloyd: I see that you did this for a test which is good, but we shouldn't leave this as a read/write property in the public API.

What we have been typically doing is:
1. Define this property in a class extension in the implementation file (`@property (nonatomic, strong) AFHTTPSessionManager* operationManager;`)
2. Create a separate category file for testing… like  WMFArticleFetcher_Tests.h
3. In the header of that category declare this method again (`@property (nonatomic, strong) AFHTTPSessionManager* operationManager;`)
4. Import this category header into your test.

Make sense?

coreyfloyd:
@montehurd looking good - left one comment on removing the readwrite property from the public header.

Wikipedia/Code/WMFArticleFetcher.h
montehurd: On it! :)

montehurd:
@coreyfloyd Ok I think this should be better now - no more exposing `operationManager`.

PR #660, opened 2016-04-14
Fix IP edits (maybe?) plus a few boyscouts

coreyfloyd:
(Potentially) Fixes IP edits by logging in the user when the app is resumed.

I say potentially because I wasn't able to reproduce. The thought is that our tokens are expiring and we have no logic to renew our tokens. We should be doing this on launch to ensure we are logged in. That is what this pull request does. We will need to deploy to see if this actually helps.

I definitely "repeated" myself here… since the logic for logins is in a view controller and not a model controller / headless object.

I would like to address that (as well as modernizing the fetcher logic) in a larger engineering task:
https://phabricator.wikimedia.org/T132700

This also fixes some issues with updates on a background thread and where we would do a double fetch and it might prevent the article from being updated after an edit.

### DIFF
--- a/Wikipedia/Code/EventLoggingFunnel.m
+++ b/Wikipedia/Code/EventLoggingFunnel.m
@@ -9,8 +9,9 @@
 #import "EventLoggingFunnel.h"
 #import "EventLogger.h"
 #import "QueuesSingleton.h"
-#import "SessionSingleton.h"
 #import "MediaWikiKit.h"
+#import "Wikipedia-Swift.h"
+#import "SessionSingleton.h"
 
 @implementation EventLoggingFunnel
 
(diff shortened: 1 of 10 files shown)


Wikipedia/Code/EventLoggingFunnel.m
```diff
@@ -28,8 +29,7 @@ - (NSDictionary*)preprocessData:(NSDictionary*)eventData {
 }
 
 - (void)log:(NSDictionary*)eventData {
-    SessionSingleton* session = [SessionSingleton sharedInstance];
-    NSString* wiki            = [session.currentArticleSite.language stringByAppendingString:@"wiki"];
```
montehurd: Could this have the effect of causing article specific events appear to come from wrong lang wiki? Not sure if we're even doing anything like that...

coreyfloyd: good catch - i thought i avoided all the "article specific" ones… but this one might be… I'll back it out here.

PR #657, opened 2016-04-14
Use "Your languages" and "Other languages" section headers with article lang picker. Restyle it too.

montehurd:
https://phabricator.wikimedia.org/T130675#2205878

- [x] Use "Your languages" and "Other languages" section headers 
- [x] Update title to "Change language"
- [x] Update text sizes
- [x] Update paddings
- [x] Add thin separator beneath filter box
- [x] Add thicker separator between cells
- [x] Hide "Your languages" and "Other languages" headers if no respective langs for article
- [x] Remove the alert (per @nirzar)

article langs mov

screen shot 2016-04-13 at 8 07 25 pm

PR #662, opened 2016-04-15
Add support for add/remove/sort of preferred languages.

montehurd:
https://phabricator.wikimedia.org/T130875

**Note:** this is on top of https://github.com/wikimedia/wikipedia-ios/pull/657
    (once 657 is merged this diff won't look so big - all the changes for this PR are in the *"Add support for add/remove/sort of preferred languages."* commit below)
Also, I took to heart the *"use existing functionality"* approach pretty strongly here, so there are some deviations from the mocks.

**Allows preferred langs to be:**
- [x] added
- [x] removed
- [x] re-ordered

**Preserves existing:**
- [x] article lang picking functionality
- [x] lang filtering
- [x] cell layout - we do still need canonical lang name (even Apple does this)

**Updated:**
- [x] search bar *exactly* presents languages in order user has specified
- [x] when user's preferred language list changes the first language is used as the primary search language (also appears as first language in search bar languages)
- [x] title to say "My languages" both in settings an atop lang interface when it's not being used to choose an article language

**Prevents:**
- [x] all preferred langs from being removed - there must be at least one
- [x] adding langs to the user's preferred list by any implicit mechanism - the user controls these now via "My languages"
- [x] preferred langs from being dragged/re-ordered when filter is not blank - you can't re-order if you don't know what other items you are reordering in relation to.

langs mov


Wikipedia/Code/LanguageCell.m
```diff
@@ -20,8 +20,6 @@ @interface LanguageCell ()
 @property (strong, nonatomic) IBOutlet UILabel* languageCodeLabel;
```
montehurd: @coreyfloyd note that all commits above this are from my other open lang PR.

reetssydney:
hey @montehurd - just s/w Nirzar and he was fine with the diffs as we discussed, but had one more comment about wanting to remove the URL being shown for the languages.
image

Would be great if you can add this to other changes discussed. Thanks!

montehurd:
@coreyfloyd hey I sent you an email update about this... sorry about the delay :(

PR #665, opened 2016-04-18

niedzielski:


coreyfloyd:
@niedzielski going to let this one sit until @montehurd returns from vacation Thursday.

PR #670, opened 2016-04-20
Fix crash on Endoplasmic Reticulum

coreyfloyd:
https://phabricator.wikimedia.org/T131351

Crash was caused by a very large gif - see ticket above.

Fixed by changing logic to not use multiple copies of the gif and by not force inflating it on creation. Instead we now let the web view and NSURLCache do their thing. And they do it much better.

We no longer need the image protocol. That has been removed.


Wikipedia/Code/SavedArticlesFetcher.m
```diff
@@ -127,30 +128,30 @@ - (void)fetchUncachedTitles:(NSArray<MWKTitle*>*)titles {
...
-        });
-    }).catch(^(NSError* error){
-        if (!self) {
-            return;
-        }
-        dispatch_async(self.accessQueue, ^{
-            [self didFetchArticle:nil title:title error:error];
+    if (articleFromDisk.isCached) {
+        // only fetch images if article was cached
+        [self downloadImageDataForArticle:articleFromDisk];
```
montehurd: @coreyfloyd 
Little tough to follow... it isCached, yet we're downloading? Does this cause images to be fetched twice? (by web view and here?)

Wikipedia/Code/WMFURLCache.h
```diff
@@ -3,6 +3,12 @@
 
 #import <Foundation/Foundation.h>
 
+@class MWKArticle;
+
 @interface WMFURLCache : NSURLCache
 
+- (void)permenantlyCacheImagesForArticle:(MWKArticle*)article;
```
montehurd: @coreyfloyd typo

Wikipedia/Code/WMFImageController.swift
```diff
@@ -246,6 +246,14 @@ public class WMFImageController : NSObject {
         self.imageManager.imageCache.clearDisk()
     }
 
+    public func cacheImageData(imageData: NSData, url: NSURL){
```
montehurd: @coreyfloyd Could we name this "addImageDataToCache"? "cache" being both a noun and a verb and all.

PR #673, opened 2016-04-23
Updates to language management

montehurd:
https://phabricator.wikimedia.org/T130875

langs mov


Wikipedia/Code/WMFLanguageCell.h
```diff
@@ -1,12 +1,14 @@
+#import "WMFTableViewCellWithCustomMinusButton.h"
 
 /// Table cell designed for displaying a language autonym & a page title in that language.
-@interface WMFLanguageCell : UITableViewCell
+@interface WMFLanguageCell : WMFTableViewCellWithCustomMinusButton
```
coreyfloyd: WMFTableViewCellWithCustomDeleteButton - minor, but minus happens to be the symbol that is used, but delete is more to the point.

Wikipedia/Code/WMFLanguageCell.m
```diff
@@ -70,14 +73,28 @@ - (void)awakeFromNib {
     [self prepareForReuse];
     [self wmf_makeCellDividerBeEdgeToEdge];
     self.showsReorderControl = YES;
+    self.primaryLabel.textColor = [UIColor whiteColor];
```
coreyfloyd: Just set this in the nib

Wikipedia/Code/WMFLanguageCell.m
```diff
@@ -70,14 +73,28 @@ - (void)awakeFromNib {
     [self prepareForReuse];
     [self wmf_makeCellDividerBeEdgeToEdge];
     self.showsReorderControl = YES;
```
coreyfloyd: This can also be set in the nib

Wikipedia/Code/WMFLanguageCell.m
```diff
@@ -70,14 +73,28 @@ - (void)awakeFromNib {
...
+    self.primaryLabel.textColor = [UIColor whiteColor];
+    self.isPrimary = NO;
 }
 
 - (void)prepareForReuse {
     [super prepareForReuse];
     self.languageNameLabel.text           = @"";
```
coreyfloyd: Is there a reason we are not setting these to nil?
Usually this is the way we clear a label as opposed to empty strings unless there is a bug you need to work around.

Wikipedia/Code/WMFLanguageCell.m
```diff
@@ -70,14 +73,28 @@ - (void)awakeFromNib {
     [self prepareForReuse];
     [self wmf_makeCellDividerBeEdgeToEdge];
     self.showsReorderControl = YES;
+    self.primaryLabel.textColor = [UIColor whiteColor];
+    self.isPrimary = NO;
```
coreyfloyd: unneeded - the default for Bool ivars is "NO"

Wikipedia/Code/WMFLanguageCell.m
```diff
@@ -70,14 +73,28 @@ - (void)awakeFromNib {
...
+}
+
+- (void)setIsPrimary:(BOOL)isPrimary {
+    _isPrimary = isPrimary;
+    if (isPrimary){
+        self.primaryLabel.text = [MWLocalizedString(@"settings-primary-language", nil) uppercaseStringWithLocale:[NSLocale currentLocale]];
+        self.primaryLabelContainerView.backgroundColor = [UIColor colorWithRed:0.8039 green:0.8039 blue:0.8039 alpha:1.0];
+    }else{
+        self.primaryLabel.text = @"";
```
coreyfloyd: Why not nil?

Wikipedia/Code/WMFLanguageCell.m
```diff
@@ -70,14 +73,28 @@ - (void)awakeFromNib {
...
-    self.languageNameLabelHeight.constant = 0.f;
+    self.localizedLanguageLabelHeight.constant = 0.f;
+    self.isPrimary = NO;
+}
+
+- (void)setIsPrimary:(BOOL)isPrimary {
+    _isPrimary = isPrimary;
+    if (isPrimary){
+        self.primaryLabel.text = [MWLocalizedString(@"settings-primary-language", nil) uppercaseStringWithLocale:[NSLocale currentLocale]];
+        self.primaryLabelContainerView.backgroundColor = [UIColor colorWithRed:0.8039 green:0.8039 blue:0.8039 alpha:1.0];
```
coreyfloyd: Should be in the style category

Wikipedia/Code/WMFLanguageCell.m
```diff
@@ -70,14 +73,28 @@ - (void)awakeFromNib {
...
 }
 
 - (void)prepareForReuse {
     [super prepareForReuse];
     self.languageNameLabel.text           = @"";
     self.articleTitleLabel.text           = @"";
     self.localizedLanguageLabel.text      = @"";
-    self.languageNameLabelHeight.constant = 0.f;
+    self.localizedLanguageLabelHeight.constant = 0.f;
+    self.isPrimary = NO;
```
coreyfloyd: What about isPreferred? Do we need to reset that?

Wikipedia/Code/WMFLanguagesViewController.m
```diff
@@ -433,6 +449,31 @@ - (void)languagesController:(WMFLanguagesViewController*)controller didSelectLan
...
+
+- (UITableViewCell*)tableView:(UITableView*)tableView cellForRowAtIndexPath:(NSIndexPath*)indexPath {
+    WMFLanguageCell* cell = (WMFLanguageCell*)[super tableView:tableView cellForRowAtIndexPath:indexPath];
+    cell.isPrimary = (indexPath.row == 0) ? YES : NO;
+    return cell;
+}
+
+- (void)tableView:(UITableView*)tableView moveRowAtIndexPath:(NSIndexPath*)sourceIndexPath toIndexPath:(NSIndexPath*)destinationIndexPath {
+    [super tableView:tableView moveRowAtIndexPath:sourceIndexPath toIndexPath:destinationIndexPath];
+    [tableView reloadData];
```
coreyfloyd: Should use insertion APIs instead of reloading.

usually the only reason to reload a table is if the entire dataset is changing or if the view is re-appearing.

Wikipedia/Code/WMFTableViewCellWithCustomMinusButton.h
```diff
@@ -0,0 +1,11 @@
+#import <UIKit/UIKit.h>
+
+/**
+ * Table cell with a custom minus button achieved by overlaying the default minus button with a custom minus button.
+ */
+
+@interface WMFTableViewCellWithCustomMinusButton : UITableViewCell
+
+@property (strong, nonatomic, readonly) UIButton* minusButton;
```
coreyfloyd: `deleteButton`

In general we should name things base don functionality rather than appearance as the designs can change but the purpose of deleting will stay the same.

Wikipedia/Code/WMFTableViewCellWithCustomMinusButton.m
```diff
@@ -0,0 +1,71 @@
...
+@property (strong, nonatomic) CALayer* minusButtonWhiteLayer;
+
+@end
+
+@implementation WMFTableViewCellWithCustomMinusButton
+
+- (void)layoutSubviews {
+    [super layoutSubviews];
+    
+    // HAX: ensure the custom button is always on top.
```
coreyfloyd: Not a hack… no need to mark as hack.

Wikipedia/Code/WMFTableViewCellWithCustomMinusButton.m
```diff
@@ -0,0 +1,71 @@
...
+    [self.minusButton.superview bringSubviewToFront:self.minusButton];
```
coreyfloyd: We should only call this is the if it isn't frontmost view instead of calling every time.

Wikipedia/Code/WMFTableViewCellWithCustomMinusButton.m
```diff
@@ -0,0 +1,71 @@
...
+
+- (void)awakeFromNib {
+    [super awakeFromNib];
+    
+    self.minusButton = [self wmf_minusButton];
+    [self.contentView.superview addSubview:self.minusButton];
+    [self.minusButton mas_makeConstraints:^(MASConstraintMaker* make) {
+        make.width.equalTo(@(50));
+
+        // HAX: attach leading edge to contentView so it tracks horizonally
```
coreyfloyd: Not a hack - good to comment… but aligning a view to another view isn't a hack. We should reserve this word for things that are actually hacks.

Wikipedia/Code/WMFTableViewCellWithCustomMinusButton.m
```diff
@@ -0,0 +1,71 @@
...
+        // with cell animations such as when enabling table view editing with
+        // animations.
+        make.leading.equalTo(self.contentView).offset(-40);
+
+        // HAX: use offsets for top and bottom so as not to overlap separators.
```
coreyfloyd: same here - good to say what you are doing, but no need to call it a hack.

Wikipedia/Code/WMFTableViewCellWithCustomMinusButton.m
```diff
@@ -0,0 +1,71 @@
...
+    [button setImage:[UIImage imageNamed:@"language-delete"] forState:UIControlStateNormal];
+    button.imageView.backgroundColor = [UIColor whiteColor];
+    
+    self.minusButtonWhiteLayer = [self whiteLayer];
+    [button.layer insertSublayer:self.minusButtonWhiteLayer atIndex:0];
+    
+    return button;
+}
+
+// HAX: need white layer beneath the custom minus button to hide the baked-in minus button.
```
coreyfloyd: Try just making the button opaque so you can't see views behind it.

https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIView_Class/index.html#//apple_ref/occ/instp/UIView/opaque


In general any view you can make opaque you should as it improves the performance of rendering

Wikipedia/Code/WMFLanguageCell.h
montehurd: I used "minus" to avoid confusion with the delete button which appears when you tap the minus button, and with the "delete" enums/methods which deal with that button... I went back and forth on it...

Wikipedia/Code/WMFLanguageCell.m
montehurd: Fixed.

Wikipedia/Code/WMFLanguageCell.m
montehurd: Fixed.

Wikipedia/Code/WMFLanguageCell.h
coreyfloyd: @montehurd no need to worry about that… ALL cells support custom delete buttons on the right. No custom class is needed for that, so there shoudn't be much confusion.

And you can comment the header to explain the exact delete button that is being customized.

Also I forgot to mention, you the last part of the class name should be the type, the "adjectives" sound come first, so:

`WMFCustomDeleteButtonTableViewCell`

A good test is that classes generally shouldn't have the word "with" in them (unlike methods).

Wikipedia/Code/WMFLanguageCell.m
montehurd: Fixed.

Wikipedia/Code/WMFLanguageCell.m
montehurd: All must have been testing cruft - removed it.

Wikipedia/Code/WMFLanguageCell.m
montehurd: Fixed.

Wikipedia/Code/WMFLanguageCell.m
montehurd: Fixed.

Wikipedia/Code/WMFLanguageCell.h
montehurd: Fixed.

Wikipedia/Code/WMFLanguageCell.m
montehurd: Probably not now that we're not showing both preferred and non-preferred in the same table any more, but I'll add it just in case.

Wikipedia/Code/WMFTableViewCellWithCustomMinusButton.h
montehurd: Fixed.

Wikipedia/Code/WMFTableViewCellWithCustomMinusButton.m
montehurd: Fair enough. Removed.

Wikipedia/Code/WMFTableViewCellWithCustomMinusButton.m
montehurd: Fixed.

Wikipedia/Code/WMFTableViewCellWithCustomMinusButton.m
montehurd: Fixed.

Wikipedia/Code/WMFTableViewCellWithCustomMinusButton.m
montehurd: I don't think it triggers a layout pass, but doesn't hurt to check it I guess. Fixed.

PR #676, opened 2016-04-26
ToC hint animation

coreyfloyd:
https://phabricator.wikimedia.org/T126703

Only shows once for each user.

@reetssydney we didn't discus whether we should cover the article with black during the hint. I did that here but it is easily removed, can you let me know what you think?


reetssydney:
Hey @coreyfloyd - think the black is a bit aggressive and would prefer the same transition as when user goes from feed to article view. I.e., with a drop shadow under the TOC and the article lightening slightly. 
Had a go at updating this here:
https://github.com/wikimedia/wikipedia-ios/pull/677
Effect is shown in this screencast:
https://youtu.be/0EeTQ_HJZDA

coreyfloyd:
@montehurd this is good for review now

PR #677, opened 2016-04-26
Toc hint2

reetssydney:
updated the visual display of the ‘peek’ with a shadow and fade closer to the transition between feed to article view.


coreyfloyd:
Added these changes in #676, closing

PR #678, opened 2016-04-27
Add new Support setting

coreyfloyd:
Uses a subclassed article VC to display the FAQ.

This touches a lot of underlying logic about MWKSite / MWKTitle. Otherwise it is straightforward.


montehurd:
Can we hide the footer for the help page?

Noticed something odd about the footer too - my simulator is in English, but a couple of the strings in the help page footer are not English:

screen shot 2016-04-27 at 12 22 38 pm

screen shot 2016-04-27 at 12 22 44 pm

Wikipedia/Code/WMFLeadingImageTrailingTextButton.h
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ *  The selected text shown to the right of the image.
+ */
+@property (nonatomic, strong) NSString* selectedLabelText;
+
+/**
+ *  The space between the elements. Default == 12
+ */
+@property (nonatomic, assign) CGFloat spaceBewteenIconAndText;
```
montehurd: Typo.

montehurd:
Getting crash in settings after tapping on "Wikipedia Zero FAQ".

Wikipedia/Code/WebViewController.m
```diff
@@ -296,6 +296,12 @@ - (WMFArticleFooterView*)footerLicenseView {
 }
 
 - (void)addFooterView {
+    if(!self.article){
+        return;
+    }
+    if([self.article.title isNonStandardTitle]){
```
montehurd: Should this be happening in the same place where we check if an article is a main page before showing a footer?

Wikipedia/Code/WMFArticleContainerViewController+TOC.swift
```diff
@@ -50,7 +50,11 @@ extension WMFArticleViewController : WMFTableOfContentsViewControllerDelegate {
     }
 
     public func tableOfContentsArticleSite() -> MWKSite {
-        return self.articleTitle.site
+        if(self.articleTitle.isNonStandardTitle()){
+            return MWKSite.init(language: "en")
```
montehurd: Interesting... so it's not that we're using localized strings so we can't rely on that english fallback.

So for now we assume all non-standard TOC sites are English... which works because at the moment only the FAQ uses the TOC in this way?

montehurd:
@coreyfloyd Hmm I think I found a blocker:

When I try to login in to the app it says a server with the specified host name could not be found.

Wikipedia/Code/WebViewController.m
coreyfloyd: It needs to be here also. Because this method is sometimes called seperately

Wikipedia/Code/WMFArticleContainerViewController+TOC.swift
coreyfloyd: Yeah. Basically. The only thing we are doing this for now is mediawiki.org sites. I don't like it but really we just need to rip out mwksite and mwktitle

Wikipedia/Code/WebViewController.m
montehurd: Gotcha.

Wikipedia/Code/WMFArticleContainerViewController+TOC.swift
montehurd: True.

montehurd:
We should probably suppress edit buttons on the FAQ too. I don't think edits will work - may crash?

PR #679, opened 2016-04-27
Nil out the username and password when auto login fail

coreyfloyd:
@montehurd this isn't working - but should. 

What I mean is that I can change my password, but wikipedia still lets me log in with the old password… very weird.

You want to take a look?


montehurd:
Checking...

montehurd:
@coreyfloyd I think this is good to go.

Wikipedia/Code/SessionSingleton.m
```diff
@@ -201,6 +201,14 @@ - (void)fetchFinished:(id)sender
                 [self cloneSessionCookies];
             }
             break;
+            case FETCH_FINAL_STATUS_FAILED: {
+                // If autoLogin fails the credentials need to be cleared out if they're no longer valid so the
+                // user has an indication that they're no longer logged in.
+                if (error.code == LOGIN_ERROR_PASSWORD_WRONG){
```
coreyfloyd: Two things:

1. You must check the domain and the error code. If you get back an error from NSURLConnection that happens to have the same error code as LOGIN_ERROR_PASSWORD_WRONG, then you will inadvertently clear the user name.

2. Looks like we should clear this out for all the errors you defined above, right?  All of the errors you defined seem to suggest the username or password needs to be fixed to properly log in.

coreyfloyd:
@montehurd all looks good except for my comments… should be quick. - Thanks again for picking this up.

Wikipedia/Code/SessionSingleton.m
montehurd: Regarding clearing out credentials, one by one...

`LOGIN_ERROR_NAME_REQUIRED`
Auto login isn't triggered unless there's a name.

`LOGIN_ERROR_NAME_ILLEGAL`
If the user enters an illegal name, it seems it would be easier to just require they edit the name - i.e. remove a the disallowed character etc, than be forced to completely re-enter the name.

`LOGIN_ERROR_NAME_NOT_FOUND`
If "corey" is taken if we blank it out they'd have to completely re-enter "corey2" instead of just appending "2".

`LOGIN_ERROR_PASSWORD_REQUIRED`
Not good to blank out the user name because they forgot to enter a pwd.

`LOGIN_ERROR_PASSWORD_WRONG`
Not good to blank out the user name because they entered wrong pwd.

`LOGIN_ERROR_THROTTLED`
Not good to blank out the user name because their connection was throttled.

`LOGIN_ERROR_BLOCKED`
Doesn't make any difference either way.

coreyfloyd:
@montehurd all of those explanations make sense to why the error occurred… but most all of them still mean we shouldn't keep the bad username or password in the keychain. The saved credentials will never work so we shouldn't keep them. 

More importantly unless we nil out the credentials - settings will still say the user is logged in, which is a lie. And the entire point of the ticket.

coreyfloyd:
Also we should never be worried about erasing credentials that do not work. Obviously the app can't use them to login so what good are they?

Wikipedia/Code/SessionSingleton.m
montehurd: @coreyfloyd 
> Also we should never be worried about erasing credentials that do not work. Obviously the app can't use them to login so what good are they?

Hmm... but we don't store bad credentials, right? So we should never even be trying to autologin with anything but credentials which worked the only way a credential goes from `good` to `bad` is the user's password changed or they were blocked. Does this make sense?

montehurd:
@coreyfloyd 
> Also we should never be worried about erasing credentials that do not work. Obviously the app can't use them to login so what good are they?

iirc we only store credentials which have actually worked so in this case we're only trying to auto login with good credentials, which can only become bad credentials in one of 2 ways - the user either changes their password or they're blocked.

coreyfloyd:
Right. They worked at one time but no longer. The server told us so. In any case he only way for the user to fix it is to log out of the app and login again


And the bigger point, to close his ticket you must not display the user has logged in on the settings screen when they have failed to login. 

That isn't true unless you nil out bad credentials. No matter why hey are bad.

PR #680, opened 2016-04-28
Fix Myanmar (Burmese) support.

montehurd:
https://phabricator.wikimedia.org/T131285

- [x] Bundle same font referenced by mobile web html - "[Tharlon-Regular.ttf](https://code.google.com/archive/p/tharlon-font/downloads)"
- [x] Bundle the font license file (see "OFL.txt" at the link above)
- [x] Also fixed Khmer! :) 

**BEFORE:**
before 1
before 2
before 3

**AFTER:**
after 1
after 2
after 3
after 4


montehurd:
@coreyfloyd If you can bundle the license file from the description above I think this is good to go. (I would have before I left but I have to head out in a minute)

Wikipedia/Code/MWKLanguageLinkController.m
```diff
@@ -35,7 +35,7 @@
     static NSArray* unsupportedLanguageCodes;
     static dispatch_once_t onceToken;
     dispatch_once(&onceToken, ^{
-        unsupportedLanguageCodes = @[@"my", @"am", @"km", @"dv", @"lez", @"arc", @"got", @"ti"];;
+        unsupportedLanguageCodes = @[@"am", @"dv", @"lez", @"arc", @"got", @"ti"];
```
coreyfloyd: Just to check: We support 2 new languages?  my and km?

coreyfloyd: nm - your comment from standup answered my question

PR #683, opened 2016-04-29
Move code files to code folder and uncrustify.

montehurd:


coreyfloyd:
@montehurd thanks for cleaning up all of this! I'm going to hold this until after the we ship the beta just in case the re-org causes some build issues.

PR #684, opened 2016-04-29
Fix crash bug on older devices

coreyfloyd:
Several fixes here all revolving around our image handling:

- No longer decode images on older devices (This is an optimization that improves performance but increases memory consumption) which have less than 1 GB of memory
- Don't inflate images into memory when the saved page fetcher is checking if they were cached
- No longer leave images that were cached by the saved page fetcher in the image cache
- Reduce the number of concurrent image downloads on older devices (only 4 instead of 6)
- Clear image cache when the app is backgrounded (Boyscout, but clears out the cache when we go to background)


Wikipedia/Code/WMFAppViewController.m
```diff
@@ -239,6 +239,7 @@ - (void)resumeApp {
 }
 
 - (void)pauseApp {
+    [[WMFImageController sharedInstance] clearMemoryCache];
```
montehurd: Should we do this in didReceiveMemoryWarning too?

PR #687, opened 2016-04-30
Fix universal links crash.

montehurd:
https://phabricator.wikimedia.org/T134028

Repro steps:
Tap wikipedia link in Safari.

### DIFF
--- a/Wikipedia/Code/MWKSite.m
+++ b/Wikipedia/Code/MWKSite.m
@@ -41,18 +41,22 @@ typedef NS_ENUM (NSUInteger, MWKSiteNSCodingSchemaVersion) {
         DDLogError(@"Can't form site from incomplete URL: %@", url);
         return nil;
     }
-    NSRange range    = NSMakeRange(hostComponents.count - 2, hostComponents.count - 1);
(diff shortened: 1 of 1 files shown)


Wikipedia/Code/MWKSite.m
```diff
@@ -41,18 +41,22 @@ - (MWKSite* __nullable)initWithURL:(NSURL* __nonnull)url {
...
-    NSRange range    = NSMakeRange(hostComponents.count - 2, hostComponents.count - 1);
-    NSString* domain = [[hostComponents subarrayWithRange:range] componentsJoinedByString:@"."];
 
     if ([url.host containsString:@"mediawiki"]) {
+        NSRange range    = NSMakeRange(hostComponents.count - 2, hostComponents.count - 1);
+        NSString* domain = [[hostComponents subarrayWithRange:range] componentsJoinedByString:@"."];
         return [self initWithDomain:domain language:nil];
-    } else {
-        [hostComponents removeObject:@"m"];
+    }else{
```
montehurd: Everything in this "else" case is the previous pre-crash logic.

PR #688, opened 2016-04-30
Fixed bug with universal link if settings had been left open.

montehurd:
https://phabricator.wikimedia.org/T134031

If you minimized the app while settings was onscreen, then went
to safari and opened a link it would open the app still showing
settings, and only when you closed settings would you see the
article from safari.


Wikipedia/Code/WMFAppViewController.m
```diff
@@ -362,6 +362,7 @@ - (BOOL)processUserActivity:(NSUserActivity*)activity {
             [[UIViewController wmf_sharedSearchViewController] setSearchTerm:[activity wmf_searchTerm]];
             break;
         case WMFUserActivityTypeArticle: {
+            [self dismissViewControllerAnimated:NO completion:NULL];
```
coreyfloyd: Looks like we are dismissing in all cases - mind DRYing it up and putting the dismiss before the case statement?

montehurd: Oh, ya!

PR #689, opened 2016-05-05
Fix feed jump

coreyfloyd:
Add code to prevent the jump from happening when a single error occurs
Change preview fetching back to GET to work around the POST body issue.


montehurd:
👍

PR #692, opened 2016-05-09
Fix random sort and request proper number of images/extracts.

montehurd:
https://phabricator.wikimedia.org/T134813

### DIFF
--- a/Wikipedia/Code/WMFRandomArticleFetcher.m
+++ b/Wikipedia/Code/WMFRandomArticleFetcher.m
@@ -66,14 +66,15 @@ NS_ASSUME_NONNULL_BEGIN
     //Sort so random results with good extracts and images come first and disambiguation pages come last.
     NSSortDescriptor* extractSorter  = [[NSSortDescriptor alloc] initWithKey:@"extract.length" ascending:NO];
     NSSortDescriptor* descripSorter  = [[NSSortDescriptor alloc] initWithKey:@"wikidataDescription" ascending:NO];
-    NSSortDescriptor* thumbSorter    = [[NSSortDescriptor alloc] initWithKey:@"thumbnailURL" ascending:NO];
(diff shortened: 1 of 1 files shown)


Wikipedia/Code/WMFRandomArticleFetcher.m
```diff
@@ -66,33 +66,34 @@ - (MWKSearchResult*)getBestRandomResultFromResults:(NSArray*)results {
     //Sort so random results with good extracts and images come first and disambiguation pages come last.
     NSSortDescriptor* extractSorter  = [[NSSortDescriptor alloc] initWithKey:@"extract.length" ascending:NO];
     NSSortDescriptor* descripSorter  = [[NSSortDescriptor alloc] initWithKey:@"wikidataDescription" ascending:NO];
-    NSSortDescriptor* thumbSorter    = [[NSSortDescriptor alloc] initWithKey:@"thumbnailURL" ascending:NO];
+    NSSortDescriptor* thumbSorter    = [[NSSortDescriptor alloc] initWithKey:@"thumbnailURL.absoluteString" ascending:NO];
     NSSortDescriptor* disambigSorter = [[NSSortDescriptor alloc] initWithKey:@"isDisambiguation" ascending:YES];
     NSSortDescriptor* listSorter     = [[NSSortDescriptor alloc] initWithKey:@"isList" ascending:YES];
-    results = [results sortedArrayUsingDescriptors:@[disambigSorter, listSorter, extractSorter, thumbSorter, descripSorter]];
+    results = [results sortedArrayUsingDescriptors:@[disambigSorter, listSorter, thumbSorter, descripSorter, extractSorter]];
```
montehurd: Extract length descriptor was moved to the end so it can act as a tie breaker for random items which have both an image and a wikidata description. 

Having it first was meaningless as the other sort criteria would only really take effect if 2 random results had extracts of the exact same length, which is unlikely.

PR #695, opened 2016-05-11
Convert CocoaPods to static libs and move frameworks to Carthage

coreyfloyd:
This is to resolve:
https://phabricator.wikimedia.org/T134805

Which is being tracked publicly in the iOS community here:
https://github.com/artsy/eigen/issues/1246

We are reducing the number of frameworks as a potential solution to the number of crashes.

~~Currently the unit test target is broken, I am posting now as I try to resolve. Carthage frameworks are not being loaded correctly on launch when running the unit tests. I filed a bug in the Carthage repo here:
https://github.com/Carthage/Carthage/issues/1299~~

There are some other small things in here:
- Changed the icon badging in beta to use the fastlane action (instead of our custom script) because we had issues with the script after the conversion 🤔
- Changed minimum deployment target to 8.3 - something hated it being 8.0 - and this isn't that big of a difference. 8.x users should all be on 8.4 actually.


orta:
Seeing as this is linked to our crash issue, I felt I should to be the bearer of bad news and say that this won't make a difference to https://github.com/artsy/eigen/issues/1246 - merely including one framework, or including swift would cause the crash to happen 

It will help against slow launches on enterprise builds with frameworks, but I don't want to see less experienced programmers cargo-culting this as a potential fix.

PR #696, opened 2016-05-11
Add custom URL scheme wikipedia to Info plist.

cromulentlabs:
Please see https://phabricator.wikimedia.org/T135048. It would be nice to be able to launch the Wikipedia app from other apps. Thanks.


montehurd:
looks simple enough :)

PR #698, opened 2016-05-13
Convert UIWebView -> WKWebView

coreyfloyd:
This pr moves over to WKWebView

It launches but the the javascript needs some work.

This may be useful but we can probably do better:
http://stackoverflow.com/questions/25792131/how-to-get-jscontext-from-wkwebview
(Maybe there are some native APIs?)

Note: WKWebView can not be loaded form a nib, so I had to move the web view instantiation to code.

(uhhh… sorry about the commit that says "temp" - I tried to rebase but it was too much trouble)


coreyfloyd:
@montehurd wanna give this a looksy and see if you can convert the js over to the WKWebView paradigm?

coreyfloyd:
Also its possible this may be of help as well:

https://github.com/XWebView/XWebView

montehurd:
@coreyfloyd cool i'll give it a go!

Wikipedia/Code/PreviewAndSaveViewController.storyboard
```diff
@@ -57,6 +57,14 @@
                                                     <constraint firstItem="LfM-01-aCF" firstAttribute="leading" secondItem="Fdd-cK-gLF" secondAttribute="leading" constant="20" symbolic="YES" id="xTZ-yd-BxN"/>
```
montehurd: I should have titled this commit "Revert **PreviewAndSaveViewController** storyboard changes to get back fickle constraints."

.gitignore
```diff
@@ -60,8 +60,6 @@ xcbaselines/
 
 # www
 node_modules/
-# TODO: remove this when assets is moved into build/
-Wikipedia/assets/
```
coreyfloyd: Why were we ignoring this?

montehurd: Not sure. It doesn't make sense we would ignore this directory - we need to add files occasionally to it. I don't think it will hurt anything to add it back.

Wikipedia/Code/WMFArticleViewController.m
```diff
@@ -1091,7 +1091,6 @@ - (nullable UIViewController*)previewingContext:(id<UIViewControllerPreviewing>)
     if (peekVC) {
         [[PiwikTracker wmf_configuredInstance] wmf_logActionPreviewInContext:self contentType:nil];
         self.webViewController.isPeeking = YES;
-        previewingContext.sourceRect     = [self.webViewController rectForHTMLElement:peekElement];
```
montehurd: This never worked anyway for web view link peeking. Not sure why.

Wikipedia/Code/WebViewController.m
```diff
@@ -123,12 +123,15 @@ - (NSString*)tableTransformJS {
 
 - (void)userContentController:(WKUserContentController*)userContentController didReceiveScriptMessage:(WKScriptMessage*)message {
     if ([message.name isEqualToString:@"peek"]) {
-        NSDictionary* peekDict = message.body;
-        self.peekURLString = peekDict[@"touchedElementURL"];
+        self.peekURLString = message.body[@"touchedElementURL"];
```
montehurd: This change wasn't related to the logging bits - just a small clean-up.

montehurd: This change wasn't related to the logging bits - just a small clean-up.

Wikipedia/Code/WebViewController.m
```diff
@@ -149,56 +303,73 @@ - (void)viewWillAppear:(BOOL)animated {
...
+    [self.KVOControllerNonRetaining unobserve:self.footerContainerView];
 }
 
-#pragma mark - Observations
+- (void)observeTopAndBottomNativeViewFrames {
+    [self.KVOControllerNonRetaining observe:self.headerView
+                                    keyPath:WMF_SAFE_KEYPATH(self.headerView, bounds)
+                                    options:NSKeyValueObservingOptionInitial
+                                      block:^(WebViewController* observer, UIView* view, NSDictionary* change) {
+
```
coreyfloyd: Might be good to put the logic in these handlers into well named methods.

Wikipedia/Code/WebViewController.m
```diff
@@ -447,17 +581,15 @@ - (void)scrollToFragment:(NSString*)fragment animated:(BOOL)animated {
         [self.webView.scrollView scrollRectToVisible:CGRectMake(0, 1, 1, 1) animated:animated];
     } else {
         if (!animated) {
-            [self.webView.wmf_javascriptContext.globalObject invokeMethod:@"scrollToFragment"
-                                                            withArguments:@[fragment]];
+            [self.webView evaluateJavaScript:[NSString stringWithFormat:@"window.wmf.utilities.scrollToFragment('%@')", fragment] completionHandler:nil];
```
coreyfloyd: A good idea is maybe to put all code that evaluates javascript into a category on WKWebView. (this goes for the the previous comment as well)

That way we can isolate js stuff from the webviewcontroller.

Wikipedia/Code/WebViewController.m
```diff
@@ -103,6 +92,166 @@ - (instancetype)initWithSession:(SessionSingleton*)aSession {
...
+            if (sectionIndex < [self.article.sections count]) {
+                [self.delegate webViewController:self didTapEditForSection:self.article.sections[sectionIndex]];
+            }
+        }else if(message.body[@"nonAnchorTouchEndedWithoutDragging"]){
+            [self referencesHide];
+        }
+    }
+}
+
+- (WKWebViewConfiguration*)configuration {
```
coreyfloyd: This can probably be moved into a category on WKWebViewConfiguration

Wikipedia/Code/WebViewController.m
```diff
@@ -452,18 +472,22 @@ - (void)addFooterView {
...
 }
 
 - (void)addFooterContainerView {
     self.footerContainerView.translatesAutoresizingMaskIntoConstraints = NO;
     [self.webView.scrollView addSubview:self.footerContainerView];
-    [self.footerContainerView mas_remakeConstraints:^(MASConstraintMaker* make) {
+    [self.footerContainerView mas_makeConstraints:^(MASConstraintMaker* make) {
         // lead/trail must be constained to webview, the scrollview doesn't define a width
         make.leading.and.trailing.equalTo(self.webView);
-        make.bottom.equalTo([[self.webView wmf_browserView] mas_bottom]);
```
montehurd: No more `wmf_browserView`! Now that's only used for peeking :)

joewalsh:
I've been looking into expanding the fix to eliminate the spinner for images that should be cached in memory by SDWebImage via WMFImageController. At first I thought it might be a schemaless vs schemed URL issue, but accounting for that didn't seem to fix the problem. Now I'm looking at whether SDWebImageCache is actually caching the images in memory or not.

montehurd:
The more I fiddle with this saved images issue the more I think we're in a tricky situation. @coreyfloyd the issue we're seeing is WKWebView's out of process structure means it doesn't work with NSURLCache ( see: https://github.com/ShingoFukuyama/WKWebViewTips ) so our `cachedResponseForRequest` method can't serve up cached images for offline mode. I've tried a couple workarounds ( i can share details monday ) but the one which works ( mostly ) involves downloading all images twice so they're present both in the Webkit cache and our SDWebImage cache ( for both native and web view cache coverage). 

I'm now experimenting with proxying via https://github.com/swisspol/GCDWebServer/. Basic idea is to set the base url to this app server, then seeing if the server is notified of relative url requests including images...

montehurd:
Ok I did a quick spike with GCDWebServer:

- set up GCDWebServer listening on port 8080
- changed the html image src paths to start with `localhost:8080` instead of `http://upload.wikimedia.org` so the bundled web server would receive these image requests
- when the server received the localhost image requests i grab the image url and restored `http://upload.wikimedia.org` so i have the original image url again
- used the GCDWebServer to then async fetch the actual image url's data
- respond to the original localhost request with the fetched data
- success! the images actually start appearing in the article again (they stopped appearing when `http://upload.wikimedia.org` was changed to `localhost:8080` in step 2)
- commented out the fetch and instead responded with a bundled image to see if cached images would show up in the web view if offline
- success! the bundled image that i forced appeared for every image in the article

joewalsh:
Great to see this working!

montehurd:
@joewalsh I pushed up the changes. I still need to add some tests and a couple optimizations for the string parsing in the morning, but you should be good to go for adding these ( https://www.cocoanetics.com/2012/03/reading-and-writing-extended-file-attributes/ ) bits.

joewalsh:
A couple of new concerns - 1. the image files are always being read from disk - I guess because we can't get at the in-memory image data, only UIImage objects? 2. for images cached directly via SDWebImage, the MIME type will be missing

joewalsh:
I investigated some ways we can get around the missing MIME type for images cached directly via SDWebImage. First method would involve our own our own `SDWebImageManager` subclass and our own `SDImageCache` subclass. It would also involve overriding some methods not publicly exposed in the interfaces for those files. Second idea I had would be to cache everything for the proxy into `NSURLCache` completely separately from `SDWebImage`. Downside of this is double caching.

joewalsh:
Ah- but I did just find `SDWebImageDownloaderUseNSURLCache` - so it might be possible to cache everything to `NSURLCache` and have `SDWebImage` read from that as well rather than trying to write everything to `SDWebImage`'s cache

joewalsh:
@montehurd I implemented the changes you mentioned in IRC to address srcset going away ^. Let me know how it looks

Wikipedia/assets/index.js
```diff
@@ -703,44 +703,32 @@ function getStretchRatio(image){
...
-            image.src = srcsetDict[largestSrcsetDictKey];
-
-            if(enableDebugBorders){
-                image.style.borderWidth = '10px';
-            }
-        }
+			var pathComponents = src.split("/");
+			var filename = pathComponents[pathComponents.length - 1];
+			var sizeRegex = /^[0-9]+(?=px-)/;
+			var sizeMatches = filename.match(sizeRegex);
```
joewalsh: should I also check for /thumb/ in the full path in addition to matching `^[0-9]+px-` on the filename?

joewalsh:
I'm noticing that the WMFArticle images have the non-scaled URLs, which causes it to require a network load when tapping an image to view in the gallery rather than using the cache. For example, the article had an image at `/640px-`, the scaling changed that to `/686px-`, but the gallery is looking for the `/640px` which wasn't loaded

PR #700, opened 2016-05-19
Return revision id everywhere we have a MWKSearchResult.

montehurd:
https://phabricator.wikimedia.org/T135752

@coreyfloyd Revision IDs!


coreyfloyd:
@montehurd thanks!

PR #702, opened 2016-05-22
Save the section id of the "continue reading" article.

montehurd:
@coreyfloyd 
With this PR the wmf_openArticleTitle will have its "fragment" set to whatever the WebViewController's currently onscreen section was. This fragment can then be used later for things like calling WebViewController's "scrollToFragment:" method. The web view controller already scrolls to a fragment if one is present in an MWKTitle when an article is opened iirc.

Note: **MWKTitle's *fragment* == MWKSection's *anchor* == that section's HTML div *id*.** 
(We may want to rename one of the first two, probably *anchor* should be changed to *fragment*, in the future. I believe *fragment* is standard url terminology and that *anchor* comes from *mobileview* api naming... iirc.)

This approach moves the "saving" of the "continue reading" title to the WebViewController. 

The following assumptions are made:
- a web view controller probably only scrolls if it is currently onscreen
- we need to record when the onscreen article's current section has changed, i.e. after scroll
- the web view can be scrolled either by the user or the TOC


Wikipedia/Code/WebViewController.m
```diff
@@ -133,6 +133,9 @@ - (void)viewDidLoad {
     [self displayArticle];
 }
 
+NSString* const WMF_UIScrollViewDidEndDeceleratingNotification = @"_UIScrollViewDidEndDeceleratingNotification";
```
montehurd: Need to observe both of these so we update both when TOC scrolls and when user scrolls the web view directly.

Wikipedia/Code/WebViewController.m
```diff
@@ -147,6 +150,34 @@ - (void)viewWillAppear:(BOOL)animated {
...
+
+- (void)saveOpenArticleTitleWithCurrentlyOnscreenFragment {
+    [[NSUserDefaults standardUserDefaults] wmf_setOpenArticleTitle:[self articleTitleWithCurrentlyOnScreenFragment]];
+}
+
+- (MWKTitle*)articleTitleWithCurrentlyOnScreenFragment {
+    return
+        [[MWKTitle alloc] initWithSite:self.article.title.site
+                       normalizedTitle:self.article.title.text
+                              fragment:[self currentVisibleSection].anchor];
```
montehurd: This will need to be changed a bit for the WKWebView pr, but the same observer approach should carry over.

Wikipedia/Code/MWKTitle.m
```diff
@@ -145,8 +145,8 @@ - (NSUInteger)hash {
 
 // Need to specify storage properties since text & site are readonly, which Mantle interprets as transitory.
 + (MTLPropertyStorage)storageBehaviorForPropertyWithKey:(NSString*)propertyKey {
-    #define IS_MWKTITLE_KEY(key) [propertyKey isEqualToString : WMF_SAFE_KEYPATH([MWKTitle new], key)]
-    if (IS_MWKTITLE_KEY(text) || IS_MWKTITLE_KEY(site)) {
+#define IS_MWKTITLE_KEY(key) [propertyKey isEqualToString : WMF_SAFE_KEYPATH([MWKTitle new], key)]
+    if (IS_MWKTITLE_KEY(text) || IS_MWKTITLE_KEY(site) || IS_MWKTITLE_KEY(fragment)) {
```
montehurd: Saves the fragment.

Wikipedia/Code/WebViewController.m
coreyfloyd: Can't we just do it in viewwilldissapear or subscribe to appwillresignactive?

Wikipedia/Code/WebViewController.m
```diff
@@ -147,6 +150,34 @@ - (void)viewWillAppear:(BOOL)animated {
                                                object:nil];
     // should happen in will appear to prevent bar from being incorrect during transitions
     [self updateZeroState];
+
+    @weakify(self);
+    void (^ saveOpenArticleTitleBlock)(NSNotification*) = ^void (NSNotification* notification) {
```
coreyfloyd: Or maybe I don't understand why did you change the logic to happen in viewdidappear instead of viewwilldissapear?

Basically I was thinking this would just be an extra piece of information on the history item. You save the scroll position and the current section title. If you want to use the fragment of the mwktitle that seems like a good idea, but I'm not sure why the timing of the logic needs to change. 

Basically why can't we just do your new logic in viewwilldissapear?

Wikipedia/Code/WebViewController.m
montehurd: @coreyfloyd viewwilldissapear doesn't do what we need, for instance - it gets called when you tap a link which causes a new view to be pushed over the old one, and we don't want to record the position of the old article in this case because the last read article is now the new one which was just pushed. Additionally, viewwilldissapear isn't called when the app resigns or is killed afaict.

Wikipedia/Code/WebViewController.m
montehurd: Can't use viewdidappear - the user hasn't scrolled yet.

montehurd: Checking on appwillresignactive...

montehurd: @coreyfloyd updated! see latest commit :)

coreyfloyd:
@montehurd thanks

montehurd:
@coreyfloyd np! after we merge this i'm gonna merge the TWN updates from last week so we can get this weeks TWN changes.

Wikipedia/Code/WebViewController.m
```diff
@@ -147,6 +147,30 @@ - (void)viewWillAppear:(BOOL)animated {
...
+        @strongify(self);
+        if (self.navigationController.topViewController == self.parentViewController) { // Ensure only the topmost article is recorded.
+            [self saveOpenArticleTitleWithCurrentlyOnscreenFragment];
+        }
+    };
+
+    [[NSNotificationCenter defaultCenter] addObserverForName:UIApplicationWillResignActiveNotification
```
coreyfloyd: @montehurd you have to retain the object that is returned by `addObserverForName`. But honestly I would just use the normal selector API instead. The block based API is pretty wonky.

Sorry I didn't catch this earlier.

Once you make that change this is good to go.

montehurd:
@coreyfloyd Fixed! :)

montehurd:
@coreyfloyd I'm gonna merge this in - I won't make a habit of self-merging but I need some of these bits on the WKWebView PR and you said it was good to go after changing to the normal observer selector api, which I did.

PR #705, opened 2016-05-31

niedzielski:


Wikipedia/Localizations/ksh.lproj/Localizable.strings
```diff
@@ -287,8 +287,10 @@
 "nearby-distance-label-miles" = "$1 Meile";
 "nearby-pull-to-refresh-prompt" = "Trick zom neu Aanzeije";
 "nearby-pull-to-refresh-is-refreshing" = "Neu aam Lahde";
-"nearby-loading" = "Loading nearby articles...";
-"nearby-loaded" = "Nearby articles loaded";
+// Fuzzy
+"nearby-loading" = "Loading nearby articles...<!-- https://phabricator.wikimedia.org/T135832 -->";
```
niedzielski: @montehurd, I wasn't sure if you wanted these comments included or not but figured they needed to be communicated. If you don't want them, please feel free to add another commit or make the change in TWN and I'll resync and force push.

montehurd:
@joein3d looks like this is passing tests... wanna give it a look-over?

Wikipedia/Localizations/ur.lproj/Localizable.strings
```diff
@@ -101,7 +101,7 @@
 "search-recent-clear-confirmation-sub-heading" = "اس عمل کو واپس نہیں کیا سکتا!";
 "search-recent-clear-cancel" = "منسوخ";
 "search-recent-clear-delete-all" = "سبھی کو حذف کریں";
-"search-result-redirected-from" = "%s سے رجوع مکرر ہوا ہے";
```
montehurd: @joewalsh 
I also logged in to https://translatewiki.net and changed it there. 

If a change is tiny like this one usually there's no issue, but sometimes we need to chat with the volunteer translators before undoing or modifying their translations. You can see an example of such a discussion I had on this translator's talk page: (i'm `mhurd` on the talk page)
https://translatewiki.net/wiki/User_talk:Purodha

joewalsh:
I found a potential truncation issue, but that could be filed separately. Here's a screenshot:

screen shot 2016-06-01 at 6 35 50 pm

 Otherwise looks good and can be merged.

montehurd:
@joewalsh ooh good catch! ya prob fix that separately... i bet it's an issue on any lang where that string is long :)

PR #706, opened 2016-06-02
T133933: Replace MWKSite and MWKTitle with NSURLComponents and app specific categories

joewalsh:
Opening this as WIP to get feedback, ensure I'm headed down the right path


joewalsh:
do we still want/need this check for mediawiki domains?

```    if ([url.host containsString:@"mediawiki"]) {
        NSRange range    = NSMakeRange(hostComponents.count - 2, hostComponents.count - 1);
        NSString* domain = [[hostComponents subarrayWithRange:range] componentsJoinedByString:@"."];
        return [self initWithDomain:domain language:nil];
    } ```

**Comment by joewalsh** [comment]
also should I match the `wmf_propertyName` naming convention instead of the `WMFPropertyName` style I have?

**Comment by montehurd** [comment]
> do we still want/need this check for mediawiki domains?

I think so. Looks like it was added here https://github.com/wikimedia/wikipedia-ios/commit/b4174a319cc6 but I can't recall what part of the app needed it... @coreyfloyd do you remember?

**Comment by joein3d** [comment]
Is the idea to re-work MWKSite and MWKTitle to use new categories on NSURLComponents or to eliminate both classes entirely?

**Comment by joewalsh** [comment]
Is the idea to re-work MWKSite and MWKTitle to use new categories on NSURLComponents or to eliminate both classes entirely?

**Comment by montehurd** [comment]
Definitely be nice to eliminate MWKTitle and just pass around NSURLs. Many (most?) places we're accessing MWKSite are in relation to a MWKTitle, but I'm not sure that all of them are... so I'm unsure about completely eliminating MWKSite for now.

**Comment by joewalsh** [comment]
@coreyfloyd @montehurd I'm at the point where MWKSite and MWKTitle work entirely on the new categories. Is it worth merging this into develop before going down the path of removing the model objects?

**Comment by joewalsh** [comment]
@coreyfloyd - I fixed the issue with the help and feedback page introduced by these changes. However, I found a failed assertion that also fails on develop branch.

When tapping on one of the animated gifs on the help and feedback page, the assertion at `MWKImageInfoFetcher.m:116` fails:

```
NSAssert(site.language.length, @"Site must have a non-empty language in order to send requests: %@", site);
```

it seems like this assertion should be removed

**Comment by joewalsh** [comment]
Ended up going back to something closer to the original logic for host parsing to handle the edge cases. Split the domain into components separated by `.` check the count, handle the various cases from there.

File: WikipediaUnitTests/Code/MWKTitleTests.m [hunk 1]
```diff
@@ -48,7 +48,6 @@ - (void)testSimple {
     MWKTitle* title = [MWKTitle titleWithString:@"Simple" site:site];
     XCTAssertEqualObjects(title.text, @"Simple", @"Text form is full");
     XCTAssertNil(title.fragment, @"Fragment is nil");
-    XCTAssertEqualObjects(title.escapedFragment, @"", @"Fragment for URL is empty string");
```
coreyfloyd: Is the big change here is that we are saying that fragments can be nil where before they would only be empty strings?

coreyfloyd:
@joewalsh looks good… the only comment I made was about the empty fragment vs nil fragment which looks like a behavior change since we modified the unit tests.

This is probably fine. If you didn't already, it might be good to confirm that fragments aren't assumed to be non-nil somewhere in the code base. For instance some method might pass a fragment as a parameter to a method that does not accept nil values.

If we don't feel we can verify this then we may want to always return an empty string for a fragment.

coreyfloyd:
@montehurd want to give this a pass when you get a sec?

WikipediaUnitTests/Code/MWKTitleTests.m
joewalsh: This is just fallout from removing `escapedFragment` entirely - it's not used anywhere in the code, only this test. https://github.com/wikimedia/wikipedia-ios/pull/706/commits/b05102e9ca86b3e093ac1935aa1b68cfa403f64d
and
https://github.com/wikimedia/wikipedia-ios/pull/706/commits/5d1eb0bca59f3088d07b94e8fdfa99ba8e9635fb
were just code pruning

coreyfloyd:
👍

Wikipedia/Code/NSURL+WMFLinkParsing.m
```diff
@@ -77,7 +77,7 @@ - (NSString*)wmf_domain {
 
 - (NSString*)wmf_language {
     NSRange dotRange = [self.host rangeOfString:@"."];
-    if (dotRange.length == 1 && dotRange.location < 4 && dotRange.location > 1) {
+    if (dotRange.location != NSNotFound) {
```
montehurd: @joewalsh Good catch. `test` wiki would be another one. See the list here: https://en.wikipedia.org/wiki/List_of_Wikipedias

Wikipedia/Code/NSURL+WMFLinkParsing.h
```diff
@@ -12,9 +12,31 @@ NS_ASSUME_NONNULL_BEGIN
...
 
-+ (NSURL*)wmf_URLWithDomain:(NSString*)domain language:(NSString* __nullable)language title:(NSString*)title fragment:(NSString* __nullable)fragment;
+/**
+ * Initialize a new URL with a Wikimedia `domain`, `language`, `title` and `fragment`.
+ *
+ * @param domain        Wikimedia domain - for example: `wikimedia.org`.
+ *
+ * @param language      An optional Wikimedia language code. Should be ISO 639-x/IETF BCP 47 @see kCFLocaleLanguageCode - for exmaple: `en`.
+ *
+ * @param title         An optional Wikimedia title. for exmaple: `Main Page`.
```
montehurd: @joewalsh 
nitpick typo: `exmaple`

montehurd:
@coreyfloyd Looks good to me - I left a couple small comments. I'm going to kick the tires for a bit...

montehurd:
@joewalsh I only found one issue, but it looks like it's not a new one, so we can address as separate ticket if you prefer.

If you go to "test wiki > bird", near the bottom of the page I have these 3 links...

`[[:es:Plancton]]`

`[[:en:Plankton]]`

`[[:en:A]]`

The html output for this wiki text is as follows:
`<a href="//es.wikipedia.org/wiki/Plancton" class="extiw" title="es:Plancton">es:Plancton</a>`

`<a href="//en.wikipedia.org/wiki/Plankton" class="extiw" title="en:Plankton">en:Plankton</a>`

`<a href="//en.wikipedia.org/wiki/A" class="extiw" title="en:A">en:A</a>`

Strangely, only the third link works in the app. All three work on the mobile site, so the pages exist.

joewalsh:
@montehurd yeah if it's not a regression i'd say it should be a separate ticket.

Wikipedia/Code/NSURL+WMFLinkParsing.h
joewalsh: ah - will fix

coreyfloyd:
@joewalsh can you resolve the conflicts and we will get it merged

@montehurd can you file that ticket?

montehurd:
@coreyfloyd will do

montehurd:
@coreyfloyd @joewalsh 
I made a ticket: https://phabricator.wikimedia.org/T137839

joewalsh:
@coreyfloyd @montehurd fixed conflicts

joewalsh:
This should get merged after #715 and I'll resolve the conflicts.

coreyfloyd:
@joewalsh #715 is merged

PR #708, opened 2016-06-08
T136974: Fix "Crash" option in settings on debug builds

joewalsh:
The HockeyApp App ID wasn't getting set in debug builds. This fixes the build script that sets that Info.plist property. Also needed to add placeholder values for the Piwik config information.

### DIFF
--- a/Wikipedia.xcodeproj/project.pbxproj
+++ b/Wikipedia.xcodeproj/project.pbxproj
@@ -6358,7 +6358,7 @@
 				GCC_PREPROCESSOR_DEFINITIONS = (
 					"$(inherited)",
 					"COCOAPODS=1",
-					"PIWIK_ENABLED=1",
+					"PIWIK_ENABLED=0",
 				);
 				GCC_TREAT_WARNINGS_AS_ERRORS = NO;
 				INFOPLIST_FILE = "Wikipedia/Wikipedia-Info.plist";
@@ -6492,7 +6492,7 @@
 				GCC_PREPROCESSOR_DEFINITIONS = (
 					"$(inherited)",
 					"COCOAPODS=1",
-					"PIWIK_ENABLED=1",
+					"PIWIK_ENABLED=0",
 				);
 				GCC_TREAT_WARNINGS_AS_ERRORS = NO;
 				INFOPLIST_FILE = "Wikipedia/Wikipedia-Info.plist";
@@ -6764,7 +6764,7 @@
 				GCC_PREPROCESSOR_DEFINITIONS = (
 					"$(inherited)",
 					"COCOAPODS=1",
-					"PIWIK_ENABLED=1",
+					"PIWIK_ENABLED=0",
 				);
 				GCC_TREAT_WARNINGS_AS_ERRORS = NO;
 				INFOPLIST_FILE = "Wikipedia/Wikipedia-Info.plist";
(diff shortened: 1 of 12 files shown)


coreyfloyd:
@joewalsh we weren't sending any reports to hockey for debug builds intentionally - we just didn't think we needed it… No problem if you want to turn it on though for testing.

Also… if you have VPN access I can get you setup on the Jenkins server so you can see where we are injecting those strings.

joewalsh:
@coreyfloyd it looked like the strings are in user defined build settings and then are injected via scripts/process_plist.sh - is that correct?

coreyfloyd:
@joewalsh was the ticket for debug builds or beta builds? I assumed @reetssydney meant beta builds… but she also runs the app locally so it could be debug as well

coreyfloyd:
@joewalsh Yep - the other part is that they are injected as environment variables by Jenkins during builds.

We do that so we don't commit the keys to the repo since its open source 😉

joewalsh:
Ah - OK. These commits might break that depending on what the environment variables are named

Wikipedia.xcodeproj/project.pbxproj
```diff
@@ -6373,6 +6373,8 @@
 				RUN_CLANG_STATIC_ANALYZER = NO;
 				SWIFT_OBJC_BRIDGING_HEADER = "Wikipedia/Code/Wikipedia-Bridging-Header.h";
 				WMF_HOCKEYAPP_IDENTIFIER = 2295c3698bbd0b050f257772dd2bdbb2;
```
joewalsh: So these values that are user defined settings are bogus? Or just the other ones need to be secure? Also... if they're being written to the plist, anyone can read them. Maybe we're talking about different config vars...

coreyfloyd:
@joewalsh yeah - the original names in scripts/process_plist.sh are the ones on the Jenkins machine. So you just may need to update your changes to use those keys based on the configuration

Wikipedia.xcodeproj/project.pbxproj
coreyfloyd: Hmm… thats our beta one. The production one really needs to be more secure, but really this shouldn't be there either.

Wikipedia.xcodeproj/project.pbxproj
```diff
@@ -6819,6 +6825,8 @@
 				RUN_CLANG_STATIC_ANALYZER = NO;
 				SWIFT_OBJC_BRIDGING_HEADER = "Wikipedia/Code/Wikipedia-Bridging-Header.h";
 				WMF_HOCKEYAPP_IDENTIFIER = 5d80da08a6761e5c6456736af7ebad88;
```
joewalsh: this production one?

coreyfloyd: well this is going to be a problem…

coreyfloyd: 🤕

reetssydney: Hi, part (b) not seeing crash dialogue was happening on the beta; though
once I deleted and reinstalled per Joe's comment earlier the
dialogue appeared again.

Part (a) where none of the tweak settings were working was on my local.

joewalsh:
@coreyfloyd here's where the keys were added: https://github.com/wikimedia/wikipedia-ios/commit/5cfbb50761512d90e1245090c45589720c082cd1 for issue https://phabricator.wikimedia.org/T101269 - does it make sense to keep them?

coreyfloyd:
@joein3d hmmm… I guess at this point we can just keep them in. The better thing may be to remove the logic from Jenkins that injects them.

So I'm good here.

👍 

@montehurd ?

montehurd:
Seems fine 👍

joewalsh:
@coreyfloyd when we do remove the Jenkins injection, we can switch back to the original version of this fix (before the revert commits)

PR #709, opened 2016-06-10
Add authmanager support

coreyfloyd:
This patch adds authmanager support - basically adds logic to support the new login and account creation flows.

The new flow is:
1. Add an extra call before beginning an auth process to detect if the auth manager is available (This call also returns a captcha url in the case of account creation)
2. Request tokens
3. Attempt login / account creation (without captcha)
4. If the account creation fails, then present the captcha
5. To request a new captcha, just perform steps 1 and 2 again
6. Repost account creation with captcha string

This flow is… messy. The old logic must be maintained until the new authmanager is deployed to all wikis. The auth logic already need refactored and it didn't make much sense to refactor it now until we can pull out the old logic.

There is a ticket for the general refactor and cleanup of the auth viewcontrollers and networking here:
https://phabricator.wikimedia.org/T132700

(but apparently this may be now? http://www.gossamer-threads.com/lists/wiki/wikitech/717118?search_string=AuthManager%20getting%20enabled%20in%20production;#717118).

To finish this up:
- [x] Test against production (we were black listed at the office after too many accounts created)
- [x] Verify banner presentations during login/account creation


coreyfloyd:
@montehurd here is the work from yesterday… I'll have a few minutes to talk before my slew of meetings.

PR #710, opened 2016-06-10
Use desktop domain for non Zero rated requests to restore img widening.

montehurd:
https://phabricator.wikimedia.org/T137595

Also fixes some article image taps not loading in gallery.

Android has switched to do same. Issue arose because "m." endpoint
html now is missing img tag "srcset" for some(most?) images. The
desktop endpoint is not affected however and still has srcset.


joewalsh:
👍

PR #714, opened 2016-06-13
Exclude image cache from iCloud backup

joewalsh:


coreyfloyd:
@joewalsh is the flag recursive? 

If not, is that reason to mark each individual article path as opposed to just excluding the entire `articles` directory?

One thought I had was that if occasionally items in the document directories can have that flag reset is that maybe we want to reset the flag each time the MWKDataStore is inited - so that may be easier if we only make the root directory.

Thoughts?

Also so you know any way to test if an item is removed from backup after we flag it?

joewalsh:
@coreyfloyd do you mean the root directory `Data/` that MWKDataStore uses or the root directory for articles for a given site `Data/sites/wikipedia.org/en/articles`? I had originally set it up to mark `Data/` on every init but changed it after re-reading the ticket that specified only articles.

coreyfloyd:
@joein3d I think Data contains other things like the saved pages… but would just `Data/sites` work for excluding just articles?

joewalsh:
@coreyfloyd it looks like `Data/sites` would work, only articles are in there now. There is the abstraction for `sitePath:` so if MWKDataStore is ever expanded, someone might put extra stuff in there.

coreyfloyd:
@joewalsh ok we should be fine for now with that - we would be moving away from this persistence model anyways - at least we won't make any changes to it before we move to a database.

If you want to add a comment about that as well it would probably be good.

joewalsh:
Ok, excluding pathForSites on every init

coreyfloyd:
👍

PR #715, opened 2016-06-14
T135960:Searching does not does not do a proper sanitization against "%" signs in strings

huayu0723:
Titles containing special characters have been well processed(unescaped). The issue is caused by unescaping again on a unescaped string.
The solution is to cancel this unnecessary unescape.


coreyfloyd:
@huayu0723 could you merge in the develop branch?

Just a heads up, we work on develop which can be significantly out of sync with master. So when you submit a PR, its easiest to branch off of develop rather than master. (We need to update our contribution instructions for that - sorry!)

coreyfloyd:
Hopefully this change will help:

https://github.com/wikimedia/wikipedia-ios/pull/717

joewalsh:
@huayu0723: Thanks for submitting this fix! Unfortunately, altering MWKTitle in this manner breaks other areas of the app. An alternative fix that won't break existing functionality would be to check where normalized titles are being passed into `[MWKTitle initWithString: site:]` and alter those calls to use `- [MWKTitle initWithSite: normalizedTitle: fragment:]`

huayu0723:
@coreyfloyd I confirmed that I created T135960 by branching the develop. For me this PR targets develop instead of master. Check it:
"huayu0723  wants to merge 1 commit into wikimedia:develop from huayu0723:T135960"
Do you know where I might be wrong?  Sorry for  inconvenience:(

huayu0723:
@joewalsh That's helpful!

joewalsh:
👍 looks great, thanks!

PR #716, opened 2016-06-14
Memory optimization/T125156

joewalsh:
Biggest win here is no longer stuck with every WMFArticleImageGalleryViewController created & associated views and images still in memory.

Also implemented suggestions from T125156- except we can’t actually clear the WebView cache anymore.


coreyfloyd:
@joewalsh just curious… did you happen to glance at what the savings in memory was?

PR #718, opened 2016-06-15
T124847 include about this article in table of contents if it's present

joewalsh:


coreyfloyd:
@montehurd can you take a look when you get a minute?

PR #722, opened 2016-06-16
T135243 add crash protection for nil language

joewalsh:
this is the most basic fix - it will unfortunately result in a blank space where the language name should be. An alternative fix would be to add an additional localization for a generic “Top Read on Wikipedia” and use that in this case. Or, figure out why language is nil in the first place.

### DIFF
--- a/Wikipedia/Code/WMFMostReadSectionController.m
+++ b/Wikipedia/Code/WMFMostReadSectionController.m
@@ -144,7 +144,13 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (NSAttributedString*)headerTitle {
     // fall back to language code if it can't be localized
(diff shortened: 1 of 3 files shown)


coreyfloyd:
@joewalsh how about we create an alternate localized string that doesn't take a parameter? Like just "Top Read" and use that instead?

Note that when you add a new string in the english strings file, you need to add a description in the "qqq" strings file. (This is used to gather translations from TranslateWiki - something @montehurd or I can show you later)

joewalsh:
@coreyfloyd - monte showed me this the other day. i made the change ^

montehurd:
👍

coreyfloyd:
👍 

@joewalsh when you get a minute - would you mind filling a ticket to investigate the root issue as well? Would like to keep track of that.

joewalsh:
@coreyfloyd https://phabricator.wikimedia.org/T137993

PR #724, opened 2016-06-16
push all HTML and image loading through our proxy server

joewalsh:


coreyfloyd:
@montehurd can you put this through its paces as well?

I'm going to test on the device

coreyfloyd:
@joewalsh I verified the content loaded on the device so we are good there - just fix up whatever @montehurd found and we are good. 👍

montehurd:
@joewalsh hey nice work! i kicked the tires on this some today and fixed one tiny issue. i'll test it a bit more tomorrow and let u know if i find anything.

joewalsh:
@montehurd thanks for the fix! i made one more minor tweak to my original code - to use `pathWithComponents` rather than `stringWithFormat`

Wikipedia/Code/AppDelegate.m
```diff
@@ -103,10 +104,28 @@ - (BOOL)application:(UIApplication*)application didFinishLaunchingWithOptions:(N
...
+    NSURL *documentsURL = [[[NSFileManager defaultManager] URLsForDirectory:NSDocumentDirectory inDomains:NSUserDomainMask] lastObject];
+    NSLog(@"%@", documentsURL);
 
     self.webServer = [[GCDWebServer alloc] init];
-
+    
+    // HAX: hardcoded for now to match WebView category
+    NSURL *assetsURL = [documentsURL URLByAppendingPathComponent:@"assets"];
+    
+    [self.webServer addDefaultHandlerForMethod:@"GET" requestClass:[GCDWebServerRequest class] processBlock:^GCDWebServerResponse *(GCDWebServerRequest *request) {
```
montehurd: @joewalsh any issue with maybe using "asyncProcessBlock" here?

joewalsh: @montehurd I verified that this occurs on the background, but with some new changes this will be using an asyncProcessBlock

montehurd: @joewalsh and maybe a path like in the other "asyncProcessBlock" call just to make it explicit?

montehurd: @joewalsh ah gotcha thx

joewalsh:
@montehurd @coreyfloyd the most recent commit consolidates everything into WMFProxyServer. This addresses a number of issues:

- The proxy server no longer accepts requests from anything but localhost so that others on your local network can't see your last browsed article or cached images
- The proxy server uses a shared secret path generated at runtime to prevent another process running locally from making a request to the proxy (though could this be sniffed?)
- The proxy server is assigned a port by the OS to avoid a conflict when 8080 is taken

montehurd:
@joewalsh wow very nice refactor! i'll dig in a bit more this afternoon.

Wikipedia/Code/AppDelegate.m
```diff
@@ -103,59 +97,9 @@ - (BOOL)application:(UIApplication*)application didFinishLaunchingWithOptions:(N
...
-            }
-        } else {
-            completionBlock(notFound);
-        }
-    }];
-
-    [self.webServer startWithPort:8080 bonjourName:nil];
+    
+    NSURL *documentsURL = [[[NSFileManager defaultManager] URLsForDirectory:NSDocumentDirectory inDomains:NSUserDomainMask] lastObject];
+    NSLog(@"%@", documentsURL);
```
coreyfloyd: We don't need this right?

joewalsh: it was there before I started editing, only without the documentsURL = part

Wikipedia/Code/WMFProxyServer.m
```diff
@@ -95,8 +95,8 @@ - (GCDWebServerAsyncProcessBlock)defaultHandler {
                        return;
                    }
 
-                   if (!([originalSrc hasPrefix:@"http"] || [originalSrc hasPrefix:@"https"])) {
-                       originalSrc = [@"http:" stringByAppendingString:originalSrc];
+                   if ([originalSrc hasPrefix:@"//"]) {
+                       originalSrc = [@"https:" stringByAppendingString:originalSrc];
```
joewalsh: @montehurd I made this change because it looked like on en.wiki it would just make those requests and be redirected to HTTPS. Also, security wise, this would prevent someone from sniffing which articles are being viewed based on image URLs. Might this be an issue on other wikis/should we build in a param that passes whether or not the wiki supports HTTPS?

Wikipedia/Code/AppDelegate.m
montehurd: @coreyfloyd I'd prefer we leave this (somewhere) so we can easily get the path. Since we removed it a few months ago I've had to go copy/paste it in at least 30 or 40 times when testing.

Wikipedia/Code/WMFProxyServer.m
montehurd: @joewalsh i think all the lang wiki's support https... should be ok.

coreyfloyd:
Just re-confirming that everything still works on the iOS 8 device

👍

coreyfloyd:
@montehurd are you good?

montehurd:
@coreyfloyd hmm did travis not run? i think the proxy string parsing tests should have failed - the string category went away, right?

coreyfloyd:
@montehurd yeah it only tests when targeted at develop this is not - @joewalsh can you run the tests locally and confirm they are passing?

joewalsh:
whoops - will update those tests

joewalsh:
They’re failing locally. Working on a fix

montehurd:
👍 when the tests are fixed

joewalsh:
tests are fixed 🎉

PR #728, opened 2016-06-21
T115157:Regression: As a user, the references popup should dismiss when I interact with the content

huayu0723:
Dismiss the reference panel once the user scrolls the article.


coreyfloyd:
👍

PR #729, opened 2016-06-21
Convert WMFImageController away from PromiseKit

joewalsh:


Wikipedia/Code/WMFImageController.swift
```diff
@@ -411,54 +390,39 @@ extension WMFImageController {
...
-    }
-    
-    /**
-     Objective-C-compatible variant of fetchImageWithURL(url:options:) returning an `AnyPromise`.
-     
-     - returns: `AnyPromise` which resolves to `WMFImageDownload`.
-     */
-    @objc public func fetchImageWithURL(url: NSURL?, options: SDWebImageOptions) -> AnyPromise {
-        return AnyPromise(bound: fetchImageWithURL(url, options: options))
+        return AnyPromise(bound: Promise<WMFImageDownload> { fulfill, reject in
```
coreyfloyd: Are we still using Promiskit here?

Also I think we no longer need to have an objective-c compatible version - the normal `fetchImageWithURL` should work now because it was only incompatible because promisekit.

coreyfloyd:
@joewalsh looking good so far - left a comment about some potential lingering promise kit stuff

Wikipedia/Code/WMFImageController.swift
joewalsh: @coreyfloyd yeah this is step 1 - wrap the new stuff in promise kit for obj-c. I did this to get it to build so that Wikimedia-swift would update and I could use the new Obj-c compatible versions

joewalsh: @coreyfloyd wanted to verify with you and monte that this is the pattern we want to use before updating all the obj-c to the new normal `fetchImageWithURL`

coreyfloyd:
@joewalsh is this ready to be reviewed after the conflicts are fixed?

joewalsh:
@coreyfloyd yeah

coreyfloyd:
👍 I'm good here seems straight forward, but extensive… c'est la vie

coreyfloyd:
@montehurd you want to have a look?

montehurd:
@coreyfloyd ya i'll jump on this after srcset pr.

montehurd:
👍

PR #730, opened 2016-06-21
Fix bug where you could not select a preferred language that was high…

coreyfloyd:
…er than 3rd on the list

Fixed Two issues:
- The preferred languages were not selectable
- The search VC did not update its UI even if one was selected


montehurd:
@coreyfloyd hmm... I must be missing something - I can neither repro the bug nor see any change when testing with this PR...

PR #731, opened 2016-06-22
check the NSURLCache if there's nothing in the SDWebImage cache in syncCachedImageWithURL

joewalsh:
https://phabricator.wikimedia.org/T138358


montehurd:
👍

PR #727, opened 2016-06-21
Hook up article image 3d touch.

montehurd:
https://phabricator.wikimedia.org/T131727

- [x] hook up article image 3d touch - peek/pops image into gallery
- [x] don't 3d touch an image if it's not in the gallery - i.e. tiny images for things like mathematical formulae
- [x] fix typo: WMFImageGalleryViewContoller - search for "Contoller"
- [x] call proper logging method
- [x] make explicit methods for getting controller configured for respective peek type - i.e. article and image
- [x] make peek element model properties be immutable, not strings if possible
- [x] added tests

I also pulled some methods out of the increasingly catch-all NSURL+Extras category into separate categories:
- **NSURL+WMFProxyServer** for bits related to the proxy server
- **NSURL+WMFQueryParameters** for bits related to url query parameters

There are now separate test files for these as well.

PR #735, opened 2016-06-22
Remove reliance on srcset

joewalsh:


joewalsh:
added benefit - the gallery works like it used to with blurry thumb ---> full resolution image.

however, maybe instead of downloading the full resolution original file we alter the thumb URL to request an appropriate size for the screen resolution?

Wikipedia/Code/WebViewController.m
```diff
@@ -151,21 +151,35 @@ - (void)userContentController:(WKUserContentController*)userContentController di
...
+            NSCParameterAssert(selectedImageURLString.length);
+            if (!selectedImageURLString.length) {
                 DDLogError(@"Image clicked callback invoked with empty URL: %@", message.body[@"imageClicked"]);
                 return;
             }
+            
+            NSURLComponents* selectedImageURLComponents = [NSURLComponents componentsWithString:selectedImageURLString];
+            for (NSURLQueryItem* item in selectedImageURLComponents.queryItems) {
+                if ([item.name.lowercaseString isEqualToString:@"originalsrc"]) {
+                    selectedImageURLString = item.value;
```
montehurd: In my image peek pr I refactored this bit - there's now a method I call here instead of checking the components.

montehurd: Here's the line:
https://github.com/wikimedia/wikipedia-ios/pull/727/files#diff-38568396230724f82d0bf5ec47cec4d8R180

I should have sync'ed up with you first but I fell down a bit of a rabbit hole :(

montehurd:
I'll test in the morning.

Wikipedia/Code/WKWebView+LoadAssetsHtml.m
```diff
@@ -25,8 +25,9 @@ - (void)loadHTML:(NSString*)string withAssetsFile:(NSString*)fileName scrolledTo
     }
     
     WMFProxyServer *proxyServer = [WMFProxyServer sharedProxyServer];
-
-    string = [proxyServer stringByReplacingImageURLsWithProxyURLsInHTMLString:string];
+    CGFloat padding = 32;// HAX: hardcoded based on measurement. can we get this value somewhere?
```
coreyfloyd: Hmmm… is this in the CSS or html somewhere?

Wikipedia/Code/MWKImage.h
```diff
@@ -46,6 +46,22 @@
 
 
 /**
+ * Width of the original image file. For example, if this image is a thumbnail, this is the width of the original image.
+ */
+@property (copy, nonatomic) NSNumber* originalFileWidth;
```
coreyfloyd: @joewalsh are we using these new properties anywhere?

I didn't see them used except for when we inject them. Is there an intended place that you think we will be using this information?

coreyfloyd: Are we handling the case of where previously serialized MWKImages won't have this information? Is there anything to do? Or any side affects to that?

joewalsh: I forgot to update the logic after making the change to separate properties. Here's that new change: 0a7a508a5100d798e46c06f67c79c3a5430f45b9

joewalsh: The answer now is yes we do use these new properties and it's OK for older serialized MWKImages to not have this information.

joewalsh:
Should I bring this logic from the js into the obj-c as well 
 
```
image.width >= 64 &&       
!image.hasAttribute('hasOverflowXContainer') &&
!utilities.isNestedInTable(image)
```

coreyfloyd:
@joewalsh maybe we could inject this constant as a parameter when we create the HTML so we can control it from Obj-C land?

coreyfloyd:
That way it is only in one place?

joewalsh:
@coreyfloyd good idea

montehurd:
Testing now...

montehurd:
Broken Image:

enwiki > barack obama > employment statistics > 3rd image

montehurd:
So far no other broken images... I've checked enwiki
- picasso
- tyrannosaurus
- goat
- mini
- barack obama
- toast
- tangent

montehurd:
@joewalsh 

Found two more images:

enwiki > fish > reproductive system > 2nd and 3rd images

These two should have been widened but weren't - their data-file-width is sufficient... hmm...

montehurd:
@joewalsh ^ I just checked and the issue above is not a regression as I see it was happening already - up to you if you think we should handle this separately.

Wikipedia/Code/WKWebView+LoadAssetsHtml.m
```diff
@@ -25,8 +25,31 @@ - (void)loadHTML:(NSString*)string withAssetsFile:(NSString*)fileName scrolledTo
...
+    NSUInteger windowPixelWidth = (NSUInteger)(MIN(self.bounds.size.width, self.bounds.size.height) * self.window.screen.scale);
+    
+    NSUInteger targetImageWidth = 320;
+    
+    if (windowPixelWidth < 640) {
+        targetImageWidth = 320;
+    } else if (windowPixelWidth < 800) {
+        targetImageWidth = 640;
+    } else if (windowPixelWidth < 1024) {
+        targetImageWidth = 800;
```
montehurd: I think 640 should be the cutoff. 

This image from the Obama article has a canonical size of 4800 x 6000, so it's a good example for an image that has lots of headroom for requesting higher resolutions...

On iPad (open these images in separate tabs to compare):

- **Before this PR**, this image was fetched at 340px width, and the entire page transfer size was **2.54 MB**:
pre change

- **This PR fetched at 640px width**, and entire page transfer size of **3.36 MB**:
this pr at 640

- **This PR fetched at 800px width**, and entire page transfer size of **4.10 MB**:
this pr at 800

640px looks great IMO and it's more detailed than the 340px we were formerly fetching.

iPhone 6 transfer sizes for Obama article:
 320  1.70 MB
 340  2.13 MB < Before this PR
 640  3.36 MB < After this PR
 800  4.10 MB
1024  6.48 MB
1280  7.77 MB
1920  11.7 MB
2560  18.9 MB
2880  20.7 MB

montehurd: Actually, now that I think of it, we should probably just use the same image scale logic we use to determine other image widths...

How about we add a method like the following to UIScreen+WMFImageWidth.m

    - (NSNumber*)wmf_articleImageWidthForScale {
        return @(self.wmf_maxScale * WMFImageWidthMedium);
    }

This would return 320 for 1x density screens, and 640 for 2x. Then in the future if we want to enable 3x we could just update wmf_maxScale and we'd get the change everywhere - in both articles *and* native code. This would also let us expose a user preference for low bandwidth by letting the user change wmf_maxScale, (should prob wait on adding that that for now).

joewalsh:
@montehurd thanks for the notes and for finding the busted images! I added the method to the UIScreen category as you suggested. I also fixed the SVG handling & broke down the larger parsing function into it's two main components. The fish images aren't being scaled because they have width=36 which is less than the threshold of 64. As you noted, this matches how the JavaScript scaler works. I'm a bit hesitant to alter that behavior because I'm not sure of the reason it was initially in there. In this case - it's clear the actual images are wide enough to be scaled based on data-file-width. If you know the reason for the 64 cutoff was just a rough estimate of file size, we could change it due to the presence of data-file-width.

Wikipedia/Code/WMFProxyServer.m
```diff
@@ -189,136 +189,122 @@ - (NSURL*)proxyURLForImageURLString:(NSString*)imageURLString {
...
+    NSMutableString *newHTMLString = [NSMutableString stringWithString:@""];
+
+    __block NSInteger location = 0;
+    [imageTagRegex enumerateMatchesInString:HTMLString options:0 range:NSMakeRange(0, HTMLString.length) usingBlock:^(NSTextCheckingResult * _Nullable imageTagResult, NSMatchingFlags flags, BOOL * _Nonnull stop) {
+        //append the next chunk that we didn't match on to the new string
+        NSString *nonMatchingStringToAppend = [HTMLString substringWithRange:NSMakeRange(location, imageTagResult.range.location - location)];
+        [newHTMLString appendString:nonMatchingStringToAppend];
+        
+        //get just the image tag contents - everything between <img and >
+        NSString *imageTagContents = [imageTagRegex replacementStringForResult:imageTagResult inString:HTMLString offset:0 template:@"$1"];
```
montehurd: Can imageTagContents ever be nil?

joewalsh: It's non-optional so it seems all it could be would be an empty string meaning the img tag looked like this: ``.

Wikipedia/Code/WMFProxyServer.m
```diff
@@ -189,136 +189,122 @@ - (NSURL*)proxyURLForImageURLString:(NSString*)imageURLString {
...
+        *stop = dataFileWidth > 0 && srcAttributeRange.location != NSNotFound && width > 0;
+    }];
+    
+    NSMutableString *newImageTagContents = [imageTagContents mutableCopy];
+    
+    
+    NSMutableArray *srcPathComponents = [[src pathComponents] mutableCopy];
+    NSString *resizedSrc = nil;
+    //check to see if we should change the image size being requested - if we know the original file width (dataFileWidth) and the width is present and greater than or equal to 64, and it's a thumbnail URL, we should request a larger size
+    if (dataFileWidth > 0 && (width == 0 || width >= 64) && srcPathComponents.count > 4 && [[srcPathComponents[srcPathComponents.count - 5] lowercaseString] isEqualToString:@"thumb"]) {
```
montehurd: I wonder if we should check for "/thumb/" anywhere in the path instead of at a specific location? Thoughts?

coreyfloyd: @montehurd at this time I'd try to scope any additional review to known problems and any further refinements we should file separately.

Wikipedia/Code/WMFProxyServer.m
```diff
@@ -189,136 +189,122 @@ - (NSURL*)proxyURLForImageURLString:(NSString*)imageURLString {
...
+            [newImageTagContents replaceCharactersInRange:srcAttributeRange withString:newSrcAttribute];
         }
     }
-    return [output componentsJoinedByString:@", "];
+    
+    
+    [newImageTagContents replaceOccurrencesOfString:@"srcset" withString:@"data-srcset-disabled" options:0 range:NSMakeRange(0, newImageTagContents.length)]; //disable the srcset since we put the correct resolution image in the src
+    
+    if (resizedSrc) {
+        [newImageTagContents appendString:@" data-image-resized=\"true\""]; //the javascript looks for this so it doesn't try to change the src again
```
montehurd: For future reference, I added `style="border-radius: 60px;"` to this string for testing. I also enabled `enableDebugBorders` in widenImages.js. That way you could easily see which native and js transforms took effect. Looked good.

montehurd:
👍

PR #736, opened 2016-06-23
T128646:Scroll to the top by pressing the Explore icon if on the Explore tab

huayu0723:
Add a key-value pair to user defaults, indicating the previous active tab index(WMFAppTabTypeExplore, WMFAppTabTypeSaved or WMFAppTabTypeRecent). It’s set as 0 when registered


Wikipedia/Code/WMFAppViewController.m
```diff
@@ -716,6 +718,12 @@ - (void)runDataMigrationIfNeededWithCompletion:(dispatch_block_t)completion {
 #pragma mark - UITabBarControllerDelegate
 
 - (void)tabBarController:(UITabBarController*)tabBarController didSelectViewController:(UIViewController*)viewController {
+    NSInteger prevActiveTabIndex = [[NSUserDefaults standardUserDefaults] integerForKey:PrevActiveTabIndex];
+    [[NSUserDefaults standardUserDefaults] setInteger:tabBarController.selectedIndex forKey:PrevActiveTabIndex];
```
coreyfloyd: Instead of saving this to user defaults, an alternate solution that doesn't require persistence would be to also implement:

```
- (BOOL)tabBarController:(UITabBarController *)tabBarController
shouldSelectViewController:(UIViewController *)viewController
```
https://developer.apple.com/library/ios/documentation/UIKit/Reference/UITabBarControllerDelegate_Protocol/index.html#//apple_ref/occ/intfm/UITabBarControllerDelegate/tabBarController:shouldSelectViewController:

You could capture the currently selected tab there, and if it is thew explore VC then you can scroll to the top - so that avoids using NSUserDefaults all together.

Wikipedia/Code/WMFAppViewController.m
```diff
@@ -716,6 +718,12 @@ - (void)runDataMigrationIfNeededWithCompletion:(dispatch_block_t)completion {
...
+    if (tabBarController.selectedIndex == WMFAppTabTypeExplore && prevActiveTabIndex==WMFAppTabTypeExplore) {
+        WMFExploreViewController *exploreViewController = (WMFExploreViewController *)[self exploreViewController];
+        [exploreViewController.tableView scrollToRowAtIndexPath:[NSIndexPath indexPathForRow:0 inSection:0] atScrollPosition:UITableViewScrollPositionTop animated:YES];
```
coreyfloyd: Generally we try not to touch the internals of another class and favor providing a specific method to accomplish a task - in this case it would be more encapsulated if we provided a public method like:

```
- (void)scrollToTop:(BOOL)animated;
```

And then we can let the explore view controller do the work as it seed fit.

huayu0723:
Thank you for your careful review!

coreyfloyd:
@huayu0723 no problem - thanks for the quick changes!

coreyfloyd:
@huayu0723 I'm also going to grant you contributor access to the repo so you don't need to maintain your own fork. Thanks again!

huayu0723:
@coreyfloyd I'll switch to the new repo soon. Thank you!

PR #738, opened 2016-06-24
remove extra border around about this article

joewalsh:


reetssydney:
Looks good on my Simulator for iPad 2, iPhone 4S and 6S.

PR #740, opened 2016-06-24
Add test for any upstream changes in image tag formatting.

montehurd:
- added quick test asking api.php for the html for a wikitext image string. it compares the response with a string of the response as of today, so we will be notified if ever the upstream image html changes - i.e. if they remove srcset or make other changes

- renamed the performance test and made its regex warmup string not use the full obama article html


WikipediaUnitTests/Code/ImageProxyParsingTests.m
```diff
@@ -200,12 +200,52 @@ - (void)testObamaArticleImageProxySubstitutionCount {
     assertThat(allObamaHTMLSplitOnImageProxy, hasCountOf(107));
 }
 
-- (void)testPerformanceExample {
+- (void)testPerformanceOfImageTagParsing {
     //initial run compiles regexes
-    [self.proxyServer stringByReplacingImageURLsWithProxyURLsInHTMLString:[self allObamaHTML] targetImageWidth:self.imageSize];
+    [self.proxyServer stringByReplacingImageURLsWithProxyURLsInHTMLString:@"Small sample string <img alt=\"Example.jpg\" src=\"//upload.wikimedia.org/wikipedia/en/thumb/a/a9/Example.jpg/20px-Example.jpg\" width=\"20\" height=\"22\" srcset=\"//upload.wikimedia.org/wikipedia/en/thumb/a/a9/Example.jpg/30px-Example.jpg 1.5x, //upload.wikimedia.org/wikipedia/en/thumb/a/a9/Example.jpg/40px-Example.jpg 2x\" data-file-width=\"275\" data-file-height=\"297\"> and stuff." targetImageWidth:self.imageSize];
```
coreyfloyd: Is this enough to performance test against?

montehurd: No, it's just to ensure the full obama article html performance test below has its static regexes already set before it runs so when it runs we're only measuring the parsing. iirc

coreyfloyd: Ahh misunderstood…

PR #741, opened 2016-06-25
Fix for image peeking not working for many images.

montehurd:
The problem is the decision whether to peek for an image was
controlled by MWKImageList's "hasImageURL" method, which checked
if the article's image list has a match for image. This check
is based on image url, but the image list no longer seems to
have the exact urls which the image tag's "src" used.

The reason things are out of sync is we are building the
imageList when the article data is saved post fetch, but the
image tag "src" attribute value isn't changed until after
this - at the time the html is about to be used.

One potential fix would be to change the image tag's "src"
values *before* we save the article data, but we can't do
that because the "secret" string we place in the url changes
between app cold starts. This means the proxy server would
ignore these requests after a restart.

I'm going to investigate doing away with the imageList
entirely. Benefits would include being able to do away with
all imageList logic *and* no longer having to do an XPath
parse before saving the article html.

### DIFF
--- a/Wikipedia/Code/MWKImageList.h
+++ b/Wikipedia/Code/MWKImageList.h
@@ -28,8 +28,6 @@
 
 - (void)addImageURL:(NSString*)imageURL;
 
-- (BOOL)hasImageURL:(NSURL*)imageURL;
-
 - (BOOL)hasImageURLString:(NSString*)imageURLString;
 
 - (MWKImage*)imageWithURL:(NSString*)imageURL;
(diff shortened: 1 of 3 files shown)


montehurd:
Ok I think I have a pretty snazzy way to efficiently parse out the list of the gallery image urls from the article html so we can consider completely getting rid of the imageLists... going to test it some more before trying to hook it up - but at the moment it's working quite nicely for the Obama article... more soon...

PR #742, opened 2016-06-27
T126667

huayu0723:
Add a new method wmf_openExternalUrl:useSafari: ignoreOperatingSystemVersion:
When the info button is pressed, the method will be called with the ignoreOperatingSystemVersion flag set as YES. In this case, whether the safari will be launched or not depends only on the useSafari param, which had been set as NO.


Wikipedia/Code/UIViewController+WMFOpenExternalUrl.h
```diff
@@ -19,4 +19,13 @@
...
 - (void)wmf_openExternalUrl:(NSURL*)url useSafari:(BOOL)useSafari;
 
+/**
+ *  Open an external URL and specifiy whether or not safari should be used, and whether or not operating system version should be ignored
+ *
+ *  @param url                           The url to use
+ *  @param useSafari                     If YES, then safari will always be used
+ *  @param ignoreOperatingSystemVersion  If NO and it's iOS 9 or later, then safari will always be used
+ */
+- (void)wmf_openExternalUrl:(NSURL*)url useSafari:(BOOL)useSafari ignoreOperatingSystemVersion:(BOOL)ignoreOperatingSystemVersion;
```
coreyfloyd: @huayu0723 we have second part of this task if you are interested. We would like to modify the behavior of our link opening to not use Safari at all unless we specify it explicitly. Since you are already working on link handling here, would you like to take on the bigger change?

If not, no worries we will merge this in and work on it separately, But if so, here is the other changes we would like to make to our link handling:

Refactor the internals of:
```
 - (void)wmf_openExternalUrl:(NSURL*)url;
```
This method should now always open within the app. In iOS 8, `SVModalWebViewController`. In iOS 9 and above, use `SFSafariViewController`

Refactor the internals of:
```
- (void)wmf_openExternalUrl:(NSURL*)url useSafari:(BOOL)useSafari;
```
If useSafari is NO, then the behavior is is identical to the method above:  Open the link within the app. In iOS 8, `SVModalWebViewController`. In iOS 9 and above, use `SFSafariViewController`
If useSafari is YES, then the url should be opened externally in Safari.

If you do make these changes, then I don't think you will need the 3rd method anymore since we never user the operating system to implicitly use Safari.

Thanks again for contributing!

huayu0723:
@coreyfloyd For sure I'll take on the change. The logic is quite clear with the new solution.

coreyfloyd:
Code is 👍 

@montehurd you want to put this through its paces when you finish helping with regression?

Wikipedia/Code/UIViewController+WMFOpenExternalUrl.m
```diff
@@ -47,16 +48,26 @@ - (void)wmf_openExternalUrl:(NSURL*)url useSafari:(BOOL)useSafari {
...
             }];
             return;
         }
+        [self wmf_presentExternalUrlWithinApp:url];
+    }
+}
+
+- (void)wmf_presentExternalUrlWithinApp:(NSURL *)url {
+//    iOS 9 and later
+    if ([[NSProcessInfo processInfo] isOperatingSystemAtLeastVersion:(NSOperatingSystemVersion){9, 0, 0}]) {
```
joewalsh: I'd suggest using `if ([SFSafariViewController class])` as per the recommendation in https://developer.apple.com/library/ios/documentation/DeveloperTools/Conceptual/cross_development/Using/using.html#//apple_ref/doc/uid/20002000-SW3

huayu0723: Sorry that I don't understand you clearly. @joewalsh

coreyfloyd:
@joewalsh lets just clean that up in a follow on.

joewalsh:
@coreyfloyd sounds good @huayu0723 nevermind re: my previous comment

joewalsh:
👍

coreyfloyd:
@joewalsh we can probably also collapse the methods in that class… it feels like it has an extra one.

PR #746, opened 2016-06-27
Small typographical touches

josve05a:
Per e.g. [Wikipedia:Username policy](https://en.wikipedia.org/wiki/Wikipedia:Username_policy) it is called "username" and not "user name" on Wikipedia.


montehurd:
👍

PR #747, opened 2016-06-27
Get rid of MWKImageList and just parse the gallery image list from the html when needed.

montehurd:


Wikipedia/Code/WMFGalleryImageListParser.m
```diff
@@ -0,0 +1,78 @@
...
+        return [[WMFGalleryImageTag alloc] initWithSrc:tagDict[@"src"]
+                                                srcset:tagDict[@"srcset"]
+                                                   alt:tagDict[@"alt"]
+                                                 width:tagDict[@"width"]
+                                                height:tagDict[@"height"]
+                                         dataFileWidth:tagDict[@"data-file-width"]
+                                        dataFileHeight:tagDict[@"data-file-height"]];
+    }];    
+    
+    NSArray<NSURL*>* imageTagURLs = [[imageTagModels bk_select:^BOOL(WMFGalleryImageTag* tag){
```
joewalsh: would there be value in a method that returns the WMFGalleryImageTag objects? This way the gallery could have one list that could be used to generate thumbnail and full res URLs without having to go through the parser again. The use case I'm thinking of is user taps an article image, the gallery shows a thumbnail - same res as was loaded for the article - while loading the full size image. The gallery could get the thumbnail and full size URLs from the same object.

Wikipedia/Code/WMFImageTag+TargetImageWidthURL.h
```diff
@@ -0,0 +1,7 @@
+#import "WMFImageTag.h"
+
+@interface WMFImageTag (TargetImageWidthURL)
+
+- (NSURL*)urlForTargetWidth:(NSUInteger)targetWidth;
```
joewalsh: minor suggestion - capitalize URL to match foundation naming conventions

joewalsh:
@montehurd I have a change that runs the WMFProxyServer src size adjustment through WMFImageTag. This helps ensure the URL we set in the article is the same we request for the gallery (I noticed some discrepancies with PNG) and keeps this logic in one place. Is it OK if I commit it to this branch?

Wikipedia/Code/WMFImageTag+TargetImageWidthURL.m
```diff
@@ -0,0 +1,62 @@
...
+ *  @param targetWidth  The desired width for images.
+ *
+ *  @return The URL returned needs to be as close to equal to the targetWidth as possible. Because the image scaler will not scale raster images above their canonical resolution the URL returned here needs to sometimes be modified to just use the canonical URL.
+ */
+- (NSURL*)urlForTargetWidth:(NSUInteger)targetWidth {
+    NSString *tagSrc = self.src;
+    
+    // One of these widths was set for us to even get here.
+    NSNumber *safeCanonicalWidthAssumption = self.dataFileWidth ? self.dataFileWidth : self.width;
+    NSString *canonicalExtension = [[[tagSrc stringByDeletingLastPathComponent] lastPathComponent] pathExtension];
```
joewalsh: Does this need to handle original URLs as well? Or is the assumption that everything will be /thumb/ URLs?

montehurd: @joewalsh i think those fall through to the empty else case below.

Wikipedia/Code/WMFImageTag+TargetImageWidthURL.h
montehurd: @joewalsh ah good point - will do.

Wikipedia/Code/WMFImageTag+TargetImageWidthURL.m
joewalsh: @montehurd ah, I see. Should this move after  `if (sizeFromSrcUrl != NSNotFound) {` then?

joewalsh: *basically everything `tagSrc =` and `sizeFromSrcUrl` can go in the first branch of that if, right?

Wikipedia/Code/WMFArticleViewController.m
```diff
@@ -1102,10 +1102,13 @@ - (nullable UIViewController*)peekViewControllerForPeekElement:(WMFPeekHTMLEleme
 }
 
 - (nullable UIViewController*)viewControllerForImageURL:(nullable NSURL*)url {
-    if (!url || ![self.article.images imageSizeVariants:url.absoluteString]) {
+
```
montehurd: @joewalsh your commit fixed this :)

Wikipedia/Code/WMFImageTag+TargetImageWidthURL.m
montehurd: @joewalsh hmm lemme think about that...

montehurd:
@joewalsh took another crack at fixing those tests... got nervous about re-adding tests. all fixed :)

montehurd:
+881 −4,241

wow

montehurd:
@coreyfloyd @joewalsh Removed the WIP label. If you have time to review I think it's good to go :)

joewalsh:
Testing it out now - looks great. Articles saved with the current shipping app load fine on this branch airplane mode. Gallery works as expected - no spinners. Only minor bug I've found is that in the _Dartmouth College_ article, tapping the hero image at the top shows a different image. This is a regression as the app store build shows the correct image.

Wikipedia/Code/MWKArticle.m
```diff
@@ -453,40 +404,75 @@ - (NSString*)debugDescription {
...
+    [imageURLs wmf_safeAddObject:articleThumbnailURL];
+
+    for (NSURL* url in imageURLs) {
+        NSAssert(url.scheme == nil, @"Non nil image url scheme detected! These must be nil here or we can get duplicate image urls in the set - ie the same url twice - once with the scheme and once without. This is because WMFImageTagParser returns urls without schemes because the html it parses does not have schemes - but the MWKImageInfo *does* have schemes in its 'canonicalFileURL' and 'imageThumbURL' properties because the api returns schemes.");
     }
 
     // remove any null objects inserted during above map/valueForKey operations
     return [imageURLs bk_reject:^BOOL (id obj) {
-        return [NSNull null] == obj;
+        return [obj isEqual:[NSNull null]];
```
joewalsh: this can stay as == since [NSNull null] is a singleton - there could be a performance impact of calling obj's isEqual method, but probably not too bad since they're NSURLs

Wikipedia/Code/WMFImageTag.m
```diff
@@ -0,0 +1,74 @@
...
+        self.width = @([width integerValue]);
+        self.height = @([height integerValue]);
+        self.dataFileWidth = @([dataFileWidth integerValue]);
+        self.dataFileHeight = @([dataFileHeight integerValue]);
+    }
+    return self;
+}
+
+- (BOOL)isWideEnoughForGallery {
+    return self.width.integerValue > WMFImageTagMinimumWidthForGalleryInclusion && self.dataFileWidth.integerValue > WMFImageTagMinimumWidthForGalleryInclusion;
```
joewalsh: it looks like the previous code checked width and height, should that be preserved?

Wikipedia/Code/MWKImage.m
```diff
@@ -272,90 +218,34 @@ - (BOOL)isEqual:(id)object {
...
 
+- (BOOL)isPathEqualToImage:(MWKImage*)image {
+    return [[[self.sourceURLString wmf_schemelessURL] stringByDeletingLastPathComponent] isEqualToString:[[image.sourceURLString wmf_schemelessURL] stringByDeletingLastPathComponent]];
+}
+
 - (BOOL)isVariantOfImage:(MWKImage*)otherImage {
     // !!!: this might not be reliable due to underscore, percent encodings, and other unknowns w/ image filenames
     return [self.fileNameNoSizePrefix isEqualToString:otherImage.fileNameNoSizePrefix]
-           && ![self isEqualToImage:otherImage];
+           && [self isPathEqualToImage:otherImage];
```
joewalsh: does path need to be checked if all images are on commons? I might need to alter the code I just added which only checks filenames.

montehurd: @joewalsh yes i think so - see the test here:
https://github.com/wikimedia/wikipedia-ios/pull/747/commits/c1da552df3dbdd851ab70ca8f25be8b4019144ef#diff-7dc0541c8d97acb98df6dc9574971ca6R20

It turns out the image scaler can use a format image name like "XXXpx-thumbnail.jpg" if the image name is very long - you can see this in the urls used for the test linked above.

montehurd: @joewalsh here is the "isVariantOfImage" test covering those "XXXpx-thumbnail.jpg" type image urls:
https://github.com/wikimedia/wikipedia-ios/pull/747/commits/c1da552df3dbdd851ab70ca8f25be8b4019144ef#diff-4f9ebe5e9be025960b5f61d5c676f7fbR36

joewalsh: @montehurd in those cases, the image filename extraction APIs should return the second to the last path component, which is the actual filename, not the thumbnail filename

Wikipedia/Code/WMFImageTagParser.h
```diff
@@ -8,6 +8,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (WMFImageTagList*)imageTagListFromParsingHTMLString:(NSString*)HTMLString;
 
+- (WMFImageTagList*)imageTagListFromParsingHTMLString:(NSString*)HTMLString withLeadImageURL:(nullable NSURL *)leadImageURL;
```
montehurd: @joewalsh I'm not sure we need "withLeadImageURL:" - the reason is the parser reduces the html you pass it to just image tags, which means if you just add an img tag for the lead image to the beginning of the HTMLString it should just work.

joewalsh: @montehurd wouldn't that cause the image to appear twice - leadImageURL is picking an image URL out of the list to be the first

WikipediaUnitTests/Code/WMFImageTagParserTests.m
```diff
@@ -335,6 +335,92 @@ - (void)testImgTagReductionNonHTMLString {
     assertThat([self.parser imgTagsOnlyFromHTMLString:@"bla bla"], is(equalTo(@"")));
 }
 
+- (void)testObamaArticleGalleryImageListExtractionWithLeadImage {
+    NSArray* parsedObamaGalleryURLS = [[self.parser imageTagListFromParsingHTMLString:[self allObamaHTML] withLeadImageURL:[NSURL URLWithString:@"https//upload.wikimedia.org/wikipedia/commons/thumb/f/f1/BarackObamaportrait.jpg/640px-BarackObamaportrait.jpg"]] imageURLsForGallery];
```
montehurd: i think `https//` should be `http://`

WikipediaUnitTests/Code/WMFImageTagParserTests.m
```diff
@@ -335,6 +335,92 @@ - (void)testImgTagReductionNonHTMLString {
...
+       @"//upload.wikimedia.org/wikipedia/commons/thumb/3/36/Seal_of_the_President_of_the_United_States.svg/640px-Seal_of_the_President_of_the_United_States.svg.png"
+       ] bk_map:^NSURL*(NSString* stringURL){
+           return [NSURL URLWithString:stringURL];
+       }];
+    
+    assertThat(parsedObamaGalleryURLS, is(equalTo(expectedObamaGalleryURLs)));
+}
+
+- (void)testObamaArticleGalleryImageListExtractionWithMistmatchedSizeLeadImage {
+    NSArray* parsedObamaGalleryURLS = [[self.parser imageTagListFromParsingHTMLString:[self allObamaHTML] withLeadImageURL:[NSURL URLWithString:@"https//upload.wikimedia.org/wikipedia/commons/thumb/f/f1/BarackObamaportrait.jpg/816px-BarackObamaportrait.jpg"]] imageURLsForGallery];
```
montehurd: `https//` should be `https://` here too

montehurd:
@joewalsh @coreyfloyd ok should be good - been tapping/peeking all sorts of article images and they seem solid.

joewalsh:
@montehurd I made some commits that eliminate dependence on NSXMLParser. The upside is here is that now WMFProxyServer and WMFImageTagParser share the same parsing logic and there's no need for an extra percent encode/decode. The new tests pass and I verified that the logarithm article behaves correctly.

www/js/transforms/widenImages.js
```diff
@@ -45,52 +32,6 @@ function makeRoomForImageWidening(image) {
...
-				}
-				image.src = newSrc;
-	            if(enableDebugBorders){
-	                image.style.borderWidth = '10px';
-	            }
-			}
-        } 
-    }
-}
-
```
joewalsh: It's great to no longer have this logic duplicated in JS

Wikipedia/Code/WMFImageTag.m
```diff
@@ -37,16 +37,22 @@ - (instancetype)initWithSrc:(NSString*)src
...
+        self.height = [height isEqual:[NSNull null]] ? nil : @([height integerValue]);
+        self.dataFileWidth = [dataFileWidth isEqual:[NSNull null]] ? nil : @([dataFileWidth integerValue]);
+        self.dataFileHeight = [dataFileHeight isEqual:[NSNull null]] ? nil : @([dataFileHeight integerValue]);
     }
     return self;
 }
 
-- (BOOL)isWideEnoughForGallery {
-    return self.width.integerValue > WMFImageTagMinimumWidthForGalleryInclusion && self.dataFileWidth.integerValue > WMFImageTagMinimumWidthForGalleryInclusion;
+- (BOOL)isSizeLargeEnoughForGalleryInclusion {
```
joewalsh: I verified that the images I was seeing before that were wide and short (and weren't in the app store version gallery) are no longer in the gallery. Looks good

PR #750, opened 2016-06-28

niedzielski:


montehurd:
👍

josve05a:
"Fussy" ?

montehurd:
@josve05a I think that's what TWN uses for strings which have changed so translators can re-check them.

PR #753, opened 2016-06-28
Ensure title equivalence in cases where the same character is encoded in different ways

joewalsh:
Ensure title text is converted using precomposedStringWithCanonicalMapping when creating a title URL


joewalsh:
Article with info on `precomposedStringWithCanonicalMapping`: https://www.objc.io/issues/9-strings/unicode/

PR #755, opened 2016-06-29
Don't show language codes in top read feed header T133420

joewalsh:
This fix has two parts corresponding to the two commits - the first commit is a fallback on showing the generic title "Top Read Articles on Wikipedia" rather than "Top Read Articles on als Wikipedia"

The second commit is an attempt to add a mapping from wikipedia language code to ISO language code, which shows "Top Read Articles on Swiss German Wikipedia". I'm not sure this part is necessary, but if it is included, it would be good to find a more complete list of the mapping from wiki language to ISO language.

### DIFF
--- a/Wikipedia/Code/NSLocale+WMFExtras.swift
+++ b/Wikipedia/Code/NSLocale+WMFExtras.swift
@@ -11,6 +11,13 @@ extension NSLocale {
         return (langCode == "en" || langCode.hasPrefix("en-")) ? true : false;
     }
     func wmf_localizedLanguageNameForCode(code: String) -> String? {
-        return self.displayNameForKey(NSLocaleLanguageCode, value: code)
+        var ISOLanguageCode:String
+        switch code {
+        case "als": //wiki language codes don't map 1:1 to ISOLanguageCodes
+            ISOLanguageCode = "gsw"
+        default:
+            ISOLanguageCode = code
+        }
+        return self.displayNameForKey(NSLocaleLanguageCode, value: ISOLanguageCode)
     }
 }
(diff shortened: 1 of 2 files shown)


Wikipedia/Code/NSLocale+WMFExtras.swift
```diff
@@ -11,6 +11,13 @@ extension NSLocale {
         return (langCode == "en" || langCode.hasPrefix("en-")) ? true : false;
     }
     func wmf_localizedLanguageNameForCode(code: String) -> String? {
-        return self.displayNameForKey(NSLocaleLanguageCode, value: code)
+        var ISOLanguageCode:String
+        switch code {
+        case "als": //wiki language codes don't map 1:1 to ISOLanguageCodes
+            ISOLanguageCode = "gsw"
+        default:
+            ISOLanguageCode = code
```
joewalsh: is there somewhere we can get this complete mapping?

coreyfloyd: @joewalsh Possibly the site matrix call may help - we have a ticket to use it for languages:

https://phabricator.wikimedia.org/T121128

coreyfloyd: @joewalsh if it is helpful, you can ask Josh if thats something that should be prioritized

coreyfloyd:
Do not merge into 5.0.5

coreyfloyd:
Actually before ei remove from 5.0.5 - @joewalsh why did you add the wip tag?

joewalsh:
@coreyfloyd waiting on whether or not I should include the site matrix work

coreyfloyd:
@joewalsh not for 5.0.5 - we should do that separately unless you think your current fix isn't worthwhile to include without the matrix work.

PR #756, opened 2016-06-29
pass article and title to web view controller so the correct fragment is used when loading the page

joewalsh:


coreyfloyd:
@joein3d I'd like to fix this in my PR (or at least after my PR) unless we need to patch the 5.0.5 build.

joewalsh:
@coreyfloyd sounds good

montehurd:
@coreyfloyd There are 2 links on testwiki > bird to different sections on testwiki > bird2 which you can use to test for this.

coreyfloyd:
@joewalsh we want to get this in since my PR will be 5.1 and not 5.0.5, I'd like to explore fixing gar underlying issue with the title on MWKArticle missing the fragment rather than passing 2 arguments. That way we avoid the ambiguity of having 2 titles in the webviewcontroller

PR #757, opened 2016-06-30
Fix auto login loop 

coreyfloyd:
Don't call finish… this happens automatically
Also make sure to call the passed in failure block


montehurd:
Ooooh was this why login was taking so long?

joewalsh:
👍

PR #758, opened 2016-07-01
Memory crashes/T139074

joewalsh:
Responsibility of cache cleanup is moved to MWKHistoryList & MWKSavedPage list so that articles not in either list are cleaned up. MWKDataStore keeps a queue of titles to be removed and deletes them while the app is in the background.


Wikipedia/Code/MWKDataStore.m
```diff
@@ -495,6 +529,53 @@ - (void)iterateOverArticles:(void (^)(MWKArticle*))block {
     }
 }
 
+- (void)startCacheRemoval {
+    dispatch_async(self.cacheRemovalQueue, ^{
+        self.cacheRemovalActive = true;
```
coreyfloyd: What do you think of using an NSOperationQueue as an ivar and then using an NSOperation? Then you could just check if it has any current operations.

This would be instead of using GCD plus a flag to check if cache removal was in progress.

The main thought here is that that it would eliminate the extra variable for tracking state and it would use a higher level API.

Wikipedia/Code/MWKDataStore.m
```diff
@@ -495,6 +529,53 @@ - (void)iterateOverArticles:(void (^)(MWKArticle*))block {
...
+    dispatch_async(self.cacheRemovalQueue, ^{
+        NSMutableArray* allTitlesToRemove = [[self cacheRemovalListFromDisk] mutableCopy];
+        if (allTitlesToRemove == nil) {
+            allTitlesToRemove = [NSMutableArray arrayWithArray:titlesToRemove];
+        } else {
+            [allTitlesToRemove addObjectsFromArray:titlesToRemove];
+        }
+        NSError* error = nil;
+        if (![self saveCacheRemovalListToDisk:allTitlesToRemove error:&error]) {
+            DDLogError(@"Errora saving cache removal list to disk: %@", error);
```
joewalsh: typo

Wikipedia/Code/MWKDataStore.m
joewalsh: `cacheRemovalActive` would still be necessary in the NSOp case - to check whether or not the op should be added to the queue. you could nil out the queue, but that would prevent necessary operations from going through - namely writing the work list to disk.

PR #761, opened 2016-07-06
Update to wikimedia/AFNetworking pod

coreyfloyd:
This contains fixes to eliminate KVO in AFURLSessionManager that potentially causes:
https://phabricator.wikimedia.org/T139106

To see more discussion on this, checkout the Pull request with the relevant changes here:
https://github.com/wikimedia/AFNetworking/pull/1

In hockey app, you can see the trace here:
https://rink.hockeyapp.net/manage/apps/152731/app_versions/51/crash_reasons/120924114?type=overview

It's not testable, since we can't reproduce - but this looks like the correct fix,


joewalsh:
👍

PR #762, opened 2016-07-06
Fix crash by not allowing fragments to fall through to the external U…

coreyfloyd:
…RL handling.

This bug was always present, but the 3rd party browser ignored the bad URL.

However the SFSafariViewController is pretty angry.


joewalsh:
👍

montehurd:
@coreyfloyd hmm that last "Code formatting" commit broke the references which had been working...

coreyfloyd:
@montehurd that is really really weird… so if you revert it - then it works?

coreyfloyd:
Just a note… @montehurd is going to fix the secondary issue here which is references not working

montehurd:
@coreyfloyd lemme try... had just updated and they stopped working - was looking at the working ones to see what was up with the "A" one...

montehurd:
@coreyfloyd confirmed

coreyfloyd:
@montehurd go ahead and revert… but if you can describe what actually happened there… did some logic actually get changed?

montehurd:
@coreyfloyd will do. i think i know what broke... confirming...

Wikipedia/Code/WebViewController.m
```diff
@@ -150,53 +150,54 @@ - (void)userContentController:(WKUserContentController*)userContentController di
...
-                                                                      width:imageClicked[@"width"]
-                                                                     height:imageClicked[@"height"]
-                                                              dataFileWidth:imageClicked[@"data-file-width"]
-                                                             dataFileHeight:imageClicked[@"data-file-height"]];
-
-            if (![imageTagClicked isSizeLargeEnoughForGalleryInclusion]) {
-                return;
-            }
+        }
+    } else if (message.body[@"imageClicked"]) {
```
montehurd: @coreyfloyd this was the problem - this needs to be nested inside of the `} else if ([message.name isEqualToString:@"clicks"]) {` if statement, not at same level.

Wikipedia/Code/WKWebView+LoadAssetsHtml.m
```diff
@@ -16,11 +16,11 @@ - (void)loadHTMLFromAssetsFile:(NSString*)fileName scrolledToFragment:(NSString*
     
     fragment = fragment ? fragment : @"top";
     NSURL* requestURL = [[WMFProxyServer sharedProxyServer] proxyURLForRelativeFilePath:fileName fragment:fragment];
-    NSURLRequest* request = [NSURLRequest requestWithURL:requestURL];
+    NSURLRequest* request = [NSURLRequest requestWithURL:requestURL cachePolicy:NSURLRequestReloadIgnoringCacheData timeoutInterval:10];
```
montehurd: hehe i hope the proxy server doesn't take more than 10 seconds!

joewalsh: oh - I can revert this line... I thought WKWebView was caching when I saw "en.wikipedia" as the base href, but really it was being changed by the JavaScript.

montehurd: Ah, cool.

montehurd:
@coreyfloyd @joewalsh got the "A" reference link on "enwiki > Airplane" working finally.

PR #763, opened 2016-07-06
Fix crash when showing main page due to tripping a toc assertion

coreyfloyd:
The crash came when trying to add items to the ToC on an article which should not show a ToC

This fix just separates the footer and ToC logic so it can be called at the appropriate times.


joewalsh:
👍

coreyfloyd:
@montehurd @joewalsh accidentally pushed the last commit which fixes https://phabricator.wikimedia.org/T139513 on this branch

Should be fine, it is a one liner.

PR #767, opened 2016-07-07
Fix crash when saving image URLs by aggressively rejecting nil (NSNul…

coreyfloyd:
…l) URLs

- Reject nil URLs when retrieving imageInfoForTitle from the datastore
- Check for the proper class (MWKImageInfo) before mapping to a URL (previously a KVO key value coding call on an unknown class)
- Check for the proper class (NSURL) before mapping to a scheme less URL (previously assumed it was an NSURL and crashed)

- Additionally refactor both scheme removal code paths into one (in NSURL category)

### DIFF
--- a/Wikipedia/Code/MWKArticle.m
+++ b/Wikipedia/Code/MWKArticle.m
@@ -431,14 +431,12 @@ static MWKArticleSchemaVersion const MWKArticleCurrentSchemaVersion = MWKArticle
     }];
 }
 
--(NSArray<NSURL*>*)schemelessURLs:(NSArray<NSURL*>*)urls {
-    return [urls bk_map:^NSURL*(NSURL* url){
-        NSString* urlStr = url.absoluteString;
-        NSRange dividerRange = [urlStr rangeOfString:@"://"];
-        if (dividerRange.location != NSNotFound) {
-            return [NSURL URLWithString:[urlStr substringFromIndex:dividerRange.location + 1]];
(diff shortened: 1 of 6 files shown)


Wikipedia/Code/MWKArticle.m
```diff
@@ -431,14 +431,12 @@ - (NSString*)debugDescription {
...
 
--(NSArray<NSURL*>*)schemelessURLs:(NSArray<NSURL*>*)urls {
-    return [urls bk_map:^NSURL*(NSURL* url){
-        NSString* urlStr = url.absoluteString;
-        NSRange dividerRange = [urlStr rangeOfString:@"://"];
-        if (dividerRange.location != NSNotFound) {
-            return [NSURL URLWithString:[urlStr substringFromIndex:dividerRange.location + 1]];
+-(NSArray<NSURL*>*)schemelessURLsRejectingNilURLs:(NSArray<NSURL*>*)urls {
+    return [urls wmf_mapAndRejectNil:^NSURL*(NSURL* url){
+        if([url isKindOfClass:[NSURL class]]){
```
montehurd: Is this class check redundant because urls is already `NSArray<NSURL*>*`?

joewalsh:
👍 pending monte's additions

Wikipedia/Code/MWKArticle.m
coreyfloyd: Already talked about in hangout but for posterity:

It can't be guaranteed - it is only a hint to the compiler

montehurd:
@coreyfloyd 👍 if my last commits look ok to you

joewalsh:
👍

PR #768, opened 2016-07-08
Fix no class checks crash

coreyfloyd:
We have another crash in hockey

This is due to making assumptions about the classes we get back from JS

See T139751 and https://rink.hockeyapp.net/manage/apps/152731/app_versions/62/crash_reasons/127780360

### DIFF
--- a/Wikipedia/Code/NSDictionary+WMFExtensions.m
+++ b/Wikipedia/Code/NSDictionary+WMFExtensions.m
@@ -1,0 +1,13 @@
+
+#import "NSDictionary+WMFExtensions.h"
+
+@implementation NSDictionary (WMFExtensions)
+
+- (BOOL)containsNullObjects {
+    NSNull* null = [self bk_match:^BOOL (id key, id obj) {
+        return [obj isKindOfClass:[NSNull class]];
+    }];
+    return (null != nil);
+}
+
+@end
--- a/Wikipedia/Code/NSDictionary+WMFExtensions.m
+++ b/Wikipedia/Code/NSDictionary+WMFExtensions.m
@@ -42,4 +42,12 @@
     return hasNull;
 }
 
+
+- (NSDictionary*)wmf_dictionaryByRemovingNullObjects{
+    return [self bk_reject:^BOOL(id key, id obj) {
+        return [obj isKindOfClass:[NSNull class]];
+    }];
+    
+}
+
 @end
--- a/Wikipedia/Code/NSDictionary+WMFExtensions.m
+++ b/Wikipedia/Code/NSDictionary+WMFExtensions.m
@@ -3,11 +3,42 @@
 
 @implementation NSDictionary (WMFExtensions)
 
-- (BOOL)containsNullObjects {
+- (BOOL)wmf_containsNullObjects {
     NSNull* null = [self bk_match:^BOOL (id key, id obj) {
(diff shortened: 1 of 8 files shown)


Wikipedia/Code/NSDictionary+WMFExtensions.m
```diff
@@ -0,0 +1,13 @@
+
+#import "NSDictionary+WMFExtensions.h"
+
+@implementation NSDictionary (WMFExtensions)
+
+- (BOOL)containsNullObjects {
```
joewalsh: should this be recursive? so if a value in a dictionary is also a dictionary, it should return TRUE if that contains NSNull

coreyfloyd: @joewalsh I could go either way - you have a preference? Should we have 2 versions - one recursive and one not?

joewalsh:
👍 this is good, maybe with a follow on ticket to recursively remove null

Wikipedia/Code/NSDictionary+WMFExtensions.m
```diff
@@ -42,4 +42,12 @@ - (BOOL)wmf_recursivelyContainsNullObjects{
     return hasNull;
 }
 
+
+- (NSDictionary*)wmf_dictionaryByRemovingNullObjects{
+    return [self bk_reject:^BOOL(id key, id obj) {
+        return [obj isKindOfClass:[NSNull class]];
```
joewalsh: a quicker check would be obj == [NSNull null]

Wikipedia/Code/WebViewController.m
```diff
@@ -86,121 +86,154 @@ - (instancetype)initWithSession:(SessionSingleton*)aSession {
...
-            
-            NSString* selectedImageSrcURLString = imageClicked[@"src"];
-            NSCParameterAssert(selectedImageSrcURLString.length);
-            if (!selectedImageSrcURLString.length) {
-                DDLogError(@"Image clicked callback invoked with empty src url: %@", imageClicked);
-                return;
+        }
+    }
+
+    if ([message.body isKindOfClass:[NSString class]]) {
```
montehurd: The message "name" checks below should, I think, be part of the message name "if" check above. Formerly the single "if" above was basically functioning like to a "switch" for all the name of messages which could be received. But here we've broken out a second "name if" structure, and inverted the logic, now checking the body before the name.

Wikipedia/Code/WMFImageTag.m
```diff
@@ -84,9 +88,14 @@ - (instancetype)initWithImageTagContents:(NSString*)imageTagContents {
         *stop = width && height && dataFileWidth && dataFileHeight && src && srcAttributeRange.location != NSNotFound;
     }];
 
+    //Don't continue initialization if we have invalid src
+    if([src length] == 0){
```
montehurd: Should this be doing the same `if([[src stringByTrimmingCharactersInSet:[NSCharacterSet whitespaceCharacterSet]] length] == 0){` check?

montehurd:
added one comment. 👍

Wikipedia/Code/WMFImageTag.m
coreyfloyd: We don't have to because its in the default initializer (thats why I added it there so we only need to do it once)

PR #769, opened 2016-07-09
Refactor WKScriptMessage handling.

montehurd:
Broke things up into individual methods.

Added WMFWKScriptMessageType so we can more easily specify and check for expected body message types.

Fixed bug introduced with the recent null check fixes which prevented peekElement from being properly reset.

Fixes bug preventing reference panel from scrolling all the way to the bottom if there's a lot of text in the panel.

Fixes bug with reference links being peekable - they should show the panel instead.

### DIFF
--- a/Wikipedia/Code/WKScriptMessage+WMFScriptMessage.m
+++ b/Wikipedia/Code/WKScriptMessage+WMFScriptMessage.m
@@ -3,24 +3,24 @@
 @implementation WKScriptMessage (WMFScriptMessage)
 
 + (WMFWKScriptMessageType)wmf_typeForMessageName:(NSString*)name {
-    if ([name isEqualToString:@"peek"]) {
-        return WMFWKScriptMessagePeek;
-    } else if ([name isEqualToString:@"sendJavascriptConsoleLogMessageToXcodeConsole"]) {
-        return WMFWKScriptMessageConsoleMessage;
+    if ([name isEqualToString:@"nonAnchorTouchEndedWithoutDragging"]) {
+        return WMFWKScriptMessageNonAnchorTouchEndedWithoutDragging;
     } else if ([name isEqualToString:@"linkClicked"]) {
         return WMFWKScriptMessageClickLink;
     } else if ([name isEqualToString:@"imageClicked"]) {
         return WMFWKScriptMessageClickImage;
+    } else if ([name isEqualToString:@"peek"]) {
+        return WMFWKScriptMessagePeek;
     } else if ([name isEqualToString:@"referenceClicked"]) {
         return WMFWKScriptMessageClickReference;
     } else if ([name isEqualToString:@"editClicked"]) {
         return WMFWKScriptMessageClickEdit;
-    } else if ([name isEqualToString:@"nonAnchorTouchEndedWithoutDragging"]) {
-        return WMFWKScriptMessageNonAnchorTouchEndedWithoutDragging;
     } else if ([name isEqualToString:@"lateJavascriptTransform"]) {
         return WMFWKScriptMessageLateJavascriptTransform;
     } else if ([name isEqualToString:@"articleState"]) {
         return WMFWKScriptMessageArticleState;
+    } else if ([name isEqualToString:@"sendJavascriptConsoleLogMessageToXcodeConsole"]) {
+        return WMFWKScriptMessageConsoleMessage;
     } else{
         return WMFWKScriptMessageUnknown;
     }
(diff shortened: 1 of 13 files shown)


Wikipedia/assets/index.js
```diff
@@ -124,7 +124,7 @@ function maybeSendMessageForTarget(event, hrefTarget){
     var href = hrefTarget.getAttribute( "href" );
     var hrefClass = hrefTarget.getAttribute('class');
     if (hrefTarget.getAttribute( "data-action" ) === "edit_section") {
-        window.webkit.messageHandlers.clicks.postMessage({"editClicked": { sectionId: hrefTarget.getAttribute( "data-id" ) }});
```
montehurd: This PR makes the JS not send dictionaries within dictionaries. 
Crashes were happening when these nested dictionaries were NSNull and we tried to access values on them with dictionary subscript syntax - i.e. `dict[@"somekey"]`

Wikipedia/Code/WebViewController.m
```diff
@@ -89,146 +90,164 @@ - (instancetype)initWithSession:(SessionSingleton*)aSession {
...
-     *  All messages from JS can contain NSNulls.
-     *  All elements must be checked for existence of NSNull before being passed on to any other methods.
-     * Additionally body is untyped, so its class must be
-     * checked before proceeding.
-     */
-
-    if ([message.body isKindOfClass:[NSNull class]]) {
+
+    WMFWKScriptMessageType messageType = [WKScriptMessage wmf_typeForMessageName:message.name];
+    if (![message.body isKindOfClass:[WKScriptMessage wmf_expectedMessageBodyClassForType:messageType]]) {
```
montehurd: All handler methods will be passed correct type - never NSNull.

montehurd:
@joewalsh @coreyfloyd should be good for review.

Wikipedia/Code/WebViewController.m
```diff
@@ -89,146 +90,164 @@ - (instancetype)initWithSession:(SessionSingleton*)aSession {
...
+            [self handleArticleStateScriptMessage:message.body];
+            break;
+        case WMFWKScriptMessageUnknown:
+            NSAssert(NO, @"Unhandled script message type!");
+            break;
+    }
+}
+
+- (void)handlePeekScriptMessage:(NSDictionary*)messageDict {
+    if(messageDict.count == 0){
```
joewalsh: Minor suggestion - check for the elements you want (tagName, src, href) rather than checking messageCount == 0 - this way the behavior is safe even if some objects are missing or the wrong keys are passed

joewalsh:
👍

Wikipedia/Code/WebViewController.m
montehurd: Ah, good point.

Wikipedia/Code/WKScriptMessage+WMFScriptMessage.m
```diff
@@ -0,0 +1,64 @@
...
+        case WMFWKScriptMessageLateJavascriptTransform:
+            return [NSString class];
+            break;
+        case WMFWKScriptMessageArticleState:
+            return [NSString class];
+            break;
+        case WMFWKScriptMessageConsoleMessage:
+            return [NSDictionary class];
+            break;
+        case WMFWKScriptMessageUnknown:
```
coreyfloyd: This is better for the "0" case - because that way its the default which makes it safer

Wikipedia/Code/WMFRotationRespectingTabBarController.swift
```diff
@@ -1,6 +1,18 @@
 
 import UIKit
 
+//HAX: work-around for iOS 9 infinite recursion loop bug with UIAlertController
+// http://stackoverflow.com/a/36476346/135557 . Noticed while attempting 3d touch
+// on link in reference panel.
+extension UIAlertController {
```
coreyfloyd: This needs to be in a separate PR

Wikipedia/Code/WebViewController.m
coreyfloyd: Ask don't tell

joewalsh:
👍 after the UIAlertController stuff is in a separate PR

montehurd:
I moved the "ref panel scroll issue" bits to https://github.com/wikimedia/wikipedia-ios/pull/773

montehurd:
I moved the "ref panel link 3d touch crash fix" to #774

montehurd:
@coreyfloyd I made those changes. Should be good to go.

montehurd:
Note: the UIAlertController bits ended up being redundant. https://github.com/wikimedia/wikipedia-ios/pull/774 fixed same issue.

PR #773, opened 2016-07-12
Fix reference panel scroll issue.

montehurd:
Fixes issue with reference panel not being able to scroll to bottom of long reference text.

**Before:**
before mov

**After:**
after mov


joewalsh:
👍

PR #774, opened 2016-07-12
Fix ref panel link 3d touch crash.

montehurd:
https://phabricator.wikimedia.org/T140168

Repro steps:

- load "enwiki > Tyrannosaurus"
- tap second reference link - i.e. "[2]"
- do 3d touch on the link in the black reference panel at the bottom of the screen
- CRASH!


joewalsh:
👍

PR #775, opened 2016-07-12
don't let NSCache evict section controller cache objects as it causes an infinite feed load loop

joewalsh:


joewalsh:
So here's what was happening:

If the user has more than totalMaxNumberOfSections, as the section controllers were all created, the section controller cache would evict the initial controllers that were added (which happened to be visible on screen). When WMFExploreViewController requested those visible sections from the cache, they were no longer available, so they were re-created and re-evicted, on and on forever.

Since WMFExploreViewController expects the visible sections/controllers to exist in the cache, I gave  it control over what gets evicted from the cache and when. The WMFExploreSectionControllerCache can no longer self evict. Also, instead of only being able to evict everything from the cache, it evicts sections that aren't visible (which has the side effect of fixing the complete section reload on app exit & relaunch).

Wikipedia/Code/WMFExploreSectionControllerCache.m
```diff
@@ -52,9 +47,7 @@ - (instancetype)initWithDataStore:(MWKDataStore*)dataStore {
     if (self) {
         self.dataStore                              = dataStore;
         self.sectionControllersBySection            = [[NSMutableDictionary alloc] init];
-        self.reverseLookup                          =
-            [NSMapTable mapTableWithKeyOptions:NSMapTableWeakMemory | NSMapTableObjectPointerPersonality
-                                  valueOptions:NSMapTableWeakMemory];
+        self.reverseLookup                          = [[NSMutableDictionary alloc] init];
```
coreyfloyd: Do we no longer need this to be weak? Is it expected that the new behavior will take care of cleaning up all the objects in the reverse lookup?

Also, is it ok that the section controllers don't conform to NSCopying - will the dictionary be able to find them based on pointer equality?

coreyfloyd: Never mind… you are now using the hash… so that we do't have to deal with the section controllers themselves

coreyfloyd: New question… do we need to implement the hash method for the controllers?

joewalsh: The new expected behavior is that the view controller owns the cache and cleans it. The hash method on NSObject seems to suffice.

coreyfloyd:
👍 

Waiting for travis… its being weird today

coreyfloyd:
Merging and will let jerkins do the testing

PR #776, opened 2016-07-14
Fix iOS link click regression. Was using JS method not present w/iOS 8.

montehurd:
https://phabricator.wikimedia.org/T140400


montehurd:
@coreyfloyd @joewalsh oops :/

joewalsh:
👍

PR #777, opened 2016-07-15
Fix article image maps.

montehurd:
https://phabricator.wikimedia.org/T140444

Example articles with image maps which were messed up:

- *"enwiki > Counties of England > Scope and structure > Local government"*
No longer scrolled side to side, links not in correct places over image.

- *"enwiki > Kingdom (biology) > first non lead image is an image map"*
Image was incorrectly widened and taps did not load respective article for tapped part of the image.

- *"enwiki > Kingdom (biology) > Three domains of life > Phylogenetic Tree of Life image is an image map"*
Image did not scroll side to side and taps didn't load respective article for tapped part of the image.

### DIFF
--- a/Wikipedia/assets/index.js
+++ b/Wikipedia/assets/index.js
@@ -337,11 +337,7 @@ var transformer = require("../transformer");
 var utilities = require("../utilities");
 
 function shouldAddImageOverflowXContainer(image) {
-    if ((image.width > (window.screen.width * 0.8)) && !utilities.isNestedInTable(image)){
-        return true;
-    }else{
-        return false;
-    }
+    return ((image.width > (window.screen.width * 0.8)) && !utilities.isNestedInTable(image)) ? true : false;
 }
 
 function addImageOverflowXContainer(image, ancestor) {
@@ -357,8 +353,7 @@ function firstAncestorWithMultipleChildren (el) {
     return el;
 }
 
-function maybeAddImageOverflowXContainer() {
-    var image = this;
+function maybeAddImageOverflowXContainer(image) {
     if (shouldAddImageOverflowXContainer(image)){
         var ancestor = firstAncestorWithMultipleChildren (image);
         if(ancestor){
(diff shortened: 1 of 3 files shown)


montehurd:
@coreyfloyd this is a regression in 5.0.5 as well!

www/js/transforms/addImageOverflowContainers.js
```diff
@@ -37,8 +32,6 @@ transformer.register( "addImageOverflowXContainers", function( content ) {
     // side to side if needed without causing the entire section to scroll side to side.
     var images = content.getElementsByTagName('img');
     for (var i = 0; i < images.length; ++i) {
-        // Load event used so images w/o style or inline width/height
```
montehurd: This needed to happen *before* image widening because image widening checks for hasOverflowXContainer, but image widening was updated to no longer wait until image 'load'. Because we were still waiting for image 'load' here it no longer happened before image widening.

www/js/transforms/addImageOverflowContainers.js
```diff
@@ -37,8 +32,6 @@ transformer.register( "addImageOverflowXContainers", function( content ) {
...
-        // attributes can still have their size determined reliably.
-        images[i].addEventListener('load', maybeAddImageOverflowXContainer, false);
```
montehurd: This needed to happen *before* image widening because image widening checks for hasOverflowXContainer, but image widening was updated to no longer wait until image 'load'. Because we were still waiting for image 'load' here it no longer happened before image widening.

montehurd:
@coreyfloyd no merge conflicts with branch `releases/5.0.5.901.1` :)

Wikipedia/assets/index.js
```diff
@@ -337,11 +337,7 @@ var transformer = require("../transformer");
 var utilities = require("../utilities");
 
 function shouldAddImageOverflowXContainer(image) {
-    if ((image.width > (window.screen.width * 0.8)) && !utilities.isNestedInTable(image)){
-        return true;
-    }else{
-        return false;
-    }
+    return ((image.width > (window.screen.width * 0.8)) && !utilities.isNestedInTable(image)) ? true : false;
```
joewalsh: can probably just `return ((image.width > (window.screen.width * 0.8)) && !utilities.isNestedInTable(image));` unless some permutation of that doesn't return a boolean in js

joewalsh:
👍

coreyfloyd:
Looks good.

I'll merge into develop now - lets check with @jtminor on Monday if he wants to patch 5.0.5.

Wikipedia/assets/index.js
montehurd: haha how did i miss that! :)

PR #778, opened 2016-07-15
SPIKE - Rough find-in-page proof of concept.

montehurd:
Load the Obama article in simulator then tap command shift M to
trigger memory warning. This will highlight all occurences of the
string "Obama" in the artice.

Native land send message to JS land to perform search for term "Obama".

JS land examines visible text elements, highlighting the matches, and
sends message back to native land with array of the id's of the spans
which were added around the matches to make them yellow.

Now if you repeatedly tap command shift M the native code will cycle
though the matches and scroll the web view to the next one.

untitled 2 mov


montehurd:
Closing. Was just proof-of-concept.

PR #779, opened 2016-07-18
Easy/hist save scroll/t139530

KVTaniguchi:
The purpose of this PR is to add a scroll to top action to the history and saved lists.  When the Saved or History tabs are pressed in the tab bar, the corresponding list should scroll to the first cell.

How to test:
Open the app with a user that has 0 saved and 0 history.  Pressing the Saved and History tabs should do nothing (the app shouldn't crash due to having 0 list items).
Load a user or use the app such that the Saved and History lists have 20+ items each.  Pressing the Saved or History tabs should scroll the list to the first cell.


coreyfloyd:
@KVTaniguchi thanks for the pr! 

The test is failing because the new file was added to the root directory of the repo. If you just move the file into the Wikipedia/Code directory everything will pass and we can merge.

coreyfloyd:
👍

PR #781, opened 2016-07-19
Regression - Fix crash related to not properly handing thumbnail url from pdfs.

montehurd:
https://phabricator.wikimedia.org/T140714

Saving "enwiki > Tatar language" was causing crash.

The image scaler handles serving up thumbnails for PDFs a little differently - note the "**page1-**" prefix preceding the "**240px-**" size prefix:
https://upload.wikimedia.org/wikipedia/commons/thumb/6/65/A_Fish_and_a_Gift.pdf/page1-240px-A_Fish_and_a_Gift.pdf.jpg

### DIFF
--- a/WikipediaUnitTests/Code/MWKImageInfo+MWKImageComparisonTests.m
+++ b/WikipediaUnitTests/Code/MWKImageInfo+MWKImageComparisonTests.m
@@ -1,11 +1,3 @@
-//
-//  MWKImageInfo+MWKImageComparisonTests.m
-//  Wikipedia
-//
-//  Created by Brian Gerstle on 2/12/15.
-//  Copyright (c) 2015 Wikimedia Foundation. All rights reserved.
-//
-
 #import "MWKImageInfo+MWKImageComparison.h"
 #import <XCTest/XCTest.h>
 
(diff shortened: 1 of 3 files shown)


WikipediaUnitTests/Code/MWKImageInfo+MWKImageComparisonTests.m
```diff
@@ -21,16 +13,16 @@ @interface MWKImageInfo_MWKImageComparisonTests : XCTestCase
 @implementation MWKImageInfo_MWKImageComparisonTests
 
 - (void)testAssociation {
-    MWKImage* image    = [MWKImage imageAssociatedWithSourceURL:@"some_file_name.jpg/400px-some_file_name.jpg"];
-    MWKImageInfo* info = [MWKImageInfo infoAssociatedWithSourceURL:@"some_file_name.jpg/800px-some_file_name.jpg"];
+    MWKImage* image    = [MWKImage imageAssociatedWithSourceURL:@"/thumb/some_file_name.jpg/400px-some_file_name.jpg"];
```
coreyfloyd: @joewalsh why do we need to fix the test? 

Worried that the tests are changing for images that aren't PDFs. Did we need to update any assumptions about the paths in the codebase?

joewalsh: @coreyfloyd the tests were failing. They're failing because they aren't using actual URLs, but rather a made up partial path. In the wild, URLs like this would have /thumb/ somewhere in the path. If we want to go a step further, we can use actual commons thumbnail URLs to test.

PR #783, opened 2016-07-20
Handle parsing of thumbnailer image urls from canonical TIF files.

montehurd:
**See phab ticket for details:**
https://phabricator.wikimedia.org/T140854

----

**TODO:**

- [x] check other file type handlers to see if there are other type-specific thumbnailer url syntaxes we need to handle. 

----

**RELEVANT MEDIAWIKI CODE:**

Mediawiki core appears to deal with common image type thumbnailing (jpg, png etc) with:
https://github.com/wikimedia/mediawiki/blob/master/thumb.php

Checked the following extensions list:
https://www.mediawiki.org/wiki/Category:Extensions_used_on_Wikimedia

From the list above these 2 seemed the relevant ones:
https://www.mediawiki.org/wiki/Extension:PdfHandler
https://www.mediawiki.org/wiki/Extension:PagedTiffHandler

----

**RESULTS:**

By examining the PagedTiffHandler, noticed "lossless" "png" syntax which needed to be supported in addition to the "lossy" "jpg" syntax.

### DIFF
--- a/Wikipedia/Code/AppDelegate.m
+++ b/Wikipedia/Code/AppDelegate.m
@@ -97,9 +97,10 @@
 
     [self updateDynamicIconShortcutItems];
 
-    
-    NSURL *documentsURL = [[[NSFileManager defaultManager] URLsForDirectory:NSDocumentDirectory inDomains:NSUserDomainMask] lastObject];
-    NSLog(@"%@", documentsURL);
+#if DEBUG
+    NSLog(@"\n\nSimulator documents directory:\n\t%@\n\n",
+          [NSSearchPathForDirectoriesInDomains(NSDocumentDirectory, NSUserDomainMask, YES) lastObject]);
+#endif
 
     return YES;
 }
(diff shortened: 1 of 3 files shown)


montehurd:
@coreyfloyd @joewalsh this should be good to go.

Wikipedia/Code/AppDelegate.m
```diff
@@ -97,9 +97,10 @@ - (BOOL)application:(UIApplication*)application didFinishLaunchingWithOptions:(N
 
     [self updateDynamicIconShortcutItems];
 
-    
-    NSURL *documentsURL = [[[NSFileManager defaultManager] URLsForDirectory:NSDocumentDirectory inDomains:NSUserDomainMask] lastObject];
-    NSLog(@"%@", documentsURL);
+#if DEBUG
+    NSLog(@"\n\nSimulator documents directory:\n\t%@\n\n",
+          [NSSearchPathForDirectoriesInDomains(NSDocumentDirectory, NSUserDomainMask, YES) lastObject]);
+#endif
```
joewalsh: whoops, i made a similar change when using this to debug

joewalsh:
👍

PR #784, opened 2016-07-21
Saved Article Image Fix

joewalsh:


Wikipedia/Code/WMFImageController.swift
```diff
@@ -219,7 +219,11 @@ public class WMFImageController : NSObject {
     public func typedDiskDataForImageWithURL(url: NSURL?) -> WMFTypedImageData {
         if let url = url {
             let path = imageManager.imageCache.defaultCachePathForKey(cacheKeyForURL(url))
-            let mimeType: String? = NSFileManager.defaultManager().wmf_valueForExtendedFileAttributeNamed(WMFExtendedFileAttributeNameMIMEType, forFileAtPath: path)
+            var mimeType: String? = NSFileManager.defaultManager().wmf_valueForExtendedFileAttributeNamed(WMFExtendedFileAttributeNameMIMEType, forFileAtPath: path)
+            if mimeType == nil {
+                let fileURL = NSURL(fileURLWithPath: path, isDirectory: false)
+                mimeType = fileURL.pathExtension?.wmf_asMIMEType()
+            }
```
joewalsh: The migrated images don't have a MIME type on disk, use this fallback at request time.

Wikipedia/Code/WMFImageURLParsing.m
```diff
@@ -101,7 +115,7 @@ NSInteger WMFParseSizePrefixFromSourceURL(NSString* sourceURL)  __attribute__((o
 
     NSString* lastPathComponent = [sourceURL lastPathComponent];
     
-    if (WMFParseSizePrefixFromSourceURL(sourceURL) == NSNotFound || !WMFIsThumbURLString(sourceURL)) {
+    if (WMFParseSizePrefixFromSourceURL(sourceURL) == NSNotFound) {
```
joewalsh: WMFParseSizePrefixFromSourceURL now checks for WMFIsThumbURLString

Wikipedia/Code/WMFImageURLParsing.m
```diff
@@ -80,6 +87,13 @@ NSInteger WMFParseSizePrefixFromSourceURL(NSString* sourceURL)  __attribute__((o
     }
 }
 
+NSString* WMFOriginalImageURLStringFromURLString(NSString *URLString) {
+    if ([URLString containsString:@"/thumb/"]) {
+        URLString = [[URLString stringByDeletingLastPathComponent] stringByReplacingOccurrencesOfString:@"/thumb/" withString:@"/"];
+    }
+    return URLString;
+}
+
```
joewalsh: No functionality is changed, just moved to be shared.

Wikipedia/Code/WMFImageTag+TargetImageWidthURL.m
```diff
@@ -37,9 +37,7 @@ - (NSURL*)URLForTargetWidth:(NSUInteger)targetWidth {
             // path component, but only if we're actually dealing with a "thumb" scaled variant.
             // This check for "/thumb/" is needed in case the canonical image name happens to
             // start with "XXXpx-", as the last image on "enwiki > Geothermal gradient" does.
-            if ([tagSrc containsString:@"/thumb/"]) {
-                tagSrc = [[tagSrc stringByDeletingLastPathComponent] stringByReplacingOccurrencesOfString:@"/thumb/" withString:@"/"];
-            }
```
joewalsh: No functionality is changed, just refactored

Wikipedia/Code/SavedArticlesFetcher.m
```diff
@@ -170,11 +172,94 @@ - (void)downloadImageDataForArticle:(MWKArticle*)article failure:(WMFErrorHandle
...
+        NSError *removalError = nil;
+        if (![fileManager removeItemAtURL:imageFolderURL error:&removalError]) {
+            DDLogError(@"Error removing old image list image: %@", removalError);
+        }
+    }
+}
 
 - (void)fetchAllImagesInArticle:(MWKArticle*)article failure:(WMFErrorHandler)failure success:(WMFSuccessHandler)success {
+    WMF_TECH_DEBT_TODO(This legacy migration can be removed after enough users upgrade to 5.0.5)
+    [self migrateLegacyImagesInArticle:article];
```
coreyfloyd: I think we need a check here so we don't call `[cache permanentlyCacheImagesForArticle:article];` if we are going to start migrating

Wikipedia/Code/WMFImageController.swift
coreyfloyd: So the downside is that some images will be broken in the browser, right? Because we needed to do this for SVGs I believe, correct?

Wikipedia/Code/SavedArticlesFetcher.m
coreyfloyd: Or I guess you are doing it synchronously so it should be finished before any further logic is called - is that right?

joewalsh: That's right - it happens synchronously so it's finished before any further logic is called.

Wikipedia/Code/WMFImageController.swift
joewalsh: we need to do this for all image types that we read from disk. the updated fix addresses the problem with the migration. we might still want a fallback here using imageURL instead of fileURL (fileURL doesn't have an extension so this change never worked).

Wikipedia/Code/SavedArticlesFetcher.m
```diff
@@ -203,52 +203,34 @@ - (void)migrateLegacyImagesInArticle:(MWKArticle *)article {
...
-        //if this is the article image, skip over it because it should be preserved
-        BOOL isArticleImage = [article.imageURL isEqualToString:imageURLString];
-        if (isArticleImage) {
-            continue;
-        }
-        
         NSUInteger width = WMFParseSizePrefixFromSourceURL(imageURLString);
-        if (width != articleImageWidth) {
+        if (width != articleImageWidth && width != NSNotFound) {
             NSURL *imageURL = [NSURL URLWithString:imageURLString];
```
coreyfloyd: `URLWithString:` can return nil - we should probably bail here?

Wikipedia/Code/SavedArticlesFetcher.m
```diff
@@ -203,52 +203,34 @@ - (void)migrateLegacyImagesInArticle:(MWKArticle *)article {
         NSDictionary *imageDictionary = [NSDictionary dictionaryWithContentsOfURL:imagePlistURL];
         NSString *imageURLString = imageDictionary[@"sourceURL"];
```
coreyfloyd: Do we need a nil check here before proceeding?

Wikipedia/Code/SavedArticlesFetcher.m
```diff
@@ -203,52 +203,34 @@ - (void)migrateLegacyImagesInArticle:(MWKArticle *)article {
...
         
-        //if this is the article image, skip over it because it should be preserved
-        BOOL isArticleImage = [article.imageURL isEqualToString:imageURLString];
```
coreyfloyd: Where is this being handled now?

joewalsh: further down

Wikipedia/Code/SavedArticlesFetcher.m
```diff
@@ -203,52 +203,34 @@ - (void)migrateLegacyImagesInArticle:(MWKArticle *)article {
...
+                    DDLogError(@"Error removing legacy cached image: %@", removalError);
                 }
             }
         }
         
-        NSError *removalError = nil;
-        if (![fileManager removeItemAtURL:imageFolderURL error:&removalError]) {
-            DDLogError(@"Error removing old image list image: %@", removalError);
+        //if this is the article image, don't delete it because it should be preserved
+        BOOL isArticleImage = [article.imageURL isEqualToString:imageURLString];
```
joewalsh: here

Wikipedia/Code/SavedArticlesFetcher.m
```diff
@@ -163,18 +179,91 @@ - (void)fetchTitle:(MWKTitle*)title failure:(WMFErrorHandler)failure success:(WM
...
+    NSEnumerator *enumerator = [[NSFileManager defaultManager] enumeratorAtURL:imagesFolderURL includingPropertiesForKeys:@[NSURLIsDirectoryKey] options:NSDirectoryEnumerationSkipsSubdirectoryDescendants errorHandler:^BOOL(NSURL * _Nonnull url, NSError * _Nonnull error) {
+        DDLogError(@"Error enumerating image directory: %@", error);
+        return YES;
+    }];
+    
+    WMFImageController *imageController = [WMFImageController sharedInstance];
+    NSUInteger articleImageWidth = [[UIScreen mainScreen] wmf_articleImageWidthForScale];
+    NSFileManager *fileManager = [NSFileManager defaultManager];
+    
+    for (NSURL *imageFolderURL in enumerator) { //Enumerate each subfolder of the image folder. There is one subfolder per saved image.
```
montehurd: No need to walk the directory structure opening an Image.plist for each image... I think there's an index of all the image urls in Image**s**.plist:

screen shot 2016-07-21 at 5 41 36 pm

montehurd: I think it's still ok to blast those folders, though, to keep it idempotent. 

However, we probably don't want to blast this Image**s**.plist file until the article record is being deleted. This is because it is the only record we have of these 5.0.4 style image urls. The reason we still need to know about them is so we can clear the SDWebImage cache of these images when we're deleting article data. 

For example, I did a clean install on 5.0.4, loaded and saved obama article, stopped the app, updated to this PR, started app, the **un**saved obama and cleared history, then stopped the app. It appeared that the obama images were removed from SDWebImage's cache folder *except* for the old 5.0.4 images - but the Image**s**.plist file has those urls, so I think we can just make sure we try to also delete any images from that plist at the time we're clearing out other article image data and we should be good. 

I hope that makes sense...

montehurd:
@joewalsh i can re-test in the morning if any of my suggestions sound ok... heading out.

joewalsh:
@montehurd pushed changes that utilize the Images.plist for the migration and also utilize it to delete those legacy files on article deletion. I don't want to move Images.plist to LegacyImages.plist at this point just to minimize changes & amount of new code. I also eliminated the folder deletion for the same reason. After thinking about it, it's easier and safer to keep that data around at this point. We lose the "progress" of the migration that was gained by deleting image folders, but all that means is it will do some extra checks but we don't actually re-coopy any of the files.

PR #785, opened 2016-07-21
Fix crash on nil/non-url array by checking before we call the swift method

coreyfloyd:
Reject any non-NSUrls
Do not call with a nil/empty array

https://rink.hockeyapp.net/manage/apps/152731/app_versions/68/crash_reasons/129245027?type=crashes
https://rink.hockeyapp.net/manage/apps/152731/app_versions/68/crash_reasons/129253881?type=crashes

### DIFF
--- a/Wikipedia/Code/SavedArticlesFetcher.m
+++ b/Wikipedia/Code/SavedArticlesFetcher.m
@@ -177,7 +177,7 @@ static SavedArticlesFetcher* _articleFetcher = nil;
 
     NSArray<NSURL*>* URLs = [[article allImageURLs] allObjects];
     
-    [self.imageController cacheImagesWithURLsInBackground:URLs failure:failure success:success];
+    [self cacheImagesWithURLsInBackground:URLs failure:failure success:success];
 }
 
 - (void)fetchGalleryDataForArticle:(MWKArticle*)article failure:(WMFErrorHandler)failure success:(WMFSuccessHandler)success {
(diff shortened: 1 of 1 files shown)


Wikipedia/Code/SavedArticlesFetcher.m
```diff
@@ -177,14 +177,14 @@ - (void)fetchAllImagesInArticle:(MWKArticle*)article failure:(WMFErrorHandler)fa
...
-    [self.imageController cacheImagesWithURLsInBackground:URLs failure:failure success:success];
+    [self cacheImagesWithURLsInBackground:URLs failure:failure success:success];
 }
 
 - (void)fetchGalleryDataForArticle:(MWKArticle*)article failure:(WMFErrorHandler)failure success:(WMFSuccessHandler)success {
     WMF_TECH_DEBT_TODO(check whether on - disk image info matches what we are about to fetch)
     @weakify(self);
     [self fetchImageInfoForImagesInArticle:article failure:^(NSError *error) {
-        
+        failure(error);
```
joewalsh: this might have been missing before the conversion away from promise kit. seems like it should be called, but dunno if it has unintended consequences.

joewalsh:
👍

Wikipedia/Code/SavedArticlesFetcher.m
coreyfloyd: @joewalsh possibly - the big reason I fixed it is because in the case of your fix you are relying on the failure block being called so you can set the migration flag. 

If we don't add this the app might get in a state where the migration is never marked as complete.

montehurd:
👍

PR #786, opened 2016-07-24
hotfix for dispatch group crash

joewalsh:


Wikipedia/Code/WMFTaskGroup.m
```diff
@@ -0,0 +1,50 @@
...
+            self.count--;
+            dispatch_group_leave(self.group);
+        } else {
+            DDLogError(@"Mismatched leave for group: %@", self);
+        }
+    }
+}
+
+- (void)waitInBackgroundWithTimeout:(NSTimeInterval)timeout completion:(nonnull dispatch_block_t)completion {
+    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_BACKGROUND, 0), ^{
```
coreyfloyd: There is a block based API that we can use here so you don't have to do a synchronous `wait` on a background thread:
`dispatch_group_notify`

coreyfloyd:
@joewalsh left one comment - otherwise looks good to me. Lets get this merged in the morning.

Wikipedia/Code/WMFTaskGroup.m
montehurd: @coreyfloyd i think that's being used on line 47. this one is so you can also have a timeout on the group independent of the tasks in the group, right?

joewalsh: @coreyfloyd @montehurd correct - the synchronous wait is only for having a timeout.  `dispatch_group_notify` is used when there is no timeout. The timeout method isn't used in the codebase, I'd be OK with removing it. I added it so we could consider it, but I think there's more risk in it timing out early based on an arbitrary guess at timeout length. The real danger is a crash due to over-leaving the dispatch group.

coreyfloyd: Got it. Misread

PR #790, opened 2016-07-26
Fix missing image caused by proxy using incomplete url.

montehurd:
https://phabricator.wikimedia.org/T141405

"enwiki > Ptolemy (name)"

# Before
screen shot 2016-07-26 at 3 33 04 pm

# After
screen shot 2016-07-26 at 3 31 15 pm


coreyfloyd:
Why is the URL incomplete? I see that we are hard coding the domain to fix it... how do we know that is the right domain? Is this something we can fix in the HTML?

montehurd:
@coreyfloyd 

> Why is the URL incomplete? 

The url is not incomplete. It is a valid *relative* url. 

Browsers will resolve relative urls to full urls by the html file's path, or, if specified the "base" tag's href path.

The proxy server has neither of these, and so can't use either of these methods to resolve relative urls to full ones.

> I see that we are hard coding the domain to fix it... how do we know that is the right domain? 

Any relative urls in mobileview html will be relative to whatever mobileview html is relative to, which is, by definition, "wikipedia.org".

montehurd:
More background:

----

Here's one of the relative urls (note that the bold "originalSrc" value is *directly* from the mobileview html):

src="http://localhost:52491/B0750864-4008-4D64-8269-7EF33D71C52E/imageProxy?originalSrc=**/w/extensions/wikihiero/img/hiero_V4.png?e648c**"

Given that mobileview html is relative to the "wiki**p**edia.org" domain, the urls in the html it outputs are free to use urls relative to that domain.

----

Here's an example of the more common case (the majority of images living on commons) of using a full url when the domain is not "wiki**p**edia.org":

src="http://localhost:52491/B0750864-4008-4D64-8269-7EF33D71C52E/imageProxy?originalSrc=**//upload.wikimedia.org/wikipedia/commons/1/1c/Homer_British_Museum.jpg**"

The full url is not on a "wiki**p**edia.org" domain, so it had so specify "wiki**m**edia.org".

joewalsh:
@montehurd good catch. I'm concerned that the underlying WMFImageTags are still incorrect, leaving the "imageURLsFor..." methods on article also incorrect. I put up an alternate proposed fix that adds baseURL handling to WMFImageTag https://github.com/wikimedia/wikipedia-ios/pull/791 let me know what you think.

montehurd:
@joewalsh good point! closing...

PR #793, opened 2016-07-27
Convert WMFExploreViewController to UICollectionView & Custom Layout

joewalsh:


joewalsh:
@coreyfloyd @montehurd this is ready to be looked at. It'd be good to break this layout into a separate library, so I'd love feedback on how to make it more understandable as a standalone component.

joewalsh:
Known Issues:
1. Refresh control can appear over cells occasionally 
2. Scroll to top is broken

Wikipedia/Code/WMFCVLAttributes.h
```diff
@@ -0,0 +1,5 @@
+#import <UIKit/UIKit.h>
+
+@interface WMFCVLAttributes : UICollectionViewLayoutAttributes
```
coreyfloyd: @joewalsh what is the purpose of the new class?

joewalsh: so that it's there if we need to add custom layout attributes

joewalsh: also it's shorter to write than UICollectionViewLayoutAttributes, but I understand if neither of these are compelling arguments

Wikipedia/Code/WMFCVLInfo.h
```diff
@@ -0,0 +1,43 @@
...
+ */
+- (nonnull instancetype)initWithMetrics:(nonnull WMFCVLMetrics *)metrics NS_DESIGNATED_INITIALIZER;
+
+- (void)enumerateSectionsWithBlock:(nonnull void(^)(WMFCVLSection * _Nonnull section, NSUInteger idx, BOOL * _Nonnull stop))block;
+- (void)enumerateColumnsWithBlock:(nonnull void(^)(WMFCVLColumn * _Nonnull column, NSUInteger idx, BOOL * _Nonnull stop))block;
+
+- (nullable WMFCVLAttributes *)layoutAttributesForItemAtIndexPath:(nonnull NSIndexPath *)indexPath;
+- (nullable WMFCVLAttributes *)layoutAttributesForSupplementaryViewOfKind:(nonnull NSString *)elementKind atIndexPath:(nonnull NSIndexPath *)indexPath;
+
+- (void)layoutForBoundsSize:(CGSize)size withDelegate:(nullable id <WMFColumnarCollectionViewLayoutDelegate>)delegate collectionView:(nullable UICollectionView *)collectionView;
```
coreyfloyd: Why do we pass in the delegate here instead of on init or as a separate property?

joewalsh: it could be passed in at init or as a separate property. could also snapshot all of the required delegate metrics and pass those in.

Wikipedia/Code/WMFColumnarCollectionViewLayout.h
```diff
@@ -0,0 +1,24 @@
+#import <UIKit/UIKit.h>
+
+@class WMFCVLMetrics;
+
+/*!
+ @class        WMFColumnarCollectionViewLayout
+ @abstract     A WMFColumnarCollectionViewLayout organizes a collection view into columns grouped by section - all items from the same section will be in the same column.
+ @discussion   ...
+ */
+@interface WMFColumnarCollectionViewLayout : UICollectionViewLayout
```
coreyfloyd: Can you add Nullability attributes for this and the other classes?

coreyfloyd: I mean the "NS_ASSUME_NONNULL_BEGIN"/"END"

Wikipedia/Code/WMFColumnarCollectionViewLayout.m
```diff
@@ -0,0 +1,165 @@
...
+    return self.info.contentSize;
+}
+
+#pragma mark - Layout
+
+- (nullable NSArray<__kindof UICollectionViewLayoutAttributes *> *)layoutAttributesForElementsInRect:(CGRect)rect {
+    
+    NSMutableArray *attributesArray = [NSMutableArray array];
+    
+    [self.info enumerateSectionsWithBlock:^(WMFCVLSection * _Nonnull section, NSUInteger idx, BOOL * _Nonnull stop) {
```
coreyfloyd: Should this be encapsulated in the Info class so you can implement this method in one line similarly to the calls below?

Wikipedia/Code/WMFColumnarCollectionViewLayout.m
```diff
@@ -0,0 +1,165 @@
...
+    return [self.info layoutAttributesForSupplementaryViewOfKind:elementKind atIndexPath:indexPath];
+}
+
+- (nullable UICollectionViewLayoutAttributes *)layoutAttributesForDecorationViewOfKind:(NSString*)elementKind atIndexPath:(NSIndexPath *)indexPath {
+    return nil;
+}
+
+- (void)prepareLayout {
+    if (self.needsLayout || self.needsNewLayout) {
+        if (self.needsNewLayout) {
```
coreyfloyd: Why do we need the needs new layout flag? Could we just set the info to nil? And override the getter? Or does that cause consistency problems?

joewalsh: I'm moving the needsNewLayout flag and other changes into a follow on PR since they're going beyond parity with the table view - but yeah, the info should stay valid until this point. I'll try some other things in the follow on.

Wikipedia/Code/WMFColumnarCollectionViewLayout.m
joewalsh: sure

Wikipedia/Code/WMFColumnarCollectionViewLayout.h
joewalsh: is that preferred instead of specifying all of the nullability attributes?

coreyfloyd: I think this is fine… I just missed that you actually did the annotations just didn't need assume non null

Wikipedia/Code/WMFCVLInfo.h
coreyfloyd: I don't think you could because this method iterates over the delegate and calls `estimatedHeightForHeaderInSection:` for each section. Also that would use make a long method. I think having access to the delegate is fine.

It's ok to pass them in every time, I just wasn't sure why the delegate was being passed on each call while the metrics was being retained strongly on init.  I guess if the metrics change you need to create a new instance of the info class, but if the delegate or collection view change that isn't necessary? Either way, either stateless or immutable is best, so this is ok to me.

Wikipedia/Code/WMFCVLSection.h
```diff
@@ -0,0 +1,40 @@
...
+/*!
+ @class        WMFCVLSection
+ @abstract     A WMFCVLSection represents a section within a column of a WMFColumnarCollectionViewLayout. It handles adjusting the size and offset of items within the section.
+ @discussion   ...
+ */
+@interface WMFCVLSection : NSObject
+
+@property (nonatomic, readonly) NSInteger index;
+@property (nonatomic) CGRect frame;
+@property (nonatomic, weak, nullable) WMFCVLColumn *column;
```
coreyfloyd: This class doesnt use column internally - does it need to have the property?

I ask because I see the copy method that needs to iterate through and reset this property on every section.

Wikipedia/Code/WMFCVLInfo.h
```diff
@@ -0,0 +1,43 @@
...
+@property (nonatomic) CGSize contentSize;
+
+/*!
+ Initialize a new WMFCVLInfo with the given metrics.
+ @param metrics     The metrics for the WMFCVLInfo
+ @return A WMFCVLInfo with the given metrics
+ */
+- (nonnull instancetype)initWithMetrics:(nonnull WMFCVLMetrics *)metrics NS_DESIGNATED_INITIALIZER;
+
+- (void)enumerateSectionsWithBlock:(nonnull void(^)(WMFCVLSection * _Nonnull section, NSUInteger idx, BOOL * _Nonnull stop))block;
```
coreyfloyd: Is there an implied order for these enumerations? I assume its in order of index path, but not sure if that is actually true. 

Either way you may want to add a comment to these 2 methods to let the user know that it is in order or the order is undefined.

Related: 
Does idx == indexPath.section?

Wikipedia/Code/WMFCVLColumn.m
```diff
@@ -0,0 +1,88 @@
...
+    [self updateHeightWithDelta:deltaY];
+}
+
+- (void)offsetSectionsByDistance:(CGFloat)deltaY startingAfterSection:(WMFCVLSection *)afterSection invalidationContext:(WMFCVLInvalidationContext *)invalidationContext {
+    if (ABS(deltaY) == 0) {
+        return;
+    }
+    
+    __block BOOL update = NO;
+    [self enumerateSectionsWithBlock:^(WMFCVLSection * _Nonnull section, NSUInteger idx, BOOL * _Nonnull stop) {
```
coreyfloyd: Should we use `enumerateObjectsAtIndexes:…` to skip we don't want to update sections or is that not worth it?

Wikipedia/Code/WMFCVLAttributes.h
coreyfloyd: > also it's shorter to write than UICollectionViewLayoutAttributes,
😆 

Thats cool with me

Wikipedia/Code/WMFCVLInfo.m
```diff
@@ -0,0 +1,337 @@
...
+    }
+    
+    WMFCVLAttributes *attributes = section.items[itemIndex];
+    assert(attributes != nil);
+    return attributes;
+}
+
+- (nullable WMFCVLAttributes *)layoutAttributesForSupplementaryViewOfKind:(NSString *)elementKind atIndexPath:(NSIndexPath *)indexPath {
+    NSInteger sectionIndex = indexPath.section;
+    if (sectionIndex < 0 || sectionIndex >= self.sections.count) {
```
coreyfloyd: Why do we have all the < 0 checks? Being extra careful or did you actually see some?

Wikipedia/Code/WMFCVLSection.h
joewalsh: it’s for this lookup:

```
- (void)updateWithInvalidationContext:(nonnull WMFCVLInvalidationContext *)context delegate:(id <WMFColumnarCollectionViewLayoutDelegate>)delegate collectionView:(UICollectionView *)collectionView {
    if (context.boundsDidChange) {
        [self resetSections];
        [self layoutForBoundsSize:context.newBounds.size withDelegate:delegate collectionView:collectionView invalidationContext:context];
    } else if (context.originalLayoutAttributes && context.preferredLayoutAttributes) {
        UICollectionViewLayoutAttributes *originalAttributes = context.originalLayoutAttributes;
        UICollectionViewLayoutAttributes *preferredAttributes = context.preferredLayoutAttributes;
        NSIndexPath *indexPath = originalAttributes.indexPath;
        
        WMFCVLSection *invalidatedSection = self.sections[indexPath.section];
        WMFCVLColumn *invalidatedColumn = invalidatedSection.column;
```

There might be a better way to do it. You could ask each column if that section index is in it’s index set.

Wikipedia/Code/WMFCVLInfo.h
joewalsh: It is in section order, idx == section index.

Wikipedia/Code/WMFCVLInfo.m
coreyfloyd: Maybe we should encapsulate this logic in a method on NSArray? indexIsInBounds:?

joewalsh: extra careful / it was in the AdvancedCollectionViewLayout code

Wikipedia/Code/WMFCVLInfo.m
```diff
@@ -0,0 +1,337 @@
...
+            currentColumn.height += interSectionSpacing;
+        }
+        CGFloat y = currentColumn.height;
+        CGPoint sectionOrigin = CGPointMake(x, y);
+        
+        [currentColumn addSection:section];
+        
+        CGFloat sectionHeight = 0;
+        
+            NSIndexPath *supplementaryViewIndexPath = [NSIndexPath indexPathForRow:0 inSection:sectionIndex];
```
coreyfloyd: Extra indent?

Wikipedia/Code/WMFCVLColumn.m
joewalsh: you'd need to find the index to start at, so this seemed ok

Wikipedia/Code/WMFCVLInfo.m
```diff
@@ -0,0 +1,337 @@
...
+                [invalidatedHeaderIndexPaths addObject:supplementaryViewIndexPath];
+            }
+        }
+        
+        sectionHeight += headerHeight;
+        y += headerHeight;
+        
+        CGFloat itemX = x + sectionInsets.left;
+        CGFloat itemWidth = columnWidth - sectionInsets.left - sectionInsets.right;
+        for (NSInteger item = 0; item < [collectionView.dataSource collectionView:collectionView numberOfItemsInSection:sectionIndex]; item++) {
```
coreyfloyd: Should we cache the results of `[collectionView.dataSource collectionView:collectionView numberOfItemsInSection:sectionIndex]` - is that called on each iteration of the loop?

Wikipedia/Code/WMFCVLInfo.m
```diff
@@ -0,0 +1,337 @@
...
+                y += sectionInsets.top;
+            } else {
+                y += interItemSpacing;
+            }
+            
+            CGFloat itemHeight = 0;
+            NSIndexPath *itemIndexPath = [NSIndexPath indexPathForItem:item inSection:sectionIndex];
+            
+            if (item >= section.items.count) {
+                itemHeight = [delegate collectionView:collectionView estimatedHeightForItemAtIndexPath:[NSIndexPath indexPathForItem:item inSection:sectionIndex] forColumnWidth:columnWidth];
```
coreyfloyd: Can we just pass in the indexPath from above?

Wikipedia/Code/WMFCVLInfo.m
```diff
@@ -0,0 +1,337 @@
...
+                CGRect itemFrame = CGRectMake(itemX, y, itemWidth, itemHeight);
+                if (!CGRectEqualToRect(itemFrame, itemAttributes.frame)) {
+                    WMFCVLAttributes *newAttributes = [itemAttributes copy];
+                    newAttributes.frame = itemFrame;
+                    [section replaceItemAtIndex:item withItem:newAttributes];
+                    [invalidatedItemIndexPaths addObject:itemIndexPath];
+                }
+            }
+            
+            assert(itemHeight > 0);
```
coreyfloyd: Isn't there a reason to use NSAssert over assert? Does it log better or something?

joewalsh: NSAssert can make it clearer to someone using the code later the reason for the assertion failure because it takes a message. Also it throws an exception rather than just crashing.

Wikipedia/Code/WMFCVLInfo.m
```diff
@@ -0,0 +1,337 @@
...
+            [invalidatedColumn setSize:sizeToSet forFooterAtIndexPath:indexPath invalidationContext:context];
+        }
+        [self updateContentSizeWithInvalidationContext:context];
+        if (self.metrics.numberOfColumns == 1 && originalAttributes.frame.origin.y < collectionView.contentOffset.y) {
+            context.contentOffsetAdjustment = CGPointMake(0, context.contentSizeAdjustment.height);
+        }
+    }
+}
+
+- (void)layoutForBoundsSize:(CGSize)size withDelegate:(id <WMFColumnarCollectionViewLayoutDelegate>)delegate collectionView:(UICollectionView *)collectionView invalidationContext:(nullable WMFCVLInvalidationContext *)context {
```
coreyfloyd: This is a pretty long method - which I know this has all the calculation logic so there might not be a way around it. 

It would be nice to maybe break it up into some methods if possible. Or maybe just some light sprinkling of high level comments to help with scannability when reading through it.

Wikipedia/Code/WMFCVLInfo.m
```diff
@@ -0,0 +1,337 @@
...
+    CGSize oldContentSize = self.contentSize;
+    
+    [self updateContentSize];
+    
+    CGSize contentSizeAdjustment = CGSizeMake(self.contentSize.width - oldContentSize.width, self.contentSize.height - oldContentSize.height);
+    context.contentSizeAdjustment = contentSizeAdjustment;
+}
+
+- (void)updateWithInvalidationContext:(nonnull WMFCVLInvalidationContext *)context delegate:(id <WMFColumnarCollectionViewLayoutDelegate>)delegate collectionView:(UICollectionView *)collectionView {
+    if (context.boundsDidChange) {
```
coreyfloyd: Is this semantically being used the same way as `invalidateEverything`? If so, I guess you needed this because `invalidateEverything` is readonly?

Wikipedia/Code/WMFCVLColumn.m
coreyfloyd: Gotcha - since if you searched the array for the section you would have to iterate it anyways.

Wikipedia/Code/WMFCVLSection.h
coreyfloyd: Would that be better? That way you could remove the state? Or is that expensive? Is it tracked via indexSets?

coreyfloyd:
@joewalsh looks like all of the visual tests now are much wider than before. Is that expected? Should they be only 640 wide?

Wikipedia/Code/WMFColumnarCollectionViewLayout.m
coreyfloyd: 👍

coreyfloyd:
@joewalsh aside from the image updates this is all looking good to me. 

Breaking up the responsibilities into info/columns/sections/metrics seem like good abstractions that keep the code understandable. I was able to get through most of it pretty easy. The only comment regarding architecture was about that long method for the actual calculations. But I don't know if there is much to do there besides some comments.

@montehurd is going to put this through its paces in a bit.

coreyfloyd:
👍

PR #794, opened 2016-07-28
Find in page

montehurd:
untitled mov


Wikipedia/Code/FindInPageKeyboardBar.h
```diff
@@ -0,0 +1,24 @@
+#import <UIKit/UIKit.h>
+
+@class FindInPageKeyboardBar;
+
+@protocol FindInPageBarDelegate <NSObject>
```
coreyfloyd: Add WMF to these classes/protocols when you get a chance

Wikipedia/Code/FindInPageKeyboardBar.h
```diff
@@ -0,0 +1,24 @@
...
+
+- (void)findInPageTermChanged:(NSString *)text sender:(FindInPageKeyboardBar *)sender;
```
coreyfloyd: "Sender" is a terminology that is usually used with target/action or responder chain. With the delegate pattern, the name of the class is preferred, here I would use something like "keyboardBar".

Also with delegate methods, it is convention to always pass the object back as the the first parameter. See the [UITableViewDelegate](https://developer.apple.com/library/ios/documentation/UIKit/Reference/UITableViewDelegate_Protocol/index.html#//apple_ref/occ/intf/UITableViewDelegate) or the [UITextViewDelegate](https://developer.apple.com/library/ios/documentation/UIKit/Reference/UITextViewDelegate_Protocol/index.html#//apple_ref/doc/uid/TP40006897) protocols for an example pattern.

Wikipedia/Code/FindInPageKeyboardBar.m
```diff
@@ -0,0 +1,88 @@
...
+
+- (IBAction)didTouchClear{
+    [self.textField setText:@""];
+    [self.delegate findInPageTermChanged:@"" sender:self];
+    [self.delegate findInPageClearButtonTapped];
+    self.clearButton.hidden = YES;
+}
+
+- (void)setNumberOfMatches:(NSUInteger)numberOfMatches {
+    _numberOfMatches = numberOfMatches;
```
coreyfloyd: Do we need to store thee properties? Can we just update the view instead of storing the state?

Wikipedia/Code/FindInPageKeyboardBar.h
montehurd: @coreyfloyd done!

Wikipedia/Code/WMFArticleViewController.m
```diff
@@ -463,6 +490,21 @@ - (UIBarButtonItem*)shareToolbarItem {
...
 }
 
+- (UIBarButtonItem*)findInPageToolbarItem {
+    if (!_findInPageToolbarItem) {
+        @weakify(self);
+        _findInPageToolbarItem = [[UIBarButtonItem alloc] bk_initWithImage:[UIImage imageNamed:@"find-in-page"]
+                                                                   style:UIBarButtonItemStylePlain
+                                                                 handler:^(id sender){
+                                                                     @strongify(self);
+                                                                     if ([self canFindInPage]) { // Needed so you can't tap find icon when text size adjuster is onscreen.
```
coreyfloyd: If its not already done, you can set a property on the popover to not allow taps to the interface while it is visible. That way you do not have to do this check.

Wikipedia/Code/FindInPageKeyboardBar.m
```diff
@@ -0,0 +1,88 @@
...
+
+- (void)dealloc {
+    [[NSNotificationCenter defaultCenter] removeObserver:self];
+}
+
+- (instancetype)initWithCoder:(NSCoder *)coder{
+    self = [super initWithCoder:coder];
+    if (self) {
+        self.tintColor = [UIColor colorWithRed:0.3373 green:0.3373 blue:0.3373 alpha:1.0];
+        [[NSNotificationCenter defaultCenter] addObserver:self
```
coreyfloyd: Usually the delegate is used here when you own the object. Normally only "outside" observers use notifications.

Wikipedia/Code/FindInPageKeyboardBar.h
```diff
@@ -0,0 +1,24 @@
...
+- (void)findInPagePreviousButtonTapped;
+- (void)findInPageNextButtonTapped;
+
+@end
+
+@interface FindInPageKeyboardBar : UIInputView
+
+@property (weak, nonatomic) id<FindInPageBarDelegate> delegate;
+
+@property (strong, nonatomic) IBOutlet UITextField *textField;
```
coreyfloyd: Its nice to hide this from outside observers… this encourages outside users to "touch it".

If other objects need something from it, its nice to provide an API for them to get/do what they need so that this class can be wholly responsible for managing the text field and no other objects can do anything unexpected with it.

Wikipedia/Code/WebViewController.m
```diff
@@ -249,6 +262,142 @@ - (void)handleArticleStateScriptMessage:(NSString*)messageString {
...
+    [[self findInPageTextField] resignFirstResponder];
+    [self.view resignFirstResponder];
+    [self.webView evaluateJavaScript:@"window.wmf.findInPage.removeSearchTermHighlights()"
+                   completionHandler:nil];
+}
+
+- (void)clearFindInPage {
+    self.findInPageMatches = @[];
+    self.findInPageResultCursorIndex = -1;
+    [[self findInPageTextField] setText:@""];
```
coreyfloyd: Example of what I mentioned above. The web view controller is setting text on the textfield, but so does the input view. So it is touch to answer the question: "which object is responsible modifying the text of this text field?" Its always best to strive to make that a single object.

Wikipedia/Code/ViewWithFindInPageKeyboardBarInputAccessoryView.h
```diff
@@ -0,0 +1,8 @@
+#import <UIKit/UIKit.h>
+#import "FindInPageKeyboardBar.h"
+
+@interface ViewWithFindInPageKeyboardBarInputAccessoryView : UIView
```
coreyfloyd: What does this class do? Do we need a container?

Wikipedia/Code/WebViewController.m
```diff
@@ -249,6 +262,142 @@ - (void)handleArticleStateScriptMessage:(NSString*)messageString {
...
+}
+
+- (void)setFindInPageMatches:(NSArray *)findInPageMatches {
+    _findInPageMatches = findInPageMatches;
+    [self findInPageKeyboardBar].numberOfMatches = [findInPageMatches count];
+}
+
+#pragma FindInPageBar cursor
+
+- (void)setFindInPageResultCursorIndex:(NSInteger)findInPageResultCursorIndex {
```
coreyfloyd: Cursor is a little technical/abstract terminology. I know what you are building so I know you mean. But something higher level may read better. 

`selectFindInPageResultAtIndex:`

Wikipedia/Code/FindInPageKeyboardBar.h
montehurd: Fixed!

Wikipedia/Code/WebViewController.m
```diff
@@ -249,6 +262,142 @@ - (void)handleArticleStateScriptMessage:(NSString*)messageString {
...
+    _findInPageResultCursorIndex = findInPageResultCursorIndex;
+    [self findInPageKeyboardBar].currentCursorIndex = findInPageResultCursorIndex;
+}
+
+- (void)advanceFindInPageResultCursorIndexReversed:(BOOL)reversed {
```
coreyfloyd: A little nit picky, but having a method start with the word "advance" and end in "reversed" sounds weird/clunky and hurts readability.

I usually use a pair of methods to make the intent more clear like:
`selectNextFindInPageResult` and `selectPreviousFindInPageResult` 

These would usually call through to a shared method… 
`selectFindInPageResultAtIndex:` 
with something like: 
`[self selectFindInPageResultAtIndex:self.findInPageResultCursorIndex+1]`

Wikipedia/Code/WMFArticleViewController.m
montehurd: Added comment per our chat.

Wikipedia/Code/WebViewController.m
```diff
@@ -249,6 +262,142 @@ - (void)handleArticleStateScriptMessage:(NSString*)messageString {
...
+
+- (void)viewWillTransitionToSize:(CGSize)size withTransitionCoordinator:(id<UIViewControllerTransitionCoordinator>)coordinator {
+    [super viewWillTransitionToSize:size withTransitionCoordinator:coordinator];
+    self.disableMinimizeFindInPage = YES;
+    [coordinator animateAlongsideTransition:nil completion:^(id<UIViewControllerTransitionCoordinatorContext> context) {
+        self.disableMinimizeFindInPage = NO;
+    }];
+}
+
+- (void)setFindInPageMatches:(NSArray *)findInPageMatches {
```
coreyfloyd: Use one terminology results or matches

Wikipedia/Code/FindInPageKeyboardBar.m
montehurd: Good point. Fixed.

Wikipedia/Code/FindInPageKeyboardBar.m
montehurd: Fixed (iirc this was cruft from before I moved the layout to a storyboard)

Wikipedia/Code/FindInPageKeyboardBar.h
montehurd: Yeah this was the "leakiness" I'd mentioned. Fixed!

Wikipedia/Code/WebViewController.m
montehurd: Good point.

Wikipedia/Code/ViewWithFindInPageKeyboardBarInputAccessoryView.h
montehurd: Ok I just looked back into this to refresh my memory... 

This approach (redeclaring `inputAccessoryView` as `readwrite`) is roughly outlined in the `inputAccessoryView` documentation: https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIResponder_Class/index.html#//apple_ref/occ/instp/UIResponder/inputAccessoryView

It lets easily transition between having the keyboardBar "docked" at the bottom of the screen or sitting atop the keyboard.

Wikipedia/Code/WebViewController.m
montehurd: Fixed!

Wikipedia/Code/WebViewController.m
montehurd: I cleaned this up a bit. Still not perfect, but better.

Wikipedia/Code/WMFFindInPageKeyboardBar.m
```diff
@@ -80,4 +82,20 @@ - (BOOL)textFieldShouldReturn:(UITextField *)textField {
...
+- (void)show {
+    [self.textField becomeFirstResponder];
+}
+
+- (void)hide {
+    [self.textField resignFirstResponder];
+}
+
+- (void)reset {
+    [self.textField setText:@""];
```
coreyfloyd: Any reason not to use nil?

montehurd: oops!

Wikipedia/Code/WMFFindInPageKeyboardBar.m
```diff
@@ -16,6 +16,18 @@ @interface WMFFindInPageKeyboardBar() <UITextFieldDelegate>
 
 @implementation WMFFindInPageKeyboardBar
 
+- (void)awakeFromNib {
+    [super awakeFromNib];
+    [self hideCopyPasteIcons];
+}
+
+- (void)hideCopyPasteIcons {
```
coreyfloyd: We should probabaly check with Rita on this, but its standard and useful… not sure if we just rip them out.

montehurd: This may not be what u think - it causes an extra 3 icons to appear in a particularly poor place on iPad.

Wikipedia/Code/ViewWithFindInPageKeyboardBarInputAccessoryView.h
coreyfloyd: According to the referenced docs, you need a subclass of `UIResponder`. That means you can use `UIViewController` as well (it is part of the responder chain as well)

There is an example of accomplishing this task without the extra view here:
https://robots.thoughtbot.com/input-accessorizing-uiviewcontroller

Wikipedia/Code/WMFFindInPageKeyboardBar.m
montehurd: Reminder to self: rename this "hideUndoRedoIcons"

montehurd:
@coreyfloyd You were right - since UIViewController is a UIResponder subclass I didn't need to subclass the view controller's view. Fixed!

PR #795, opened 2016-07-28
Make sure pull to refresh always downloads the latest article 

coreyfloyd:
…and ignores the cache

- Also add a test to verify the behavior
- Use default timeouts for these tests (boyscout)

@montehurd verified that we already handle the case where no article is returned in the normal code path


joewalsh:
:+1:

PR #797, opened 2016-07-29
Fix Swift UInt crashing issue and NSArray+WMFExtensions.swift bugs.

montehurd:
https://phabricator.wikimedia.org/T141600

I threw this together pretty quickly... may have missed something, but pretty shocked to see our "safe" array methods weren't actually safe...

### DIFF
--- a/Wikipedia/Code/MWKSection+TOCItem.swift
+++ b/Wikipedia/Code/MWKSection+TOCItem.swift
@@ -37,10 +37,10 @@ extension MWKSection : TableOfContentsItem {
         }
     }
 
-    public var indentationLevel: UInt {
+    public var indentationLevel: Int {
         get {
-            if let level = toclevel?.unsignedIntegerValue where level > 1 {
-                return max(UInt(level - 2), 0)
+            if let level = toclevel?.integerValue where level > 1 {
+                return max((level - 2), 0)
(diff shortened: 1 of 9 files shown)


coreyfloyd:
@montehurd I see the APIs are changed to take `Int` instead of `UInt`. Did that solve an issue? `NSArray` always returns/takes an `NSUInteger` but funny enough Swift `NSArray` methods take `Int`:
https://developer.apple.com/library/mac/documentation/Cocoa/Reference/Foundation/Classes/NSArray_Class/#//apple_ref/occ/instm/NSArray/objectAtIndex:

Which is commented on in the Programming Guide:
> Use UInt only when you specifically need an unsigned integer type with the same size as the platform’s native word size. If this is not the case, Int is preferred, even when the values to be stored are known to be non-negative. A consistent use of Int for integer values aids code interoperability, avoids the need to convert between different number types, and matches integer type inference, as described in Type Safety and Type Inference.

However they still use NSuInteger for the Obj-C APIs… 

(I am assuming that the generated header for the updated methods now uses NSIntegers instead - correct?)


Meta comment: The commit message doesn't actually say how you fixed the bug. Commit messages are best when they say **how** you fixed the issue so its clear why you made the changes that you did.

PR #800, opened 2016-07-29
Fix crash caused by adjusting image view on a background thread

joewalsh:


montehurd:
👍

coreyfloyd:
@joewalsh should we be fixing this at another level? It seems something else is calling a method on UIImageView on the wrong thread. I would think that code should know better than to interact with a UIImageView on a background thread.

joewalsh:
@coreyfloyd the dispatch_async to the main thread is there for the UIView calls, it just didn't encompass the `wmf_cropContentsByVerticallyCenteringFrame` and `wmf_resetContentsRect`. Those methods interact with the view's layer, which should only occur on the main thread. This is the  only place `wmf_cropContentsByVerticallyCenteringFrame` is called. The other place to fix it would be to have the cropping code always callback on the main thread. The downside of this is if that you ever want to face crop purely in the background, it will always bounce back to the main thread. I would think it's the responsibility of the code using the face detection to async to the main thread after making the crop, as is the case with this change.

This is currently the top non-8badf00d crash in 5.0.5: https://rink.hockeyapp.net/manage/apps/152731/app_versions/70/crash_reasons/130361537

joewalsh:
@coreyfloyd here's where I made the change that introduced the crash: https://github.com/wikimedia/wikipedia-ios/commit/98497c3b287db957b7167abd1d310f8d581116fa

Previously, it used a promise resolver block which is always called on the main queue. I ensured the image set was happening on the main queue, but missed the layer frame alterations made by the methods outside of the async.

coreyfloyd:
> The other place to fix it would be to have the cropping code always callback on the main thread. The downside of this is if that you ever want to face crop purely in the background, it will always bounce back to the main thread. I would think it's the responsibility of the code using the face detection to async to the main thread after making the crop, as is the case with this change.

Yeah I don't think it is a good idea to do it there either. But shouldn't we re-dispatch where the completion handler is invoked? Instead of inside one of the methods invoked within the completion handler?

In this case, we have 2 places in this category where we invoke an async API. Seems like if we know that these success and failure blocks will not return on the main thread, Should we handle it at the source before any other methods are invoked? What do you think?

```
    [self.wmf_imageController fetchImageWithURL:imageURL failure:failure success:^(WMFImageDownload* _Nonnull download) {
        @strongify(self);
        if (!WMF_EQUAL([self wmf_imageURLToFetch], isEqual:, imageURL)) {
            failure([NSError cancelledError]);
        } else {
            [self wmf_setImage:download.image detectFaces:detectFaces animated:YES failure:failure success:success];
        }
    }];
```

```
    [self wmf_getFaceBoundsInImage:image failure:failure success:^(NSValue* bounds) {
        if (!WMF_EQUAL([self wmf_imageURLToFetch], isEqual:, imageURL)) {
            failure([NSError cancelledError]);
        } else {
            [self wmf_setImage:image faceBoundsValue:bounds animated:animated failure:failure success:success];
        }
    }];
```

joewalsh:
^ altering those works for me. I'll make the change

joewalsh:
@coreyfloyd ^ changes made

PR #805, opened 2016-08-01
Fixing migration of MWKTitles to URLs in the black list

coreyfloyd:
Was not converting from MWKTitle properly
Added a unit test for migration


joewalsh:
👍

PR #807, opened 2016-08-01
Fix the captcha presentation code

coreyfloyd:
https://phabricator.wikimedia.org/T141805

We need to differentiate between the first attempt and subsequent attempts to solve a captcha.

In reality, the user never enters text to solve the captcha on the first attempt. We immediately try to create an account with the supplied credentials.
This is because we want to get any errors back on the username (is it a duplicate, is it valid) before letting them solve the captcha.

So on the first attempt, we try and fail, once we get the captcha error, we present the captcha to the user
On subsequent attempts, if we fail it is because the user actually messed up the captcha, and in this case we need to request a new captcha for the user.

### DIFF
--- a/Wikipedia/Code/WMFAuthenticationManager.m
+++ b/Wikipedia/Code/WMFAuthenticationManager.m
@@ -92,6 +92,7 @@ NS_ASSUME_NONNULL_BEGIN
 
     self.captchaBlock = captcha;
     self.failBlock    = failure;
+    self.successBlock = nil;
 
     [self fetchCreationAuthManagerInfoWithSuccess:^(WMFAuthManagerInfo* _Nonnull info) {
         self.accountCreationAuthManagerInfo = info;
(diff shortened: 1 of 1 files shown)


coreyfloyd:
@joewalsh we still need to refactor this logic to use completion handlers all the way down so this logic will be much easier. While we have the delegate pattern with these old style fetchers we have to handle all this weird state.

coreyfloyd:
@joewalsh @montehurd to test this, try to create an account. Make sure the captcha is displayed.
Then enter a wrong captcha solution and make sure you see a new captcha. Then solve it correctly, the account should be created.

If the captcha image doesn't change, or no captcha is displayed, that means there is still a bug.

joewalsh:
👍

montehurd:
@coreyfloyd as part of the editing workflow i created an account and saved my changes but my changes were unfortunately not saved

montehurd:
@coreyfloyd checking develop...

coreyfloyd:
@montehurd if it isn't on develop, can you verify that it was an issue with the account by attaching the debugger?

montehurd:
@coreyfloyd before i even do that i'm trying to just save a change in general...

montehurd:
@coreyfloyd yet another false alarm :/ the test wiki site is *super* slow right now. after about a minute the edit *did* show up properly attributed to the account i created. the change just hadn't propagated by the time the edit window had disappeared

montehurd:
👍

PR #808, opened 2016-08-02
Fix "More Nearby" by bringing back the equivalent Table View Cells

joewalsh:


coreyfloyd:
@joewalsh are the other "more" view controllers table views as well - or are they collection views?

joewalsh:
@coreyfloyd all still table views. they can be converted.

PR #809, opened 2016-08-02
Collection View Layout & Crash Fixes

joewalsh:


Wikipedia/Code/WMFColumnarCollectionViewLayout.m
```diff
@@ -103,6 +103,14 @@ - (void)prepareLayout {
     [super prepareLayout];
 }
 
+- (CGPoint)targetContentOffsetForProposedContentOffset:(CGPoint)proposedContentOffset {
```
coreyfloyd: Do these methods need to be implemented if you aren't modifying the content offset?

Wikipedia/Code/WMFColumnarCollectionViewLayout.m
```diff
@@ -121,6 +121,7 @@ - (UICollectionViewLayoutInvalidationContext *)invalidationContextForBoundsChang
     WMFCVLInvalidationContext *context = (WMFCVLInvalidationContext *)[super invalidationContextForBoundsChange:newBounds];
     context.boundsDidChange = YES;
     context.newBounds = newBounds;
+    self.info = [self.info copy];
```
coreyfloyd: Did you find the that you need the copy?

Wikipedia/Code/WMFColumnarCollectionViewLayout.m
joewalsh: In AdvancedCollectionView they implemented the withScrollingVelocity version without modifying it so I figured they should be implemented. Seemed OK either way, but we'll eventually need to modify the values.

Wikipedia/Code/WMFColumnarCollectionViewLayout.m
joewalsh: no - was testing and accidentally pushed after making the other fixes. reverted the revert commit.

Wikipedia/Code/WMFColumnarCollectionViewLayout.m
```diff
@@ -145,12 +145,9 @@ - (UICollectionViewLayoutInvalidationContext *)invalidationContextForPreferredLa
 
 - (void)invalidateLayoutWithContext:(WMFCVLInvalidationContext *)context {
     assert([context isKindOfClass:[WMFCVLInvalidationContext class]]);
-    if (context.invalidateEverything) {
+    if (context.invalidateEverything || context.invalidateDataSourceCounts) {
```
coreyfloyd: @joewalsh can you explain how this fixes the no internet issue?

joewalsh: @coreyfloyd when there is no internet, the layout has fewer sections and items in it. After the copy, I need to remove the extraneous sections and items during updateWithInvalidationContext: since they cause a crash when returned to the collection view.

PR #810, opened 2016-08-02
Fix bug where the search results aren't updated on the first view of …

coreyfloyd:
…the search screen

https://phabricator.wikimedia.org/T139103

2 issues:
Didn't update the language bar when the search results re-appear
Didn't update the results weren't updated for the new language.

Now whenever the view comes back on screen, we reselect the last selected language
Then check to see if the language is different from the displayed results, and if it is we run the search again


joewalsh:
:+1:

PR #812, opened 2016-08-02
Log important statistics to help debug memory issues

coreyfloyd:
Logging on launch/background and when an article is pushed

Also adding in:
https://phabricator.wikimedia.org/T141629

For sending log files with our support messages

### DIFF
--- a/Wikipedia/Code/WMFAppViewController.m
+++ b/Wikipedia/Code/WMFAppViewController.m
@@ -284,7 +284,7 @@ static NSTimeInterval const WMFTimeBeforeRefreshingExploreScreen = 24 * 60 * 60;
 
 - (void)logImportantStatistics {
     NSUInteger historyCount       = [self.session.dataStore.userDataStore.historyList countOfEntries];
-    NSUInteger saveCount          = [self.session.dataStore.userDataStore.historyList countOfEntries];
+    NSUInteger saveCount          = [self.session.dataStore.userDataStore.savedPageList countOfEntries];
     NSUInteger exploreCount       = [self.exploreViewController numberOfSectionsInExploreFeed];
     UINavigationController* navVC = [self navigationControllerForTab:self.rootTabBarController.selectedIndex];
     NSUInteger stackCount         = [[navVC viewControllers] count];
--- a/Wikipedia/Code/WMFAppViewController.m
+++ b/Wikipedia/Code/WMFAppViewController.m
@@ -257,6 +257,8 @@ static NSTimeInterval const WMFTimeBeforeRefreshingExploreScreen = 24 * 60 * 60;
     if (FBTweakValue(@"Alerts", @"General", @"Show message on launch", NO)) {
         [[WMFAlertManager sharedInstance] showAlert:@"You have been notified" sticky:NO dismissPreviousAlerts:NO tapCallBack:NULL];
     }
+    DDLogWarn(@"Resuming… Logging Important Statistics");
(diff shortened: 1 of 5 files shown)


Wikipedia/Code/WMFAppViewController.m
```diff
@@ -275,6 +280,21 @@ - (void)didReceiveMemoryWarning {
     [[[SessionSingleton sharedInstance] dataStore] clearMemoryCache];
 }
 
+#pragma mark - Logging
+
+- (void)logImportantStatistics {
+    NSUInteger historyCount       = [self.session.dataStore.userDataStore.historyList countOfEntries];
+    NSUInteger saveCount          = [self.session.dataStore.userDataStore.historyList countOfEntries];
```
montehurd: you have "historyList" twice

Wikipedia/Code/WMFHelpViewController.m
```diff
@@ -62,6 +63,10 @@ - (void)sendEmail {
         [vc setSubject:[WMFSettingsEmailSubject stringByAppendingString:[WikipediaAppUtils versionedUserAgent]]];
         [vc setToRecipients:@[WMFSettingsEmailAddress]];
         [vc setMessageBody:[NSString stringWithFormat:@"\n\n\n\nVersion: %@", [WikipediaAppUtils versionedUserAgent]] isHTML:NO];
+        NSData* data = [[DDLog wmf_currentLogFile] dataUsingEncoding:NSUTF8StringEncoding];
+        if (data) {
+            [vc addAttachmentData:data mimeType:@"text/plain" fileName:@"Log Data.txt"];
```
montehurd: maybe name "Item Counts.txt" so we're less tempted to use this a catch-all for log info?

coreyfloyd: @montehurd it is all the log info, not just the counts

coreyfloyd: @montehurd sorry this was a second ticket just putting on the same PR:

https://phabricator.wikimedia.org/T141629

Wikipedia/Code/WMFAppViewController.m
coreyfloyd: Fixed

montehurd:
@coreyfloyd oh gotcha. should we get the @"Log Data.txt" string from a common var then? (i thought it was newly added)

coreyfloyd:
@montehurd it only used in one place for the title of the file we send to ourselves, so I don't think we need to make a variable out of it

joewalsh:
:+1:

Wikipedia/Code/WMFExploreViewController.m
```diff
@@ -154,6 +154,10 @@ - (WMFExploreSectionControllerCache*)sectionControllerCache {
     return _sectionControllerCache;
 }
 
+- (NSUInteger)numberOfSectionsInExploreFeed{
```
montehurd: Oh were weren't we gonna standardize on signed integers?

montehurd:
👍

Wikipedia/Code/WMFExploreViewController.m
coreyfloyd: Did we decide to do that in Obj-C as well?

montehurd: I think we prob should? To ease future Swift inter-op and/or conversion to Swift...

PR #813, opened 2016-08-03
Add missing template value for "CFBundleDisplayName"

montehurd:


montehurd:
merging so @niedzielski can run a TWN sync test - we're trying to see what's caused https://phabricator.wikimedia.org/T141752 and he needs to kick off a TWN sync

montehurd:
@coreyfloyd @joewalsh Hey guys sorry I merged this w/o review. Stephen and I were trying to debug https://phabricator.wikimedia.org/T141752. 

Background: The phab ticket ^ says the Greek "InfoPlist.strings" strings file wasn't able to be parsed, and was breaking TWN so NikeRabbit had to disable TWN automatic imports for wikipedia-iso. 

The Problem: When we add a localization, as we did recently with Greek, a copy is made of the English "Localizable.strings" and "InfoPlist.strings" files, but when we did this for Greek we didn't add the correct "InfoPlist.strings".

I pinged NikeRabbit to try to re-enable our auto-imports, which should hopefully not break now.

I am also adding a test to ensure that each plist lang has the same keys in its "InfoPlist.strings" that the English file does, which will help us catch this issue in the future.

PR #815, opened 2016-08-03
build: Update PROD_CSS_PREFIX to wikipedia.org

Krinkle:
The internal bits.wikimedia.org domain has been deprecated and
is on its way to be decommissioned.

See https://phabricator.wikimedia.org/T107430 for more info.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,7 +94,7 @@ exec-check:  ##Check that executable dependencies are installed
 web: ##Make web assets
 web: css grunt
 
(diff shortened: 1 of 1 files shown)


coreyfloyd:
These responses are identical:

https://en.wikipedia.org/w/load.php?debug=false&lang=en&only=styles&skin=vector&modules=skins.minerva.base.reset|skins.minerva.content.styles|

https://bits.wikimedia.org/en.wikipedia.org/load.php?debug=false&lang=en&only=styles&skin=vector&modules=skins.minerva.base.reset|skins.minerva.content.styles|

Merging

PR #819, opened 2016-08-08
remove null objects from script dictionaries

joewalsh:


joewalsh:
https://rink.hockeyapp.net/manage/apps/152731/app_versions/72/crash_reasons/131698821

PR #821, opened 2016-08-08
set GCDWebServer log level to warning

joewalsh:
to reduce log spam, make debugging on older devices more tolerable


montehurd:
👍

PR #824, opened 2016-08-09
Remove cruft methods

montehurd:


Wikipedia/assets/index.js
```diff
@@ -304,14 +304,6 @@ function isEndnote( href ) {
     return href.indexOf("#endnote_") > -1;
 }
 
-function isCitation( href ) {
```
montehurd: Both of these methods already exist above.

Wikipedia/assets/index.js
```diff
@@ -16,13 +16,6 @@ function stringEndsWith(str, suffix) {
     return str.indexOf(suffix, str.length - suffix.length) !== -1;
 }
 
-function getZoomLevel() {
```
montehurd: @coreyfloyd this buggy method was causing a TOC bug in @joewalsh's iPad article refactor.

joewalsh:
👍

PR #825, opened 2016-08-10
Fix crash bug due to image decompression memory consumption

coreyfloyd:
Also remove anything that was dowloaded in the background from the memory cache

### DIFF
--- a/Wikipedia/Code/WMFImageController.swift
+++ b/Wikipedia/Code/WMFImageController.swift
@@ -73,13 +73,13 @@ public class WMFImageController : NSObject {
         let downloader = SDWebImageDownloader.sharedDownloader()
         let cache = SDImageCache.wmf_appSupportCacheWithNamespace(defaultNamespace)
         let memory = NSProcessInfo.processInfo().physicalMemory
+        //Don't enable this, it causes crazy memory consumption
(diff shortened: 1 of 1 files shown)


coreyfloyd:
https://phabricator.wikimedia.org/T142602

coreyfloyd:
accidentally merged this to dev already… 😞

Wikipedia/Code/WMFImageController.swift
```diff
@@ -454,6 +454,7 @@ extension WMFImageController {
             }
 
             let success = { (didCache: Bool) in
+                self.imageManager.wmf_removeImageForURL(imageURL, fromDisk: false)
```
montehurd: Does this mean if we've successfully cached an image, we're then going to remove it from the cache?

joewalsh:
👍

Wikipedia/Code/WMFImageController.swift
montehurd: Ohh that's just removing it from the mem cache... gotcha.

montehurd: Although... if we're not using the mem cache is there any point in using SDWebImage ;)

PR #827, opened 2016-08-11
reduce memory pressure of saved article fetch

joewalsh:


montehurd:
👍

coreyfloyd:
@joewalsh can you resubmit this against Dev. I'll ask Josh but I suspect we won't hold the release for this since it primarily affects users with a large number of saved pages.

montehurd:
ship it! :)

montehurd:
just really hoping we don't get those same users immediately posting 1 star reviews saying crashed after startup if we can help it :/

montehurd:
@joewalsh corey had mentioned maybe creating a "Save 1500 article" scheme... that would be cool

joewalsh:
@montehurd @coreyfloyd my only concern is that there could be a decent chunk of users with this many saved articles. Also, even with 200 random saved articles, memory use jumped to 240MB. This is compounded for longer articles, as the main issue is several copies of the article HTML in memory. As far as I can tell, the formula is roughly ArticleCount * AvgArticleHTMLSize * AvgImagesPerArticle since it's bringing in a new copy of the article HTML for every iteration of the article image save loop.

PR #829, opened 2016-08-12
Handle wikipedia:// urls

jberkel:
* `wikipedia://explore`
* `wikipedia://saved`
* `wikipedia://history`
* `wikipedia://en.wikipedia.org/w/index.php?search=dog`
* `wikipedia://en.wikipedia.org/wiki/Foo`

https://phabricator.wikimedia.org/T141578

### DIFF
--- a/Wikipedia/Code/SavedPageSpotlightManager.swift
+++ b/Wikipedia/Code/SavedPageSpotlightManager.swift
@@ -3,25 +3,29 @@ import UIKit
 import MobileCoreServices
 import CoreSpotlight
 
-@available(iOS 9.0, *)
 
-public extension CSSearchableItemAttributeSet {
-    
-    public class func attributes(article: MWKArticle) -> CSSearchableItemAttributeSet {
-        
+public extension NSURL {
+    @available(iOS 9.0, *)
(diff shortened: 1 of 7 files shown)


Wikipedia/Code/SavedPageSpotlightManager.swift
```diff
@@ -3,25 +3,29 @@ import UIKit
...
 
-@available(iOS 9.0, *)
 
-public extension CSSearchableItemAttributeSet {
-    
-    public class func attributes(article: MWKArticle) -> CSSearchableItemAttributeSet {
-        
+public extension NSURL {
+    @available(iOS 9.0, *)
+    func searchableItemAttributes() -> CSSearchableItemAttributeSet {
```
coreyfloyd: This should probably only work for Wikipedia Article URLs.
You can check the `wmf_isWikiResource` property of the URL to make sure that is true.

coreyfloyd:
@jberkel looks good!

Left one comment about urls… otherwise I think its good to go.

jberkel:
Disclaimer: I've only tested this on 9.x. There was one change in the AppDelegate, so it might make sense to test briefly on 8.x.

PR #830, opened 2016-08-12
Fix layout glitch by turning off css animation.

montehurd:
https://phabricator.wikimedia.org/T141876

Tweaks to color from Rita.


joewalsh:
👍

PR #832, opened 2016-08-12
iPad Article Layout

joewalsh:


coreyfloyd:
@joewalsh putting it through its paces… so far looking really nice

joewalsh:
Found a bug with highlighting the footer sections... looking into a fix

coreyfloyd:
@joewalsh is the spacing of the button bar items right?

They look a little more spread out in the mocks

coreyfloyd:
(we should probably spread them out a little further even so)

joewalsh:
fixed - good catch

Wikipedia/Code/WebViewController.m
```diff
@@ -335,6 +336,23 @@ - (void)minimizeFindInPage {
     }
 }
 
+- (CGFloat)marginWidthForSize:(CGSize)size {
```
montehurd: @joewalsh Should this be marginWidthForWidth? Height isn't used and not sure we'd ever need to consider it...

PR #834, opened 2016-08-12
Increase max random items in explore to 10.

montehurd:
https://phabricator.wikimedia.org/T140779

I've been testing this by setting the explore VC's view layer scale to 0.5 so I can see more items onscreen at once.

I set my laptop's calendar back a few days, then restarted the app. Then moved the calendar one day forward and restarted the app, then repeated this step a few times. 

Problems I need to fix yet:
- [x] even though multiple random entries are now appearing in the feed they all show the same article after restart and this article is different after each restart (individual articles are not being remembered)
- [x] tapping refresh on any one of the random items only causes the 1st random item to refresh (even if that's not the one you tapped refresh on)

^ these shouldn't be to tricky to fix, but I'll need more time Monday.

### DIFF
--- a/Wikipedia/Code/WMFExploreSection.m
+++ b/Wikipedia/Code/WMFExploreSection.m
@@ -238,10 +238,10 @@ NS_ASSUME_NONNULL_BEGIN
         case WMFExploreSectionTypeMostRead:
         case WMFExploreSectionTypeNearby:
         case WMFExploreSectionTypePictureOfTheDay:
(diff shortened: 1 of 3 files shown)


montehurd:
@coreyfloyd basics are working, with the exception of the 2 bugs I noted above. Will track down next week.

joewalsh:
@montehurd it looks like this introduced a warning after removing the use of `WMFTimeBeforeRefreshingRandom`.

PR #836, opened 2016-08-15
Clang Format

joewalsh:


WikipediaUnitTests/Code/SavedArticlesFetcherTests.h
```diff
@@ -1,86 +0,0 @@
-//
```
montehurd: Oops looks like it gutted a couple test files...

PR #838, opened 2016-08-15
iPad Article View Refinements

joewalsh:
- Pillbox for narrow images when ToC is collapsed
- Save state of ToC across articles
- Smoother animation


www/less/widenImages.less
```diff
@@ -4,31 +4,9 @@
     display: block;
     margin-left: auto;
     margin-right: auto;
-}
-
-@media (orientation: portrait) {
-    .wideImageOverride {
```
montehurd: I don't think we should remove these... It breaks the image widening javascript logic and makes images take up way too much of the screen on landscape on phones...

montehurd: Before:
before mov

After:
after mov

montehurd: See the `shouldWidenImage` JS function for the other things which are related to the image widening logic.

www/less/widenImages.less
```diff
@@ -4,31 +4,9 @@
...
-        width:100% !important;
-        max-width:100% !important;
-        height:auto !important;
-    }
-
-    /* Tablet override */
-    @media (min-device-width: 768px) {
-        .wideImageOverride {
-            width:60% !important;
```
montehurd: Perhaps you'll get the same effect if you just change 60% to 100% here... or just remove lines 17 to 23 leaving the rest...

montehurd: That would leave in place the landscape narrowing and prevent any of the widening JS logic from breaking.

www/less/widenImages.less
joewalsh: wideImageOverride isn't removed, just the @ media variants. I forgot about the landscape phone (was thinking it was portrait only) I'll update so that it remains at 50% for the landscape phone.

montehurd: ah I forgot the base implementation was at the top!

joewalsh:
@montehurd 1b1684d restores 50% width for landscape phones. Even though I kept the iPad landscape width at 100%, I could be convinced that should be bumped down to 80% or the margins should be wider.

www/less/widenImages.less
```diff
@@ -29,6 +20,15 @@
         max-width:50% !important;
         height:auto !important;
     }
+	
+    /* Tablet override */
+    @media (min-device-width: 768px) {
+        .wideImageOverride {
+            width:100% !important;
```
montehurd: I think the css isn't quite right now - if you drop these values to 50% and run `make grunt` the image width isn't taking on landscape iPad...

montehurd:
I added some screenshots and comments to the Phab ticket so Rita can see the no-margin issue.

montehurd:
Getting a crash if I start when the simulator is in landscape. It's the assert on line 221 of WMFCLVInfo.m

joewalsh:
@montehurd I've reverted the article image width CSS changes as I realized it wasn't a part of the article image changes (they only applied to the lead article). I filed a new ticket here: https://phabricator.wikimedia.org/T143087

I'm unable to repro the assertion crash. Which device is it on? Are there any visual issues if you comment out the assertion?

PR #843, opened 2016-08-16
iOS 10 Project Update

coreyfloyd:
- Migrating to Swift 2.3
- Migrating to iOS 10 SDK
- Migrating to iOS 9 minimum deployment target (removing support for iOS 8)
- Migrating to Xcode 8 Project settings
- Migrating to Cocoapods 1.x and updating Podfile
- Moving back to frameworks for dependencies
- Remove Carthage from tool chain
- Removing OAStackView (no longer needed without iOS 8)

- [ ] Fix Tweaks crash


montehurd:
@coreyfloyd since this diff's gonna be hefty anyway, would u mind updating GCDWebServer too?

Podfile
```diff
@@ -51,7 +51,6 @@ abstract_target 'Foundation' do
     pod 'GCDWebServer', '~> 3.3'
 
     # Views
-    pod 'OAStackView', :git => 'git@github.com:wikimedia/OAStackView.git'
```
montehurd: Woohoo!

coreyfloyd:
@montehurd I don't want to update anything more than is necessary - its already been a bit precarious getting this to build with all the SDK/Swift/Toolchain changes.

If there are some important changes in GCDWebserver you think we need, we can do it in a separate PR after this one.

montehurd:
@corefloyd ah makes sense

coreyfloyd:
@joewalsh @montehurd travis for Xcode 8 is still hosed. The tests are passing locally. I'm trying to see if I can get Jenkins to do it now… I need to see if I can get iOS 9 and iOS 10 builds going side by side.

PR #844, opened 2016-08-17
Fix Article Footer Width on iPad

joewalsh:


joewalsh:
This was previously approved and merged, but not into develop.

PR #853, opened 2016-08-20
Made top nearby section always use current location.

montehurd:
https://phabricator.wikimedia.org/T133721


Wikipedia/Code/WMFNearbySectionController.m
```diff
@@ -145,14 +158,50 @@ - (NSString*)analyticsContentType {
...
     }
 
-    return [self fetchTitlesForLocation:self.location];
+    if ([self.date isToday]){
+
+        dispatchOnMainQueueAfterDelayInSeconds(0.1, ^{
+            [self.currentLocationManager startMonitoringLocation];
+        });
+        
+        return [[AnyPromise alloc] initWithResolver:&currentLocationResolver].then(^(PMKResolver resolve) {
```
montehurd: @coreyfloyd Wondering if I should just use https://github.com/mxcl/PromiseKit/blob/master/Categories/CoreLocation/CLLocationManager%2BAnyPromise.h instead of this new `currentLocationManager`...

PR #856, opened 2016-08-22
Be sure to unsubscribe from notifications to prevent crashing

coreyfloyd:
https://rink.hockeyapp.net/manage/apps/152731/app_versions/75/crash_reasons/133142162


montehurd:
👍

PR #857, opened 2016-08-24
Disable feed section header peeks. Is there other work-around/fix?

montehurd:
https://phabricator.wikimedia.org/T143570

Formerly a peek on a section header would cause the first item in that section to peek:

peek mov

This seems to be broken now, so I disabled header peeking since it was showing the wrong content. 

@joewalsh If there's a quick fix feel free to push a commit.


Wikipedia/Code/WMFExploreViewController.m
```diff
@@ -922,6 +922,11 @@ - (void)sectionSchema:(WMFExploreSectionSchema*)schema didRemoveSection:(WMFExpl
 - (nullable UIViewController*)previewingContext:(id<UIViewControllerPreviewing>)previewingContext
                       viewControllerForLocation:(CGPoint)location {
     NSIndexPath* previewIndexPath                     = [self.collectionView indexPathForItemAtPoint:location];
```
montehurd: Before the collection view refactor, I think the table view version of "indexPathForItemAtPoint" would actually return an indexPath if you tried to peek a header. Were our headers actually just regular cells then?

PR #859, opened 2016-08-25
Fix save button not working

coreyfloyd:
Notifications were not being subscribed to because the designated initializer was not being called

Fix by overriding init


joewalsh:
👍

PR #861, opened 2016-08-26
Refactor DynamicHeightPopover methods to also work with CGRect.

montehurd:
Follow-up to https://phabricator.wikimedia.org/T143278

Still does the UIBarButtonItem popovers, but keep watching the gif below and you'll see it can now show messages at any location too:

refactor mov


montehurd:
( For reference here is the testing code i used to make the gif: https://github.com/montehurd/apps-ios-wikipedia/commit/a52c516bc1c28a265f9367abb090350227a08b1c )

coreyfloyd:
👍 nice

PR #869, opened 2016-09-01
Make Wikipedia Zero "on / off" messages appear at appropriate times.

montehurd:


montehurd:
@coreyfloyd @joewalsh I think this is ready for review. Will test a bit more then remove WIP.

montehurd:
About to push up some naming changes which should make understanding the Zero bits easier in the future...

Wikipedia/Code/WMFZeroConfigurationManager.m
```diff
@@ -9,10 +9,9 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-NSString *const WMFIsZeroRatedDidChange = @"WMFIsZeroRatedDidChange";
-
-NSString *const ZeroOnDialogShownOnce = @"ZeroOnDialogShownOnce";
-NSString *const ZeroWarnWhenLeaving = @"ZeroWarnWhenLeaving";
+NSString *const WMFIsZeroRatedChanged = @"WMFZeroDispositionDidChange";
```
montehurd: I reverted this string for backward compatibility.

montehurd:
Renamed a bunch of bits and rebased on top of develop changes. Re-adding WIP and heading to office to re-test.

Wikipedia/Code/UIViewController+WMFOpenExternalUrl.m
```diff
@@ -88,8 +88,8 @@ - (void)wmf_presentExternalUrlAsSFSafari:(NSURL *)url {
     [self presentViewController:[[SFSafariViewController alloc] initWithURL:url] animated:YES completion:nil];
 }
 
-- (BOOL)isPartnerInfoConfigValid:(WMFZeroMessage *)msg {
-    return msg.partnerInfoText != nil && msg.partnerInfoUrl != nil;
+- (BOOL)isPartnerInfoConfigValid:(WMFZeroConfiguration *)config {
```
coreyfloyd: Can we move this to a method on `WMFZeroConfiguration` instead? Its functionality fits better there than being a `NSURL` method in the `WMFOpenExternalURL` category

Wikipedia/Code/WMFZeroConfigurationManager.m
```diff
@@ -0,0 +1,166 @@
...
+    return [self.zeroConfigurationFetcher fetchZeroConfigurationForSiteURL:[[[MWKLanguageLinkController sharedInstance] appLanguage] siteURL]];
+}
+
+- (void)updateZeroRatingAndZeroConfigurationForResponseHeadersIfNecessary:(NSDictionary *)headers {
+    NSAssert(headers != nil, @"Expecting response headers.");
+    if(!headers){
+        return;
+    }
+
+    bool zeroEnabled = self.isZeroRated;
```
coreyfloyd: should be `BOOL` - `bool` is standard a C type where `BOOL` is a typedef that Apple defines

Wikipedia/Code/WMFZeroConfigurationManager.m
```diff
@@ -0,0 +1,166 @@
...
+ *   rating is not presently enabled. (It would be nice if the query fetching the
+ *   zeroConfiguration returned an "enabled" key/value, but it doesn't - it nils out the
+ *   values instead apparently.) So in the nil message case we set isZeroRated "NO".
+ */
+- (void)fetchZeroConfigurationAndSetIsZeroRatedIfNecessary {
+    
+    // Note: don't nil out self.zeroConfiguration in this method
+    // because if we do we can't show its exit message strings!
+
+    //TODO: ensure thread safety so we can do this work off the main thread...
```
coreyfloyd: Not sure about this comment - what needs to move to a background thread? It doesn't look like any code on line 93 or 94 depends on the main thread.

Also line 99 is not guaranteed to be invoked on the main thread… which to me looks to be the only cause for concern of thread safety. 

In this simple case, you can always use `@synchronized(self){/*code that needs to be executed synchronously*/}` to make sure you do not access from multiple threads at the same time. I would recommend this approach for all the lines where you manipulate the `isZeroRated` ivar.

Related: Why do we need to re dispatch to the main queue below and then immediately invoking a promise network call (which immediately starts an asynchronous network call). I think this can be safely removed.

Wikipedia/Code/WMFZeroConfigurationManager.m
```diff
@@ -0,0 +1,166 @@
...
+ *
+ * - If the fetched zeroConfiguration has a nil "message" string, this means even though
+ *   there was a header, leading us to believe this network was Zero rated, its Zero
+ *   rating is not presently enabled. (It would be nice if the query fetching the
+ *   zeroConfiguration returned an "enabled" key/value, but it doesn't - it nils out the
+ *   values instead apparently.) So in the nil message case we set isZeroRated "NO".
+ */
+- (void)fetchZeroConfigurationAndSetIsZeroRatedIfNecessary {
+    
+    // Note: don't nil out self.zeroConfiguration in this method
```
coreyfloyd: Instead of this comment, why not remove the `nullable` attribute from the zeroConfiguration property? If we don't have one on launch since it needs to be downloaded from the network, you can make yourself a "default" one. Try to favor using language constructs to enforce behavior over comments.

This will be nice to hold some of the default logic you have if a the configuration is nil like: WMFAppViewController line 884

Wikipedia/Code/WMFZeroConfigurationManager.m
coreyfloyd: What thread am I on? This is invoked by the networking stack right?

Wikipedia/Code/WMFZeroConfigurationManager.m
```diff
@@ -0,0 +1,166 @@
...
+
+- (BOOL)hasChangeHappenedToCarrier:(NSString*)xCarrier orCarrierMeta:(NSString*)xCarrierMeta {
+    return !(
+             [self isNullableString:self.previousPartnerXCarrier equalToNullableString:xCarrier]
+             &&
+             [self isNullableString:self.previousPartnerXCarrierMeta equalToNullableString:xCarrierMeta]
+             );
+}
+
+- (BOOL)isNullableString:(nullable NSString*)stringOne equalToNullableString:(nullable NSString*)stringTwo {
```
coreyfloyd: why does `-[NSString isEqual:]` not work here?

Wikipedia/Code/WMFZeroConfigurationManager.m
```diff
@@ -0,0 +1,166 @@
...
+    
+    NSString* xCarrierFromHeader = [headers objectForKey:WMFZeroXCarrier];
+    bool hasZeroHeader = (xCarrierFromHeader != nil);
+    if (hasZeroHeader) {
+        NSString* xCarrierMetaFromHeader = [headers objectForKey:WMFZeroXCarrierMeta];
+        if ([self hasChangeHappenedToCarrier:xCarrierFromHeader orCarrierMeta:xCarrierMetaFromHeader]) {
+            self.previousPartnerXCarrier = xCarrierFromHeader;
+            self.previousPartnerXCarrierMeta = xCarrierMetaFromHeader;
+            [self fetchZeroConfigurationAndSetIsZeroRatedIfNecessary];
+        }
```
coreyfloyd: no else statement?

Wikipedia/Code/UIViewController+WMFOpenExternalUrl.m
```diff
@@ -19,10 +19,10 @@ - (void)wmf_openExternalUrl:(NSURL *)url useSafari:(BOOL)useSafari {
     NSParameterAssert(url);
 
     //If zero rated, don't open any external (non-zero rated!) links until user consents!
-    if ([SessionSingleton sharedInstance].zeroConfigState.disposition && [[NSUserDefaults wmf_userDefaults] boolForKey:@"ZeroWarnWhenLeaving"]) {
-        WMFZeroMessage *zeroMessage = [SessionSingleton sharedInstance].zeroConfigState.zeroMessage;
-        NSString *exitDialogTitle = zeroMessage.exitTitle ?: MWLocalizedString(@"zero-interstitial-title", nil);
-        NSString *messageWithHost = [NSString stringWithFormat:@"%@\n\n%@", zeroMessage.exitWarning ?: MWLocalizedString(@"zero-interstitial-leave-app", nil), url.host];
+    if ([SessionSingleton sharedInstance].zeroConfigurationManager.isZeroRated && [[NSUserDefaults wmf_userDefaults] boolForKey:WMFZeroWarnWhenLeaving]) {
```
coreyfloyd: 2 comments that both point to similar conclusions:

1.  "Tell don't ask" - object oriented design usually takes the form of this statement.  For example, If you are talking to a car object:
"Car are you unlocked? if so, then open the door, if not put the key in"  or, 
"Unlock, open door"

In the first example you are examining the state of an object and making decisions for it. In the second example you are letting the object that owns the state make decisions it knows about because it should have all the info it needs to know to perform the action.

2. This category is called `WMFOpenExternalURL` but there is a lot of Zero logic in here. A developer would probably not think to look in this file for such logic. 


Both of these point to putting this logic in the configuration manager - which is the object responsible for managing the zero state.

Wikipedia/Code/WMFZeroConfigurationManager.m
montehurd: They were equivalent except in the nil case:

     NSString* s1 = nil;
     NSString* s2 = nil;

     [s1 isEqualToString:s2] // NO

     [self isNullableString:s1 equalToNullableString:s2] // YES

I suppose I could do the nil check to catch it else do 'isEqualToString:s2'...

coreyfloyd: gotcha…  brian made something for you in this case:

`WMF_IS_EQUAL`

montehurd: Ah, awesome!

montehurd: Fixed!

Wikipedia/Code/UIViewController+WMFOpenExternalUrl.m
montehurd: Yup, makes sense. Done.

Wikipedia/Code/WMFZeroConfigurationManager.m
montehurd: Fixed the BOOL issue.

montehurd: Ya it's invoked by networking stack - not on main thread here.

Wikipedia/Code/WMFZeroConfigurationManager.m
montehurd: I pretty much just carried over the thread safety comment from before, though it was in a different place. 

Ya I thought I had `@synchronized` where needed, but will double-check that and the re-dispatch to the main queue you mentioned...

Wikipedia/Code/WMFZeroConfigurationManager.m
montehurd: Hmm... I don't follow... if the network isn't Zero rated, I think it makes sense that the zeroConfiguration is nil.

Wikipedia/Code/WMFZeroConfigurationManager.m
montehurd: Correct. If there has been no carrier change as signified by `hasChangeHappenedToCarrier` then we don't need to check for or display anything Zero related.

Wikipedia/Code/UIViewController+WMFOpenExternalUrl.m
montehurd: I thought about refactoring this file but kept changes fairly minimal instead. I'm cool with doing so as part of this PR - sounds like you're in favor?

Wikipedia/Code/WMFZeroConfigurationManager.m
montehurd: I had similar suspicions regarding threading, but wasn't sure why things were structured quite as they were. I can hack at the changes you suggested but will prob wait until I'm back at the office so I can confirm with "WMF-Testing" that things don't get funky.

Wikipedia/Code/UIViewController+WMFOpenExternalUrl.m
coreyfloyd: yep… sounds good!

Wikipedia/Code/WMFAppViewController.m
```diff
@@ -854,34 +854,36 @@ - (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer shouldRecogni
...
-- (void)showFirstTimeZeroOnAlertIfNeeded:(WMFZeroMessage *)zeroMessage {
+- (void)showFirstTimeZeroOnAlertIfNeeded:(WMFZeroConfiguration *)zeroConfiguration {
     if ([self zeroOnDialogShownOnce]) {
         return;
     }
 
     [self setZeroOnDialogShownOnce];
 
-    UIAlertController *dialog = [UIAlertController alertControllerWithTitle:zeroMessage.message message:MWLocalizedString(@"zero-learn-more", nil) preferredStyle:UIAlertControllerStyleAlert];
+    NSString* title = zeroConfiguration.message ? zeroConfiguration.message : MWLocalizedString(@"zero-free-verbiage", nil);
```
coreyfloyd: I would put this inside of the zeroConfiguration object. That way the message always comes from the same place.

Wikipedia/Code/WMFZeroConfigurationManager.m
coreyfloyd: Ahh ok - I think I misunderstood… adding some different comments…

coreyfloyd:
@montehurd i think this looks good aside from the threading stuff that already existed. I may merge this before you get in. but we can do the threading in a follow on.

There may be better ways we can handle the configuration message == nil stuff besides comments… but that is less important.

PR #871, opened 2016-09-05

niedzielski:


Wikipedia/Localizations/qqq.lproj/Localizable.strings
```diff
@@ -266,10 +267,10 @@
 "license-footer-text" = "Brief footer text linking to CC-BY-SA license. '$1' is placeholder for the {{msg-wm|Wikipedia-ios-license-footer-name}} message";
 "license-footer-name" = "License short name; usually leave untranslated as CC-BY-SA 3.0\n{{Identical|CC BY-SA}}";
 "table-of-contents-heading" = "Header text appearing above the first section in the table of contents\n{{Identical|Content}}";
-"table-of-contents-button-label" = "Accessibility label for the Table of Contents button";
+"table-of-contents-button-label" = "Accessibility label for the Table of Contents button\n{{Identical|Table of contents}}";
 "table-of-contents-close-accessibility-label" = "Accessibility label for closing table of contents";
 "table-of-contents-close-accessibility-hint" = "Accessibility hint for closing table of contents\n{{Identical|Close}}";
-"table-of-contents-popover-description" = "Description of Table of Contents which can appear over its icon in a tip bubble";
+"table-of-contents-popover-description" = "Description of Table of Contents which can appear over its icon in a tip bubble. “Overview” refers to a view of the article’s structure, not a summary of the article.";
```
niedzielski: Not sure if you guys want these curly doubly quotes around "overview".

montehurd: Oh good catch! I can add a test for this...

niedzielski: I don't know that it's wrong. It might actually be more correct depending on your angle.

montehurd: @niedzielski so I did some poking around and it looks like some languages are using those quotes - for example, the Romanian strings contains `„reîncarcă”` which uses the right quote `”` on the right side and bottom aligned left quote `„` on the left side. 

So I think we're good with allowing these.

👍

PR #873, opened 2016-09-07
Fix not opening the last article on launch

coreyfloyd:
The code for setting the last article was never being invoked.

- Move setting the open URL to the article VC
- Move setting the open URL to nil in the AppVC
- Remove unused delegate code


joewalsh:
:+1:

PR #879, opened 2016-09-08
Fix feedback link.

montehurd:
https://phabricator.wikimedia.org/T140834

Removed unneeded use of SVModalWebViewController to show external
web pages in modals - not needed now since we've dropped iOS 8.

### DIFF
--- a/Wikipedia/Code/UIViewController+WMFOpenExternalUrl.m
+++ b/Wikipedia/Code/UIViewController+WMFOpenExternalUrl.m
@@ -1,5 +1,4 @@
 #import "UIViewController+WMFOpenExternalUrl.h"
-#import "SVModalWebViewController.h"
 
 #import "Global.h"
 #import "WMFZeroConfigurationManager.h"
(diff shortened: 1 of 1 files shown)


montehurd:
Note: we can remove `pod 'SVWebViewController', '~> 1.0'` from our podfile now. I would have but running `./setup` on my machine is being funky...

coreyfloyd:
two issues:
1. It's unclear what the fix was
What was wrong and how did this fix it? Remember your commit messages should explain **WHY** this fix worked. Even if you want to explain that here now as a comment, this info should also be in your commit messages. They are just as important as the fix itself as they give us a way to know why something was changed if we ever have to refactor or debug in the future (using git blame).

2. Behavior change
We still want the ability to display the `SFSafariViewController` within the app. We should leave the logic and public API of this class intact and just remove references to the SVWebViewController

montehurd:
Oops!

Re: #1 I assumed SVWebViewController had a bug, but I didn't waste time verifying it because we only added it for iOS 8 and the Apple method worked fine first try. 

Re: #2 Was following YAGTNI principal.

montehurd:
Gah! I force pushed without thinking (gotta stop doing that). I should have reverted the last commit then pushed the new commit :/

PR #880, opened 2016-09-09
Fix TOC and Find in Page toasts appearing at top of screen if an article link is quickly tapped before the toasts are shown

montehurd:
https://phabricator.wikimedia.org/T145166


Wikipedia/Code/UIViewController+WMFDynamicHeightPopoverMessage.m
```diff
@@ -38,15 +42,32 @@ - (void)wmf_presentDynamicHeightPopoverViewControllerForSourceRect:(CGRect)sourc
...
+                         if(duration > 0){
+                             [self performSelector:@selector(dismissPopover:) withObject:popoverVC afterDelay:duration];
+                         }
+                     }];
+}
+
+- (void)dismissPopover:(UIViewController*)popoverVC {
+    // Ensure the popover is still the presented view controller.
+    if(self.presentedViewController == popoverVC){
+        [self dismissViewControllerAnimated:YES
```
coreyfloyd: Probably a good time to start using the completion blocks as well - then you can actually wait for the dismissal before showing the next one.

PR #882, opened 2016-09-09
Fix top read not updating on language changes

coreyfloyd:
- Made the siteURL a computed property so its always fresh
- Updated language link controller to not cache values, always compute the language arrays.
-- This is necessary because now we have multiple processes. We have to make sure we check user defaults when we come back to see if anything changed.


montehurd:
Looked like travis failed so I merged `develop` in to restart travis.

While that's churning I ran the tests locally and am seeing a few failing tests in `MWKLanguageLinkControllerTests`.

montehurd:
It may be that the tests are failing because I added a new language `Tulu` to `languages.json` a bit ago...

montehurd:
Hmm I removed `Tulu` from the `languages.json` and the tests still failed so it's not that...

PR #885, opened 2016-09-10
Expand continue reading and top read widget text fields if no image.

montehurd:
-----
The `continue reading` and `top read` cells are a bit different - I tried to keep this diff as minimal as possible. Can DRY up things in the future.

-----

# Top Read and Continue Reading with images
screen shot 2016-09-09 at 11 43 29 pm

# Continue Reading without image with text fields extending to right edge
screen shot 2016-09-09 at 11 44 43 pm

# Top Read without image with text fields extending to right edge (see the last item - I faked out not having an image url for that cell)
screen shot 2016-09-09 at 11 50 58 pm


montehurd:
@coreyfloyd should work

PR #888, opened 2016-09-12
crash fixes and compiler warning reduction

joewalsh:


Wikipedia/Code/WMFExploreViewController.m
```diff
@@ -63,7 +63,8 @@ @interface WMFExploreViewController () <WMFExploreSectionSchemaDelegate,
     UIViewControllerPreviewingDelegate,
     WMFAnalyticsContextProviding,
     WMFAnalyticsViewNameProviding,
-    WMFColumnarCollectionViewLayoutDelegate>
+    WMFColumnarCollectionViewLayoutDelegate,
+    UINavigationControllerDelegate>
```
coreyfloyd: @joewalsh  what navigation controller is this the delegate of? I removed it the other day because I thought it had some extra delegate methods that didn't belong to it - did I mis something else?

joewalsh: The settings VC - line 261 through 269:

```
#pragma mark - Actions

- (void)showSettings {
    UINavigationController *settingsContainer =
        [[UINavigationController alloc] initWithRootViewController:
                                            [WMFSettingsViewController settingsViewControllerWithDataStore:self.dataStore]];
    settingsContainer.delegate = self;
    [self presentViewController:settingsContainer
                       animated:YES
                     completion:nil];
}
```

coreyfloyd: @joewalsh I don't see any delegate methods implemented though - are we missing some behavior?

PR #889, opened 2016-09-12
fixing crash when select/scrolling the toc, again

coreyfloyd:
move cell registration and most table view config inside the lazy loading logic.
now the toc should have cells registered before it is called

also boyscout an extra guard statement for out of bounds index


joewalsh:
👍

PR #895, opened 2016-09-12
Fix for blank toolbar on Explore? Needs testing...

montehurd:
So far I've not been able to reproduce the bug using this PR :)

@joewalsh @coreyfloyd 
This WIP basically just side-steps the layout portion of the `secondToolBar` bits which I thought were causing the trouble... not sure why we aren't just setting `self.toolbarItems` which seems to work fine... can we strip out the `secondToolBar` bits now? Were they needed for compatibility issues or something?

Feel free to push commits or close this PR as needed.

### DIFF
--- a/Wikipedia/Code/WMFArticleNavigationController.m
+++ b/Wikipedia/Code/WMFArticleNavigationController.m
@@ -19,7 +19,7 @@ static const NSTimeInterval WMFArticleNavigationControllerSecondToolbarAnimation
     [super setDelegate:self];
 
     self.secondToolbarHidden = YES;
-    self.secondToolbar = [[UIToolbar alloc] initWithFrame:CGRectZero];
+//    self.secondToolbar = [[UIToolbar alloc] initWithFrame:CGRectZero];
     [self.view addSubview:self.secondToolbar];
     [self layoutSecondToolbarForViewBounds:self.view.bounds hidden:self.isSecondToolbarHidden animated:NO];
 }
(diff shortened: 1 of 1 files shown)


Wikipedia/Code/WMFArticleNavigationController.m
```diff
@@ -19,7 +19,7 @@ - (void)viewDidLoad {
     [super setDelegate:self];
 
     self.secondToolbarHidden = YES;
-    self.secondToolbar = [[UIToolbar alloc] initWithFrame:CGRectZero];
+//    self.secondToolbar = [[UIToolbar alloc] initWithFrame:CGRectZero];
```
montehurd: Looks like we don't need this second toolbar?

If not we can get rid of all it's layout methods too... i.e. `layoutSecondToolbarForViewBounds` and others in this file...

Wikipedia/Code/WMFArticleNavigationController.m
```diff
@@ -75,11 +76,11 @@ - (void)navigationController:(UINavigationController *)navigationController will
     }
     //Ideally this observes secondToolbarItems for changes, but this is all we need for our use case at the moment
     NSArray *newItems = viewController.secondToolbarItems;
-    [self.secondToolbar setItems:newItems animated:animated];
+    self.toolbarItems = newItems;
```
montehurd: Per http://stackoverflow.com/a/17937215

joewalsh:
@montehurd the second toolbar is for the new random article view (only enabled by the tweak) - it appears above the regular toolbar. I think we should keep it around until we decide to keep (or scrap) the new random feature. I fixed the visual glitch and it's merged into develop.

joewalsh:
@montehurd does this fix the UI freezing glitch or just the visual glitch?

PR #898, opened 2016-09-13
Fix Widget Titles, Placeholders, and Label Widths

joewalsh:
- Updates widget titles to match the design mocks
- Fixes xib placeholders showing instead of the empty placeholder
- Fixes labels not expanding to full width when thumbnail URL is missing

### DIFF
--- a/TopReadWidget/WMFTodayTopReadWidgetViewController.swift
+++ b/TopReadWidget/WMFTodayTopReadWidgetViewController.swift
@@ -214,6 +214,8 @@ class WMFTodayTopReadWidgetViewController: UIViewController, NCWidgetProviding {
                 }) {
                     vc.collapseImageAndWidenLabels = false
                 }
+            } else {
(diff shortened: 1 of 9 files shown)


TopReadWidget/WMFTodayTopReadWidgetViewController.swift
```diff
@@ -214,6 +214,8 @@ class WMFTodayTopReadWidgetViewController: UIViewController, NCWidgetProviding {
                 }) {
                     vc.collapseImageAndWidenLabels = false
                 }
+            } else {
+                vc.collapseImageAndWidenLabels = true
```
montehurd: Do we need same thing on Continue reading widget?

joewalsh: good catch. added.

montehurd:
👍

PR #899, opened 2016-09-13
don't allow the interactive pop gesture recognizer to start if there's only 1 view controller on the stack

joewalsh:


montehurd:
@joewalsh hey I'm seeing something weird on the welcome VC with edge drag... (maybe unrelated to this PR?)

welcome mov

montehurd:
@joewalsh batcave? irc cloud is down :(

joewalsh:
What are the repro steps? I’m not seeing it

montehurd:
@joewalsh I'm seeing it about 50% of the time - most often when tapping through to the last welcome panel, then dragging from left edge to go back to the first panel.

montehurd:
+1 (the issue from the animated gif above looks to be unrelated)

PR #904, opened 2016-09-14
Fix Widget Article Tap Mismatch

joewalsh:
Pushing two VCs simultaneously with animation causes the issue. These transitions aren't visible, so should be animated:NO anyway.

https://phabricator.wikimedia.org/T145031


montehurd:
👍

PR #905, opened 2016-09-14
Fix for saved page placeholder message not always showing when it should

montehurd:
https://phabricator.wikimedia.org/T145689

The problem was the call to `updateEmptyAndDeleteState` from `viewWillAppear`
doesn't always result in the placeholder being updated because the
`updateEmptyState` method bails if `self.view.superview == nil`, which it
appears to be in the state outlined by the repro steps outlined in the
phab ticket.


joewalsh:
This same change is part of the fix for missing saved items: https://github.com/wikimedia/wikipedia-ios/pull/910

PR #906, opened 2016-09-15
Add visual layout tests for Top read widget cells.

montehurd:
I tried something a little different with this visual test - I made it use outlines so small color tweaks won't make us have to re-generate the images:

testwithimage_iphone10_0_375x667 2x


joewalsh:
👍

PR #907, opened 2016-09-15
Non-edge swipe to show TOC.

montehurd:
https://phabricator.wikimedia.org/T145720

- Doesn't interfere with edge drag to go back or vertical article scrolling.
- Tuned to trigger if the swipe is fairly fast so as to not interfere with dragging horizontal overflow images from side-to-side.
- RTL ready.

Note: is not progressive drag.


montehurd:
@joewalsh @coreyfloyd hey can u guys try this out on device and let me know what you think?

montehurd:
there's a peek bug with this i think

joewalsh:
Seems to only work for me 50-75% of the time (either it doesn't work or I accidentally trigger the pop gesture). I think it's too similar to the interactive pop gesture, and too difficult to activate reliably.

What if we restored the swipe from right gesture and showed the ToC on the right only when it's activated in that manner and only on the phones?

montehurd:
Closing for now. Can play with this later if need be.

PR #910, opened 2016-09-15
Fix Saved Article List Sync

joewalsh:
For Saved Article View, only observe data source changes from willAppear to didDisappear


montehurd:
👍  if @coreyfloyd's cool with removing superview check

Wikipedia/Code/WMFArticleListTableViewController.m
```diff
@@ -165,10 +165,6 @@ - (void)updateDeleteButton {
 #pragma mark - Empty State
 
 - (void)updateEmptyState {
-    if (self.view.superview == nil) {
```
coreyfloyd: Why are we removing the superview check?

joewalsh:
@coreyfloyd this PR fixes this bug: https://phabricator.wikimedia.org/T145658

PR #913, opened 2016-09-17
Reference popovers!

montehurd:
https://phabricator.wikimedia.org/T116857
https://phabricator.wikimedia.org/T133039
https://phabricator.wikimedia.org/T138041 
https://phabricator.wikimedia.org/T138965 
https://phabricator.wikimedia.org/T141741 

- [x] work on phones and tablets
- [x] respect user's article text size preference
- [x] accommodate larger chunks of reference html
- [x] hook up link taps (to articles, external urls, other sections)
- [x] increase reference link touch targets a bit to make tapping them easier
- [x] remove all old crufty reference bits
- [x] ~~swipe between references in a group~~

refs mov


joewalsh:
simulator screen shot sep 20 2016 10 49 35 am
simulator screen shot sep 20 2016 10 49 01 am
@montehurd I'm noticing the arrow isn't attached on some left-pointing citations on the iPhone. It happens on citation 26 of today's featured article - _An Introduction to Animals and Political Theory_. I couldn't repro it on the iPad, but I also couldn't find any citations that pointed left. Seems that if it's phone only, and phone is changing to a sheet instead of a popover, it's not an issue.

Wikipedia/Code/UIColor+WMFHexColor.h
```diff
@@ -5,6 +5,6 @@
 + (UIColor *)wmf_colorWithHex:(NSInteger)hex
                         alpha:(CGFloat)alpha;
 
-- (NSString *)wmf_hexString;
+- (NSString *)wmf_hexStringWithAlpha:(BOOL)useAlpha;
```
coreyfloyd: Two methods would be better than one here. Instead of requiring that alpha parameter to be passed on every use.

Also, the naming could be a little better, I wasn't totally clear what would happen until I looked at the implementation.

`wmf_hexStringIncludingAlpha:` seems more clear.

Also, since this is returning a string - it would be nice to have a comment in the header file showing the format of the output.

Wikipedia/Code/WMFBarButtonItemPopoverBackgroundView.m
```diff
@@ -1,121 +1,15 @@
...
-@synthesize arrowDirection = _arrowDirection;
-@synthesize arrowOffset = _arrowOffset;
-
 - (instancetype)initWithFrame:(CGRect)frame {
     self = [super initWithFrame:frame];
     if (self) {
-        self.arrowDirection = UIPopoverArrowDirectionAny;
-        self.arrowImageView = [[UIImageView alloc] initWithImage:[self arrowImageWithSize:CGSizeMake([[self class] arrowBase], [[self class] arrowHeight])]];
-        [self addSubview:self.arrowImageView];
+        self.arrowColor = [UIColor wmf_barButtonItemPopoverMessageBackgroundColor];
```
coreyfloyd: Instead of an entire class to set one property, it oust be better if this was a parameter. If you want you can set it on initialization, but I think providing a sensible default and then allowing users to set it is fine.

coreyfloyd: Sorry my last comment may be a little unclear, just a TL;DR
I think the superclass is good. - I think this subclass is unneeded since it is just one line.

Wikipedia/Code/UIColor+WMFHexColor.h
montehurd: sounds good

Wikipedia/Code/WMFBarButtonItemPopoverBackgroundView.m
montehurd: The subclass is needed because `UIPopoverPresentationController` lets you set a custom background view, but by class (via `popoverBackgroundViewClass`) not by setting an instance. Doesn't seem to let you access the instance it creates from your class either iirc.

montehurd: In other words, that one line is the reason the subclass exists.

Lemme know if you can think of a workaround.

Wikipedia/Code/WMFPopoverBackgroundView.m
```diff
@@ -0,0 +1,132 @@
...
+        angle = M_PI_2;
+        center = CGPointMake(frame.size.width + height * 0.5, frame.size.height * 0.5 + self.arrowOffset);
+    }
+
+    self.arrowImageView.center = center;
+    self.arrowImageView.transform = CGAffineTransformMakeRotation(angle);
+}
+
+- (UIImage *)arrowImageWithSize:(CGSize)size {
+    UIGraphicsBeginImageContextWithOptions(size, NO, 0);
```
joewalsh: Might have been easier with `CAShapeLayer` for future reference.

Wikipedia/Code/WMFReferencePopoverMessageViewController.m
```diff
@@ -0,0 +1,122 @@
...
+      "</head>"
+      "<body>"
+      "%@"
+      "</body>"
+      "</html>", baseUrl, (long)fontSize.integerValue, [[UIColor wmf_referencePopoverTextColor] wmf_hexStringIncludingAlpha:NO], self.HTML];
+}
+
+- (NSAttributedString*)attributedStringForHTML:(NSString*)html {
+    return
+    [[NSAttributedString alloc] initWithData:[html dataUsingEncoding:NSUnicodeStringEncoding]
```
joewalsh: there used to be performance issues with this - has everything been tested on device?

Wikipedia/Code/WMFPopoverBackgroundView.m
joewalsh: (to clarify i mean the the whole background, including the arrow as a CAShapeLayer, eliminating the arrow image)

Wikipedia/Code/WMFReferencePopoverMessageViewController.m
montehurd: Ya I remembered those issues, but I seems fine for our tiny HTML snippets.

Wikipedia/Code/WMFPopoverBackgroundView.m
montehurd: UIPopoverPresentationController lets you set a custom background view, but by class (via popoverBackgroundViewClass) not by setting an instance. It takes that view and rounds the corners and adds shadowing and other bits.

joewalsh: Yeah - this class would have a `CAShapeLayer` as a sublayer to draw the background.

montehurd: Ya. Went with the UIPopoverPresentationController built-in way for now.

montehurd: UIPopoverPresentationController manipulates the view - i.e. I can't control the rounding etc, so we don't control the final shape, if that makes sense.

Wikipedia/Code/WMFBarButtonItemPopoverBackgroundView.m
coreyfloyd: @montehurd understood that one class is needed, but you have 2:
`WMFPopoverBackgroundView` and `WMFBarButtonItemPopoverBackgroundView`

I am suggesting that we only need one class. since `WMFBarButtonItemPopoverBackgroundView` only contains a single line

coreyfloyd: nm… I see there are 2 subclass… doh!

montehurd: no worries!

PR #914, opened 2016-09-19

niedzielski:


Wikipedia/Localizations/he.lproj/Localizable.strings
```diff
@@ -415,4 +408,8 @@
 "find-in-page-popover-description" = "חיפוש טקסט בערכים";
 "continue-reading-empty-title" = "אין ערכים שנקראו לאחרונה";
 "continue-reading-empty-description" = "לגלות עוד ערכים לקרוא בוויקיפדיה";
+"top-read-see-more" = "צפייה בערכים פופולריים נוספים";
 "top-read-see-more-trending" = "לראות עוד נושאים חמים";
+"top-read-reader-count-thousands" = "$1k";
+"top-read-header-with-language" = "ויקיפדיה ה{{GRAMMAR:תחילית|$1}}";
```
joewalsh: Generally in this case do we remove the entire localization or reduce everything in the brackets to $1?

joewalsh: To get a build out, I removed that whole string for now.

niedzielski: @joewalsh, looking at the other translations, replacing the bracketed template with `$1` is what I would do.

PR #915, opened 2016-09-19
debounce yapdb cleanup operations

joewalsh:


Wikipedia/Code/MWKDataStore.m
```diff
@@ -195,18 +197,25 @@ - (void)syncDataStoreToDatabase {
...
-        NSMutableArray *keysToRemove = [NSMutableArray array];
-        [view enumerateKeysInGroup:[[view allGroups] firstObject]
-                        usingBlock:^(NSString *_Nonnull collection, NSString *_Nonnull key, NSUInteger index, BOOL *_Nonnull stop) {
-                            [keysToRemove addObject:key];
-                        }];
-        [transaction removeObjectsForKeys:keysToRemove inCollection:[MWKHistoryEntry databaseCollectionName]];
-    }];
+    id previousCleanup = self.previousCleanup;
+    if (previousCleanup != nil) {
+        [NSObject bk_cancelBlock:previousCleanup];
```
coreyfloyd: This will only cancel if called before the delay. Is this the intention?

Basically, I am asking is worthwhile to us NSOperation and look at canceling within the write block. I'm guessing not, but since I see that we are bothering with cancelling at all, I figured I would ask.

joewalsh: @coreyfloyd this is the intention. the idea is to debounce the cleanup so that it doesn't happen on every database change. this way, it should keep the cleanup from happening until the database is at rest for a second.

PR #918, opened 2016-09-20
Add top gradient to gallery nav bar.

montehurd:
https://phabricator.wikimedia.org/T139507

# Before:
screen_shot_2016-08-10_at_11 46 16_am

# After:
screen shot 2016-09-21 at 4 00 32 pm


Wikipedia/Code/UINavigationBar+WMFImageGalleryGradient.swift
```diff
@@ -0,0 +1,38 @@
...
+    func wmf_applyGalleryTopGradientBackground() {
+        if (!CGRectIsNull(bounds)){
+            if (CGRectGetWidth(bounds) > 0 && CGRectGetHeight(bounds) > 0){
+                setBackgroundImage(wmf_galleryTopGradientBackgroundImage(), forBarMetrics: .Default)
+            }
+        }
+    }
+    
+    func wmf_galleryTopGradientBackgroundImage() -> UIImage {
+        UIGraphicsBeginImageContextWithOptions(bounds.size, false, UIScreen.mainScreen().scale)
```
joewalsh: you could just draw this as 1px wide and stretch the image. if performance isn't an issue, probably not worth the change.

Wikipedia/Code/UINavigationBar+WMFImageGalleryGradient.swift
```diff
@@ -0,0 +1,38 @@
...
+    
+    func wmf_applyGalleryTopGradientBackground() {
+        if (!CGRectIsNull(bounds)){
+            if (CGRectGetWidth(bounds) > 0 && CGRectGetHeight(bounds) > 0){
+                setBackgroundImage(wmf_galleryTopGradientBackgroundImage(), forBarMetrics: .Default)
+            }
+        }
+    }
+    
+    func wmf_galleryTopGradientBackgroundImage() -> UIImage {
```
coreyfloyd: Looking here it seems you do not have to render to an image. This library uses CAGradientLayer to achieve the same effect:

https://github.com/chroman/CRGradientNavigationBar/blob/master/CRGradientNavigationBar/CRGradientNavigationBar.m

coreyfloyd:
The height of the bar can be adjusted by adding a constraint:

```
   [self.overlayView.navigationBar mas_remakeConstraints:^(MASConstraintMaker *make) {
        make.height.equalTo(@100.0); //or whatever
    }];
```

Wikipedia/Code/UINavigationBar+WMFImageGalleryGradient.swift
montehurd: Nav bar tiles the background you provide. With a 1 pixel wide image wouldn't its pixels perforce be copied into the resulting pixel buffer *width* times?

montehurd:
> The height of the bar can be adjusted by adding a constraint:

Presuming NYTPhotosOverlayView is using auto layout and not manually fiddling with frames... let me check...

joewalsh:
In the past, I’ve observed that tiling or stretching the image was more performant than drawing the whole gradient. This could have changed, and would require re-profiling. Either way, if there’s not a noticeable lag, it’s fine to keep as is.

PR #919, opened 2016-09-21
point to the correct acknowledgements file

joewalsh:


joewalsh:
https://phabricator.wikimedia.org/T145763

Wikipedia/Code/AboutViewController.m
```diff
@@ -259,7 +259,7 @@ - (void)webView:(WKWebView *)webView decidePolicyForNavigationAction:(WKNavigati
     NSURL *requestURL = navigationAction.request.URL;
 
     if ([[self class] isLicenseURL:requestURL]) {
-        VTAcknowledgementsViewController *vc = [VTAcknowledgementsViewController acknowledgementsViewController];
```
montehurd: Hmm why did that stop working?

joewalsh: Stopped working after CocoaPods config changed

montehurd:
I see. Do you know what config? Wondering if we need to check for other places this may have broken things...

PR #920, opened 2016-09-21
add ability to asynchronously save articles

joewalsh:


Wikipedia/Code/WMFArticleFetcher.m
```diff
@@ -41,6 +41,67 @@ @interface WMFArticleBaseFetcher ()
...
+    NSMutableDictionary *operations = [self articleSaveOperations];
+    @synchronized (queue) {
+        NSString *key = article.url.wmf_databaseKey;
+        if (!key) {
+            return;
+        }
+        
+        NSOperation *op = operations[key];
+        if (op) {
+            [op cancel];
```
coreyfloyd: Any reason to cancel the first op instead of not starting the second one?

Wikipedia/Code/WMFArticleFetcher.m
```diff
@@ -41,6 +41,67 @@ @interface WMFArticleBaseFetcher ()
...
+            [operations removeObjectForKey:key];
+        }
+        
+        op = [NSBlockOperation blockOperationWithBlock:^{
+            [article save];
+            [operations removeObjectForKey:key];
```
coreyfloyd: Do we need to bounce back to the serial queue here?

Wikipedia/Code/WMFArticleFetcher.m
```diff
@@ -41,6 +41,67 @@ @interface WMFArticleBaseFetcher ()
 
 @implementation WMFArticleBaseFetcher
 
++ (NSOperationQueue *)articleSaveQueue {
+    static dispatch_once_t onceToken;
+    static NSOperationQueue *articleSaveQueue;
+    dispatch_once(&onceToken, ^{
+        articleSaveQueue = [NSOperationQueue new];
```
coreyfloyd: should we set a low QoS here?

Wikipedia/Code/WMFArticleFetcher.m
joewalsh: my thought was if the later save had an updated version of the article, we should have that one go through

Wikipedia/Code/WMFArticleFetcher.m
joewalsh: yeah, good catch

Wikipedia/Code/WMFArticleFetcher.m
joewalsh: yup

joewalsh:
@coreyfloyd updated the PR

montehurd:
I was foggy on how `@synchronized` used the parameter passed to it. 
Found a good article here:
http://rykap.com/objective-c/2015/05/09/synchronized/

This jumped out at me as something to keep in mind...
> This brings up a tricky debugging scenario: if you’re passing nil to @synchronized, then you’re not taking any locks and your code may not be thread-safe! If you’re wondering why you’re getting unexpected races, be sure that you’re not ever passing nil to your @synchronized blocks.

montehurd:
Would it be too tricky to pull these async save bits out into a separate object the fetcher could reference?

PR #923, opened 2016-09-23
Avoid caching zeroconfig responses.

dr0ptp4kt:
Bug T139615


joewalsh:
Looks good. Only mild concern is that it would also match `some_action=zerconfig_nope` but the chances of another param getting that naming convention is low.

👍

PR #927, opened 2016-09-24
Fix Welcome "Explore" string wrapping for longer translations

montehurd:
https://phabricator.wikimedia.org/T138153

# Before
screen shot 2016-09-23 at 5 42 17 pm

# After
screen shot 2016-09-23 at 5 39 24 pm

### DIFF
--- a/Wikipedia/Code/WelcomeAnimationView.swift
+++ b/Wikipedia/Code/WelcomeAnimationView.swift
@@ -38,4 +38,10 @@ public class WelcomeAnimationView : UIView {
     var wmf_scaleZeroAndLowerRightTransform: CATransform3D {
           return CATransform3DConcat(wmf_scaleZeroAndRightTransform, wmf_lowerTransform)
     }
+    
+    override public func didMoveToSuperview() {
+        super.didMoveToSuperview()
+        // Fix for: http://stackoverflow.com/a/39614714
+        self.layoutIfNeeded()
+    }
 }
--- a/Wikipedia/Code/WMFWelcome.storyboard
+++ b/Wikipedia/Code/WMFWelcome.storyboard
@@ -81,7 +81,7 @@
                                                     <constraint firstAttribute="width" constant="143" id="R9n-VI-TEI"/>
                                                 </constraints>
                                             </view>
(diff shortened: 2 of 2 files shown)


Wikipedia/Code/WelcomeAnimationView.swift
```diff
@@ -38,4 +38,10 @@ public class WelcomeAnimationView : UIView {
     var wmf_scaleZeroAndLowerRightTransform: CATransform3D {
           return CATransform3DConcat(wmf_scaleZeroAndRightTransform, wmf_lowerTransform)
     }
+    
+    override public func didMoveToSuperview() {
+        super.didMoveToSuperview()
+        // Fix for: http://stackoverflow.com/a/39614714
+        self.layoutIfNeeded()
```
joewalsh: might this need to be in `layoutSubviews` or `setFrame:` or are the values only incorrect when first added to a view?

montehurd: Copy paste from the linked SO post (**emphasis mine**):

> "Sending layoutIfNeeded to a view is not expected to move the view, but in earlier releases, if the view had translatesAutoresizingMaskIntoConstraints set to NO, and if it was being positioned by constraints, layoutIfNeeded would move the view to match the layout engine **before sending layout to the subtree**. These changes correct this behavior, and the receiver’s position and usually its size won’t be affected by layoutIfNeeded.

> Some existing code may be relying on this incorrect behavior that is now corrected. There is no behavior change for binaries linked before iOS 10, but when building on iOS 10 you may need to correct some situations by sending -layoutIfNeeded to a superview of the translatesAutoresizingMaskIntoConstraints view that was the previous receiver, or else positioning and sizing it before (or after, depending on your desired behavior) layoutIfNeeded.

>  Third party apps with custom UIView subclasses using Auto Layout that override layoutSubviews and dirty layout on self before calling super are at risk of triggering a layout feedback loop when they rebuild on iOS 10. When they are correctly sent subsequent layoutSubviews calls they must be sure to stop dirtying layout on self at some point (note that this call was skipped in release prior to iOS 10)."

The constraints seem to be working fine without adding the call to `layoutSubviews` or `setFrame` - it's just the initial layout, I think, based on the way I was adding the animation views, which wasn't set to (as the comment above says) `match the layout engine` in time.

PR #931, opened 2016-09-27
Fixes T139980 - Voiceover Home button - extra "button"

anirudh24seven:
Remove the word "Button" from the Home button accessibility label.


coreyfloyd:
@anirudh24seven thanks!

PR #934, opened 2016-09-27
take domain into account when checking for wiki resources

joewalsh:
https://phabricator.wikimedia.org/T131078


montehurd:
Article links stopped working on today's featured article. iPhone 6s 9.3

joewalsh:
@montehurd added fix

Wikipedia/Code/NSURL+WMFLinkParsing.m
```diff
@@ -180,7 +180,13 @@ + (NSURL *)wmf_desktopURLForURL:(NSURL *)url {
 #pragma mark - Properties
 
 - (BOOL)wmf_isWikiResource {
-    return [self.path wmf_isWikiResource];
```
montehurd: That last commit fixed the article link taps. 

Was wondering if it's ok that this logic now diverges from the NSString `wmf_isWikiResource` that it had formerly called? Should that method be updated or should this logic move there?

joewalsh: wmf_isWikiResource on a string should only be used for paths. If anything, we can remove that public interface and only allow checks on a URL

coreyfloyd: Sounds good to me… prefer to get rid of duplicate APIs

Wikipedia/Code/NSURL+WMFLinkParsing.m
```diff
@@ -180,7 +180,13 @@ + (NSURL *)wmf_desktopURLForURL:(NSURL *)url {
...
+    static NSString *wikiResourceSuffix = nil;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        wikiResourceSuffix = [NSString stringWithFormat:@".%@", WMFDefaultSiteDomain];
+    });
+    NSString *lowercaseHost = self.host.lowercaseString;
+    return (!lowercaseHost || [lowercaseHost isEqualToString:WMFDefaultSiteDomain] || [lowercaseHost hasSuffix:wikiResourceSuffix]) && [self.path wmf_isWikiResource];
```
coreyfloyd: `[lowercaseHost isEqualToString:WMFDefaultSiteDomain]` - that doesn't mean a wiki resource I don't think

coreyfloyd: Or wait… never mind…

If we check for  "wikipedia.org", we also need to check for other "known" domains. For instance this breaks for "mediawiki.org"

coreyfloyd: Just for reference… our FAQ is on mediawiki.org and follows the same logic. There was a poorly named `wmf_isNonStandardURL` property to check URLs like this that don't have language domains and special case some things.

PR #936, opened 2016-09-29
iOS 9 Support for Widgets

joewalsh:
- Enable iOS 9 support for widgets
- Update build target to 9.3 per Apple’s recommendation
- Enable Bitcode
- Add activity indicator to Top Read widget to fix inconsistent layout while loading

PR #937, opened 2016-10-01
Fix for iPad using phone layout.

montehurd:
Looks like we accidentally disabled iPad layout...

screen shot 2016-09-30 at 6 21 47 pm

Happened here (in case that PR needs further attention after re-enabling iPad bits):
https://github.com/wikimedia/wikipedia-ios/pull/936/commits/450e8f17024154d418baa4aed848a0ed64c60bc4


montehurd:
Looks like this was fixed a few hours ago.

PR #938, opened 2016-10-01
Fix for "Adding a new primary language on iOS 9 doesn't set that language as the search language"

montehurd:
https://phabricator.wikimedia.org/T146403

The reported issue was really hard to repro consistently, but I noticed some of the search lang bar layout/configuration bits were being called multiple times (iirc 3 times on first display of the search interface). 

I ended up pulling *all* of the search lang bar button logic out of the `WMFSearchViewController` and into its own controller - `WMFSearchLanguagesBarViewController`. I also switched it to use normal view containment to get the new controller's view into `WMFSearchViewController`'s view.

* haven't seen the iOS 9 issue again
* no more repeated calls
* methods in the new controller are explicit about what they do
* explicit delegate protocol for being notified when search lang changes


Wikipedia/Code/SearchLanguagesBarViewController.swift
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    private var _currentlySelectedSearchLanguage: MWKLanguageLink?
+    private(set) var currentlySelectedSearchLanguage: MWKLanguageLink? {
+        get {
+            return _currentlySelectedSearchLanguage
+        }
+        set {
+            _currentlySelectedSearchLanguage = newValue
+            NSUserDefaults.wmf_userDefaults().wmf_setCurrentSearchLanguageDomain(newValue?.siteURL())
```
coreyfloyd: We should handle the optional here. I would recommend setting the current search language tot he primary language if this value is nil

Wikipedia/Code/SearchLanguagesBarViewController.swift
```diff
@@ -0,0 +1,145 @@
...
+            delegate?.searchLanguagesBarViewController(self, didChangeCurrentlySelectedSearchLanguage: newValue!)
+            updateLanguageBarLanguageButtons()
+        }
+    }
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        _ = languageButtons.map{$0.tintColor = UIColor.wmf_blueTintColor()}
```
coreyfloyd: shouldn't this just be enumerate?

Wikipedia/Code/SearchLanguagesBarViewController.swift
```diff
@@ -0,0 +1,145 @@
...
+                heightConstraint!.constant = 0
+                view.hidden = true
+            }else{
+                heightConstraint!.constant = 44
+                view.hidden = false
+            }
+        }
+    }
+
+    private var _currentlySelectedSearchLanguage: MWKLanguageLink?
```
coreyfloyd: confused at the difference between `_currentlySelectedSearchLanguage` and `selectedLanguage` - this logic should probably be rolled into `selectedLanguage`

Wikipedia/Code/SearchLanguagesBarViewController.swift
```diff
@@ -0,0 +1,145 @@
...
+    private(set) var currentlySelectedSearchLanguage: MWKLanguageLink? {
+        get {
+            return _currentlySelectedSearchLanguage
```
coreyfloyd: Introducing extra state - `currentlySelectedSearchLanguage` should never be different than `NSUserDefaults.wmf_userDefaults().wmf_currentSearchLanguageDomain`. This should just return/set that value.

Wikipedia/Code/SearchLanguagesBarViewController.swift
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    override func viewDidAppear(animated: Bool) {
+        super.viewDidAppear(animated)
+        selectFirstLanguageIfNoneSelectedOrIfFirstLanguageHasChanged()
+        previousFirstLanguage = languageBarLanguages().first
+    }
+    
+    private func languageBarLanguages() -> [MWKLanguageLink] {
+        return MWKLanguageLinkController.sharedInstance().preferredLanguages.enumerate().flatMap{ $0.index < 3 ? $0.element : nil }
```
coreyfloyd: `preferredLanguages.prefix(3)` will grab the first 3 elements without flatmapping and is bounds safe

Wikipedia/Code/SearchLanguagesBarViewController.swift
```diff
@@ -0,0 +1,145 @@
...
+    
+    override func viewWillAppear(animated: Bool) {
+        super.viewWillAppear(animated)
+        updateLanguageBarLanguageButtons()
+        hidden = !NSUserDefaults.wmf_userDefaults().wmf_showSearchLanguageBar()
+    }
+
+    override func viewDidAppear(animated: Bool) {
+        super.viewDidAppear(animated)
+        selectFirstLanguageIfNoneSelectedOrIfFirstLanguageHasChanged()
```
coreyfloyd: No need to have a special method to do checking,  just call your logic to set it to the right language:
`updateLanguageBarLanguageButtons()`

Wikipedia/Code/SearchLanguagesBarViewController.swift
```diff
@@ -0,0 +1,145 @@
...
+        previousFirstLanguage = languageBarLanguages().first
+    }
+    
+    private func languageBarLanguages() -> [MWKLanguageLink] {
+        return MWKLanguageLinkController.sharedInstance().preferredLanguages.enumerate().flatMap{ $0.index < 3 ? $0.element : nil }
+    }
+
+    private func selectFirstLanguageIfNoneSelectedOrIfFirstLanguageHasChanged(){
```
coreyfloyd: As noted above, you should be able to remove this method all together

Wikipedia/Code/SearchLanguagesBarViewController.swift
```diff
@@ -0,0 +1,145 @@
...
+        if(isEveryButtonUnselected() || isFirstLanguageDifferentFromLastTime()){
+            setLanguage(withSender:languageButtons.first!)
+        }
+    }
+    
+    private func isEveryButtonUnselected() -> Bool{
```
coreyfloyd: Should be able to remove this one too…

Wikipedia/Code/SearchLanguagesBarViewController.swift
```diff
@@ -0,0 +1,145 @@
...
+        for button in languageButtons {
+            if button.selected {
+                return false
+            }
+        }
+        return true
+    }
+
+    private func isFirstLanguageDifferentFromLastTime() -> Bool{
```
coreyfloyd: …And this one

Wikipedia/Code/SearchLanguagesBarViewController.swift
```diff
@@ -0,0 +1,145 @@
+import Foundation
+
+@objc protocol SearchLanguagesBarViewControllerDelegate: class {
+    func searchLanguagesBarViewController(controller: SearchLanguagesBarViewController, didChangeCurrentlySelectedSearchLanguage language: MWKLanguageLink)
+}
+
+class SearchLanguagesBarViewController: UIViewController, WMFPreferredLanguagesViewControllerDelegate {
```
coreyfloyd: We should discus removing `WMF` from classes in defined in Swift - so far we have been prefixing our Swift classes.

I could go either way, but would be good to set a best practice here.

@joewalsh any thoughts?

coreyfloyd:
@montehurd most of the refactoring looks good. Left a few comments.

In regards to the bug itself. I'm not sure we can close this issue unless we can do one of the following:
1. Find a logic error in the previous code where there was an opportunity to be in a bad state and show how the new logic handles it correctly.
Or
2. Determine some repro steps to cause the bug and show it now works with the new code.

I saw the notes about the repeated calls updating the search bar. While its nice to not call view update logic unless an update is actually needed, it's not uncommon for views to be updated multiple times before they are are displayed on screen and doesn't necessarily point to a bug.

Wikipedia/Code/SearchLanguagesBarViewController.swift
montehurd: Oh cool! Forgot it was bounds safe...

Wikipedia/Code/SearchLanguagesBarViewController.swift
montehurd: Sure, if you feel strongly about it.

Wikipedia/Code/SearchLanguagesBarViewController.swift
montehurd: Part of what was confusing about the old logic was it was hard to follow why certain things were being set. Made this explicit so it was super clear why it was happening.

Wikipedia/Code/SearchLanguagesBarViewController.swift
montehurd: Cool! I had thought we had said we didn't have to prefix the Swift classes, but I kinda prefer prefixing them personally.

joewalsh: @coreyfloyd seems like what we should be doing is dropping the prefix and using @objc(WMFClassNameHere).

`@objc (WMFSearchLanguagesBarViewController) class SearchLanguagesBarViewController {`

Anything exposed to or used by Objective-C should have a prefix.

Wikipedia/Code/SearchLanguagesBarViewController.swift
coreyfloyd: In this case, yes… one of the things about swift is that it is functional… when you want to transform an object not just iterate, it provides you a map function to make your intentions clear.

This case is just the opposite, you really just want to enumerate, and it would be best to use the method that clearly identifies that intent of just iterating.

Wikipedia/Code/SearchLanguagesBarViewController.swift
coreyfloyd: I can see the value in removing complexity , but looking at the fallout in my other comments, it seems the complexity has just moved. There are now 3 methods and an ivar where previously there was just one method. 

I'm not sure the trade off between calling a method multiple times is worth the added complexity of 3 methods and an ivar that we now have to keep in sync with user defaults.

Wikipedia/Code/SearchLanguagesBarViewController.swift
montehurd: Re-added the prefix.

Wikipedia/Code/SearchLanguagesBarViewController.swift
montehurd: Fixed.

Wikipedia/Code/SearchLanguagesBarViewController.swift
montehurd: Fair enough. I fell victim to "new shiney" syndrome...
http://www.mokacoding.com/blog/when-to-use-map-flatmap-for/

montehurd: Fixed.

Wikipedia/Code/SearchLanguagesBarViewController.swift
montehurd: Fixed.

Wikipedia/Code/SearchLanguagesBarViewController.swift
montehurd: Fixed.

Wikipedia/Code/SearchLanguagesBarViewController.swift
montehurd: Fixed.

Wikipedia/Code/WMFSearchLanguagesBarViewController.swift
```diff
@@ -24,13 +24,21 @@ class WMFSearchLanguagesBarViewController: UIViewController, WMFPreferredLanguag
         }
     }
 
-    private var _currentlySelectedSearchLanguage: MWKLanguageLink?
     private(set) var currentlySelectedSearchLanguage: MWKLanguageLink? {
```
coreyfloyd: Still have variable here… see line 33. 
To remove the backing store you should make this into a computed property. See:
http://www.codingexplorer.com/computed-properties-in-swift/

Wikipedia/Code/WMFSearchLanguagesBarViewController.swift
```diff
@@ -71,7 +76,7 @@ class WMFSearchLanguagesBarViewController: UIViewController, WMFPreferredLanguag
 
     private func selectFirstLanguageIfNoneSelectedOrIfFirstLanguageHasChanged(){
```
coreyfloyd: You shouldn't need this method anymore… just call setCurrentlySelectedLanguageToButtonLanguage() directly from viewDidAppear

Wikipedia/Code/WMFSearchLanguagesBarViewController.swift
```diff
@@ -0,0 +1,141 @@
...
+}
+
+class WMFSearchLanguagesBarViewController: UIViewController, WMFPreferredLanguagesViewControllerDelegate {
+    weak var delegate: WMFSearchLanguagesBarViewControllerDelegate?
+    
+    @IBOutlet private var languageButtons: [UIButton] = []
+    @IBOutlet private var otherLanguagesButton: UIButton?
+    @IBOutlet private var heightConstraint: NSLayoutConstraint?
+    
+    private var previousFirstLanguage: MWKLanguageLink?
```
coreyfloyd: removing that method also removes the need for this variable

Wikipedia/Code/WMFSearchLanguagesBarViewController.swift
```diff
@@ -0,0 +1,141 @@
...
+    private func isEveryButtonUnselected() -> Bool{
+        for button in languageButtons {
+            if button.selected {
+                return false
+            }
+        }
+        return true
+    }
+
+    private func isFirstLanguageDifferentFromLastTime() -> Bool{
```
coreyfloyd: It also removes the need for this function

coreyfloyd:
@montehurd just left a few more comments… 

I saw you made some changes, but the backing ivar is still present and the extra logic for checking state is still here as well.  Not sure if you are finished yet or now. 

Also tests are failing - you should check out travis.

montehurd:
@coreyfloyd oh sorry I have one more commit I haven't pushed yet - wanted to talk w josh first because if he's cool with only changing the selected lang when the previously selected lang is no longer in the top 3 I can simplify it even further...

montehurd:
(that's why I threw the WIP tag on it last night)

montehurd:
@coreyfloyd ok seems pretty solid now

montehurd:
@coreyfloyd ok seems pretty solid now... tested quite a bit... fingers crossed...

coreyfloyd:
👍

PR #947, opened 2016-10-07
Add common actions to article peek

montehurd:
https://phabricator.wikimedia.org/T116944

**Add "Read now", "Save for later" and "Share" peek actions when peeking:**
- [x] Article links
- [x] Explore cards
- [x] List items (top read, search, saved, history, read more at bottom of article, etc)

**Other:**
- [x] Don't add actions for image or external url peeks
- [x] i18n strings
- [x] If the peeked article is already saved the save for later button should say "Saved for later" and tapping it should un-save it (same text and functionality as explore card save buttons)
- [x] Hide save item if peeking main page
- [x] Test everything on ipad including old article share functionality

### DIFF
--- a/Wikipedia/Code/WMFArticleListTableViewController.m
+++ b/Wikipedia/Code/WMFArticleListTableViewController.m
@@ -114,8 +114,8 @@
     UIViewController *vc = self.delegate ?
         [self.delegate listViewController:self viewControllerForPreviewingArticleURL:url] : [[WMFArticleViewController alloc] initWithArticleURL:url dataStore:self.dataStore];
     
-    if ([[vc class] conformsToProtocol:@protocol(WMFArticlePreviewingActionsProviding)]){
-        ((id< WMFArticlePreviewingActionsProviding >)vc).articlePreviewingActionsDelegate = self;
+    if ([[vc class] conformsToProtocol:@protocol(WMFArticlePreviewingActionsDelegate)]){
+        ((WMFArticleViewController*)vc).articlePreviewingActionsDelegate = self;
     }
     return vc;
 }
--- a/Wikipedia/Code/WMFArticleListTableViewController.m
+++ b/Wikipedia/Code/WMFArticleListTableViewController.m
@@ -136,12 +136,12 @@
 
 #pragma mark - WMFArticlePreviewingActionsDelegate
 
-- (void)articleReadMorePreviewAction:(UIPreviewAction*)previewAction selectedWithArticleViewController:(WMFArticleViewController *)articleViewController {
-    [self commitViewController:articleViewController];
+- (void)readMoreArticlePreviewActionSelectedWithArticleController:(WMFArticleViewController *)articleController {
+    [self commitViewController:articleController];
 }
 
-- (void)articleSharePreviewAction:(UIPreviewAction*)previewAction selectedWithArticleShareActivityViewController:(UIActivityViewController*)shareActivityViewController {
-    [self presentViewController:shareActivityViewController animated:YES completion:NULL];
+- (void)shareArticlePreviewActionSelectedWithArticleShareActivityController:(UIActivityViewController*)shareActivityController {
+    [self presentViewController:shareActivityController animated:YES completion:NULL];
 }
 
 #pragma mark - Delete Button
--- a/Wikipedia/Code/WMFArticleListTableViewController.m
+++ b/Wikipedia/Code/WMFArticleListTableViewController.m
@@ -140,10 +140,8 @@
     [self commitViewController:articleViewController];
 }
 
-- (void)presentPreviewedArticleShareActivityViewController:(nullable UIActivityViewController*)shareActivityViewController {
-    if (shareActivityViewController){
-        [self presentViewController:shareActivityViewController animated:YES completion:NULL];
-    }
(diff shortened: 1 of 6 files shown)


Wikipedia/Code/WMFArticleListTableViewController.m
```diff
@@ -111,15 +111,21 @@ - (nullable UIViewController *)previewingContext:(id<UIViewControllerPreviewing>
...
     [[PiwikTracker wmf_configuredInstance] wmf_logActionPreviewInContext:self contentType:contentType];
 
-    if (self.delegate) {
-        return [self.delegate listViewController:self viewControllerForPreviewingArticleURL:url];
-    } else {
-        return [[WMFArticleViewController alloc] initWithArticleURL:url dataStore:self.dataStore];
+    UIViewController *vc = self.delegate ?
+        [self.delegate listViewController:self viewControllerForPreviewingArticleURL:url] : [[WMFArticleViewController alloc] initWithArticleURL:url dataStore:self.dataStore];
+    
+    if ([[vc class] conformsToProtocol:@protocol(WMFArticlePreviewingActionsDelegate)]){
```
montehurd: @coreyfloyd Realized the new previewing actions delegate only needed to be set if we're *actually peeking a view controller*. Was much simpler this way too - i.e. we don't have to set it everywhere we were init'ing an article view controller.

Only needed to be set it in 3 places - here for lists, and in explore vc and article vc (for link peeking).

joewalsh:
@montehurd looks good.  Would pushing the article to share onto the stack before sharing make sense? Having the share dialogue over the previously open article could lead to confusion about which article is being shared.

montehurd:
@joewalsh 
> Would pushing the article to share onto the stack before sharing make sense? 

Good catch... lemme try that real quick...

montehurd:
@joewalsh looks like it's gonna be a bit of work - our various "commitViewController" methods don't have completion blocks so i can't easily chain the article and share pushes... ok if we look at that later?

montehurd:
@joewalsh i added the peeked view controller to the share delegate method so we'll have it ready when we add completion blocks to the various push methods... hopefully we can dry those bits up in near future...

Wikipedia/Code/WMFArticleListTableViewController.m
coreyfloyd: Is this the right check? Doesn't self conform to `WMFArticlePreviewingActionsDelegate`?

Do we even need to check a class here?

Wikipedia/Code/WMFExploreViewController.m
```diff
@@ -957,6 +958,11 @@ - (nullable UIViewController *)previewingContext:(id<UIViewControllerPreviewing>
     UIViewController *vc = [sectionController detailViewControllerForItemAtIndexPath:previewIndexPath];
     self.sectionOfPreviewingTitle = sectionController;
     [[PiwikTracker wmf_configuredInstance] wmf_logActionPreviewInContext:self contentType:sectionController];
+    
+    if ([[vc class] conformsToProtocol:@protocol(WMFArticlePreviewingActionsDelegate)]){
```
coreyfloyd: Same comment here… seems like this check is wrong

coreyfloyd:
@montehurd looks good except for possible an incorrect class check… which may be unnecessary anyways

Wikipedia/Code/WMFArticleListTableViewController.m
coreyfloyd: @montehurd right… but if self is the one that needs to conform to WMFArticlePreviewingActionsDelegate why are we checking if the article view controller conforms to WMFArticlePreviewingActionsDelegate?

montehurd: Remember "vc" here may or may not be an instance of WMFArticleViewController.

coreyfloyd: Notice your next line… if you needed VC to be a WMFArticlePreviewingActionsDelegate, then your next line would be:

```
((id< WMFArticlePreviewingActionsDelegate >)vc).articlePreviewingActionsDelegate = self;
```

montehurd: This is the vc that's being peeked.

montehurd: Oh I see! Sorry lemme fix that.

montehurd: Ok, backing up...

> Is this the right check? Doesn't self conform to WMFArticlePreviewingActionsDelegate?

Self conforms to WMFArticlePreviewingActionsDelegate, but the VC being peeked is the one who's going to send messages to self.

montehurd:
@coreyfloyd ah so you were saying no need to check for protocol compliance, just check if it's a WMFArticleViewController?

montehurd:
@coreyfloyd I guess that makes sense. It's not like we're likely to be peeking articles w/o WMFArticleViewController...

montehurd:
@coreyfloyd Makes sense now... the article view controller only implemented the protocol because it handles article link peeking, so checking the peeked VC for compliance was only true because of that.

montehurd:
This was right on...

> @montehurd right… but if self is the one that needs to conform to WMFArticlePreviewingActionsDelegate why are we checking if the article view controller conforms to WMFArticlePreviewingActionsDelegate?

PR #948, opened 2016-10-08
Fix for missing send email button and toolbar from help and feedback page

montehurd:
https://phabricator.wikimedia.org/T146682

## Main fix
- [x] Fix missing toolbar
- [x] Move TOC button to other side (same side as actual TOC)

-----
## Minor "wrong-toolbar" fixes

The help page and the pages it links to are `mediawiki.org` pages so we don't want the **Standard Article Toolbar**...

screen shot 2016-10-07 at 6 55 03 pm

...which isn't wired up to deal with non lang wiki pages - i.e. pages not from `enwiki`, `dewiki`, etc. 

We want all the help pages to use the **Send Feedback Toolbar**...

screen shot 2016-10-07 at 6 54 46 pm

The following scenarios which would cause these media wiki help pages to still be shown with the standard article toolbar were fixed:

- [x] Help page link taps - 
*Send Feedback Toolbar* is now used
- [x] Help page link peeks - *All peeking on the help pages is now disabled*
- [x] Viewed help pages being added to history and later selected from history - *Disabled history adding for help pages*
- [x] Last viewed help page presented as last read article on app restart - *Disabled remembering last page if it's a help page*


Wikipedia/Code/WMFSettingsViewController.m
```diff
@@ -89,6 +89,7 @@ - (void)viewDidAppear:(BOOL)animated {
 
 - (void)viewWillAppear:(BOOL)animated {
     [super viewWillAppear:animated];
+    self.navigationController.toolbarHidden = YES;
```
montehurd: I had formerly set the nav controller's toolbar to hidden in `WMFHelpViewController`'s `viewWillDisappear` but for some reason after tapping a few links and then tapping the "W" at the top to return to the settings screen *sometimes* the toolbar would still be visible when the settings screen appeared...

Wikipedia/Code/WMFHelpViewController.m
```diff
@@ -33,6 +41,16 @@ - (void)viewDidLoad {
     self.navigationItem.rightBarButtonItem = nil;
 }
 
+- (void)viewWillAppear:(BOOL)animated {
+    [super viewWillAppear:animated];
+    self.navigationController.toolbarHidden = NO;
+}
+
+- (void)webViewController:(WebViewController *)controller didTapOnLinkForArticleURL:(NSURL *)url {
+    WMFHelpViewController *articleViewController = [[WMFHelpViewController alloc] initWithArticleURL:url dataStore:self.dataStore];
```
montehurd: Link taps were using the regular `WMFArticleViewController` instead of `WMFHelpViewController`.

montehurd:
@joewalsh @coreyfloyd should be ready for review - got the merge conflict resolved

Wikipedia/Code/UIViewController+WMFArticlePresentation.m
```diff
@@ -42,10 +42,12 @@ - (void)wmf_pushArticleViewController:(WMFArticleViewController *)viewController
     }
     [[PiwikTracker wmf_configuredInstance] wmf_logView:viewController];
 
-    dispatchOnMainQueueAfterDelayInSeconds(0.5, ^{
-        MWKHistoryList *historyList = viewController.dataStore.userDataStore.historyList;
-        [historyList addPageToHistoryWithURL:viewController.articleURL];
-    });
+    if(viewController.isAddingArticleToHistoryListEnabled){
+        dispatchOnMainQueueAfterDelayInSeconds(0.5, ^{
```
montehurd: hmm not sure... perhaps it was needed pre-yap?

coreyfloyd: It was there to prevent the history view controller from updating before pushing the article VC. It can b removed if there if:
1. Open the history tab
2. Tap on an older item
3. The article pushes in without the history tableview re-ordering during the push animation 

cc @joewalsh @montehurd

montehurd: @coreyfloyd ah cool thx i'll check those steps...

coreyfloyd: @montehurd i forgot to say, you need to remove the delay before testing… but i figure you know anyways

montehurd: @coreyfloyd gotcha thx will do

montehurd: @coreyfloyd confirmed that we still need that delay or the history tableview re-orders during the push animation

montehurd: @joewalsh ^

coreyfloyd: @montehurd maybe add a quick note to why it is there… thats my fault I should have done that

montehurd: @coreyfloyd ah ok will do

PR #949, opened 2016-10-08
Adjusted text slider line width to single pixel width.

montehurd:
https://phabricator.wikimedia.org/T146137

-----

# BEFORE

screen shot 2016-10-07 at 9 35 14 pm

# AFTER

screen shot 2016-10-07 at 9 30 15 pm


Wikipedia/Code/WMFFontSliderViewController.xib
```diff
@@ -35,16 +33,20 @@
                         <userDefinedRuntimeAttribute type="number" keyPath="value">
                             <integer key="value" value="2"/>
                         </userDefinedRuntimeAttribute>
+                        <userDefinedRuntimeAttribute type="number" keyPath="trackHeight">
```
montehurd: The `trackHeight` here and the `tickWidth` below are the only things I changed. The rest of the diff is from Xcode updating from an older storyboard version.

PR #950, opened 2016-10-11

niedzielski:


montehurd:
👍 added the new lang (latvian) so it should pass tests. everything else looks good

montehurd:
Test comment

PR #957, opened 2016-10-17
fix missing analytics name

coreyfloyd:
- Make type not nullable
- Make sure event types are always defined\

https://phabricator.wikimedia.org/T146501


joewalsh:
👍

PR #962, opened 2016-10-20
Added visual hide class to styleoverrides.css to fix display of fractions on certain articles

reetssydney:


montehurd:
Oh we're missing the change to **Wikipedia/www/styleoverrides.css** - that's where overrides go, then we run **make grunt** which processes it and outputs **Wikipedia/assets/styleoverrides.css**. 

If we don't have the change in **Wikipedia/www/styleoverrides.css** it will be lost the next time we run **make grunt**.

montehurd:
For the record, my comment above was wrong: there is no *Wikipedia/www/styleoverrides.css* file, rather, all of the files in the *Wikipedia/www/less/* folder are digested by grunt to produce *Wikipedia/assets/styleoverrides.css*: 

`          
{ src: ["less/**/*.less"], dest: distFolder + "styleoverrides.css"}
`

montehurd:
Per conversation with Rita we're going to hold off on merging this until we chat with Android guys tomorrow about adding this fix up steam (in mobile app css extension) so Android will get the fix too.

montehurd:
I chatted with Dmitry and he's gonna add this upstream then I'll sync the iOS app css so we'll get it.

Closing this for now until the upstream change lands. I moved the phab ticket into "Blocked or waiting".

PR #963, opened 2016-10-20
Disabling bitcode for the project to re-enable symbolication of crash…

coreyfloyd:
… reports


Pods/Manifest.lock
```diff
@@ -259,6 +259,6 @@ SPEC CHECKSUMS:
   VTAcknowledgementsViewController: 44da49d57a1dbf414e653688793a130d24af151e
   YapDatabase: b1e43555a34a5298e23a045be96817a5ef0da58f
 
-PODFILE CHECKSUM: 33cf190d29231c54b305de44490116fca02272ad
+PODFILE CHECKSUM: 71ea2f0bdc1b5fa1953ac241f9de4b1f75f83869
 
-COCOAPODS: 1.1.0.rc.2
```
joewalsh: might want to re-up to the latest version of cocoapods

joewalsh: i think i had to put beta.1 in the bundle file to get it to install a non-released version that fixed an issue we were having

coreyfloyd: oh we locked "1.1.0.beta.1" in the gem file…

PR #968, opened 2016-10-21
Use UIPageViewController to present reference groups on phones

montehurd:
https://phabricator.wikimedia.org/T116857

- [x] Basic functionality showing a panel for each reference in a group
- [x] Scroll the Webview so the tapped reference is near top of page when panel(s) appear
- [x] Highlight tapped reference group - i.e. "[x] [y] [z]"
- [x] RTL

dino mov


WMFReference.swift
```diff
@@ -0,0 +1,34 @@
+
+@objc class WMFReference: NSObject {
+    
+    var html:String? = ""
+    var refId:String? = ""
+    var rect:CGRect = CGRectZero
+    var text:String? = ""
+    
+    init(html:String, refId:String, rect:CGRect, text:String) {
```
coreyfloyd: Why are the vars not lets if they are all required?

Wikipedia/Code/WMFReferencePageBackgroundView.swift
```diff
@@ -0,0 +1,21 @@
+
+class WMFReferencePageBackgroundView: UIView {
+    var clearRect:CGRect = CGRectZero {
+        didSet {
+            setNeedsDisplay()
+        }
+    }
+
+    override func drawRect(rect: CGRect) {
```
coreyfloyd: Not a big deal, but any reason you went with draw rect instead of just setting the corner radius on the self.layer?

coreyfloyd: Oh wait… are you cutting out a hole?

coreyfloyd: nm… yep… I see it… disregard

Wikipedia/Code/WMFReferencePageViewController.swift
```diff
@@ -0,0 +1,108 @@
+
+import Foundation
+
+extension UIViewController {
+    class func wmf_viewControllerFromReferencePanelsStoryboard() -> Self {
```
coreyfloyd: Shouldn't this just be a category on this view controller, not all view controllers?

WMFReference.swift
joewalsh: yeah, looks like they could be non optional lets without initial values

montehurd: Hmm I may be missing something - how do I set them if they're lets?

coreyfloyd: @montehurd you can set lets on init

Wikipedia/Code/WMFReferencePageViewController.swift
montehurd: It's used to get different VC's from that storyboard - WMFReferencePanelViewController *or* WMFReferencePageViewController.

WMFReference.swift
montehurd: Ah cool will update.

montehurd: Fixed.

PR #984, opened 2016-10-26
Fix Warnings & Analyzer Errors

joewalsh:


Wikipedia/Code/WMFBaseDataStore.h
```diff
@@ -11,7 +11,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @protocol WMFDatabaseChangeHandler <NSObject>
 
-- (void)processChanges:(NSArray<YapDatabaseViewRowChange *> *)changes onConnection:(YapDatabaseConnection *)connection;
+- (void)processChanges:(NSArray<NSNotification *> *)changes onConnection:(YapDatabaseConnection *)connection;
```
coreyfloyd: It actually is a `YapDatabaseViewRowChange`… the Yap API is lying

joewalsh: Should we just make it NSArray * to match what YapDB expects?

coreyfloyd:
👍

PR #988, opened 2016-10-26

niedzielski:


montehurd:
@joewalsh:

@niedzielski re-ran the TWN sync so those crufty qqq keys would be removed from https://translatewiki.net

joewalsh:
looks good, removed HTML, tests now passing locally

PR #989, opened 2016-10-27
Fix Image Cancellation Crash

joewalsh:
In the news collection view cell would call `wmf_configureWithDefaultPlaceholder` after an image load failed by cancellation - which would deadlock on the cancellation queue


coreyfloyd:
👍 Already passed several times…

PR #995, opened 2016-10-27
Feature/cache chinese/t149034

coreyfloyd:
This fix is:

Check to see if we are chinese
If so, cache the header that we are sending to the API on the article
When deciding to re download, look at the cached header and compare to the current header
If they are different we can assume the user has changed their languages and we will re download the chinese article to be sure


Wikipedia/Code/WMFArticleFetcher.m
```diff
@@ -173,6 +173,10 @@ - (id)serializedArticleWithURL:(NSURL *)url response:(NSDictionary *)response {
     MWKArticle *article = [[MWKArticle alloc] initWithURL:url dataStore:self.dataStore];
     @try {
         [article importMobileViewJSON:response];
+        if([NSLocale wmf_isCurrentLocaleChineseVariant]){
```
joewalsh: Instead of checking the locale, this should check the article language. A user  with an English locale can have a preferred order of Chinese language variants in settings that would affect how zh-wiki articles are loaded.

PR #1002, opened 2016-10-28
Feature/fix remove section

coreyfloyd:
Since the refresh is now manually triggered we need to do that now.

Also added animation to the deletion


coreyfloyd:
👍 looks good  - thanks @joewalsh

PR #1005, opened 2016-10-29
added clear functionality for W for Voice Over Suer

dmcglinchey2779:
When on Explore screen, W button -"Wikipedia button, scroll to top of Explore"

When not on Explore screen, W button- "Wikipedia button, return to [Explore|Saved|History]"


coreyfloyd:
👍

PR #1007, opened 2016-10-29
Group Labels into One Accessibilty Element on Explore Article Preview Cell

joewalsh:


joewalsh:
adding WIP tag as subview order doesn't necessarily equal visual order. looking at other options

Wikipedia/Code/GroupedAccessibilityView.swift
```diff
@@ -0,0 +1,71 @@
+import UIKit
+
+class GroupedAccessibilityView: UIView {
+
+    var arrangedAccessibilityViews: [UIView] = []
+    
+    //convienene outlets
+    @IBOutlet weak var accessibilityView0: UIView?
```
coreyfloyd: Do we we use an outlet collection? or does that matter for this purpose?

coreyfloyd:
👍

Wikipedia/Code/GroupedAccessibilityView.swift
joewalsh: Can't really guarantee the order of an outlet collection - we could tag all the views and sort on that but this seems easier? Open to suggestion

coreyfloyd:
👍 waiting on travis

PR #1010, opened 2016-10-31
Fixes T138091. Prevent article from reloading when back button is hit.

anirudh24seven:
Article should not reload when back button in Edit Article screen is selected.


joewalsh:
Thanks!

PR #1016, opened 2016-11-01
Fixes T132638. Navigate to Article if user clicks link in edit preview.

anirudh24seven:
The strings need to be localized though. I am not yet familiar with submitting requests for localizing strings.


Wikipedia/Code/PreviewAndSaveViewController.m
```diff
@@ -122,13 +122,21 @@ - (BOOL)prefersStatusBarHidden {
 }
 
 - (void)wmf_showAlertForTappedAnchorHref:(NSString *)href {
+    
     UIAlertController *alertController =
-        [UIAlertController alertControllerWithTitle:href
-                                            message:nil
+        [UIAlertController alertControllerWithTitle:@"Your changes have not been saved yet"
```
coreyfloyd: @anirudh24seven would you mind adding these strings to the english locatable.strings file? 
Also, there is a locatable.strings file for "qqq" - in that file we add a description to the string so translators know the context. Just open it up and you will see what I mean. 

After that you can use the `MWLocalizedString` macro to reference those strings his this file.

Let me know if try to do this and run into issues and thanks for the contribution!

Wikipedia/Code/PreviewAndSaveViewController.m
```diff
@@ -197,6 +205,15 @@ - (void)goForward {
     }
 }
 
+- (void)goToArticleFor:(NSString *)href {
```
coreyfloyd: Instead of adding this logic here, would you mind adding an additional method to the `PreviewAndSaveViewControllerDelegate`?

```
- (void)previewViewControllerDidCancel:(PreviewAndSaveViewController *)previewViewController;

```

This way the delegate can handle any navigation. You also won't need the data stores in this class anymore or need to manually construct the URL.

Thanks and let me know if this makes sense

Wikipedia/Code/PreviewAndSaveViewController.m
anirudh24seven: I was able to do this without issues. Will send in a new commit once I'm done with the other comment too.

Wikipedia/Code/PreviewAndSaveViewController.m
anirudh24seven: I thought about this but I don't understand how adding the method to the delegate will prevent the manual construction of the URL. Is it possible for you to do this part? That way, I can look at the code and learn how to do it too.

Will update this Pull Request with a fix for the previous comment.

coreyfloyd:
@anirudh24seven thanks… I'll check out the delegate stuff in the next couple days (we are prepping for a release right now, but I won't let this sit to long)

PR #1037, opened 2016-11-07
Re-enable the html beta test with Joe's mods

coreyfloyd:


joewalsh:
tests passed locally for me

PR #1051, opened 2016-11-09
Showing offline view

coreyfloyd:


joewalsh:
@coreyfloyd is it intended behavior that it only shows when the view is empty? I thought previously the offline view was shown regardless of whether or not the view was empty

PR #1058, opened 2016-11-14

niedzielski:


montehurd:
Fixing a plurals syntax entry...

montehurd:
👍 should be good now...

PR #1063, opened 2016-11-15
Dynamic text

montehurd:
https://phabricator.wikimedia.org/T148703

First pass at converting welcome screens to use Dynamic Type.


montehurd:
New branch here: https://github.com/wikimedia/wikipedia-ios/pull/1087

PR #1064, opened 2016-11-16
Core Data, AppSee Integration

joewalsh:


WMFModel/WMFArticle+CoreDataProperties.h
```diff
@@ -0,0 +1,29 @@
...
+
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface WMFArticle (CoreDataProperties)
+
++ (NSFetchRequest<WMFArticle *> *)fetchRequest;
+
+@property (nonatomic) BOOL isBlocked;
+@property (nullable, nonatomic, copy) NSString *key;
```
coreyfloyd: Is this the url? If so, should we just call it URL? 

If not, can we add the URL?

WMFModel/WMFArticle+CoreDataProperties.h
```diff
@@ -0,0 +1,29 @@
...
+@property (nullable, nonatomic, copy) NSDate *viewedDate;
+@property (nullable, nonatomic, copy) NSDate *viewedCalendarDate;
```
coreyfloyd: Not clear what this property is from the name. Looks like it is the date without time? Is that right? If so, maybe call it something like that viewedDateWithoutTime?

WMFModel/WMFArticle+CoreDataProperties.h
```diff
@@ -0,0 +1,29 @@
...
+@property (nullable, nonatomic, copy) NSString *viewedFragment;
+@property (nonatomic) double viewedScrollPosition;
+@property (nullable, nonatomic, copy) NSDate *newsNotificationDate;
+@property (nullable, nonatomic, copy) NSDate *savedDate;
+@property (nonatomic) BOOL wasSignificantlyViewed;
+@property (nullable, nonatomic, copy) NSString *displayTitle;
+@property (nullable, nonatomic, copy) NSString *wikidataDescription;
+@property (nullable, nonatomic, copy) NSString *snippet;
+@property (nullable, nonatomic, copy) NSString *thumbnailURLString;
+@property (nonatomic) double latitude;
```
coreyfloyd: We could add a category to return the CLLocation for the lat and lon

WMFModel/WMFArticle+CoreDataProperties.h
```diff
@@ -0,0 +1,29 @@
...
+@property (nonatomic) double longitude;
+@property (nullable, nonatomic, retain) NSDictionary *pageViews;
```
coreyfloyd: Can we add generics here?

WMFModel/WMFArticle+Extensions.h
```diff
@@ -0,0 +1,13 @@
+#import "WMFArticle+CoreDataClass.h"
+
+@interface WMFArticle (WMFExtensions)
+
+@property (nonatomic, readonly, nullable) NSURL *URL;
```
coreyfloyd: looks like you already answered my question…

Only issue I see is that its not clear what "key" is. How does the developer know how to key these?

Maybe add a method like `setKeyWithURL:(NSURL*)url`?

WMFModel/WMFArticle+CoreDataProperties.h
coreyfloyd: nm… see it below

WMFModel/WMFArticle+CoreDataClass.h
```diff
@@ -0,0 +1,12 @@
+#import <Foundation/Foundation.h>
+#import <CoreData/CoreData.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface WMFArticle : NSManagedObject
```
coreyfloyd: Any reason why the class file is not "WMFArticle.h"? 

Haven't set up a stack from the ground up in a while. Is this new convention?

WMFModel/WMFArticlePreview.h
```diff
@@ -1,22 +0,0 @@
-#import <Mantle/Mantle.h>
-@import CoreLocation;
-
-@interface WMFArticlePreview : MTLModel
```
coreyfloyd: Probably going to have to undelete all of these to migrate, right? Or do you plan to rebuild them all?

WMFModel/WMFArticle+CoreDataProperties.h
coreyfloyd: nm… see below

WMFModel/WMFArticle+CoreDataProperties.h
```diff
@@ -0,0 +1,29 @@
+#import "WMFArticle+CoreDataClass.h"
+
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface WMFArticle (CoreDataProperties)
+
++ (NSFetchRequest<WMFArticle *> *)fetchRequest;
+
+@property (nonatomic) BOOL isBlocked;
```
coreyfloyd: More personal preference than anything… but any reason to change the name to blocked? Blacklisted seems more clear to me than blocked which has a more generic sound… 

Related: now that the model is being changed, should we separate this into 2 properties? one for not using in the feed and one for not using in notifications? So the intent of each is more clear?

WMFModel/WMFArticle+CoreDataClass.h
joewalsh: Yeah - this is what "Create NSManagedObjectSubclass" does now... not sure if we can revert to the old behavior.

WMFModel/WMFArticlePreview.h
joewalsh: Plan to rebuild them all

WMFModel/WMFArticle+CoreDataProperties.h
joewalsh: yeah

WMFModel/WMFArticle+CoreDataProperties.h
joewalsh: Yeah - makes sense.

WMFModel/WMFArticle+CoreDataProperties.h
joewalsh: This could also be a CLLocation directly via a transformable property. We probably wouldn't be able to reliably query on it though.

coreyfloyd: Just noticed that this is a transformable on the content group. We should probably do both as transformable or both as separate properties. Either is fine by me

WMFModel/WMFArticle+CoreDataProperties.h
coreyfloyd: Related: should we instead may this a proper object? 

I only bring this up because migrations can be a pain, so if we want this to be an object, now would be the time.

WMFModel/WMFArticle+CoreDataClass.h
coreyfloyd: No its fine, just my first time seeing it

Wikipedia/Code/NSUserDefaults+WMFExtensions.swift
```diff
@@ -285,12 +285,12 @@ public extension NSUserDefaults {
...
 
     public func wmf_setDidMigrateToNewFeed(didMigrate: Bool) {
-        self.setBool(didMigrate, forKey: WMFDidMigrateToDatabaseFeedKey)
+        self.setBool(didMigrate, forKey: WMFDidMigrateToCoreDataFeedKey)
         self.synchronize()
     }
     
     public func wmf_didMigrateToNewFeed() -> Bool {
-        return self.boolForKey(WMFDidMigrateToDatabaseFeedKey)
+        return self.boolForKey(WMFDidMigrateToCoreDataFeedKey)
```
coreyfloyd: Just to check, your migration code handles:
1. Users who have migrated the feed to Yap
2. Users who have not migrated the feed to Yap, but have migrated to the shared container
3. Users who have not migrated to the shared container
4. Users who have not migrated history to Yap

Wikipedia/Code/WMFContentGroupDataStore.h
```diff
@@ -1,50 +1,40 @@
...
-#import "WMFDataSource.h"
-
-@class WMFContentGroup;
-@class WMFRelatedPagesContentGroup;
+#import "WMFContentGroup+Extensions.h"
 
 NS_ASSUME_NONNULL_BEGIN
 
-@interface WMFContentGroupDataStore : WMFBaseDataStore
+@interface WMFContentGroupDataStore : NSObject
```
coreyfloyd: Is it worth keeping the content group data store around at this point? I mean these methods could just go on the data store so we only have one object to pass around. For separation we could put content group methods in a category on the data store. ¯\_(ツ)_/¯

WMFModel/WMFArticleDataStore.h
```diff
@@ -0,0 +1,25 @@
+#import "WMFBaseDataStore.h"
+
+@class MWKSearchResult, MWKLocationSearchResult, MWKArticle, WMFFeedArticlePreview, WMFArticle;
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface WMFArticleDataStore : NSObject
```
coreyfloyd: Same goes for this class as the content store… could we just collapse into a single "store"?

Wikipedia/Code/MWKDataStore.m
```diff
@@ -114,25 +122,173 @@ - (void)dealloc {
...
 - (instancetype)init {
-    self = [self initWithDatabase:[YapDatabase sharedInstance]];
+    self = [self initWithContainerURL:[[NSFileManager defaultManager] wmf_containerURL]];
     return self;
 }
 
-- (instancetype)initWithDatabase:(YapDatabase *)database {
-    self = [self initWithDatabase:database legacyDataBasePath:[[MWKDataStore class] mainDataStorePath]];
-    return self;
+static pid_t currentPid() {
```
coreyfloyd: Since this is static, is there an implicit requirement that only a single MWKDataStore be instantiated per process? If so, should we somehow enforce this through a singleton or some other mechanism?

coreyfloyd: Actually looking through, it seems like not. You really only want to process the notification once per process not once per instance

Wikipedia/Code/NSUserDefaults+WMFExtensions.swift
joewalsh: Correct

Wikipedia/Code/WMFContentGroupDataStore.h
joewalsh: Yeah - I think long term that's the solution, but for simplicity on this first pass, keeping them around.

WMFModel/WMFArticleDataStore.h
joewalsh: ^ same comment as above. Long term yes, first pass no.

coreyfloyd:
@joewalsh i commented on a couple of old things above too

WMFModel/WMFArticle+CoreDataProperties.h
joewalsh: Is the extra info on CLLocation used anywhere? Will we ever need to fetch by lat/lon?

Wikipedia/Code/MWKDataStore.m
```diff
@@ -119,20 +129,167 @@ - (instancetype)init {
...
+}
+
+- (void)setupCoreDataStackWithContainerURL:(NSURL *)containerURL {
+    NSURL *modelURL = [[NSBundle bundleWithIdentifier:@"org.wikimedia.WMFModel"] URLForResource:@"Wikipedia" withExtension:@"momd"];
+    NSManagedObjectModel *model = [[NSManagedObjectModel alloc] initWithContentsOfURL:modelURL];
+    NSString *coreDataDBName = @"Wikipedia.sqlite";
+
+    NSURL *coreDataDBURL = [containerURL URLByAppendingPathComponent:coreDataDBName isDirectory:NO];
+    NSPersistentStoreCoordinator *persistentStoreCoordinator = [[NSPersistentStoreCoordinator alloc] initWithManagedObjectModel:model];
+    [persistentStoreCoordinator addPersistentStoreWithType:NSSQLiteStoreType configuration:nil URL:coreDataDBURL options:nil error:nil];
```
joewalsh: Wait until we need to

montehurd:
Now that's a diff! ;) Nice work!

joewalsh:
retest this please

joewalsh:
!jenkins test

montehurd:
When I cleared the history it did a very strange animation where the cells all moved to the upper right corner of the screen.

montehurd:
When I cleared saved pages it did an animation that looked normal.

montehurd:
Tapping the title bar on Explore doesn't scroll it to the top.

montehurd:
👍 Tested various parts of the app for a while... only found the 2 issues noted above.

PR #1066, opened 2016-11-18
Add announcements to feed

coreyfloyd:
- [x] Model
- [x] Fetcher
- [x] Content Source
- [x] Cell
- [x] Clean up expired announcements
- [x] Filter announcements based on IP and platform
- [x] Fix height of caption text view in cell
- [x] Add logic to hide cell when dismissed


WMFModel/WMFContentGroup.m
```diff
@@ -174,6 +234,70 @@ - (WMFContentType)contentType {
...
+    self.wasDismissed = YES;
+}
+
+- (BOOL)updateVisibility{
+    if(self.wasDismissed){
+        if(self.isVisible){
+            self.isVisible = NO;
+            return YES;
+        }
+    }
```
joewalsh: this if block should always return, otherwise the announcement will be made re-visible based on dates even if it was dismissed

PR #1067, opened 2016-11-19
Don’t adjust font size of password field on login

stevenschobert:
Hey folks!

This patch fixes a bug in iOS 10, where setting resize-to-fit text on a secure text field causes the app to hang, peak memory, and eventually crash.

To fix it I just disabled the resizing. To me, since in secure text fields you can't see the actual characters anyways, there doesn't seem to be much harm in turning off the resizing, but let me know if there something else you'd want to do instead!

**bug**: https://phabricator.wikimedia.org/T150982
**radar**: https://openradar.appspot.com/28590124

### Before patch
before-patch

### After patch
after-patch


joewalsh:
Thanks for the fix!

stevenschobert:
Happy to help! 🍻

PR #1071, opened 2016-11-22
Prioritize foreground image downloads

joewalsh:
also up the max concurrent image download count


montehurd:
On clean install with 3G Network Link Conditioner setting...



Time to load "Continent" article when tapping link from "Europe" article:
<pre>
Branch				Seconds
----------------------------------------
develop				~54
bug/image_downloads/T151266	~58
</pre>


Time to load "Continent" article directly from search results:
<pre>
Branch				Seconds
----------------------------------------
develop				~3
bug/image_downloads/T151266	~3
</pre>

I only ran this once...

PR #1093, opened 2016-12-06
Update feed for new primary language

coreyfloyd:
We didn't examine the siteURL of content groups when looking for matches. Also we didn't update the UI properly until the app was quit. This fixes both.

POTD is a special case where the image is the same across languages, but the translation is different. Just need to get guidance before merging.

- [x] Get guidance from @jtminor on POTD
- [x] Implement new POTD behavior


montehurd:
Found one issue - looks like the Main page card isn't updating quite right. 
- On clean install I first looked at the feed with EN as the primary lang. 
- The EN main page card appeared.
- Switched primary lang to DE.
- DE main page card appeared.
- Switched primary lang back to EN.
- DE main page card remained, but _no EN main page card_.

Also, the very first time I tested was not on a clean install, and I saw the following blank main page card (can't remember exact repro steps but iirc I used same steps as above but just not on clean install):
screen shot 2016-12-07 at 11 17 10 am

coreyfloyd:
@montehurd wasn't able to repro the issue… but this PR doesn't touch the main page logic. If you can reproduce again maybe file a ticket for it - and we can try to fix

joewalsh:
@montehurd that looks more like a main page layout issue I've seen before - still trying to track all of those down so definitely file a ticket if you can repro

montehurd:
@coreyfloyd @joewalsh will do

PR #1096, opened 2016-12-07
Ensure xcarchive is imported to Xcode after fastlane build

joewalsh:
```
/**
 *  EXPERIMENTAL: Enable heuristics to detect the app not terminating cleanly
 *
 *  This allows it to get a crash report if the app got killed while being in the foreground
 *  because of one of the following reasons:
 *
 *  - The main thread was blocked for too long
 *  - The app took too long to start up
 *  - The app tried to allocate too much memory. If iOS did send a memory warning before killing the app because of this reason, `didReceiveMemoryWarningInLastSession` returns `YES`.
 *  - Permitted background duration if main thread is running in an endless loop
 *  - App failed to resume in time if main thread is running in an endless loop
 *  - If `enableMachExceptionHandler` is not activated, crashed due to stack overflow will also be reported
 *
 *  The following kills can _NOT_ be detected:
 *
 *  - Terminating the app takes too long
 *  - Permitted background duration too long for all other cases
 *  - App failed to resume in time for all other cases
 *  - possibly more cases
 *
 *  Crash reports triggered by this mechanisms do _NOT_ contain any stack traces since the time of the kill
 *  cannot be intercepted and hence no stack trace of the time of the kill event can't be gathered.
 *
 *  The heuristic is implemented as follows:
 *  If the app never gets a `UIApplicationDidEnterBackgroundNotification` or `UIApplicationWillTerminateNotification`
 *  notification, PLCrashReporter doesn't detect a crash itself, and the app starts up again, it is assumed that
 *  the app got either killed by iOS while being in foreground or a crash occurred that couldn't be detected.
 *
 *  Default: _NO_
 *
 * @warning This is a heuristic and it _MAY_ report false positives! It has been tested with iOS 6.1 and iOS 7.
 * Depending on Apple changing notification events, new iOS version may cause more false positives!
 *
 * @see lastSessionCrashDetails
 * @see didReceiveMemoryWarningInLastSession
 * @see `BITCrashManagerDelegate considerAppNotTerminatedCleanlyReportForCrashManager:`
 * @see [Apple Technical Note TN2151](https://developer.apple.com/library/ios/technotes/tn2151/_index.html)
 * @see [Apple Technical Q&A QA1693](https://developer.apple.com/library/ios/qa/qa1693/_index.html)
 */
@property (nonatomic, assign, getter = isAppNotTerminatingCleanlyDetectionEnabled) BOOL enableAppNotTerminatingCleanlyDetection;
```


joewalsh:
Seems like this is prone to false positives - especially since their doc is talking about testing it on iOS 6 and 7 - this is what reports the 8badf00d crashes

coreyfloyd:
@joewalsh anything wrong with getting false positives? If we disable this then we lose all signal of 8badf00d crashes

joewalsh:
@coreyfloyd updated to remove the change to `enableAppNotTerminatingCleanlyDetection`. It's good to have this data, although potentially problematic. We should keep in mind going forward that these aren't actual 8badf00d crashes - it's based on the app not writing a user default on termination: https://github.com/bitstadium/HockeySDK-iOS/blob/develop/Classes/BITCrashManager.m#L1301

PR #1098, opened 2016-12-07
Fix for unit tests not running locally.

montehurd:
Was seeing "_Test target X encountered an error (Early unexpected exit, operation never finished bootstrapping - no restart will be attempted_" error message when trying to run unit tests locally.

This fixed it: https://github.com/Carthage/Carthage/issues/1002#issue-122848182

### DIFF
--- a/Wikipedia.xcodeproj/project.pbxproj
+++ b/Wikipedia.xcodeproj/project.pbxproj
@@ -7840,7 +7840,7 @@
 					"$(inherited)",
 				);
 				INFOPLIST_FILE = WikipediaUnitTests/Info.plist;
-				LD_RUNPATH_SEARCH_PATHS = "$(inherited) /Applications/Xcode.app/Contents/Developer/Library/Frameworks/ @executable_path/Frameworks @loader_path/Frameworks";
+				LD_RUNPATH_SEARCH_PATHS = "$(inherited) /Applications/Xcode.app/Contents/Developer/Library/Frameworks/ @executable_path/Frameworks $(PROJECT_DIR)/Carthage/Build/iOS @loader_path/Frameworks";
 				MTL_ENABLE_DEBUG_INFO = NO;
 				PRODUCT_BUNDLE_IDENTIFIER = "org.wikimedia.$(PRODUCT_NAME:rfc1034identifier)";
 				PRODUCT_NAME = "$(TARGET_NAME)";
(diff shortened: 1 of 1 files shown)


montehurd:
Hmm... also looks like we're no longer committing the visual test reference images? I had to set _-WMFVisualTestBatchRecordMode_ to **NO** once to generate them. Doesn't not saving these defeats the purpose of the visuals test though? Without tracked persistent reference images we can't see if our changes broke anything because we'll have to re-gen them to run the test, but the test will then just produce the exact image we just re-gen'ed...

joewalsh:
@montehurd we are committing the visual test reference images - you might have been running on a platform that didn't have the images generated. They're there for iOS 10.0 and iOS 9.3 on a 6S

joewalsh:
@montehurd I've been able to run tests locally - were you running tests using the app's scheme or one of the other targets?

montehurd:
@joewalsh Ya I tried iOS 9.3 on a 6S sim but no luck. Same issue with "Wikipedia", "WikipediaRTL" and "Wikipedia Test Language" schemes.

montehurd:
Also tried 6 sim.

montehurd:
@joewalsh But you're right about the reference images - with this PR's config tweak the tests do run for me with iOS 9.3 sim without needing to regenerate them.

joewalsh:
@montehurd do you have carthage installed? if not, could you try installing carthage and then re-running the tests without this tweak

joewalsh:
@montehurd also, if you do have it installed, try upgrading to the latest version

montehurd:
@joewalsh lemme will check... 

Oh also I just pushed another PR with a tiny tweak to one of our visual test macros: https://github.com/wikimedia/wikipedia-ios/pull/1099

montehurd:
@joewalsh Ya looks like it's installed. 
"carthage version" says _0.9.2_

joewalsh:
@montehurd try upgrading to 0.18.1 and re-running the tests on develop. if you installed with homebrew, should just be `brew upgrade carthage`

montehurd:
@joewalsh ya updating brew presently... taking a while

montehurd:
@joewalsh Yup updating Carthage fixed it! Thanks for the help!

joewalsh:
@montehurd no problem. I think this is another strike against Carthage though

PR #1099, opened 2016-12-07
Add slight tolerance for image differences to visual test macro.

montehurd:
A couple visual test were failing for me locally even though their images look identical. For example...

-----
**Reference image:**
-----
reference_testlayoutwithshortextractandimage_9_3_ltr_iphone9_3_375x667 2x

-----
**Test image:**
-----
failed_testlayoutwithshortextractandimage_9_3_ltr_iphone9_3_375x667 2x

-----
Adding a tiny bit of tolerance to the macro used by these tests fixed it.


WikipediaUnitTests/Code/FBSnapshotTestCase+WMFConvenience.h
```diff
@@ -21,7 +21,7 @@
  *  @param view The view to verify.
  */
 #define WMFSnapshotVerifyViewForOSAndWritingDirection(view) \
-    FBSnapshotVerifyView((view), [[UIApplication sharedApplication] wmf_systemVersionAndWritingDirection]);
+    FBSnapshotVerifyViewWithOptions((view), [[UIApplication sharedApplication] wmf_systemVersionAndWritingDirection], FBSnapshotTestCaseDefaultSuffixes(), 0.1);
```
joewalsh: Was this the lowest tolerance that would work? Seems like 10% of pixels off is a lot of pixels to be off

montehurd:
@joewalsh I think any visual test failure is fixed if you update the images ;)

I used the 0.1 percent tolerance because I noticed we were using it elsewhere.

PR #1106, opened 2016-12-08
implement collection view prefetching

joewalsh:


coreyfloyd:
Not sure if you are interested in doing the core data apart as well. Not sure if its worth the performance or not.

> Speaking of NSFetchResultsController, if you're using UICollection view, sorry there we go, UICollection view, it's really easy to adopt the new data source prefetching feature if you're using core data.
> All you need is an asynchronous fetch request, to get the request off the main thread and you want to make sure that you're not returning objects as faults.
> For more information about data source prefetching, check out Steve and Peter's talk from yesterday, what's new in UICollection view.


http://asciiwwdc.com/2016/sessions/242

PR #1107, opened 2016-12-08
match font size multiplier default to system setting

joewalsh:


Wikipedia/Code/NSUserDefaults+WMFApplicationDefaults.swift
```diff
@@ -0,0 +1,44 @@
+import Foundation
+
+extension NSUserDefaults {
+    public func wmf_setReadingFontSize(fontSize: NSNumber) {
+        self.setObject(fontSize, forKey: "ReadingFontSize")
+        self.synchronize()
+    }
+    
+    public func wmf_readingFontSize() -> NSNumber {
```
montehurd: Can we rename this to 'wmf_fontSizeMultiplier' or 'wmf_webViewFontSizeMultiplier'?

Wikipedia/Code/NSUserDefaults+WMFApplicationDefaults.swift
```diff
@@ -0,0 +1,44 @@
...
+        if let fontSize = self.objectForKey("ReadingFontSize") as? NSNumber {
+            return fontSize
+        } else {
+            let preferredContentSizeCategory = UIApplication.sharedApplication().preferredContentSizeCategory
+            switch preferredContentSizeCategory {
+            case UIContentSizeCategoryExtraSmall:
+                return NSNumber(integer: 70)
```
montehurd: Maybe use constants for these `70, 85, 100`... etc sizes so we can DRY between this method and those in `- (NSArray<NSNumber *> *)fontSizeMultipliers`?

montehurd:
The article text size appears a bit bigger than it should be at the default size setting... taking a closer look...

montehurd:
I just double-checked on my device and it looks like there are a coupe of issues:

- The first one is the default text size on the web view article appears to have increased on my device when I leave it's Dynamic font adjustment slider on the default value. You can see this by just loading an article from the feed. The feed extract text was formerly about the same size as the web view article text but now the web view article text appears a step or two larger.

- Second issue is if I go to settings and adjust my dynamic text size preference, then go back to the app, the web view text size slider isn't in the new position, even if I go back to explore and load a different article. i.e. if I set the system's dynamic type slider to the 4th tick, when I go back to the app the web view text size slider should be at the 4th tick as well I believe.

joewalsh:
@montehurd the issue with the default text size is that the app's default is the third tick:
simulator screen shot dec 8 2016 8 59 57 pm
and the system default is the middle tick:
simulator screen shot dec 8 2016 9 00 01 pm

I updated them to match - but that means we run out of sizes on the small end.

Also, as per the phab ticket, if you set the text size at all using the app slider, it no longer moves with the system setting. So if you've set it at all at any point with the app slider, it won't move with the system setting. If I do a clean install, then view an article, then change the system size, I verified that the text size and slider position do change.

joewalsh:
another alternative is to tweak the scale so instead of 15% increments it works as follows:
```
@objc enum WMFFontSizeMultiplier: Int {
    case ExtraSmall = 70
    case Small = 80
    case Medium = 90
    case Large = 100
    case ExtraLarge = 110
    case ExtraExtraLarge = 130
    case ExtraExtraExtraLarge = 160
}
```

montehurd:
> another alternative is to tweak the scale so instead of 15% increments it works as follows 

Ya that sounds reasonable.

joewalsh:
just found https://phabricator.wikimedia.org/T148935 which describes exactly how to deal with the slider mismatch issue - updating this PR

PR #1118, opened 2016-12-14
Fix for tapping on grey area not dismissing ref panel on phone.

montehurd:
Unsure why state `.ended` is not seen when using bk_recognizer... (it's always `.possible` but never `.ended`)

### DIFF
--- a/Wikipedia/Code/WMFReferencePanelViewController.swift
+++ b/Wikipedia/Code/WMFReferencePanelViewController.swift
@@ -1,6 +1,5 @@
 
 import Foundation
-import BlocksKitUIKitExtensions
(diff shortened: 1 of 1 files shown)


montehurd:
https://phabricator.wikimedia.org/T153156

joewalsh:
@montehurd good catch - this was the same underlying issue that caused the header/footer tap problem you found last week. I fixed the problem in BlocksKit but ended up switching to standard gesture recognizers like you did here. Ultimately BlocksKit wasn't providing all that much of a convenience in these cases. https://github.com/wikimedia/wikipedia-ios/pull/1120

PR #1128, opened 2016-12-14
Fix empty view being shown after deleting all

coreyfloyd:


joewalsh:
addressed this here: https://github.com/wikimedia/wikipedia-ios/pull/1129/commits/cb29da4e3cb5e3412cc00a581b60d045aeeede8a

PR #1132, opened 2016-12-16
Explore layout bugs

joewalsh:


montehurd:
Seems to work but size isn't being retrieved if the announcement goes offscreen then back onscreen:
111 mov

WMFModel/WMFAnnouncementCollectionViewCell.m
```diff
@@ -49,12 +49,13 @@ - (void)prepareForReuse {
 
 - (void)setImageURL:(NSURL *)imageURL {
     if (imageURL) {
-        [self.imageView wmf_setImageWithURL:imageURL
-                                detectFaces:YES
-                                    failure:^(NSError *_Nonnull error) {
-                                    }
-                                    success:^{
-                                    }];
+        __block UIImage *img = [[UIImage alloc] init];
```
montehurd: Not sure why, but the face aware method was messing up the layout...

Wikipedia/Code/WMFExploreCollectionViewCell.m
```diff
@@ -4,26 +4,17 @@
 @implementation WMFExploreCollectionViewCell
 
 - (UICollectionViewLayoutAttributes *)preferredLayoutAttributesFittingAttributes:(UICollectionViewLayoutAttributes *)attributesToFit {
-    if ([attributesToFit isKindOfClass:[WMFCVLAttributes class]] && [(WMFCVLAttributes *)attributesToFit precalculated]) {
-        return attributesToFit;
```
montehurd: Bailing here seemed to prevent the size from being calculated - maybe when cells were recycled?

WMFModel/WMFAnnouncementCollectionViewCell.m
montehurd: Interestingly, if you apply just this chunk to "spike/swift3", the flicker problem seems to mostly go away...

montehurd:
It looks like https://github.com/wikimedia/wikipedia-ios/pull/1132/files/23317a361d75bb184415f67e8e5782028fabbc38..7ffd38af515edd859fa4e5250bfdddc29be6f500#r92910467 fixes the problem all by itself w/o the rest of this PR's changes...

montehurd:
Here's a PR with the announcement image fetch flicker fix... https://github.com/wikimedia/wikipedia-ios/pull/1134

Wikipedia/Code/WMFExploreCollectionViewCell.m
joewalsh: `precalculated` is set when the layout was calculated ahead of time and returned for the estimated layout. In theory, that should be the correct height so it doesn't need to be re-calculated here. Unfortunately in practice, it seems auto-layout gives two different height results.

PR #1133, opened 2016-12-16
Set Image Download/Article download prio

joewalsh:


joewalsh:
!jenkins test

coreyfloyd:
@joewalsh So I guess it was the shared process pool that was causing the issue?

Only question I have is how is memory consumption now with decent size stacks of WebViews (say like 5-10 pages in a stack)? I know the shared process pool was used to help manage resources with large navigation stacks

joewalsh:
@coreyfloyd yes - 10+ articles on the stack on my older iPad works fine and added a note for testers on the ticket to verify this on other older devices

PR #1134, opened 2016-12-17
Fix for explore flicker on clean install.

montehurd:


montehurd:
Crap... didn't fix when using network link conditioner...

PR #1135, opened 2016-12-19
compare article keys rather than URLs to ensure the same article with different fragments isn't pushed twice

joewalsh:


joewalsh:
!jenkins test

PR #1147, opened 2016-12-23
Increase distance cell vertical compression resistance.

montehurd:
https://phabricator.wikimedia.org/T154062

img_2344


joewalsh:
Will this work with a compression resistance of 751 or 999? Also, was it tested on iPad?

montehurd:
@joewalsh checked on develop and can no longer repro! :)

PR #1150, opened 2017-01-03
[Bug] Fix an issue with running `make bootstrap`

samsymons:
The `bootstrap` command had extra indentation which caused the command to fail when I ran it.

I've fixed that issue, and added an additional check on `xcode-select`, since it will throw an error if you pass it `--install` when Xcode is already installed. I ran this script on my machine after these two fixes and it ran through to completion.


joewalsh:
thanks!

PR #1153, opened 2017-01-03
don't use NSAttributedString initWithData

joewalsh:
potentially causing our top crash


montehurd:
NSHTMLTextDocumentType makes me sad... why is it so bad?!

PR #1155, opened 2017-01-04
[Bug] Prevent image duplication in galleries

samsymons:
This PR fixes an issue where images were duplicated in the gallery if they were referenced multiple times in an article. It uses an ordered set to de-duplicate the image URLs before being turned into images, for both the gallery URLs and URLs for saving.

It also appeared that the `imagesForSaving` method was not being used anywhere, so I've removed that as well.

Fixes [T153260](https://phabricator.wikimedia.org/T153260), [T148499](https://phabricator.wikimedia.org/T148499).


Wikipedia/Code/MWKArticle.m
```diff
@@ -357,7 +357,8 @@ - (MWKProtectionStatus *)requiredProtectionStatus:(NSString *)key dict:(NSDictio
             }
         }
     }
-    return imageURLs;
+
+    return [[NSOrderedSet orderedSetWithArray:imageURLs] array];
```
montehurd: I think the de-dupe could (should?) probably happen inside the 2 methods in `WMFImageTagList+ImageURLs.m` instead...

Wikipedia/Code/MWKArticle.m
```diff
@@ -368,13 +369,9 @@ - (MWKProtectionStatus *)requiredProtectionStatus:(NSString *)key dict:(NSDictio
...
     WMFImageTagList *tagList = [[[WMFImageTagParser alloc] init] imageTagListFromParsingHTMLString:self.articleHTML withBaseURL:self.url];
-    return [tagList imageURLsForSaving];
-}
+    NSArray *imageURLs = [tagList imageURLsForSaving];
 
-- (NSArray<MWKImage *> *)imagesForSaving {
-    return [[self imageURLsForSaving] bk_map:^id(NSURL *url) {
-        return [[MWKImage alloc] initWithArticle:self sourceURL:url];
-    }];
+    return [[NSOrderedSet orderedSetWithArray:imageURLs] array];
```
montehurd: Same as above.

samsymons:
@montehurd Agreed, that is a better place for this code. I'll push an update for this, and update the unit tests to match.

samsymons:
Closing this ticket for now — caught up with travel at the moment, so I will update this branch and resubmit once I have time.

PR #1162, opened 2017-01-06
Always request and use fetched edit token.

montehurd:
https://phabricator.wikimedia.org/T153106

Also removed now unused persisting of edit tokens to keychain.


Wikipedia/Code/PreviewAndSaveViewController.m
```diff
@@ -501,43 +501,25 @@ - (void)fetchFinished:(id)sender
...
-                    self.keychainCredentials.editTokens;
-                NSString *domain = self.section.url.wmf_language;
-                if (domain && tokenFetcher.token) {
-                    editTokens[domain] = tokenFetcher.token;
-                    self.keychainCredentials.editTokens = editTokens;
-                }
-                upload();
+                EditTokenFetcher *tokenFetcher = (EditTokenFetcher *)sender;
+
+                (void)[[WikiTextSectionUploader alloc] initAndUploadWikiText:tokenFetcher.wikiText
```
joewalsh: seems like the `WikiTextSectionUploader` (and the `EditTokenFetcher` initialized on line 675) should be assigned to properties and retained. Having unowned objects like this could lead to them being dealloc'd and not actually completing the fetch.

montehurd: Ah, good catch. Want me to fix that as part of this PR?

joewalsh: Yeah, would be good to fix it in this PR

PR #1172, opened 2017-01-25
Refactor authentication bits and add "Forgot password" button and interfaces

montehurd:
This adds a "Forgot your password?" link to the login page and presents a functional interface for doing so when tapped. 

However, we discovered we needed to present to the user a different login interface if they try to log in with the temporary password sent to their email address...

-----

The `clientlogin` API (which we use now for login) indicates a new password must be entered by returning a `status` of `UI`:
https://www.mediawiki.org/wiki/API:Login#clientlogin

Once new pwd is obtained from the user we'll need to send it along with the previous login token and other fields. I believe all of the following:
- action=clientlogin
- logincontinue=1
- logintoken=*same token* (update: gergo said requesting another token for this step is ok - will be issued same token as login call)
- password=*new pwd*
- retype=*new pwd* (don't allow resubmit unless these match)

-----

So some login networking and UI need to be refactored to make this happen. Instead of continuing to wrangle crufty old delegation based networking code, I'm preparing to refactor some of those old fetchers to use Swift/blocks...

**The plan:**
- [x] fold the old delegate style token fetchers (LoginTokenFetcher and AccountCreationTokenFetcher) into my new WMFCSRFTokenFetcher (rename it to WMFAuthTokenFetcher)
- [x] Swift-ify/block-ify AccountLogin and AccountCreator fetchers
- [x] make new Swifty/blocky WMFPasswordResetter
- [x] Modify login workflow to present the change password interface at appropriate time


Wikipedia/Code/AccountCreationViewController.swift
```diff
@@ -0,0 +1,285 @@
...
+    @IBOutlet var createAccountContainerView: UIView!
+
+    fileprivate var captchaId: NSString? = ""
+    fileprivate var rightButton: UIBarButtonItem?
+    public var funnel: CreateAccountFunnel?
+    fileprivate var captchaViewController: CaptchaViewController?
+    
+    func scrollViewDidScroll(_ scrollView: UIScrollView) {
+        // Disable horizontal scrolling.
+        scrollView.contentOffset = CGPoint(x: 0.0, y: scrollView.contentOffset.y)
```
coreyfloyd: I think you can just turn off bounceHorizontally and make sure the content width is the same as the width of the scroll view.

Wikipedia/Code/AccountCreationViewController.swift
```diff
@@ -0,0 +1,285 @@
...
+        // Reminder: spaceBeneathCaptchaContainer constraint is space *below* captcha container -
+        // that's why below for the show case we don't have to "convertPoint".
+        spaceBeneathCaptchaContainer.constant = (showCaptchaContainer)
+            ? (view.frame.size.height - (captchaContainer.frame.size.height / 2))
+            : (view.frame.size.height - loginButton.convert(CGPoint.zero, to:scrollView).y)
+    }
+
+    override func viewWillTransition(to size: CGSize, with coordinator: UIViewControllerTransitionCoordinator) {
+        super.viewWillTransition(to: size, with: coordinator)
+        // Ensure adjustScrollLimitForCaptchaVisiblity gets called again after rotating.
```
coreyfloyd: Won't this be called before rotating? (viewWILLTransistion) Maybe thats ok, just noticed.

Also if you want the change to be animated you can call it within the coordinater's animation block.

Wikipedia/Code/AccountCreationViewController.swift
montehurd: Oh I'll check... was carry-over from old obj-c.

Wikipedia/Code/AccountCreationViewController.swift
montehurd: I'll check... more carry-over from the old obj-c.

Wikipedia/Code/AccountCreationViewController.swift
joewalsh: I don't see`alwaysBounceHorizontal` enabled here and it should be off by default. Is it enable d in the nib? The other key is ensuring the scrollView's `contentSize.width` is less than or equal to its `bounds.size.width`

Wikipedia/Code/AccountCreationViewController.swift
joewalsh: ```public func animate(alongsideTransition animation: ((UIViewControllerTransitionCoordinatorContext) -> Swift.Void)?, completion: ((UIViewControllerTransitionCoordinatorContext) -> Swift.Void)? = nil) -> Bool```

Wikipedia/Code/AccountCreationViewController.swift
```diff
@@ -0,0 +1,285 @@
...
+    func didTapClose(_ tap: UITapGestureRecognizer) {
+        if (showCaptchaContainer) {
+            showCaptchaContainer = false
+        } else {
+            self.dismiss(animated: true, completion: nil)
+        }
+    }
+
+    func loginButtonPushed(_ recognizer: UITapGestureRecognizer) {
+        if recognizer.state == .ended {
```
joewalsh: This should be a `guard` with an early return to limit indentation.

```
guard recognizer.state == .ended, 
     let presenter = self.presentingViewController else {
           return
}
```

Wikipedia/Code/AccountCreationViewController.swift
```diff
@@ -0,0 +1,285 @@
...
+                        self.textFieldDidChange(nil)
+                    })
+                })
+            }
+        }
+    }
+    
+    private var _captchaUrl: URL? = nil
+    
+    fileprivate var captchaUrl: URL? {
```
joewalsh: This can be replaced with a single `var` and a `didSet {`

```
fileprivate var captchaURL: URL? {
     didSet {
          guard let newURL = captchaURL else {
               return;
          }
          refreshCaptchaImage()
     }
}
```

Also, URL should be all caps

Wikipedia/Code/AccountCreationViewController.swift
```diff
@@ -0,0 +1,285 @@
...
+        save()
+    }
+    
+    override func updateViewConstraints() {
+        adjustScrollLimitForCaptchaVisiblity()
+        super.updateViewConstraints()
+    }
+
+    fileprivate func save() {
+        if !areRequiredFieldsPopulated() {
```
joewalsh: These ifs should be guard statements so that the compiler enforces the early return

Wikipedia/Code/LoginViewController.swift
```diff
@@ -0,0 +1,138 @@
...
+        }, failure: {
+            (error: Error) in
+            self.enableProgressiveButton(true)
+            WMFAlertManager.sharedInstance.showErrorAlert(error as NSError, sticky: true, dismissPreviousAlerts: true, tapCallBack: nil)
+            self.funnel?.logError(error.localizedDescription)
+        })
+    }
+    
+    func forgotPasswordButtonPushed(_ recognizer: UITapGestureRecognizer) {
+        if recognizer.state == .ended {
```
joewalsh: Same comment about using a guard and an early return as above

Wikipedia/Code/AccountCreationViewController.swift
montehurd: Fixed!

Wikipedia/Code/AccountCreationViewController.swift
montehurd: Good catch. didSet is awesome.

montehurd: Fixed.

Wikipedia/Code/AccountCreationViewController.swift
montehurd: Fixed.

Wikipedia/Code/LoginViewController.swift
montehurd: Fixed.

Wikipedia/Code/AccountCreationViewController.swift
montehurd: Fixed via storyboard.

Wikipedia/Code/AccountCreationViewController.swift
montehurd: Fixed!

montehurd:
@joewalsh I think I addressed all of the code review feedback. 

(Working on the 2FA bits now but would rather submit 2FA as a follow-on.)

PR #1175, opened 2017-02-01
Initial places implementation

joewalsh:


Wikipedia/Code/QuadKey.swift
```diff
@@ -0,0 +1,298 @@
+//https://msdn.microsoft.com/en-us/library/bb259689.aspx
+//http://wiki.openstreetmap.org/wiki/QuadTiles
+
+public typealias QuadKey = UInt64
+public typealias QuadKeyPart = UInt32
+public typealias QuadKeyPrecision = UInt16
+public typealias QuadKeyDegrees = Double
+
+public extension QuadKeyPrecision {
+    public static let maxPrecision: QuadKeyPrecision = 32
```
montehurd: This is probably a stupid question, but why isn't this 16?

Wikipedia/Code/UIViewController+WMFAlerts.swift
```diff
@@ -8,4 +8,10 @@ extension UIViewController {
         present(alert, animated: true, completion: nil)
     }
     
+    func wmf_showAlertWithMessage(_ message: String) {
+        let alert = UIAlertController(title: nil, message: message, preferredStyle: .alert)
+        alert.addAction(UIAlertAction(title: "OK", style:.default, handler: nil))
```
montehurd: probably want a localized string for "OK" here

Wikipedia/Code/WMFLocationSearchFetcher.h
```diff
@@ -5,6 +5,14 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+
+typedef NS_ENUM(NSUInteger, WMFLocationSearchSortStyle) {
+    WMFLocationSearchSortStyleNone = 0,
```
montehurd: Can you use interference here too? `None` vs `WMFLocationSearchSortStyleNone`

Wikipedia/Code/WMFLocationSearchFetcher.m
```diff
@@ -113,24 +148,61 @@ - (nullable NSURLRequest *)requestBySerializingRequest:(NSURLRequest *)request
...
-        @"pilimit": numberOfResults,
-        @"wbptterms": @"description",
-        @"generator": @"geosearch",
-        @"ggscoord": coords,
-        @"codistancefrompoint": coords,
-        @"ggsradius": @"10000",
-        @"ggslimit": numberOfResults,
-        @"format": @"json"
-    };
+    if (params.region.radius >= 10000 || params.searchTerm || params.sortStyle != WMFLocationSearchSortStyleNone) {
```
montehurd: MEGA `IF`! ;)

Wikipedia/Code/WMFTableOfContentsPresentationController.swift
```diff
@@ -71,23 +71,23 @@ open class WMFTableOfContentsPresentationController: UIPresentationController {
             _ = make?.width.equalTo()(44)
             _ = make?.height.equalTo()(44)
             switch self.displaySide {
-            case WMFTableOfContentsDisplaySideLeft:
+            case .left:
```
montehurd: Nice and clean. I forgot NS_ENUM made this possible for Obj-C enums.

Wikipedia/Code/WikidataFetcher.swift
```diff
@@ -0,0 +1,90 @@
...
+            let keys = ["P1332", "P1333", "P1334", "P1335"] //bounding coordinates
+
+            let coordinates = keys.flatMap({ (key) -> CLLocationCoordinate2D? in
+                guard let values = claims[key] as? [Any] else {
+                    return nil
+                }
+                guard let point = values.first as? [String: Any] else {
+                    return nil
+                }
+                guard let mainsnak = point["mainsnak"] as? [String: Any] else {
```
montehurd: lol `mainsnak`. this doesn't have anything to do with lunch does it? ;)

Wikipedia/Localizations/en.lproj/Localizable.strings
```diff
@@ -545,3 +545,9 @@
 "status-saved-for-later" = "Saved for later";
 
 "announcements-dismiss" = "No thanks";
+
+"places-search-top-articles" = "Top articles";
+"places-search-search-mapped-area" = "Search mapped area";
```
montehurd: Wondering if "mapped" will give translators any issues...

Wikipedia/Localizations/qqq.lproj/Localizable.strings
```diff
@@ -474,3 +474,9 @@
 "status-saving-for-later" = "Indicates to the user that the article is being saved for later";
 "status-saved-for-later" = "Indicates to the user that the article has been saved for later";
 "announcements-dismiss" = "Button text indicating a user wants to dismiss an announcement\n{{Identical|No thanks}}";
+
+"places-search-top-articles" = "A search suggestion for top articles";
+"places-search-search-mapped-area" = "A button title that indicates the search will be redone in the mapped area";
```
montehurd: May want to consider describing this without using the word "mapped" just in case. 

Maybe something like...
> "Title for search button for finding results which are located within the currently visible area on the map"

WikipediaUnitTests/Code/QuadKeyTests.swift
```diff
@@ -0,0 +1,57 @@
+import XCTest
+@testable import Wikipedia
+
+class QuadKeyTests: XCTestCase {
+
```
montehurd: A brief comment would be nice for future reference. This is a bit... inscrutable :)

montehurd:
Looks great!! Just a couple tiny comments. 👍

Wikipedia/Code/QuadKey.swift
joewalsh: so there are two bits necessary to store each precision level, 32 total fit in the 64 bit integer of the quad key - this is the maximum precision we can store

the view controller uses 16 as the max because that's the max level that we care about grouping locations together - below that we'd rather all individual pins be visible

Wikipedia/Code/UIViewController+WMFAlerts.swift
joewalsh: good catch

Wikipedia/Code/WikidataFetcher.swift
joewalsh: lol, yeah i wish

sadly, no
https://www.mediawiki.org/wiki/Wikibase/DataModel/JSON#Snaks

Wikipedia/Localizations/en.lproj/Localizable.strings
joewalsh: After design review this was changed to "search this area" - i also removed mapped from the key

WikipediaUnitTests/Code/QuadKeyTests.swift
joewalsh: Added comments to this and QuadKey.swift - let me know if they're helpful or could be clarified or expanded

montehurd:
@joewalsh merge and do rest as follow-on?

PR #1176, opened 2017-02-07
Two factor authentication.

montehurd:
https://phabricator.wikimedia.org/T150699

2fa mov

- [x] Hook up 2FA.
- [x] Fix changing pwd interface not appearing if you log in with temp pwd when 2FA is on.
- [x] Ask Josh if we want auto-login between re-installs when 2FA is on. (*answer is no - issue is cookies are gone so we'd have to present the 2FA token interface*)
- [x] Add scrollviews to pwd change, 2FA and forgot pwd storyboards (for landscape when keyboard can cover form fields).
- [x] Use 6 number fields for 2FA token instead of single field.


jdforrester:
Fancy.

PR #1178, opened 2017-02-09
show article lead image when tapping share-a-fact

julienbodet:
Hi! 
I don't know if there is a better fix for this bug. But this one works.

Note: After tapping "share a fact", a reload of the article view seems to happen (a white screen appears for some milliseconds) before the completion block sets the image and presents the overlay view.
Do you have any idea why that behaviour happens?
Thanks.

https://phabricator.wikimedia.org/T155759


joewalsh:
@julienbodet this looks great, thanks for the fix! I was unable to repro the white flash when tapping share a fact - is it happening for you with all articles or just some?

julienbodet:
@joewalsh I tested again and I figured out it's was also happening before my fix, for all articles. To see what I am talking about, you can compare by tapping the "Share" button. Here is 2 GIFs showing that:
test_2
test_3

PR #1179, opened 2017-02-09
Follow-up refactors to auth manager and other login/acct creation bits

montehurd:
https://phabricator.wikimedia.org/T158145

These are follow-up changes to the recent [refactor](https://github.com/wikimedia/wikipedia-ios/pull/1172) of login/acct creation bits.

Now that the authentication networking bits are blocks based the auth manager can be fixed to not park blocks all over (which it had to do to bridge blocks based methods to the old delegation ones).

- [x] swift-ify remaining auth manager login methods
- [x] then switch all the new authentication fetchers to use swift style enum errors (w associated values where needed) once final swift conversion is done
- [x] new authentication fetchers won't need to inherit from NSObject or use `@objc` either (one or 2 exceptions)
- [x] swift-ify the *ancient* keychain credentials object

_Legacy auth Obj-C is gone! :)_
screen shot 2017-02-14 at 2 18 01 pm


Wikipedia/Code/WMFAuthAccountCreationInfoFetcher.swift
```diff
@@ -22,6 +22,17 @@ public class WMFAuthAccountCreationInfo: NSObject {
         self.canCreateAccounts = canCreateAccounts
         self.captchaID = captchaID
         self.captchaURLFragment = captchaURLFragment
+    }    
+    var captchaImageURL: URL? {
+        guard let appLang = MWKLanguageLinkController.sharedInstance().appLanguage else {
+            return nil
+        }
+        let siteURL = appLang.siteURL() as NSURL
+        guard let domain = siteURL.wmf_domain else {
```
montehurd: Hey @joewalsh, 
Will we need to convert NSURL category methods from `#import "NSURL+WMFLinkParsing.h"` to Swift URL extensions before we can use them with Swift URLs?

I added `#import "NSURL+WMFLinkParsing.h"` to `Wikipedia-Bridging-Header.h` but wasn't able to call `wmf_domain` without casting to NSURL...

Wikipedia/Code/WMFAuthenticationManager.swift
```diff
@@ -0,0 +1,127 @@
...
+    private let tokenFetcher = WMFAuthTokenFetcher()
+    private let accountLogin = WMFAccountLogin()
+    private let currentlyLoggedInUserFetcher = WMFCurrentlyLoggedInUserFetcher()
+    
+    /**
+     *  Get the shared instance of this class
+     *
+     *  @return The shared Authentication Manager
+     */
+    fileprivate static let _sharedInstance = WMFAuthenticationManager()
```
montehurd: Should I just change this to... 
`public static let sharedInstance = WMFAuthenticationManager()`
... and get rid of the sharedInstance() func? That is, any opposition to or reason not to have the sharedInstance be a property instead of func?

joewalsh: having this as a property instead of a func should be fine

WMF Framework/WMFCaptchaViewController.swift
```diff
@@ -37,18 +41,23 @@ class WMFCaptchaViewController: UIViewController {
...
+        }
+
         // Allow whatever view controller is using this captcha view controller
         // to monitor changes to captchaTextBox and also when its keyboard done/next
         // buttons are tapped.
         
-        if self.parent!.conforms(to: WMFCaptchaViewControllerRefresh.self){
-            captchaTextBox.delegate = self.parent as? UITextFieldDelegate
+        if parentVC.conforms(to: WMFCaptchaViewControllerRefresh.self){
+            captchaTextBox.delegate = parent as? UITextFieldDelegate
```
joewalsh: same here - ```guard let parentVC = parent as? UITextFieldDelegate, parentVC.conforms(to: WMFCaptchaViewControllerRefresh.self)```

WMF Framework/WMFCaptchaViewController.swift
```diff
@@ -27,8 +27,12 @@ class WMFCaptchaViewController: UIViewController {
     }
     
     func reloadCaptchaPushed(_ sender: AnyObject) {
-        if self.parent!.responds(to: #selector(reloadCaptchaPushed(_:))) {
-            self.parent!.performSelector(onMainThread: #selector(reloadCaptchaPushed(_:)), with: nil, waitUntilDone: true)
+        guard let parentVC = parent else{
+            assert(false, "Expected parent view controller not found")
+            return
+        }
+        if parentVC.responds(to: #selector(reloadCaptchaPushed(_:))) {
```
joewalsh: this all could be moved up into the guard statement

Wikipedia/Code/WMFAuthenticationManager.swift
montehurd: Ah cool will fix.

Wikipedia/Code/WMFAccountCreator.swift
```diff
@@ -1,38 +1,25 @@
...
-        return type.rawValue
-    }
-    public var errorUserInfo: [String : Any] {
-        var userInfo = [String: Any]()
-        if let localizedDescription = self.localizedDescription {
-            userInfo[NSLocalizedDescriptionKey] = localizedDescription
+    public var errorDescription: String? {
+        switch self {
+        case .cannotExtractStatus:
+            return "Could not extract status"
```
joewalsh: Do these need to be localized?

montehurd: I doubt these will ever be shown to the user...

WMF Framework/WMFCaptchaViewController.swift
montehurd: Fixed.

WMF Framework/WMFCaptchaViewController.swift
montehurd: Fixed!

Wikipedia/Code/WMFAuthenticationManager.swift
montehurd: Fixed.

PR #1180, opened 2017-02-10
Add link to alternative account creation methods

montehurd:
https://phabricator.wikimedia.org/T154728

req mov


Wikipedia/Localizations/en.lproj/Localizable.strings
```diff
@@ -49,6 +49,9 @@
 "account-creation-captcha-required" = "CAPTCHA verification required.";
 "account-creation-captcha-retry" = "Incorrect entry. Please try again.";
 "account-creation-captcha-obtaining" = "Obtaining a new CAPTCHA...";
+"account-creation-captcha-title" = "CAPTCHA security check";
+"account-creation-captcha-cannot-see-image" = "Can't see the image? $1";
+"account-creation-captcha-request-account" = "Request an account.";
 "account-creation-logging-in" = "Logging in...";
```
montehurd: Quick note for Andy: localized strings added to this file get automatically added to translatewiki.net. We periodically get translations back - about weekly, as PRs, which are then tested (occasionally translators will accidentally add things like html comments which are fine if the string was going to used with html for a web page but which don't make sense for the native app).

Here's a link to the iOS TWN page: https://translatewiki.net/w/i.php?title=Special:Translate&group=out-wikimedia-mobile-wikipedia-ios&filter=translated&language=de&action=translate
Usually we don't have to do anything with this, but it's nice to see where the translations actually happen.

Wikipedia/Localizations/qqq.lproj/Localizable.strings
```diff
@@ -59,6 +59,9 @@
 "account-creation-captcha-required" = "Alert message shown when account creation require a captcha verification";
 "account-creation-captcha-retry" = "Alert message shown when the captcha the user entered did not match the captcha image";
 "account-creation-captcha-obtaining" = "Alert shown when user wants a new captcha when creating account";
+"account-creation-captcha-title" = "Title for account creation CAPTCHA interface";
```
montehurd: Any time a new translation string is added this is the place you must add a description of the string. This is presented to translators on TWN to provide context.

Wikipedia/Code/WMFAccountCreationViewController.swift
```diff
@@ -97,8 +101,34 @@ class WMFAccountCreationViewController: UIViewController, WMFCaptchaViewControll
...
+        
+        // Reminder: used a label instead of a button for subtitle because of multi-line string issues with UIButton.
+        captchaSubtitleLabel.attributedText = getCaptchaSubtitleAttributedString()
+        captchaSubtitleLabel.addGestureRecognizer(UITapGestureRecognizer(target: self, action: #selector(self.requestAnAccountTapped(_:))))
+
+        view.wmf_configureSubviewsForDynamicType()
     }
 
+    func requestAnAccountTapped(_ recognizer: UITapGestureRecognizer) {
+        wmf_openExternalUrl(URL.init(string: "https://en.wikipedia.org/wiki/Wikipedia:Request_an_account"))
```
amrox: I guess this page only exists on the english site? Or is it worth trying to localize?

montehurd: Good point. I just double-checked with Carolyn and she said this is ok for now (until we can figure out the other pages).

montehurd:
@joewalsh this should be good to go.

Wikipedia/Code/WMFAccountCreationViewController.swift
```diff
@@ -97,8 +101,34 @@ class WMFAccountCreationViewController: UIViewController, WMFCaptchaViewControll
...
+    }
+        
+    fileprivate func getCaptchaSubtitleAttributedString() -> NSAttributedString {
```
joewalsh: stylistically, this should probably be 
```
fileprivate var captchaSubtitleAttributedString: NSAttributedString {
    get {
        -implementation here-
    }
}
```

joewalsh:
!jenkins test

Wikipedia/Code/WMFAccountCreationViewController.swift
montehurd: Fixed!

PR #1181, opened 2017-02-13
Open keyboard when clear button of 'Find in page' textfield is touched

julienbodet:
This PR resolves this bug: https://phabricator.wikimedia.org/T143024

In the collapse view, if the user touches the clear button, re-open the keyboard to enter new text.


montehurd:
Seems solid! 👍

PR #1183, opened 2017-02-13
Heading arrow & action title updates for places

joewalsh:


joewalsh:
!jenkins test

Wikipedia/Code/ArticlePopoverViewController.swift
```diff
@@ -19,11 +19,29 @@ class ArticlePopoverViewController: UIViewController {
...
-    var article: WMFArticle?
+    let article: WMFArticle
+    
+    required init(_ article: WMFArticle) {
+        self.article = article
+        super.init(nibName: "ArticlePopoverViewController", bundle: nil)
+    }
+    
+    required init?(coder: NSCoder) {
+        return nil
```
joewalsh: So it's instantiated using `initWithNibName` (see line 26 above) but that doesn't seem to call `initWithCoder`. The other option would be to have `initWithCoder` look like this:
```
    required init?(coder: NSCoder) {
        self.article = WMFArticle()
        super.init(coder: coder)
    }
```
However, it seemed more appropriate to fail & return nil since it's not initializing with a valid article. Open to other suggestions though.

amrox: Nah, I like the fail and return.

PR #1184, opened 2017-02-13
Fix for T153909: Header image appears too large on iPad

amrox:
https://phabricator.wikimedia.org/T153909

 - [x] Contrain height to 400px
 - [x] exclude these images from face detection and use aspect fit built-in


joewalsh:
UIImageViewAligned looks handy, however, it appears to conflict with our existing face detection. For example, on the Obama article, the face should be centered:
simulator screen shot feb 14 2017 6 58 18 am
vs with UIImageViewAligned:
simulator screen shot feb 14 2017 6 57 39 am

Rather than using UIImageViewAligned, you could alter our existing face detection code (which sets the crop box on the image view using `layer.contentsRect`) to set an area just below the top as the crop rect. The place to look is `UIImageView+WMFImageFetchingInternal.m` line 164:
```
- (void)wmf_setImage:(UIImage *)image
     faceBoundsValue:(nullable NSValue *)faceBoundsValue
            animated:(BOOL)animated
             failure:(WMFErrorHandler)failure
             success:(WMFSuccessHandler)success {
    NSAssert([NSThread isMainThread], @"Interaction with a UIImageView should only happen on the main thread");
    CGRect faceBounds = [faceBoundsValue CGRectValue];
    if (!CGRectIsEmpty(faceBounds)) {
        [self wmf_cropContentsByVerticallyCenteringFrame:[image wmf_denormalizeRect:faceBounds]
                                     insideBoundsOfImage:image];
    } else {
        [self wmf_resetContentsRect];
    }
```
^ `wmf_resetContentsRect` resets it to the whole frame, causing the image to be centered. I think it could set a rect along the lines of `CGRectMake(0, 0, 1, 0.5)` to cause the image to be top-aligned.

The other thing is that `T153909` specifically pertains to the view that appears when tapping a news item `InTheNewsViewController`. It's not super prominent in the ticket, and it looks almost identical to the article view header image, so I can see the cause for confusion. Really it's two issues - 1. When face detection fails app-wide, the default crop rect should be the top portion of the image rather than the center 2. The image scaling in the "In the news" view controller is too large on iPad

PR #1185, opened 2017-02-15
Top-aligning images if face detection fails

amrox:
Related to https://phabricator.wikimedia.org/T153909

This change will top-align images in an UIImageView when face-detection fails.

Questions:
 - Is top-alignment for header images really what we want from a design perspective?
 - Are there any code paths which we still want the centering behavior?


montehurd:
We'll never merge a WIP PR until you remove the WIP tag. "Do Not Merge Yet" is more explicit, but it's fine to leave things "WIP" too :)

amrox:
@montehurd I removed WIP and added Do Not Merge Yet, because the work is done, I'm just not sure the change is actually wanted. See questions in the description. Going to remove the Do Not Merge Tag and see what happens :)

PR #1192, opened 2017-02-21
Checking in face bounds are less than 5% of the image

amrox:
Ticket: https://phabricator.wikimedia.org/T156217


Wikipedia/Code/UIImageView+WMFImageFetchingInternal.m
```diff
@@ -168,7 +168,8 @@ - (void)wmf_setImage:(UIImage *)image
              success:(WMFSuccessHandler)success {
     NSAssert([NSThread isMainThread], @"Interaction with a UIImageView should only happen on the main thread");
     CGRect faceBounds = [faceBoundsValue CGRectValue];
-    if (!CGRectIsEmpty(faceBounds)) {
+    CGFloat faceArea = faceBounds.size.width * faceBounds.size.height;
```
amrox: Since we're in normalized coordinates, I'm fairly sure we can calculate the area percentage just by multiplying the bounds width and height

Wikipedia/Code/UIImageView+WMFImageFetchingInternal.m
```diff
@@ -168,7 +168,8 @@ - (void)wmf_setImage:(UIImage *)image
...
+    if (!CGRectIsEmpty(faceBounds) && (faceArea >= 0.25)) {
```
amrox: - [ ] Is there a better place to put this constant?

PR #1194, opened 2017-02-22
Unit Test Suite should work on at least one iPad model

amrox:
- [x] Automatically run tests on reference Phone and iPad automatically on PRs


amrox:
Once this is merged, the jenkins command should be updated to `verify_pull_request`

fastlane/Fastfile
```diff
@@ -54,6 +54,16 @@ platform :ios do
...
   end
 
+  desc "Runs tests on select platforms for verifying pull requests"
+  lane :verify_pull_request do
+    verify(
+     sim: 'iPhone 6s (9.3)'
+    )
+    verify(
+     sim: 'iPad Air 2 (9.3)'
+    )
```
joewalsh: These should be 10.2, right?

amrox: I guess thats debatable? What's more valuable to test, the latest OS or the oldest supported?

Or if we can't test 9.3 on the Jenkins box then I guess 10.3 it is

joewalsh: Ideally both, so we could do iPad 9.3 and iPhone 10.2

amrox:
@joewalsh I updated it to be iPhone (10.2), iPad Air 2 (9.3)

Is that what you were suggesting?

It would be trivial to add each combination of (iPhone, iPad) (9.3, 10.2) but that would double the build time...

joewalsh:
@amrox that was my suggestion, looks good now

PR #1195, opened 2017-02-23
Allow longer text snippets in Share-A-Fact cards

julienbodet:
Fix this bug: https://phabricator.wikimedia.org/T156078

Currently the property numberOfLines is set to 5 but the minimum font size is set to 38 in the storyboard, so it never displays up to 5 (max 4 in my tests) lines as 38 is too high.

To fix this, I set the numberOfLines to 0 to use as many lines as needed (https://developer.apple.com/reference/uikit/uilabel/1620539-numberoflines) and the minimum font size to 24. So, when the text goes longer, the font size is reduced and more lines are used. The text is truncated when minimum font size of 24 is reached and there is no more space to add new lines.

Also I deleted some code that sets some label properties for iOS 7 and lower. I think it's good as the required version is 9.3.


joewalsh:
Looks great, thank you!

joewalsh:
ok to test

joewalsh:
ok to test

joewalsh:
ok to test

joewalsh:
retest this please

joewalsh:
ok to test

PR #1201, opened 2017-02-24
WMFDatabaseHouseKeeper Cleanups

amrox:


Wikipedia/Code/WMFDatabaseHouseKeeper.swift
```diff
@@ -0,0 +1,97 @@
...
+                    break
+                    
+                default:
+                    assertionFailure("Unknown Content Type")
+                }
+            }
+        }
+        
+        let articlesToDeleteFetchRequest = WMFArticle.fetchRequest()
+        var articlesToDeletePredicate = NSPredicate(format: "viewedDate == NULL && savedDate == NULL && placesSortOrder == NULL && isExcludedFromFeed == %@", NSNumber(booleanLiteral: true))
```
amrox: @joewalsh could you explain this predicate, especially `viewedDate == NULL`. Intuitively it seems like articles that were viewed would never be cleaned up

joewalsh: A `viewedDate` of null indicates that the article was never viewed, and is only a preview. `savedDate` of null indicates the article isn't saved. Saved and history are populated by looking for articles with viewed dates and saved dates. A non-null `placesSortOrder` indicates that the article is currently visible on the places map. `isExcludedFromFeed` here is incorrect - it should check for items not excluded from the feed. Items with `isExcludedFromFeed == YES` need to stay in the database so that they will continue to be excluded from the feed. Previous code: ```NSPredicate *articlesToDeletePredicate = [NSPredicate predicateWithFormat:@"viewedDate == NULL && savedDate == NULL && placesSortOrder == NULL && isExcludedFromFeed == %@", @(NO)];```

amrox: oops good catch, I flipped the boolean somewhere

Wikipedia/Code/WMFLegacyDatabaseHouseKeeper.h
```diff
@@ -1,6 +1,6 @@
 #import <Foundation/Foundation.h>
 
-@interface WMFDatabaseHouseKeeper : NSObject
+@interface WMFLegacyDatabaseHouseKeeper : NSObject
```
joewalsh: can the legacy housekeeper be deleted now?

amrox: Yes. Was leaving it around for reference because the code is non-trivial and they're aren't any tests to verify the new behavior. I did hand-verify it though.

Wikipedia/Code/WMFDatabaseHouseKeeper.swift
```diff
@@ -0,0 +1,113 @@
...
+    
+    private func deleteStaleUnreferencedArticles(_ moc: NSManagedObjectContext) throws {
+        
+        /**
+ 
+        Find `WMFContentGroup`s more than 30 days old.
+ 
+        */
+        
+        let midnightTodayUTC = (Date() as NSDate).wmf_midnightUTCDateFromLocal!
```
joewalsh: I'd rather see this as a `guard let midnightTodayUTC = ...` so that it won't crash if for whatever reason this fails. Would be a good idea to add an assertion inside the guard block so that it still would crash on debug builds

Wikipedia/Code/WMFDatabaseHouseKeeper.swift
```diff
@@ -0,0 +1,113 @@
...
+        let utcCalendar = NSCalendar.wmf_utcGregorian() as Calendar
+        let thirtyDaysAgoMidnightUTC = utcCalendar.date(byAdding: .day, value: -30, to: midnightTodayUTC, wrappingComponents: true)
```
joewalsh: Is `wrappingComponents: true` the better way to do this instead of `NSCalendarMatchStrictly`?

Wikipedia/Code/WMFDatabaseHouseKeeper.swift
```diff
@@ -0,0 +1,113 @@
...
+            if group.midnightUTCDate?.compare(thirtyDaysAgoMidnightUTC!) == .orderedAscending {
+                moc.delete(group)
+                continue
+            }
+            
+            if let articleURLDatabaseKey = (group.articleURL as NSURL?)?.wmf_articleDatabaseKey {
+                referencedArticleKeys.insert(articleURLDatabaseKey)
+            }
+            
+            guard let content = group.content as Array! else {
```
joewalsh: Shouldn't this be `as? Array` to avoid a crash in the case where content is nil or not an array?

Wikipedia/Code/WMFDatabaseHouseKeeper.swift
```diff
@@ -0,0 +1,113 @@
...
+                    referencedArticleKeys.insert(key)
+                    
+                case (.topReadPreview, let preview as WMFFeedTopReadArticlePreview):
+                    guard let key = (preview.articleURL as NSURL).wmf_articleDatabaseKey else {
+                        continue
+                    }
+                    referencedArticleKeys.insert(key)
+                    
+                case (.story, let story as WMFFeedNewsStory):
+                    if let articlePreviews = story.articlePreviews {
```
joewalsh: this could be a `guard` to limit nesting

Wikipedia/Code/WMFDatabaseHouseKeeper.swift
amrox: This line was a little weird to translate from ObjC. Actually I think we're ok just unwrapping the optional instead of the explicit cast because it's already defined as `[NSCoding]?`

PR #1202, opened 2017-02-24
Nearby feed logic changes & Places updates

joewalsh:
- Update the movement limit that triggers a nearby feed update from 1km to 25km
- Force a feed update every 10 days for users who haven't used the Places tab (use the lower distance limit of 1km as well)
- Limit places search history to 15 items
- Bundle update


joewalsh:
jenkins, test this please

PR #1205, opened 2017-02-25
Minimize keyboard when tapping the 'Search' button in Find In Page

julienbodet:
This PR resolves this bug: https://phabricator.wikimedia.org/T158457


joewalsh:
jenkinks, ok to test

PR #1206, opened 2017-02-27
Animate article's lead image for GIFs

julienbodet:
Resolves this bug: https://phabricator.wikimedia.org/T153905


Wikipedia/Code/UIImageView+WMFImageFetchingInternal.m
```diff
@@ -187,6 +190,14 @@ - (void)wmf_setImage:(UIImage *)image
     };
     
     self.image = image;
+    
+    if (image.isGIF && [self isKindOfClass:[FLAnimatedImageView class]]) {
+        NSURL *imageURL = [self wmf_imageURLToFetch];
+        FLAnimatedImage *animatedImage = [FLAnimatedImage animatedImageWithGIFData: [[self wmf_imageController] diskDataForImageWithURL:imageURL]];
```
joewalsh: Since `diskDataForImageWithURL:` could return nil (even though it should exist in pretty much every case) this should be nil checked. Also, I'm not sure if `animatedImageWithGIFData` will return nil if it fails, so it would probably good to check that for nil as well. If either fails, leave the original image.

Wikipedia/Code/UIImageView+WMFImageFetchingInternal.m
```diff
@@ -187,6 +190,14 @@ - (void)wmf_setImage:(UIImage *)image
...
+        FLAnimatedImageView *animatedImageView = ((FLAnimatedImageView*) self);
```
joewalsh: There's a potential that self isn't actually an `FLAnimatedImageView` - this should be verified with `if ([self isKindOfClass:[FLAnimatedImageView class]])`

Wikipedia/Code/UIImageView+WMFImageFetchingInternal.m
joewalsh: Actually, instead you could add the image data to the `WMFImageDownload` returned by `fetchImageWithURL:...` in `wmf_fetchImageDetectFaces:...` - then pass that data to this method to use rather than re-querying the image manager for `diskDataForImageWithURL`. It should still be nil-checked. Passing the data in this way also avoids a potentially costly disk read on the main thread.

Wikipedia/Code/UIImageView+WMFImageFetchingInternal.m
```diff
@@ -109,7 +109,8 @@ - (void)wmf_fetchImageDetectFaces:(BOOL)detectFaces onGPU:(BOOL)onGPU failure:(W
 
     UIImage *cachedImage = [self wmf_cachedImage];
     if (cachedImage) {
-        [self wmf_setImage:cachedImage detectFaces:detectFaces onGPU:onGPU animated:NO failure:failure success:success];
+        NSData *cachedImageData = cachedImage.isGIF ? [[self wmf_imageController] diskDataForImageWithURL:imageURL] : nil;
```
joewalsh: This whole cached image check can be removed (lines 110-115) - the image load function below checks the cache as well.

julienbodet:
Hi @joewalsh is it good now or do you need more changes?

joewalsh:
@julienbodet I'll take a look. Thanks again for submitting this fix!

joewalsh:
jenkins ok to test

joewalsh:
jenkins test this please

joewalsh:
Merging and will update visual tests if necessary.

PR #1209, opened 2017-02-27
Capping image height to 1.0 (100%)

amrox:
Values greater cause the image to be stretched

Also added a test case

https://phabricator.wikimedia.org/T159167


Wikipedia/Code/UIImageView+WMFContentOffset.m
```diff
@@ -38,7 +38,7 @@ - (void)wmf_resetContentsRect {
 }
 
 - (void)wmf_topAlignContentsRect:(UIImage *)image {
-    self.layer.contentsRect = CGRectMake(0, 0, 1, (image.size.width / image.size.height) / 2.f);
+    self.layer.contentsRect = CGRectMake(0, 0, 1, MIN(image.size.width / image.size.height / 2.f, 1.0f));
```
joewalsh: I meant to mention this before - but i'm not sure if the f suffix is still required - with 64 bit phones aren't they all doubles anyway?

PR #1216, opened 2017-02-28
Updates to auth interface styling

montehurd:
https://phabricator.wikimedia.org/T158142


montehurd:
@joewalsh @amrox 
This covers the bulk of the restyling. Still need to finish the alert changes, but was thinking about doing those bits in a separate PR... this one's pretty big already.

Oh the failing test doesn't appear to be related to my changes:
> (([defaults wmf_inTheNewsMostRecentDateNotificationCount] == 3) is true) failed

Wikipedia/Code/UILabel+WMFAuth.swift
```diff
@@ -0,0 +1,13 @@
+
+extension UILabel {
+    /// Makes an attributed string from two strings performing a **$1** substitution and applying a common color scheme (black/blue) used by a few auth view controllers.
+    /// Also re-uses the label's font so the resulting attributed string respects storyboard dynamic type font choice and responds to dynamic type size changes.
+    ///
+    /// - Parameters:
+    ///   - dollarSignString: A localization string which has a **$1** subtitution for the string from the **substitutionString** parameter.
+    ///   - substitutionString: The string which will replace **$1** in the **dollarSignString** parameter.
+    /// - Returns: Attributed string from the two parameter strings, with the **dollarSignString** in black and the **substitutionString** in blue.
+    func wmf_authAttributedStringReusingFont(withDollarSignString dollarSignString: String, substitutionString: String) -> NSAttributedString {
```
amrox: Can you add an example? I read through the doc several times but my brain still isn't getting it :)

montehurd: @amrox 

Sure! It's used in 3 places:

untitled mov

For the login place it says "Don't have an account? Join Wikipedia.", which comes from:
> "login-no-account" = "Don't have an account? $1";
> "login-join-wikipedia" = "Join Wikipedia.";

I did these as 2 separate strings so those parts could be styled independently.

One other thing to keep in mind is we generally avoid simply appending strings which are to be combined in some way, as in this case. "$1" is used instead so translators can choose a language appropriate combination.

montehurd: Oh and it's a UILabel extension so it can re-use the label's _font_ so the dynamic type auto-adjustment still happens with the attributed string...

[NSForegroundColorAttributeName : UIColor.black, NSFontAttributeName : **_font_**]

^ hmm maybe in this case I should have left "self" for clarity... **_self.font_**...

montehurd: @amrox I'm totally open ideas of ways to clarify this too! Or alternative approaches! :)

PR #1218, opened 2017-03-01
remove www from the workspace

joewalsh:
- Xcode isn’t a very good editor for html, js, css
- Having this folder in the project significantly slows down whole
project searches and introduces bogus results
- It’s not necessary to build and run the app


amrox:
👍👍👍👍👍👍

PR #1219, opened 2017-03-01
If the app is launched via a universal link, delay feed load until after article load

joewalsh:


Wikipedia/Code/WMFAppViewController.m
```diff
@@ -425,64 +420,75 @@ - (void)resumeApp {
...
-    NSDate *now = [NSDate date];
-
-    BOOL locationAuthorized = [WMFLocationManager isAuthorized];
     
-    if (!feedRefreshDate || [now timeIntervalSinceDate:feedRefreshDate] > WMFTimeBeforeRefreshingExploreFeed || [[NSCalendar wmf_gregorianCalendar] wmf_daysFromDate:feedRefreshDate toDate:now] > 0) {
-        [self updateFeedSourcesWithCompletion:NULL];
-    } else if (locationAuthorized != [defaults wmf_locationAuthorized]) {
-        [self.exploreViewController updateNearby:NULL];
-    }
+    dispatch_block_t done = ^{
```
amrox: Did you consider pulling all the code in the `done` block into a separate method? Honest question. I'm actually not sure which is more clear and readable.

joewalsh: I didn't consider this - the separate method is probably more clear and readable - will update.

PR #1229, opened 2017-03-02
[Experimental] Including Carthage sources in the project

amrox:
So the are searchable and easy to add breakpoints to


joewalsh:
I kinda enjoyed having these as a black box, but I'd be fine with them in the project (especially since some were just incorporated anyway to limit the number of frameworks).

amrox:
@joewalsh yeah the black box is appealing, but it's made debugging various image stuff in SDImage and NYTPhoto so much easier...

And it's just a folder reference. We can remove it again if it's problematic.

joewalsh:
screen shot 2017-03-02 at 4 16 33 pm

joewalsh:
^ There are also now duplicate app delegates and a lot of extra search results. I'd vote to take this out and only add it to the project on a per-case basis while debugging something that involves an external framework.

PR #1236, opened 2017-03-03
Don't copy the assets folder

joewalsh:
This slows down launch, potentially crashes the widget, and is no longer necessary since we don’t write to the assets folder


amrox:
jenkins test this please

montehurd:
Hmm let me test this a bit before merging...

montehurd:
Per @joewalsh the proxy server is already accessing these bundled assets. (Pre-proxy the webview needed them to be in the user's documents directory.) The grunt process for editing "www" content to produce changes in "assets" remains the same. i.e. never edit assets files directly, edit "www" and run 'make grunt' to rebuild "assets".

PR #1237, opened 2017-03-03
Separate target for Alpha, new build config for Beta Cluster

joewalsh:


amrox:
Why is the nomenclature "Alpha/BetaCluster", not "Alpha/Beta" or "AlphaCluster/BetaCluster"?

joewalsh:
Alpha points to the production servers, Beta Cluster points to the beta server cluster

PR #1238, opened 2017-03-03
fix article removal from disk cache

joewalsh:


Wikipedia/Code/WMFDatabaseHouseKeeper.swift
```diff
@@ -109,23 +111,31 @@ import Foundation
         */
         
         let articlesToDeleteFetchRequest = WMFArticle.fetchRequest()
-        var articlesToDeletePredicate = NSPredicate(format: "viewedDate == NULL && savedDate == NULL && placesSortOrder == NULL && isExcludedFromFeed == %@", NSNumber(booleanLiteral: false))
+        var articlesToDeletePredicate = NSPredicate(format: "viewedDate == NULL && savedDate == NULL && placesSortOrder == 0 && isExcludedFromFeed == FALSE")
```
amrox: You're sure that `FALSE` works? I did it the only way due to the way it was in ObjC and googling around, but that's obviously much simpler if it has the same effect

Wikipedia/Code/WMFDatabaseHouseKeeper.swift
```diff
@@ -2,9 +2,11 @@ import Foundation
 
 @objc class WMFDatabaseHouseKeeper : NSObject {
     
-    @objc func performHouseKeepingOnManagedObjectContext(_ moc: NSManagedObjectContext)  throws {
+    // Returns deleted URLs
+    @objc func performHouseKeepingOnManagedObjectContext(_ moc: NSManagedObjectContext) throws -> [URL] {
```
amrox: Just a note for future - if this object grows any further I'd like to refactor it to use a command pattern (or something).

Wikipedia/Code/WMFDatabaseHouseKeeper.swift
joewalsh: Yeah, I wasn't sure either - I tested it when I made the change, but will re-test to verify. I also read up on here: https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/Predicates/Articles/pSyntax.html

joewalsh:
Adding WIP pending addition of migration that removes articles that should have been deleted previously.

Wikipedia/Code/MWKDataStore.m
```diff
@@ -988,7 +987,59 @@ - (NSArray *)imageInfoForArticleWithURL:(NSURL *)url {
...
+    NSError *fetchError = nil;
+    NSArray *arrayOfAllValidArticles = [self.viewContext executeFetchRequest:allValidArticleKeysFetchRequest error:&fetchError];
+
+    if (fetchError) {
+        failure(fetchError);
+        return;
+    }
+
+    if (arrayOfAllValidArticles.count == 0) {
+        [[NSFileManager defaultManager] removeItemAtPath:[self pathForSites] error:nil];
```
amrox: Is this just an optimization, or does removing the whole sites dir have some added effect?

joewalsh: Just an optimization, rather than adding them all to the to-be-deleted list

joewalsh: Might be good to move this onto the background queue though in case it hangs long enough to cause an `8badf00d`

joewalsh:
test this please

PR #1239, opened 2017-03-03
Auth alert updates.

montehurd:
More from:
https://phabricator.wikimedia.org/T158142

The first commit was supposed to be part of the first auth styling PR but I forgot to push it.

This PR will hopefully address the auth alert changes from the phab ticket, but it will have some pixel-pushing and other tweaks too.


montehurd:
@joewalsh @amrox 

This PR should be reviewable. I'll still need to do some pixel pushing with Carolyn when she gets back, but probably as a follow-up PR if that's ok.

montehurd:
Looks like the failing test is that "in the news" one again...

> (([defaults wmf_inTheNewsMostRecentDateNotificationCount] == 3) is true) failed

joewalsh:
^ Should have separated those out into a separate PR, but the investigation and fix kind of snowballed. TL;DR - tests were broken when notifications were scheduled for the next day instead of the current day. There was also a bug with next day notifications where the notification count for the current day would increment to 4 rather than setting it to one on the next day.

amrox:
> notification logic should probably be updated to take a "current time" so we can test all of the cases without changing the system time.

Or if you want to go the super DI way, you create something like a TimeProvider. In the app it just provides the real time, but in tests you can mock it out to provide whatever time you like.

WikipediaUnitTests/Code/WMFNotificationTests.m
```diff
@@ -21,6 +23,9 @@ @implementation WMFNotificationTests
 
 - (void)setUp {
     [super setUp];
+
+    [[NSUserDefaults wmf_userDefaults] wmf_resetToDefaultValues];
```
amrox: This is more of a tech-debt thing, but it would be better if tests didn't rely on user defaults (or any shared state).

WMF Framework/WMFCaptchaViewController.swift
```diff
@@ -4,6 +4,23 @@ import UIKit
 // Presently it is assumed this view controller will be used only as a
 // child view controller of another view controller.
 
+extension UIView {
+    fileprivate func wmf_subviews(hide: Bool) {
```
joewalsh: naming here is a little confusing - could it be a `var wmf_subviewsHidden: Bool` with set & get implementations?

WMF Framework/WMFCaptchaViewController.swift
```diff
@@ -4,6 +4,23 @@ import UIKit
...
+        isHidden = hide
+        for subview in self.subviews {
+            subview.wmf_subviews(hide:hide)
+        }
+    }
+}
+
+extension UIStackView {
+    fileprivate func wmf_arrangedSubviews(hide: Bool) {
```
joewalsh: same question as above

montehurd: checking...

WMF Framework/WMFCaptchaViewController.swift
montehurd: hmm unsure what the set/get would add... presently this recursively hides subviews...

WMF Framework/WMFCaptchaViewController.swift
montehurd: I deleted this one - forgot it wasn't needed.

WMF Framework/WMFCaptchaViewController.swift
joewalsh: It just would have clearer naming -  instead of `self.wmf_subviews(hide: true)` it could read more clearly as `self.wmf_areSubviewsHidden = true` or `self.wmf_hideSubviews(true)`

montehurd: Ah, I see thx. Lemme tweak that...

joewalsh:
```
    fileprivate var wmf_areSubviewsHidden: Bool {
        set {
            for subview in self.subviews {
                subview.wmf_areSubviewsHidden = newValue
            }
        }
        get {
            return subviews.first?.isHidden ?? false
        }
    }
```

PR #1256, opened 2017-03-07
Update places map annotations

joewalsh:


Wikipedia/Code/ArticlePlaceView.swift
```diff
@@ -3,14 +3,20 @@ import MapKit
 import WMF
 
 class ArticlePlaceView: MKAnnotationView {
-    private let imageView: UIImageView
-    private let selectedImageView: UIImageView
+    private let imageView: UIView
```
montehurd: Just curious... why doesn't ArticlePlaceView use IB? No worries and not a blocker, just wondering.

joewalsh: Originally it was simple enough to not need IB, now, given all of the overlapping views and hierarchies, it's easier to not use IB

PR #1259, opened 2017-03-07
only select the article after searching if it was a search suggestion

joewalsh:


amrox:
Just deletions really? Nice

PR #1260, opened 2017-03-07
Add Get Directions or Open in Apple Maps actions to share sheet when article has location

amrox:
https://phabricator.wikimedia.org/T159784

 - [x] Verify test coverage
- [x] Add placeholder images


Wikipedia/Code/WMFArticleViewController.m
```diff
@@ -98,8 +98,16 @@ - (nullable UIActivityViewController *)sharingActivityViewControllerWithTextSnip
     if (url) {
         [items addObject:[[NSURL alloc] initWithString:[NSString stringWithFormat:@"%@?%@", url.absoluteString, @"wprov=sfsi1"]]];
     }
+    
+    if (CLLocationCoordinate2DIsValid(self.coordinate)) {
```
joewalsh: Would probably be safer to create and assign the MKMapItem to a var, then check that it != nil before adding it to the array.

Wikipedia/Code/WMFMapsActivity.swift
```diff
@@ -0,0 +1,94 @@
...
+    override open var activityImage: UIImage? {
+        get {
+            return UIImage(named: "places-map")
+        }
+    }
+    
+    override func perform() {
+        
+        guard let mapItem = self.mapItem else {
+            assertionFailure("MapItem should have been set")
```
joewalsh: should the activity also finish in this case?

Wikipedia/Code/WMFArticleViewController.m
amrox: Ah good point. Originally it was all coordinate-based. Then I switched it to be MapItem based and didn't change this condition

Wikipedia/Code/WMFMapsActivity.swift
amrox: Yeah

Wikipedia/Code/WMFArticleViewController.m
```diff
@@ -99,7 +99,15 @@ - (nullable UIActivityViewController *)sharingActivityViewControllerWithTextSnip
         [items addObject:[[NSURL alloc] initWithString:[NSString stringWithFormat:@"%@?%@", url.absoluteString, @"wprov=sfsi1"]]];
     }
 
-    UIActivityViewController *vc = [[UIActivityViewController alloc] initWithActivityItems:items applicationActivities:@[[[TUSafariActivity alloc] init]]];
+    if ([self mapItem]) {
```
joewalsh: This might be overkill, but I usually do something along the lines of 
```
MKMapItem *mapItem = [self mapItem];
if (mapItem) {
    [items addObject:mapItem];
}
```
In case there's every anything funky about the getter that returns nil even after it just didn't return nil

PR #1262, opened 2017-03-08
Let two factor login handle 16 digit backup codes

montehurd:
https://phabricator.wikimedia.org/T158241


Wikipedia/Code/WMFTwoFactorPasswordViewController.swift
```diff
@@ -17,18 +19,51 @@ class WMFTwoFactorPasswordViewController: WMFScrollViewController, UITextFieldDe
...
     public var captchaWord:String?
     
+    func backupOathTokenToggleTapped(_ recognizer: UITapGestureRecognizer) {
+        guard recognizer.state == .ended else {
+            return
+        }
+        useBackupOathToken = !useBackupOathToken
+    }
+    
+    fileprivate var useBackupOathToken: Bool = false {
```
amrox: Is Oath supposed to be Oauth?

montehurd: Oh, it's Oath. I asked Gergo about the difference a couple weeks ago but I don't recall how he explained it. Check out the url parameters in the example at the very bottom of this page: 
https://www.mediawiki.org/wiki/API:Login#The_clientlogin_action

montehurd: I think it was this:
> OAuth is a service that is complementary to and distinct from OpenID. OAuth is also distinct from OATH, which is a reference architecture for authentication, not a standard for authorization.

https://en.wikipedia.org/wiki/OAuth

montehurd:
I think the only thing I'm missing is the inline alert, but if it's ok I'll add that with the results from my upcoming pixel pushing session with Carolyn.

Wikipedia.xcodeproj/project.pbxproj
```diff
@@ -141,6 +141,8 @@
 		B01662B51D1B8E3B006F4544 /* NSURL+WMFProxyServerTests.m in Sources */ = {isa = PBXBuildFile; fileRef = B01662B41D1B8E3B006F4544 /* NSURL+WMFProxyServerTests.m */; };
 		B016DCEB1D497C74005BABA1 /* WMFFindInPageKeyboardBar.xib in Resources */ = {isa = PBXBuildFile; fileRef = B016DCEA1D497C74005BABA1 /* WMFFindInPageKeyboardBar.xib */; };
 		B017BF951E68EE3E009F8FAF /* UILabel+WMFAuth.swift in Sources */ = {isa = PBXBuildFile; fileRef = B017BF941E68EE3E009F8FAF /* UILabel+WMFAuth.swift */; };
+		B01CFC5F1E70B00100B3546A /* WMFDeleteBackwardReportingTextField.swift in Sources */ = {isa = PBXBuildFile; fileRef = B01CFC5E1E70B00100B3546A /* WMFDeleteBackwardReportingTextField.swift */; };
+		B01CFC611E71069000B3546A /* String?+WMFExtras.swift in Sources */ = {isa = PBXBuildFile; fileRef = B01CFC601E71069000B3546A /* String?+WMFExtras.swift */; };
```
joewalsh: Are these both added to the Alpha target as well?

joewalsh: nope screen shot 2017-03-09 at 9 39 23 am

joewalsh:
test this please

PR #1263, opened 2017-03-08
Find In Page: stop scrolling to let the keyboard open

julienbodet:
Resolves the scroll bug describe here: https://phabricator.wikimedia.org/T143024

This fix stop the view from scrolling when Find in page icon or textfield clear button is touched to let the keyboard open.

I found how to stop scrolling on this thread: http://stackoverflow.com/questions/3410777/how-can-i-programmatically-force-stop-scrolling-in-a-uiscrollview


montehurd:
Oh interesting! Another place we could put that method would maybe be as a category method in the `UIScrollView+WMFContentOffsetUtils.m` file, then we could use it for any scroll view.

joewalsh:
ok to test

joewalsh:
Looks good, thanks @julienbodet! @montehurd the category can be addressed as a follow-on.

PR #1265, opened 2017-03-08
Distance squished on Nearby article with two line header and two line subhead on Explore feed Nearby card

amrox:
https://phabricator.wikimedia.org/T154035


amrox:
test this please

PR #1269, opened 2017-03-09
Fixing a bug where all articles would have View on a map even if they

amrox:
didn't have a location

### DIFF
--- a/Wikipedia/Code/WMFArticleViewController.m
+++ b/Wikipedia/Code/WMFArticleViewController.m
@@ -1762,8 +1762,9 @@ static const CGFloat WMFArticleViewControllerTableOfContentsSectionUpdateScrollD
                                      }
                                  }];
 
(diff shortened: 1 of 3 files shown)


Wikipedia/Code/WMFArticleViewController.m
```diff
@@ -1762,8 +1762,9 @@ - (void)commitViewController:(UIViewController *)viewControllerToCommit {
                                      }
                                  }];
 
+    WMFArticle *wmfarticle = [self.previewStore itemForURL:self.articleURL];
```
joewalsh: I'd suggest dataStore `fetchArticleForURL` (even though it's poorly named) since it caches the results.

joewalsh: lol nevermind

```
- (nullable WMFArticle *)itemForURL:(NSURL *)url {
    return [self.dataStore fetchArticleForURL:url];
}
```

PR #1270, opened 2017-03-09
Piwik logging for places, update footer icon for "show on map"

joewalsh:


montehurd:
Hmm may want to update the title :)

PR #1272, opened 2017-03-10
Follow-up auth tweaks

montehurd:
https://phabricator.wikimedia.org/T160133

### DIFF
--- a/Wikipedia/Localizations/en.lproj/Localizable.strings
+++ b/Wikipedia/Localizations/en.lproj/Localizable.strings
@@ -68,8 +68,8 @@
 "two-factor-login-title" = "Log in to your account";
 "two-factor-login-instructions" = "Please enter two factor verification code";
 "two-factor-login-continue" = "Continue log in";
-"two-factor-login-with-regular-code" = "Use short numeric code";
-"two-factor-login-with-backup-code" = "Use longer backup code";
+"two-factor-login-with-regular-code" = "Use verification code";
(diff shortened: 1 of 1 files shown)


Wikipedia/Code/WMFAuthLinkLabel.swift
```diff
@@ -7,8 +15,27 @@ class WMFAuthLinkLabel: UILabel {
...
+
+        guard let strings = strings else {
+            // Not combining two strings.
+            font = boldSubheadlineFont
+            return
+        }
+
+        // Combine and style 'dollarSignString' and 'substitutionString'.
+        // https://github.com/wikimedia/wikipedia-ios/pull/1216#discussion_r104224511
+        let subheadlineFont = UIFont.wmf_preferredFontForFontFamily(WMFFontFamily.system, withTextStyle: .subheadline, compatibleWithTraitCollection: self.traitCollection)!
```
montehurd: @joewalsh do you think we can make "wmf_preferredFontForFontFamily" return font instead of an optional font?

I'll fix the "!" here, but was just curious.

montehurd:
@joewalsh @amrox Should be reviewable :)

Wikipedia/Code/WMFAuthLinkLabel.swift
montehurd: We left the fonts optional and I fixed the forced unwrapping here:
https://github.com/wikimedia/wikipedia-ios/pull/1285

PR #1276, opened 2017-03-10
Make UIColor+WMFStyle colors consistent.

montehurd:
I was having a hard time knowing if we already had certain colors, and Carolyn said hex would be easiest when we later update the app colors to use wmf style guide colors.


montehurd:
test this please

PR #1278, opened 2017-03-11
Display article quotebox in the normal HTML flow

julienbodet:
https://phabricator.wikimedia.org/T156977

The `www/.node-version` file contains `v0.10`, but `nodenv` is not able to find that exact version. 
So I installed the last 0.10 version which is 0.10.48 and I changed to `v0.10.48` in the `.node-version` file to run `npm install`.

Maybe `.node-version` should be updated?


montehurd:
Seems reasonable. I'll test this a bit, but may not get to it today.

montehurd:
Good catch about `.node-version` too... I'll try updating that too...

joewalsh:
jenkins ok to test

montehurd:
@julienbodet Hey I think this PR https://github.com/wikimedia/wikipedia-ios/pull/1289 should bring the `.node-version` up to date. Thanks for the recommendation!

julienbodet:
@montehurd Thanks for updating it! But I still can't get `npm install` work to generate the `node_modules` folder.

Node version `6.10.0` is available via `nodenv`, but not the `6.10`. So when I run `npm install` I get this error:
`nodenv: version v6.10 is not installed (set by ~/wikipedia-ios/www/.node-version)`

I also tried `make grunt` as you mention in #1289 and I got the same error. Maybe I missed something. Please, do you have any suggestion?

To get it work I changed to `v6.10.0` in .node-version and `npm install` or `make grunt` works fine.

PR #1282, opened 2017-03-13
Updated place group interaction

joewalsh:
This updates the places group interaction, but also has a number of other fixes. Merging this doesn't commit to the new interaction, it's simple to return to the old style.


joewalsh:
jenkins test this please

joewalsh:
jenkins test this please

Wikipedia/Code/ArticlePlaceGroupViewController.swift
```diff
@@ -0,0 +1,332 @@
...
+        tintView.image = image
+        UIGraphicsEndImageContext()
+    }
+    
+    func show(center: CGPoint) {
+        self.center = center
+       
+        drawGradient(center: center)
+        
+        let rotationTransform = CGAffineTransform(rotationAngle: CGFloat(-0.99*M_PI))
```
amrox: Curious why the `-0.99` and `0.99` below

montehurd: Same.

joewalsh: To force it to rotate in a certain direction. Exactly M_PI and negative M_PI don't go the same way, this way it forces clockwise out and counter-clockwise back in.

montehurd: Ah, maybe something like `let forcedClockwiseAngle = 0.99` or some such to make that explicit?

PR #1293, opened 2017-03-16
Hide the places feature

joewalsh:


montehurd:
lol "montehurd self-requested a review 16 seconds ago" - i hit the wrong button! :)

PR #1295, opened 2017-03-16
Add "applib" npm package (this is where the consolidated js bits live for now)

montehurd:
More bits related to https://phabricator.wikimedia.org/T158686

The "applib" repo at https://github.com/wikimedia/applib.git
contains common js bits used by both the iOS and Android apps.


joewalsh:
It looks like the package wasn't actually installed? Shouldn't there be changes to the js in assets and/or www?

montehurd:
@joewalsh 

> It looks like the package wasn't actually installed? Shouldn't there be changes to the js in assets and/or www?

This just adds the package. I split things out - I had too much going on in the former PR :)

PR #1298, opened 2017-03-17
Test to ensure number of substitions matches in source and description strings

amrox:
Please verify the new text added to qqq.lproj


Wikipedia/Localizations/qqq.lproj/Localizable.strings
```diff
@@ -136,7 +136,7 @@
 "wikitext-upload-save-anonymously" = "Button text for publishing without logging in";
 "wikitext-upload-save-anonymously-warning" = "Button sub-text informing user or draw-backs of not signing in before saving wikitext. Parameters:\n* $1 - sign in button text";
 "wikitext-upload-save-terms-and-license" = "Button text for information about the Terms of Use and edit license. Parameters:\n* $1 - 'Terms of Use' link ([[Wikimedia:Wikipedia-ios-wikitext-upload-save-terms-name]])\n* $2 - license name link";
-"wikitext-upload-save-terms-name" = "This message is used as $1 in the message [[Wikimedia:Wikipedia-ios-wikitext-upload-save-terms-and-license]].\n{{Identical|Terms of use}}";
+"wikitext-upload-save-terms-name" = "This message is used in the message [[Wikimedia:Wikipedia-ios-wikitext-upload-save-terms-and-license]].\n{{Identical|Terms of use}}";
```
amrox: @montehurd This one was the only one I struggled with. The `$1` is confusing my very naive algorithm since it refers to a substation is a another string, not this one. I took it out so the test would pass. Thoughts?

montehurd: @amrox This does seem like a case where a mis-match in number of "$" is acceptable... hmm... how about if we do this:

- make the test fail (as it does now) if the number of "$" in the QQQ value doesn't match the number of "$" in the EN value
- but don't fail if the EN value has no "$"

That way we'd enforce "$" being added to QQQ for any EN having "$", but we'd still be able to add "$" to QQQ in cases like the one you found...

amrox: I thought about this solution as well but obviously it doesn't catch the case where a substitution was removed from EN but not qqq. Do you think losing that check matters?

montehurd: Ya that's a tough call... I'm 50/50 about being able to reference other "$" keys in the QQQ, but then, that's the only QQQ value that does this, right, so maybe it's not so valuable...

PR #1304, opened 2017-03-20
Update visual design of search box / search page and filtered search functionality on Places based on user feedback

amrox:
https://phabricator.wikimedia.org/T160577

### Things to look at:
 - **TouchOutsideOverlayView** This is a piece of code that will let you know when you touch *outside* a view (really some rects). It's used to dismiss the filter dropdown on iPad
 - **Search History** It is now saving separate search history for Top and Saved filters. This is because the filters seem like a "mode" and it seemed confusing to list searches from another mode
 - **NSTextAttachment+WMFExtras** I snagged this snippet from stackoverflow to properly rendered the dropdown chevron
 - **Top-level views in Places.storyboard** I have a few views that aren't children of the main view in the storyboard. They're referenced by outlets in the PlacesViewController and added programmatically. Any problems with this?
 - **New `origin` field in PlaceSearch** has values `.user` or `.system` to distinguish between searches initiated by the user or automatic, default searches. `.system` searches don't appear in the search history

### TODO

 - [x] summary count for Top articles 
      - it'd be best probably to just populate when users open the dropdown, we could use the default 
progress bar to show that the count is being fetched
     - https://phabricator.wikimedia.org/T163567
- [x] left align placeholder text in search box (this is non-trivial)
     - https://phabricator.wikimedia.org/T163570
- [x] right side of places list seems tight
    - https://phabricator.wikimedia.org/T163571
- [x] bottom border of the navigation bar is missing when pushing the article vc from places.
- [x] add more spacing under search bar on iPhone
 - [x] double check the my location icon is centered in code
  - [x] darker grey for no search state (#f5f5f5)
 - [x] dark grey instead of blue on close button
 - [x] align right hand side of of X close to right side of toggle
 - [x] header text for no search should be Medium
 - [x] disable drop down on search screen
 - [x] Show help overlay when there are no recent searches
- [x] filter selector dropdown placement on iPad
- [x] Bug where selecting filter dropdown title doesn't always show dropdown
 - [x] ~~disable filter selector dropdown when searching?~~
- [x] ~~search suggestions for Saved places~~


Wikipedia/Localizations/en.lproj/Localizable.strings
```diff
@@ -587,6 +587,7 @@
 "places-search-saved-articles" = "Saved articles";
 "places-search-suggested-searches-header" = "Suggested searches";
 "places-search-recently-searched-header" = "Recently searched";
+"places-search-articles-that-match" = "Articles matching \“$1\”";
```
montehurd: Are those directional quotes (or whatever they're called) or is GitHub just making them italics?

amrox: I think they are directional quotes. Should they not be?

joewalsh:
I'm fine with top level views in the storyboard that get added programatically. All the other changes and additions you called out look good as well. Only small thing I noticed was that the bottom border of the navigation bar is missing when pushing the article vc from places.

amrox:
@joewalsh good catch, I need to reset the navigation bar style when leaving the Places view

ContinueReadingWidget/Info.plist
```diff
@@ -19,7 +19,7 @@
 	<key>CFBundleShortVersionString</key>
 	<string>5.4.1</string>
 	<key>CFBundleVersion</key>
-	<string>0</string>
+	<string>1111</string>
```
joewalsh: these should stay at 0

Wikipedia.xcodeproj/project.pbxproj
```diff
@@ -10961,7 +10987,7 @@
 				CLANG_ENABLE_CODE_COVERAGE = NO;
 				CODE_SIGN_ENTITLEMENTS = Wikipedia/Wikipedia.entitlements;
 				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
-				CURRENT_PROJECT_VERSION = 0;
+				CURRENT_PROJECT_VERSION = 1111;
```
joewalsh: Same comment as above (and for all the build numbers below)

Wikipedia/Code/PlacesViewController.swift
```diff
@@ -790,20 +843,36 @@ class PlacesViewController: UIViewController, MKMapViewDelegate, UISearchBarDele
...
+        updateNavigationBar(removeUnderline: true)
+
+        let searchBarHeight: CGFloat = 32
+        let searchBarLeftPadding: CGFloat = 7.5
+        let searchBarRightPadding: CGFloat = 2.5
+        
         searchBar = titleViewSearchBar
-        navigationController?.setNavigationBarHidden(false, animated: animated)
+        
+        filterSelectorView.frame = CGRect(x: searchBarLeftPadding, y: 0, width: view.bounds.size.width - searchBarLeftPadding - searchBarRightPadding, height: 32)
```
joewalsh: should `32` be `searchBarHeight`?

Wikipedia/Code/PlacesViewController.swift
```diff
@@ -1879,24 +2141,45 @@ class PlacesViewController: UIViewController, MKMapViewDelegate, UISearchBarDele
                 let limit = min(count, searchHistoryCountLimit)
                 recentSearches = try results[0..<limit].map({ (kv) -> PlaceSearch in
                     guard let ps = PlaceSearch(object: kv.value) else {
-                            throw NSError()
+                            throw PlaceSearchError.deserialization(object: kv.value)
```
joewalsh: Thank you for making this a real error

PR #1312, opened 2017-03-24
try to delete & re-add the persistent store if it fails

joewalsh:


amrox:
I would be really useful to have events recorded for the primary and secondary failures, but we still need to talk about analytics.

How do you feel about adding a 

`// TODO: Analytics Events`

as a reminder?

montehurd:
@amrox ya sounds like we'll want to add event logging for this error/issue once we have the ability to log errors by type.

PR #1313, opened 2017-03-24
Pull widenImage transform from applib.

montehurd:
https://phabricator.wikimedia.org/T161019

Image widening JS/CSS was moved to applib:
https://github.com/wikimedia/applib/pull/8


www/.jshintrc
```diff
@@ -7,6 +7,7 @@
 	"nonew": true,
 	"undef": true,
 	"unused": true,
+	"esversion": 6,
```
niedzielski: This seems like a bad idea unless all ES6 features are available in all iOS versions you support. By the byte, long term I hope we can switch to ESLint!

www/package.json
```diff
@@ -4,7 +4,7 @@
   "repository": "https://gerrit.wikimedia.org/r/apps/ios/wikipedia",
   "version": "0.1.0",
   "dependencies": {
-    "applib": "1.0.0"
+    "applib": "1.2.0"
```
niedzielski: :100:

www/js/transforms/widenImages.js
```diff
@@ -1,87 +1,15 @@
...
-        }
-        if(el.style.maxWidth){
-            el.style.maxWidth = '100%';
-        }
-        if(el.style.float){
-            el.style.float = 'none';
-        }
-    }
-}
+const isGalleryImage = function(image) {
```
niedzielski: Do your iOS versions support this ES6 feature or are you transpiling down?

niedzielski:
@montehurd, sorry for the late feedback inline. It's probably all cool but seemed worth mentioning.

PR #1318, opened 2017-04-05
Remove SDWebImage

joewalsh:
These are the underlying changes that would allow us to do https://phabricator.wikimedia.org/T131350. It might also be a fix for https://phabricator.wikimedia.org/T161216

- SDWebImage is replaced with:
  - a wrapper around NSURLCache for temporarily cached images
  - a filesystem cache with a Core Data index for permanently cached images

Also addressed:
https://phabricator.wikimedia.org/T162452 
Since saving a lot of pages to test this issue exposed the explore feed issue referenced there.


joewalsh:
The latest changes add isDownloaded to WMFArticle to track whether or not a given article and all of its images are saved. This allows us to check for articles that need to be cached (or removed from the cache) with core data fetches rather than an iteration over all saved articles.

amrox:
Should images for saved articles be backed up by iCloud?

joewalsh:
@amrox i'd guess so - If I saved a bunch of articles, got a new device, and restored it from backup, I'd expect them to still be viewable offline without having to re-request the images.

joewalsh:
@amrox ultimately probably a product question (cc @jtminor) though

amrox:
Also obligatory 😢 🐼 for removing `WMFImageControllerTests` and not adding any new tests

Wikipedia/Code/WMFNotificationsController.m
```diff
@@ -162,70 +162,69 @@ - (void)sendNotificationWithTitle:(NSString *)title body:(NSString *)body catego
...
+                                                    //                                                            CGFloat originX = MAX(0, faceMidX - halfNormalizedWidth);
+                                                    //                                                            CGFloat normalizedHeight = MAX(faceRect.size.height, WMFNotificationImageCropNormalizedMinDimension);
+                                                    //                                                            CGFloat originY = 0.5 * (1 - normalizedHeight);
+                                                    //                                                            cropRect = CGRectMake(originX, originY, normalizedWidth, normalizedHeight);
+                                                    //                                                        }
+                                                    //                                                    }
+                                                    
+                                                    //Since face cropping is broken, don't attach images with faces
```
joewalsh: @amrox broken specifically in the notifications (not on our end) - the content rects I passed weren't respected by the system notification display. this code wasn't added, just updated to use the new image loading

joewalsh:
jenkins, test this please

amrox:
exciting times

PR #1323, opened 2017-04-07
Ensure the splash screen isn't hidden before migrations complete, ensure the user activity is processed instead of showing the explore feed or the last read article

joewalsh:


Wikipedia/Code/WMFAppViewController.m
```diff
@@ -240,7 +243,7 @@ - (void)configureArticleListController:(WMFArticleListTableViewController *)cont
 - (void)appWillEnterForegroundWithNotification:(NSNotification *)note {
     self.unprocessedUserActivity = nil;
     self.unprocessedShortcutItem = nil;
-    [self resumeApp];
+    [self resumeApp:^{}];
```
amrox: nitpicky, but why just just pass `nil`?

joewalsh: I've been making completions "nonnull" to avoid the nil check, but ultimately end up doing the nil check, so probably could just make them nullable. In this case nil would be fine since the completion is nullable. In Swift I've also been making them non-optional.

PR #1327, opened 2017-04-10
Bump Continue Reading widget title content compression resistance.

montehurd:
https://phabricator.wikimedia.org/T160773


montehurd:
@joewalsh 

Looks like screenshot device was using Turkish. So I switched to Turkish but was still unable to repro with the article from the screenshot. 

However, I was able to fake it by increasing the height of the "Days ago" label:


Also confirmed that bumping the title label's vertical content compression resistance (to 751) seems to have a reasonable effect - i.e. the description sacrifices some height and drops to a single line:


Based on that, I'm leaning towards going with this PR's `751` bump... thoughts?

joewalsh:
Sure

PR #1332, opened 2017-04-12
Fix url encoding when sharing an article title with an exclamation mark

nambatee:


Wikipedia/Code/WMFArticleViewController.m
```diff
@@ -97,7 +97,11 @@ - (nullable UIActivityViewController *)sharingActivityViewControllerWithTextSnip
     NSURL *url = [NSURL wmf_desktopURLForURL:self.url];
 
     if (url) {
-        [items addObject:[[NSURL alloc] initWithString:[NSString stringWithFormat:@"%@?%@", url.absoluteString, @"wprov=sfsi1"]]];
+        NSString *urlAsString = url.absoluteString;
+        if([urlAsString hasSuffix:@"!"]) {
+            urlAsString = [urlAsString stringByReplacingOccurrencesOfString:@"!" withString:@"%21"];
+        }
+        [items addObject:[[NSURL alloc] initWithString:[NSString stringWithFormat:@"%@?%@", urlAsString, @"wprov=sfsi1"]]];
```
joewalsh: Regardless of whether or not the `!` gets percent-encoded, the old code should be updated to use `NSURLComponents` and `NSURLQueryItem` to add `wprov=sfsi1` to the URL ([here's a similar use](https://github.com/wikimedia/wikipedia-ios/blob/7402cc2db9105910da1989c7dd4e0eca50eaad4e/Wikipedia/Code/WMFProxyServer.m#L260)). Also, instead of using a hardcoded `%21`, the built in `stringByAddingPercentEncodingWithAllowedCharacters:` could be used on the `path` property with a character set based on `[NSCharacterSet URLPathAllowedCharacterSet]` with `!` removed. That would then be set to `percentEncodedPath`.

nambatee: I tried using `stringByAddingPercentEncodingWithAllowedCharacters:` on the `path` property with  `URLPathAllowedCharacterSet ` but it didn't encode `!`. I asked on [stack overflow](http://stackoverflow.com/questions/43401456/urlpathallowedcharacterset-not-encoding-exclamation-marks/43414635#43414635) and 1 answer suggested that `URLPathAllowedCharacterSet` did not include `!`.

The way I tried it:

`components.path = [components.path stringByAddingPercentEncodingWithAllowedCharacters:[NSCharacterSet URLPathAllowedCharacterSet]];`

So that part of the code is now using `NSURLComponents ` but the link to the article is still broken on Twitter 😟

Wikipedia/Code/WMFArticleViewController.m
```diff
@@ -95,13 +95,14 @@ - (nullable UIActivityViewController *)sharingActivityViewControllerWithTextSnip
...
-        if([urlAsString hasSuffix:@"!"]) {
-            urlAsString = [urlAsString stringByReplacingOccurrencesOfString:@"!" withString:@"%21"];
-        }
-        [items addObject:[[NSURL alloc] initWithString:[NSString stringWithFormat:@"%@?%@", urlAsString, @"wprov=sfsi1"]]];
+        NSURLComponents *components = [NSURLComponents componentsWithURL:url resolvingAgainstBaseURL:NO];
+        
+        NSURLQueryItem *queryItem = [NSURLQueryItem queryItemWithName: @"wprov" value: @"sfsi1"];
+        components.queryItems = @[queryItem];
+
+        [items addObject:components.URL];
```
joewalsh: `components.URL` could be nil (the property on `NSURLComponents` is declared as `nullable`, also `components` itself could be nil) and adding nil to an array causes a crash. This just needs to be assigned to a variable and nil checked before adding it to the array. Otherwise looks good!

joewalsh:
jenkins, ok to test

Wikipedia/Code/WMFArticleViewController.m
joewalsh: @nambatee the character set would need to be modified to remove the `!`. It'd look like this:
```
        NSMutableCharacterSet *characterSet = [[NSCharacterSet URLPathAllowedCharacterSet] mutableCopy];
        [characterSet removeCharactersInString:@"!"];
```

joewalsh: However, I think since `!` is actually allowed in the path and the bug is on twitter's end, we should merge this PR as is (since there are worthwhile fixes in it) and discuss whether or not we want to add the workaround for the twitter bug as a follow on (still tracked by https://phabricator.wikimedia.org/T100321).

PR #1333, opened 2017-04-12
Coalesced background feed loading

joewalsh:
Feed updates are done on a background managed object context and the UI is updated in a single pass when those changes are merged into the view. `WMFArticleDataStore` and `WMFContentGroupDataStore` were removed and replaced with categories on `NSManagedObjectContext` and the associated object classes. The feed also now automatically loads older content (if less than 30 days old) when the user scrolls past 80% of the existing content.

https://phabricator.wikimedia.org/T162167
https://phabricator.wikimedia.org/T142577


montehurd:
Seeing a good deal more red than green on the diff... I like it.

Wikipedia/Code/WMFExploreViewController.m
```diff
@@ -682,6 +686,14 @@ - (WMFLayoutEstimate)collectionView:(UICollectionView *)collectionView estimated
         } break;
         case WMFFeedDisplayTypePageWithPreview: {
             WMFArticle *article = [self articleForIndexPath:indexPath];
+            NSString *key = article.key;
+            NSString *cacheKey = [NSString stringWithFormat:@"%@-%lli", key, (long long)columnWidth];
+            NSNumber *cachedValue = [self.cachedHeights objectForKey:cacheKey];
+            if (cachedValue) {
+                estimate.height = [cachedValue doubleValue];
```
montehurd: Will this play nice with rotation and Dynamic Type changes?

joewalsh: @montehurd  good catch, I addressed this in the separate PR I opened for this change

coreyfloyd:
Just wanted to say this is awesome… so happy this is going in! 💪 💯

PR #1335, opened 2017-04-13
Find In Page: display string "0 / 0" when no results found

julienbodet:
https://phabricator.wikimedia.org/T161271


montehurd:
Thanks for the patch! 

In this case I'd consider approaching it a little differently. 

In most cases we usually only directly edit these two localization files:
- `en.lproj/Localizable.strings` (the master file translators use)
- `qqq.lproj/Localizable.strings` (the file which explains the strings in the file above)

In this case I'm leaning towards still having "0 / 0" be translated because, even though in most languages I imagine the translation will still be "0 / 0", in a few cases translators may prefer a different " / " separator.

I'd propose leaving the old `find-in-page-no-matches` strings as-is (reverting d4a33c2) and instead adding and using a new string. Maybe call it `find-in-page-zero-of-zero`? You'd only have to add it to the `en` and `qqq` files mentioned above.

Thanks again for the patch! I hope my feedback made sense :)

julienbodet:
@montehurd Thanks for your feedback! It makes sense, but I assumed that the separator in "0 / 0" will be the same in all languages. It's actually the case when matches are found, we always use "/" as a separator no matter what language.

Assuming the separator might be different depending on the language, we can use a string with parameters (`@"%@ / %@"`) when there is results or not.

montehurd:
> Assuming the separator might be different depending on the language, we should maybe use a string with parameters (@"%@ / %@") when there is results or not. What do you think?

@julienbodet That's a really good point! Yeah if you want to switch it to do that I think it makes sense.

julienbodet:
> @julienbodet That's a really good point! Yeah if you want to switch it to do that I think it makes sense.

@montehurd I switched it!

Wikipedia/Localizations/en.lproj/Localizable.strings
```diff
@@ -547,6 +547,7 @@
 "explore-randomizer" = "Randomizer";
 
 "find-in-page-no-matches" = "0 matches";
+"find-in-page-number-matches" = "%lu / %lu";
```
joewalsh: Since we use translatewiki.net for translations, we can't use standard string tokens. `%lu / %lu` should be `$1 / $2`

Wikipedia/Code/WMFFindInPageKeyboardBar.m
```diff
@@ -110,9 +110,9 @@ - (void)updateLabelTextForCurrentMatchIndex:(NSInteger)index matchesCount:(NSUIn
     } else if (count > 0 && index == -1) {
         labelText = [NSString stringWithFormat:@"%lu", (unsigned long)count];
     } else if (count == 0) {
-        labelText = MWLocalizedString(@"find-in-page-no-matches", nil);
+        labelText = [NSString stringWithFormat:MWLocalizedString(@"find-in-page-number-matches", nil), 0, 0];
     } else {
-        labelText = [NSString stringWithFormat:@"%lu / %lu", (unsigned long)(index + 1), (unsigned long)count];
+        labelText = [NSString stringWithFormat:MWLocalizedString(@"find-in-page-number-matches", nil), (unsigned long)(index + 1), (unsigned long)count];
```
joewalsh: Since the localization string needs to change to  ` $1 / $2 `, this also needs to be updated to replace `$1` and `$2` in the string. Another potential (but not fully necessary) enhancement would be to use `[NSNumberFormatter localizedStringFromNumber:@(index + 1) numberStyle:NSNumberFormatterDecimalStyle]` to format the numbers before inserting them into the string (this will ensure that thousands separators are added if necessary).

Wikipedia/Code/WMFFindInPageKeyboardBar.m
```diff
@@ -109,10 +109,13 @@ - (void)updateLabelTextForCurrentMatchIndex:(NSInteger)index matchesCount:(NSUIn
         labelText = nil;
     } else if (count > 0 && index == -1) {
         labelText = [NSString stringWithFormat:@"%lu", (unsigned long)count];
-    } else if (count == 0) {
-        labelText = [NSString stringWithFormat:MWLocalizedString(@"find-in-page-number-matches", nil), 0, 0];
     } else {
```
julienbodet: I also removed this unnecessary `else if` block. The `else` clause catch it, `count` and `index + 1` are equal to 0.

joewalsh:
jenkins, ok to test

PR #1355, opened 2017-04-20
Fixes for notification header visibility, refresh control visibility, background feed update save

joewalsh:


Wikipedia/Code/WMFExploreViewController.m
```diff
@@ -433,6 +433,10 @@ - (void)showNotificationHeader {
     [header.enableNotificationsButton addTarget:self action:@selector(enableNotificationsButtonPressed) forControlEvents:UIControlEventTouchUpInside];
 
     [[NSUserDefaults wmf_userDefaults] wmf_setDidShowNewsNotificationCardInFeed:YES];
+    
+     if (self.collectionView.contentOffset.y <= 0) {
```
amrox: minor, but this non-obvious (to me) condition is in 2 places. Does it make sense to refactor into a method?

PR #1356, opened 2017-04-20
Fix background fetch - Ensure the app doesn't fully launch when launched for a background fetch

joewalsh:


Wikipedia/Code/AppDelegate.m
```diff
@@ -114,6 +113,7 @@ - (void)applicationWillEnterForeground:(UIApplication *)application {
 - (void)applicationDidBecomeActive:(UIApplication *)application {
     // Restart any tasks that were paused (or not yet started) while the application was inactive. If the application was previously in the background, optionally refresh the user interface.
     [[NSUserDefaults wmf_userDefaults] wmf_setAppBecomeActiveDate:[NSDate date]];
+    [self resumeAppIfNecessary];
```
joewalsh: Resume is now always delayed until at least `applicationDidBecomeActive:` - this way the app isn't fully resumed for a background fetch

Wikipedia/Code/WMFAppViewController.m
```diff
@@ -280,9 +280,15 @@ - (void)appLanguageDidChangeWithNotification:(NSNotification *)note {
 #pragma mark - Background Fetch
 
 - (void)performBackgroundFetchWithCompletion:(void (^)(UIBackgroundFetchResult))completion {
-    [self.dataStore.feedContentController updateBackgroundSourcesWithCompletion:^{
-        completion(UIBackgroundFetchResultNewData);
-    }];
+    dispatch_async(dispatch_get_main_queue(), ^{
```
joewalsh: async to give no-op launch migrations a chance to complete

Wikipedia/Code/WMFExploreViewController.m
```diff
@@ -142,12 +143,16 @@ - (void)titleBarButtonPressed {
 
 #pragma mark - Accessors
 
-- (void)setRefreshControl:(UIRefreshControl *)refreshControl {
-    [_refreshControl removeFromSuperview];
-
-    _refreshControl = refreshControl;
+- (UIRefreshControl *)refreshControl {
+    WMFAssertMainThread(@"Refresh control can only be accessed from the main thread");
```
joewalsh: When the feed was refreshed before the view loaded, the refresh control wouldn't be shown. This fixes that issue by ensuring the refresh control & view load in those cases.

PR #1358, opened 2017-04-21
[5.4.1] Consolidate summary logic, Fix Core Data issues

joewalsh:
Same fix as https://github.com/wikimedia/wikipedia-ios/pull/1349 but based on the 5.4.1 branch instead of develop

Also fixes Core Data errors caused by setting `propertiesToFetch` without also setting a dictionary result type.

### DIFF
--- a/WMF Framework/SummaryExtensions.swift
+++ b/WMF Framework/SummaryExtensions.swift
@@ -1,0 +1,114 @@
+extension WMFArticle {
+    public func update(withSummary summary: [String: Any]) {
+        if let originalImage = summary["originalimage"] as? [String: Any],
+            let source = originalImage["source"] as? String,
+            let width = originalImage["width"] as? Int,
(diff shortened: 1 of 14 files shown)


WMF Framework/SummaryExtensions.swift
```diff
@@ -0,0 +1,114 @@
...
+            let coordinate = coordinates.first,
+            let lat = coordinate["lat"] as? Double,
+            let lon = coordinate["lon"] as? Double {
+            self.coordinate = CLLocationCoordinate2D(latitude: lat, longitude: lon)
+        }
+    }
+}
+
+extension URLSession {
+    public func jsonDictionaryTask(with url: URL, completionHandler: @escaping ([String: Any]?, URLResponse?, Error?) -> Swift.Void) -> URLSessionDataTask {
```
amrox: mostly out of curiosity, why `Swift.Void` here?

PR #1366, opened 2017-04-25
[5.4.1] Potential crash fixes, new debugging methods

joewalsh:
- Ensure `feedImportContext` is created and torn down on the main thread
- Save the main context after any `feedImportContext` save (previously would hold changes in memory until saving on app exit)
- Fixes iOS 9 crashes by removing unsupported `automaticallyMergesChangesFromParent` and `dispatch_assert_queue_debug`
- Ensure feed imports on any context are saved & the callback occurs on the main thread
- `debugChaos` for randomly moving, inserting, deleting feed sections

### DIFF
--- a/WMF Framework/WMFExploreFeedContentController.m
+++ b/WMF Framework/WMFExploreFeedContentController.m
@@ -318,7 +318,8 @@ static NSTimeInterval WMFFeedRefreshBackgroundTimeout = 30;
         return;
     }
     self.taskGroup = [WMFTaskGroup new];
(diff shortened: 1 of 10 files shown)


WMF Framework/WMFExploreFeedContentController.m
```diff
@@ -300,4 +312,57 @@ - (void)debugSendRandomInTheNewsNotification {
     });
 }
 
+
+- (void)debugChaos {
```
amrox: lol

PR #1367, opened 2017-04-25
Tweaks for places filters

amrox:
- Top filter always has active icon
 - Saved filter has active icon unless there are no saved places
 - Different message appears if there are no saved places

### DIFF
--- a/Wikipedia/Code/PlaceSearchFilterListController.swift
+++ b/Wikipedia/Code/PlaceSearchFilterListController.swift
@@ -50,21 +50,15 @@ class PlaceSearchFilterListController: UITableViewController {
         if (indexPath.row == 0) {
             myCell.titleLabel.text = localizedStringForKeyFallingBackOnEnglish("places-filter-top-articles")
             myCell.subtitleLabel.text = localizedStringForKeyFallingBackOnEnglish("places-filter-top-articles-count").replacingOccurrences(of: "$1", with: String(delegate.placeSearchFilterListController(self, countForFilterType: .top)))
-            
-            if (currentFilterType == .top) {
-                myCell.iconImageView?.image = #imageLiteral(resourceName: "places-suggestion-top")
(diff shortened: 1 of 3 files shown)


Wikipedia/Code/PlaceSearchFilterListController.swift
```diff
@@ -50,21 +50,15 @@ class PlaceSearchFilterListController: UITableViewController {
...
-                myCell.iconImageView?.image = #imageLiteral(resourceName: "places-filter-top-disabled")
-            }
-
+            myCell.iconImageView?.image = #imageLiteral(resourceName: "places-suggestion-top")
         } else if (indexPath.row == 1) {
             myCell.titleLabel.text = localizedStringForKeyFallingBackOnEnglish("places-filter-saved-articles")
-            myCell.subtitleLabel.text = localizedStringForKeyFallingBackOnEnglish("places-filter-saved-articles-count").replacingOccurrences(of: "$1", with: String(delegate.placeSearchFilterListController(self, countForFilterType: .saved)))
-            
-            if (currentFilterType == .saved) {
+            let savedCount = delegate.placeSearchFilterListController(self, countForFilterType: .saved)
```
joewalsh: Do the saved articles get locations added somewhere before this count is taken? Users with existing libraries will have saved articles in the db that, even though they have locations on Wikipedia, don't reflect that in core data (since we weren't fetching it previously)

amrox: they don't get updated specifically. Both counts for Saved and Top are a little wonky because they only needed to be somewhat accurate for user-testing

amrox: @joewalsh I filed a ticket https://phabricator.wikimedia.org/T163821 

it's out of scope for this PR though, which is minor presentation tweaks

PR #1378, opened 2017-04-27
Refactor places search routines

amrox:
### Tickets
 - https://phabricator.wikimedia.org/T163567
 - https://phabricator.wikimedia.org/T163821

### Notes
 - Refactored some (not all) of the search logic in `PlaceSearchService`
 - Created a `PlaceSearchResult` struct
    - originally it was going to return a `NSFetchRequest` for either type of search (`.top` or `.saved`) but in practice that was more trouble than it was worth. So `.top` returns `[MWKLocationSearchResult]` and `.saved` returns `NSFetchRequest<WMFArticle>`
 - I would like to move the `performWikiDataQuery`logic into the `PlaceSearchService` too, but it's working as is, and it's a reasonable checkpoint

### Issues
 - If you're on the Saved filter, and you pan around enough so the "Redo Search" button appears:
     - The count of Top Articles will be cleared
     - But it won't be re-fetched until the user performs a search
     - I didn't want to auto-trigger a top articles search in the background just to update that count
     - However I think we should put something better than "0 top articles in there"
         - maybe "Tap to see Top Articles"


Wikipedia/Code/PlaceSearchService.swift
```diff
@@ -0,0 +1,144 @@
...
+                let savedPagesWithoutLocation = try moc.fetch(savedPagesWithoutLocationRequest)
+                guard savedPagesWithoutLocation.count > 0 else {
+                    done()
+                    return
+                }
+                let urls = savedPagesWithoutLocation.flatMap({ (article) -> URL? in
+                    return article.url
+                })
+                // TODO: ARM: I don't understand this --------------------V
+                //var allArticlesWithLocation = savedPagesWithLocation // this should be re-fetched
```
amrox: @joewalsh what does the comment 

```
// this should be re-fetched
```
mean?

Wikipedia/Code/PlacesViewController.swift
```diff
@@ -611,74 +628,108 @@ class PlacesViewController: UIViewController, MKMapViewDelegate, UISearchBarDele
...
-        switch search.type {
-        case .location:
-            guard search.needsWikidataQuery else {
+        if (search.filter == .top && search.type == .location) {
+            if (search.needsWikidataQuery) {
+                performWikidataQuery(forSearch: search)
+                return
+            } else {
+                // TODO: ARM: I don't understand this
                 tracker?.wmf_logActionTapThrough(inContext: searchTrackerContext, contentType: AnalyticsContent(siteURL))
```
amrox: @joewalsh can you explain this analytics call?

Wikipedia/Code/PlacesViewController.swift
```diff
@@ -611,74 +628,108 @@ class PlacesViewController: UIViewController, MKMapViewDelegate, UISearchBarDele
...
+                guard let request = result.fetchRequest else {
+                    DDLogError("Error fetching saved articles: fetchRequest was nil")
+                    return
+                }
+                
+                self.articleFetchedResultsController = NSFetchedResultsController<WMFArticle>(fetchRequest: request, managedObjectContext: self.dataStore.viewContext, sectionNameKeyPath: nil, cacheName: nil)
+                
+                do {
+                    let articlesToShow = try moc.fetch(request)
+                    self.articleKeyToSelect = articlesToShow.first?.key
```
amrox: @joewalsh I don't think a saved filter was every pre-selecting an article. I added this. Good/bad?

Wikipedia/Code/PlaceSearchService.swift
joewalsh: It meant since we had made a network request and returned to the main queue, those core data objects might have been deleted while we were away, so they weren't safe to use. The comment is invalid now that this code uses `dataStore.viewContext.updateOrCreateArticleSummariesForArticles(withURLs: urls) { (articles)` instead of updating the articles here.

Wikipedia/Code/PlacesViewController.swift
joewalsh: Logs the search action as defined by https://phabricator.wikimedia.org/T159769

joewalsh: Looks like it's in the wrong place

Wikipedia/Code/PlacesViewController.swift
joewalsh: I think after discussing with Josh & Carolyn the decision was to not auto-select after the saved articles search, but I could be mis-remembering.

PR #1379, opened 2017-04-27
override isEqual for MWKSearchResult

nambatee:
https://phabricator.wikimedia.org/T142797


Wikipedia/Code/MWKSearchResult.m
```diff
@@ -235,4 +235,33 @@ - (nullable NSURL *)articleURLForSiteURL:(NSURL *)siteURL {
...
+        return NO;
+    }
+    
+    return [self isEqualToSearchResult:(MWKSearchResult *)object];
+    
+}
+
+- (NSUInteger)hash {
+    return [self.displayTitle hash];
+}
```
julienbodet: @nambatee I thought about the same solution as yours, but I don't think that overriding `isEqual:` and `hash` methods is the right way to correct this. MWKSearchResults is a subclass of `MTLModel` that checks equality of two `MTLModel` objects not based on their adress, but on `MTLModel` implementation which uses propertie keyx defined inthe objects. 

So, I think the problem is that one of the property doesn't return `true` when `isEqual:` is called on it, while it should.

nambatee: Thanks @julienbodet, I'm not 100% confident about the solution, my understanding is that `containsObject` returns `false` cause those objects are equal, but not identical.

joewalsh: This solution works, but I'm not sure if overriding `isEqual` has other consequences for subclasses of `MTLModel`. The reason they aren't equal is due to the `index` property - the results have different index values so `MTLModel`'s `isEqual` implementation returns false. A lower impact fix would be to alter [mergeResultsFromModel:](https://github.com/wikimedia/wikipedia-ios/blob/dff6ea281f0fa262a008e49843680778215d5c57/Wikipedia/Code/WMFSearchResults.m#L98) to check for conflicting display titles rather than checking for `containsObject:`.

coreyfloyd:
@joewalsh @nambatee 

If the index properly is the problem, you can exclude the index properly form the equality check by setting this flag for the index property:
MTLPropertyStorageTransitory

You may need to set this explicitly to make sure it is still encoded Shane archived though:
MTLModelEncodingBehaviorUnconditional

That way you don't need to override isequal

joewalsh:
@coreyfloyd yeah, I tried that yesterday. It caused an infinite recursion crash. I updated the model version and overrode `description` similar to `MWKCitation`. I'd rather not mess with the model object at all. The better fix is what I described in my previous comment (updating the merge). CC @nambatee

Wikipedia/Code/WMFSearchResults.m
```diff
@@ -96,8 +96,13 @@ - (void)mergeRedirectMappingsFromModel:(WMFSearchResults *)searchResults {
 }
 
 - (void)mergeResultsFromModel:(WMFSearchResults *)searchResults {
+    NSMutableArray *displayTitlesOfExistingResults = [NSMutableArray new];
```
joewalsh: This should be a `NSMutableSet` so that the lookup below is constant time instead of linearly increasing with the size of `displayTitlesOfExistingResults`.

Wikipedia/Code/WMFSearchResults.m
```diff
@@ -96,8 +96,13 @@ - (void)mergeRedirectMappingsFromModel:(WMFSearchResults *)searchResults {
...
+    for (MWKSearchResult *obj in self.results) {
+        [displayTitlesOfExistingResults addObject:obj.displayTitle];
```
joewalsh: Needs a nil check on displayTitle

Wikipedia/Code/WMFSearchResults.m
```diff
@@ -96,8 +96,13 @@ - (void)mergeRedirectMappingsFromModel:(WMFSearchResults *)searchResults {
...
+    }
+    
     NSArray *newResults = [searchResults.results wmf_reject:^BOOL(MWKSearchResult *obj) {
-        return [self.results containsObject:obj];
+        return [displayTitlesOfExistingResults containsObject:obj.displayTitle];
```
joewalsh: Also needs a nil check on displayTitle

nambatee: Thanks Joe! If displayTitle is nil, should we then check on object like before?

joewalsh: if `obj.displayTitle` is nil, `return YES;` and exclude the result from the new array. `displayTitle` _should_ never be nil, but technically it's `nullable`, so this is just to prevent a crash if that ever happens, since `containsObject:` and `addObject:` will crash when passed nil.

joewalsh:
ok to test

joewalsh:
@nambatee ^ it looks like there were tests written for that merge code that need to be updated. `[WMFSearchResultMergeTests merging_results__should_preserve_result_ordering__putting_original_results_first]`
and
`[WMFSearchResultMergeTests merging_results__should_not_insert_results_already_present_in_the_original]`

joewalsh:
@nambatee on second thought, delete `WMFSearchResultMergeTests.m` since 1. they didn't actually work, it was supposed to be checking for exactly this bug and 2. they're written using Quick and we're moving away from that. We can re-write them as a follow on.

joewalsh:
ok to test

joewalsh:
jenkins, retest this please

joewalsh:
Merging. Thanks @nambatee !

PR #1384, opened 2017-04-28
When switching the viewMode, the redoSearchButton was set to visible …

amrox:
…if the mapView was visible

What we really want is to hide the redoSearchButton if the map view is not visible, but only show it if necessary when it is visible


amrox:
!jenkins test this please

PR #1388, opened 2017-04-28
Footer pixel push round two

montehurd:
https://phabricator.wikimedia.org/T164117

## Hairline menu item dividers and moved image to right side of read more cards:
screen shot 2017-04-28 at 6 23 36 pm

## Inertial scrolling!
scroll mov


montehurd:
For the record, if we ever want to stop using side-to-side scrolling for read more pages the following CSS changes will do the trick:

To stack read more items vertically:
<pre>
#footer_container_readmore_pages
	/* width: auto; */
	/* overflow-x: scroll; */
	/* -webkit-overflow-scrolling: touch; */
	/* white-space: nowrap; */

.footer_readmore_page
	/* display: inline-block; */
	/* width: 250px; */
</pre>

screen shot 2017-04-28 at 7 23 54 pm

montehurd:
Testing image (phab thinks it's too big to attach directly to the ticket):

d mov

PR #1390, opened 2017-04-29
Remove old crufty 'transformer.js' file.

montehurd:
Wasn't adding anything. At one time it may have handled JS-to-native bridge communications bits.


amrox:
!jenkins test this please

PR #1395, opened 2017-05-01
change row animation from automatic to fade when clearing history

nambatee:
https://phabricator.wikimedia.org/T163730


joewalsh:
ok to test

PR #1397, opened 2017-05-02
Make TOC text direction always match article text direction.

montehurd:
https://phabricator.wikimedia.org/T149066

When device lang is LTR...

# Before
The TOC was still LTR when the article was RTL:
before mov

# After
The TOC is RTL when the article is RTL:
after mov


montehurd:
Oops forgot to test when device lang *is* RTL and you load a LTR article...

Wikipedia/Code/WMFTableOfContentsViewController.swift
```diff
@@ -32,6 +32,8 @@ open class WMFTableOfContentsViewController: UIViewController,
                                                WMFTableOfContentsAnimatorDelegate {
     
     let tableOfContentsFunnel: ToCInteractionFunnel
+
+    let semanticContentAttribute: UISemanticContentAttribute
```
montehurd: Considered naming this `semanticContentAttributeOverride`...

joewalsh:
retest this please

Wikipedia/Code/WMFTableOfContentsViewController.swift
amrox: actually I like `semanticContentAttributeOverride`

PR #1402, opened 2017-05-03
Update to latest version of page-library.

montehurd:
https://phabricator.wikimedia.org/T164425

Formerly know as "applib".

https://github.com/wikimedia/wikimedia-page-library


Wikipedia/assets/index.js
```diff
@@ -1195,11 +1200,11 @@ var findClosest = function findClosest(el, selector) {
  * @return {boolean}        Whether table ancestor of 'el' is found
  */
 var isNestedInTable = function isNestedInTable(el) {
-  return findClosest(el, 'table') !== null;
+  return !!findClosestAncestor(el, 'table');
```
joewalsh: should this be `!!` or just `!`?

montehurd: Ya I had same question [here](https://github.com/wikimedia/wikimedia-page-library/pull/21#issuecomment-297226657). Was changed in the "favor undefined to null" [PR](https://github.com/wikimedia/wikimedia-page-library/pull/21). @niedzielski would you be open to adding a comment to the places this is used that it's intentional?

niedzielski: Fixed in https://github.com/wikimedia/wikimedia-page-library/pull/27

PR #1403, opened 2017-05-04
Integrate collapse table JS from page-library

montehurd:
o mov


montehurd:
Pushing a clean version.

PR #1410, opened 2017-05-05
T130159 More prominent search

amrox:
https://phabricator.wikimedia.org/T130159

### Items of note:

- Explore now has it's own storyboard
- The old WMFExploreViewController has been renamed to WMFExploreCollectionViewController is now embedded inside a new WMFExploreViewController
 - the snapping behavior when you scroll so the search bar is only partially exposed may need some tweaking

out

### Todo
 - [x] Fix tests
 - [x] Add analytics categories to Wrapper view controller
 - [x] Fix tint colors
 - [x] "Snap" scrolling
- [x] Fade in search icon in upper right
 - [x] Transition from "Wikipedia" title to "W" mark
 

### Progress

search collapse mov


Wikipedia/Code/ExploreViewController.swift
```diff
@@ -0,0 +1,277 @@
...
+    }
+    
+    func exploreCollectionViewController(_ collectionVC: WMFExploreCollectionViewController, willBeginScrolling scrollView: UIScrollView) {
+        isUserScrolling = true
+    }
+    
+    func exploreCollectionViewController(_ collectionVC: WMFExploreCollectionViewController, didScroll scrollView: UIScrollView) {
+        //DDLogDebug("scrolled! \(scrollView.contentOffset)")
+        
+        guard self.view != nil else {
```
joewalsh: doesn't calling self.view cause the view to load?

joewalsh: (or is causing the view to load intentional? otherwise there's `@available(iOS 3.0, *)
    open var isViewLoaded: Bool { get }`)

Wikipedia/Code/UIViewController+WMFNavigationBarUtilities.swift
```diff
@@ -0,0 +1,40 @@
...
+        // Setup extended navigation bar
+        //   Borrowed from https://developer.apple.com/library/content/samplecode/NavBar/Introduction/Intro.html
+  
+        guard let bgColor = view.backgroundColor else {
+            assertionFailure("Could not get background color of view")
+            return
+        }
+        assert(bgColor != UIColor.clear, "Background color must not be clear")
+        
+        view.shadowOffset = CGSize(width: 0, height: CGFloat(1) / UIScreen.main.scale)
```
joewalsh: `UIScreen.main.scale` should be `traitCollection.displayScale` to account for presentation places other than the main screen

joewalsh:
@amrox looks good overall, only minor questions

Wikipedia/Code/ExploreViewController.swift
amrox: I think that was some weirdness with the original segue embed technique I was doing. I think it can be removed

PR #1413, opened 2017-05-07
Add maps item to the list of libraries used. Replace old buggy 3rd party acknowledgements library with simple swift replacement.

montehurd:
https://phabricator.wikimedia.org/T159974

- adds the maps item from the phab ticket
- replaces the buggy [third party acknowledgements library](https://github.com/vtourraine/VTAcknowledgementsViewController) we were using with cleaner Swift controllers

### Looks as it did previously, but now with TWN VC title, better readability, better RTL layout and Dynamic Type support:
maps mov

Note: the VC's title is localized, but the library names themselves (and their license copy) are not localized. This remains as it was.

### New bits implemented in 3 "Libraries Used" files:
screen shot 2017-05-07 at 6 01 42 pm

### Third party library's bugs / issues which finally led me to just use Swift instead:
- not configured for dynamic type
- view controller title localization was incomplete and not via TWN
- poor handing of line breaks in license text making word wrapping behave badly
- convoluted handing of items in its plist data file (odd [stripping of first and last items](https://github.com/vtourraine/VTAcknowledgementsViewController/blob/master/Classes/VTAcknowledgementsParser.m#L61))
- the bug seen in this screenshot (which was related in some unclear way to the item above - finally got to the point where just using Swift seemed easier than debugging the library):

screen_shot_2017-05-04_at_6 43 18_pm

It turned out there were RTL issues as well (fixed below) but I didn't realize this until after implementing the Swift file.


Wikipedia/Code/LibrariesUsed.swift
```diff
@@ -0,0 +1,180 @@
...
+            let multiNewlineRegex = newlineBoundedWhitespaceRegex,
+            let whitespaceRegex = whitespaceRegex
+        else {
+            assertionFailure("Regex's didn't compile!")
+            return licenseString
+        }
+        var string = licenseString
+        let breaksPlaceholder = "#breaks_placeholder#"
+        string = multiNewlineRegex.stringByReplacingMatches(in: string, options: [], range: NSRange(location: 0, length: string.characters.count), withTemplate: breaksPlaceholder)
+        string = string.replacingOccurrences(of: "\n", with: " ")
```
joewalsh: Might be safer to use `newlineCharacterSet`

montehurd: Ah that's right... lemme tweak that...

montehurd: Fixed!

montehurd:
@joewalsh Unsure about non-versioned link...

PR #1414, opened 2017-05-08
Start & end background tasks for feed content fetches

joewalsh:
With our core data sqlite db stored in the shared app group container, the OS is more sensitive about us keeping a lock on that file when the app is suspended. These changes ensure we start a background task for feed import operations that will access the db. This is in addition to the existing background tasks for database migration and housekeeping.

More info here:
http://blog.supertop.co/post/152615019837/how-not-to-crash-1


WMF Framework/WMFExploreFeedContentController.m
```diff
@@ -9,26 +9,50 @@
 #import "WMFAssertions.h"
 #import <WMF/WMF-Swift.h>
 
+NSString * const WMFExploreFeedContentControllerBusyStateDidChange = @"WMFExploreFeedContentControllerBusyStateDidChange";
+
 static const NSTimeInterval WMFFeedRefreshTimeoutInterval = 12;
 static NSTimeInterval WMFFeedRefreshBackgroundTimeout = 30;
+static const NSString *kvo_WMFExploreFeedContentController_operationQueue_operationCount = nil;
```
amrox: super minor, but I used to do 

static const NSString *kvo_WMFExploreFeedContentController_operationQueue_operationCount = "kvo_WMFExploreFeedContentController_operationQueue_operationCount";

just in case I wanted to print out the context in the debugger. In practice I think it helped maybe once ¯\_(ツ)_/¯

WMF Framework/WMFExploreFeedContentController.m
```diff
@@ -9,26 +9,50 @@
...
 
 @interface WMFExploreFeedContentController ()
 
 @property (nonatomic, strong) NSArray<id<WMFContentSource>> *contentSources;
-@property (nonatomic, strong) WMFTaskGroup *taskGroup;
-@property (nonatomic, strong) NSMutableArray <dispatch_block_t>*queue;
+@property (nonatomic, strong) NSOperationQueue *operationQueue;
```
amrox: +1 for a real `NSOperationQueue`

WMF Framework/WMFExploreFeedContentController.m
joewalsh: Good point, updated

joewalsh:
@montehurd consolidated "state" key path usage with b5494a6

PR #1398, opened 2017-05-02
Filter dropdown button improvements

amrox:
- Change filter dropdown to standard blue button color
    - Goes back to black when not active (i.e. in search)
 - increased height to 44
 - added 22 px to either side

https://phabricator.wikimedia.org/T164223

screenshot 2017-05-02 16 08 25
 
screenshot 2017-05-02 16 11 24

PR #1418, opened 2017-05-09
forcing blue in search filter bar

amrox:
Looks like the default tint color wasn't reliable

see https://github.com/wikimedia/wikipedia-ios/pull/1398/files


Wikipedia/Code/UIColor+WMFStyle.swift
```diff
@@ -51,5 +51,6 @@ public extension UIColor {
     public static let wmf_orange: UIColor = .wmf_color(withHex: 0xff5b00, alpha: 1.0)
     public static let wmf_authTitle: UIColor = .wmf_color(withHex: 0x72777d, alpha: 1.0)
     public static let wmf_filterDropDownBackground: UIColor = .wmf_color(withHex: 0xf6f6f6, alpha: 1.0)
+    public static let wmf_blueTitleColor: UIColor = .wmf_color(withHex: 0x274ec0, alpha: 1.0)
```
montehurd: Looks to close(ish) to wmf_darkBlueTint, but not too worried about it.

amrox: I pulled the color from Zeplin but let me double check with design

PR #1419, opened 2017-05-10
Exclude pronunciation guides from article extracts

montehurd:
https://phabricator.wikimedia.org/T155573

- [x] Determine repo steps. 
  - clean install
  - save `enwiki > Ken MacKinnon`
  - save `Scottish Gaelic`
  - refresh feed twice
  - find `Sar GHaidheal Fellowships` and `Alba` in feed `Because you read` recommendations
  - tap `Save for later` on each ^
  - refresh feed
  - find `Sar GHaidheal Fellowships` and `Alba` again in feed `Because you read` recommendations
  - each will now have bracketed and/or parenthetic substrings in their extracts

- [x] Audit methods involved
  - MWKSection `summary`
  - NSString+WMFHTMLParsing `wmf_stringByRecursivelyRemovingParenthesizedContent`
  - NSString+WMFHTMLParsing `wmf_stringByRemovingBracketedContent`
  - NSString+WMFHTMLParsing `wmf_summaryFromText`
  - WMFArticlePreviewCollectionViewCell `snippetText`

- [x] Resolve Inconsistent naming? In some places we use `extract`, in other places we use `snippet`, in other places `summary`... (Added TODO note as reminder)
- [x] Quick hygiene commit removing some unused native code I noticed
- [x] Wrangle some hairy string cleaning code and ensure it gets called such that repro steps no longer repro (creating follow-up ticket to re-examine these bits after some upstream code is merged)
- [x] Ensure tests still pass

-----
Found other explore bug while debugging this one:
- clean install
- save `enwiki > Dr. Jekyll and Mr. Hyde (character)`
- refresh feed twice
- find `Strange Case` in feed `Because you read` recommendations
- tap save on ^
- refresh feed
- scroll down to where `Strange Case` was in the `Because you read` recommendations
- it now has an image, no title, description or extract:
screen shot 2017-05-09 at 7 54 20 pm


montehurd:
@joewalsh Hey I put some notes in the description above with repro steps and some of the issues I think we'll want to deal with. Wanted to ping you before going further as I suspect there are a few ways we could handle these bits...

joewalsh:
@montehurd `SummaryExtensions.swift` line 21 just needs `extract.wmf_summaryFromText()`. That's the only remaining place where parenthesized content is being added to the `WMFArticle` object. I don't think we should remove parenthesized content from articles the user already has cached (would require a migration) but that's a decision for @jtminor.

I don't think it's worth unifying naming at this time (again, because it would require a database migration) however, "extract" seems to be the most common naming with the endpoints we use.

Don't file the `Strange Case` bug unless it's repro-able with a featured article. The "because you read" section is being completely re-done with different cells.

PR #1420, opened 2017-05-10
Update becasue you read explore card design

joewalsh:
https://phabricator.wikimedia.org/T141767


Wikipedia/Code/SaveButtonsController.swift
```diff
@@ -0,0 +1,63 @@
...
+        self.dataStore = dataStore
+        super.init()
+        NotificationCenter.default.addObserver(self, selector: #selector(articleUpdated(notification:)), name: NSNotification.Name.WMFArticleUpdated, object: nil)
+    }
+    
+    deinit {
+        NotificationCenter.default.removeObserver(self)
+    }
+    
+    @objc(willDisplaySaveButton:forArticle:) func willDisplay(saveButton: SaveButton, for article: WMFArticle) {
```
amrox: I slightly prefer the `@objc` and `func` on separate lines

```
@objc(willDisplaySaveButton:forArticle:)
func willDisplay(saveButton: SaveButton, for article: WMFArticle)
```

I think I've seen that format elsewhere in the code. What do you think?

joewalsh: Yeah, you're right, that is easier to read. Same rule for class names too or just functions?

amrox: Yeah, I like for class names too. The inline `@objc` isn't as bad if it doesn't have parameters, but I still prefer 2 lines. Maybe it's just because I'm used to other languages that use that formatting.

http://www.saltycrane.com/blog/2010/03/simple-python-decorator-examples/
https://docs.oracle.com/javase/tutorial/java/annotations/basics.html

PR #1422, opened 2017-05-11
Fix Explore status bar tap "scroll to top" after 3D-peek

montehurd:
https://phabricator.wikimedia.org/T128688

Fix Explore status bar tap "scroll to top" when:
- [x] coming back to Explore after peeking AND popping to an article
- [x] coming back to Explore after peeking but NOT popping to an article

The two cases above are now working properly:
title mov

@joewalsh @amrox This fixes both cases but the [latter fix](https://github.com/wikimedia/wikipedia-ios/pull/1422/commits/fa48a882d2afb5048055da545ac28680a4cda736) seems more complicated than it should be :/
Perhaps there's an easier way to know when [a peek has ended without popping](http://stackoverflow.com/a/33778293/135557)? (`viewDidAppear` doesn't seem to re-fire if you peek but then don't pop)

### DIFF
--- a/Wikipedia/Code/WMFExploreViewController.m
+++ b/Wikipedia/Code/WMFExploreViewController.m
@@ -1583,7 +1583,37 @@ const NSInteger WMFExploreFeedMaximumNumberOfDays = 30;
         ((WMFArticleViewController *)vc).articlePreviewingActionsDelegate = self;
     }
 
+    [previewingContext.previewingGestureRecognizerForFailureRelationship addObserver:self
+                                                                          forKeyPath:kvo_WMFExploreViewController_peek_state_keypath
+                                                                             options:NSKeyValueObservingOptionNew
+                                                                             context:&kvo_WMFExploreViewController_peek_gesture_recognizer_for_failure_relationship];
+
     return vc;
+}
+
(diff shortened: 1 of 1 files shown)


Wikipedia/Code/WMFExploreViewController.m
```diff
@@ -1581,9 +1582,39 @@ - (nullable UIViewController *)previewingContext:(id<UIViewControllerPreviewing>
...
 
+    [previewingContext.previewingGestureRecognizerForFailureRelationship addObserver:self
+                                                                          forKeyPath:kvo_WMFExploreViewController_peek_state_keypath
+                                                                             options:NSKeyValueObservingOptionNew
+                                                                             context:&kvo_WMFExploreViewController_peek_gesture_recognizer_for_failure_relationship];
+
     return vc;
 }
 
+static const NSString *kvo_WMFExploreViewController_peek_gesture_recognizer_for_failure_relationship = nil;
```
joewalsh: @amrox pointed out that for debugging it's helpful to assign this to `@"kvo_WMFExploreViewController_peek_gesture_recognizer_for_failure_relationship"` - it's not that way other places in the codebase but would be good to do going forward (and fix as we encounter them)

PR #1425, opened 2017-05-11
Update featured article feed appearance

joewalsh:
https://phabricator.wikimedia.org/T164979
https://phabricator.wikimedia.org/T164526
https://phabricator.wikimedia.org/T164527


joewalsh:
@montehurd is `WMFDisambiguationPagesViewController` reachable anywhere? I don't see it in the footer section of disambiguation pages. Asking because that's the last place the old article preview table view cell is referenced.

montehurd:
Looks like this fixes https://phabricator.wikimedia.org/T165111

montehurd:
Looks like the mock has multiple line extracts, but I'm seeing just one line..

screen shot 2017-05-11 at 6 59 50 pm

Otherwise looks great!!!

montehurd:
Interesting... I tested a little more and some featured cards do layout with multiple lines... but then if I scroll around and refresh sometimes I get no lines...

screen shot 2017-05-11 at 7 03 45 pm

montehurd:
I also tried saving some articles and refreshing while testing, so that may help for repro-steps.

joewalsh:
@montehurd I can't repro the single line issue, that featured article shows multiple lines for me. I have repro'd the issue where the text is entirely missing (in addition, we have two featured articles that are missing extracts server-side - http://en.wikipedia.org/api/rest_v1/page/summary/Belgium_national_football_team and http://en.wikipedia.org/api/rest_v1/page/summary/Kuiper_belt)

joewalsh:
@montehurd pushed some changes that should help with the missing text issue, but I was unable to find a reliable repro. Let me know if you can still repro.

montehurd:
iirc I saved a few articles and refreshed and rotated. Will retest when I get to office.

PR #1427, opened 2017-05-12
Fix explore feed layout issues

joewalsh:
- `CollectionViewCell` is the base class for cells with manual layout
- `ArticleCollectionViewCell` is the base class for cells that represent articles
- `wmf_preferredFrame` methods help with sizing & positioning subviews


montehurd:
Can we talk about the constraints issues before switching to manually laying things out? I'm not doubting there were reasons to take a stab at a manual approach - I'd just like to see what the auto layout issue are first. I actually enjoy debugging auto layout fickleness ... I'm weird like that :)

joewalsh:
@montehurd thanks for getting at the root cause! 

However, in this case, the manual layout fixes a second issue - layout speed. With 30 days of feed history, calculating the layout for the collection view with auto layout takes an extra 600ms on a 6S, causing the app to hang at the splash screen. With these changes, it's down to 80ms.

It would be possible to update the collection view layout to have only a rough estimate of layout height that gets updated with a more refined estimate as the user scrolls. However, rather than trying to maintain accurate estimates for different cell widths (and update the collection view content size on the fly), I'd rather have the simplicity of a faster layout.

montehurd:
@joewalsh Any opposition to me doing a quick patch job on the other xib so we can performance test with all ambiguous layouts resolved?

montehurd:
Manual layouts are just so much harder to grok and maintain... especially if we have to manually deal with things like RTL. I'm open to it of course, but would rather performance test against an optimized xib implementation.

montehurd:
Or... could you re-test using the xib I edited disabling the other layout type for the comparison?

joewalsh:
@montehurd sure, if you can make the auto layout as performant (or within the same order of magnitude) as the manual layout, then it's the best of both worlds.

However, I did run into this issue on your fork:
simulator screen shot may 13 2017 6 16 31 am
Repro steps were clean install, go through onboarding to the feed, rotate back and forth a few times, save featured article, pull to refresh. There were no constraint error logs.

Also, I'd be happy to work with you to clean up and properly document the manual layout so it's easier to grok (and re-use). It is messy at the moment, I just wanted to verify it would all work within the confines of the collection view. I used a similar setup before auto layout existed and it worked well.

montehurd:
@joewalsh 

Thanks I'll take a stab at it and ping you with branch. Thx for screenshot too (I hid all 'because you read' cards while fixing the xib I hacked.) 

----- 

Tangential performance question that came to mind: 

Are we caching these cell sizes once they're determined for a given cell? If not could we make a size checksum from a digit indicating screen orientation, digits for lengths of each of the cell's label's text, digit for if there's an image, digit indicating dynamic type size, and a string indicating cell class name  - that way *any* changes affecting cell layout size would change the checksum, so we'd know to recalculate and cache and re-use the cached size for that cell. (Note that I think the checksum could be article title agnostic - one method could calculate checksum for all cell types.)

Then we could just have a global set 'cellSizeChecksumCache' of these checksums and their respective sizes. (We'd ask the cell for its size checksum before asking the system to do potentially expensive fitting size calculation and we'd see if we had that checksum already cached.) Could limit its size in some way, or use a first-in-last-out array of size 1000 or something reasonable. Seems super easy to implement and understand and then we'd drastically reduce number of times the layout would have to ask the system to figure out layout size.

montehurd:
I should probably have waited to ask caching questions until I was not on my phone so I could search the code base to see if we're already caching sizes in some way :)

joewalsh:
@montehurd we are caching the sizes once they're determined. The slowness is the initial layout pass calculating the content size of the collection view, which caching doesn't help with.

montehurd:
>  The slowness is the initial layout pass calculating the content size of the collection view, which caching doesn't help with.

Doesn't that kind of depends on a few things? Implementation details like persisting cache across restarts vs not. 

 Still interesting to ponder real world use case from clean install where user isn't loading so many cells that performance becomes an issue but by time they do have so many that it would be the caching alleviates it.

montehurd:
Anyway just throwing out ideas - sorry for blowing up your PR thread ;)

joewalsh:
> Doesn't that kind of depend on a few things? Implementation details like persisting size cache across restarts vs not.

I think the added complexity of a persistent layout cache would be more of a burden long term than the added complexity of cells with manual layout code. Manual layout code replaces one piece of functionality with another, a persistent layout cache is a whole new piece of functionality that would need to be built, maintained, and tested.

I agree that manual layout code can be a burden. However, if we work together to make the manual layout code easier to understand, we can lessen or eliminate that burden. For these cells, even with xibs there is a code-side component for managing layout (image vs no image, save button vs no save button, different semantic content attributes). Because of this existing code-side complexity, going full manual isn't as high of a cost as converting a view that was entirely xib-based. The working manual layout took a few hours to implement, works out of the box, and had the unintended upside of helping with the layout speed issue. I've spent much longer than that building and debugging auto layout versions of the cells (even before the incorrect stack view implementation).

I'm not advocating for manual layout over auto layout everywhere - I still think xib based auto layout should be the default. However, in this case, I think the upsides of manual layout outweigh the downsides.

montehurd:
For the record, I think Joe's reasoning here is solid.

WMF Framework/LabelGroupAccessibilityElement.swift
```diff
@@ -0,0 +1,35 @@
...
+    }
+    
+    func update() {
+        guard let firstLabel = labels.first else {
+            return
+        }
+        var combinedLabel = firstLabel.accessibilityLabel ?? firstLabel.text ?? ""
+        var combinedFrame = firstLabel.frame
+        for label in labels[1..<labels.count] {
+            combinedFrame = combinedFrame.union(label.frame)
```
montehurd: These rect methods are so handy!

WMF Framework/UIView+SemanticContent.swift
```diff
@@ -7,3 +7,34 @@ extension UIView {
         }
     }
 }
+
+extension UIView {
+    public func wmf_prefferedFrame(at point: CGPoint, fitting size: CGSize, alignedBy semanticContentAttribute: UISemanticContentAttribute, apply: Bool) -> CGRect {
```
montehurd: Oops typo.

WMF Framework/UILabel+HasText.swift
```diff
@@ -0,0 +1,5 @@
+extension UILabel {
+    public var wmf_hasText: Bool {
+        return (text as NSString?)?.length ?? 0 > 0
```
montehurd: Was NSString used here because of issues related to Swift strings not currently being collections of characters and how that makes character counts screwy in some langs?

WMF Framework/UIView+SemanticContent.swift
```diff
@@ -0,0 +1,40 @@
...
+        if #available(iOS 10.0, *) {
+            return self.effectiveUserInterfaceLayoutDirection
+        } else {
+            return UIView.userInterfaceLayoutDirection(for: semanticContentAttribute)
+        }
+    }
+}
+
+extension UIView {
+    public func wmf_prefferedFrame(at point: CGPoint, fitting size: CGSize, alignedBy semanticContentAttribute: UISemanticContentAttribute, apply: Bool) -> CGRect {
```
montehurd: Can these methods go in the first UIView extension in the file?

Wikipedia/Code/ArticleCollectionViewCell+WMFFeedContentDisplaying.swift
```diff
@@ -0,0 +1,60 @@
...
+    }
+    
+    public func isSaveButtonHidden(for displayType: WMFFeedDisplayType) -> Bool {
+        return displayType == .pageWithPreview ? false : true
+    }
+    
+    public func configure(article: WMFArticle, contentGroup: WMFContentGroup, layoutOnly: Bool) {
+        let displayType = contentGroup.displayType()
+        
+        let imageWidthToRequest = imageView.frame.size.width < 300 ? traitCollection.wmf_nearbyThumbnailWidth : traitCollection.wmf_leadImageWidth
```
montehurd: For future reference, can we comment why we chose `300` here, or give it a named var for a context hint?

Wikipedia/Code/ArticleCollectionViewCell.swift
```diff
@@ -1,109 +1,109 @@
...
+        addSubview(titleLabel)
+        addSubview(descriptionLabel)
+        descriptionLabel.textColor = UIColor.wmf_customGray
+        addSubview(saveButton)
+        saveButton.tintColor = UIColor.wmf_blueTint
+        saveButton.setTitleColor(UIColor.wmf_blueTint, for: .normal)
+        saveButton.saveButtonState = .longSave
+        saveButton.addObserver(self, forKeyPath: "titleLabel.text", options: .new, context: &kvoButtonTitleContext)
+        backgroundColor = .white
+        super.setup()
```
montehurd: When we're doing frames manually, as guidance for future volunteers do we maybe want assertions somewhere verifying that we haven't set autoresizingMasks or constraints on such views?

montehurd: I don't feel strongly about it - just a thought.

Wikipedia/Code/ArticleFullWidthImageCollectionViewCell.swift
```diff
@@ -1,17 +1,57 @@
...
+                imageView.frame = CGRect(x: 0, y: 0, width: size.width, height: imageViewHeight)
+            }
+            origin.y += imageViewHeight
+        }
+        
+        origin.y += margins.top
+        
+        
+        let titleFrame = titleLabel.wmf_prefferedFrame(at: origin, fitting: widthMinusMargins, alignedBy: articleSemanticContentAttribute, apply: apply)
+        origin.y += titleFrame.height > 0 ? titleFrame.height + spacing : 0
```
montehurd: Could this height with spacing be DRY'ed to a rect extension? I see it's used a few places below...

Wikipedia/Code/CollectionViewCell.swift
```diff
@@ -0,0 +1,81 @@
+import UIKit
+
+// CollectionViewCell is the base class of collection view cells that use manual layout.
```
montehurd: Looks good!

montehurd:
Added some comments on some very minor things. Overall looks good. Gonna test a bit more.

montehurd:
I noticed a couple small issues when testing on iOS 9 in landscape:

Tight spacing here:
screen shot 2017-05-15 at 5 17 27 pm


Couldn't scroll all the way to the bottom here:
screen shot 2017-05-15 at 5 18 33 pm

montehurd:
I was clicking around the feed on iOS 10.3 and somehow got this `UIButtonLabel was deallocated while key value observers were still registered with it` crash:

<pre>
2017-05-15 17:31:10.853 Wikipedia[11059:531345] *** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'An instance 0x7ff803d25770 of class UIButtonLabel was deallocated while key value observers were still registered with it. Current observation info: <NSKeyValueObservationInfo 0x60800082cb00> (
<NSKeyValueObservance 0x608000859a10: Observer: 0x608000859980, Key path: text, Options: <New: YES, Old: NO, Prior: NO> Context: 0x0, Property: 0x60000024de30>
)'
*** First throw call stack:
(
	0   CoreFoundation                      0x0000000106d0cb0b __exceptionPreprocess + 171
	1   libobjc.A.dylib                     0x0000000106771141 objc_exception_throw + 48
	2   CoreFoundation                      0x0000000106d75625 +[NSException raise:format:] + 197
	3   Foundation                          0x0000000105d739d3 NSKVODeallocate + 294
	4   UIKit                               0x0000000104868dc3 -[UIButton .cxx_destruct] + 126
	5   libobjc.A.dylib                     0x000000010676f9bc _ZL27object_cxxDestructFromClassP11objc_objectP10objc_class + 127
	6   libobjc.A.dylib                     0x000000010677bd34 objc_destructInstance + 129
	7   libobjc.A.dylib                     0x000000010677bd66 object_dispose + 22
	8   UIKit                               0x00000001046886d1 -[UIResponder dealloc] + 145
	9   UIKit                               0x000000011ece2b7e -[UIResponderAccessibility dealloc] + 55
	10  UIKit                               0x0000000104507fc8 -[UIView dealloc] + 1874
	11  UIKit                               0x00000001045f7957 -[UIControl dealloc] + 62
	12  UIKit                               0x000000010485b715 -[UIButton dealloc] + 89
	13  Foundation                          0x0000000105d7390d NSKVODeallocate + 96
	14  libswiftCore.dylib                  0x0000000107d5193d _ZN5swift12metadataimpl14ValueWitnessesINS0_20OpaqueExistentialBoxILj0EEEE12destroyArrayEPNS_11OpaqueValueEmPKNS_14TargetMetadataINS_9InProcessEEE + 45
	15  libswiftCore.dylib                  0x0000000107b6b18c _TFCs23_ContiguousArrayStorageD + 92
	16  libswiftCore.dylib                  0x0000000107c4c25e _TFCs21_SwiftDeferredNSArrayD + 110
	17  libobjc.A.dylib                     0x0000000106780723 _object_remove_assocations + 364
	18  libobjc.A.dylib                     0x000000010677bd40 objc_destructInstance + 141
	19  libobjc.A.dylib                     0x000000010677bd66 object_dispose + 22
	20  UIKit                               0x00000001046886d1 -[UIResponder dealloc] + 145
	21  UIKit                               0x000000011ece2b7e -[UIResponderAccessibility dealloc] + 55
	22  UIKit                               0x0000000104507fc8 -[UIView dealloc] + 1874
	23  UIKit                               0x0000000104dde325 -[UICollectionViewCell dealloc] + 106
	24  CoreFoundation                      0x0000000106d632ad common_removeAllObjects + 189
	25  CoreFoundation                      0x0000000106c443d3 -[__NSArrayM dealloc] + 19
	26  libobjc.A.dylib                     0x0000000106785b8e _ZN11objc_object17sidetable_releaseEb + 202
	27  UIKit                               0x0000000104da738a -[UICollectionView _pruneCachedPrefetchViewsForCacheValidationBounds:] + 405
	28  UIKit                               0x0000000104da71ec -[UICollectionView _pruneCachedPrefetchViewsForVelocity:] + 195
	29  UIKit                               0x0000000104daf56a -[UICollectionView layoutSubviews] + 649
	30  UIKit                               0x000000010453620b -[UIView(CALayerDelegate) layoutSublayersOfLayer:] + 1268
	31  QuartzCore                          0x00000001031d8904 -[CALayer layoutSublayers] + 146
	32  QuartzCore                          0x00000001031cc526 _ZN2CA5Layer16layout_if_neededEPNS_11TransactionE + 370
	33  QuartzCore                          0x00000001031cc3a0 _ZN2CA5Layer28layout_and_display_if_neededEPNS_11TransactionE + 24
	34  QuartzCore                          0x000000010315be92 _ZN2CA7Context18commit_transactionEPNS_11TransactionE + 294
	35  QuartzCore                          0x0000000103188130 _ZN2CA11Transaction6commitEv + 468
	36  QuartzCore                          0x00000001030e56cf _ZN2CA7Display11DisplayLink14dispatch_itemsEyyy + 601
	37  CoreFoundation                      0x0000000106c9fb61 __CFMachPortPerform + 161
	38  CoreFoundation                      0x0000000106c9faa9 __CFRUNLOOP_IS_CALLING_OUT_TO_A_SOURCE1_PERFORM_FUNCTION__ + 41
	39  CoreFoundation                      0x0000000106c9fa21 __CFRunLoopDoSource1 + 465
	40  CoreFoundation                      0x0000000106c97ba0 __CFRunLoopRun + 2352
	41  CoreFoundation                      0x0000000106c97016 CFRunLoopRunSpecific + 406
	42  GraphicsServices                    0x000000010c92ba24 GSEventRunModal + 62
	43  UIKit                               0x00000001044730d4 UIApplicationMain + 159
	44  Wikipedia                           0x00000001025bf419 main + 105
	45  libdyld.dylib                       0x000000010a40b65d start + 1
	46  ???                                 0x0000000000000003 0x0 + 3
)
</pre>

Sorry I don't have better repro-steps :(

joewalsh:
@montehurd fixed the crash 384b055, the spacing issue at the bottom of the cards fe81526, and the scroll issue (was related to the search changes, not these changes) e75ed7b

Will continue to test and look at the other suggestions tomorrow.

amrox:
iPhone 4S is borken

screenshot 2017-05-16 10 00 30

amrox:
oh it's broken on develop too, so separate issue

WMF Framework/UILabel+HasText.swift
joewalsh: Yeah since that API is in flux, the labels are using NSStrings anyway, and this is faster than `characters.count`

Wikipedia/Code/ArticleCollectionViewCell.swift
joewalsh: Added assertions when `DEBUG` preprocessor macro is set

PR #1428, opened 2017-05-12
Test for beta changes to mobileview html output.

montehurd:
https://phabricator.wikimedia.org/T165207

This is to try to give us early warning for changes to mobileview section html output which have been staged to the beta cluster. That way we can examine such changes as soon as they land on beta instead of when they land in production. Hopefully we can avoid issues like this: https://phabricator.wikimedia.org/T165115

<strike>I added a pretty giant comment to the test failure message. Open to moving this to comments in the test itself if anyone prefers.</strike> _I moved it_ :)


montehurd:
BearND had some [suggestions](https://phabricator.wikimedia.org/T165207#3264854) I'd like to look into before we consider merging this.

montehurd:
Closing this until I have time to fix.

PR #1437, opened 2017-05-17
Decrease search complexity on Places

amrox:
https://phabricator.wikimedia.org/T163860
---

The default result now gets elevated to the first position in the list:
screenshot 2017-05-22 13 29 53

Did you mean button:
screenshot 2017-05-22 13 31 52

---

 - [x] Don't allow empty searches
 - [x] Disable keyboard searches while suggestions are being loaded
 - User types "Pa", results are loaded
     - [x] taps keyboard Search button, top result "Pakistan" is loaded
         - This was existing behavior, no changes necessary
     - [x] taps "Top read matching 'Pa'", top read articles in current area are loaded
         - This was existing behavior, no changes necessary
 - [x] User types "Paris", taps "Top read matching 'Paris'", no articles in current area, button appears with "Did you mean *_Paris_*"
 - [x] When user taps on the 'x' in the search field to clear their search and then closes the search overlay their search should be cleared (their previous search should NOT persist) and the default search should be re-enabled (eg. 'Top read' with no search string)
   - This was existing behavior, no changes necessary
 - [x] Add a progress bar below the search field and above the search results to show users that a search is in progress.
 - [x] Move the top suggestion to the top of the search list once the search finishes


Wikipedia/Code/PlacesViewController.swift
```diff
@@ -661,6 +672,15 @@ class PlacesViewController: UIViewController, MKMapViewDelegate, UISearchBarDele
                 
                 guard result.error == nil else {
                     WMFAlertManager.sharedInstance.showWarningAlert(result.error!.localizedDescription, sticky: false, dismissPreviousAlerts: true, tapCallBack: nil)
+                    
+                    let error: Error = result.error!
```
joewalsh: Could this be a guard let rather than a force unwrap?

joewalsh: I see that this is in the "else" condition of `== nil` but maybe instead it's just `if let error = ` with an early return inside the if

PR #1442, opened 2017-05-18
Don't import translation strings with mismatched tokens, fix Explore & Places tab bar localized titles

joewalsh:


WikipediaUnitTests/Code/TWNStringsTests.m
```diff
@@ -225,6 +226,29 @@ - (void)testiOSTranslationStringForBracketSubstitutions {
                         XCTAssertTrue(false, @"Unsupported {{ }} in localization");
```
montehurd: Could the assert string report which localization failed?

WikipediaUnitTests/Code/TWNStringsTests.m
```diff
@@ -209,7 +208,9 @@ - (void)testIncomingTranslationStringForBracketSubstitutions {
     }
 }
 
-- (void)testiOSTranslationStringForBracketSubstitutions {
+- (void)testiOSTranslationStringForBracketSubstitutionsAndMismatchedTokens {
```
montehurd: Thoughts on maybe keeping this as two separate tests and possibly a utility method for any common bits?

joewalsh: The failures are logged separately, so that's enough for me.

montehurd: Ya, was thinking more in terms of keeping tests "unit"y if it wasn't a hassle.

PR #1451, opened 2017-05-22
Fix search bar buttons being truncated to two characters.

montehurd:
https://phabricator.wikimedia.org/T166084

This reverts commit d7d67bb7c50c3300ea929e3bc8179659ed1898ba.

Xcode bug? Got no warning removing these classes even though they are referenced in a storyboard:

screen shot 2017-05-22 at 1 56 54 pm


joewalsh:
screen shot 2017-05-22 at 5 05 42 pm

I don't think it's an Xcode bug, it never has validated the classes in xibs. However, before removing, it's good to do a full workspace search for the class name.

montehurd:
@joewalsh I'm guessing I did a search in .swift and .m and instead of the full workspace :/

PR #1453, opened 2017-05-23
Keep 'Continue reading' widget up to date after 'Back' button tap.

montehurd:
https://phabricator.wikimedia.org/T162507

- Switched `Continue reading` to key off of `wmf_openArticleURL` instead of history `mostRecentEntry` - per our preference to fix this [without re-ordering history](https://phabricator.wikimedia.org/T162507#3190854).
- Ensured `wmf_openArticleURL` is also updated by `WMFArticleViewController viewDidAppear`.

zone mov


ContinueReadingWidget/WMFTodayContinueReadingWidgetViewController.swift
```diff
@@ -86,19 +86,10 @@ class WMFTodayContinueReadingWidgetViewController: UIViewController, NCWidgetPro
     }
     
     func hasNewData() -> Bool{
-        
-        guard let session = SessionSingleton.sharedInstance() else {
-            return false
-        }
-        
-        guard let historyEntry = session.dataStore.historyList.mostRecentEntry() else {
+        guard let openArticleURL = UserDefaults.wmf_userDefaults().wmf_openArticleURL() else {
```
joewalsh: Ultimately, I'd like to move this out of user defaults and into a `WMFKeyValue`. However, since this property exists in `userDefaults` already, it's fine to keep for now.

PR #1457, opened 2017-05-23
Fix initial places animation

joewalsh:
Wikidata query is only necessary if region is nil and the search result isn't nil


joewalsh:
@amrox hold off on merging, found a bug

PR #1464, opened 2017-05-25
Rough attempt to hook up custom "places" accessibility rotor.

montehurd:
This isn't working quite right...

To see it sorta-work (it's easiest to test on device and easier yet to test on iPad):
- turn on Voice Over (I set up a home button triple-click shortcut for Voice Over to make toggling it on and off easier)
- load places
- tap the map
- do 2 finger twist gesture to show rotor, repeat twist until you see  "Places" rotor
- now that places rotor is in effect do one finger swipe up or down to move from place to place

Issues:
- order isn't respecting order from the filter table
- after swiping up or down a few times it gets somehow knocked out of the "Places" mode (something is flakey)
- isn't moving through entire list correctly (swipe up / down for previous / next only seems to move through a couple of the onscreen places)

-----

VoiceOver gesture list:
https://developer.apple.com/library/content/technotes/TestingAccessibilityOfiOSApps/TestAccessibilityonYourDevicewithVoiceOver/TestAccessibilityonYourDevicewithVoiceOver.html

High-level rotor doc:
https://support.apple.com/en-us/HT204783

Custom rotor in WWDC video:
https://developer.apple.com/videos/play/wwdc2016/202/


montehurd:
Keep in mind the variable naming is very WIP - same for using `guard` etc...

Wikipedia/Code/PlacesViewController.swift
```diff
@@ -89,6 +89,57 @@ class PlacesViewController: UIViewController, MKMapViewDelegate, UISearchBarDele
         return PlaceSearchService(dataStore: self.dataStore)
     }()
     
+    // MARK: - Accessibility
+    
+    func configurePlacesAccessibilityRotor() {
+        if #available(iOS 10.0, *) {
+            let favoritesRotor = UIAccessibilityCustomRotor(name: "Places") { predicate in
+                
+                let allArticlePlaceAnnotations = Array(self.mapView.annotations(in: self.mapView.visibleMapRect)).filter { $0 is ArticlePlace } as! [ArticlePlace]
```
joewalsh: You could fix `order isn't respecting order from the filter table` by using the `articleFetchedResultsController.fetchedObjects` - however this would also include items that are offscreen.

joewalsh: Add `placesSortOrder` to `ArticlePlace` and apply set it to the same value as the associated `WMFArticle` - then sort this array.

PR #1467, opened 2017-05-26
Update location detail view

joewalsh:
simulator screen shot may 26 2017 12 27 39 pm
simulator screen shot may 26 2017 12 27 36 pm

https://phabricator.wikimedia.org/T166241


amrox:
Looks reasonable, but can we get a description, screenshot or ticket? :)

joewalsh:
@amrox ticket was in the branch name (same as other detail view updates), but added link & screenshots

amrox:
Ah, didn’t notice the branch name, sorry

montehurd:
@joewalsh For the `international federation` results, in the first screenshot the title is 3 lines and there's no wikidata description, but in the second screenshot the title is 2 lines with a description... is that difference intentional?

joewalsh:
@montehurd that's autolayout preferring to use the space for the title label. Same thing happens in the explore feed. There may be updated designs for these cells at some point, at which time we could clarify expected behavior of the cells.

PR #1484, opened 2017-05-31
Analytics Revamp

amrox:
Features:
  - Events are assuming to be of type `Dictionary<String, Any>`
  - All events are stored in CoreData under and `EventRecord`, which has the fields
    - event: Transformable
      - the event dictionary
    - recorded: Date
      - when the event was recorded on the device
    - posted: Date
      - when the event was successfully posted to the server
    - postAttempts: Int16
       - incremented each time a post is attempted. Currently not used, but intended to prune events after so many retries
    - failed: Boolean
      - if the event failed in a fatal way, like a JSON serialization error
      - *not* if it failed due to connection error
  - `logEvent` is meant to be non-blocking and thread-safe
  - monitors reachability to the logging host, specifically
  - attempts to post events in batches of `postBatchSize` (default 10)
    - if all events in batch succeed, attempts to post more
  - Reachability and Posting
    - event posting is paused if host is not reachable
    - if any reachability, events are posted on a timer (`postInterval` default 10m)
    - events are always posted immediately on WiFi
    - on WWAN, events are only posted immediately if there was a recent network request (`sendImmediatelyOnWWANThreshhold`, default 30s), or the periodic timer fires

Items of note:
 - Added a notification in WMFURLCache for when a real network request is going to happen
     - Originally I started down the path of subleasing NSURLSession and AFHTTPSessionManager, but that was cumbersome and not as accurate anyway
 - I *think* I got the CoreData background thread-ness correct, but please check


montehurd:
I merged in develop to kick off jenkins on the build server - sorry I prob should have just opened a PR to test. Anyway tests are running fine - looks like the files are just in wrong folder (the `CodeFileLocationTests.test_code_files_are_not_in_root_folder` test failed).

Wikipedia/Code/WMFURLCache.m
```diff
@@ -28,9 +28,16 @@ - (NSCachedURLResponse *)cachedResponseForRequest:(NSURLRequest *)request {
         }
     }
     
+    dispatch_block_t done = ^{
+        if (!response) {
+            wmf_postNetworkRequestBeganNotification(request.HTTPMethod, request.URL.absoluteString);
```
amrox: @joewalsh  What do you think about adding this notification here?

joewalsh: looks good to me - maybe have the response passed to the block instead of a captured var. Does this capture the value when done is assigned? I thought it would require `__block` to be updated after the block creation.

amrox:
!jenkins test this please

montehurd:
Quick questions - are the old event logging funnel bits going to be removed in separate ticket? Same with wiring up to `shouldSendUsageReports`.

amrox:
@montehurd yeah because it's complex, but relatively invisible code I wanted to have it easy to turn back to the old way. I'll probably refactor more of the bits out while preserving as much of the current interface as I can, but I think this a good first step

PR #1485, opened 2017-05-31
Speed up swift compilation for functions that are slowest to compile

joewalsh:


montehurd:
Weird that those are taking so long... would be interesting to circle back after next Xcode update and see how they do.

PR #1494, opened 2017-06-01
[T166819] Add ranks to top read

joewalsh:
https://phabricator.wikimedia.org/T166819


Wikipedia/Code/ArticleCollectionViewCell+WMFFeedContentDisplaying.swift
```diff
@@ -84,3 +84,28 @@ public extension ArticleCollectionViewCell {
...
+        var r2: CGFloat = 0
+        var g2: CGFloat = 0
+        var b2: CGFloat = 0
+        var a2: CGFloat = 0
+        endColor.getRed(&r2, green: &g2, blue: &b2, alpha: &a2)
+        let percent = CGFloat(index + 1) / CGFloat(count)
+        let r = r1 + percent * (r2 - r1)
+        let g = g1 + percent * (g2 - g1)
+        let b = b1 + percent * (b2 - b1)
+        let a = a1 + percent * (a2 - a1)
```
montehurd: I think this line can be `let a: CGFloat = 1`

montehurd:
Only issue I noticed is the margin is still present in the detail view when RTL:
screen shot 2017-06-01 at 3 51 46 pm

Wikipedia/Code/ArticleCollectionViewCell+WMFFeedContentDisplaying.swift
joewalsh: Alpha is expected to always be one in this case. I wrote the code to be generalizable (because it could be re-usable elsewhere where you might want a gradient with an alpha change).

joewalsh:
Merging to clean up layout changes in current tickets - some changes here affect overall layout

PR #1502, opened 2017-06-05
[T166898] Add target with OSM map integration via Mapbox

joewalsh:
https://phabricator.wikimedia.org/T166898

The `Wikipedia OSM` target integrates Mapbox framework with workaround for using WMF tile server.

The `OSM` preprocessor macro controls whether Mapbox or MapKit is used.

To run the OSM variant:

1. Select the `Wikipedia OSM` build scheme
screen shot 2017-06-05 at 3 45 07 pm

2. Build & run (the first build will take awhile as it compiles the carthage dependencies)
3. There's no step 3


joewalsh:
retest this please

montehurd:
Wow nice! 

...looks like we'll need to flip some bit to tell map box to show bodies of water - the coast blends with the land atm.

joewalsh:
@montehurd the water issue is a result of not having a fully defined style that matches WMF's tileserver vector format. If you comment out the entire `func mapView(_ mapView: MGLMapView, didFinishLoading style: MGLStyle) {` function, it'll use Mapbox's tile server and styles.

montehurd:
Ah gotcha. Makes sense.

pnorman:
Water information is included in the `water` layer: https://github.com/kartotherian/osm-bright.tm2source/blob/a9dd06ed15ac62f079c02ee796dc168f765e4baf/data.yml#L123-L166

PR #1506, opened 2017-06-06
[T167142] Fall back on most recent history entry for continue reading widget

joewalsh:
https://phabricator.wikimedia.org/T167142


montehurd:
Oh good catch!!

PR #1507, opened 2017-06-06
[T167146] Fix searches for Antartica (among other things)

joewalsh:
- Fix region calculations to be safe to pass to MKMapView
- Incorporate fixed small dot asset (scale animation was off)
before:
jun-06-2017 16-27-42
after:
jun-06-2017 17-00-15

When showing a map from an article that isn't in your search language (or if the article just wasn't in the search results) the article wouldn't show up. This is fixed by:
- Add siteURL to PlaceSearch
- Add the result to the search results if it isn't included

There will be duplicates in search history since the keys were updated. This is expected.


montehurd:
Looks good. Only thing to maybe consider tweaking would some the latitude/longitude delta logic - seems the diff had 3 or 4 places where a category method on CLLocationDegrees (Double) could DRY things a bit.

montehurd:
iirc you can do that in Swift... anyways, looks good.

montehurd:
Oh, one thing - looks like the old `places-dot-small.pdf` is still hanging around. Did you mean for the updated small dot asset to replace it? It just tacked on `(3)` to the name instead of replacing the old one.

PR #1510, opened 2017-06-07
Use same fonts/colors on "Enable locations" as we do on cards

nambatee:
https://phabricator.wikimedia.org/T164890

Happy to adjust this/create followup tickets if anything's off 🤔


Wikipedia/Code/WMFExploreCollectionViewController.m
```diff
@@ -1081,12 +1081,13 @@ - (nonnull UICollectionReusableView *)collectionView:(UICollectionView *)collect
             return [collectionView dequeueReusableSupplementaryViewOfKind:UICollectionElementKindSectionFooter withReuseIdentifier:WMFFeedEmptyHeaderFooterReuseIdentifier forIndexPath:indexPath];
         case WMFFeedMoreTypeLocationAuthorization: {
             WMFTitledExploreSectionFooter *footer = (id)[collectionView dequeueReusableSupplementaryViewOfKind:UICollectionElementKindSectionFooter withReuseIdentifier:[WMFTitledExploreSectionFooter wmf_nibName] forIndexPath:indexPath];
-
-            for (UIGestureRecognizer *gr in footer.gestureRecognizers) {
-                [footer removeGestureRecognizer:gr];
+            
+            for (UIGestureRecognizer *gr in footer.enableLocationButton.gestureRecognizers) {
```
joewalsh: Generally the tap recognizers are for adding gesture recognition to plain views that aren't subclasses of `UIControl` (like `UIButton`). With controls, you can programatically add target objects that respond to the actions of the control - `addTarget:action:forControlEvents:` or connect them with Interface Builder via an `IBAction`.

In this case, it's probably easiest to remove & add the target, similar to how the header button works (line 981ish):

```
        [header.rightButton removeTarget:self action:@selector(headerRightButtonPressed:) forControlEvents:UIControlEventTouchUpInside]; // ensures the view controller isn't duplicated in the target list, causing duplicate actions to be sent
        header.rightButton.tag = indexPath.section;
        [header.rightButton addTarget:self action:@selector(headerRightButtonPressed:) forControlEvents:UIControlEventTouchUpInside];
```

Alternatively, you could add the `IBAction` to the view class, and a delegate protocol & property that could be set up to relay actions back to the view controller (Similar to how `WMFFindInPageKeyboardBar` works). However, since this is only one button and action, the method above is probably better.

Wikipedia/Code/WMFTitledExploreSectionFooter.xib
```diff
@@ -1,57 +1,71 @@
...
-                    <rect key="frame" x="50" y="16" width="312" height="42.5"/>
-                    <fontDescription key="fontDescription" style="UICTFontTextStyleHeadline"/>
-                    <color key="textColor" red="0.20000000000000001" green="0.40000000000000002" blue="0.80000000000000004" alpha="1" colorSpace="calibratedRGB"/>
+                <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" verticalHuggingPriority="251" verticalCompressionResistancePriority="801" ambiguous="YES" text="Explore articles near your current location" textAlignment="center" lineBreakMode="tailTruncation" numberOfLines="0" baselineAdjustment="alignBaselines" adjustsFontSizeToFit="NO" translatesAutoresizingMaskIntoConstraints="NO" id="Jze-hP-209">
+                    <rect key="frame" x="14" y="10" width="349" height="19.5"/>
+                    <fontDescription key="fontDescription" type="system" pointSize="14"/>
```
joewalsh: A system font of fixed size won't be resized for dynamic type, so this should be the font style that matches (or closely matches) the target style - in this case `Subheadline`. The other fonts in this xib should also be updated.

Wikipedia/Code/WMFTitledExploreSectionFooter.xib
```diff
@@ -1,57 +1,71 @@
...
-                    <rect key="frame" x="50" y="16" width="312" height="42.5"/>
-                    <fontDescription key="fontDescription" style="UICTFontTextStyleHeadline"/>
-                    <color key="textColor" red="0.20000000000000001" green="0.40000000000000002" blue="0.80000000000000004" alpha="1" colorSpace="calibratedRGB"/>
+                <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" verticalHuggingPriority="251" verticalCompressionResistancePriority="801" ambiguous="YES" text="Explore articles near your current location" textAlignment="center" lineBreakMode="tailTruncation" numberOfLines="0" baselineAdjustment="alignBaselines" adjustsFontSizeToFit="NO" translatesAutoresizingMaskIntoConstraints="NO" id="Jze-hP-209">
```
joewalsh: Ambiguous location should be resolved

joewalsh:
jenkins, ok to test

joewalsh:
Looks great!

PR #1514, opened 2017-06-07
[T167308] Add translations to enable location explore feed card & add missing wmf_configureSubviewsForDynamicType

joewalsh:
- Adds the same translations used in places
- Adds missing `wmf_configureSubviewsForDynamicType`
- Uses the ability to override extension functions to eliminate the need for the type check in `wmf_configureSubviewsForDynamicType`

simulator screen shot jun 7 2017 11 20 36 am

### DIFF
--- a/Wikipedia/Code/EnableLocationViewController.swift
+++ b/Wikipedia/Code/EnableLocationViewController.swift
@@ -3,8 +3,12 @@ import UIKit
 protocol EnableLocationViewControllerDelegate: NSObjectProtocol {
     func enableLocationViewController(_ enableLocationViewController: EnableLocationViewController, didFinishWithShouldPromptForLocationAccess  shouldPromptForLocationAccess: Bool)
 }
-class EnableLocationViewController: UIViewController {
 
+class EnableLocationViewController: UIViewController {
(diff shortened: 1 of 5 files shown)


montehurd:
hmm that "South Wales" title and description looks like it's coming from enwiki...

joewalsh:
@montehurd yes, the content is from enwiki but the system language is French

joewalsh:
simulator screen shot jun 7 2017 1 18 48 pm

^ same screen but with French feed content

montehurd:
> @montehurd yes, the content is from enwiki but the system language is French

oh so system lang was FR, but first search lang was EN?

joewalsh:
yes

Wikipedia/Code/EnableLocationViewController.swift
```diff
@@ -3,8 +3,12 @@ import UIKit
 protocol EnableLocationViewControllerDelegate: NSObjectProtocol {
     func enableLocationViewController(_ enableLocationViewController: EnableLocationViewController, didFinishWithShouldPromptForLocationAccess  shouldPromptForLocationAccess: Bool)
 }
-class EnableLocationViewController: UIViewController {
 
+class EnableLocationViewController: UIViewController {
+    static let localizedEnableLocationTitle = WMFLocalizedString("places-enable-location-title", value:"Explore articles near your location by enabling Location Access", comment:"Explains that you can explore articles near you by enabling location access. \"Location\" should be the same term, which is used in the device settings, under \"Privacy\".")
```
montehurd: > \"Location\" should be the same term, which is used in the device settings

hehe guessing most translators aren't iOS users ;)

PR #1516, opened 2017-06-07
Use style JSON rather than trying to adapt a Mapbox style

joewalsh:


Wikipedia/Code/mapstyle.json
```diff
@@ -0,0 +1,30 @@
+{
```
montehurd: Nice!!!

montehurd:
I can see water! :)

PR #1520, opened 2017-06-08
Fix missing `Similar pages` footer menu items.

montehurd:
The upstream css changed adding an additional class and we were testing for an extact match.

https://phabricator.wikimedia.org/T167476

# Before
before1 mov

# After
after2 mov

### DIFF
--- a/Wikipedia/Code/MWKSection.m
+++ b/Wikipedia/Code/MWKSection.m
@@ -244,7 +244,7 @@ static NSString *const WMFSectionSummaryXPathSelector = @"\
     }] componentsJoinedByString:@" "] wmf_summaryFromText];
 }
 
-static NSString *const WMFSectionDisambiguationTitlesXPathSelector = @"//div[@class='hatnote']//a/@href";
+static NSString *const WMFSectionDisambiguationTitlesXPathSelector = @"//div[contains(@class, 'hatnote')]//a/@href";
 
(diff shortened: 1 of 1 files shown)


Wikipedia/Code/MWKSection.m
```diff
@@ -244,7 +244,7 @@ - (nullable NSString *)summary {
     }] componentsJoinedByString:@" "] wmf_summaryFromText];
 }
 
-static NSString *const WMFSectionDisambiguationTitlesXPathSelector = @"//div[@class='hatnote']//a/@href";
+static NSString *const WMFSectionDisambiguationTitlesXPathSelector = @"//div[contains(@class, 'hatnote')]//a/@href";
 
 - (nullable NSArray<NSURL *> *)disambiguationURLs {
```
montehurd: Note: thinking we should use a bit of standard JS to mine these disambiguation URLs (and the page issues strings) from the web view's *existing* DOM instead of reconstructing a DOM via Hpple/XPath here. Android has such JS functions and these should be moved to the page-library soon-ish.

montehurd: iirc `disambiguationURLs` and `pageIssues` are only used by the footer menu transform anyway (they do hand-off to native land when tapped, but shouldn't make much difference if they figure out how to get the disambig urls & page-issues strings)

PR #1529, opened 2017-06-10
[Need review] Adding GFDL

josve05a:
Since they (should) release it under GFDL as well, we need to mention this, which the desktop and mobileWeb do, but not they iOS app.

Please check all edits, and please help add the proper strings were missing, I just used the search bar on Github to find relevant documents to add GFDL to, but I do not understand how they all connect.

https://phabricator.wikimedia.org/T167561

PR #1530, opened 2017-06-12
Updated #1529 with requested changes

joewalsh:


montehurd:
There appears to be a bug when tapping the `CC BY-SA 3.0` link:

wot mov

josve05a:
https://phabricator.wikimedia.org/T167561

PR #1531, opened 2017-06-12
Design tweaks for 'Use same fonts/colors on "Enable locations" as we do on cards'

nambatee:
https://phabricator.wikimedia.org/T164890


Wikipedia/Code/WMFExploreCollectionViewController.m
```diff
@@ -1080,9 +1080,6 @@ - (nonnull UICollectionReusableView *)collectionView:(UICollectionView *)collect
             return [collectionView dequeueReusableSupplementaryViewOfKind:UICollectionElementKindSectionFooter withReuseIdentifier:WMFFeedEmptyHeaderFooterReuseIdentifier forIndexPath:indexPath];
         case WMFFeedMoreTypeLocationAuthorization: {
             WMFTitledExploreSectionFooter *footer = (id)[collectionView dequeueReusableSupplementaryViewOfKind:UICollectionElementKindSectionFooter withReuseIdentifier:[WMFTitledExploreSectionFooter wmf_nibName] forIndexPath:indexPath];
-            footer.titleLabel.text = [EnableLocationViewController localizedEnableLocationTitle];
-            footer.descriptionLabel.text = [EnableLocationViewController localizedEnableLocationDescription];
-            [footer.enableLocationButton setTitle:[EnableLocationViewController localizedEnableLocationButtonTitle] forState: UIControlStateNormal];
```
joewalsh: What was the reason for removing these lines? The text won't be localized without them.

nambatee: My bad! I removed it cause It is overriding the "Explore articles near your current location" label to be "Explore articles near your location by enabling Location Access". I can add a new WMFLocalizedString with "Explore articles near your current location" and set `footer.titleLabel.text` to that (and obviously revert the rest). Does that sound like a good solution or can you think of a better one?

joewalsh: @nambatee that sounds good - I messed that up in my PR, I didn't realize it was just "Explore articles near your location" instead of "Explore articles near your location by enabling Location Access"

montehurd: I think [Carolyn](https://phabricator.wikimedia.org/T164890#3341991) wanted the title to be "Explore articles near your current location", which is set in the storyboard, so deleting these lines make that storyboard string appear. So yeah we'll need new localized string for "Explore articles near your current location".

montehurd: lol never mind you guys resolved while i was typing :)

joewalsh:
ok to test

joewalsh:
Thanks!

PR #1532, opened 2017-06-12
[Contains #1526] Fix linting

joewalsh:
Contains #1526 - Merge that first


montehurd:
btw our `WMF_HexColor` methods are imo much better than the 3rd party `HexColors.m`... would totally be in favor of removing `HexColors.m` if we can.

joewalsh:
@montehurd `TSMessage` relies on `HexColors`, but probably could be updated to use our category

PR #1533, opened 2017-06-12
Add a check on article's displayTitle to avoid adding it to History i…

nambatee:
…f article doesn't exist

https://phabricator.wikimedia.org/T167339


Wikipedia/Code/MWKHistoryList.m
```diff
@@ -154,6 +154,11 @@ - (void)addPageToHistoryWithURL:(NSURL *)URL {
     NSDate *now = [NSDate date];
 
     WMFArticle *article = [self.dataStore fetchOrCreateArticleWithURL:URL];
+    
+    if (!article.displayTitle) {
```
joewalsh: This check will skip history for any article opened via a universal link from Safari that hasn't been viewed in the app yet. The data store will only have a `WMFArticle` object with a display title if the article appeared in the feed or search results.

As a result, with this change, if you do a clean install of the app, exit the app, enter an article URL into Safari that isn't in the feed (I tested with https://en.wikipedia.org/wiki/Barack_Obama), then tap "Open" on the banner on the top of the page, it won't show up in history.

joewalsh: I opened #1535 with a change that fixes this issue.

joewalsh:
ok to test

joewalsh:
Merging. Tests fixed in #1535

PR #1542, opened 2017-06-14
[5.5] Fix invalid quad key crashes, iOS 9 places crashes

joewalsh:
https://phabricator.wikimedia.org/T167884
https://phabricator.wikimedia.org/T167889

### DIFF
--- a/Wikipedia/Code/QuadKey.swift
+++ b/Wikipedia/Code/QuadKey.swift
@@ -229,6 +229,22 @@ public extension QuadKey {
         }
         return string
     }
+    
+    public func coordinate(precision: QuadKeyPrecision) -> QuadKeyCoordinate {
+        return QuadKeyCoordinate(quadKey: self, precision: precision)
+    }
+    
+    public var coordinate: QuadKeyCoordinate {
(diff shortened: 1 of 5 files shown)


Wikipedia/Code/QuadKey.swift
```diff
@@ -96,18 +98,30 @@ public extension QuadKeyDegrees {
     public static func max(_ a: QuadKeyDegrees, _ b: QuadKeyDegrees) -> QuadKeyDegrees {
         return a > b ? a : b
     }
+    
+    public static func min(_ a: QuadKeyDegrees, _ b: QuadKeyDegrees) -> QuadKeyDegrees {
+        return a < b ? a : b
```
montehurd: Any possible equality issues? Guessing not, but looks like both methods always return `b` in that case.

joewalsh: If a == b, it doesn't matter if it returns a or b

montehurd: lol that's what i get for trying to think pre-coffee

PR #1545, opened 2017-06-14
On this day!

montehurd:
https://phabricator.wikimedia.org/T135769

- [x] fetch  
- [x] model
- [x] store
- [x] hook up to feed
- [x] event tap
- [x] event article tap
- [x] event peek
- [x] event article peek
- [x] detail vc
- [x] detail view header
- [x] format display of `eventYear` to use localized `BC|AD`
- [x] format display of `yearsSinceEvent` so number formatting is localized
- [x] timeline
- [x] side-to-side article scroll covers timeline
- [x] feed icon
- [x] timeline year dots
- [x] timeline year dot animation on scroll
- [x] enable year dots on explore card
- [x] show year of next event at bottom of explore card
- [ ] style (WIP)
- [ ] timeline clip at top of first card and end of last card on detail view
- [ ] timeline fade out at bottom and top of explore card
- [ ] enable only for langs supported by endpoint
- [ ] fix all the `TODO|FIX` comments in the WIP diff
- [ ] fix endpoint `BC` off-by-one bug (no year zero)
- [ ] iOS 9
- [ ] rtl
- [ ] dynamic type 
- [ ] should we not create event if for some reason there are no pages associated with the event? could then revert [this commit](https://github.com/wikimedia/wikipedia-ios/pull/1545/commits/91d94f6c8dff28b600b691f508b44107b1c2fb26)

# So far...

ccccc mov


WMF Framework/WMFContentGroup+Extensions.m
```diff
@@ -82,6 +85,7 @@ - (void)updateContentType {
             self.contentType = WMFContentTypeTopReadPreview;
             break;
         case WMFContentGroupKindNews:
+        case WMFContentGroupKindOnThisDay:
```
joewalsh: Would be good to make this a separate content type since it's a different content model object

WMF Framework/WMFOnThisDayEventsFetcher.m
```diff
@@ -0,0 +1,65 @@
+#import "WMFOnThisDayEventsFetcher.h"
+#import "WMFFeedOnThisDayEvent.h"
+#import <WMF/WMF-Swift.h>
```
joewalsh: The new syntax for this is `@import  WMF.Swift;` - apparently the old syntax is equivalent to this anyway, but trying to move everything in the app to the new syntax.

Wikipedia/Code/WMFExploreCollectionViewController.m
```diff
@@ -227,6 +227,10 @@ - (nullable WMFContentGroup *)sectionForIndexPath:(NSIndexPath *)indexPath {
         content = [content wmf_map:^id(WMFFeedNewsStory *obj) {
             return [[obj featuredArticlePreview] articleURL] ?: [[[obj articlePreviews] firstObject] articleURL];
         }];
+    } else if ([group contentType] == WMFContentTypeOnThisDayEvent) {
+        content = [content wmf_map:^id(WMFFeedOnThisDayEvent *obj) {
+            return [[[obj articlePreviews] firstObject] articleURL];
```
montehurd: @joewalsh was a little foggy about whether this line and [L281](https://github.com/wikimedia/wikipedia-ios/pull/1545/commits/0261e7baac7d509eedfdddb711943ea797316922#diff-075df5f5584af1efc1a6c2e7138578d5R281) and [L332](https://github.com/wikimedia/wikipedia-ios/pull/1545/commits/0261e7baac7d509eedfdddb711943ea797316922#diff-075df5f5584af1efc1a6c2e7138578d5R332) were right... once things are wired up a bit more it'll probably become apparent, but thought I'd ask in the mean time.

Wikipedia/Code/WMFExploreCollectionViewController.m
```diff
@@ -314,6 +324,13 @@ - (nullable NSURL *)imageURLForIndexPath:(NSIndexPath *)indexPath {
         }
         articleURL = [[content[indexPath.row] featuredArticlePreview] articleURL] ?: [[[content[indexPath.row] articlePreviews] firstObject] articleURL];
         width = self.traitCollection.wmf_nearbyThumbnailWidth;
+    } else if ([section contentType] == WMFContentTypeOnThisDayEvent) {
```
joewalsh: This whole block can be removed since there's no lead image on the card, returning nil is expected for this case.

montehurd: Ah, great thanks!

Wikipedia/Code/WMFExploreCollectionViewController.m
joewalsh: This can be removed (along with the same code for news story). If you look at where this is used in `presentMoreViewControllerForGroup` - the URLs aren't used for presenting the news vc. You'd also have to remove the assertion there and re-arrange that method a bit.

Wikipedia/Code/WMFExploreCollectionViewController.m
```diff
@@ -269,6 +273,12 @@ - (nullable NSURL *)contentURLForIndexPath:(NSIndexPath *)indexPath {
             return nil;
         }
         return [[content[indexPath.row] featuredArticlePreview] articleURL] ?: [[[content[indexPath.row] articlePreviews] firstObject] articleURL];
+    } else if ([section contentType] == WMFContentTypeOnThisDayEvent) {
```
joewalsh: This (and the news equivalent) can probably be removed as well.

joewalsh: You would need to implement the case for on this day in `detailViewControllerForItemAtIndexPath` similar to how news is implemented (where it checks for an index path of more than two elements, section, item, then sub-item, where sub-item is the index of the item in the side scrolling list)

montehurd:
@joewalsh Feed is showing something close to useful now. I added a checklist of remaining items I can start running down next week.

WMF Framework/WMFOnThisDayEventsFetcher.m
montehurd: @joewalsh 
Hmm... for some reason it doesn't compile for me using that syntax in that file...

montehurd:
retest this please

Wikipedia/Code/NewsViewController.swift
```diff
@@ -56,6 +56,9 @@ extension NewsViewController {
             header.label.text = headerTitle(for: indexPath.section)
             return header
         default:
+            
+//FIXME: According to docs looks like this will crash - "The view that is returned must be retrieved from a call to -dequeueReusableSupplementaryViewOfKind:withReuseIdentifier:forIndexPath:"
```
joewalsh: Execution should never reach this point. Could be updated to include an assertion.

montehurd:
> Looks good so far! Let's merge this where it is and make the remaining tasks a follow on PR.

@joewalsh oh awesome that works for me!

montehurd:
Part 2 [here](https://github.com/wikimedia/wikipedia-ios/pull/1577).

PR #1555, opened 2017-06-20
[T168062] add check for days between news date and current date

joewalsh:


Wikipedia/Code/WMFFeedContentSource.m
```diff
@@ -390,18 +390,36 @@ - (void)scheduleNotificationsForFeedDay:(WMFFeedDayResponse *)feedDay onDate:(NS
...
+        NSDate *midnightUTCDate = date.wmf_midnightUTCDateFromLocalDate;
+        NSDateComponents *nowComponents = [utcCalendar components:NSCalendarUnitYear | NSCalendarUnitMonth | NSCalendarUnitDay fromDate:midnightUTCDate];
+        NSDateComponents *newsComponents = [utcCalendar components:NSCalendarUnitYear | NSCalendarUnitMonth | NSCalendarUnitDay fromDate:newsDate];
+        NSInteger year = nowComponents.year;
+        if (nowComponents.month == 1 && newsComponents.month == 12) {
+            year--;
+        }
+        newsComponents.year = year;
+        newsDate = [utcCalendar dateFromComponents:newsComponents];
+        if (newsDate && midnightUTCDate && [utcCalendar wmf_daysFromDate:newsDate toDate:midnightUTCDate] > 1) {
```
montehurd: Would it be ok to pull this new logic into a bool method? Then a year from now when we try to remember why we're calling `done()` we can just read the method signature.

joewalsh: I abstracted this into an extension on calendar. It's not a boolean, but it's clearer.

PR #1556, opened 2017-06-21
[5.5.1] [T168082] Fix "enable notifications" header spacing issues

joewalsh:
- Converts the header to a WMFContentGroup
- Updates style

https://phabricator.wikimedia.org/T168082


montehurd:
This is great! RTL looked good too. Only think I noticed is the text doesn't seem to respond to dynamic type size changes.

PR #1567, opened 2017-06-23
remove leading 0 for days < 10 in date formatter

nambatee:
https://phabricator.wikimedia.org/T168642

Instead of changing the existing `wmf_dayNameMonthNameDayOfMonthNumberDateFormatter`,
another solution would be to add a new `wmf_dayNameMonthNameDayOfMonthNumberWithoutLeadingZeroDateFormatter`to `NSDateFormatter+WMFExtensions` and make all sections of Explore feed in `WMFContentGroup+WMFFeedContentDisplaying`use that.


joewalsh:
ok to test

PR #1569, opened 2017-06-26
Show '9' instead of '09' for dates where day < 10 in Explore feed

waffleboot:
https://phabricator.wikimedia.org/T168642


montehurd:
Thanks for the patch! Look good but someone [beat you to it](https://github.com/wikimedia/wikipedia-ios/pull/1567) :)

PR #1573, opened 2017-06-27
Use the same sort order for main page feed on iPad as iPhone

waffleboot:
https://phabricator.wikimedia.org/T168512


waffleboot:
Looks like my fix doesn't work, Main page again in top right corner on iPad, I need time to check why Core Data fetch results controller returns such resultset.

Wikipedia/Code/WMFMainPageContentSource.m
```diff
@@ -103,6 +104,10 @@ - (void)fetchAndSaveMainPageForSiteURL:(NSURL *)siteURL intoManagedObjectContext
                             WMFContentGroup *section = [moc fetchOrCreateGroupForURL:groupURL ofKind:WMFContentGroupKindMainPage forDate:[NSDate date] withSiteURL:siteURL associatedContent:nil customizationBlock:NULL];
                             [moc fetchOrCreateArticleWithURL:data.mainPageURL updatedWithSearchResult:[results firstObject]];
                             section.content = @[data.mainPageURL];
+                            if (![[NSUserDefaults wmf_userDefaults] wmf_didMigrateMainPageDailySortPriority]) {
```
joewalsh: Could you remove the new user defaults key & just `updateDailySortPriority` on every refresh? This way, if we ever change the sort priority in the future it'll be applied. The object is being updated at this point anyway so it's a low cost op.

waffleboot: Done, I removed new user default settings for Main page feed sort order.

PR #1577, opened 2017-06-27
On this day part 2

montehurd:
https://phabricator.wikimedia.org/T135769

This ticket is follow-up to [this](https://github.com/wikimedia/wikipedia-ios/pull/1545) ticket (which added "On this day").

- [x] timeline clip at top of first card on detail view
screen shot 2017-06-27 at 4 31 04 pm

- [x] timeline fade out at bottom and top of explore card
screen shot 2017-06-27 at 4 31 25 pm

- [x] dynamic type 
- [x] enable only for langs supported by endpoint
- [x] iOS 9
- [x] <strike>should we not create event if for some reason there are no pages associated with the event? could then revert [this commit](https://github.com/wikimedia/wikipedia-ios/pull/1545/commits/91d94f6c8dff28b600b691f508b44107b1c2fb26)</strike>
- [x] database housekeeper also has to be updated to handle the new content type
- [ ] rtl
- [ ] fix all the `TODO|FIX` comments in the WIP diff
- [ ] fix endpoint `BC` off-by-one bug (no year zero)


montehurd:
@joewalsh 

Found the [cause](https://github.com/wikimedia/wikipedia-ios/pull/1577/files#diff-218347b5f64cb41cb142b6b7fa35c181L70) of the slight height miscalculation on rotate.

montehurd:
@joewalsh Hey some quick notes on the rotation issue...

Repo steps:
- scroll to second "On this day" item
- rotate
- the bottom of the bottom label is clipped in half

Interestingly, when I added `[self.collectionView reloadData]` to WMFExploreCollectionViewController's `didReceiveMemoryWarning` then ran the repo steps and triggered the sim memory warning, the issue goes away, so I think the side scrolling cell's `sizeThatFits` is at least ok:

720 mov

I tried adding a call to `[self.collectionView reloadData]` to `traitCollectionDidChange` but it didn't seem to have the same effect...

Any ideas?

joewalsh:
@montehurd rotation issue appears to be fixed by changes I made that fully recalculate the layout on rotation. Our layout subclass would fully update the layout but it seems the collection view wouldn't re-query for the new layout information. I also made this fix so that it can be pulled into 5.5.1 https://github.com/wikimedia/wikipedia-ios/pull/1579. I reverted the frame rounding changes but kept the spacing rounding changes, since the factor of 0.4 frequently led to an odd spacing value.

Wikipedia/Code/OnThisDayCollectionViewCell.swift
```diff
@@ -1,16 +1,41 @@
...
+    @objc(configureWithOnThisDayEvent:previousEvent:dataStore:layoutOnly:)
+    func configure(with onThisDayEvent: WMFFeedOnThisDayEvent,  previousEvent: WMFFeedOnThisDayEvent, dataStore: MWKDataStore, layoutOnly: Bool) {
         bottomTitleLabel.textColor = .wmf_blue
         bottomTitleLabel.text = previousEvent.yearWithEraString()
-        configure(with: onThisDayEvent, dataStore: dataStore, layoutOnly: layoutOnly, shouldAnimateDots: false)
+        super.configure(with: onThisDayEvent, dataStore: dataStore, layoutOnly: layoutOnly, shouldAnimateDots: false)
+    }
+
+    private lazy var whiteTopAndBottomGradientLayer: CAGradientLayer = {
+        let layer = CAGradientLayer.init()
```
joewalsh: This can be written as `CAGradientLayer()` - also saw a few other places, for example `CGPoint.init(` that can be written as `CGPoint(`

montehurd: Ah, good catch, will fix!

montehurd: Fixed!

montehurd: Feel free to fix any stragglers I missed...

Wikipedia/Code/OnThisDayCollectionViewCell.swift
```diff
@@ -1,16 +1,41 @@
...
-class OnThisDayCollectionViewCell: SideScrollingCollectionViewCell {
-
-    let timelineView = OnThisDayTimelineView()
+@objc(WMFOnThisDayExploreCollectionViewCell)
+class OnThisDayExploreCollectionViewCell: OnThisDayCollectionViewCell {
 
-    @objc(configureForExploreWithOnThisDayEvent:previousEvent:dataStore:layoutOnly:)
-    func configureForExplore(with onThisDayEvent: WMFFeedOnThisDayEvent,  previousEvent: WMFFeedOnThisDayEvent, dataStore: MWKDataStore, layoutOnly: Bool) {
+    @objc(configureWithOnThisDayEvent:previousEvent:dataStore:layoutOnly:)
+    func configure(with onThisDayEvent: WMFFeedOnThisDayEvent,  previousEvent: WMFFeedOnThisDayEvent, dataStore: MWKDataStore, layoutOnly: Bool) {
```
joewalsh: This should be moved into a class extension

montehurd: Sounds good! Will move.

montehurd: @joewalsh oh wait do you mean don't use a concrete `OnThisDayExploreCollectionViewCell` subclass altogether?

joewalsh: I mean to have `extension OnThisDayExploreCollectionViewCell {` either in `ArticleCollectionViewCell+WMFFeedContentDisplaying.swift` or a new `OnThisDayExploreCollectionViewCell+WMFFeedContentDisplaying.swift`

joewalsh: @montehurd just for the configuration methods - basically don't have those methods that take model objects in the main view files.

montehurd: @joewalsh ah I see thx!

WMF Framework/WMFOnThisDayEventsFetcher.m
```diff
@@ -27,9 +27,13 @@ - (BOOL)isFetching {
     return [[self.operationManager operationQueue] operationCount] > 0;
 }
 
+- (NSArray*)supportedLanguages {
```
joewalsh: This should return an `NSSet` instead of an array due it's use as a lookup. It should also be a class method with a `static NSSet` created within a `dispatch_once` block.

montehurd: That's better - will fix.

montehurd: Fixed!

Wikipedia/Code/OnThisDayCollectionViewCell.swift
montehurd: Fixed!

PR #1591, opened 2017-07-01
Fix various article footer issues

montehurd:
- [x] ensure 'Read more' section header can always be scrolled to top https://phabricator.wikimedia.org/T166609
- [x] fix footer layout on tablet (use gray background, full width footer menu, etc)
- [x] make footer read more item 'Save for later' button use article lang
- [x] fix not using wikidata description (always using extract) bug on read more items

Pics of footer layout fixes on tablet:
# Before
before 111

# After
after 111


montehurd:
@joewalsh I think this is good for review.

Wikipedia/Code/WKWebView+WMFWebViewControllerJavascript.swift
```diff
@@ -130,11 +130,13 @@ extension WKWebView {
...
         }
         
+        evaluateJavaScript("window.addEventListener('resize', window.wmf.footerContainer.updateBottomPaddingToAllowReadMoreToScrollToTop);", completionHandler: nil)
+        
         let heading = WMFLocalizedString("article-read-more-title", language: article.url.wmf_language, value: "Read more", comment: "The text that is displayed before the read more section at the bottom of an article\n{{Identical|Read more}}").wmf_stringByReplacingApostrophesWithBackslashApostrophes().uppercased(with: Locale.current)
         evaluateJavaScript("window.wmf.footerReadMore.setHeading('\(heading)', 'footer_container_readmore_heading');", completionHandler: nil)
 
-        let saveForLaterString = SaveButton.saveTitle.wmf_stringByReplacingApostrophesWithBackslashApostrophes()
-        let savedForLaterString = SaveButton.savedTitle.wmf_stringByReplacingApostrophesWithBackslashApostrophes()
+        let saveForLaterString = WMFLocalizedString("button-save-for-later", language: article.url.wmf_language, value: "Save for later", comment: "Longer button text for save button used in various places.")
```
joewalsh: Having the same localization in multiple places can cause issues with generating the localization strings - if they get out of sync, it's not clear which comment and value are used. This should be added to save button  as `SaveButton.saveTitle(with language:...`. Then, `SaveButton.saveTitle` should be updated to use the new method with language nil. This way, the translation only exists in one place in source.

montehurd: Good point! I was wondering about how to best handle dupes...

montehurd: Fixed!

PR #1592, opened 2017-07-01
W icon tooltip

montehurd:
https://phabricator.wikimedia.org/T166270

- [x] ask Carolyn about corner rounding


montehurd:
Note to self: docs say we can use "resizableImage:withCapInsets:" https://developer.apple.com/documentation/uikit/uipopoverbackgroundview?changes=latest_minor

montehurd:
@joewalsh our `WMFDynamicHeightPopoverMessage` convenience methods don't return handles, so punted on rotation bits for now.

PR #1598, opened 2017-07-05
Fix iOS 9 article loading regression

montehurd:
https://phabricator.wikimedia.org/T169660

# Before
screen shot 2017-07-05 at 3 28 10 pm

# After
screen shot 2017-07-05 at 3 29 33 pm

[This PR](https://github.com/wikimedia/wikimedia-page-library/pull/48) on the page library fixed the bug preventing pages from loading correctly on iOS 9 (and older Android) devices.


Wikipedia/assets/index.js
```diff
@@ -1181,38 +1202,40 @@ var newCaption = function newCaption(title, headerText) {
...
 
   var tables = content.querySelectorAll('table');
-  for (var i = 0; i < tables.length; ++i) {
+
+  var _loop = function _loop(i) {
     var table = tables[i];
 
     if (elementUtilities.findClosestAncestor(table, '.app_table_container') || !shouldTableBeCollapsed(table)) {
-      continue;
+      return 'continue';
```
montehurd: @niedzielski Assuming babel changed this (and below in this method)? Works fine, but do you know what bit got flipped to turn this on?

niedzielski: @montehurd, I'm not sure why this changed but it looks correct albeit obfuscated. Looking a couple lines up, this changed from a for loop to a function called loop so the `continue` keyword no longer applies and you can see how it's used further down like you mentioned. If you're feeling doubtful, you might try removing your node_modules directory and reinstalling to verify you have the correct versions of the tools and then just toggling a section.

PR #1599, opened 2017-07-06
T164598 - Fix place popover position

julienbodet:
This PR fixes the bug described here: https://phabricator.wikimedia.org/T164598

I also corrected the calculation of the location of the popover. When positioned on top or left, the distance between the place annotation and the place popover was 2 times more as when on bottom or right.


joewalsh:
ok to test

Wikipedia/Code/PlacesViewController.swift
```diff
@@ -1832,13 +1833,13 @@ class PlacesViewController: PreviewingViewController, UISearchBarDelegate, Artic
...
-        var y = annotationCenter.y > viewCenter.y ? viewSize.height - popoverSize.height - spacing : spacing
+        var y = annotationCenter.y > viewCenter.y ? viewSize.height - popoverSize.height - spacing : spacing + navBarHeight
 
         let canFitTopOrBottom = viewSize.width - annotationCenter.x > 0.5*popoverSize.width && annotationCenter.x > 0.5*popoverSize.width
-        let fitsTop = top < 0 && canFitTopOrBottom
+        let fitsTop = top < -navBarHeight && canFitTopOrBottom
         let fitsBottom = bottom < 0 && canFitTopOrBottom
         
-        let canFitLeftOrRight = viewSize.height - annotationCenter.y > 0.5*popoverSize.height && annotationCenter.y > 0.5*popoverSize.width
+        let canFitLeftOrRight = viewSize.height - annotationCenter.y > 0.5*popoverSize.height && annotationCenter.y - navBarHeight > 0.5*popoverSize.height
```
joewalsh: good catch

PR #1601, opened 2017-07-07
 Swiping left on a list item should show relevant actions for that item in Saved and History

nambatee:
I'm not sure whether the way I'm building `shareActivityController ` is ok or whether it's necessary at all in this case (maybe I should be using something that's already there).
Also, I think that string descriptions in `WMFLocalizedStringWithDefaultValue ` might not be descriptive enough, is there a convention that I should be following? 

Happy to adjust 👍


joewalsh:
ok to test

Wikipedia/Code/WMFArticleListTableViewController.m
```diff
@@ -259,4 +252,83 @@ - (void)tableViewUpdater:(WMFTableViewUpdater *)updater didUpdateTableView:(UITa
...
+
+- (NSString *)shareActionText {
+    return WMFLocalizedStringWithDefaultValue(@"article-share", nil, nil, @"Share", @"Text of the article list row action shown on swipe which allows the user to choose the sharing option");
+}
+
+- (NSString *)saveActionText {
+    return WMFLocalizedStringWithDefaultValue(@"article-save", nil, nil, @"Save", @"Text of the article list row action shown on swipe which allows the user to save the article");
+}
+
+- (UITableViewRowAction *)shareAction:(NSIndexPath *)indexPath {
```
joewalsh: These row actions can lead to some conflicting animations. The simplest workaround would be to add a helper method to this VC that doesn't apply the edit until the table view end editing animation is complete (similar to how places handles it):

```
- (UITableViewRowAction *)rowActionWithStyle:(UITableViewRowActionStyle)style title:(nullable NSString *)title handler:(void (^)(UITableViewRowAction *action, NSIndexPath *indexPath))handler {
    return [UITableViewRowAction rowActionWithStyle:style title:title handler:^(UITableViewRowAction *action, NSIndexPath *indexPath){
        [CATransaction begin];
        [CATransaction setCompletionBlock:^{
            if (handler) {
                handler(action, indexPath);
            }
        }];
        [self.tableView setEditing:NO animated:YES];
        [CATransaction commit];
    }];
}
```

before:
before

after:
after

Also, `WMFTableViewUpdater.m` should have `UITableViewRowAnimationNone` for any `reloadRowsAtIndexPaths:`

Wikipedia/Code/WMFArticleListTableViewController.m
```diff
@@ -259,4 +252,83 @@ - (void)tableViewUpdater:(WMFTableViewUpdater *)updater didUpdateTableView:(UITa
...
+}
+
+- (void)shareArticle:(UIActivityViewController *)shareActivityController {
+    [self presentViewController:shareActivityController animated:YES completion:NULL];
+}
+
+#pragma mark - Row actions
+
+- (NSString *)deleteActionText {
+    return WMFLocalizedStringWithDefaultValue(@"article-delete", nil, nil, @"Delete", @"Text of the article list row action shown on swipe which deletes the article");
```
joewalsh: It seems like your build environment isn't running the localization scripts. On every build, any new strings should be added to the strings files, but they aren't in this diff. They do get updated when I run the build locally.

Wikipedia/Code/WMFArticleListTableViewController.m
nambatee: Thanks, Joe! I added `rowActionWithStyle ` to the VC and set `reloadRowsAtIndexPaths ` in `WMFTableViewUpdater.m` to `UITableViewRowAnimationNone ` but the animation still looks the same for me (as in 'before')

nambatee:
fixing the conflicts

Wikipedia/Code/WMFArticleListTableViewController.m
joewalsh: @nambatee are the other calls updated to `[self rowActionWithStyle:` or are they still `[UITableViewRowAction rowActionWithStyle:`

nambatee: thanks, overlooked :/ 😳!

PR #1609, opened 2017-07-11
Themes part 4

joewalsh:
- Themes for log in, sign up
- Fix places search bar
- Fix article footer VC background and nav bar (still not completely themed)


montehurd:
Noticed some keyboard funkiness when toggling dark mode when search interface is visible:
wot mov

joewalsh:
@montehurd should be fixed with latest commit

PR #1610, opened 2017-07-12
Centralize share activity controller

nambatee:
First attempt at centralizing the share activity controller, follow-up on @joewalsh suggestion in #1601.

I got rid of the `wprov` parameter anywhere but in `WMFShareOptionsController.h`. I looked at the latest implementation of share activity controller in `PlacesViewController.swift` and didn't see that parameter being added to the url. Depending on whether `wprov` parameter is necessary, I will add it to other `inits` in `ShareActivityController` or remove the one that I left.


Wikipedia/Code/ShareActivityController.swift
```diff
@@ -0,0 +1,94 @@
+@objc(WMFShareActivityController)
+class ShareActivityController: UIActivityViewController {
+    
+    init(url: URL, userDataStore: MWKDataStore, context: AnalyticsContextProviding) {
```
joewalsh: `url:` should be renamed `articleURL` to make it clear it's expecting an `articleURL`

Wikipedia/Code/ShareActivityController.swift
```diff
@@ -0,0 +1,94 @@
...
+        let article = userDataStore.fetchArticle(with: url)
+        var items = [Any]()
+        let text: String
+        
+        if let article = article {
+            if let title = article.displayTitle {
+                text = "\"\(title)\" on @Wikipedia"
```
joewalsh: The declaration for text can be moved here since it's only used within the scope of this if statement.

Wikipedia/Code/ShareActivityController.swift
```diff
@@ -0,0 +1,94 @@
...
+    
+    init(article: WMFArticle, context: AnalyticsContextProviding) {
+        let tracker = PiwikTracker.sharedInstance()
+        tracker?.wmf_logActionShare(inContext: context, contentType: article)
+        
+        var items = [Any]()
+        let text: String
+        
+        if let title = article.displayTitle {
+            text = "\"\(title)\" on @Wikipedia"
```
joewalsh: Same comment re: text

Wikipedia/Code/ShareActivityController.swift
```diff
@@ -0,0 +1,94 @@
...
+        
+        items.append(title)
+        
+        //TODO: Is wprov param needed?
+        if let image = image {
+            param = "wprov=sfii1"
+            items.append(image)
+        }
+        
+        param = "wprov=sfti1"
```
joewalsh: Will this over-write the `sfii1` param in the case it's an image? Could be moved to an else after the `if let image`

joewalsh: Also, I confirmed that we should add `wprov=sfti1` to all shared links, `wprov=sfii1` only if it's share a fact with an image.

nambatee:
In the last commit, passing `imageDownload.image.staticImage` instead of `imageDownload.image` fixes the no "Save Image" button bug reported today [T170451](https://phabricator.wikimedia.org/T170451)

Wikipedia/Code/ShareActivityController.swift
```diff
@@ -0,0 +1,93 @@
...
+        items.append(title)
+        
+        if let image = image {
+            param = "wprov=sfii1"
+            items.append(image)
+        } else {
+            param = "wprov=sfti1"
+        }
+        
+        if article.url != nil, let url = URL(string: "\(article.url.absoluteString)?\(param)") {
```
joewalsh: Missed this in the earlier review - but this should use the same code that was in article list table vc for adding the query params (but in Swift): ```NSURLComponents *components = [NSURLComponents componentsWithURL:url resolvingAgainstBaseURL:NO];
        
        NSURLQueryItem *queryItem = [NSURLQueryItem queryItemWithName:@"wprov" value:@"sfsi1"];
        components.queryItems = @[queryItem];
        
        NSURL *componentsURL = components.URL;```

joewalsh: (class name is `URLComponents` in swift instead of `NSURLComponents`)

PR #1611, opened 2017-07-12
Move disambig and page issues determination to the footer menu transform

montehurd:
Phase one of https://phabricator.wikimedia.org/T164137

The footer menu JS now extracts the disambiguation URLs and the page issues strings and relays them to native land. Formerly these were extracted by parsing the article HTML in native land.

Consolidating this to the footer menu transform so both platforms can use same code.

(Next phase will start moving the transform JS to the page lib.)

woooo mov


montehurd:
@joewalsh Hey I noticed we have a bunch of code files in the `WikipediaUnitTests/code` folder... guessing this was accidental? (I left them for now.)

PR #1623, opened 2017-07-14
Update reading themes controls in the article reader

nambatee:


Wikipedia/Code/WMFReadingThemesControlsViewController.swift
```diff
@@ -0,0 +1,171 @@
...
+    }
+    
+    @IBAction func darkThemeButtonPressed(_ sender: Any) {
+        let theme = ["theme": Theme.dark]
+        
+        applyBorder(to: darkThemeButton)
+        removeBorderFrom(standardThemeButton)
+        removeBorderFrom(lightThemeButton)
+        
+        NotificationCenter.default.post(name: Notification.Name(WMFReadingThemesControlsViewController.WMFUserDidSelectThemeNotification), object: nil, userInfo: theme)
```
joewalsh: The contents of the `ThemeButtonPressed` methods could be abstracted into a single update method.

Wikipedia/Code/WMFReadingThemesControlsViewController.swift
```diff
@@ -0,0 +1,171 @@
...
+            case "dark":
+                applyBorder(to: darkThemeButton)
+            default:
+                break
+            }
+        }
+    }
+    
+    @IBAction func brightnessSliderValueChanged(_ sender: UISlider) {
+        UIScreen.main.brightness = CGFloat(sender.value)
```
joewalsh: May want to also listen for `UIScreenBrightnessDidChangeNotification` and update the slider if the user changes the brightness using the system slider while this UI is up.

Wikipedia/Code/WMFAppViewController.m
```diff
@@ -1450,6 +1458,17 @@ - (void)applyTheme:(WMFTheme *)theme {
     [self setNeedsStatusBarAppearanceUpdate];
 }
 
+- (void)changeTheme:(NSNotification *)note {
+    NSDictionary* userInfo = note.userInfo;
+    WMFTheme* theme = (WMFTheme*)userInfo[@"theme"];
```
joewalsh: It's helpful to define user info key strings as constants as well to avoid typos with the string - in this case, something like `WMFUserDidSelectThemeNotificationThemeKey`

Wikipedia/Code/WMFReadingThemesControlsViewController.swift
```diff
@@ -0,0 +1,171 @@
+import UIKit
+
+@objc public protocol WMFReadingThemesControlsViewControllerDelegate {
+    
+    func fontSizeSliderValueChangedInController(_ controller: WMFReadingThemesControlsViewController, value: Int)
+}
+
+open class WMFReadingThemesControlsViewController: UIViewController {
```
joewalsh: Have been moving towards `@objc(WMFReadingThemesControlsViewController)` while keeping the swift class named without the WMF prefix

PR #1625, opened 2017-07-17
[T170699] Fix spacing bug in 'On this day' detail view

joewalsh:
https://phabricator.wikimedia.org/T170699

Also updates the on this day explore feed card sort order based on the description in https://phabricator.wikimedia.org/T170700


nambatee:
Looks good to me

PR #1627, opened 2017-07-17
Themes part 8

joewalsh:
- Apply themes to article reference popovers & tool tips
- Bump history/saved image view size to 40


Wikipedia/Code/WMFAppViewController.m
```diff
@@ -192,8 +192,20 @@ - (void)loadMainUI {
...
-    [[NSUserDefaults wmf_userDefaults] wmf_setAppTheme:self.theme.name];
+    
+    NSString *themeName = [[NSUserDefaults wmf_userDefaults] wmf_appThemeName];
+    
+    NSArray *themes = @[WMFTheme.light, WMFTheme.sepia, WMFTheme.dark];
+    
+    WMFTheme *themeToApply = WMFTheme.standard;
+    for (WMFTheme *theme in themes) {
+        if ([themeName isEqualToString:theme.name]) {
+            themeToApply = theme;
```
montehurd: Curious about use of `for / if` over `wmf_match`...

Wikipedia/Code/WMFReferencePopoverMessageViewController.m
```diff
@@ -129,4 +123,29 @@ - (BOOL)accessibilityPerformEscape {
...
+    self.theme = theme;
+    if (self.viewIfLoaded == nil) {
+        return;
+    }
+
+    self.view.backgroundColor = theme.colors.paperBackground;
+    
+    self.textView.linkTextAttributes = @{NSForegroundColorAttributeName: theme.colors.link};
+    
+    [self.textView setAttributedText:[self attributedStringForHTML:[self referenceHTMLWithSurroundingHTML]]];
```
montehurd: Could things be re-arranged a bit so this call to `setAttributedText` (and the one on 144) could be made after the call to `applyTheme` rather than inside it?

joewalsh: The attributed text depends on the colors of the theme, so it needs to be updated anytime the theme changes, so it needs to be inside of `applyTheme:`

PR #1631, opened 2017-07-18
[T170701] Re-add separators to table views & list detail views

joewalsh:
https://phabricator.wikimedia.org/T170701


Wikipedia/Code/ArticleRightAlignedImageCollectionViewCell.swift
```diff
@@ -50,6 +54,11 @@ open class ArticleRightAlignedImageCollectionViewCell: ArticleCollectionViewCell
             origin.y += saveButtonFrame.height
         }
         origin.y += margins.bottom
+        
+        if (apply) {
+            separator.frame = CGRect(x: 0, y: origin.y - 0.5, width: size.width, height: 0.5)
```
montehurd: Can we use `1.0 / UIScreen.main.scale` here so non-retina devices will get single pixel hairlines too?

joewalsh: Yeah, good catch. Should be `self.traitCollection.displayScale` though - view might not necessarily be on the main screen.

montehurd: Oh interesting. We're using mainscreen in a few other places... may want to follow up later changing those to trait collection too.

PR #1633, opened 2017-07-19
Apply reading themes to article view

montehurd:
https://phabricator.wikimedia.org/T169148

theme mov

There are a couple parts of the article which won't "take" the new colors yet (footer, collapsed table in sepia, etc), but these will be fixed by updating the pagelib transform - at which point I'll open a separate PR to slurp up those fixes. 

So this PR is good to go so we can start using it and identifying any other areas which need attention.


montehurd:
For future reference, [this](https://gerrit.wikimedia.org/r/#/c/365181/) was the Android integration patch.

niedzielski:
This looks great! In the future, maybe we can play with making auto-night mode a gradient instead of a binary.

montehurd:
@niedzielski oh ya that could be neat. iirc the auto-night mode may not ship until version after initial iOS night mode release.

PR #1634, opened 2017-07-19
Add controls for reading themes to Settings view

nambatee:


WMF Framework/Theme.swift
```diff
@@ -160,11 +160,11 @@ public class Theme: NSObject {
     public let blurEffectStyle: UIBlurEffectStyle
     public let keyboardAppearance: UIKeyboardAppearance
     
-    public static let light = Theme(colors: .light, preferredStatusBarStyle: .default, blurEffectStyle: .light, keyboardAppearance: .light, name: "standard")
+    public static let light = Theme(colors: .light, preferredStatusBarStyle: .default, blurEffectStyle: .light, keyboardAppearance: .light, name: "Default")
```
joewalsh: This should probably be "light". Also, they should stay lowercased and not user-facing.

joewalsh: A user facing name could be added as `localizedName` or what is currently `name` move to `key` and `name` could be a localized name.

montehurd:
Looks like `settings-image-dimming` icon needs to be set in the asset catalog to be single scale.

montehurd:
Tapping the `Image dimming` cell unchecks the reading theme selection above it.

montehurd:
The "D" in "Image Dimming" should probably be lower case. Translators tend to prefer it this way.

montehurd:
A couple parts aren't responding to "Dynamic type" text size changes. If you haven't played with the "Accessibility Inspector" Xcode comes bundled with it's pretty slick:

inspect mov

montehurd:
One more thing to check - there's a console message saying there's some constraint breakage happening when your VC appears.

montehurd:
Oh I added a quick note to the [phab ticket](https://phabricator.wikimedia.org/T169255#3455305) too.

Wikipedia/Localizations/en.lproj/Localizable.strings
```diff
@@ -45,8 +45,11 @@
 "action-saved-accessibility" = "Saved. Activate to unsave.";
 "action-share" = "Share";
 "action-unsave" = "Unsave";
+"adjust-text-sizing" = "Adjust text sizing";
```
montehurd: May want to consider renaming "adjust-text-sizing" and "adjust-text-sizing-footer" to start with "appearance-settings-".

montehurd: Same with the other adding strings except "settings-appearance".

montehurd: @joewalsh Thoughts on whether we should keep adhering to the TWN string naming convention we had before your localization refactors? Now that strings "live" where they're used the verbose naming we had used before is arguably less useful...

joewalsh: We should keep the prefix for strings in similar contexts. It's nice to have the extra info, and keeps them ordered together on the TWN side.

Wikipedia/Code/AppearanceSettingsViewController.swift
```diff
@@ -0,0 +1,245 @@
...
+    
+    
+    static var disclosureText: String {
+        var text = WMFLocalizedString("default-theme-display-name", value: "Default", comment: "Default theme name presented to the user")
+        let currentAppTheme = UserDefaults.wmf_userDefaults().wmf_appTheme
+        
+        if currentAppTheme == Theme.sepia {
+            text = WMFLocalizedString("sepia-theme-display-name", value: "Sepia", comment: "Sepia theme name presented to the user")
+        } else if currentAppTheme == Theme.dark || currentAppTheme == Theme.darkDimmed {
+            text = WMFLocalizedString("dark-theme-display-name", value: "Dark", comment: "Dark theme name presented to the user")
```
joewalsh: The localized theme names can be added as `localizedName` or `displayName` to the theme object.  That way, you can do `let text = currentAppTheme.localizedName`.

Wikipedia/Code/AppearanceSettingsViewController.swift
```diff
@@ -0,0 +1,245 @@
...
+        return WMFLocalizedString("default-theme-display-name", value: "Default", comment: "Default theme name presented to the user")
+    }
+    
+    func sepiaThemeDisplayName() -> String {
+        return WMFLocalizedString("sepia-theme-display-name", value: "Sepia", comment: "Sepia theme name presented to the user")
+    }
+    
+    func darkThemeDisplayName() -> String {
+        return WMFLocalizedString("dark-theme-display-name", value: "Dark", comment: "Dark theme name presented to the user")
+    }
```
joewalsh: ^ these should be removed, using the above suggestion, you could do `Theme.sepia.localizedName` etc.

Wikipedia/Code/AppearanceSettingsViewController.swift
```diff
@@ -0,0 +1,245 @@
...
+        footer?.setShortTextAsProse(sections[section].footerText)
+        return footer
+    }
+    
+    public func tableView(_ tableView: UITableView, heightForFooterInSection section: Int) -> CGFloat {
+        guard let footerText = sections[section].footerText else {
+            return 0
+        }
+        
+        let footer = WMFTableHeaderLabelView.wmf_viewFromClassNib()
```
joewalsh: use `guard let footer` instead of the force unwrap below

Wikipedia/Code/AppearanceSettingsViewController.swift
```diff
@@ -0,0 +1,245 @@
...
+        footer?.text = footerText
+        return footer!.height(withExpectedWidth: self.view.frame.width)
+    }
+    
+    public func tableView(_ tableView: UITableView, heightForHeaderInSection section: Int) -> CGFloat {
+        let header = WMFTableHeaderLabelView.wmf_viewFromClassNib()
```
joewalsh: `guard let header`

joewalsh: Nevermind - the custom headers & footers can be replaced with the standard headers and footers. This also fixes the layout attribute issue. The custom headers only exist in Notification settings due to having non-uppercase text in a header.

So all the current viewForHeader/Footer and heightForHeader/Footer can be replaced with:

```
    public func tableView(_ tableView: UITableView, titleForHeaderInSection section: Int) -> String? {
        return sections[section].headerTitle
    }
    
    public func tableView(_ tableView: UITableView, titleForFooterInSection section: Int) -> String? {
        return sections[section].footerText
    }
```

Wikipedia/Code/AppearanceSettingsViewController.swift
nambatee: great idea, thanks!!!

nambatee:
@montehurd, thanks for the comment! I'll look into the constraints and accessibility stuff! I have a quick question about the "Image dimming" string - the string value is "Image dimming" with a lowercase "D" and I'm not capitalizing it anywhere in the code. Looking at the implementation of `WMFSettingsTableViewCell` and `SSBaseTableCell` it inherits from, there's no capitalization happening anywhere 🤔
Do you know what might be causing the capitalization? It's not happening for cell titles in main Settings ... 🤔

joewalsh:
@nambatee it looks like the "image-dimming" `WMFLocalizedString` exists in two places in the code, so it's undefined which value is used. In this case, it's taking the "Image Dimming" value from `ReadingThemesControlsViewController`. Take a look at `CommonStrings.swift` and add any strings shared between the settings view and the popover to that class.

Wikipedia/Code/AppearanceSettingsViewController.swift
```diff
@@ -0,0 +1,183 @@
+import UIKit
+
+@objc public protocol WMFAppearanceSettingsViewControllerDelegate {
+    func themeChangedInController(_ controller: AppearanceSettingsViewController, theme: Theme)
```
joewalsh: This should be removed, the `WMFUserDidSelectThemeNotification` should cause `WMFAppViewController` to properly update the theme - there could be a bug in the app controllers code if something isn't getting properly updated.

nambatee:
@joewalsh you were right, I swapped `image-dimming` for `appearance-settings-dim-images` in `AppearanceSettingsViewController` and it's not capitalizing anymore. Is there any good practice for naming the the same strings (`{{Identical|String}}`?)?

joewalsh:
@nambatee the translators add `{{Identical|String}}` if it's identical to another key (which can happen if the same string is presented in different contexts so we have different keys for it). However, in this case, I think the contexts are similar enough that they should share the same localization strings. To share the same strings in different places, add an entry to `CommonStrings` similar to how `shortSavedTitle` works and use it in settings and the article popover.

nambatee:
The "Dynamic type" issues disappeared when I stopped using custom headers & footers

Wikipedia/Localizations/qqq.lproj/Localizable.strings
```diff
@@ -73,6 +78,8 @@
 "compass-direction" = "Spoken description of compass direction, e.g. \"at 3 o'clock\" means \"to the right\", \"at 11 o'clock\" means \"slightly to the left\", etc. $1 is the hour.";
 "continue-reading-empty-description" = "Explore Wikipedia for more articles to read";
 "continue-reading-empty-title" = "No recently read articles";
+"dark-theme-display-name" = "Dark theme name presented to the user";
```
montehurd: Oh, can we name "dark-theme-display-name", "default-theme-display-name" and "sepia-theme-display-name" to start with "theme-" so they group together?

nambatee: sure, will throw it in in some other pr!

PR #1640, opened 2017-07-20
[T171139] Fix incorrect colors on article load

joewalsh:
theme


montehurd:
Oh nice! I may have one tweak to this based on the theme JS - I'll double-check when I get to office.

PR #1643, opened 2017-07-21
Part 2 of initial article load theme setting to avoid flicker

montehurd:
https://phabricator.wikimedia.org/T171139

- This should ensure that all the theme color bits are set before the web view renders anything.
- Theme colors now used on article edit preview web view.
- Reduces some duplication between index.html and preview.html
- I held off on further refactor of the `late` JS transform invocations in the webVC for now.


Wikipedia/Code/PreviewAndSaveViewController.m
```diff
@@ -703,6 +703,10 @@ - (void)applyTheme:(WMFTheme *)theme {
     if (self.viewIfLoaded == nil) {
         return;
     }
+
+    self.previewWebViewContainer.webView.opaque = NO;
+    self.previewWebViewContainer.webView.scrollView.backgroundColor = [UIColor clearColor];
```
niedzielski: I believe we do something similar on Android.

Wikipedia/Code/WKWebView+WMFWebViewControllerJavascript.swift
```diff
@@ -156,27 +156,30 @@ extension WKWebView {
...
 
     public func wmf_classifyThemeElements(){
+        // 'themes.classifyElements()' needs to happen once after body elements are present. it
+        // classifies some tricky elements like math formula images (see 'enwiki > Quadradic formula')
         evaluateJavaScript("window.wmf.themes.classifyElements(document);", completionHandler: nil)
     }
     
-    public func wmf_applyTheme(_ theme: Theme){
+    static public func wmf_themeApplicationJavascript(with theme: Theme) -> String{
         var jsThemeConstant = "DEFAULT"
```
niedzielski: This is practical and good as is but in case it wasn't clear, these constants are also exposed in JavaScript as `pagelib.ThemeTransform.THEME.DEFAULT` / `DARK` / `SEPIA`.

Wikipedia/Code/WKWebView+WMFWebViewControllerJavascript.swift
```diff
@@ -156,27 +156,30 @@ extension WKWebView {
         let readMoreItemCount = 3
         evaluateJavaScript("window.wmf.footerReadMore.add('\(title)', \(readMoreItemCount), 'pagelib_footer_container_readmore_pages', '\(proxyURL)', \(saveButtonTapHandler), \(titlesShownHandler), document);", completionHandler: nil)
     }
-    
-    public func wmf_enableCompatibilityTransformSupport(){
-        evaluateJavaScript("window.wmf.compatibility.enableSupport(document);", completionHandler: nil)
-    }
 
     public func wmf_classifyThemeElements(){
```
niedzielski: I'm not sure where you call this in the browsing view but don't forget to do the same in the preview view. (BTW, think I'd like to consider merging bundle.js and preview.js later.)

Wikipedia/Code/WKWebView+WMFWebViewControllerJavascript.swift
montehurd: Oh yes that's what these reference on line 178.

Wikipedia/Code/WKWebView+WMFWebViewControllerJavascript.swift
montehurd: I'll double check thanks! (Was thinking similar thoughts earlier about just using the regular bundle on preview.)

joewalsh:
simulator screen shot jul 21 2017 6 31 18 am
simulator screen shot jul 21 2017 6 30 42 am
simulator screen shot jul 21 2017 6 30 37 am

It appears that some of the preview text colors are incorrect. On `develop`, the previews are unthemed entirely.

Wikipedia/Code/WKWebView+WMFWebViewControllerJavascript.swift
montehurd: Good catch it wasn't getting called correctly on preview! Fixed!

montehurd:
@joewalsh 

> It appears that some of the preview text colors are incorrect. On develop, the previews are unthemed entirely.

I double-checked those screenshot issues and it looks like these are some of the places we'll need to make follow-up patches to the pagelib theme transform. I'll file a ticket :)

montehurd:
@joewalsh here's the ticket: https://phabricator.wikimedia.org/T171326

PR #1651, opened 2017-07-22
Hook up new article image dimming transform

montehurd:
https://phabricator.wikimedia.org/T169151

(see comment [here](https://phabricator.wikimedia.org/T169151#3462686) for follow-ons we can probably now knock out)
toast mov


montehurd:
Is it just me or is the bookmark icon changing to wrong color in the animation? (away from laptop a.t.m)

PR #1654, opened 2017-07-24
[T171336] Update on this day header

joewalsh:
simulator screen shot jul 24 2017 10 06 46 am


WMF Framework/UIFont+WMFDynamicType.swift
```diff
@@ -117,11 +117,11 @@ let fontSizeTable: [WMFFontFamily:[UIFontTextStyle:[UIContentSizeCategory:CGFloa
                 .extraSmall: 12
             ],
             UIFontTextStyle.body: [
-                .accessibilityExtraExtraExtraLarge: 23,
-                .accessibilityExtraExtraLarge: 23,
-                .accessibilityExtraLarge: 23,
-                .accessibilityLarge: 23,
-                .accessibilityMedium: 23,
```
montehurd: Ya iirc we deliberately cap the accessibility sizes at the point at which layout completely explodes. Or at least I think we probably should until we have time to come up with layouts which don't fall apart at those huge sizes.

PR #1655, opened 2017-07-24
[T169941] Pick an event to feature for on this day

joewalsh:
https://phabricator.wikimedia.org/T169941#3466369


joewalsh:
@montehurd updated the property name to `score`. Rather than enumerate `event.articlePreviews` twice at this point (as a calculate score method would have to do), it can stay inline with the existing implementation.

PR #1656, opened 2017-07-24
Accessibility for Reading Themes Controls

nambatee:


Wikipedia/Code/ReadingThemesControlsViewController.swift
```diff
@@ -69,7 +81,7 @@ open class ReadingThemesControlsViewController: UIViewController {
         button.borderWidth = 2
         button.isEnabled = false
         button.borderColor = theme.colors.link
-
+        button.accessibilityTraits = UIAccessibilityTraitSelected
```
joewalsh: I think this needs a corresponding `button.accessibilityTraits = UIAccessibilityTraitButton` in `func removeBorderFrom(_ button: UIButton) {`

Wikipedia/Code/ReadingThemesControlsViewController.swift
```diff
@@ -55,7 +55,19 @@ open class ReadingThemesControlsViewController: UIViewController {
         brightnessSlider.value = Float(UIScreen.main.brightness)
         
         imageDimmingLabel.text = CommonStrings.dimImagesTitle
-    
+        
+        brightnessSlider.accessibilityLabel = WMFLocalizedString("reading-themes-controls-brightness-slider", value: "Brightness slider", comment: "Accessibility label for the brightness slider in the Reading Themes Controls popover")
```
joewalsh: These should include accessibility in the key: for example `reading-themes-controls-accessibility-brightness slider` or `reading-themes-controls-brightness slider-accessibility-label`. I'm leaning towards adding `-accessibility-label` but either is fine. This is to ensure it's extra clear that they're accessibility labels (even though it's indicated in the comment)

PR #1662, opened 2017-07-25
Dismissing reading themes controls popover when tapping on other toolbar item

nambatee:
- Reading Themes Controls Popover is dismissed when other `UIBarButtonItem` is tapped
-  Dim images switch in Reading Controls Popover is more prominent for disabled & off state (Default, Sepia)

Awaiting other design review tweaks

### DIFF
--- a/Wikipedia/Code/ReadingThemesControlsViewController.swift
+++ b/Wikipedia/Code/ReadingThemesControlsViewController.swift
@@ -164,12 +164,7 @@ open class ReadingThemesControlsViewController: UIViewController {
     
     func makeImageDimmingSwitchMoreProminentForDisabledState(_ isEnabled: Bool) {
         imageDimmingSwitch.layer.cornerRadius = imageDimmingSwitch.frame.height / 2
-        if isEnabled {
-            imageDimmingSwitch.backgroundColor = nil
-        } else {
-        imageDimmingSwitch.tintColor = UIColor.wmf_lighterGray
-        imageDimmingSwitch.backgroundColor = UIColor.wmf_lighterGray
-        }
+        imageDimmingSwitch.backgroundColor = isEnabled ? nil : UIColor.wmf_lighterGray
     }
     
     func screenBrightnessChangedInApp(notification: Notification){
--- a/Wikipedia/Code/ReadingThemesControlsViewController.swift
+++ b/Wikipedia/Code/ReadingThemesControlsViewController.swift
@@ -137,6 +137,7 @@ open class ReadingThemesControlsViewController: UIViewController {
         removeBorderFrom(darkThemeButton)
         removeBorderFrom(sepiaThemeButton)
         imageDimmingSwitch.isEnabled = false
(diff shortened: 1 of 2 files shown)


Wikipedia/Code/ReadingThemesControlsViewController.swift
```diff
@@ -154,12 +155,18 @@ open class ReadingThemesControlsViewController: UIViewController {
...
         case Theme.dark.name:
             imageDimmingSwitch.isEnabled = true
+            makeImageDimmingSwitchMoreProminentForDisabledState(imageDimmingSwitch.isEnabled)
             applyBorder(to: darkThemeButton)
         default:
             break
         }
     }
     
+    func makeImageDimmingSwitchMoreProminentForDisabledState(_ isEnabled: Bool) {
```
montehurd: Would makeEnabledImageDimmingSwitchMoreProminent better describe what this does?

Wikipedia/Code/WMFArticleViewController.m
```diff
@@ -559,9 +559,12 @@ - (UIBarButtonItem *)findInPageToolbarItem {
 }
 
 - (void)findInPageButtonPressed {
-    if ([self canFindInPage]) { // Needed so you can't tap find icon when text size adjuster is onscreen.
```
montehurd: Is the 'canFind' check no longer needed?

Wikipedia/Code/WMFArticleViewController.m
```diff
@@ -1244,19 +1251,10 @@ - (void)shareArticleWhenReady {
     }
 }
 
-#pragma mark - Find-in-page
-
-- (void)showFindInPage {
-    if (self.presentedViewController != nil) {
```
montehurd: Oh, is the nil check also now not needed?

Wikipedia/Code/ReadingThemesControlsViewController.swift
montehurd: Or makeImageDimmingSwitchMoreProminentWithEnabledState per the Swifty convention...

joewalsh: If you're going to change this at all, it should be combined with setting `isEnabled` on the switch:

`var isImageDimmingSwitchEnabled { didSet { ... } }`

Wikipedia/Code/WMFArticleViewController.m
joewalsh: It now dismisses the themes popover when you tap the button.

Wikipedia/Code/WMFArticleViewController.m
joewalsh: Correct, since the intended behavior is now to dismiss the themes/text size popover when the user taps find in page.

montehurd: What about reference popovers?

joewalsh: @montehurd they cover the toolbar, disallowing any interaction with it

PR #1668, opened 2017-07-26
Track basic usage of appearance options

nambatee:
https://phabricator.wikimedia.org/T171669

Not sure if I'm implementing the target (normal/sepia/dark) for switching to a different theme right (`contentType: AnalyticsContent(self.theme.displayName)` on line 174 in `ReadingThemesControlsViewController` and line 113 in `AppearanceSettingsViewController`)

I was following the existing naming convention for actions in `PiwikTracker+WMFExtensions` (e.g., `                             action:@"Adjust Brightness"`) even though the ticket is asking for a different one (e.g., `adjust_brightness )


joewalsh:
@jtminor ^ Should the names keep with the existing style with spaces (`Adjust Brightness`) or should the be as the ticket describes with underscores(`adjust_brightness`)?

nambatee:
Josh confirmed that we should use the existing pattern

PR #1680, opened 2017-07-31
[T172013] Fix transparent article image background color

joewalsh:
https://phabricator.wikimedia.org/T172013


montehurd:
I added a note to the phab ticket - (iirc we don't want to do this because it will cause issues with other images with transparency).

PR #1682, opened 2017-07-31
[T172019] Don't show older news stories at the end of newer feed content

joewalsh:
Only add or show a news card that matches the feed date being loaded.

https://phabricator.wikimedia.org/T172019


joewalsh:
retest this please

PR #1687, opened 2017-08-01
[T172014] Update themes feed card image

joewalsh:


joewalsh:
This is just an asset change. Merging.

PR #1690, opened 2017-08-01
Users do not understand why 'Dim images' is deactivated on Default and Sepia

nambatee:
https://phabricator.wikimedia.org/T172010

Not sure about:

- `settings-image-dimming-example` in assets is set to `Individual Scales`, for vector files we normally choose `Single Scale` but in this case assets are `.jpg`
- `Content Mode` in `ImageDimmingExampleViewController.xib` is set to `Aspect Fit`


joewalsh:
retest this please

montehurd:
@nambatee Hey would you mind double-checking one thing on this patch - this image https://github.com/wikimedia/wikipedia-ios/pull/1690/files#diff-548805877a66f8d00a3ba488b840dc4e looks really blurry. If you can confirm could you get a better version from Carolyn?

PR #1691, opened 2017-08-01
Fix network notification crash, disable new event logging

joewalsh:


Wikipedia/Code/WMFNetworkUtilities.m
```diff
@@ -28,11 +32,16 @@
...
+    if (request) {
+        userInfo[WMFNetworkRequestBeganNotificationRequestKey] = request;
+    }
     [[NSNotificationCenter defaultCenter] postNotificationName:WMFNetworkRequestBeganNotification
                                                         object:nil
-                                                      userInfo:@{ @"method": method,
-                                                                  @"URLString": URLString }];
+                                                      userInfo:userInfo];
+#endif
 }
```
joewalsh: ^ crash was here. Either `method` or `URLString` was nil.

PR #1696, opened 2017-08-02
[T170465] Fix search URL handling

joewalsh:
https://phabricator.wikimedia.org/T170465

Search URLs were being constructed using a format string rather than `NSURLQueryItem`s.

Also, setting a search term on a search VC that hadn't been presented wasn't triggering a search.

### DIFF
--- a/Wikipedia/Code/WMFSearchViewController.m
+++ b/Wikipedia/Code/WMFSearchViewController.m
@@ -180,10 +180,6 @@ static NSUInteger const kWMFMinResultsBeforeAutoFullTextSearch = 12;
 
     [self updateUIWithResults:nil];
     [self updateRecentSearchesVisibility:NO];
-    
-    if (self.searchTerm) {
-        self.searchTerm = self.searchTerm;
-    }
 }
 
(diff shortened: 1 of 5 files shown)


Wikipedia/Code/WMFSearchViewController.m
```diff
@@ -364,16 +368,19 @@ - (BOOL)textFieldShouldClear:(UITextField *)textField {
...
 
-- (void)setSearchTerm:(NSString *)searchTerm {
-    if (searchTerm.length == 0) {
+- (void)performSearchWithCurrentSearchTerm {
+    if (!self.viewIfLoaded) {
         return;
     }
-    [self setSearchFieldText:searchTerm];
-    [self searchForSearchTerm:searchTerm];
+    if (self.searchTerm.length == 0) {
```
montehurd: Oh do we want to trip whitespace for this check?

joewalsh: In the old version, there wasn't a trim here. This line was just moved down, not added.

PR #1697, opened 2017-08-02
Add text size control to Appearance settings

nambatee:
https://phabricator.wikimedia.org/T172307


Wikipedia/Code/ImageDimmingExampleViewController.swift
```diff
@@ -20,6 +20,6 @@ extension ImageDimmingExampleViewController: Themeable {
         }
         
         view.backgroundColor = theme.colors.baseBackground
-        view.alpha = theme.imageOpacity
+        view.subviews.first?.alpha = theme.imageOpacity
```
nambatee: dimming the entire view looks bad

dimmingview

Wikipedia/Code/AppearanceSettingsViewController.swift
```diff
@@ -126,6 +124,7 @@ open class AppearanceSettingsViewController: UIViewController, UITableViewDataSo
             case Theme.dark:
                 cell.disclosureSwitch.isEnabled = true
                 cell.disclosureSwitch.addTarget(self, action: #selector(self.handleImageDimmingSwitchValueChange(_:)), for: .valueChanged)
+                userDidSelect(theme: currentAppTheme.withDimmingEnabled(cell.disclosureSwitch.isOn))
```
nambatee: without this, after changing the theme from Dark (image dimming on) to Sepia or Default and changing back to Dark, the example image loses the dim:

img_0022

Wikipedia/Code/ImageDimmingExampleViewController.swift
joewalsh: Good catch & fix - probably should use an IBOutlet for the image view in case the view is modified in the future and subviews.first is no longer the image view

Wikipedia/Code/FontSizeSliderViewController.swift
```diff
@@ -32,7 +32,7 @@ class FontSizeSliderViewController: UIViewController {
         
         slider.isAccessibilityElement = true
         slider.accessibilityTraits = UIAccessibilityTraitAdjustable
-        slider.accessibilityLabel = WMFLocalizedString("reading-themes-controls-accessibility-text-size-slider", value: "Text size slider", comment: "Accessibility label for the text size slider in the Appearance Settings\n{{Identical|Text size slider}}")
+        slider.accessibilityLabel = WMFLocalizedString("appearance-settings-accessibility-text-size-slider", value: "Text size slider", comment: "Accessibility label for the text size slider in the Appearance Settings\n{{Identical|Text size slider}}")
```
joewalsh: This can be consolidated in `CommonStrings` - use `reading-themes-controls-accessibility-text-size-slider` since we have translations for that

joewalsh: Updated here 8ce8247

PR #1699, opened 2017-08-02
Theme related tweaks to article reference panel presentation

montehurd:
https://phabricator.wikimedia.org/T171677

- [x] update bg color of refs overlay used by sepia theme
- [x] update bg alpha of refs overlay used by dark theme
- [x] fix currently selected reference link coloring
- [x] use white for selected reference link in dark themes

Note: click images below for higher res.

Before | After
--- | ---
screen shot 2017-08-02 at 3 08 36 pm | screen shot 2017-08-02 at 3 09 57 pm
screen shot 2017-08-02 at 3 08 26 pm | screen shot 2017-08-02 at 3 10 04 pm


joewalsh:
It looks like there's still a diff in the CSS. Is this intentional?

montehurd:
> We have a ton of theme colors, so I can understand the hesitance to add another one. However, that's preferred over having the colors stored inside the view subclass with a switch on the theme name. Having them all in the theme makes them easier to re-use (or update) and keeps the view side unaware of any color info. I made the change and pushed to this branch.

Makes sense!

montehurd:
@joewalsh

Yes the CSS change is intentional - is because when I moved the ref CSS to its own file I also increased the specificity of the selectors. (needed to play nice with the forthcoming article theme CSS refactor)

PR #1700, opened 2017-08-03
Fix for 'On this day' detail view sometimes not changing to date when you scroll down.

montehurd:
https://phabricator.wikimedia.org/T171205

### DIFF
--- a/Wikipedia/Code/OnThisDayViewController.swift
+++ b/Wikipedia/Code/OnThisDayViewController.swift
@@ -22,7 +22,17 @@ class OnThisDayViewController: ColumnarCollectionViewController {
     
     var isDateVisibleInTitle: Bool {
         didSet {
-            guard isDateVisibleInTitle, let language = events.first?.language else {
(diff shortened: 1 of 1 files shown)


Wikipedia/Code/OnThisDayViewController.swift
```diff
@@ -22,7 +22,17 @@ class OnThisDayViewController: ColumnarCollectionViewController {
     
     var isDateVisibleInTitle: Bool {
         didSet {
-            guard isDateVisibleInTitle, let language = events.first?.language else {
+            
+            // Work-around for: https://phabricator.wikimedia.org/T169277
+            // Presently the event looks to its first article preview when you ask it for the language, so if the event has no previews, no lang!
+            let firstEventWithArticlePreviews = events.first(where: {
```
joewalsh: This might also cause an issue with the header not being localized in the correct language. See `OnThisDayHeaderViewController.swift:19`

PR #1709, opened 2017-08-04
Remove SSDataSource from WMFSettingsViewController & fix weird animation

nambatee:
https://phabricator.wikimedia.org/T171867

Without `SSDataSource`, I'm not seeing the weird animation anymore


Wikipedia/Code/WMFSettingsViewController.m
```diff
@@ -387,50 +382,66 @@ - (void)tweakViewControllerPressedDone:(FBTweakViewController *)tweakViewControl
...
-        return ((WMFSettingsMenuItem *)[self.elementDataSource itemAtIndexPath:indexPath]).type == type;
+        return ([self.tableView cellForRowAtIndexPath:indexPath].tag == type);
     }];
 }
 
 - (void)reloadVisibleCellOfType:(WMFSettingsMenuItemType)type {
     NSIndexPath *indexPath = [self indexPathForVisibleCellOfType:type];
     if (indexPath) {
-        [self.elementDataSource replaceItemAtIndexPath:indexPath withItem:[WMFSettingsMenuItem itemForType:type]];
+        [self.tableView reloadRowsAtIndexPaths:@[indexPath] withRowAnimation:UITableViewRowAnimationNone];
```
montehurd: I think `reloadVisibleCellOfType` will still need to update the datasource's respective row with a new `[WMFSettingsMenuItem itemForType:type]`, otherwise `reloadRowsAtIndexPaths` will just show what it was showing before. For example, I don't think login will show you that you've logged in otherwise. (Oh that reminds me - another thing to be sure to test after fix is that you can log in, delete the app, reinstall, go back to settings, and ensure it still shows you as being logged in.)

joewalsh: You could probably delete this method entirely and just use reload sections instead. There probably isn't a huge performance advantage to just reloading the rows that changed.

PR #1714, opened 2017-08-05
Bug/t171094

GianmarcoSalerno:
Check to see if the user is nil. If not then proceed with adding the revision.


Wikipedia/Code/PageHistoryFetcher.swift
```diff
@@ -180,24 +180,26 @@ extension HistoryFetchResults {
...
-            if let revisionDate = revision.revisionDate {
-                var title: String?
-                let getSectionTitle = {
-                    title = DateFormatter.wmf_long().string(from: revisionDate)
-                }
-                if Thread.isMainThread {
-                    getSectionTitle()
-                } else {
-                    DispatchQueue.main.sync(execute: getSectionTitle)
+        if (revision.user != nil) {
```
montehurd: Consider using 'guard' here instead of 'if' - pretty handy way to keep the diff (and rightward indentation) minimal.

montehurd:
Thanks for the PR! Thought I was being too picky about the 'guard' so I removed the comment but you already fixed it :)

GianmarcoSalerno:
I actually appreciate you being picky! It helps me learn, since I am new to iOS developing.

GianmarcoSalerno:
Hey @joewalsh the change should have already been reverted. If you look, there is no diff in the original and what I am committing

screen shot 2017-08-07 at 1 46 01 pm

, but for some reason the file still shows up under the category of being changed.

joewalsh:
ok to test

PR #1716, opened 2017-08-07
Tapping left or right of the text slider does not change the text size

nambatee:
https://phabricator.wikimedia.org/T171204

Starting off by proposing a slightly different solution - instead of incrementing/decrementing by 1 step, tap the preferred step:

sliiiider


Wikipedia/Code/ReadingThemesControlsViewController.swift
```diff
@@ -79,6 +79,27 @@ open class ReadingThemesControlsViewController: UIViewController, AnalyticsConte
...
     }
     
+    open override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+        let tap = UITapGestureRecognizer(target: self, action: #selector(self.handleTap(_:)))
+        slider.addGestureRecognizer(tap)
+    }
+    
+    func handleTap(_ sender: UIGestureRecognizer) {
+        let pointTapped: CGPoint = sender.location(in: self.view)
```
joewalsh: might want to do `location(in: slider)` since the tap gesture recognizer is added to the slider

Wikipedia/Code/ReadingThemesControlsViewController.swift
```diff
@@ -79,6 +79,27 @@ open class ReadingThemesControlsViewController: UIViewController, AnalyticsConte
...
+        
+        let positionOfSlider: CGPoint = slider.frame.origin
+        let widthOfSlider: CGFloat = slider.frame.size.width
```
joewalsh: if you change to `location(in: slider)` above, the resulting point is in the slider's view coordinates, so you would use `slider.bounds` since `slider.frame` is in the main view's coordinates

Wikipedia/Code/ReadingThemesControlsViewController.swift
```diff
@@ -79,6 +79,27 @@ open class ReadingThemesControlsViewController: UIViewController, AnalyticsConte
...
+        let newValue = ((pointTapped.x - positionOfSlider.x) * CGFloat(slider.numberOfSteps) / widthOfSlider)
```
joewalsh: you could also remove the position in slider adjustment here since the point returned by `location(in: slider)` should have an x coordinate between 0 and `slider.bounds.size.width` - make sure to test trying to tap outside the slider bounds though

Wikipedia/Code/ReadingThemesControlsViewController.swift
```diff
@@ -79,6 +79,27 @@ open class ReadingThemesControlsViewController: UIViewController, AnalyticsConte
...
+        
+        CATransaction.begin()
+        CATransaction.setDisableActions(true)
```
joewalsh: there's `[UIView performWithoutAnimations:]` that you could use here if it works

Wikipedia/Code/ReadingThemesControlsViewController.swift
```diff
@@ -79,6 +79,25 @@ open class ReadingThemesControlsViewController: UIViewController, AnalyticsConte
...
+        slider.addGestureRecognizer(tap)
+    }
+    
+    func handleTap(_ sender: UIGestureRecognizer) {
+        let pointTapped: CGPoint = sender.location(in: slider)
+        
+        let widthOfSlider: CGFloat = slider.bounds.size.width
+        let newValue = pointTapped.x * (CGFloat(slider.numberOfSteps) / widthOfSlider)
+
+        slider.value = Int(newValue)
```
joewalsh: Should this be `newValue.rounded()`? That way, for example, values from 0.5 --> 1.49 become 1 instead of 1.0 --> 1.99 becoming 1

nambatee: With `rounded()` not all steps are being respected 🤔
rounded

nambatee:
Didn't notice that the slider jumps back on swiping to the closest step on either side:

sliderjumps

Wikipedia/Code/ReadingThemesControlsViewController.swift
joewalsh: @nambatee ah, i should have tested - looks like it works fine without `rounded()`

nambatee: np, thanks for the suggestion!

montehurd:
Oh don't forget to test RTL!

PR #1717, opened 2017-08-07
Update contributors, fix ar.lproj strings

joewalsh:


montehurd:
@joewalsh Did the arabic strings change upstream?

joewalsh:
@montehurd the strings were out of sync after merging https://github.com/wikimedia/wikipedia-ios/pull/1714

montehurd:
@joewalsh Ah, perhaps my comment here https://github.com/wikimedia/wikipedia-ios/pull/1714#commitcomment-23497492 was misleading then and I should have pointed the volunteer to the translate wiki page for that particular arabic string?

joewalsh:
@montehurd your comment was correct, but the advice wasn't followed. I took it as a GitHub UTF-16 fluke that it was showing a diff on the strings file when there wasn't any change. Also, since the PR wasn't for the arabic translation bug, it shouldn't have had a change on the strings file.

montehurd:
Gotcha.

PR #1720, opened 2017-08-08
[Bug] 3D touch of article cards on the On this day feed card does not always lead to the correct article

nambatee:
https://phabricator.wikimedia.org/T172762


Wikipedia/Code/WMFExploreCollectionViewController.m
```diff
@@ -1509,7 +1509,10 @@ - (nullable UIViewController *)previewingContext:(id<UIViewControllerPreviewing>
             UIView *view = [sideScrollingCell viewForSubItemAtIndex:index];
             CGRect sourceRect = [view convertRect:view.bounds toView:self.collectionView];
             previewingContext.sourceRect = sourceRect;
-            NSUInteger indexes[3] = {previewIndexPath.section, previewIndexPath.item, index};
+            
+            NSUInteger articleToPeekIndex = group.articleURLString.integerValue;
```
nambatee: Since `NSUInteger` can never be nil, we might want to use a different value if `articleToPeekIndex == NSNotFound` but not sure what's the best solution in this case

joewalsh: The logic for the featured content index should be added to `WMFContentGroup+Extensions` as `@property (nonatomic) NSInteger featuredContentIndex;` and `@property (nonatomic, readonly, nullable) id<NSCoding> featuredContentObject;` so that the view controllers don't need to know about the `articleURLString.integerValue` hack (and we can make it a proper field in the future).

Also, the root cause is in `detailViewControllerForItemAtIndexPath`. For "On this day", the section only has one item in it, so the second index path should always be zero (even though that corresponds to a different index inside of the content array). The third index should be the index of the article in the side scrolling list. `detailViewControllerForItemAtIndexPath` should always use the  contentGroup's `featuredContentObject` to get the featured event.

PR #1727, opened 2017-08-14
Read more text line height with smallest-size system font is too low, text runs into itself

nambatee:
https://phabricator.wikimedia.org/T169262

Before

before

After

after


nambatee:
https://github.com/wikimedia/wikimedia-page-library/pull/82

PR #1728, opened 2017-08-14
Swiping left on a list item should show relevant actions for that item in Search Results

nambatee:
https://phabricator.wikimedia.org/T173343

After

img_0046
img_0047
img_0048
img_0049
img_0050


montehurd:
@nambatee I may be missing something, but the phab ticket is talking about `Recently Searched`, and the ticket doesn't define what these actions are which a user could take for a given search term:

screen shot 2017-08-15 at 3 34 05 pm

montehurd:
Did you hear back from Josh about this one?

PR #1732, opened 2017-08-16
[BUG] "Read more" section title should not be shown on screen when there are no related articles

nambatee:
https://phabricator.wikimedia.org/T172906


nambatee:
The first solution is invalid cause `User:` is being translated to different languages. For example, on Polish Wikipedia, user pages start with `Wikipedysta:`.

PR #1733, opened 2017-08-16
Pages not adhering to dark mode

nambatee:
https://phabricator.wikimedia.org/T171799

Before

img_0043
img_0044
img_0045

After

img_0034
img_0035
img_0036
img_0037
img_0038
img_0039
img_0040
img_0041
img_0042


nambatee:
There are changes to localization files because I branched off develop after recent twn sync. I think we'll be adding those translations to 5.6.1 anyway so it shouldn't matter.

montehurd:
@nambatee 

> There are changes to localization files because I branched off develop after recent twn sync. I think we'll be adding those translations to 5.6.1 anyway so it shouldn't matter.

If you rebase on the `releases/5.6.1` branch I think those TWN commits will disappear.

montehurd:
Just be sure to do `git pull origin releases/5.6.1` before rebasing.

nambatee:
@montehurd Rebasing on the `releases/5.6.1` didn't remove the localization strings (with a previous pull).

montehurd:
Hmm we should only be seeing the commits added by your PR, not my commits. Maybe we can batcave for a few tomorrow and get to the bottom of it?

PR #1734, opened 2017-08-17
Pages not adhering to dark mode

nambatee:
See #1733.


Wikipedia/Code/AboutViewController.m
```diff
@@ -59,6 +59,15 @@ - (void)wmf_setTextFontSize {
     [self evaluateJavaScript:fontSizeJS completionHandler:nil];
 }
 
+- (void)wmf_setTextFontColor:(WMFTheme *)theme {
+    NSString *fontColorJS = [NSString stringWithFormat:@"document.body.style.color = '#%@';\
+                             document.styleSheets[0].rules[2].style.color = '#%@';\
```
nambatee: @montehurd if you add an id to `style`, it won't let you access its css rules. It looks like it's useful in different cases: https://stackoverflow.com/questions/3797221/why-would-you-give-a-style-tag-an-id/3797236

Wikipedia/assets/about.html
```diff
@@ -26,6 +26,15 @@
                 color:#000000;
```
montehurd: These changes need to be made to `/www/about.html` or they'll be overwritten when we run `make grunt`.

Everything in `/assets/` gets overwritten each time we run `make grunt`.

Wikipedia/Code/AboutViewController.m
montehurd: @nambatee 
I think I got some JS accessing the selectors by name - ok if I push a commit?

montehurd: I used the inspector and was able to access the rules from style id handle. 
`document.getElementById(#ID_HERE#).sheet.rules`

Was also able to get the styles by selector name. Ping me if you want a quick pair session.

montehurd: Or I can push a commit if changes are welcome. Either way works for me.

nambatee: Sure, feel free to push

PR #1736, opened 2017-08-18
SPIKE - Highlight every onscreen section in the TOC so we can remove buggy dynamic bottom space JS

montehurd:
https://phabricator.wikimedia.org/T173529

Would allow us to remove the footer container transform's JS which dynamically adjusts the space beneath the legal footer at the bottom of the page.

The reason it would let us get rid of that `dynamic bottom space` code is the TOC would no longer be looking at what article section intersects the `top of the screen` (actually it's a little ways down from the top) to determine which TOC sections to highlight but instead it would just highlight whatever article sections were `anywhere` onscreen (so there would be no need for dynamic space JS just to allow the bottom section to be able to scroll to near top of the screen).

To ship this proof of concept would need to be updated to:
- remove unused toc section indicator dot logic
- dry up new `deselectRow` logic
- make TOC footer read more and footer menu logic also be able to highlight both items if they're both onscreen
- remove footer container JS which dynamically adjusts space at bottom

11111 mov


montehurd:
This reverted the spike: https://github.com/montehurd/apps-ios-wikipedia/commit/2f6b39fae3388b962a1c5e177bf461851500e074

montehurd:
Here's this same spike updated to work with develop as of 2019-02-06:
https://github.com/wikimedia/wikipedia-ios/tree/spike/toc-multiple-section-highlight/T173529-2

PR #1741, opened 2017-08-23
[5.6.1] Widget updates

joewalsh:
https://phabricator.wikimedia.org/T172807
- Eliminate unnecessary view updates & retained managed objects from the top read widget
- Update Featured Article title font to match other widgets
- Tweak colors & spacing


TopReadWidget/WMFTodayTopReadWidgetViewController.swift
```diff
@@ -347,33 +347,41 @@ class WMFTodayTopReadWidgetViewController: UIViewController, NCWidgetProviding {
...
     
-    func updateUIWithTopReadFromContentStoreForSiteURL(siteURL: URL, date: Date) -> Bool {
+    func updateUIWithTopReadFromContentStoreForSiteURL(siteURL: URL, date: Date) -> NCUpdateResult {
         if let topRead = self.userStore.viewContext.group(of: .topRead, for: date, siteURL: siteURL) {
             if let content = topRead.content as? [WMFFeedTopReadArticlePreview] {
+                if let previousGroupKey = self.group?.key,
+                    let topReadKey = topRead.key,
+                    self.results.count > 0,
+                    topReadKey == previousGroupKey {
+                    return .noData
```
montehurd: Thoughts on using `.dataUnchanged` instead of `.noData`? (just based on the commit message)

FeaturedArticleWidget/FeaturedArticleWidget.swift
```diff
@@ -146,5 +151,21 @@ class FeaturedArticleWidget: UIViewController, NCWidgetProviding {
...
+            return
+        }
+        guard let article = self.article, let articleURL = article.url else {
+            return
+        }
+
+        let URL = articleURL as NSURL?
+        let URLToOpen = URL?.wmf_wikipediaScheme ?? NSUserActivity.wmf_baseURLForActivity(of: .explore)
+
+        self.extensionContext?.open(URLToOpen, completionHandler: { (success) in
```
montehurd: Was the leaving the completion block empty intentional? Could maybe pass it nil if nothing needs to happen on completion.

TopReadWidget/WMFTodayTopReadWidgetViewController.swift
joewalsh:
```public enum NCUpdateResult : UInt {

    
    case newData

    case noData

    case failed
}
```
`.dataUnchanged` isn't an option

FeaturedArticleWidget/FeaturedArticleWidget.swift
joewalsh: updated

TopReadWidget/WMFTodayTopReadWidgetViewController.swift
montehurd: Oh sorry didn't realize was existing enum!

PR #1743, opened 2017-08-23
[BUG] "Read more" section title should not be shown on screen when there are no related articles

nambatee:
https://phabricator.wikimedia.org/T172906


Wikipedia/Code/WebViewController.m
```diff
@@ -303,7 +303,8 @@ - (void)handleLateJavascriptTransformScriptMessage:(NSString *)messageString {
         [self.webView wmf_setPageProtected:!self.article.editable];
     } else if ([messageString isEqualToString:@"addFooterContainer"]) {
         [self.webView wmf_addFooterContainer];
-    } else if ([messageString isEqualToString:@"addFooterReadMore"]) {
+        // ns of articles with Read More pages equals 0.
+    } else if ([messageString isEqualToString:@"addFooterReadMore"] && self.article.ns == 0) {
```
joewalsh: should also check for `!self.article.isMain` - could add a computed readonly property `hasReadMore` to `MWKArticle` to use here and in `WMFArticleViewController`

nambatee: sounds great, will add that!

Wikipedia/Code/MWKArticle.m
```diff
@@ -491,4 +491,11 @@ - (NSString *)articleHTML {
     return htmlStr;
 }
 
+#pragma mark Read More
+
+- (BOOL)hasReadMore {
+    // ns of articles with Read More pages equals 0.
+    return self && !self.isMain && self.ns == 0;
```
joewalsh: The check for `self` is unnecessary, but doesn't hurt

Wikipedia/Code/WMFArticleViewController.m
```diff
@@ -379,7 +379,7 @@ - (BOOL)hasAboutThisArticle {
     return self.article && !self.article.isMain;
 }
 
-- (void)getShareText:(void (^)(NSString *text))completion {
```
joewalsh: there was a `^C` between `)` and `c`

PR #1755, opened 2017-08-25
Fix location cell image scale

joewalsh:
Before:
simulator screen shot aug 25 2017 11 31 57 am

After:
simulator screen shot aug 25 2017 11 30 58 am


joewalsh:
It's expected that the Jenkins runs will fail for 5.6.1 PRs after the Xcode 9 update. Ran `fastlane verify_pull_request` locally

PR #1757, opened 2017-08-28
Remove SS data source from Page Issues VC

nambatee:
There should be no visual changes to the Page Issues view.

To test:

1. Search for ator
2. Open the first result
3. Go to About this article
4. Open Page Issues


Wikipedia/Code/BaseTableViewCell.swift
```diff
@@ -0,0 +1,8 @@
+import UIKit
+
+@objc(WMFBaseTableViewCell)
+class BaseTableViewCell: UITableViewCell {
```
joewalsh: You could remove this class and use `UITableViewCell` directly (unless there are plans to put more functionality here?)

Wikipedia/Code/PageIssuesTableViewController.swift
```diff
@@ -0,0 +1,70 @@
...
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        self.title = WMFLocalizedStringWithDefaultValue("page-issues", nil, nil, "Page issues", "Label for the button that shows the \"Page issues\" dialog, where information about the imperfections of the current page is provided (by displaying the warning/cleanup templates).\n{{Identical|Page issue}}")
+        
+        self.tableView.estimatedRowHeight = 90.0
+        self.tableView.rowHeight = UITableViewAutomaticDimension
+        self.tableView.separatorStyle = UITableViewCellSeparatorStyle.none
+        
+        self.tableView.register(BaseTableViewCell.self, forCellReuseIdentifier: "BaseTableViewCell")
```
joewalsh: If `BaseTableViewCell` stays, might as well use `BaseTableViewCell.identifier()`

PR #1758, opened 2017-08-28
[5.6.1] Keep label background clear while selected & highlighted states are animating

joewalsh:
https://phabricator.wikimedia.org/T174341


joewalsh:
Tests were run here: https://github.com/wikimedia/wikipedia-ios/pull/1759

PR #1776, opened 2017-09-01
Prevent edit pencil peeking.

montehurd:
https://phabricator.wikimedia.org/T174791

### DIFF
--- a/Wikipedia/Code/WKWebView+WMFWebViewControllerJavascript.m
+++ b/Wikipedia/Code/WKWebView+WMFWebViewControllerJavascript.m
@@ -2,6 +2,7 @@
 @import WMF;
 #import "Wikipedia-Swift.h"
 #import "WMFProxyServer.h"
+#import <WMF/NSURL+WMFLinkParsing.h>
 
 // Some dialects have complex characters, so we use 2 instead of 10
 static int const kMinimumTextSelectionLength = 2;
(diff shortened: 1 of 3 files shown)


Wikipedia/Code/WKWebView+WMFWebViewControllerJavascript.m
```diff
@@ -18,7 +19,11 @@ - (void)wmf_collapseTablesForArticle:(MWKArticle *)article {
 
 - (void)wmf_addEditPencilsForArticle:(MWKArticle *)article {
     if (!article.isMain) {
-        [self evaluateJavaScript:@"window.wmf.editButtons.add(document);" completionHandler:nil];
+        [self evaluateJavaScript: [NSString stringWithFormat:@""
+                                   "window.wmf.editButtons.add(document);"
+                                   "document.querySelectorAll('.pagelib_edit_section_link').forEach(anchor => {anchor.href = '%@'});"
```
montehurd: Had to set the edit pencil href to an identifiable string because:
- the transform doesn't set an href for the edit pencil anchors
- when the browser sees the href is unset it seems to assume the url of the current article (hence peeking the pencil showing the same article)
- WKPreviewElementInfo only provides access to `linkURL`, so there's no other thing you can check in `- (BOOL)webView:(WKWebView *)webView shouldPreviewElement:(WKPreviewElementInfo *)elementInfo`

PR #1778, opened 2017-09-01
Remove now-unneeded version checks.

montehurd:


joewalsh:
Removing iOS 9 support isn't on the board for 5.7 - Josh wants to gather upgrade data after iOS 11 is released before making the call

PR #1779, opened 2017-09-02
Update js syntax

montehurd:
Now that we've dropped iOS 9 we can use more modern JS syntax.

Also simplified file page edit pencil hiding.


joewalsh:
Removing iOS 9 support isn't on the board for 5.7 - Josh wants to gather upgrade data after iOS 11 is released before making the call

PR #1780, opened 2017-09-02
Fix edit preview article theme

montehurd:
https://phabricator.wikimedia.org/T174842


montehurd:
@joewalsh one more last-minute regression fix!

montehurd:
Should be good to go despite test failure.

PR #1788, opened 2017-09-08
Start removing Masonry

joewalsh:
Rationale here: https://phabricator.wikimedia.org/T175373

Only the first part of the work.


Wikipedia/Code/UIViewController+WMFChildViewController.swift
```diff
@@ -8,35 +8,6 @@ extension UIViewController {
...
-            let childView = childController.view
-            else{
-                assertionFailure("Expected child controller view")
-                return
-        }
-        wmf_configureForDynamicHeight(childView: childView, containerView: containerView)
-        addChildViewController(childController)
-        containerView.addSubview(childView)
-        childController.view.mas_makeConstraints(constraints)
-        childController.didMove(toParentViewController: self)
```
montehurd: I think the `didMove` call is needed for container-ized vc's...

https://developer.apple.com/documentation/uikit/uiviewcontroller/1621405-didmove
`If you are implementing your own container view controller, it must call the didMove(toParentViewController:) method of the child view controller after the transition to the new controller is complete or, if there is no transition, immediately after calling the addChildViewController(_:) method.`

joewalsh: Good catch, unintentionally removed that - re-added with latest commit

montehurd:
So far so good! Want to leave this open or submit separate follow-ons?

PR #1791, opened 2017-09-11
[5.6.1] Eliminate unnecessary updates to label background colors

joewalsh:
The Core Data fetch and inflation of Mantle objects still causes some hang, but this helps mitigate https://phabricator.wikimedia.org/T175543

Before:
screen shot 2017-09-11 at 7 31 28 am

After:
screen shot 2017-09-11 at 7 35 15 am


joewalsh:
Ran verify_pull_request locally with 0 failures

PR #1794, opened 2017-09-12
Implement custom swipe actions for collection view layout

nambatee:
https://phabricator.wikimedia.org/T174721
https://phabricator.wikimedia.org/T174713

simulator screen shot - iphone 6 - 2017-09-13 at 02 54 16

simulator screen shot - iphone 6 - 2017-09-13 at 02 54 25

simulator screen shot - iphone 6 - 2017-09-13 at 02 54 46

simulator screen shot - ipad air - 2017-09-13 at 02 55 43

simulator screen shot - iphone 6 - 2017-09-13 at 02 55 27


WMF Framework/CollectionViewCellActionsView.swift
```diff
@@ -0,0 +1,193 @@
...
+            self.addSubview(wrapper)
+            maxButtonWidth = max(maxButtonWidth, button.intrinsicContentSize.width)
+        }
+        
+        maximumWidth = maxButtonWidth * CGFloat(self.subviews.count)
+    }
+    
+    public weak var delegate: SwipeableDelegate?
+    
+    @objc func didTapDelete() {
```
joewalsh: This might benefit from being consolidated into a single method that passes the action or action type to the delegate. After consolidating, it would be easier to add new actions without adding new methods to this class and every delegate.

WMF Framework/CollectionViewSwipeToEditController.swift
```diff
@@ -0,0 +1,235 @@
...
+    
+    public init(collectionView: UICollectionView, theme: Theme) {
+        self.collectionView = collectionView
+        self.theme = theme
+        super.init()
+        
+        addPanGesture(to: self.collectionView)
+        addLongPressGesture(to: self.collectionView)
+        
+        if let gestureRecognizers = self.collectionView.gestureRecognizers {
```
joewalsh: Can this be set up with the `panGestureRecognizer`, `pinchGestureRecognizer`, etc properties? Or, move the addPanGesture, addLongPressGesture to after this (error in the console - it's adding a gesture recognizer as a requirement to fail for itself)

Wikipedia/Code/ArticleCollectionViewCell.swift
```diff
@@ -148,4 +153,221 @@ open class ArticleCollectionViewCell: CollectionViewCell {
...
+            return self.actionsView?.actions ?? []
+        }
+        set {
+            self.actionsView?.actions = newValue
+        }
+    }
+    
+    public var isSaved: Bool {
+        if let delegate = delegate, let indexPath = indexPathForActiveCell {
+            return delegate.isArticleSaved(at: indexPath)
```
joewalsh: Rather than give the cell specific knowledge of save/unsaved state, the action controller could ask the delegate for which actions are available for a given row.

PR #1795, opened 2017-09-13
Crash on sharing article from table view swipe action on iPad

nambatee:
https://phabricator.wikimedia.org/T175789

I introduced this bug when I was setting up swipe actions for Saved & History table views. I forgot to set `sourceView` and `sourceRect` for the popoverPresentationController.

I'm not sure whether we want to present the controller from the tab bar, from the cell, or from somewhere else.

Tab bar:
simulator screen shot - ipad air 2 - 2017-09-12 at 23 07 28

Cell:
simulator screen shot - ipad air 2 - 2017-09-12 at 23 18 10


nambatee:
We should present from the cell https://phabricator.wikimedia.org/T175789#3609223

PR #1798, opened 2017-09-14
Update Share-a-fact card design

joewalsh:
https://phabricator.wikimedia.org/T174213


Wikipedia/Images.xcassets/share-building.imageset/Contents.json
```diff
@@ -2,18 +2,7 @@
...
-    {
-      "idiom" : "universal",
-      "filename" : "shareIllo@2x.png",
-      "scale" : "2x"
-    },
-    {
-      "idiom" : "universal",
-      "filename" : "shareIllo@3x.png",
-      "scale" : "3x"
+      "filename" : "group4.pdf"
```
montehurd: May want a more descriptive file name here.

joewalsh: These filenames aren't really relevant - they're obscured by the asset catalog.

montehurd: Ah, fair enough.

montehurd:
I noticed on iOS 11 once I select some article text and try to share it, if I cancel I can't de-select the text by clicking anywhere else on the article. (not sure if such refinements were coming as follow-ons...)

joewalsh:
@montehurd try to pull, that was an issue with older revisions

montehurd:
Ah will do...

montehurd:
Still happening on 39d0147.

joewalsh:
@montehurd it's working here. could you take a video of the issue?

montehurd:
Couple small things can maybe tweak later:

- Half of the nav bar is showing, may consider just covering it:
screen shot 2017-09-14 at 11 05 53 am

- The "Building a share card" screen comes and goes a bit quickly.

- The Apple share interface is so big it's covering part of the image on an iPhone 6, so it's a bit tough to see what's being shared.

montehurd:
> @montehurd it's working here. could you take a video of the issue?

Will do!

montehurd:
@joewalsh 

click mov

montehurd:
Not sure what to do about the landscape experience... w/o a separate preview step we get this:
screen shot 2017-09-14 at 11 15 54 am

montehurd:
The landscape issue is more of a product/design thing. I'll add a note to the phab ticket.

nambatee:
@montehurd selecting & deselecting the text works for me on iOS 11, following your repro steps (it was an issue before that last fix but it works for me now)

montehurd:
Noticed one more:
- The "Copy | Lookup | Share" buttons aren't appearing on iOS 9 when I select text.

montehurd:
> @montehurd selecting & deselecting the text works for me on iOS 11, following your repro steps (it was an issue before that last fix but it works for me now)

Weird! I'm on the latest...

montehurd:
Slight difference on iOS 10 and 11 - wasn't sure if intentional:

- 10
screen shot 2017-09-14 at 11 29 54 am


- 11
screen shot 2017-09-14 at 11 31 34 am

montehurd:
@joewalsh if you want to merge and address the iOS 9 issue and https://github.com/wikimedia/wikipedia-ios/pull/1798#issuecomment-329570718 as follow-ons that's fine w me.

joewalsh:
@montehurd the iOS 9/10 vs 11 differences are intentional. iOS 11 allows us to override the share button, iOS 9/10 doesn't

montehurd:
@joewalsh ah gotcha on the override diff. that's a separate issue from it not working at all in iOS 9 though.

joewalsh:
@montehurd it's working for me on iOS 9.3, 6S simulator

simulator screen shot - iphone 6s - 2017-09-14 at 17 00 48
simulator screen shot - iphone 6s - 2017-09-14 at 17 00 45
simulator screen shot - iphone 6s - 2017-09-14 at 17 00 42

is it broken for you on simulator or device?

montehurd:
@joewalsh 

> is it broken for you on simulator or device?

sim, making quick gif...

joewalsh:
@montehurd it stopped working for me after re-launching the app on the iOS 9 simulator. I tried another clean install after sim reset, and it didn't work that time.

I switched to develop, and it has the same problem.

montehurd:
ah gotcha.

montehurd:
i had used the 4s sim and when i selected text i never saw the buttons appear

joewalsh:
@montehurd when replacing the article HTML with a basic page, the menu works fine. Also, pointing the webview to en.wikipedia.org directly also works fine. Points to an issue with the article JS.

simulator screen shot - iphone 6s - 2017-09-14 at 17 26 58
simulator screen shot - iphone 6s - 2017-09-14 at 17 25 28

joewalsh:
Since it's unrelated to this PR, it can be filed as a separate issue.

PR #1799, opened 2017-09-14
GDFL » GFDL

jhsoby:
https://phabricator.wikimedia.org/T175909


joewalsh:
@jhsoby good catch! I updated the rest of the instances in the codebase that had the same problem with #1800 . Thanks!

jhsoby:
Thanks @joewalsh! I considered doing that, but was afraid it might cause some hickups on the Translatewiki end, but hopefully it'll work out. :-)

PR #1801, opened 2017-09-14
Build server test - do not merge or delete until we see if build kickoff happens.

montehurd:


montehurd:
I handed off build server to IT for the move. It's back online.

PR #1803, opened 2017-09-19
Welcome screens updates

montehurd:
https://phabricator.wikimedia.org/T174207

- [x] update strings and layout to match new designs
- [x] add new exploration page
- [x] wire up new Skip/Next buttons to either side of UIPageControl dots
- [x] fade out new Skip/Next buttons when going to last page
- [x] remove masonry
- [x] don't show custom delete button on lang screen (woohoo!)
- [x] first pass at RTL/tablet/4s/5/6Plus layout compatibility
- [x] update animations
- [x] get design feedback on iOS 11 change to [word wrapping in UILabels](https://stackoverflow.com/q/46200027/135557)
- [x] use smaller top animation on iPhone 5
- [x] fix missing lang filter
- [x] further strings tweaks from Josh once they're done on [T175309](https://phabricator.wikimedia.org/T175309), including pop-up links on 1st and last page
- [ ] <strike>theme compliance (including smart invert)</strike>
- [x] address new `TODO:` comments in welcome swift files

I'll work on these as follow-ons if that's ok...
- [ ] fix missing "PRIMARY" lang indicator
- [ ] update lang picker popup title
- [ ] test/harden more with RTL/tablet/4s/5/6Plus etc
- [ ] use smaller title font on 4s and 5?
- [ ] dynamic type hardening
- [ ] use UIButtons instead of labels 

**Animations are WIP in the gif below:**

mov


Wikipedia/Code/WMFWelcomeAnalyticsViewController.swift
```diff
@@ -17,29 +36,40 @@ class WMFWelcomeAnalyticsViewController: UIViewController {
             toggle.isOn = false
             BITHockeyManager.shared().crashManager.crashManagerStatus = .alwaysAsk
         }
-        self.view.wmf_configureSubviewsForDynamicType()
+        view.wmf_configureSubviewsForDynamicType()
+        
+        linkLabel.addGestureRecognizer(UITapGestureRecognizer.init(target: self, action: #selector(showPrivacyAlert(_:))))
```
joewalsh: If possible, this should be a button instead of a label with a tap gesture recognizer.

Wikipedia/Code/WMFWelcomeAnalyticsViewController.swift
```diff
@@ -17,29 +36,40 @@ class WMFWelcomeAnalyticsViewController: UIViewController {
...
             BITHockeyManager.shared().crashManager.crashManagerStatus = .alwaysAsk
             UserDefaults.wmf_userDefaults().wmf_setSendUsageReports(false)
-        
         }
         updateToggleLabelTitleForUsageReportsIsOn(sender.isOn)
     }
+    
+    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        super.traitCollectionDidChange(previousTraitCollection)
+        linkLabel.font = UIFont.wmf_preferredFontForFontFamily(.systemBold, withTextStyle: .footnote, compatibleWithTraitCollection: traitCollection)
```
joewalsh: If `linkLabel` is converted to a button, you can use `linkButton.titleLabel.font`

Wikipedia/Code/WMFWelcomeAnimationBackgroundView.swift
```diff
@@ -0,0 +1,82 @@
...
+    }()
+    
+    override open func addAnimationElementsScaledToCurrentFrameSize(){
+        super.addAnimationElementsScaledToCurrentFrameSize()
+        removeExistingSubviewsAndSublayers()
+        guard let imageViewsAndModels = imageViewsAndModels else {return}
+        imageViewsAndModels.forEach{ (imageViewAndModel) in
+            addSubview(imageViewAndModel.imageView)
+            // Start all images in the center. imageModel unitDestination assumes center origin.
+            imageViewAndModel.imageView.center = CGPoint(x: bounds.midX, y: bounds.midY);
```
joewalsh: unnecessary semicolon

Wikipedia/Code/WMFWelcomeAnimationBackgroundView.swift
```diff
@@ -0,0 +1,82 @@
...
+         print("unitSize \(imageUnitSize)")
+     }
+    */
+    override open func beginAnimations() {
+        super.beginAnimations()
+        CATransaction.begin()
+        guard let imageViewsAndModels = imageViewsAndModels else {return}
+        imageViewsAndModels.forEach{ (imageViewAndModel) in
+            let dest = imageViewAndModel.model.unitDestination.wmf_denormalizeUsingSize(frame.size)
+            let tx = CATransform3DMakeTranslation(dest.x, dest.y, 1.0)
```
joewalsh: If there's no translation or scaling in the Z coordinate, this could all be accomplished with `UIView`'s `transform` property and `CGAffineTransform`

Wikipedia/Code/WMFWelcomeAnimationViewControllers.swift
```diff
@@ -0,0 +1,85 @@
...
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        guard let animationView = animationView else {return}
+        animationView.translatesAutoresizingMaskIntoConstraints = false
+        view.wmf_addSubviewWithConstraintsToEdges(animationView)
+    }
+    
+    override func didMove(toParentViewController parent: UIViewController?) {
+        super.didMove(toParentViewController: parent)
+        view.superview?.layoutIfNeeded() // Fix for: http://stackoverflow.com/a/39614714
```
joewalsh: Seems like this should be fixed in the parent view controller rather than having this VC reach outside of itself to tell a superview to layout.

Wikipedia/Code/WMFWelcomeAnimationViewControllers.swift
```diff
@@ -0,0 +1,85 @@
+class WMFWelcomeAnimationViewController: UIViewController {
+    var welcomePageType:WMFWelcomePageType = .intro
+    fileprivate var hasAlreadyAnimated = false
+    
+    open lazy var animationView: WMFWelcomeAnimationView? = nil
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        guard let animationView = animationView else {return}
```
joewalsh: There should always be a newline for braces, especially in a guard. It needs to be easy to spot (and set a breakpoint inside of) an early return.

Wikipedia/Code/WMFWelcomeAnimationViewControllers.swift
```diff
@@ -0,0 +1,85 @@
...
+        animationView.translatesAutoresizingMaskIntoConstraints = false
+        view.wmf_addSubviewWithConstraintsToEdges(animationView)
+    }
+    
+    override func didMove(toParentViewController parent: UIViewController?) {
+        super.didMove(toParentViewController: parent)
+        view.superview?.layoutIfNeeded() // Fix for: http://stackoverflow.com/a/39614714
+        guard let animationView = animationView else {return}
```
joewalsh: Same comment as above. Also update the other guards in this PR

Wikipedia/Code/WMFWelcomeAnimationViewControllers.swift
```diff
@@ -0,0 +1,85 @@
...
+        super.didMove(toParentViewController: parent)
+        guard let animationView = animationView else {return}
+        switch welcomePageType {
+        case .exploration:
+            fallthrough
+        case .languages:
+            fallthrough
+        case .analytics:
+            animationView.backgroundColor = UIColor(0xdee6f6)
+            animationView.layer.cornerRadius = animationView.frame.size.width / 2.0
```
joewalsh: This breaks if the view is resized after moving to the parent

Wikipedia/Code/WMFWelcomeContainerViewController.swift
```diff
@@ -42,59 +45,64 @@ class WMFWelcomeContainerViewController: UIViewController {
...
+            useBottomAlignmentIfPhone()
+            if isiPhone4s() {
+                hideAndCollapseTopContainerView()
+            }
+            needsDeviceAdjustments = false
         }
     }
     
+    fileprivate func isiPhone5() -> Bool {
+        return view.frame.size.height == 568
```
joewalsh: Maybe this should be `> 400 && < 568` instead of looking for an exact match? It's possible new short devices are introduced that aren't exactly 568 points tall

Wikipedia/Code/WMFWelcomeContainerViewController.swift
```diff
@@ -42,59 +45,64 @@ class WMFWelcomeContainerViewController: UIViewController {
...
+    }
+
+    fileprivate func isiPhone4s() -> Bool {
+        return view.frame.size.height == 480
```
joewalsh: Same comment as above but with `< 480`

Wikipedia/Code/WMFWelcomeIntroductionViewController.swift
```diff
@@ -1,58 +1,33 @@
...
+        
+        descriptionLabel.text = WMFLocalizedString("welcome-intro-free-encyclopedia-description", value:"Wikipedia is written collaboratively by volunteers and consists of more than 40 million articles in nearly 300 languages.", comment:"Description for introductory welcome screen")
+
+        linkLabel.text = WMFLocalizedString("welcome-intro-free-encyclopedia-more", value:"Learn more about Wikipedia", comment:"Text for link for learning more about Wikipedia on introductory welcome screen")
+        
+        linkLabel.textColor = .wmf_blue
+        
+        view.wmf_configureSubviewsForDynamicType()
+        
+        linkLabel.addGestureRecognizer(UITapGestureRecognizer.init(target: self, action: #selector(showLearnMoreAlert(_:))))
```
joewalsh: same comment as above on button vs label

Wikipedia/Code/WMFWelcomeAnimationViewControllers.swift
joewalsh: I have the fix for this if you'd rather me just commit it, let me know. The custom view can handle all of this in `layoutSubviews`. You could also adjust the corner radius property there.

Wikipedia/Code/WMFWelcomeAnalyticsViewController.swift
```diff
@@ -2,11 +2,30 @@
...
+    @IBOutlet fileprivate var getStartedButton:UIButton!
+
+    @IBOutlet fileprivate var descriptionLabel:UILabel!
+    @IBOutlet fileprivate var linkLabel:UILabel!
     
     override func viewDidLoad() {
         super.viewDidLoad()
-                
+
+        // TODO: figure out how to make this color public - naming and such (it's in the theme colors)
```
joewalsh: It's already public as `wmf_green`

Wikipedia/Code/WMFWelcomePageViewController.swift
```diff
@@ -60,38 +77,75 @@ class WMFWelcomePageViewController: UIPageViewController, UIPageViewControllerDa
...
     }
+    
+    fileprivate func configureAndAddNextButton(){
+        nextButton.translatesAutoresizingMaskIntoConstraints = false
+        nextButton.addTarget(self, action: #selector(nextButtonTapped), for: .touchUpInside)
+        nextButton.isUserInteractionEnabled = true
+        nextButton.titleLabel?.adjustsFontSizeToFitWidth = true
+        nextButton.setTitle(CommonStrings.nextTitle, for: .normal)
+        nextButton.setTitleColor(.wmf_blue, for: .normal)
+        // TODO: figure out what to do about 0xA2A9B1 - not sure what to name this gray based on the other gray names - we already have  'light(er|est)Gray'
```
joewalsh: Ask Carolyn if we can use one of the existing gray colors, otherwise `lightererGray`

Wikipedia/Code/WMFWelcomeLanguageTableViewCell.swift
```diff
@@ -1,5 +1,5 @@
-
-class WMFWelcomeLanguageTableViewCell: WMFCustomDeleteButtonTableViewCell {
+//TODO: should this be themable?
```
joewalsh: If welcome is always in the light theme, it doesn't need to be for now.

Wikipedia/Code/WMFWelcomeAnalyticsViewController.swift
montehurd: Formerly was having trouble getting buttons with longer strings to wrap lines properly (w long translations and or big dynamic type size), but I can re-check to see if this is still an issue.

Wikipedia/Code/WMFWelcomeAnimationBackgroundView.swift
montehurd: Fixed.

Wikipedia/Code/WMFWelcomeAnimationBackgroundView.swift
montehurd: Ya formerly custom CSShapeLayers were being animated as well, but no longer. I can switch, but it doesn't make any visual difference.

Wikipedia/Code/WMFWelcomeAnimationViewControllers.swift
montehurd: Fixed.

Wikipedia/Code/WMFWelcomeAnimationViewControllers.swift
montehurd: Fixed.

Wikipedia/Code/WMFWelcomePageViewController.swift
montehurd: lol :)

Wikipedia/Code/WMFWelcomeAnimationViewControllers.swift
montehurd: @joewalsh 

Turns out this was needed to fix the issue online in the comment's stack overflow link.

You can see the difference on iPhone 5 - we tweak constraints in `reduceTopAnimationsSizes`.

With it the animations appear with reduced size:
screen shot 2017-10-02 at 4 35 52 pm

Without it they don't:
screen shot 2017-10-02 at 4 37 56 pm

montehurd: iirc this started happening in iOS8... not sure if there's a better work-around - open to ideas... but this fix seems to have been working ok

montehurd: i think it's specific to using container views...

Wikipedia/Code/WMFWelcomeContainerViewController.swift
montehurd: Fixed!

Wikipedia/Code/WMFWelcomeContainerViewController.swift
montehurd: Fixed!

Wikipedia/Code/WMFWelcomeAnalyticsViewController.swift
montehurd: Fixed!

Wikipedia/Code/WMFWelcomeLanguageTableViewCell.swift
montehurd: That's what I suspected.

Wikipedia/Code/WMFWelcomePageViewController.swift
montehurd: Fixed. She said it could just use lighterGray.

montehurd:
@joewalsh I think I got mostly everything addressed except the UIButtons. I added that to the checklist so I don't forget to do it with the other follow-up changes.

Wikipedia/Code/WMFWelcomeContainerViewController.swift
montehurd: Not using UIScreen.main.bounds.size.height for these ended up being the real reason why my code reducing the size of the top animation views didn't work without calling view.superview?.layoutIfNeeded() here: https://github.com/wikimedia/wikipedia-ios/pull/1803/files#diff-7b043c3999562fd7b1ebad3e7dcc4244L31

PR #1805, opened 2017-09-19
Convert Saved, History, Disambiguation pages, and search table views to collection views

joewalsh:
Any issues with swipe actions (cells not closing or opening , etc) are expected and will be fixed with follow on PRs.

This PR is for the first chunk of swipe action work (with known issues) along with the broader changes made to utilize collection views (and the new swipe action code) across the app.


montehurd:
Testing now...

- "Save for later" button on feed goes to article instead of saving.

montehurd:
- If you've scrolled down in search results, then enter new search term, the list isn't scrolled to the top when the new results appear

WMF Framework/UIFont+WMFDynamicType.swift
```diff
@@ -195,20 +198,28 @@ public extension UIFont {
         let styleTable: [UIContentSizeCategory:CGFloat]? = familyTable?[style]
         let size: CGFloat = styleTable?[preferredContentSizeCategory] ?? 21
 
+        let cacheKey = "\(fontFamily.rawValue)-\(size)"
```
montehurd: ooh nice!

PR #1817, opened 2017-09-22
Update app to recheck feed more often to prevent vandalism from being retained

nambatee:
https://phabricator.wikimedia.org/T175329


Wikipedia/Code/WMFAppViewController.m
```diff
@@ -597,6 +594,22 @@ - (void)finishResumingApp {
 #endif
 }
 
+- (NSTimeInterval)WMFTimeBeforeRefreshingExploreFeed {
```
joewalsh: should be `timeBeforeRefreshingExploreFeed` to match method naming conventions

Wikipedia/Code/WMFFeedContentFetcher.m
```diff
@@ -90,6 +90,17 @@ - (void)fetchFeedContentForURL:(NSURL *)siteURL date:(NSDate *)date force:(BOOL)
                                           userInfo:nil]);
 
             } else {
+                NSHTTPURLResponse *response = ((NSHTTPURLResponse *)[operation response]);
+                NSDictionary *headers = [response allHeaderFields];
+                NSString *cacheControllHeader = headers[@"Cache-Control"];
+
+                NSRange range1 = [cacheControllHeader rangeOfString:@"max-age="];
+                NSRange range2 = NSMakeRange(cacheControllHeader.length, 1);
+                NSRange range = NSMakeRange(range1.location + range1.length, range2.location - range1.location - range1.length);
```
joewalsh: `range2.length` is never used - can `range2.location` just be `cacheControlHeader.length`?

Wikipedia/Code/WMFFeedContentSource.m
```diff
@@ -103,6 +103,23 @@ - (void)fetchContentForDate:(NSDate *)date force:(BOOL)force completion:(void (^
         }
         success:^(WMFFeedDayResponse *_Nonnull feedDay) {
 
+            NSString *key = [WMFFeedDayResponse WMFFeedDayResponseMaxAgeKey];
+            NSFetchRequest *request = [WMFKeyValue fetchRequest];
+            request.predicate = [NSPredicate predicateWithFormat:@"key == %@", key];
+            request.fetchLimit = 1;
+            NSManagedObjectContext *moc = self.userDataStore.viewContext;
```
joewalsh: It's possible this isn't the main thread, so `viewContext` shouldn't be used - you can move all of this into  `- (void)saveContentForFeedDay:(WMFFeedDayResponse *)feedDay pageViews:(NSDictionary<NSURL *, NSDictionary<NSDate *, NSNumber *> *> *)pageViews onDate:(NSDate *)date inManagedObjectContext:(NSManagedObjectContext *)moc completion:(dispatch_block_t)completion` and use the provided context

joewalsh: (inside of the `performBlock:^{`)

Wikipedia/Code/WMFFeedContentFetcher.m
```diff
@@ -90,6 +90,17 @@ - (void)fetchFeedContentForURL:(NSURL *)siteURL date:(NSDate *)date force:(BOOL)
...
+                NSInteger maxAge = [[cacheControllHeader substringWithRange:range] intValue];
```
joewalsh: This will fail if the order changes - for example if the header becomes `max-age=10, s-maxage=100`, should be updated to handle this case - maybe a regex?

joewalsh: This will fail if the order changes - for example if the header becomes `max-age=10, s-maxage=100`, should be updated to handle this case - maybe a regex?

nambatee: I swapped it for `(?<=max-age=)\\d{2}` but is there a way to match all the digits after the string, not just 2?

Wikipedia/Code/WMFFeedContentSource.m
```diff
@@ -202,6 +202,23 @@ - (void)removeAllContentInManagedObjectContext:(NSManagedObjectContext *)moc {
...
+        NSArray<WMFKeyValue *> *results = [moc executeFetchRequest:request error:nil];
+        WMFKeyValue *keyValue = results.firstObject;
+        
+        if (keyValue == nil) {
+            NSEntityDescription *entity = [NSEntityDescription entityForName:@"WMFKeyValue" inManagedObjectContext:moc];
+            keyValue = [[WMFKeyValue alloc] initWithEntity:entity insertIntoManagedObjectContext:moc];
+            keyValue.key = key;
+            NSNumber *value = @(feedDay.maxAge);
+            keyValue.value = value;
+            [moc save:nil];
```
joewalsh: The save here is unnecessary, the moc will be saved after all the content is imported.

PR #1819, opened 2017-09-23
T174840. Settings > Appearance does slight upward animation when you tap

vpeschenkov:
The [issue] occurs only on iOS 11 or later. In order to get the needed behavior, we'd add the code like following:
```
if (@available(iOS 11, *)) {
    self.contentInsetAdjustmentBehavior = UIScrollViewContentInsetAdjustmentNever;
}
``` 
However, the code should be added to the all instances of ```UIScrollView```, ```UITableView```, etc.

I’ve written ```UIScrollView+WMFAdjustmentBehavior``` category, that replaces```-initWithFrame:``` and ```-initWithCoder:``` methods, and also sets ```contentInsetAdjustmentBehavior``` to ```Never``` (The default value is ```Automatic```). The category changes just the default value and we still can set that value.

I've used Objective-C, because, I couldn't do similar code on Swift. It's possible, but the result seems less beautiful than on Objective-c (In Swift 4 isn't possible overrides ```+load``` or ```+initialize``` methods). 

| Before           | After           |
| ---------------- | --------------- |
| ![][BeforeImage] | ![][AfterImage] |

[issue]: https://phabricator.wikimedia.org/T174840
[AfterImage]: https://monosnap.com/file/d00fN2c3mSeLT26JMyXrIJ67ioN9uR.png
[BeforeImage]: https://monosnap.com/file/VeqlceiQjEnLxVzYglAv9CpPnD1DWI.png

Also, I've fixed this bug on iPhone X:

| Before                       | After                           |
| ---------------------------- | ------------------------------- |
| ![][iPhoneXFixedImageBefore] | ![][iPhoneXFixedImageAfter]     |

[iPhoneXFixedImageAfter]: https://monosnap.com/file/ck9bWPq7TRF7VPX29EryGsiOXh4em1.png
[iPhoneXFixedImageBefore]: https://monosnap.com/file/465ooc9Yy77JJ4LHomiGMb7tlxGU3D.png


joewalsh:
@vpeschenkov swizzling the initializers and globally disabling inset adjustment could have a lot of unintended consequences. Generally, it's best to either subclass or adjust the properties of objects where needed. There's a lower-impact workaround I found thanks to this thread https://twitter.com/marcoarment/status/912056135435374592 . I've opened a PR here:  https://github.com/wikimedia/wikipedia-ios/pull/1823

PR #1834, opened 2017-09-28
In Settings view cell background color (esp. on Dark / Sepia mode) should extend to fill the full width on landscape

nambatee:
https://phabricator.wikimedia.org/T176784

### DIFF
--- a/Wikipedia/Code/AppearanceSettingsViewController.swift
+++ b/Wikipedia/Code/AppearanceSettingsViewController.swift
@@ -103,6 +103,13 @@ open class AppearanceSettingsViewController: UIViewController, UITableViewDataSo
                 cell.backgroundColor = vc.view.backgroundColor
             }
             
+            if vc is ImageDimmingExampleViewController {
+                let bottomView = UIView(frame: CGRect(x: 0, y: cell.frame.height - ImageDimmingExampleViewController.bottomSpacing, width: cell.bounds.width, height: ImageDimmingExampleViewController.bottomSpacing))
+                bottomView.autoresizingMask = [.flexibleWidth]
+                bottomView.backgroundColor = tableView.backgroundColor
+                cell.addSubview(bottomView)
+            }
+            
             cell.selectionStyle = .none
             return cell
         }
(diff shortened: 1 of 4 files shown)


Wikipedia/Code/AppearanceSettingsViewController.swift
```diff
@@ -205,6 +205,12 @@ open class AppearanceSettingsViewController: UIViewController, UITableViewDataSo
                 break
             }
             
+        } else if let customViewItem = sections[indexPath.section].items[indexPath.item] as? AppearanceSettingsCustomViewItem {
+            if customViewItem.viewController is ImageDimmingExampleViewController {
+                let bottomView = UIView(frame: CGRect(x: 0, y: cell.frame.height - ImageDimmingExampleViewController.bottomSpacing, width: cell.frame.width, height: ImageDimmingExampleViewController.bottomSpacing))
+                bottomView.backgroundColor = tableView.backgroundColor
+                cell.addSubview(bottomView)
```
montehurd: Usually best to avoid adding subviews in `willDisplay`... could you set the bg color on the white views on the sides of https://phab.wmfusercontent.org/file/data/dblft4xffury3xn2exz4/PHID-FILE-gxcq6s2d73u2s7cpghbc/IMG_0024.PNG ?

nambatee: The bg color is already set. This is adding a view that makes it look like there's spacing between the cells.

montehurd: If you set a breakpoint on line 212 where it's adding this view, then rotate landscape and scroll up and down, it's hitting that breakpoint each time you scroll.

montehurd: It's also off when initially rotating from portrait to landscape:

screen shot 2017-09-28 at 4 23 17 pm

montehurd: > The bg color is already set. This is adding a view that makes it look like there's spacing between the cells.

Here's the view I'm talking about:

screen shot 2017-09-28 at 4 43 40 pm

Looks like the spacing is already there, the view is just the wrong color:

screen shot 2017-09-28 at 4 40 25 pm

montehurd: Oh I see now - the lighter color extends to the edges.

I'd maybe try adding a view to ImageDimmingExampleViewController.xib for the lighter color, then set that view's color in ImageDimmingExampleViewController.swift's `apply(theme: Theme)`.

joewalsh:
jenkins, retest this please

PR #1845, opened 2017-10-04
Follow-up Welcome tweaks

montehurd:
Follow-ups for https://phabricator.wikimedia.org/T174207

- [x] fix missing "PRIMARY" lang indicator (use the WMFLanguageCell which already has it)
- [x] allow WMFLanguageCell collapse side padding (needed on Welcome lang screen)
- [x] fix uipagecontrol dots on iphone x
- [x] fix being able to go landscape on iPhone X
- [x] hide battery indicator
- [x] add top animation bottom space constraint of 20 for everything except iPhone 5 and 4s per Carolyn
- [x] Joe's comment here: https://github.com/wikimedia/wikipedia-ios/pull/1803#pullrequestreview-66720366
- [x] use smaller title font on 4s and 5?
- [x] fix transparency on the new initial vc - on sides when you drag panels there's a white margin
- [x] dynamic type hardening - exempt the title - just use fixed size
- [x] update lang picker popup title (Wikipedia languages - everywhere the lang picker is used is fine)
- [x] use UIButtons instead of labels 
- [x] fix shadow artifact on last welcome screen's UISwitch
- [ ] fix incorrect tint on "X" close button and nav bar background when you tap "Add or edit preferred languages > Add another language" from 3rd welcome screen
- [ ] fix weird crash when changing dynamic type size with accessibility inspector


montehurd:
@joewalsh @nambatee feel free to fix those last 2 checkbox items - i didn't have time to chase them down today :)

PR #1847, opened 2017-10-04
[T176800] Fix alert layout on iPhone X

joewalsh:
https://phabricator.wikimedia.org/T176800

Switches from TSMessage to RMessage library. Updates color schemes.


montehurd:
There's one alert which doesn't seem to be appearing - the "Loading content..." alert shown when you tap an edit pencil. Called in `loadLatestWikiTextForSectionFromServer`. Noticed on iPhone X.

montehurd:
iirc we can delete the Hex Colors third party code now too - i think TSMessage was the only thing which used it.

PR #1849, opened 2017-10-04
Don't use custom chevron for back button

nambatee:
https://phabricator.wikimedia.org/T176809


nambatee:
simulator screen shot - iphone x - 2017-10-04 at 14 45 50
simulator screen shot - iphone x - 2017-10-04 at 14 46 02

Leaving the asset in place as it's still being used in `AboutViewController` and `PreviewAndSaveViewController`

PR #1850, opened 2017-10-04
Remove Masonry

joewalsh:
https://phabricator.wikimedia.org/T175373


Wikipedia/Code/WMFReferencePanelViewController.swift
```diff
@@ -55,13 +55,12 @@ class WMFReferencePanelViewController: UIViewController, Themeable {
...
-        containerController.willMove(toParentViewController: self)
-        containerController.view.translatesAutoresizingMaskIntoConstraints = false
-        containerView.addSubview(containerController.view!)
-        containerView.bringSubview(toFront: containerController.view!)
-        containerController.view.mas_makeConstraints { make in
-            _ = make?.top.bottom().leading().and().trailing().equalTo()(self.containerView)
+        guard let view = containerController.view else {
+            return
         }
+        containerController.willMove(toParentViewController: self)
```
montehurd: do we need matching "didmove" call?

Wikipedia/Code/PreviewWebViewContainer.m
```diff
@@ -58,14 +57,11 @@ - (void)awakeFromNib {
     [super awakeFromNib];
     WKWebView *webview = [[WKWebView alloc] initWithFrame:CGRectZero configuration:[self configuration]];
     webview.translatesAutoresizingMaskIntoConstraints = NO;
-    [self addSubview:webview];
+    [self wmf_addSubviewWithConstraintsToEdges:webview];
```
montehurd: The `wmf_addSubview` way using left and right instead of leading and trailing. Should be just switch it to use leading and trailing too?

joewalsh: It doesn't really matter in this case since both edges are zero- it's just potentially confusing to have non-directional insets (`UIEdgeInsets` - `NSDirectionalInsets` is iOS 11 only) and directional layout anchors for the main method. We can try leading/trailing and see if there are any issues since that should be the default over left/right.

PR #1852, opened 2017-10-05
Swipe refinements

joewalsh:
- Fix divide by zero issue (disappearing cell on swipe)
- Add icons
- Implement accessibility actions

https://phabricator.wikimedia.org/T174713
https://phabricator.wikimedia.org/T177487


nambatee:
In Top Read "Save" action is not being swapped for "Unsave"

notsaved

PR #1853, opened 2017-10-05
Fix layout margins on iOS 10

joewalsh:


joewalsh:
Added a max readable width to match the previous behavior, fixed theme-ing of action views

PR #1856, opened 2017-10-05
Use "death" weights for "On this day" explore card heuristic scoring on EN.

montehurd:
https://phabricator.wikimedia.org/T169941

Non-EN uses event overall image count same way as before.

### DIFF
--- a/Wikipedia/Code/WMFOnThisDayContentSource.m
+++ b/Wikipedia/Code/WMFOnThisDayContentSource.m
@@ -99,7 +99,6 @@ NS_ASSUME_NONNULL_BEGIN
         NSDateComponents *components = [[NSCalendar wmf_gregorianCalendar] components:NSCalendarUnitDay | NSCalendarUnitMonth | NSCalendarUnitYear fromDate:date];
         NSInteger month = [components month];
         NSInteger day = [components day];
-        NSInteger year = [components year];
         @weakify(self)
             [self.fetcher fetchOnThisDayEventsForURL:self.siteURL
                 month:month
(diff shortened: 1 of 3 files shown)


Wikipedia/Code/WMFOnThisDayContentSource.m
```diff
@@ -120,24 +119,18 @@ - (void)loadContentForDate:(NSDate *)date inManagedObjectContext:(NSManagedObjec
...
                         }];
 
                         NSInteger featuredEventIndex = NSNotFound;
 
                         NSArray *eventsSortedByScore = [onThisDayEvents sortedArrayUsingDescriptors:@[[NSSortDescriptor sortDescriptorWithKey:@"score" ascending:NO]]];
                         if (eventsSortedByScore.count > 0) {
-                            NSInteger index = ((year % 10) % eventsSortedByScore.count);
-                            WMFFeedOnThisDayEvent *featuredEvent = eventsSortedByScore[index];
-                            featuredEventIndex = featuredEvent.index.integerValue;
+                            featuredEventIndex = ((WMFFeedOnThisDayEvent*)eventsSortedByScore.firstObject).index.integerValue;
```
joewalsh: Removing the `year % 10` means you'll likely repeat the same events every year

montehurd: Assuming the event text and article previews image counts remained static over the course of a year that would definitely be true. Not sure it's likely, but I can try to make the score use some weight based on current year... will have to think a bit about how that will play with the current mechanism... will take a stab at it :)

montehurd: @joewalsh 

So if I just leave this as it was...

     NSInteger index = ((year % 10) % eventsSortedByScore.count);
     WMFFeedOnThisDayEvent *featuredEvent = eventsSortedByScore[index];		
     featuredEventIndex = featuredEvent.index.integerValue;

... it seems like it would be ok. Guessing that's better than moving any current year logic into `calculateScore`?

Wikipedia/Code/WMFOnThisDayContentSource.m
```diff
@@ -130,7 +131,12 @@ - (void)loadContentForDate:(NSDate *)date inManagedObjectContext:(NSManagedObjec
 
                         NSArray *eventsSortedByScore = [onThisDayEvents sortedArrayUsingDescriptors:@[[NSSortDescriptor sortDescriptorWithKey:@"score" ascending:NO]]];
                         if (eventsSortedByScore.count > 0) {
-                            featuredEventIndex = ((WMFFeedOnThisDayEvent*)eventsSortedByScore.firstObject).index.integerValue;
+                            // Rotate through the 10 highest scoring events based on current year. Ensures unlikely to see a repeat for at least 10 years
```
montehurd: @joewalsh I used some of your comments from the phab ticket for future reference.

PR #1862, opened 2017-10-07
Various Welcome improvements for larger dynamic text sizes.

montehurd:
Nothing should appear very different at normal text sizes, but Carolyn and I identified various things which needed tweaks at larger sizes (and/or long translations - especially on smaller screen sizes). 

- Titles now scroll with content.
- Scroll indicator no longer overlaps edge of content (was covering things like the "PRIMARY" lang indicator).
- Switch now fully left aligned (formerly doing this caused its shadow to be clipped a bit).
- Small top and bottom gradient on scrollview (can only see if content is so large you need to scroll - additional visual hint that you can scroll)

Some things had to be moved around and containerized slightly differently, but overall this approach is more flexible (yay stack views!). I have a couple more tweaks coming. I'll amend or open new PR if this one gets merged.


toppole69:
U

Wikipedia/Code/WMFWelcomePanelViewController.swift
```diff
@@ -15,9 +18,15 @@ class WMFWelcomePanelViewController: UIViewController {
         if view.bounds.size.height <= 568 {
             titleLabel.font = UIFont.systemFont(ofSize: 28, weight: .medium)
         }
-            
+        
+        nextButton.backgroundColor = .wmf_blue
```
joewalsh: this should be pulled from the theme

PR #1864, opened 2017-10-10
Language indicator line is below the search area with large dynamic text sizes

nambatee:
https://phabricator.wikimedia.org/T176813


joewalsh:
It looks like this caused the language bar to always be visible, even if it's empty.

PR #1868, opened 2017-10-11
More Welcome tweaks

montehurd:
More issues noticed at larger dynamic text sizes, w German strings etc
- [x] make skip/next button text not go off sides of screen if text large (but until that point maintain consistent horizontal spacing from the center)
- [x] make skip/next button not truncate
- [x] fix bug with dynamic type method setting unrelated props on UIButton's label 
- [x] make buttons on 1st, 3rd and 4th screens wrap to 2 lines instead of shrinking if string gets super long

### DIFF
--- a/Wikipedia/Code/ColumnarCollectionViewController.swift
+++ b/Wikipedia/Code/ColumnarCollectionViewController.swift
@@ -39,11 +39,15 @@ class ColumnarCollectionViewController: UICollectionViewController, Themeable {
                 cellWithSubItems.deselectSelectedSubItems(animated: animated)
             }
         }
+        NotificationCenter.default.addObserver(forName: .UIContentSizeCategoryDidChange, object: nil, queue: nil) {_ in
+            self.collectionView?.reloadData()
+        }
     }
     
     override func viewDidDisappear(_ animated: Bool) {
         super.viewDidDisappear(animated)
         unregisterForPreviewing()
+        NotificationCenter.default.removeObserver(self, name: .UIContentSizeCategoryDidChange, object: nil)
     }
     
     // MARK - Cell & View Registration
--- a/Wikipedia/Code/ColumnarCollectionViewController.swift
+++ b/Wikipedia/Code/ColumnarCollectionViewController.swift
@@ -39,11 +39,15 @@ class ColumnarCollectionViewController: UICollectionViewController, Themeable {
                 cellWithSubItems.deselectSelectedSubItems(animated: animated)
             }
         }
+        NotificationCenter.default.addObserver(forName: .UIContentSizeCategoryDidChange, object: nil, queue: nil) {_ in
+            self.collectionView?.reloadData()
+        }
     }
     
     override func viewDidDisappear(_ animated: Bool) {
         super.viewDidDisappear(animated)
         unregisterForPreviewing()
+        NotificationCenter.default.removeObserver(self, name: .UIContentSizeCategoryDidChange, object: nil)
     }
     
     // MARK - Cell & View Registration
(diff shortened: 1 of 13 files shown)


montehurd:
I removed the properties we talked about from `wmf_configureSubviewsForDynamicType` and didn't notice any issues so far elsewhere when increasing dynamic type sizes. (there was an issue with collection views not being reloaded when the text size changed but I don't think it was related to these props)

Wikipedia/Code/ColumnarCollectionViewController.swift
```diff
@@ -39,11 +39,15 @@ class ColumnarCollectionViewController: UICollectionViewController, Themeable {
                 cellWithSubItems.deselectSelectedSubItems(animated: animated)
             }
         }
+        NotificationCenter.default.addObserver(forName: .UIContentSizeCategoryDidChange, object: nil, queue: nil) {_ in
+            self.collectionView?.reloadData()
```
joewalsh: this can probably be put inside of `traitCollectionDidChange`

Wikipedia/Code/ColumnarCollectionViewController.swift
```diff
@@ -39,11 +39,15 @@ class ColumnarCollectionViewController: UICollectionViewController, Themeable {
...
+        }
     }
     
     override func viewDidDisappear(_ animated: Bool) {
         super.viewDidDisappear(animated)
         unregisterForPreviewing()
+        NotificationCenter.default.removeObserver(self, name: .UIContentSizeCategoryDidChange, object: nil)
```
joewalsh: can be removed after addressing comment above

Wikipedia/Code/ColumnarCollectionViewController.swift
montehurd: ah thanks lemme try that...

Wikipedia/Code/ColumnarCollectionViewController.swift
montehurd: Fixed!

Wikipedia/Code/ColumnarCollectionViewController.swift
montehurd: Fixed!

montehurd:
@joewalsh oh wait I just noticed an issue with moving the re-load to traitCollectionDidChange...

montehurd:
@joewalsh the problem with having it in `traitCollectionDidChange` is that also gets called on rotation and it makes the rotation animation do this weird fade. I think if it's ok I'd prefer it to not be in `traitCollectionDidChange` so it only happens explicitly when the text size has changed...

joewalsh:
@montehurd does the resize issue happen when resizing from settings or only from the accessibility inspector? If it's only from the accessibility inspector, you can revert your original change as well.

montehurd:
> @montehurd does the resize issue happen when resizing from settings or only from the accessibility inspector? If it's only from the accessibility inspector, you can revert your original change as well.

good point...checking...

montehurd:
@joewalsh 

Ya I'm seeing funkiness on device too.

Recorded from my phone: 

 # Before: #
before111 mov

# After: #
after111 mov

joewalsh:
@montehurd the observer removal you have in didDisappear doesn't work with the block-based addObserver. The explore VC has the same problem. Also, if the observer was actually removed on didDisappear, changes that happened while the view wasn't visible wouldn't trigger the update. As it stands, a new block based observer is added every time the views appear, and never removed.

The observer should be added in viewDidLoad and removed in deinit (swift) / dealloc (objc). It would probably be easier to switch to the`addObserver:selector:name:object:` method instead of the block based approach. If you do keep the block based approach, be sure to make sure the block doesn't capture self causing a retain cycle (might just need to add `[unowned self]`).

montehurd:
@joewalsh gah! good catch... looking...

joewalsh:
Yeah. And the block would be called (calling reloadData) once for every time the view appeared

montehurd:
@joewalsh It's showing up here - very end of the animation you'll see Explore hasn't adjusted quite right...
weird mov

PR #1875, opened 2017-10-12
Migrate WMFContentGroup content into a separate, related entity

joewalsh:
Speeds up explore feed fetch by not requiring MTLModel inflation for every piece of explore feed detail content (for example, every "On this day" event).

https://phabricator.wikimedia.org/T166125


Wikipedia/Code/WMFExploreCollectionViewController.m
```diff
@@ -394,6 +400,7 @@ - (void)viewWillAppear:(BOOL)animated {
     }
 
     if (!self.fetchedResultsController) {
+        [self.userStore prefetchArticles]; // articles aren't linked to content groups by a core data relationship, they're fetched on demand. it helps to warm up the article cache with one big fetch instead of a lot of individual fetches.
```
montehurd: May want to test on an iPhone 4s...

Wikipedia/Code/WMFExploreCollectionViewController.m
```diff
@@ -1170,30 +1182,21 @@ - (void)configurePhotoCell:(WMFPicOfTheDayCollectionViewCell *)cell withImageInf
...
-        while (events.count > featuredIndex + 1 && previousEvent.year == event.year && attempts < 4) {
-            WMFFeedOnThisDayEvent *potentialEvent = events[featuredIndex + 1];
-            if ([potentialEvent isKindOfClass:[WMFFeedOnThisDayEvent class]]) {
-                previousEvent = potentialEvent;
-            }
-            attempts++;
+    NSArray *previewEvents = (NSArray *)contentGroup.contentPreview;
+    if ([previewEvents isKindOfClass:[NSArray class]]) {
+        WMFFeedOnThisDayEvent *event = previewEvents.firstObject;
+        WMFFeedOnThisDayEvent *previousEvent = previewEvents.count > 1 ? previewEvents.lastObject : nil;
```
montehurd: `previousEvent` should be the first event before `event` which doesn't have the same year. Otherwise on the feed card it's possible to see the same year twice. On Oct 2, for example, it should show `1990` as the faded out year at the bottom:
screen shot 2017-10-12 at 3 29 13 pm

I double-checked on develop and it's broken there too, so fix doesn't have to come with this PR, but saw this PR was changing the `previousEvent` logic, so thought I'd mention it. (also double-checked with Carolyn that that's what was wanted).

montehurd: (I'll file ticket if you'd rather fix separately)

PR #1879, opened 2017-10-12
Fix extra padding at top of settings on iOS 9 and 10.

montehurd:
Not sure where the extra content offset is coming from...

<strike>I double-checked this fix on iPhone X sim and seems ok, but would rather find out what's causing the offset... ideas? (tried symbolic breakpoint tagged to this tableview and none of our code seems to be setting it...)</strike>

Turning off `automaticallyAdjustsScrollViewInsets` for iOS 9 and 10 fixed it... still uncertain why it gets added - unless maybe it's related to the nav controller we're using for pushing the appearance vc?

# Before #
screen shot 2017-10-12 at 2 36 03 pm

# After #
screen shot 2017-10-12 at 2 34 39 pm


montehurd:
Looks like it may be `automaticallyAdjustsScrollViewInsets`...

PR #1880, opened 2017-10-13
Only show one image download error alert per gallery session

joewalsh:
https://phabricator.wikimedia.org/T177872


montehurd:
Strange that it would show as you swipe between images given that the "imageinfo" fetch is asking for info for all the gallery images for a given article... wondering if whatever other fetcher is firing on swipe can be switched to fetch for all the images too...

montehurd:
> given that the "imageinfo" fetch is asking for info for all the gallery images for a given article
I may be wrong about that...

montehurd:
Looks like it was just the request for the high res and not anything "imageinfo" related.

PR #1883, opened 2017-10-13
Hide nav bar on scroll

nambatee:


Wikipedia/Code/ExploreViewController.swift
```diff
@@ -230,6 +227,24 @@ class ExploreViewController: UIViewController, WMFExploreCollectionViewControlle
...
+    var statusBarUnderlay: UIView?
+    
+    func exploreCollectionViewController(_ collectionVC: WMFExploreCollectionViewController, willEndDragging scrollView: UIScrollView, velocity: CGPoint) {
+        self.navigationController?.setNavigationBarHidden(velocity.y > 0, animated: true)
+    }
+    
+    func addStatusBarUnderlay() {
+        let statusBarUnderlay = UIView()
+        statusBarUnderlay.translatesAutoresizingMaskIntoConstraints = false
+        let topConstraint = self.view.topAnchor.constraint(equalTo: statusBarUnderlay.topAnchor)
```
montehurd: According to the docs the subview should call `constraint(equalTo`. Same bug in the web vc's `addHeaderView`.

Wikipedia/Code/WebViewController.storyboard
```diff
@@ -38,14 +39,19 @@
                         </subviews>
                         <color key="backgroundColor" red="1" green="1" blue="1" alpha="1" colorSpace="custom" customColorSpace="sRGB"/>
                         <constraints>
-                            <constraint firstItem="WeL-Mj-Zsh" firstAttribute="leading" secondItem="kh9-bI-dsS" secondAttribute="leading" id="3Zs-jL-a48"/>
+                            <constraint firstItem="WeL-Mj-Zsh" firstAttribute="leading" secondItem="dho-2s-f1C" secondAttribute="leading" id="3Zs-jL-a48"/>
```
montehurd: What were these constraint changes? On the last branch I reverted all the changes in this file and it seemed to work the same...

Wikipedia/Code/ExploreViewController.swift
joewalsh: The `subview` in the docs is just an example. In practice, for constraints that are just `equalTo` it doesn't matter which item is first. Think of it like an equation `view.topAnchor = subview.topAnchor` is equivalent to `subview.topAnchor = view.anchor`.

Wikipedia/Code/WebViewController.storyboard
joewalsh: The status bar underlay was added and the container view constraints were updated to not be relative to margins.

Wikipedia/Code/ExploreViewController.swift
montehurd: I thought it made a difference though - you want to set the new item equal to something that already exists... that's why even in storyboard it has option for reversing first and second item...

joewalsh: It can make a difference if it's not just an equality constraint. In this case, it's fine.

Wikipedia/Code/WebViewController.storyboard
montehurd: Coolness. Will probably want to double-check layout of wikipedia zero bits...

Wikipedia/Code/ExploreViewController.swift
montehurd: This https://stackoverflow.com/a/28164113/135557  cleared it up for me... order makes a difference but only when it evaluates the constant. 

This reply also helped:

The formula is an equation, not an assignment. It's not "set firstItem.firstItemAttribute to be the result of …". It's "keep this equation true" or "keep firstItem.firstItemAttribute equal to …". That is symmetric. The system tries to find values for all of the attributes of all of the items that satisfies all of the constraints. Logically, it doesn't start with some values and work from those to compute the others.

montehurd:
Small issues:
- on the article view if you scroll then tap to stop the scrolling the nav bar re-appears
- nav doesn't disappear if you drag - you have to flick and release. safari hides nav bar in both cases

montehurd:
- title bar tap on either explore or web view should prob re-show the nav bar

montehurd:
- sometimes on iOS 9.3 if you scroll so the nav bar hides, tapping the title bar doesn't scroll to top
- on explore if the "On this day" "More historical events" footer is onscreen after scrolling stops and you tap on it when the detail view appears there is no nav bar

montehurd:
@joewalsh @nambatee 

Wondering if it would be better for the web VC to remain ignorant of the navigation controller... It looks like WMFArticleViewController was handling anything related to it before.

We could ping Corey to double-check, but iirc the nav controller bits were split out into WMFArticleViewController deliberately.

joewalsh:
Approved, but waiting to merge pending discussion & approval to pull into 5.7.1

PR #1890, opened 2017-10-25
Wire up headless section JS transforming

montehurd:
https://phabricator.wikimedia.org/T175320
https://phabricator.wikimedia.org/T175322
https://phabricator.wikimedia.org/T175321

- [x] load sections via [DocumentFragments](https://developer.mozilla.org/en-US/docs/Web/API/DocumentFragment)
- [x] apply JS transforms to these fragments before they are attached to the main DOM

Below is Obama article from saved when offline. Note - my phone *was* offline when I recorded this through quicktime, but quicktime added the title bar which shows wifi on, even though on my device wifi and data were off. (It also changes the time to '9:41' am which is the time Jobs announced the iPhone iirc.)

# Before
o old mov

# After
o new mov


www/js/sectionTransformation.js
```diff
@@ -0,0 +1,235 @@
...
+  }
+  
+  //TODO pass the hideTables params
+  wmf.tables.hideTables(fragment, article.ismain, 'pageTitle', 'infoboxTitle', 'otherTitle', 'footerTitle')
+
+  //TODO when proxy delivers section html ensure it sets both data-image-gallery and image variant widths! (at moment variant width isnt' set so images
+  //dont get widened even though i'm forcing "data-image-gallery" to true here - the css *is* being changed though)
+  var images = fragment.querySelectorAll('img').forEach(function(image){
+    // maybeWidenImage(image)
+    const isGallery = parseInt(image.width) > 70 ? 'true' : 'false'
```
montehurd: This image logic is temporary.

Wikipedia/Code/WebViewController.m
```diff
@@ -323,6 +323,17 @@ - (void)handleLateJavascriptTransformScriptMessage:(NSString *)messageString {
 
 - (void)handleArticleStateScriptMessage:(NSString *)messageString {
     if ([messageString isEqualToString:@"articleLoaded"]) {
+        
+        
+
+        
+
```
montehurd: This is also going to change - just invoked it here for testing.

montehurd:
I'm about half-way through hooking up the last footer transform - the menu. Will be deleting a bunch of code once that's done :)

Wikipedia/Code/WKWebView+LoadAssetsHtml.swift
```diff
@@ -0,0 +1,96 @@
+import WebKit
+import WMF
+
```
montehurd: @joewalsh heads-up I refactored `WKWebView+LoadAssetsHtml.m` to Swift.

I don't think it needs to call `stringByReplacingImageURLsWithProxyURLs` any more so I removed the call. See comments on this commit: https://github.com/wikimedia/wikipedia-ios/pull/1890/commits/bd9140615f1dc915f423ecaf643a0f7e556e29df

Wikipedia/Code/WebViewController.m
```diff
@@ -301,7 +301,10 @@ - (void)handleEditClickedScriptMessage:(NSDictionary *)messageDict {
 - (void)handleArticleStateScriptMessage:(NSString *)messageString {
     if ([messageString isEqualToString:@"indexHTMLDocumentLoaded"]) {
 
-        [self.webView wmf_fetchTransformAndAppendSectionsToDocument:self.article];
+        // TODO: figure out why these fragments are double-encoded - mobile front-end no longer encodes section id's so we don't need to either... let alone double-encode
```
montehurd: @joewalsh any ideas why fragments are double-encoded?

montehurd:
I spent a some time today (Sunday) on the UI testing harness branch @joewalsh started. A 2017 wwdc session helped some too, but still hashing the first paint timing details out. Getting much more comfortable with Xcode’s UI testing and its accessiblity bindings... it’s kind of fun actually :)

PR #1891, opened 2017-10-26
As a 3d touch user, I expect previews to load quickly enough to be useful

nambatee:
https://phabricator.wikimedia.org/T128650


Wikipedia/Code/ArticlePeekPreviewViewController.swift
```diff
@@ -0,0 +1,94 @@
...
+
+        self.titleLabel.text = article.displayTitle
+        self.descriptionLabel.text = article.capitalizedWikidataDescription
+        self.textLabel.text = article.snippet
+        
+        self.preferredContentSize = self.view.systemLayoutSizeFitting(CGSize(width: self.view.bounds.size.width, height: UILayoutFittingCompressedSize.height), withHorizontalFittingPriority: UILayoutPriority.required, verticalFittingPriority: UILayoutPriority.fittingSizeLevel)
+        self.parent?.preferredContentSize = self.preferredContentSize
+    }
+    
+    override func viewDidLoad() {
```
joewalsh: missing a `super.viewDidLoad()`

Wikipedia/Code/ArticlePeekPreviewViewController.swift
```diff
@@ -0,0 +1,94 @@
...
+        fetchArticle()
+        updateFonts()
```
joewalsh: Does dynamic type work as expected? `updateFonts()` might need to be in `viewWillAppear` so that the trait collection is up-to-date. If it's working fine, don't bother moving it.

Wikipedia/Code/UIViewController+Peekable.swift
```diff
@@ -0,0 +1,17 @@
+import UIKit
+
+extension UIViewController {
+    // Adds a peekable view controller on top of another view controller. Since the view controller is the one we want to show when the user peeks through, we're saving time by loading it behind the peekable view controller.
+    @objc func wmf_setupPeekable(_ peekableViewController: UIViewController, on viewController: UIViewController) -> UIViewController {
```
joewalsh: you could remove the `onViewController` param and return value if you assume you're calling this on the VC that is getting the peekable added - `wmf_addPeekable(_` then use `self` instead of `viewController`

montehurd:
@nambatee This is sooooo smooth!!! Very nice!

Wikipedia/Code/ArticlePeekPreviewViewController.swift
```diff
@@ -0,0 +1,104 @@
...
+            self.leadImageView.wmf_setImage(with: imageURL, detectFaces: true, onGPU: true, failure: { (error) in
+                self.leadImageView.isHidden = true
+            }, success: {
+                //handle success
+            })
+        } else {
+            leadImageView.isHidden = true
+        }
+
+        self.titleLabel.text = article.displayTitle
```
montehurd: This is a nitpick, but iirc you don't need 'self' for these IBOutlets.

Wikipedia/Code/UIViewController+Peekable.swift
```diff
@@ -0,0 +1,28 @@
...
+protocol Peekable {
+    
+}
+
+extension UIViewController {
+    // Adds a peekable view controller on top of another view controller. Since the view controller is the one we want to show when the user peeks through, we're saving time by loading it behind the peekable view controller.
+    @objc func wmf_addPeekableChildViewController(for articleURL: URL, dataStore: MWKDataStore, theme: Theme) {
+        let articlePeekPreviewViewController = ArticlePeekPreviewViewController(articleURL: articleURL, dataStore: dataStore, theme: theme)
+        
+        self.addChildViewController(articlePeekPreviewViewController)
```
montehurd: May not need 'self' here either.

montehurd:
Oh may want to do a quick RTL check if you haven't already.

Wikipedia/Code/UIViewController+Peekable.swift
nambatee: 👍 thanks!

PR #1896, opened 2017-10-27
Default to any orientation and autorotate to YES

joewalsh:
Potential fix for landscape launch bug


montehurd:
On phone iirc we want the welcome screens portrait - would that still be the case?

joewalsh:
Yes, the welcome screens are still portrait

PR #1900, opened 2017-10-31
Text size slider is not usable when using voiceover

nambatee:


nambatee:
Closing in favor of #1901

PR #1909, opened 2017-11-03
Add build scheme for page library development

joewalsh:
- Link and unlink commented out in the script for now (you have to manually link pagelib before running, and manually unlink when you're done)

- Should automatically rebuild pagelib and update assets for the app on build and run of the "Page Library Development" scheme

- Added package-lock.json since those are usually committed to the repo


montehurd:
I forgot to link and ran it and I'm still seeing assets changes, which doesn't seem right. Ideas?

montehurd:
This is 💯.

PR #1915, opened 2017-11-07
Use ArticleFullWidthImageCollectionViewCell for peek preview

nambatee:
We can reuse existing `ArticleFullWidthImageCollectionViewCell` instead of adding a new .xib. If I can't make this PR work, #1916 is using the original peek view controller. Opening 2 PRs so we can decide which approach is better.


nambatee:
- [ ] Set header view (titleLabel and descriptionLabel) bg color to `theme.colors.midBackground` 

simulator screen shot - iphone x - 2017-11-06 at 21 02 12

simulator screen shot - iphone x - 2017-11-06 at 21 03 04

PR #1916, opened 2017-11-07
Design tweaks for peek previews

nambatee:
https://phabricator.wikimedia.org/T128650#3739182
https://phabricator.wikimedia.org/T179746

This PR addresses design review comments. It is using the original peek preview view controller. #1915 is using existing `ArticleFullWidthImageCollectionViewCell`. Opening 2 PRs so we can decide which approach is better.


nambatee:
Closing in favor of #1915.

PR #1922, opened 2017-11-09
Enable scrolls to top after peek with no pop

nambatee:
https://phabricator.wikimedia.org/T180059

Is this breaking something that I'm not seeing?


montehurd:
iirc it was something like only one scroll view on nav stack should have this on. not sure if os updates have changed this. i’m afk but title bar taps i’d test:
- toc (since more than one scroll view is onscreen - webview’s being other one), phone and tablet
- article when you have a few articles on nav stack

joewalsh:
@nambatee feel free to revert my commit, but it looks like none of the recursive application of `scrollsToTop` settings are necessary anymore. I tested on iPhone X/iOS 11 - peek/pop, explore (from W  button, status bar, and tab bar), articles, history, and the edit VC. Also 

If we do have a VC with more than one scroll view visible that does break scroll to top, we can address it in that specific VC instead of generally.

nambatee:
Looks great, thanks @joewalsh!

PR #1924, opened 2017-11-09
As a user who searches in multiple languages I'd like to be alerted to the option to add an additional search language from the search screen

nambatee:
https://phabricator.wikimedia.org/T176653


Wikipedia/Code/WMFSearchLanguagesBarViewController.swift
```diff
@@ -121,6 +126,21 @@ class WMFSearchLanguagesBarViewController: UIViewController, WMFPreferredLanguag
         }
     }
     
+    fileprivate func showMoreLanguagesIfNecessary() {
```
montehurd: Could we call this `showMoreLanguagesTooltipIfNecessary`?

Wikipedia/Code/WMFSearchLanguagesBarViewController.swift
```diff
@@ -121,6 +126,21 @@ class WMFSearchLanguagesBarViewController: UIViewController, WMFPreferredLanguag
         }
     }
     
+    fileprivate func showMoreLanguagesTooltipIfNecessary() {
+        guard !hidden && languageBarLanguages().count >= 2 && !UserDefaults.standard.wmf_didShowMoreLanguagesTooltip() else {
+            return
+        }
+        UserDefaults.standard.wmf_setDidShowMoreLanguagesTooltip(true)
+        self.perform(#selector(showMoreLanguagesTooltip), with: nil, afterDelay: 1.0)
```
joewalsh: needs an equivalent `NSObject.cancelPreviousPerform...` call in `deinit` or `willDisappear` - otherwise it can be sent after this vc is deallocated, causing a crash. You could move the `wmf_setDidShow...` to `showMoreLangugaesTooltip` to ensure it's shown even if the call is cancelled.

PR #1925, opened 2017-11-10
First section appearance performance test

montehurd:
https://phabricator.wikimedia.org/T179436

Adds a test which, going forward, should raise a flag if our first section appearance performance regresses.

Loads Obama article from test fixture.

Configures a web view controller to fire two events:
- `startTime` when `.atDocumentEnd` fires, which happens when index.html's DOM is inflated, but **before** our sections are loaded
- `firstSectionAppearance` using method outlined [here](https://davidwalsh.name/detect-node-insertion) when first section appears

Measures difference in time between these two events.

Confirmed effectiveness by measuring time it took to load second section after increasing the 'head-start delay' used after first section is loaded. The tests reflected this increase.


montehurd:
@joewalsh Looks like there's an issue on develop with localized strings. A weird file appears in the Latvian strings folder and compilation hoses...

Wikipedia/Code/WebViewController.h
```diff
@@ -27,6 +28,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (nonatomic) BOOL navBarHidden;
 
+@property (nonatomic, copy, nullable) void (^wkUserContentControllerTestingConfigurationBlock)(WKUserContentController *);
```
joewalsh: wrap in `#IF TEST` `#endif`

Wikipedia/Code/WebViewController.m
```diff
@@ -640,6 +640,11 @@ - (WKWebViewConfiguration *)configuration {
                                                    forMainFrameOnly:YES]];
 
     WKWebViewConfiguration *configuration = [[WKWebViewConfiguration alloc] init];
+    
+    if(self.wkUserContentControllerTestingConfigurationBlock) {
```
joewalsh: wrap in `#IF TEST` `#endif`

Wikipedia/Code/WebViewController.h
montehurd: Fixed!

montehurd:
@joewalsh oh weird - the compiler is giving me a warning now in the test file that the method I wrapped in '#if test' doesn't exist...

montehurd:
Re-adding WIP tag 
- tracking down why 'test' isn't set when this test runs
- checking why there seems to be more difference than i'd expect between each of the 10 'measure' runs

montehurd:
Perhaps related to [this](https://stackoverflow.com/a/39435346/135557)?

montehurd:
I found a [way](https://davidwalsh.name/detect-node-insertion) to check for first section appearance without polling! Got it wired up - testing now...

montehurd:
@nambatee I ran into some issues with my test's `measureMetrics` loop's accuracy - finally got them ironed out. I should be able to review/test your (and @mdholloway's) pagelib PRs tomorrow.

montehurd:
@joewalsh @nambatee I think I'm done with tweaks to this :)

montehurd:
Last one :)

PR #1928, opened 2017-11-13
Exclude special pages from deep links

nambatee:
https://phabricator.wikimedia.org/T143267


Wikipedia/Code/WMFArticleViewController.m
```diff
@@ -1164,6 +1164,9 @@ - (void)fetchArticleForce:(BOOL)force {
                                                dismissPreviousAlerts:NO
                                                          tapCallBack:NULL];
                 }
+            } else if ([error wmf_isWMFErrorMissingTitle]) {
+                [self wmf_openExternalUrl:self.articleURL];
+                [self.navigationController popToRootViewControllerAnimated:YES];
```
joewalsh: Probably just want to pop the current VC instead of popping to root. If the user is more than one article deep on the nav stack, they should only go back to the previous article, not all the way back to explore, saved, history, or places.

Wikipedia/Code/WMFArticleViewController.m
```diff
@@ -1525,6 +1528,11 @@ - (void)showTalkPage {
     [self.navigationController presentViewController:safariController animated:YES completion:nil];
 }
 
+- (void)showSpecialPage {
```
joewalsh: it looks like `showSpecialPage ` isn't used any more and can be removed

PR #1929, opened 2017-11-14
Ensure popover is shown even if the vc disappears after the delay is scheduled but before the popover is shown

nambatee:


joewalsh:
retest this please

montehurd:
Oh man because the FB snapshot image file names change when version is bumped it makes it tricky to visually compare the re-generated images against previous versions... not much we can do about it I guess...

PR #1935, opened 2017-11-17
Fix jumpy text on languages footer

montehurd:
https://phabricator.wikimedia.org/T178729


Wikipedia/Code/WMFArticleLanguagesSectionFooter.swift
```diff
@@ -0,0 +1,30 @@
+class WMFInterfaceBuilderFriendlyUITableViewHeaderFooterView : UITableViewHeaderFooterView {
+    public override func layoutSubviews() {
+        super.layoutSubviews()
+        // There is no IB UITableViewHeaderFooterView component, which means you have to use a UIView.
+        // But you can't set UITableViewHeaderFooterView's 'contentView' from UIView IB component.
+        // This work-around fixes weird autolayout issues (likely related to this?) where multi-line
+        // labels from a xib using a UITableViewHeaderFooterView subclass appear at the wrong size for
+        // their text length, then jump to thier proper width a second after appearing.
+        frame = CGRect(origin: frame.origin, size: systemLayoutSizeFitting(CGSize(width: bounds.size.width, height: UILayoutFittingCompressedSize.height), withHorizontalFittingPriority: .required, verticalFittingPriority: .defaultHigh)
```
joewalsh: A view shouldn't set its own frame inside of `layoutSubviews`, it should only adjust the frame of its subviews.

Wikipedia/Code/WMFArticleLanguagesSectionFooter.swift
```diff
@@ -0,0 +1,30 @@
...
+        titleLabel.text = WMFLocalizedString("settings-primary-language-details", value:"The first language in this list is used as the primary language for the app. Changing this language will change daily content (such as Featured Article) shown on Explore.", comment:"Explanation of how the first preferred language is used. \"Explore\" is {{msg-wm|Wikipedia-ios-home-title}}.")
+        wmf_configureSubviewsForDynamicType()
+        apply(theme: .standard)
+    }
+    func apply(theme: Theme) {
+        backgroundView?.backgroundColor = theme.colors.baseBackground
+        titleLabel.textColor = theme.colors.secondaryText
+        addButton.setTitleColor(theme.colors.link, for: .normal)
+    }
+}
```
joewalsh: Was the conversion to Swift here necessary? It seems the class is largely the same and the same fix could have been implemented in Obj-C without too much hassle (including the superclass)

montehurd: Explained in the commit message.

Wikipedia/Code/WMFArticleLanguagesSectionFooter.swift
montehurd: Ya I wasn’t liking that either. Forgot to add Changes Welcome tag :)

PR #1939, opened 2017-11-19
Bug/t180078

onlyrandomness:
https://phabricator.wikimedia.org/T180078

The simplest solution to this is to disable the user interaction of the entire vc while the requests are active. The downside to this approach is the UI won't scroll during this time. For this reason I made dismissing the alert re-enable the user interaction so the app doesn't appear frozen during the request.

Another option would be to disable the text fields and buttons individually, as this may create a smoother user experience.


Wikipedia/Code/WMFLoginViewController.swift
```diff
@@ -319,4 +324,8 @@ class WMFLoginViewController: WMFScrollViewController, UITextFieldDelegate, WMFC
         scrollContainer.backgroundColor = theme.colors.paperBackground
         captchaViewController?.apply(theme: theme)
     }
+    
+    func setViewControllerUserInteraction(enabled: Bool) {
+        self.view.isUserInteractionEnabled = enabled
```
joewalsh: This should dim the view to indicate that user interaction is disabled. Can be addressed in a follow on PR.

joewalsh:
ok to test

joewalsh:
Thanks!

PR #1945, opened 2017-11-22
Add guards to didMoveToSuperview.

montehurd:


joewalsh:
what is this fixing?

montehurd:
@joewalsh these seem to fire when the view is removed from the superview too. no need to do the override code in that case - especially where colors/frames are concerned.

joewalsh:
@montehurd none of the one-time setup code should really be in `didMoveToSuperview` - it should be triggered on init for views created in code or in `awakeFromNib` when it relies on other views loaded from a nib

montehurd:
@joewalsh lgtm

PR #1947, opened 2017-11-23
Fix find-in-page next result vertical centering.

montehurd:
https://phabricator.wikimedia.org/T181210


Wikipedia/Code/WebViewController.m
```diff
@@ -552,7 +552,11 @@ - (void)scrollToAndFocusOnSelectedMatch {
                                                  @strongify(self);
                                                  self.disableMinimizeFindInPage = YES;
 
-                                                 CGFloat halfSpaceAboveKeyboardBar = [self.findInPageKeyboardBar convertPoint:CGPointZero toView:self.webView].y / 2.f;
+                                                 CGFloat spaceAboveKeyboardBar = [self.findInPageKeyboardBar convertPoint:CGPointZero toView:self.webView].y - self.webView.scrollView.contentInset.top;
+                                                 if (self.navBarHidden) {
+                                                     spaceAboveKeyboardBar = spaceAboveKeyboardBar - self.navigationController.navigationBar.bounds.size.height;
+                                                 }
+                                                 CGFloat halfSpaceAboveKeyboardBar = spaceAboveKeyboardBar / 2.f;
```
joewalsh: Using an `f` to force `float` can be removed since `CGFloat` is a `double` on all platforms. I see it was just a carry-over here, but adding as a note for future reference.

montehurd: oh good point!

PR #1950, opened 2017-11-27
New build system

joewalsh:
The new build system couldn’t handle a script that succeeded but had error output. In this case, the errors aren’t new and are ignorable, but the new build system treats the output as a build failure. Moved the files that generate the errors to a folder that the localization script doesn’t run on. These files shouldn’t have any localizable content, and they are the only place in code where we can pass arguments to `WMFLocalizedString` that aren’t string literals (the source of the error).


joewalsh:
retest this please

montehurd:
Very exciting!

PR #1954, opened 2017-11-28
Macro for testing announcements

joewalsh:


montehurd:
I had a [WIP](https://github.com/montehurd/apps-ios-wikipedia/commit/a32ef3676d78bb3851ccf9d85aed741f3d10b2b2) a while back where I hard-coded a couple bits to point to a localhost copy of an announcements json file... not sure if that approach would be preferable...

joewalsh:
Not really - this is more contained. If you want to be able to copy/paste a JSON string you could add a string var then turn it into a dictionary with `NSJSONSerialization` to pass to the `MTLJSONAdapter`

montehurd:
> If you want to be able to copy/paste a JSON string you could add a string var then turn it into a dictionary with NSJSONSerialization to pass to the MTLJSONAdapter

Testing with actual files the endpoint will deliver may be simpler...

PR #1959, opened 2017-11-30
[BUG] Explore "Save for later" button touch target is too small

onlyrandomness:
https://phabricator.wikimedia.org/T180632

In order to increase the hit box of the "Save for later" button, I changed the `contentEdgeInset` `top` and `bottom` values to 5 to give the button a little extra "padding"

In an effort to make this solution more flexible moving forward I added optional padding parameters to the `SaveButton` initializer so these values could be set at time of instantiation.

The only place where the `SaveButton` is instantiated programmatically is the `ArticleCollectionViewCell.swift` so maybe this is overkill, and can simply be replaced by default values for `contentEdgeInset` `top` and `bottom`


joewalsh:
Thanks for the PR! Good idea to add padding options to the button. I think it can be moved up into  `AlignedImageButton` and consolidated to a single parameter (for now, since we only set the same value for top and bottom). This way `AlignedImageButton` remains in full control of its insets.

The solution also added some extra spacing to the explore cells, so the cells need to account for the extra padding. Here are some screenshots:

Before:
simulator screen shot - iphone se - 2017-11-30 at 06 39 23

After:
simulator screen shot - iphone se - 2017-11-30 at 06 38 10

I opened a PR (#1961) based off of this one with the suggested updates. Let me know if that looks good to you (feel free to merge the branch back in here and make other changes if you'd like). Thanks again!

onlyrandomness:
@joewalsh Everything in #1961 looks great, nice work!

joewalsh:
ok to test

PR #1971, opened 2017-12-04
Add lazy image loading

montehurd:
https://phabricator.wikimedia.org/T175323

- [x] integrate lazy image transform
- [x] fix image taps (gallery isn't loading presently - certain `data-` attributes aren't being preserved). <strike>Awaiting page lib [PR](https://github.com/wikimedia/wikimedia-page-library/pull/112) review/merge.</strike> 
- [x] fix image placeholder taps
- [x] figure out lazy image transform placeholder size issue (not same size as image - appears to not be widened correctly). <strike>Awaiting page lib [PR](https://github.com/wikimedia/wikimedia-page-library/pull/111) review/merge.</strike>


www/js/sections.js
```diff
@@ -165,6 +177,9 @@ const applyTransformationsToFragment = (fragment, article, isLead) => {
   // 'enwiki > Quadradic equation' and 'enwiki > Away colors > Association football'). See the
   // 'classifyElements' method itself for other examples.
   requirements.themes.classifyElements(fragment)
+
+  lazyImageLoadingTransformer.convertImagesToPlaceholders(fragment)
```
niedzielski: You may want to forbid lazy load transforms on file pages.

www/js/sections.js
```diff
@@ -165,6 +177,9 @@ const applyTransformationsToFragment = (fragment, article, isLead) => {
...
+  lazyImageLoadingTransformer.loadPlaceholders()
```
niedzielski: This should be fine but depending on how your loading works, you may only need to do this once after the page is loaded instead of after every section load.

www/js/sections.js
```diff
@@ -113,7 +125,7 @@ const processResponseStatus = response => {
 const extractResponseJSON = response => response.json()
 
 const fragmentForSection = section => {
-  const fragment = document.createDocumentFragment()
+  const fragment = lazyDocument.createDocumentFragment()
```
niedzielski: I've tried updating the demos to use a DocumentFragment instead of a div but they seem to work fine. Since DocumentFragments avoid reflows, it would be a good debugging test to see if using a div changes the lazy load image reflow behavior.

www/js/sections.js
```diff
@@ -165,6 +177,9 @@ const applyTransformationsToFragment = (fragment, article, isLead) => {
   // 'enwiki > Quadradic equation' and 'enwiki > Away colors > Association football'). See the
   // 'classifyElements' method itself for other examples.
   requirements.themes.classifyElements(fragment)
+
```
niedzielski: After passing through the lazy load transform, what do your placeholders look like? This would be one of the first things to check if the placeholders appeared with the wrong dimensions. If you can copy and paste an image before the transform, after the transform, and after the inverse, I can take a look too. Setting the multiplier to a lower value like .1 should allow you easily examine the different states of a transformed image and verify whether the transform is correct or not. Example here: https://docs.google.com/document/d/1K4bFdzDhSuh2FYs_9XrdD5efW7mMMVvPazWwOctSB_s/edit#heading=h.1ll3i14x7eaw

www/js/sections.js
montehurd: I'll check thanks!

www/js/sections.js
montehurd: The cool thing about the way we're using fragments is all transforms are complete before a fragment is attached to the `live` document :)

niedzielski: Sorry, I was pointing at the call to `loadPlaceholders()` which selectively undoes the transforms.

montehurd: Ah, cool thanks!

montehurd: Thoughts on maybe naming `loadPlaceholders` a bit differently to better distinguish what it does from what `convertImagesToPlaceholders` does?

niedzielski: `loadPlaceholders()` transforms placeholders to images. The intent was for the client to think of images as being loaded into the placeholders since the download can fail but naming is tough. Feel free to submit naming improvements in a separate patch that bumps the major version and flag an Android dev.

www/js/sections.js
montehurd: Here's an example from today's featured article. 

Before:
screen shot 2017-12-06 at 12 38 29 pm

After:
screen shot 2017-12-06 at 12 38 33 pm

montehurd: Oh this may be a clue... on `enwiki > Picasso's Blue Period` the first image has the issue, but all of the images after the first one have their placeholders correctly widened... hmm...

niedzielski: (In my best Jerry Maguire impression) show me the code! Sorry, I was referring to the raw HTML in each stage (check out the link I shared where you can see how I show what it looks like before the transform, after the transform, and after loading, etc). We can then see if it differs from the demos.

montehurd: lol oops! :) ya i spent a while yesterday live-tweaking the CSS comparing it with your lifecycle examples here https://github.com/wikimedia/apps-android-wikipedia/commit/620538d961221942e340ca7ac7f429393d1309d6

gotta grab some foods atm - will post screenshots after

montehurd: Screenshots from `enwiki > Picasso's Blue Period`

# The first image exhibits the bug:

html w/o lazy loading


html w lazy loading, placeholder (shows the bug - i.e. the placeholder dimensions are not widened)


html w lazy loading, image fetched



# The second image does not exhibit the bug:

html w/o lazy loading


html w lazy loading, placeholder


html w lazy loading, image fetched

niedzielski: Just started looking at this but the first thing I saw was that `pagelib_widen_image_override` appears to be incorrectly limited to `img` in WidenImage.css? If you change the CSS rule from `img.pagelib_widen_image_override` to `pagelib_widen_image_override`, maybe it'll work? (I'm resuming inspection now.)

montehurd: iirc I tried that yesterday and it didn't make any difference. Will try again to make sure...

montehurd: Ok, so changing `img.pagelib_widen_image_override` to `.pagelib_widen_image_override` makes the lazy placeholders disappear completely... wot!

www/js/sections.js
montehurd: Tested with div and saw same behavior.

www/js/sections.js
montehurd: ^ actually the placeholder's only disappear for images exhibiting the bug - i.e. the 1st image in the screenshots above in this comment thread.

niedzielski: I've modified the ThemeTransform demo to perform image widening and lazy loading. I get a significant reflow when I make the screen width very small. Can you try these changes to [WidenImage.css](https://github.com/wikimedia/wikimedia-page-library/pull/110/files#diff-13d24612474e53852200f7cbc0ae483f)?

montehurd: I tried those changes - looks like they break image widening on about half of the images on `enwiki > Pablo Picasso`...

montehurd: ^ Looks like the affected images have the `thumbimage` class...

montehurd: I think I have a fix! Still testing, but adding the following to the first line of `widenAncestors` seems to do the trick...
<pre>
if(el.parentElement.tagName === 'A'){
    el.parentElement.style.width = '100%'
}
</pre>

montehurd:
@niedzielski 

Heya! Question about the `data-` attribute copying by lazy load: There are 3 items on iOS which would be nice to carry over from img > span > img. Specifically, `data-file-width`, `data-file-height`, and `data-image-gallery`. 

I initially wrote a generic method to ensure any `data-` items would just carry over, without having to add them to COPY_ATTRIBUTES, but it seemed simpler for now to just add those 3 items to COPY_ATTRIBUTES. 

The complexity in the more generic solution had to do with potentially overwriting `data-` items (ie if the orig img had a `data-src` and our COPY_ATTRIBUTES had `src`, etc). I thought we could change the prefix used when parking COPY_ATTRIBUTES items to something like `data-lazy-` to avoid potential overwrites, but then I ran into testing issues with Domino and HTMLElement.attributes, which Domino doesn`t [correctly implement](https://github.com/fgnass/domino/commit/cf8f7ae6e57a4caf2d36711336d0f9d6b4ec9500)

Anyway I felt uncomfortable switching to a not-easily-testable approach... so I guess my question is, do you think it's ok to just add  `data-file-width`, `data-file-height`, and `data-image-gallery` to COPY_ATTRIBUTES for now if I add comment to remind us these are used by iOS?

niedzielski:
@montehurd 

> it seemed simpler for now to just add those 3 items to COPY_ATTRIBUTES.

:+1: That sounds fine.

> The complexity in the more generic solution had to do with potentially overwriting data- items (ie if the orig img had a data-src and our COPY_ATTRIBUTES had src, etc).

I think in a previous implementation I just copied all `data-*` attributes first and then copied the specific non-`data-*` attributes like `src` second. If there was a conflict, it was overwritten which seems reasonable to me. I think also experimented with `data-data-*` which is the approach your new implementation uses. 

> issues with Domino and HTMLElement.attributes, which Domino doesn`t correctly implement

I'm not sure I follow. We already test all the currently copied attributes AFAIK: https://github.com/wikimedia/wikimedia-page-library/blob/master/test/transform/LazyLoadTransform.test.js#L114-L124. Additionally, a very simple workaround is described in the ticket.

montehurd:
@niedzielski 

> issues with Domino and HTMLElement.attributes, which Domino doesn`t correctly implement

Oh, sorry, my generic copying implementation iterated the img tags `attributes` checking for any keys with `data-` prefix.

Wikipedia/assets/preview.js
```diff
@@ -1745,8 +1787,7 @@ var IMAGE_LOADED_CLASS = 'pagelib_lazy_load_image_loaded'; // Download completed
 
 // Attributes copied from images to placeholders via data-* attributes for later restoration. The
 // image's classes and dimensions are also set on the placeholder.
-// The 3 data-* items are used by iOS.
-var COPY_ATTRIBUTES = ['class', 'style', 'src', 'srcset', 'width', 'height', 'alt', 'data-file-width', 'data-file-height', 'data-image-gallery'];
```
montehurd: These are the bits I accidentally merged too soon - they are required for this PR, but will be merged once we update pagelib version once these bits are merged there.

PR #1975, opened 2017-12-07
Interactive navigation bar hiding

joewalsh:
Each view controller manages a `NavigationBar` using a `NavigationBarHider`


montehurd:
Looks like there's some extra whitespace above the search box. 

# Before/After
screen shot 2017-12-11 at 3 20 42 pm

montehurd:
I tested a bit for those scrolling related issues I recently patched.

Looks like [this](https://github.com/wikimedia/wikipedia-ios/pull/1938) regressed and is causing regressions with:
- TOC selection. The selected section isn't scrolled to top. There's a gap of nav bar height.
- Find-in-Page next result vertical centering is also off by nav bar height.

PR #1983, opened 2017-12-14
Fix edit captchas

montehurd:
https://phabricator.wikimedia.org/T182079


Wikipedia/Code/PreviewAndSaveViewController.m
```diff
@@ -500,7 +499,7 @@ - (void)fetchFinished:(id)sender
 
                         NSURL *captchaUrl = [[NSURL alloc] initWithString:error.userInfo[@"captchaUrl"]];
                         NSString *captchaId = error.userInfo[@"captchaId"];
-                        [[WMFAlertManager sharedInstance] showErrorAlert:error sticky:YES dismissPreviousAlerts:YES tapCallBack:NULL];
```
montehurd: We may have a bug with `dismissPreviousAlerts` because there were 2 alerts, on on top of the other.

PR #1986, opened 2017-12-14
Use iOS 11+ recommended API change, fix iOS 10 bottom inset bug

joewalsh:
Fix issue where bottom inset is too tall on iOS 10 in some situations.

Repro steps:

From explore, search for "North Korea" and tap the article. Scroll sideways and observe the location of the horizontal scroll indicator.

https://phabricator.wikimedia.org/T182901

Also fixes issue where swipe to back would fail on articles with side scrolling.

Repro steps:

From explore, search for "North Korea" and tap the article. Attempt to swipe to go back to explore.

https://phabricator.wikimedia.org/T182900

### DIFF
--- a/Wikipedia/Code/WebViewController.m
+++ b/Wikipedia/Code/WebViewController.m
@@ -673,6 +673,11 @@ typedef NS_ENUM(NSUInteger, WMFFindInPageScrollDirection) {
                                              selector:@selector(articleUpdatedWithNotification:)
                                                  name:WMFArticleUpdatedNotification
                                                object:nil];
+
+    UIGestureRecognizer *interactivePopGR = self.navigationController.interactivePopGestureRecognizer;
+    if (interactivePopGR) {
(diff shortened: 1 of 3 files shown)


Wikipedia/Code/WebViewController.m
```diff
@@ -673,6 +673,11 @@ - (void)viewWillAppear:(BOOL)animated {
                                              selector:@selector(articleUpdatedWithNotification:)
                                                  name:WMFArticleUpdatedNotification
                                                object:nil];
+
+    UIGestureRecognizer *interactivePopGR = self.navigationController.interactivePopGestureRecognizer;
+    if (interactivePopGR) {
+        [self.webView.scrollView.panGestureRecognizer requireGestureRecognizerToFail:interactivePopGR];
```
montehurd: When I saw the ticket this was the fix I was thinking about... nice!

montehurd:
Testing now...

PR #1987, opened 2017-12-14
Black theme tweaks.

montehurd:
https://phabricator.wikimedia.org/T181847#3838774
https://phabricator.wikimedia.org/T182843


montehurd:
There are a couple article tweaks that go with this but waiting until I hear back if they need iOS platform specifiers...

montehurd:
Here's the pagelib pr: https://github.com/wikimedia/wikimedia-page-library/pull/116

PR #1990, opened 2017-12-15
Peeking "Next random article" footer of random doesn't show the next random article

nambatee:
https://phabricator.wikimedia.org/T182984

Is there a better way to do this?

Fetching the url in `peekViewControllerForItemAtIndexPath` is too slow and won't peek.

### DIFF
--- a/Wikipedia/Code/ArticlePeekPreviewViewController.swift
+++ b/Wikipedia/Code/ArticlePeekPreviewViewController.swift
@@ -20,7 +20,7 @@ class ArticlePeekPreviewViewController: UIViewController, Peekable {
         return nil
     }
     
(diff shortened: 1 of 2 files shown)


joewalsh:
@nambatee instead of trying to fetch the random article and show it in the peek card, i'd say just peek/pop `WMFRandomArticleViewController` without the card for now. I think ultimately random is going to change to using the preview card instead of the full article view each time you shuffle, so this is an intermediate fix.

nambatee:
Closing this PR and moving the ticket back to the backlog. Given the current Random implementation, there's no good way to solve this. I will revisit this when we begin rethinking Random implementation.

PR #1993, opened 2017-12-15
Log WMF api warnings so we can catch and fix them.

montehurd:
I noticed a few different API warnings when playing with various feed requests. Figured we could catch these more easily in the future if we logged warnings.

### DIFF
--- a/Wikipedia/Code/WMFApiJsonResponseSerializer.m
+++ b/Wikipedia/Code/WMFApiJsonResponseSerializer.m
@@ -27,6 +27,14 @@
         }
         return nil;
     }
+
+#if DEBUG
+    NSDictionary *warnings = json[@"warnings"];
+    if (warnings) {
+        NSLog(@"\n\n\nWarnings for request:\n%@\n%@", response.URL.absoluteString, warnings);
+    }
+#endif
+
     return json;
 }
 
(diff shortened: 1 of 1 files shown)


Wikipedia/Code/WMFApiJsonResponseSerializer.m
```diff
@@ -27,6 +27,14 @@ - (id)responseObjectForResponse:(NSURLResponse *)response
         }
         return nil;
     }
+
+#if DEBUG
+    NSDictionary *warnings = json[@"warnings"];
+    if (warnings) {
+        NSLog(@"\n\n\nWarnings for request:\n%@\n%@", response.URL.absoluteString, warnings);
```
montehurd: I suppose we could also assert that there are no warnings... would that be preferred?

joewalsh: I think I'd rather just log for now - I think there were a few cases of expected warnings in places search. Also use `DDLogWarn` for consistency (even though it's wrapped in #if DEBUG)

montehurd: Ah will do.

montehurd: Fixed!

PR #1996, opened 2017-12-15
Disable w icon popover

joewalsh:


montehurd:
Reminder: this was to fix edit pencil taps.

PR #1999, opened 2017-12-16
before adding @Wikipedia, check if sharing via Twitter

nambatee:
https://phabricator.wikimedia.org/T183060


joewalsh:
I think long term `ShareAFactActivityTextItemProvider` could be updated to be re-used here, but this is a good short term fix. I made one small modification to simplify the if condition - the !twitter check covers all of the other cases.

PR #2000, opened 2017-12-18
Fix "Help and feedback" view

joewalsh:
https://phabricator.wikimedia.org/T183143#3844703

### DIFF
--- a/Wikipedia/Code/WKWebView+WMFWebViewControllerJavascript.swift
+++ b/Wikipedia/Code/WKWebView+WMFWebViewControllerJavascript.swift
@@ -142,10 +142,7 @@ extension WKWebView {
     }
     
     private func languageJS(for article: MWKArticle) -> String {
-        guard let lang = (article.url as NSURL).wmf_language else {
-            assertionFailure("Expected lang")
-            return ""
-        }
+        let lang = (article.url as NSURL).wmf_language ?? MWKLanguageLinkController.sharedInstance().appLanguage?.languageCode ?? "en"
         let langInfo = MWLanguageInfo(forCode: lang)
         let langCode = langInfo.code
         let langDir = langInfo.dir
--- a/Wikipedia/Code/ViewController.swift
+++ b/Wikipedia/Code/ViewController.swift
@@ -12,12 +12,7 @@ class ViewController: UIViewController, Themeable {
     }
     
     let navigationBar: NavigationBar = NavigationBar()
-    open var showsNavigationBar: Bool {
-        guard let navigationController = navigationController else {
-            return false
-        }
-        return parent == navigationController && navigationController.isNavigationBarHidden
-    }
+    open var showsNavigationBar: Bool = false
     
     open var scrollView: UIScrollView? {
         return nil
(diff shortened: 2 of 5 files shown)


nambatee:
simulator screen shot - iphone x - 2017-12-18 at 10 22 44
simulator screen shot - iphone x - 2017-12-18 at 10 22 47

scroll view inset breaks for similar pages

joewalsh:
@nambatee good catch, pushed the fix

montehurd:
This diff is really confusing...

Wikipedia/Code/WKWebView+WMFWebViewControllerJavascript.swift
```diff
@@ -142,10 +142,7 @@ extension WKWebView {
     }
     
     private func languageJS(for article: MWKArticle) -> String {
-        guard let lang = (article.url as NSURL).wmf_language else {
-            assertionFailure("Expected lang")
-            return ""
-        }
+        let lang = (article.url as NSURL).wmf_language ?? MWKLanguageLinkController.sharedInstance().appLanguage?.languageCode ?? "en"
```
montehurd: Are we sure we want `appLanguage` here? The assertion seemed to be useful to know asap that we're getting what we expect...

joewalsh: @montehurd appLangugae then english seemed like a reasonable fallback path for sites without language. The help page is https://m.mediawiki.org/wiki/Wikimedia_Apps/iOS_FAQ/

nambatee: @montehurd look at the commits after d563e42 (if the diff is confusing)

montehurd: @joewalsh defaulting to EN is fine but it's no guarantee the title on EN matches whatever that article's title is in device lang. Just suspecting the assert is useful for early warning in case something ever goes crazy wrong...

montehurd:
@nambatee Going forward it's probably better we have clean diffs. It's confusing if the PR includes a bunch of unrelated stuff.

nambatee:
@montehurd yeah, ideally but in this case, the commits after d563e42 are relevant and that's all I was trying to tell you (not make a statement about whether diffs should be dirty/clean) ¯\_(ツ)_/¯.

joewalsh:
@montehurd this was my PR - the problem was originally targeting it into 5.7.3, then bringing develop into 5.7.3 along with this PR. 
ed0454b...d0bd76e are the commits that matter.

Wikipedia/Code/WKWebView+WMFWebViewControllerJavascript.swift
joewalsh: @montehurd feel free to submit an alternate fix or re-add the assertion without the guard/return. falling back on default languages seemed like the lowest impact fix for the help & feedback page for 5.7.3.

montehurd:
@nambatee No worries! I added a quick note to the eng-sync etherpad suggesting in cases like this we just add a quick comment to any such PR's identifying relevant commit range(s) so we as reviewers don't have to circle back with comments for clarity. I forgot until seeing @joewalsh's link that github supports `compare` `commit` `...` `commit` range link syntax, which is pretty cool :) iirc we've also had cases where unrelated commits were _not_ intentional, so a comment would speed clarification around such questions as well.

PR #2004, opened 2017-12-19
Fix scroll jump when scrolled to bottom by preserving offset, fix places list view content inset

joewalsh:
https://phabricator.wikimedia.org/T183237 2a15a0b...521391e

https://phabricator.wikimedia.org/T183259 f7c8d2c

### DIFF
--- a/WMF Framework/UIScrollView+Limits.swift
+++ b/WMF Framework/UIScrollView+Limits.swift
@@ -18,16 +18,16 @@ extension UIScrollView {
     }
 
     fileprivate var wmf_isAtTop: Bool {
-        return contentOffset.y == wmf_topOffsetY
+        return contentOffset.y <= wmf_topOffsetY
     }
 
     fileprivate var wmf_isAtBottom: Bool {
-        return contentOffset.y == wmf_bottomOffsetY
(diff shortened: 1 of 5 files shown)


joewalsh:
retest this please

WMF Framework/UIScrollView+Limits.swift
```diff
@@ -0,0 +1,46 @@
...
+        let wasAtTop = wmf_isAtTop
+        let wasAtBottom = wmf_isAtBottom
+        scrollIndicatorInsets = updatedScrollIndicatorInsets
+        contentInset = updatedContentInset
+        if wasAtTop {
+            contentOffset = wmf_topOffset
+        } else if wasAtBottom {
+            contentOffset = wmf_bottomOffset
+        }
+        if wmf_isAtTop {
```
montehurd: Is *is* (vs *was*) intentional here?

joewalsh: Yes, sometimes the rotation will put it at the top even if it wasn't at the top. On an X in portrait, view explore feed, scroll just the search bar invisible, rotate to landscape.

PR #2012, opened 2018-01-02
Fix missing thumbnails in search

joewalsh:
https://phabricator.wikimedia.org/T183918
and
https://phabricator.wikimedia.org/T183543 in 
aea54e8


nambatee:
couldn't reproduce the thumbnail bug, the nav bar in airplane mode looks good 👍

PR #2021, opened 2018-01-15
[BUG] [OTRS] When I delete recently searched, not all search terms get deleted

nambatee:
https://phabricator.wikimedia.org/T184932


Wikipedia/Code/WMFSearchViewController.m
```diff
@@ -190,6 +191,8 @@ - (void)viewDidLoad {
     [self updateUIWithResults:nil];
     [self updateRecentSearchesVisibility:NO];
 
+    self.clearedAllRecentSearches = NO;
```
joewalsh: Should this be in `viewWillAppear`? Seems like if you clear searches, leave search, re-enter search, perform a search, leave again, it won't save the last search.

PR #2026, opened 2018-02-07
Initial Reading List Changes

joewalsh:
- UI/UX changes for multiple reading lists
- First pass of sync is implemented but not enabled


WMF Framework/CollectionViewCellActionsView.swift
```diff
@@ -147,15 +151,15 @@ public class ActionsView: SizeThatFitsView, Themeable {
 
     public weak var delegate: ActionDelegate?
     
-    @objc func didPerformAction(_ sender: UIButton) {
+    @objc func shouldPerformAction(_ sender: UIButton) {
```
joewalsh: A method starting with `should` implies that a `BOOL` is returned that determines whether or not the action is performed. It seems the return value is ignored here.

WMF Framework/CollectionViewEditController.swift
```diff
@@ -240,6 +279,13 @@ public class CollectionViewSwipeToEditController: NSObject, UIGestureRecognizerD
...
+    var areSwipeActionsDisabled: Bool = false {
+        didSet {
+            longPressGestureRecognizer.isEnabled = !areSwipeActionsDisabled
+            panGestureRecognizer.isEnabled = !areSwipeActionsDisabled
+        }
+    }
+    
     // MARK: - States
     
     func openActionPane(_ completion: @escaping (Bool) -> Void = {_ in }) {
```
montehurd: Naming question - how do we feel about implicit `completion` parameters? I think I prefer explicit vs "_", but I don't feel super strongly about it beyond maybe wanting consistency...

joewalsh: It's fine unlabeled here as it's the first and only param. Autocomplete handles the case as well:
screen shot 2018-02-08 at 7 36 57 am
screen shot 2018-02-08 at 7 36 13 am

PR #2028, opened 2018-02-08
Education panels

montehurd:
https://phabricator.wikimedia.org/T177559

- [x] panel strings and images
- [x] scroll tall panels
- [x] hide image on compact vertical size classes (per Carolyn)
- [x] darken overlay behind panel
- [x] light blur (translucency) effect for the panel itself
- [x] wire up panels

# Includes the following panels:

screen shot 2018-02-14 at 6 48 36 pm
screen shot 2018-02-14 at 6 48 56 pm
screen shot 2018-02-14 at 6 49 22 pm
screen shot 2018-02-14 at 6 49 30 pm
screen shot 2018-02-14 at 6 50 45 pm
screen shot 2018-02-14 at 7 03 25 pm
screen shot 2018-02-14 at 6 49 46 pm

# Hides image on compact vertical size classes:

screen shot 2018-02-14 at 6 49 59 pm

# Scrolls if content is too tall (long translations/large dynamic text size etc):
(Note: discussed with Carolyn the 'X' scrolling up and away with the content - this is intentional.)
aaa


Wikipedia/Code/Panels.swift
```diff
@@ -0,0 +1,111 @@
...
+        image = UIImage.init(named: "reading-list-syncing")
+        heading = WMFLocalizedString("reading-list-sync-enable-title", value:"Turn on reading list syncing?", comment:"Title describing reading list syncing.")
+        subheading = WMFLocalizedString("reading-list-sync-enable-subtitle", value:"Your saved articles and reading lists can now be saved to your Wikipedia account and synced across devices.", comment:"Subtitle describing reading list syncing.")
+        primaryButtonTitle = WMFLocalizedString("reading-list-sync-enable-button-title", value:"Enable syncing", comment:"Title for button enabling reading list syncing.")
+    }
+}
+
+extension UIViewController {
+    @objc func wmf_showEnableReadingListSyncPanelOnce(theme: Theme) {
+        guard !UserDefaults.wmf_userDefaults().wmf_didShowEnableReadingListSyncPanel() else {
```
montehurd: @joewalsh You had mentioned not using user defaults for these `once per user login` panels... what was the alternative again? (The way I've wired this one up is `once ever` a.t.m. - I'll will rename it when it's switched to `once per user login`.)

joewalsh: In retrospect, it'ss probably fine to use defaults for these

Wikipedia/Code/WMFAuthenticationManager.swift
```diff
@@ -144,6 +144,8 @@ class WMFAuthenticationManager: NSObject {
         }
         SessionSingleton.sharedInstance()?.dataStore.clearMemoryCache()
         
+        SessionSingleton.sharedInstance().dataStore.readingListsController.setSyncEnabled(false, shouldDeleteLocalLists: false, shouldDeleteRemoteLists: false)
```
montehurd: @joewalsh Do these parameters look ok for calling this on logout?

montehurd:
@joewalsh @nambatee I think the wiring all works now. Should be good for some CR.

Wikipedia/Code/Panels.swift
```diff
@@ -0,0 +1,203 @@
...
+    @objc func wmf_showKeepSavedArticlesOnDevicePanelIfNecessary(theme: Theme, completion: @escaping (() -> Swift.Void) = {}) {
+        guard self.hasSavedArticles() else {
+            completion()
+            return
+        }
+        let panelVC = KeepSavedArticlesOnDevicePanelViewController(showCloseButton: false, primaryButtonTapHandler: { sender in
+            SessionSingleton.sharedInstance().dataStore.readingListsController.setSyncEnabled(false, shouldDeleteLocalLists: false, shouldDeleteRemoteLists: false)
+            sender.dismiss(animated: true, completion: nil)
+        }, secondaryButtonTapHandler: { sender in
+            SessionSingleton.sharedInstance().dataStore.readingListsController.setSyncEnabled(false, shouldDeleteLocalLists: true, shouldDeleteRemoteLists: false)
```
montehurd: @joewalsh 

This line gets called on logout when I tap "No, delete articles from device", but it doesn't seem to be working - the local lists are still there. 

screen shot 2018-02-15 at 2 09 35 pm

Is that a bug or am I mis-reading how that method is used?

montehurd: Ah it looks like `.setSyncEnabled()` only tries to delete local lists if `enabled` is being set to `true`. Would de-coupling `shouldDeleteLocalLists` from the `enabled` parameter be ok?

Wikipedia/Code/WMFAuthenticationManager.swift
joewalsh: yes

PR #2032, opened 2018-02-12
[BUG] Tapping 'Back' button as soon as 'Tap to go home' tooltip appears goes back to feed but leaves app frozen

nambatee:
Do we also want to dismiss if the user scrolls down? Currently, the popover will persists on scroll and now that we're hiding the nav bar it might look weird:
img_2175247aa9d9-1


joewalsh:
@nambatee I'd say it should dismiss on any interaction, including scroll down

PR #2038, opened 2018-02-15
Reading list updates

joewalsh:
- Update `SavedArticlesFetcher` to be able to handle influx of a large number of articles (previously would try to download them all simultaneously at high priority, now will download one at a time at low network priority)
- Consolidate sync logic and use a persistent state variable to handle transitions between enable/disable/update/sync
- Enable sync when another device enables sync remotely


joewalsh:
retest this please

PR #2042, opened 2018-02-20
As a user of reading lists, I don't expect my saved articles to be unsaved when I delete the reading list they are associated with

nambatee:
https://phabricator.wikimedia.org/T187602


Wikipedia/Localizations/en.lproj/Localizable.strings
```diff
@@ -431,9 +430,8 @@
 "two-factor-login-title" = "Log in to your account";
 "two-factor-login-with-backup-code" = "Use one of your backup codes";
 "two-factor-login-with-regular-code" = "Use verification code";
-"unsave-article-and-remove-from-reading-list-with-name-message" = "Unsaving this article will also remove it from the reading list “$1”.";
-"unsave-article-and-remove-from-reading-lists-message" = "Unsaving this article will remove it from all reading lists.";
-"unsave-article-and-remove-from-reading-lists-title" = "Unsave article?";
+"unsave-article-and-remove-from-reading-lists-message" = "Un-saving {{PLURAL:$1|this articles will remove it|these articles will remove them}} from all associated reading lists";
```
montehurd: Looking at the other strings in this file, we usually start the keys with the feature.

PR #2043, opened 2018-02-20
Fix/simplify sync logic

joewalsh:


Wikipedia/Code/WMFSettingsMenuItem.m
```diff
@@ -136,6 +136,16 @@ + (WMFSettingsMenuItem *)itemForType:(WMFSettingsMenuItemType)type {
                                        disclosureText:nil
                                            isSwitchOn:[SessionSingleton sharedInstance].dataStore.readingListsController.isSyncEnabled];
         }
+        case WMFSettingsMenuItemType_StorageAndSyncingDebug: {
+            return
+            [[WMFSettingsMenuItem alloc] initWithType:type
+                                                title:@"Reading list danger zone"
```
montehurd: Looks like we need a localized string here.

nambatee: This is only for debugging though

montehurd: Haha you're right! I just saw a string and didn't even read it :-P

joewalsh: simulator screen shot - iphone x - 2018-02-21 at 07 45 52

PR #2045, opened 2018-02-21
Theme updates for reading lists

montehurd:
https://phabricator.wikimedia.org/T187415

(this also updates the "Add to reading list" hint, but the screenshots below don't show this.)

| Before        | After |
| ------------- |:-------------:|
| screen shot 2018-02-21 at 12 12 29 pm | screen shot 2018-02-21 at 11 58 50 am|
| screen shot 2018-02-21 at 12 12 17 pm | screen shot 2018-02-21 at 11 59 03 am|
| screen shot 2018-02-21 at 12 12 06 pm | screen shot 2018-02-21 at 11 59 17 am|


Wikipedia/Code/BatchEditSelectView.swift
```diff
@@ -20,7 +24,7 @@ public class BatchEditSelectView: SizeThatFitsView {
     }
     
     fileprivate func updateMultiSelectIndicatorImage() {
-        let image = isSelected ? UIImage(named: "selected", in: Bundle.main, compatibleWith: nil) : UIImage(named: "unselected", in: Bundle.main, compatibleWith: nil)
+        let image = isSelected ? theme.multiSelectIndicatorImage : UIImage(named: "unselected", in: Bundle.main, compatibleWith: nil)
```
nambatee: any reason not to theme the image via tintColor = theme.colors.link like we do with the hint?

montehurd: Ya the white part inside the image.

nambatee: simulator screen shot - iphone x - 2018-02-20 at 21 45 46

works for me with
`let image = isSelected ? UIImage(named: "selected", in: Bundle.main, compatibleWith: nil) : UIImage(named: "unselected", in: Bundle.main, compatibleWith: nil)` in `updateMultiSelectIndicatorImage()` of `BatchEditSelectView`
and
`multiSelectIndicator?.tintColor = theme.colors.link` in `apply(theme: Theme)` of `BatchEditSelectView`

montehurd: @nambatee 
Hmm that pic doesn't appear to reflect the needed change (making the check image blue a lighter color). Did you switch the image to use "template" rendering in the assets catalog? My understanding of `template` style image handling w/imageview's tint is that the tint color replaces all non-clear parts of the image. In this case that makes even the white check become the tint color, which makes a solid circle. It was the first thing I tried.

nambatee:
If you change the theme while the hint is visible, the hint won't be themed

simulator screen shot - iphone x - 2018-02-21 at 15 11 05

nambatee:
`ReadingThemesControlsViewController` and `AppearanceSettingsViewController` post `WMFUserDidSelectThemeNotification` when the theme changes, so you can wait for that in `ReadingListHintViewController` and re-apply theme

montehurd:
@nambatee ah good catch on the hint not changing in time. Will check it...

Wikipedia/Code/BatchEditSelectView.swift
montehurd: @nambatee The plus/add icon only has one non-transparent color so the tint trick works - ie the plus is just transparent (vs white), so using tint to change all the non-transparent parts does what we want.

PR #2046, opened 2018-02-21
[Bug] Save for later expanding outside of card boundaries on the Explore feed on iPad

nambatee:
https://phabricator.wikimedia.org/T187509


nambatee:
thanks @joewalsh, will change it in some other pr 👍

PR #2055, opened 2018-02-23
Reading list tweaks

joewalsh:


www/less/misc.less
```diff
@@ -7,6 +7,7 @@ a:hover {
 
 a {
     color: #3366CC;
+    -webkit-tap-highlight-color: rgba(245,166,35,0.2);
```
montehurd: This is why I suggested we wait ;) It just seemed too wonky that we needed any JS to do this. Would you mind moving this to a separate PR? It's a bit hard to argue it's a reading list tweak hehe.

Wikipedia/Code/ReadingListDetailViewController.swift
```diff
@@ -274,7 +291,7 @@ extension ReadingListDetailViewController: ActionDelegate {
         }
     }
     
-    func didPerformAction(_ action: Action) -> Bool {
+    func didPerformAction(_ action: Action, from sender: UIButton?) -> Bool {
```
montehurd: Looks like `sender` is unused by this method.

www/less/misc.less
montehurd: Fingers-crossed the upstream setting for this may get changed... https://phabricator.wikimedia.org/T188043#3997885

joewalsh: @montehurd the JS will likely be necessary again if the design team decides the active state should persist through the navigation. You were able to find and comment on commits related to this change several times within the context of a larger PR, so it doesn't make sense to split out at this point.

joewalsh:
Merged as this was getting unwieldy and out of scope, future reading list PRs should be more atomic.

PR #2066, opened 2018-02-27
Disable clear button when collection view is empty

nambatee:
This fixes clear button being enabled even when collection view is empty. We don't always go through `emptyStateDidChange` so we need to make sure we set `navigationItem.leftBarButtonItem?.isEnabled` in `emptyStateDidChange` and after we set `leftButtonType`


nambatee:
retest this please

PR #2067, opened 2018-02-27
Remote config check for sync state

joewalsh:
- Old assets file system was attempting to write to the WMF.framework asset folder (as a part of launch time improvement, the copy of these assets files to the user's documents folder was removed  due to an expensive filesystem scan and check on every launch. The attempted write of the config file to the non-documents location remained)
-Adds new config check, uses that config check to set whether or not reading list sync is remotely enabled


joewalsh:
Per conversation with @jtminor this shouldn't touch the UI, it should only prevent hitting the sync API endpoints

PR #2070, opened 2018-02-27
Fix share/save actions on search results view

joewalsh:
Ensure WMFArticle object is created if it doesn't exist


montehurd:
https://phabricator.wikimedia.org/T188444

PR #2082, opened 2018-03-01
Remove upstream tap highlight override, don't set custom color

joewalsh:


joewalsh:
Can be pulled from upstream css after https://gerrit.wikimedia.org/r/#/c/415556/

PR #2086, opened 2018-03-01
Proof-of-concept overlaying tab bar icons with indicator.

montehurd:
https://phabricator.wikimedia.org/T187823

Trigger proof-of-concept in simulator by triggering memory warning when viewing feed.

Would need to finish wiring individual icons were addressable (so individual icons can have indicator toggled) and themes reapplied when they change.

screen shot 2018-03-01 at 1 49 33 pm


montehurd:
Closing in favor of using `UITabBarItem` `badgeValue` with an existing character. Can maybe use a custom glyph character later if needed.

PR #2087, opened 2018-03-01
Saved progress indicators 

montehurd:
https://phabricator.wikimedia.org/T187823

- [x] Saved tab bar icon badge
- [x] Saved tab UIProgressView
- [x] Theme-ing
- [x] Handle long translation string (per Carolyn had the progress panel grow as tall as it needs to)

screen shot 2018-03-20 at 7 12 01 pm

There are a couple remaining todo's which it sounds like it's fine to address as follow-ons (per Josh and Carolyn), so removing WIP tag.


Wikipedia/Code/WMFAppViewController.m
```diff
@@ -302,6 +307,18 @@ - (void)appLanguageDidChangeWithNotification:(NSNotification *)note {
     [self configureExploreViewController];
 }
 
+- (void)readingListsSyncProgressDidChange:(NSNotification *)note {
+    NSNumber *progress = note.userInfo[WMFReadingListsController.syncProgressDidChangeFractionCompletedKey];
+
+    // TODO: minimum busy time before showing and a minimum show length
+    UITabBarItem *item = [[self navigationControllerForTab:WMFAppTabTypeSaved] tabBarItem];
```
montehurd: @nambatee  @joewalsh I used the `reading list danger zone` to add 10 lists with 10 articles but `progress` was always `1` here. 

I'm looking at the `Initial progress implementation` PR https://github.com/wikimedia/wikipedia-ios/pull/2077 to get a handle on what's happening with the fraction completed logic, but any pointers are welcome! :)

montehurd: (for future reference: ^ this bug was iOS 10.3 only)

montehurd:
Open questions:

- Best way to make `readingListsSyncProgressDidChange` reflect new ticket requirement: 

> The progress indicator should reflect the state of the article data downloaded to the user's device, not just the status received from the API (eg. when the progress bar is filled/complete, this should mean that all of the article data including images is saved to the user's device).

- Why is `readingListsSyncProgressDidChange` firing when sync is disabled?

coreyfloyd:
@montehurd is this what you are currently working on? Or is this blocked?

montehurd:
@coreyfloyd Yes this is the one.

montehurd:
I'll push more WIP commits on this later today with some open questions I have.

montehurd:
@nambatee This should be ready for review 😀

nambatee:
anyone else crashing on `NSException *** Collection <__NSArrayM: 0x60c000640570> was mutated while being enumerated.` when tapping on "Saved" tab bar item?

Wikipedia/Code/SavedArticlesFetcher.m
```diff
@@ -81,6 +87,29 @@ - (instancetype)initWithDataStore:(MWKDataStore *)dataStore
                   imageInfoFetcher:[[MWKImageInfoFetcher alloc] init]];
 }
 
+#pragma mark - Progress
+
+// Reminder: due to the internal structure of this class and how it is presently being used, we can't simply check the 'count' of 'fetchOperationsByArticleTitle' dictionary for the total. (It doesn't reflect the actual total.) Could re-plumb this class later.
+- (NSUInteger)calculateTotalArticlesToFetchCount {
```
coreyfloyd: Any reason to always require the DB rather than just keep the fetch in memory and observe the changes (This may require an internal NSFRC but maybe not). 

I guess this is probably less error prone. Im not sure if the performance is a concern. 

Well actually, performance may be a concern since you are using the "viewContext". Actually, why not use a background context and then at least the UI won't ever get held up while performing the fetch.

Also is do you need to ensure you are on the main queue? Or should you use the a block from the moc to make sure you are on the right queue?

Wikipedia/Code/SavedArticlesFetcherProgressManager.swift
```diff
@@ -0,0 +1,63 @@
...
+    private static func newProgress() -> Progress {
+        return Progress.discreteProgress(totalUnitCount: -1)
+    }
+
+    private func resetProgress() {
+        // Reminder - see the note above for why this gets reset after finish.
+        progress = SavedArticlesFetcherProgressManager.newProgress()
+    }
+    
+    var fetcher: SavedArticlesFetcher? {
```
coreyfloyd: Any reason not to init this class with the fetcher?

Wikipedia/Code/SavedArticlesFetcher.m
montehurd: I'll check...

Wikipedia/Code/SavedArticlesFetcherProgressManager.swift
```diff
@@ -0,0 +1,63 @@
+/*
+ This class acts as a bridge between the SavedArticlesFetcher (which retrieves article html and images) and VC's which want to display save article fetching progress. It does so while being minimally invasive to SavedArticlesFetcher, observing only the SavedArticlesFetcher's 'fetchesInProcessCount'. Interested VC's can access the SavedArticlesFetcherProgressManager.shared.progress - but be aware that this "progress" object is not re-used, so you'll need to observe it to know when it gets re-set, and you'll then have to observe any properties you are interested in of the re-set progress. Reason:
+ 
+ "NSProgress objects cannot be reused. Once they’re done, they’re done. Once they’re cancelled, they’re cancelled. If you need to reuse an NSProgress, instead make a new instance and provide a mechanism so the client of your progress knows that the object has been replaced, like a notification." ( Source: https://developer.apple.com/videos/play/wwdc2015/232/ by way of https://stinkykitten.com/index.php/2017/08/13/nsprogress/ )
+*/
+
+@objcMembers class SavedArticlesFetcherProgressManager: NSObject, ProgressReporting {
+
+    static let shared = SavedArticlesFetcherProgressManager()
```
coreyfloyd: Is there only one saved articles fetcher? If so, couldn't this just be a property on that class? it seems tightly coupled to that class so it would make sense. Also would get rid of the this being shared and make it easier to kill.

montehurd:
@nambatee 

> anyone else crashing on NSException *** Collection <__NSArrayM: 0x60c000640570> was mutated while being enumerated. when tapping on "Saved" tab bar item?

Oh weird no I've never seen that. Do you know which array it's referring to?

Wikipedia/Code/WMFAppViewController.m
```diff
@@ -1045,6 +1062,9 @@ - (SavedArticlesFetcher *)savedArticlesFetcher {
             [[SavedArticlesFetcher alloc] initWithDataStore:[[SessionSingleton sharedInstance] dataStore]
 
                                               savedPageList:[self.dataStore savedPageList]];
+        
+        SavedArticlesFetcherProgressManager* savedArticlesProgressManager = [SavedArticlesFetcherProgressManager shared];
+        savedArticlesProgressManager.fetcher = _savedArticlesFetcher;
```
coreyfloyd: it seems like this only gets set once and is always associated with the fetcher. So I think it could make sense to make this just a property of the fetcher itself. (Unless there is something else I am missing which is entirely possible)

coreyfloyd:
My major comment was around the use of the MOC and updating the fetch. Secondary was structuring the code to remove the shared class property and instead making it a property on the fetcher itself.

Aside from those and the crash that @nambatee found and the looks pretty good to me code wise

coreyfloyd:
Oh I meant to say… Natalia's crash and my comments on the MOC may be related

Wikipedia/Code/SavedArticlesFetcherProgressManager.swift
montehurd: Yes I struggled with this.

The SavedArticlesFetcher is currently only init'ed by the app view controller, but there's more than one thing that needs to have a way to know about the saved article fetching progress:
- the progress view (currently using it) 
- the icon badge (currently only reflecting the reading list API activity - this was one of the follow-up issues - Josh is aware and we'll prob chat about it at meeting tomorrow)

So making the one instance of the SavedArticlesFetcher accessible to these other areas means we'd need a way to expose it. I was reluctant to make the whole thing just a singleton for this purpose, and my initial attempt to relay it otherwise were (way) more invasive, so I settled on this manager which was solely responsible for exposing (via `shared` and conformance to `ProgressReporting`) and managing as much progress logic as it could.

I can take a stab again at exposing the SavedArticlesFetcher instead - perhaps I should just make it a singleton?

Wikipedia/Code/SavedArticlesFetcherProgressManager.swift
montehurd: Oops I'll try.

Wikipedia/Code/WMFAppViewController.m
montehurd: Context in my earlier comment: https://github.com/wikimedia/wikipedia-ios/pull/2087/files#r176246309

nambatee:
Corey's comments sound good, looks good to me 👍

montehurd:
@nambatee Are you able to repro the crash on clean install?

nambatee:
@montehurd let me check on clean install, first time it was when I had ~5000 articles, and the badge & progress view were visible

nambatee:
@montehurd no crash on clean install, it's still reproducible if you add articles on one device and then open the previously installed app (while being logged in & sync enabled) on another device

montehurd:
@nambatee oh sweet thanks for the repro steps i'll take a look

Wikipedia/Code/SavedArticlesFetcher.m
montehurd: > Any reason to always require the DB rather than just keep the fetch in memory and observe the changes (This may require an internal NSFRC but maybe not).

Ya the reason is the way the `_update` method currently works, which is a bit odd and hard to follow. It's using the same context and predicate and an `updateAgain` closure to recursively continue kicking off `fetchArticleURL` calls, one by one, until they've all wrapped up. These `fetchArticleURL` calls are what actually populate a `fetchOperationsByArticleTitle` dict, but the `fetchOperationsByArticleTitle` `count` doesn't reflect the total we need - it's usually `0` or `1` because of the aforementioned structure, so we can't monitor it for the total.

> Well actually, performance may be a concern since you are using the "viewContext". Actually, why not use a background context and then at least the UI won't ever get held up while performing the fetch.

I think because I'm not fetching (inflating) the actual `WMFArticle`'s, but rather only the count matching the predicate, it's pretty snappy. On my machine it looks to be around 0.0005 seconds with only a few articles, but also around 0.0005 seconds with a hundred or so articles, so seems safe.

Re the view context, it's the same one used everywhere else within the `SaveArticlesFetcher`, (such as in its `update` method). Was hesitant to re-plumb more than necessary at this time. I can add a main thread assert (as seen in the `update` method) to make sure we're on the expected one.

coreyfloyd: > These fetchArticleURL calls are what actually populate a fetchOperationsByArticleTitle dict, but the fetchOperationsByArticleTitle count doesn't reflect the total we need - it's usually 0 or 1 because of the aforementioned structure, so we can't monitor it for the total.

Oh I wasn't proposing monitoring the dictionary. I am asking rather than executing a DB fetch every time to update the count, why not fetch once and then update the results by monitoring changes to the MOC… BUT I think that complexity is probably not warranted given your performance numbers - so disregard :)

> I think because I'm not fetching (inflating) the actual WMFArticle's, but rather only the count matching the predicate, it's pretty snappy. On my machine it looks to be around 0.0005 seconds with only a few articles, but also around 0.0005 seconds with a hundred or so articles, so seems safe.

> Re the view context, it's the same one used everywhere else within the SaveArticlesFetcher, (such as in its update method). Was hesitant to re-plumb more than necessary at this time. 

Hmm… interesting… do we know why we are using the view context in a fetcher that I am guessing is running a "background process"? Maybe there is something I am missing here… I wouldn't want you to change this without understanding why.

Usually, if DB fetches can be done off the main thread, it is good to do so (In general, I wouldn't use the view context for fetching object unless I then had to the manipulate the fetched objects on the main thread).

Who knows what types of environmental conditions may arise that cause the fetch to be slow (old device, large DB, constrained memory, etc). So even if performance is good in your testing, is such issues come up for users in the real world, the UI may lock up and could lead to hard to debug performance issues down the road.

> I can add a main thread assert (as seen in the update method) to make sure we're on the expected one.

Depends… is this method designed only to be called from the main thread? If so, then yes. But if it is expected to be able to be called from any thread, I would use the MOC block method to ensure the work is done on the right queue.

Wikipedia/Code/SavedArticlesFetcherProgressManager.swift
coreyfloyd: Got it… so you are saying that progress needs to be known by several objects, but those objects don't necessarily need to be able to control the fetcher - is that right?

In that case what sticks out to me is that the fetcher is just global anyways… it is a variable on the progress manager. And the progress manager is global. 

So by that logic… if the appVC nils out it's fetcher, the progress manager would still be retaining the old one. So that feels strange to me. 

Basically in the current setup, the progress manager is kinda controlling the life cycle of the fetcher… which feels inverted to me.

Honestly, the progress feels like it could be owned by the fetcher. I mean for the most part I don't think anything else in the system even needs to know about the progress manager class except for the fetcher. If the fetcher just exposed the resulting progress of the manager as a var that could be monitored, it could hide the complexity of the manager class from the rest of the world - which may be nice. 

But given your other comments… lets just say that the progress manager is now a variable of the fetcher and any object that has access to the fetcher can monitor the progress… do you actually need to make the fetcher global? Can it be passed by dependency injection instead? Or is it just to "far away" to do cleanly?

Wikipedia/Code/SavedArticlesFetcher.m
montehurd: > Hmm… interesting… do we know why we are using the view context in a fetcher that I am guessing is running a "background process"? Maybe there is something I am missing here… I wouldn't want you to change this without understanding why.

No, I wasn't sure why it was this way. I'll take a look at the blame log for hints...

Wikipedia/Code/SavedArticlesFetcherProgressManager.swift
montehurd: > Got it… so you are saying that progress needs to be known by several objects, but those objects don't necessarily need to be able to control the fetcher - is that right?

Yes I believe so.

> So by that logic… if the appVC nils out it's fetcher, the progress manager would still be retaining the old one.n Basically in the current setup, the progress manager is kinda controlling the life cycle of the fetcher… which feels inverted to me.

Ah, good point on the inversion/lifecycle, though presently the appVC doesn't appear to ever nil out the fetcher. 

> If the fetcher just exposed the resulting progress of the manager as a var that could be monitored, it could hide the complexity of the manager class from the rest of the world - which may be nice.

That does sound better, though I'll have to check again on the "too far away" issue you mentioned. I'll take a peek...

montehurd: Looking at the view hierarchy this may not be so bad... taking a stab at it...

montehurd:
@coreyfloyd 

I pushed some updates. 

- Mainly the SavedArticlesFetcher is now the only one with knowledge of the SavedArticlesFetcherProgressManager.
- SavedArticlesFetcherProgressManager no longer singleton.

I did use a shared object to keep the wiring easier, but now it's a single property object rather than the whole manager.

coreyfloyd:
@montehurd cool… looks good 💪 Will merge after the test failure is fixed.

I suggest flag the viewContext usage with a warning to be investigated why it is being used and flip it to a background context if it can be at a later time. (nothing to worry about this minute)

Wikipedia/Code/SavedArticlesFetcher.m
```diff
@@ -48,39 +43,64 @@ - (void)dealloc {
...
                     articleFetcher:[[WMFArticleFetcher alloc] initWithDataStore:dataStore]
                    imageController:[WMFImageController sharedInstance]
                   imageInfoFetcher:[[MWKImageInfoFetcher alloc] init]];
 }
 
+#pragma mark - Progress
+
+// Reminder: due to the internal structure of this class and how it is presently being used, we can't simply check the 'count' of 'fetchOperationsByArticleTitle' dictionary for the total. (It doesn't reflect the actual total.) Could re-plumb this class later.
+- (NSUInteger)calculateTotalArticlesToFetchCount {
+    NSAssert([NSThread isMainThread], @"Must be called on the main thread");
```
michalciurus: It doesn't really matter as it's just an assert but it may be of interest to you, I know I was surprised:

[I think it should say](https://developer.apple.com/documentation/coredata/nsmanagedobjectcontextconcurrencytype/nsmainqueueconcurrencytype) "Must be called on the main queue" not "thread" to be precise. Queue != Thread

Queues reuse threads, so it may happen that a background queue will execute on a main thread and [apparently it's very easy to reproduce](http://blog.benjamin-encz.de/post/main-queue-vs-main-thread/).

[More on that topic](http://blog.krzyzanowskim.com/2016/06/03/queues-are-not-bound-to-any-specific-thread/)

As I said, not that it matters or produces any problem but just a little brain teaser 😁

montehurd: Oh thanks! Good catch!

montehurd: That `main-queue-vs-main-thread` article is great!!!

montehurd:
retest this please

montehurd:
retest this please

PR #2089, opened 2018-03-02
Fix duplicate lists on initial sync

joewalsh:
Ensure reading list IDs are associated with created lists.

### DIFF
--- a/WMF Framework/ReadingListsSyncOperation.swift
+++ b/WMF Framework/ReadingListsSyncOperation.swift
@@ -417,10 +417,10 @@ internal class ReadingListsSyncOperation: ReadingListsOperation {
                     return
                 }
                 guard let readingListID = localReadingList.readingListID?.int64Value else {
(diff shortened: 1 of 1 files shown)


WMF Framework/ReadingListsSyncOperation.swift
```diff
@@ -455,36 +455,37 @@ internal class ReadingListsSyncOperation: ReadingListsOperation {
...
                         if let apiError = createError as? APIReadingListError {
-                            for list in listsToCreate {
+                            for list in listsToCreateSubarray {
                                 failedReadingLists.append((list, apiError))
                             }
                         }
                         return
                     }
                     for (index, readingList) in readingListIDs.enumerated() {
-                        guard (start + index) < listsToCreate.count else {
```
joewalsh: Start isn't captured when this block is created, it's equivalent to an __block as a var, and it's value is past the end of the array when this callback is called.

PR #2090, opened 2018-03-02
ensure we don't deselect the selected tabBarItem when presenting alerts on iPad

nambatee:
https://phabricator.wikimedia.org/T187515

Before
simulator screen shot - ipad air - 2018-03-01 at 19 39 37

After
simulator screen shot - ipad air - 2018-03-01 at 19 36 41


nambatee:
Instead of asking for the view of the selected tab bar item and setting `passthroughViews` to include that view, I can set `passthroughViews` to include the entire tab bar. However, if I do that, after dismissing the alert (via `tabBarController:(UITabBarController *)tabBarController didSelectViewController:(UIViewController *)viewController` in `WMFAppViewController`) the view that was presenting the alert will not know that the alert was dismissed and as a result, will not update the `tintColor` of its buttons.

montehurd:
I added quick note to the ticket: https://phabricator.wikimedia.org/T187515#4024679

montehurd:
Looks like setting 

> tabBar.tabBar.tintAdjustmentMode = UIViewTintAdjustmentModeNormal;

on the line before 

> self.rootTabBarController = tabBar;

in `loadMainUI` fixes it

screen shot 2018-03-05 at 11 19 47 am

PR #2095, opened 2018-03-02
make hint tap gesture recognizers optional to avoid crash

nambatee:


Wikipedia/Code/ReadingListHintViewController.swift
```diff
@@ -71,8 +76,12 @@ class ReadingListHintViewController: UIViewController {
     
     deinit {
         NotificationCenter.default.removeObserver(self)
-        hintView?.removeGestureRecognizer(tapGestureRecognizer.hint)
-        confirmationView?.removeGestureRecognizer(tapGestureRecognizer.confirmation)
+        if let tapHintGestureRecognizer = tapHintGestureRecognizer {
```
montehurd: Looks like you don't have to remove recognizers in de-init ( https://stackoverflow.com/a/47209664/135557 ) - were the `removeGestureRecognizer` calls triggering the crash?

PR #2097, opened 2018-03-02
Make the gesture recognizer in WMFTableOfContentsAnimator a let

nambatee:


montehurd:
Oh, sorry, my comment in chat...

> btw i see another place where we're dealing with an optional gesture recognizer in a deinit - WMFTableOfContentsAnimator - in a similar fashion.

... was just pointing out there was another place we handle these like you did in https://github.com/wikimedia/wikipedia-ios/pull/2095.

Did you notice a crash here in WMFTableOfContentsAnimator?

nambatee:
@montehurd this will ensure that the lazy var isn't instantiated in deinit

montehurd:
Thanks. I think that generally we shouldn't have vars be lazy if we need to touch them in deinit, I was just wondering if this particular case had crashes.

montehurd:
(Not saying we shouldn't merge this if no crashes, just curious.)

nambatee:
didn't notice any crashes here

coreyfloyd:
@nambatee is this ready to go? I see it has been sitting for a bit

PR #2100, opened 2018-03-06
As a user of 'Save for later' I would like to be able to continue saving articles without sorting them into reading lists even when I have a lot of saved articles

nambatee:
This PR covers showing the "Limit hit for unsorted articles" panel when the user hits 5000 one-tap articles / unsorted articles.

I also removed the hardcoded max entries per list limit and started using the reading lists config we get from site info.


montehurd:
It's generally easier to grok/review individual commits with their own commit messages over single commits with large commit messages. Would you mind breaking the logical parts of the first commit up? (The Xcode diff viewer makes this pretty easy to do by checking/unchecking which parts you want in which commits.)

I'll try to be better about doing this myself to keep your CR of my code more digestible too :)

Wikipedia/Code/MWKSiteInfo.h
```diff
@@ -12,8 +12,13 @@ NS_ASSUME_NONNULL_BEGIN
 /// Raw title for the receiver's main page.
 @property (readonly, copy, nonatomic) NSString *mainPageTitleText;
 
+@property (readwrite, copy, nonatomic) NSNumber *readingListsConfigMaxEntriesPerList;
```
montehurd: Looks like `readingListsConfigMaxEntriesPerList` and `readingListsConfigMaxListsPerUser` should be `readonly` here.

Wikipedia/Code/MWKSiteInfo.m
```diff
@@ -14,16 +14,21 @@ typedef NS_ENUM(NSUInteger, MWKSiteInfoNSCodingSchemaVersion) {
 @interface MWKSiteInfo ()
 @property (readwrite, copy, nonatomic) NSURL *siteURL;
 @property (readwrite, copy, nonatomic) NSString *mainPageTitleText;
+
```
montehurd: Here's where the `readwrite` versions of `readingListsConfigMaxEntriesPerList` and `readingListsConfigMaxListsPerUser` can go. That way internally we can write to them but externally they're immutable.

Wikipedia/Code/ReadingListsController.swift
```diff
@@ -1,16 +1,20 @@
 import Foundation
 
-internal let WMFReadingListSyncStateKey = "WMFReadingListsSyncState"
-internal let WMFReadingListDefaultListEnabledKey = "WMFReadingListDefaultListEnabled"
-fileprivate let WMFReadingListSyncRemotelyEnabledKey = "WMFReadingListSyncRemotelyEnabled"
+// Sync keys
```
montehurd: Looks like all of these can remain restricted to WMF module via `internal`.

Wikipedia/Code/ReadingListsController.swift
```diff
@@ -1,16 +1,20 @@
...
+let WMFReadingListSyncStateKey = "WMFReadingListsSyncState"
+private let WMFReadingListSyncRemotelyEnabledKey = "WMFReadingListSyncRemotelyEnabled"
```
montehurd: This can remain `fileprivate`.

Wikipedia/Code/ReadingListsController.swift
```diff
@@ -1,16 +1,20 @@
...
-internal let WMFReadingListBatchSizePerRequestLimit = 500
+// Batch size keys
+let WMFReadingListBatchSizePerRequestLimit = 500
+let WMFReadingListCoreDataBatchSize = 500
 
-internal let WMFReadingListCoreDataBatchSize = 500
-
-internal let WMFReadingListEntryLimit = 1000
+// Reading lists config keys
+let WMFReadingListsConfigMaxEntriesPerList = "WMFReadingListsConfigMaxEntriesPerList"
```
montehurd: These new one can be use `internal` access control too.

WMF Framework/ReadingList+CoreDataClass.swift
```diff
@@ -14,17 +17,21 @@ public class ReadingList: NSManagedObject {
         return existingKeys
     }
     
-    public func updateCountOfEntries() {
```
montehurd: Looks like `updateCountOfEntries` wasn't used, but do you recall if its `entry.isDeletedLocally` filtering logic isn't needed? Just curious.

Wikipedia/Code/ReadingListsAlertController.swift
```diff
@@ -68,4 +68,25 @@ public class ReadingListsAlertController: NSObject {
         actions.forEach { alert.addAction($0) }
         return alert
     }
+    
+    // MARK: - ScrollableEducationPanelViewController presentation
+    
+    @objc func showLimitHitForDefaultListPanelIfNecessary(presenter: UIViewController, dataStore: MWKDataStore, readingList: ReadingList, theme: Theme) {
+        guard Thread.isMainThread, dataStore.readingListsController.isSyncEnabled else {
```
montehurd: Consider moving the main thread check to its own guard statement with a `assertionFailure("Expected main thread")` so we'll know right away if this ever happens on an unexpected thread.

Wikipedia/Code/ReadingListsAlertController.swift
```diff
@@ -75,6 +75,9 @@ public class ReadingListsAlertController: NSObject {
         guard Thread.isMainThread, dataStore.readingListsController.isSyncEnabled else {
             return
         }
+        guard !UserDefaults.wmf_userDefaults().wmf_didShowLimitHitForUnsortedArticlesPanel() else {
```
montehurd: Does this need to show any time the limit is hit vs just once?

montehurd:
Looks pretty good overall - added a few small comments.

Wikipedia/Code/ReadingListsController.swift
nambatee: Internal is the default access control so we don't need to add it explicitly https://developer.apple.com/swift/blog/?id=5

Wikipedia/Code/ReadingListsController.swift
nambatee: private is more restrictive and it looks like we don't need to make it fileprivate in this case

Wikipedia/Code/ReadingListsController.swift
nambatee: same as above, internal is the default

WMF Framework/ReadingList+CoreDataClass.swift
nambatee: inside `updateArticlesAndEntries`  we're updating `countOfEntries` with the count of `validEntries`, to get validEntries we're filtering on `isDeletedLocally`, `let validEntries = (entries ?? []).filter { !$0.isDeletedLocally }`, line `37` in the same file

Wikipedia/Code/ReadingListsController.swift
montehurd: > private is more restrictive

whoa interesting swift 4 change! reading that now... (good catch!)

WMF Framework/ReadingList+CoreDataClass.swift
montehurd: Cool thanks!!!

Wikipedia/Code/ReadingListsAlertController.swift
montehurd: Per https://phabricator.wikimedia.org/T188057#4036175 just once.

PR #2101, opened 2018-03-06
Update the name of the Default list from Bookmarks to Saved to match Android

nambatee:
https://phabricator.wikimedia.org/T188662


WMF Framework/CommonStrings.swift
```diff
@@ -86,7 +86,7 @@ public class CommonStrings: NSObject {
     
     @objc public static let unknownError = WMFLocalizedString("error-unknown", value: "An unknown error occurred", comment: "Message displayed when an unknown error occurred")
     
-    @objc public static let readingListsDefaultListTitle = WMFLocalizedString("reading-lists-default-list-title", value: "Bookmarks", comment: "The title of the default saved pages list")
+    @objc public static let readingListsDefaultListTitle = WMFLocalizedString("reading-lists-default-list-title", value: "Saved", comment: "The title of the default saved pages list\n{{Identical|Saved}}")
```
montehurd: It's fine to add `\n{{Identical|Saved}}` here, but generally you don't have to - TWN folks (or an upstream script?) usually add these then next time we sync we get them.

montehurd: Looks like `readingListsDefaultListTitle` is also used in `ReadingList+CoreDataProperties.swift` to determine if a list is the `default` one - I only mention this in case we notice any weirdness with the old default list created before this change going forward.

nambatee: in `ReadingList+CoreDataProperties.swift` `readingListsDefaultListTitle ` is used so that we give the correct reading list name to the views since the default list has no `canonicalName`

PR #2105, opened 2018-03-08
As a user of reading lists, I would like to have a place in Settings to manage my reading list preferences

nambatee:
https://phabricator.wikimedia.org/T187965

additionally fixes https://phabricator.wikimedia.org/T189015

what to watch out for:

~~in the current implementation, the Erase saved articles footer label ("Erasing your saved articles will remove...") will not fit longer text. Potential solutions: 1) layout manually so we can get the placeholder height before the view loads 2) embed it in a scroll view 3) something else that I haven't thought of, suggestions welcome~~


nambatee:
@montehurd the tests are failing because I changed the `content mode` of the `WMFSettingsTableViewCell` `titleIcon` from `aspectFit` to `center` - ok if I regenerate reference images based on the new content mode?

Wikipedia/Localizations/ast.lproj/Localizable.strings
```diff
@@ -324,8 +324,7 @@
 "reading-list-api-error-list-limit" = "Esta llista nun ta sincronizada porque aportasti a la llende del númberu de llistes sincronizaes.";
 "reading-list-deleted-accessibility-notification" = "Desaniciada la llista de llectura";
 "reading-list-do-not-keep-button-title" = "Non, desaniciar los artículos del preséu";
-// Fuzzy
```
montehurd: One trick I use in cases where develop has moved ahead since I started a branch but before I've started to split things into individual commits is rebasing on develop before using the diff viewer to split into commits - that way my diff only reflects my changes.

Wikipedia/Localizations/en.lproj/Localizable.strings
```diff
@@ -328,11 +328,10 @@
 "reading-list-login-title" = "Sync your saved articles?";
 "reading-list-new-list-description-placeholder" = "optional short description";
 "reading-list-new-list-name-placeholder" = "reading list title";
-"reading-list-show-default-list-setting" = "Show default list";
 "reading-list-sync-enable-button-title" = "Enable syncing";
 "reading-list-sync-enable-subtitle" = "Your saved articles and reading lists can now be saved to your Wikipedia account and synced across devices.";
 "reading-list-sync-enable-title" = "Turn on reading list syncing?";
-"reading-list-sync-setting" = "Reading list sync";
+"reading-list-sync-setting" = "Saved articles storage and syncing";
```
montehurd: Looks like this uses same string as `settings-storage-and-syncing-title`. Consider consolidating these and exposing the shared string via CommonString if it seems like they're most likely going to remain the same.

Wikipedia/Code/StorageAndSyncingSettingsViewController.swift
```diff
@@ -0,0 +1,213 @@
...
+        
+        let eraseSavedArticlesSection = SettingsSection(headerTitle: nil, footerText: nil, items: [eraseSavedArticles])
+        let syncWithTheServerSection = SettingsSection(headerTitle: nil, footerText: WMFLocalizedString("settings-storage-and-syncing-server-sync-footer-text", value: "Request a sync from the server for an update to your synced articles and reading lists", comment: "Footer text of the settings button that initiates saved articles and reading lists server sync"), items: [syncWithTheServer])
+        
+        return [syncSavedArticlesAndListsSection, showSavedReadingListSection, eraseSavedArticlesSection, syncWithTheServerSection]
+    }
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        title = WMFLocalizedString("settings-storage-and-syncing-title", value: "Saved articles storage and syncing", comment: "Title of the saved articles storage and syncing settings screen")
```
montehurd: Something weird is happing when this text appears and after rotate - it jumps a bit on appearance and is vertically offset a bit after rotate.

Wikipedia/Code/StorageAndSyncingSettingsViewController.swift
```diff
@@ -0,0 +1,213 @@
+struct SettingsSection {
```
montehurd: This struct can be private, at which point we can just call it `Section`.

Wikipedia/Code/StorageAndSyncingSettingsViewController.swift
```diff
@@ -0,0 +1,213 @@
...
+    let headerTitle: String?
+    let footerText: String?
+    let items: [SettingsItem]
+}
+
+struct SettingsItem {
```
montehurd: This can be private too, then simplified to `Item`.

Wikipedia/Code/StorageAndSyncingSettingsViewController.swift
```diff
@@ -0,0 +1,213 @@
...
+    let itemType: StorageAndSyncingSettingsItemType
+    let title: String?
+    let iconName: String?
+    let iconColor: UIColor?
+    let iconBackgroundColor: UIColor?
+    let isSwitchOn: Bool?
+    let buttonTitle: String?
+}
+
+enum StorageAndSyncingSettingsItemType: Int {
```
montehurd: Also a candidate for being made private, then simplified to `ItemType`.

Wikipedia/Code/StorageAndSyncingSettingsViewController.swift
```diff
@@ -0,0 +1,213 @@
+struct SettingsSection {
+    let headerTitle: String?
+    let footerText: String?
+    let items: [SettingsItem]
+}
+
+struct SettingsItem {
+    let disclosureType: WMFSettingsMenuItemDisclosureType?
```
montehurd: `WMFSettingsMenuItemDisclosureType` has a None case so doesn't need to be optional.

Wikipedia/Code/StorageAndSyncingSettingsViewController.swift
```diff
@@ -0,0 +1,213 @@
...
+    let itemType: StorageAndSyncingSettingsItemType
+    let title: String?
+    let iconName: String?
```
montehurd: Looks like `iconName` is never not `nil` so can be removed from Item and just pass nil to WMFSettingsMenuItemDisclosureType's configure method (`iconColor` and `iconBackgroundColor` too)

Wikipedia/Code/StorageAndSyncingSettingsViewController.swift
```diff
@@ -0,0 +1,213 @@
...
+    let iconColor: UIColor?
+    let iconBackgroundColor: UIColor?
+    let isSwitchOn: Bool?
```
montehurd: `isSwitchOn` would be simpler if not optional - just default it to false and use ` ?? false`  at the end of the `let showSavedReadingList = ` line.

Wikipedia/Code/StorageAndSyncingSettingsViewController.swift
```diff
@@ -0,0 +1,213 @@
...
+    let buttonTitle: String?
+}
+
+enum StorageAndSyncingSettingsItemType: Int {
+    case syncSavedArticlesAndLists, showSavedReadingList, eraseSavedArticles, syncWithTheServer
+    
+    public func settingsItem(isSwitchOnByDefault: Bool? = nil) -> SettingsItem {
```
montehurd: This func is essentially an Item constructor, so should be part of the Item struct as something like: `init(for itemType: StorageAndSyncingSettingsItemType, disclosureType: WMFSettingsMenuItemDisclosureType, ...`

Wikipedia/Code/StorageAndSyncingSettingsViewController.swift
```diff
@@ -0,0 +1,213 @@
...
+        
+        return SettingsItem(disclosureType: disclosureType, itemType: self, title: title, iconName: nil, iconColor: nil, iconBackgroundColor: nil, isSwitchOn: isSwitchOn, buttonTitle: buttonTitle)
+    }
+}
+
+@objc(WMFStorageAndSyncingSettingsViewController)
+class StorageAndSyncingSettingsViewController: UIViewController {
+    private let customViewCellReuseIdentifier = "CustomViewTableViewCell"
+    private var theme: Theme = Theme.standard
+    private var tableView: UITableView!
```
montehurd: Using an IBOutlet & storyboard(or xib) for the `tableView` would eliminate the bang here and the boilerplate tableview config code in `viewDidLoad`.

Wikipedia/Code/StorageAndSyncingSettingsViewController.swift
```diff
@@ -0,0 +1,213 @@
...
+    }
+    
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+        tableView.reloadRows(at: indexPathsForCellsWithSwitches, with: .none)
+    }
+    
+    private var isSyncEnabled: Bool {
+        guard let dataStore = dataStore else {
+            return false
```
montehurd: Consider an assertionFailure msg here if we don't expect this to ever happen but would like to know if it starts happening.

montehurd:
> @montehurd the tests are failing because I changed the content mode of the WMFSettingsTableViewCell titleIcon from aspectFit to center - ok if I regenerate reference images based on the new content mode?

Was the image added not centering properly? If so can we get an updated asset? I think we want to keep `aspectFit` if possible.

montehurd:
Could you move the https://phabricator.wikimedia.org/T189015 bits to a separate PR? It would be easier to grok & comment about it separately - it's changing some panel parts I have questions about and I don't want those parts to block these new settings interface bits :)

Wikipedia/Code/StorageAndSyncingSettingsViewController.swift
montehurd: (I don't feel super strongly about this, but it would be nice.)

nambatee:
@montehurd do you want to ask questions about the panel in the ticket? I'd like to keep it here cause it touches the Settings

Wikipedia/Code/StorageAndSyncingSettingsViewController.swift
nambatee: I'd rather keep it an optional here

montehurd: Why is that? WMFSettingsMenuItemDisclosureType is an enum. As an optional it gets wrapped inside *another* enum...
> Under the hood optional types are merely an enum with two cases — `None` meaning no value is set and `Some` meaning the value is set and it’s the value associated with it. [sauce](https://hackernoon.com/swift-optionals-explained-simply-e109a4297298)

montehurd:
I do think I'll have comments on the T189015 bits, but they should be in a separate PR. If the T189015 bits are needed by this PR just add a note in the description that this PR needs the other one merged first.

nambatee:
@montehurd if you don't have comments on those bits, why can't you review them as part of this PR? It's a small change and I don't want to go back and split it. I split all of my previous PRs like you asked, I would appreciate it if you met me half way here.

coreyfloyd:
@nambatee need any review yet?

nambatee:
@coreyfloyd still figuring out why I'm truncating the label on X's landscape 🤦🏻‍♀️

nambatee:
taking off do not merge and putting changes welcome, I can't figure out why the middle label gets truncated on iPhone X rotation. I will revisit this, up to the reviewer if it should happen in this pr or if ok merging what we have and patching later

simulator screen shot - iphone x - 2018-03-13 at 21 47 25

Wikipedia/Code/StorageAndSyncingSettingsViewController.swift
```diff
@@ -183,6 +182,7 @@ extension StorageAndSyncingSettingsViewController: UITableViewDelegate {
         guard let footer = self.tableView(tableView, viewForFooterInSection: section) as? WMFTableHeaderFooterLabelView else {
             return 0
         }
+        footer.prepareForReuse()
```
coreyfloyd: Is there a reason you need to invoke this directly at all? I thought `prepareForReuse` was invoked by the framework automatically?

Or is there a side effect here of calling this? (maybe this is also a pattern that is being use elsewhere?)

coreyfloyd: nm… out of date

coreyfloyd:
Ok, this looks good to me… waiting on tests to complete

Wikipedia/Code/Panels.swift
```diff
@@ -63,13 +63,28 @@ class ReLoginFailedPanelViewController : ScrollableEducationPanelViewController
...
         primaryButtonTitle = WMFLocalizedString("reading-list-login-or-create-account-button-title", value:"Log in or create account", comment:"Title for button to login or create account to sync saved articles and reading lists.")
     }
+    
+    override func primaryButtonTapped(_ sender: Any) {
+        tappedPrimaryButton = true
+        super.primaryButtonTapped(sender)
+    }
+    
+    override func viewWillDisappear(_ animated: Bool) {
+        guard !tappedPrimaryButton else {
```
coreyfloyd: Can you say why you are avoiding calling super in this case? This will also prevent all other super classes from running cleanup code (like UIViewController).

Wikipedia/Code/WMFLoginViewController.swift
```diff
@@ -13,6 +13,8 @@ class WMFLoginViewController: WMFScrollViewController, UITextFieldDelegate, WMFC
     @IBOutlet fileprivate var loginButton: WMFAuthButton!
     @IBOutlet weak var scrollContainer: UIView!
     
+    public var enableSyncOnLogin: Bool = false
```
coreyfloyd: So rather than giving the login controller knowledge of sync, another possibility is to create a var for a completion block to run.

This way any object that pushes the login VC can run arbitrary code after login completes (or fails). This both generalizes the problem (allows login to trigger many different actions) and encapsulates logic (keeps the login control only concerned with login). 

Thoughts?

Wikipedia/Code/ScrollableEducationPanelViewController.swift
```diff
@@ -182,7 +182,7 @@ class ScrollableEducationPanelViewController: UIViewController, Themeable {
         dismiss(animated: true, completion: nil)
     }
     
-    @IBAction fileprivate func primaryButtonTapped(_ sender: Any) {
+    @IBAction open func primaryButtonTapped(_ sender: Any) {
```
coreyfloyd: It looks like this class was designed to put actions in the tap handler rather than override the IBAction. 

Either strategy is fine, but now it looks like both are happening. Probably want to settle on one strategy so subclass implementers have an understanding of what is expected

Wikipedia/Code/WMFSettingsTableViewCell.m
```diff
@@ -21,6 +27,11 @@ - (void)setTitle:(NSString *)title {
     self.titleLabel.text = title;
 }
 
+- (void)setButtonTitle:(NSString *)buttonTitle {
```
coreyfloyd: not necessary, but you could also override the getter to read the title from the button (and eliminate the need for storage all together).

coreyfloyd: Oh, is the problem you need to set the string before the view is ever on screen?

Wikipedia/Code/WMFTableHeaderFooterLabelView.m
```diff
@@ -1,44 +1,53 @@
-#import "WMFTableHeaderLabelView.h"
+#import "WMFTableHeaderFooterLabelView.h"
```
coreyfloyd: good generalization!

Wikipedia/Code/Panels.swift
coreyfloyd: @nambatee this is probably the only thing I think would need to be at least commented on before merging. The other things are nice cleanup/questions to consider but non-blocking.

nambatee: yeah, I missed that, thanks! I wanted to avoid calling `dismissHandler` because if user dismisses the panel, I want the sync switch to be turned off but if they tap the button to 'log in', I don't want them to see the switch being turned off right before they go to the login view. With the superclass calling `dismissHandler` in `viewWillDisappear`, I didn't have a good way to prevent that. I decided to have a flag in my own panel, override `viewWillDisappear` and decide whether I want to let `super` call `dismissHandler` or not. But that won't work cause you're right about missing the cleanup part, I guess the `tappedPrimaryButton` flag should be in the superclass or is there a better way to do this?

Wikipedia/Code/WMFLoginViewController.swift
nambatee: that sounds good 👍

Wikipedia/Code/Panels.swift
coreyfloyd: Hmm… I see…

So you may want to change where the dismiss handler is being called in the super class
Or add a flag to the super class to not call the dismiss handler
Or you could clear the button variable so no UI updates happen during dismissal - basically disconnect the UI from the VC
Or you could defer the UI update using a dispatch_async - so the UI update doesn't happen until after the view disappears. 
Also maybe move logic into viewDidDissapear instead?

All or none of those may work… probably have to play around with them to see

Wikipedia/Code/WMFSettingsTableViewCell.m
nambatee: that'll work, thanks for the suggestion 👍

Wikipedia/Code/Panels.swift
nambatee: ended up adding a flag to the superclass, I set it in init because there's at least one case (`wmf_showKeepSavedArticlesOnDevicePanelIfNecessary` in `Panels`, line 228) when we don't want to discard `dismissHandler` simply because we tap on the button (setting it to `true` in `primaryButtonTapped` and then guarding against it in `viewWillDisappear` would prevent `dismissHandler` from happening every time we tap on the button)

PR #2106, opened 2018-03-08
Add UI test cases

michelleXue:
Add UI test cases with 60% code coverage.


WikipediaUITests/WikipediaUITestCase.swift
```diff
@@ -0,0 +1,154 @@
...
+        app.navigationBars["Settings"].buttons["Close"].tap()
+    }
+    
+    /**
+     Scrolls to a particular element until it is rendered in the visible rect
+     - Parameter elememt: the element we want to scroll to
+     */
+    func scrollToElement(element: XCUIElement)
+    {
+        while element.exists == false
```
montehurd: Would some sort of maximum tries safety check make sense here? On my machine the [testOnThisDay]( https://github.com/wikimedia/wikipedia-ios/pull/2106/files#diff-6d50462001956c9435c46081da5a7e9bR1038) func was calling this func which failed to find the element (I think because of the `today` / `yesterday` issue I mentioned) and this `while` loop never stopped until I stopped it manually.

joewalsh:
Closing this for now as inactive. Please re-open if the suggested changes are made

PR #2108, opened 2018-03-09
Update bundled CSS and update Makefile to request consolidated style modules

mdholloway:
The first patch in this PR updates the bundled stylesheets with the latest from Wikimedia production.

The second and third update the modules requested to handle the consolidation work in https://gerrit.wikimedia.org/r/#/q/topic:pagestyles+(status:open+OR+status:merged).

The fourth adds a CSS snippet to hide the (listen) parenthetical in IPA spans.  It has a soft dependency on https://gerrit.wikimedia.org/r/#/c/419843/.

Notes:

- The bundles created by the second patch should have the same styles as in the preceding patch (though not necessarily in the same order); no functional changes are intended by the module consolidation.

- The third patch adds styles to preview.css but these should have no effect in practice.


mdholloway:
This is ready for review.  Note that the last commit I just pushed has a soft dependency on https://gerrit.wikimedia.org/r/#/c/419843/.  The MobileApp dependency for the earlier commits is now live in production.

coreyfloyd:
@montehurd can you look at this when you get a moment?

PR #2111, opened 2018-03-12
[BUG] Edit button is enabled when the only visible reading list is the default one

nambatee:
https://phabricator.wikimedia.org/T189476


WMF Framework/CollectionViewEditController.swift
```diff
@@ -408,18 +417,17 @@ public class CollectionViewEditController: NSObject, UIGestureRecognizerDelegate
...
+            rightBarButtonSystemItem = .done
         case .open:
-            newBarButtonSystemItem.right = UIBarButtonSystemItem.cancel
+            rightBarButtonSystemItem = .cancel
             fallthrough
         case .closed:
             transformBatchEditPane(for: editingState)
         case .empty:
             isBatchEditToolbarHidden = true
-            enabled = false
```
coreyfloyd: Dumb question: Some states set the left button, some set the right button, some set none. 

I would expect all states to set all buttons… otherwise is there a potential to get the buttons in an inconsistent state? Or maybe I am missing how the button state gets cleaned up

nambatee: 👍 good call, fixed

PR #2122, opened 2018-03-21
On clearing all reading list articles disable and re-enable sync instead of hitting delete endpoint for each article

nambatee:
https://phabricator.wikimedia.org/T190326


Wikipedia/Code/StorageAndSyncingSettingsViewController.swift
```diff
@@ -118,7 +118,7 @@ class StorageAndSyncingSettingsViewController: UIViewController {
         let alert = UIAlertController(title: WMFLocalizedString("settings-storage-and-syncing-erase-saved-articles-alert-title", value: "Erase all saved articles?", comment: "Title of the alert shown before erasing all saved article."), message: WMFLocalizedString("settings-storage-and-syncing-erase-saved-articles-alert-message", value: "Erasing your saved articles will remove them from your user account if you have syncing turned on as well as from this device. You cannot undo this action.", comment: "Message for the alert shown before erasing all saved articles."), preferredStyle: .alert)
         let cancel = UIAlertAction(title: CommonStrings.cancelActionTitle, style: .cancel)
         let erase = UIAlertAction(title: CommonStrings.eraseAllSavedArticles, style: .destructive) { (_) in
-            self.dataStore?.readingListsController.unsaveAllArticles({})
+            self.dataStore?.readingListsController.setSyncEnabled(true, shouldDeleteLocalLists: true, shouldDeleteRemoteLists: true)
```
coreyfloyd: Just checking… did this actually work as intended.

Side comment, should we we delete the code in unsaveAllArticles? I assume that is a dead code path now ¯\_(ツ)_/¯

nambatee: yeah, checked and it worked 🎉 will do

nambatee: ok, removed, this branch should be merged after #2121 cause that's where we're removing the 2nd `unsaveAllArticles` call

coreyfloyd: 👍

PR #2125, opened 2018-03-22
Turn on reading list syncing? panel is presented when sync is already enabled

nambatee:
In the current implementation, we're attempting to present the "Turn on reading list syncing?" panel before we kick off the first sync. It means that on the first launch, before we hit line `160` in `ReadingListsSyncOperation`, sync is marked as disabled. You can confirm that that's what's happening if you remove the app while being logged in & having sync enabled, re-install, launch and go to Saved articles storage and syncing in Settings. The switch for Sync saved articles and lists will be off. If you tap back and then go back to that view, the switch will be on.

This PR doesn't solve the problem in Settings, it fixes the panel on launch problem, but in a dubious way. Putting this up with Changes welcome and Question so we can decide how to solve it.


coreyfloyd:
I don't think I quite have the logic figured out, but one thought I have from reading your issue is are you using a number or a bool for the "sync disabled" flag? 

If it is a number you can check if it was ever set (nil) and then if it was enabled or disabled (1 or 0).

This may not be helpful advice… feel free to hit me up and discuss if you like

nambatee:
On the first launch, the number will be nil before we present the panel, which will tell us that it's the first sync. The problem is that it doesn't get set before we present the panel, so at the time of presenting we have no idea if sync is enabled or disabled, we're assuming that nil == sync is disabled

When we finish resuming app, we're attempting to log in with a completion block:

    [self attemptLogin:^{
        [self checkRemoteAppConfigIfNecessary];
        [self.dataStore.readingListsController start];
        [self.savedArticlesFetcher start];
        self.resumeComplete = YES;
    }];

`[self.dataStore.readingListsController start];` call kicks off the sync

in `- (void)attemptLogin:(dispatch_block_t)completion`, we're showing the panel right after dispatching the completion block:

    [[WMFAuthenticationManager sharedInstance] loginWithSavedCredentialsWithSuccess:^(WMFAccountLoginResult *_Nonnull success) {
        DDLogDebug(@"\n\nSuccessfully logged in with saved credentials for user '%@'.\n\n", success.username);
        dispatch_async(dispatch_get_main_queue(), completion);
        [self wmf_showEnableReadingListSyncPanelOncePerLoginWithTheme:self.theme];
    }

in `wmf_showEnableReadingListSyncPanelOncePerLoginWithTheme ` we're guarding against sync being disabled

        guard !UserDefaults.wmf_userDefaults().wmf_didShowEnableReadingListSyncPanel(),
            SessionSingleton.sharedInstance().dataStore.readingListsController.isSyncRemotelyEnabled,
            !SessionSingleton.sharedInstance().dataStore.readingListsController.isSyncEnabled else {
            return
        }

but at that time (first launch), `SessionSingleton.sharedInstance().dataStore.readingListsController.isSyncEnabled` is always false

and that's how we end up with sync disabled on the first launch

We can check if the number is nil (first sync) and if so, present the panel after we complete the sync pass (and find out if sync is enabled/disabled for the account), which is pretty much what this PR is doing, except for the first sync flag

nambatee:
retest this please

PR #2126, opened 2018-03-22
Fix dupe articles in reading list 

michalciurus:
[T189271 Issue](https://phabricator.wikimedia.org/T189271)

Hey there 👋 This PR fixes a bug where duplicate entries in a reading list would present themselves. The bug has been fixed with two simple checks to prevent duplicate entries to be saved to the persistence layer.


WMF Framework/ReadingListsSyncOperation.swift
```diff
@@ -661,7 +661,6 @@ internal class ReadingListsSyncOperation: ReadingListsOperation {
         }
         let group = WMFTaskGroup()
         var remoteEntriesToCreateLocallyByArticleKey: [String: APIReadingListEntry] = [:]
-        var requestedArticleKeys: Set<String> = []
```
michalciurus: During debugging I noticed that `requestedArticleKeys` isn't used anywhere so I took the liberty of removing it to unclutter the code. Please take a look to be double sure 🙏

joewalsh: `requestedArticleKeys` ensures the same article summary request isn't made twice. For example, if this function is importing the same article to two different lists (which presents as two separate entries), the app only needs to get the summary for that article once.

michalciurus: Oh yes, I read the code wrong :+1: Thanks

WMF Framework/ReadingListsSyncOperation.swift
```diff
@@ -754,21 +749,23 @@ internal class ReadingListsSyncOperation: ReadingListsOperation {
...
-                }
-                if entry.updatedDate == nil {
-                    entry.updatedDate = entry.createdDate
-                }
-                entry.list = readingList
-                entry.articleKey = article.key
-                entry.displayTitle = article.displayTitle
-                if article.savedDate == nil {
-                    article.savedDate = entry.createdDate as Date?
+                if !readingList.articleKeys.contains(articleKey) {
```
joewalsh: By the time we get to this function, the dupe check should have already occurred in `createOrUpdate(remoteReadingListEntries: [APIReadingListEntry], for readingListID: Int64? = nil, deleteMissingLocalEntries: Bool = false, inManagedObjectContext moc: NSManagedObjectContext)`. So I'd suggest keeping this function as it was before and seeing if the bug is one level up in `createOrUpdate`

WMF Framework/ReadingListsSyncOperation.swift
joewalsh: No problem, I wrote it and still had to do a double take to ensure it was still used. Open to suggestions for improving readability.

WMF Framework/ReadingListsSyncOperation.swift
```diff
@@ -892,9 +893,12 @@ internal class ReadingListsSyncOperation: ReadingListsOperation {
...
         for (readingListID, readingListEntriesByKey) in remoteReadingListEntriesByReadingListID {
             try autoreleasepool {
-                let localReadingListEntryFetch: NSFetchRequest<ReadingListEntry> = ReadingListEntry.fetchRequest()
-                localReadingListEntryFetch.predicate = NSPredicate(format: "list.readingListID == %@ && isDeletedLocally != YES", NSNumber(value: readingListID)) // this is != YES instead of == NO to match NULL values as well
-                let localReadingListEntries = try moc.fetch(localReadingListEntryFetch)
+
+                let localReadingListsFetch: NSFetchRequest<ReadingList> = ReadingList.fetchRequest()
+                localReadingListsFetch.predicate = NSPredicate(format: "readingListID == %@", NSNumber(value: readingListID))
+                let localReadingLists = try moc.fetch(localReadingListsFetch)
+                let localReadingListEntries = localReadingLists.first?.entries?.filter { !$0.isDeletedLocally } ?? []
```
michalciurus: @joewalsh You were right, the problem was much more subtle. 
It was so subtle that I still have no idea why it occured.

Apparently fetching `ReadingListEntry` by `list` relationship:

`localReadingListEntryFetch.predicate = NSPredicate(format: "list.readingListID == %@ && isDeletedLocally != YES", NSNumber(value: readingListID))`

Doesn't bring expected results. Specifically: the first fetch brings `0` entries while there definitely are entries and:

`localReadingListsFetch.predicate = NSPredicate(format: "readingListID == %@", NSNumber(value: readingListID))`

Does indeed work and fetches the expected entries. It's weird as these queries should have the same result. I'll try to debug some more to find out what's the reason behind it.

michalciurus: This is very puzzling. I've spent another hour debugging this and the result is the same: `list.readingListID == %@` sometimes returns `0` results even though clearly there should be results. It doesn't seem to do with faulting too.

The new query fixes the issue.

nambatee:
jenkins retest this please

WMF Framework/ReadingListsSyncOperation.swift
nambatee: If we don't include `&& isDeletedLocally != YES` in the predicate, lists previously deleted locally will show up. You can test it by logging in, adding a couple of articles to list Foo, logging out (while keeping articles on your device), creating another list Bar (can be empty) && deleting it, logging in and enabling sync. We'll see both Foo and Bar, and we only care about Foo since Bar was deleted.

PR #2128, opened 2018-03-22
[Bug] As a user who has exceeded my limit of unsorted articles, I would still like to be able to sort my saved articles by recency even if they aren't synced

nambatee:
https://phabricator.wikimedia.org/T190186


montehurd:
michalciurus

nambatee:
retest this please

PR #2129, opened 2018-03-22
As a user of reading list syncing, I would like a way to know that articles are downloaded to my device

nambatee:
https://phabricator.wikimedia.org/T187968


montehurd:
retest this please

WMF Framework/AlertType.swift
```diff
@@ -0,0 +1,21 @@
+public enum AlertType {
```
montehurd: Do you think it would be ok to name this something like `ReadingListAlertType`? Just wondering about confusion with alerts with types elsewhere in the future...

WMF Framework/AlertType.swift
```diff
@@ -0,0 +1,21 @@
...
+    case listLimitExceeded(limit: Int)
+    case entryLimitExceeded(limit: Int)
+    case genericNotSynced
+    case downloading
+}
+
+extension AlertType: Equatable {
+    public static func ==(lhs: AlertType, rhs: AlertType) -> Bool {
+        switch (lhs, rhs) {
```
montehurd: Swift is so cool :)

WMF Framework/AlertType.swift
```diff
@@ -0,0 +1,21 @@
...
+        case let (.listLimitExceeded(a), .listLimitExceeded(b)),
+             let (.entryLimitExceeded(a), .entryLimitExceeded(b)):
+            return a == b
```
montehurd: For equality do we care if a == b or is it enough to know both are of same type?

nambatee: We have to compare associated values, it's not enough to compare types

nambatee:
jenkins retest this please

WMF Framework/AlertType.swift
nambatee: @montehurd feel free to post a solution without Equatable 👍

montehurd: You are right on 👍 about equatable and associated values: 
https://andybargh.com/swift-enumerations/#Boolean_Comparison_of_Associated_Values_Enumerations

Given that it sounded like we want this equality check to be associated value agnostic, it looks like it can be simplify it to this:

     extension ReadingListAlertType: Equatable {
          public static func ==(lhs: ReadingListAlertType, rhs: ReadingListAlertType) -> Bool {
               switch (lhs, rhs) {
                    case  (.listLimitExceeded, .listLimitExceeded),
                          (.entryLimitExceeded, .entryLimitExceeded),
                          (.genericNotSynced, .genericNotSynced),
                          (.downloading, .downloading):
                         return true
                    default:
                         return false
               }
          }
     }

Being able to do multi-argument switches still feels like magic :)

PR #2112, opened 2018-03-15
Sync limiting behaviors, edge cases and alerts

nambatee:
https://phabricator.wikimedia.org/T187400

### Reading lists per account limit

- [x] _You have reached the limit of 100 reading lists per account_ standard error when user attempts to create a 101st reading list
- [x] _You have exceeded the limit of 100 reading lists per account. This reading list and the articles saved to it will not be synced, please decrease your number of lists to 100 to resume syncing of this list._ in-line alert on individual reading list when due to syncing error user now has 101 lists
- [x] _List not synced, limit exceeded_ on Reading list view, list item when due to syncing error user now has 101 lists

### Articles per reading list limit
- [x] _Article cannot be added to this list. You have reached the limit of 5000 articles per reading list for list name_ when user attempts to add 5001st article to a list
- [x] _You have exceeded the limit of 5000 articles per account Please decrease your number of articles in this list to 5000 resume syncing of all articles in this list._ on individual reading list when due to syncing error user now has 5001 articles in a list
- [x] _Some articles not synced, limit exceeded_ on Reading list view, list item when due to syncing error user now has 5001 articles in a list
- [x] _Not synced_ on All articles view, list item when due to syncing error user now has 5001 articles in a list

### Reading lists per account limit & articles per reading list limit
- [x] _Article cannot be added to this list. You have reached the limit of 5000 articles per reading list for list name <BR >You have reached the limit of 100 reading lists per account_ standard error when user attempts to add 5001st article and user has hit the 100 reading list limit

### Other TODOs

- [x] Fix tags sizing and collapsing to accommodate for _Not synced_ alert label on All articles view
- [x] Ensure that the extended view in the reading list detail view controller is not stretched if there is no alert
- [x] Ensure that the alert in the reading list detail extended view disappears ~~nicely~~ if user takes action in the detail view
- [x] Ensure that the reading list detail extended view looks ok on landscape, especially on iPhone X
- [x] Theme and avoid blended layers where possible

PR #2130, opened 2018-03-23
User education for reading lists and account creation: Alerts

nambatee:
https://phabricator.wikimedia.org/T187403

- [x] You have reached the limit of 100 reading lists per account #2112
- [x] You do not have enough space on your device to save this article
- [x] Syncing will resume when internet connection is available
- [x] Your list ‘Saved’ has been renamed ‘Saved_[user created]’
- [x] Article cannot be added to this list. You have reached the limit of 5000 articles per list for list name #2112
- [x] n articles and n reading lists synced from your account

_There is a limit of 5000 articles per reading list. Existing lists with more than this limit have been split into multiple lists_ alert will be implemented in https://phabricator.wikimedia.org/T187382

Questions:
1. Should we label the articles that weren't saved? https://phabricator.wikimedia.org/T187403#


nambatee:
jenkins retest this please

nambatee:
jenkins retest this please

nambatee:
jenkins retest this please

nambatee:
retest this please

Wikipedia/Code/MWKImage.m
```diff
@@ -156,10 +156,6 @@ - (NSString *)fileNameNoSizePrefix {
 
 #pragma mark - Import / Export
 
-- (void)save {
```
coreyfloyd: Just curious, how is this handled now? I couldn't find where images are saved now in the diff

coreyfloyd: MWKImages I should say…

Wikipedia/Code/WMFArticleFetcher.m
```diff
@@ -99,13 +99,25 @@ - (nullable NSURLSessionTask *)fetchArticleForURL:(NSURL *)articleURL
...
     if (![SessionSingleton sharedInstance].zeroConfigurationManager.isZeroRated) {
         useDeskTopURL = YES;
     }
 
     NSURL *url = useDeskTopURL ? [NSURL wmf_desktopAPIURLForURL:articleURL] : [NSURL wmf_mobileAPIURLForURL:articleURL];
+    
+    __block WMFErrorHandler saveToDiskFailure = nil;
+    if (saveToDisk) {
+        saveToDiskFailure = ^(NSError *error) {
+            if (error.domain == NSCocoaErrorDomain && error.code == NSFileWriteOutOfSpaceError) {
```
coreyfloyd: Does this error stop other disk save operations from being attempted? I'm wondering if the user will just see like 100 of these alerts in a row.

coreyfloyd: Also, does the app keep trying to download articles/images even though saves are failing? IOW will the app keep trying to re-download the same articles over an over again only to have saves fail?

We are getting into edge cases here, but something to think about.

nambatee: Yup, the way it's implemented now, both are true. The error should be displayed only when user attempts to save via the UI so I should modify this PR and pass that failure block only when that happens (right now, I'm passing it in the wrong place and I end up with an error anytime save happens, regardless of whether it was initiated by user action). I should check with Carolyn how we want to handle the 'your account has more articles than your device can handle' case, maybe that should add another label on the article cell rather than an alert

montehurd:
lgtm

@coreyfloyd did you have any other questions?

PR #2131, opened 2018-03-24
Revert "Saved progress indicators "

nambatee:
Reverts wikimedia/wikipedia-ios#2087

`'*** Collection <__NSArrayM: 0x60000084a260> was mutated while being enumerated.'` crash still occurs, I think we should revert or provide a fix asap


montehurd:
Fix here #2132

montehurd:
@nambatee Closing since the fix at #2132 seems to be working. Thanks again for catching this!

PR #2132, opened 2018-03-24
Fix iPhone X crash on Saved icon tap.

montehurd:
Followup to #2087


Wikipedia/Code/SavedViewController.swift
```diff
@@ -154,7 +154,9 @@ class SavedViewController: ViewController {
         navigationBar.addExtendedNavigationBarView(extendedNavBarView)
         navigationBar.addUnderNavigationBarView(underBarView)
         navigationBar.isBackVisible = false
-        
+
+        wmf_add(childController:savedProgressViewController, andConstrainToEdgesOfContainerView: progressContainerView)
```
montehurd: Diving into why exactly this works if it's before `currentView` but not after. Should be ok to roll with this fix in the mean time.

PR #2133, opened 2018-03-24
Make saved progress background view extend to screen edge on iPhoneX.

montehurd:
Follow-up for https://phabricator.wikimedia.org/T187823

Should merge #2132 first since it fixes the iPhoneX save button tap crash.

<strike>I'm pinging Carolyn before removing the question tag to double-check this is wanted.</strike> (She said it's cool)

Before:
screen shot 2018-03-23 at 7 46 29 pm

After:
screen shot 2018-03-23 at 7 39 18 pm

### DIFF
--- a/Wikipedia/Code/Saved.storyboard
+++ b/Wikipedia/Code/Saved.storyboard
@@ -30,13 +30,13 @@
                             </containerView>
                         </subviews>
                         <constraints>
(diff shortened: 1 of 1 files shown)


carolynlimadeo:
@montehurd the 'After' looks great to me!

PR #2134, opened 2018-03-24
Fix leaking view controllers

michalciurus:
This PR fixes leaking view controllers that I managed to find:

`AppearanceSettingsViewController`
`NotificationSettingsViewController`
`ReadingListHintController`


Wikipedia/Code/AppearanceSettingsViewController.swift
```diff
@@ -72,8 +72,14 @@ open class AppearanceSettingsViewController: UIViewController, UITableViewDataSo
...
         
+        let themeCheckmarkItem: (Theme) -> (AppearanceSettingsCheckmarkItem) = { [weak self] theme in
+            return AppearanceSettingsCheckmarkItem(title: theme.displayName, theme: theme) {
+                self?.userDidSelect(theme: theme)
+            }
+        }
+
         let readingThemesSection =
-            AppearanceSettingsSection(headerTitle: WMFLocalizedString("appearance-settings-reading-themes", value: "Reading themes", comment: "Title of the the Reading themes section in Appearance settings"), footerText: nil, items: [AppearanceSettingsCheckmarkItem(title: Theme.light.displayName, theme: Theme.light, checkmarkAction: {self.userDidSelect(theme: Theme.light)}), AppearanceSettingsCheckmarkItem(title: Theme.sepia.displayName, theme: Theme.sepia, checkmarkAction: {self.userDidSelect(theme: Theme.sepia)}), AppearanceSettingsCheckmarkItem(title: Theme.dark.displayName, theme: Theme.dark, checkmarkAction: {self.userDidSelect(theme: Theme.dark)}), AppearanceSettingsCheckmarkItem(title: Theme.black.displayName, theme: Theme.black, checkmarkAction: {self.userDidSelect(theme: Theme.black)})])
```
michalciurus: This one was *very* tricky to find 😅

Wikipedia/Code/AppearanceSettingsViewController.swift
```diff
@@ -72,8 +72,14 @@ open class AppearanceSettingsViewController: UIViewController, UITableViewDataSo
     
     func sectionsForAppearanceSettings() -> [AppearanceSettingsSection] {
         
+        func checkmarkItemFor(theme: Theme) -> (AppearanceSettingsCheckmarkItem) {
```
montehurd: Not a blocker for merge, but this method signature would also be acceptable:

> func checkmarkItem(for theme: Theme)

montehurd:
Great catches!

Wikipedia/Code/AppearanceSettingsViewController.swift
michalciurus: :+1: good catch, that's more swifty ;)

PR #2137, opened 2018-03-26
More CSS updates

mdholloway:
Summary of commits:

The first updates the bundled CSS from production;

The second switches CSS loading to MW-Vagrant and pulls in the new styles about to be deployed to hide the (listen) parenthetical in IPA spans, and makes some other changes to the request params for correctness (with no current effect);

The third removes styles that are no longer needed;

The fourth removes the bundled legacy enwiki.less styles and adds support for loading the current MediaWiki:Mobile.css from ResourceLoader as mobile.site;

The fifth cleans up by switching default CSS loading back to production.

Note that the third and fourth are based on MobileApp extension patches currently awaiting review.  Reviews on those in time to make this week's MediaWiki train deployment (i.e., by ~18:00 UTC Tuesday) are very welcome!


montehurd:
@mdholloway coolness! i'll kick the tires...

montehurd:
Hey I noticed one small missing piece - the changes made to the HTML files would need to be made to the HTML files in `www` too (when we run `grunt` it overwrites the `assets` html files with those in `www`). 

Otherwise looks good - I tested it for a while and seemed solid :)

mdholloway:
Thanks, @montehurd!  Updated and rebased.

montehurd:
@mdholloway looks like https://gerrit.wikimedia.org/r/#/c/421947/ got merged. Did the other dependencies get deployed already?

mdholloway:
@montehurd https://gerrit.wikimedia.org/r/#/c/419843/ and https://gerrit.wikimedia.org/r/#/c/421947/ are merged and rolling out to production this week, https://gerrit.wikimedia.org/r/#/c/420375/ (dependency for cda5d94) is still waiting on review and won't go out until next week at the earliest.

montehurd:
@mdholloway thanks! i just +1'ed https://gerrit.wikimedia.org/r/#/c/420375/ :)

montehurd:
@mdholloway If it's ok I'd probably lean towards waiting to merge this PR until https://gerrit.wikimedia.org/r/#/c/420375/ goes live.

mdholloway:
@montehurd  That sounds fine.  Or I could split it up if you wanted the IPA change sooner (i.e., this Thursday).  Either way works for me!

montehurd:
Added the "Do not merge" tag to remind us to wait until https://gerrit.wikimedia.org/r/#/c/420375/ is live.

mdholloway:
I took out the commit with the outstanding dependency since it will probably need some changes in light of the ongoing discussion on [T190091](https://phabricator.wikimedia.org/T190091) anyway.  I'll open a new PR for that.  The rest is rebased and should be ready to merge.

montehurd:
@mdholloway heya! i pulled this PR locally then ran 'make css' to fetch the latest using the new URL - the diff was bit larger than i expected (you can see it here: https://github.com/montehurd/apps-ios-wikipedia/commit/465cb86b2b0ad8221d197a2c2a3df14d4db404d1 ). just double-checking that it's expected :)

mdholloway:
@montehurd Howdy!  I believe the sizable diff is expected because we've removed the old enwiki Mobile.css (enwiki.less) from the CSS provided by the MobileApp extension.  This actually probably shouldn't be merged until we have a strategy for re-incorporating site specific CSS from the newly deployed REST CSS endpoints (e.g., https://en.wikipedia.org/api/rest_v1/data/css/mobile/site).  On Android this is actually proving to be a little trickier than anticipated; see https://gerrit.wikimedia.org/r/#/c/428692/ for discussion.  I'm not sure if the fine details of page loading are quite the same on iOS but I imagine you might run into similar issues.  Any ideas on how best to proceed?

montehurd:
@mdholloway I think we're a little ways out from tackling the site specific CSS bits on iOS. You mentioned this PR should prob wait on that... would we be missing some important bits if we merged this in the mean time?

mdholloway:
Just to be clear, when you say "new URL," do you mean `$(CSS_PREFIX)/load.php?only=styles&target=mobile&skin=minerva&modules=skins.minerva.base.reset|skins.minerva.content.styles|ext.math.styles|ext.pygments|mobile.app`, or `/api/rest_v1/data/css/mobile/base`?  The latter is meant for a Parsoid-consuming client, so you won't want that until after making the switch.

As I understand it from digging through the commit history, the enwiki styles were bundled mainly for dealing with horizontal lists (`.hlist`).  I imagine you would introduce some visual regressions by just omitting them, but not many.  If you've been testing with the bundle from your latest updates and not noticed any issues, then that's a good sign.  

If there are issues, the easiest way to resolve them for now would be to append `|mobile.site.styles`, which provides the current mobile site CSS, to `CSS_ORIGIN` in your Makefile and rebuild the style bundle.  I see you are loading from enwiki, so that would approximate the status quo.

This should be fine to merge either way.

montehurd:
oh, by new URL i meant the former:
`$(CSS_PREFIX)/load.php?only=styles&target=mobile&skin=minerva&modules=skins.minerva.base.reset|skins.minerva.content.styles|ext.math.styles|ext.pygments|mobile.app`

montehurd:
Test comment 2.

montehurd:
I added `Do not merge` because Michael's gonna re-run `make css` shortly (the fetched files have diverged a bit from the ones in this PR since it was opened).

montehurd:
@mdholloway 

Hey I noticed something unexpected in the diff - there's a bunch of new `.mw-parser-output` classes near the bottom of each CSS file. Are these wanted?

mdholloway:
Yes, I believe this is the relevant change:

https://en.wikipedia.org/w/index.php?title=MediaWiki%3AMobile.css&type=revision&diff=829072432&oldid=829071767

Seems like a classic case of consumers external to MediaWiki not being considered.  On the one hand, I don't think the scoping to .mw-parser-output is really appropriate, but on the other, this keeps coming up.  

In Parsoid HTML, the `mw-parser-output` class is actually added to the `<body>` element in a post-processing step presumably for compatibility's sake.  Maybe the apps would be better off adding it to their bundled HTML files as well.

mdholloway:
Also, after looking more closely at the Android app, it appears to be omitting the site-specific styles completely at this point, with no fallout at least as far as I've heard.

I'll take `mobile.site.styles` back out and rerun `make css` so you can give that a spin when you get a chance.

montehurd:
@mdholloway 

> In Parsoid HTML, the mw-parser-output class is actually added to the <body> element in a post-processing step presumably for compatibility's sake. Maybe the apps would be better off adding it to their bundled HTML files as well.

This is interesting! If we add it on iOS as they [did on Android](https://github.com/wikimedia/apps-android-wikipedia/commit/b2916cb6d1e7a2309cb7240dae75544bdc14f4a5#diff-d280d067bb2d871e9e6381350f3e769aR12), would we need to add [mediawiki.skinning.content.parsoid](https://github.com/wikimedia/apps-android-wikipedia/commit/b2916cb6d1e7a2309cb7240dae75544bdc14f4a5#diff-05422bba61ddc96a04e7e3a840d2f8b8) too as mentioned in the commit message for that PR?

mdholloway:
@montehurd  Nope.  `mediawiki.skinning.content.parsoid` is just for providing extra features to clients, like auto-incrementing external link counters, that the MediaWiki PHP parser includes but Parsoid doesn't, in order to provide feature parity (mostly for VE).  You'll want it when you switch to using Parsoid HTML, but it's of no interest to clients consuming HTML from Parser.php.

montehurd:
I'll do separate PR to use `<body class="stable mw-parser-output">`.

montehurd:
Confirmed in chat we're good to go on this PR.

PR #2140, opened 2018-03-27
Use NSSortDescriptor init supporting Swift 4 KeyPaths.

montehurd:
Noticed there's a NSSortDescriptor convenience init supporting Swift 4 key paths. (safer than string literals)


WMF Framework/EventLoggingService.swift
```diff
@@ -283,7 +283,7 @@ public class EventLoggingService : NSObject, URLSessionDelegate {
         let moc = self.managedObjectContext
         moc.perform {
             let fetch: NSFetchRequest<EventRecord> = EventRecord.fetchRequest()
-            fetch.sortDescriptors = [NSSortDescriptor.init(key: "recorded", ascending: true)]
+            fetch.sortDescriptors = [NSSortDescriptor.init(keyPath: \EventRecord.recorded, ascending: true)]
```
nambatee: `.init` not necessary

montehurd: Fixed!

PR #2143, opened 2018-03-28
Don't show saved progress icon badge or progress view until there's been at least 2 seconds of progress

montehurd:
Follow-up to: https://phabricator.wikimedia.org/T187823

- reduces saved progress badge flicker if single articles are saved
- also safe in cases where extra progress items are added during initial progression


montehurd:
@carolynlimadeo I started with 2 seconds. If you want to check this branch out and kick the tires it should do what we wanted :)

Wikipedia/Code/SavedTabBarItemProgressBadgeManager.swift
```diff
@@ -16,22 +16,33 @@
         // Observe any time a new Progress object is set. (NSProgress are not re-usable so you need to reset it if you're tracking a new progression)
         progressObjectWasSetObservation = ProgressContainer.shared.observe(\ProgressContainer.articleFetcherProgress, options: [.new, .initial]) { [weak self] (progressContainer, change) in
             self?.progressFractionCompletedObservation?.invalidate()
+            
+            var exceededMinSyncInProgressDuration = false
```
coreyfloyd: @montehurd Is there a way to reuse this code rather than write it once for each component? 

Since both the badge and the progress bar follow the same rule, it makes sense if they are updated by the same code. It is now possible for them to get out of sync if a developer doesn't realize they have to update this code in two places.

montehurd:
jenkins retest this please

PR #2136, opened 2018-03-26
Refactor EditSummaryViewController to Swift

michalciurus:
[T164633 Issue](https://phabricator.wikimedia.org/T164633)

This PR refactors one of the classes mentioned in the issue (`EditSummaryViewController.m`) to Swift.

PR #2146, opened 2018-03-29
Extract grouping code from PlacesViewController

michalciurus:
`PlacesViewController.swift` is now 2,8k LOC. My rule of thumb is that a VC should be about 1k LOC maximum. I've picked some low hanging fruit: the grouping algorithm and extracted it to a separate class. It's still not very pretty but it's a small step.

In a perfect scenario the algorithm would be completely abstracted from the view layer (with a presenter pattern for example) but it's a very challenging task given the complexity of the algorithm.

`PlacesViewController` is 2,500 LOC after extracting the grouping algorithm.

It's just a suggestion/experiment PR. Feel free to close it if you don't like it or if you have suggestions on how to improve the refactor. Also, obviously it's very low priority so review it whenever you have extra time on your hands.


Wikipedia/Code/PlacesViewController.swift
```diff
@@ -52,13 +52,14 @@ class PlacesViewController: PreviewingViewController, UISearchBarDelegate, Artic
...
     @objc public var dataStore: MWKDataStore!
+    var articleGrouper: ArticleLocationGrouper!
 
     fileprivate let locationSearchFetcher = WMFLocationSearchFetcher()
     fileprivate let searchFetcher = WMFSearchFetcher()
     fileprivate let wikidataFetcher = WikidataFetcher()
     fileprivate let locationManager = WMFLocationManager.fine()
-    fileprivate let animationDuration = 0.6
-    fileprivate let animationScale = CGFloat(0.6)
+    internal let animationDuration = 0.6
```
michalciurus: I suggest declaring a universal way of declaring Swift constants in the project.
In my [other PR](https://github.com/wikimedia/wikipedia-ios/pull/2136/files#diff-573fdbd6a3e8d237b1cd2e577595fed6R4) I suggested [this](clusteringViewContext.animationScale).

montehurd:
@michalciurus 
Looks like some of our new bits caused a merge conflict - could you check?

joewalsh:
Closing this for now - there might be bigger changes to Places coming with https://phabricator.wikimedia.org/T193312 which would be a more advantageous time to refactor

PR #2156, opened 2018-04-03
For existing users with over 5000 articles per list, split reading lists with more articles than the limit per list into multiple lists

nambatee:
https://phabricator.wikimedia.org/T187382

Questions

1. If we're about to split Foo into "Foo" and "Foo_1" but user already has their own list named "Foo_1", what do we name the slit list?
2. Do we split the default list as well?
3. Do we split lists every time, or only on the first sync?
https://phabricator.wikimedia.org/T187382#
WIP til questions answered.


WMF Framework/ReadingListsSyncOperation.swift
```diff
@@ -400,6 +400,46 @@ internal class ReadingListsSyncOperation: ReadingListsOperation {
         }
     }
     
+    private func splitReadingList(_ readingList: ReadingList, intoReadingListsOfSize size: Int, in moc: NSManagedObjectContext) throws -> [ReadingList] {
```
michalciurus: Insignificant detail but: `split(readingList...` seems more "Swifty" 😅?

PR #2160, opened 2018-04-03
[Design bug] 'Move to list' view should have the title 'Move [n] article(s) to reading list' not 'Add [n] article(s) to reading list'

nambatee:
https://phabricator.wikimedia.org/T188566


nambatee:
@montehurd looks like other references to `CommonStrings.articleCountFormat` are ok, no more composed strings

PR #2162, opened 2018-04-03
Hierarchy issues on individual reading list views

nambatee:
https://phabricator.wikimedia.org/T191254


Wikipedia/Code/ReadingListDetailExtendedViewController.swift
```diff
@@ -141,7 +142,21 @@ class ReadingListDetailExtendedViewController: UIViewController {
     
     private var isAlertViewHidden: Bool = true {
         didSet {
-            collapseAlert(isAlertViewHidden)
+            if isAlertViewHidden {
+                alertView?.removeFromSuperview()
```
montehurd: It looks like `alertView` would need to be `strong` to keep it around after `removeFromSuperview`. I toggled `isAlertViewHidden` in a `didReceiveMemoryWarning` to test and that seemed to be the case. After I removed `weak` it started working. Could you take a peek?

nambatee: good catch, thanks! 👍 should be fixed

PR #2169, opened 2018-04-04
Hierarchy issues on individual reading list views followup

nambatee:
https://phabricator.wikimedia.org/T191254#

- [x] make the HR between the search bar and the top list item the same thickness as the list items HRs
- [x] have the Title of the reading list move into the header after the user has scrolled past the large title (similar to the header in iOS settings)

### DIFF
--- a/WMF Framework/NavigationBar.swift
+++ b/WMF Framework/NavigationBar.swift
@@ -7,6 +7,8 @@ public class NavigationBar: SetupView {
     fileprivate let shadow: UIView = UIView()
     fileprivate let progressView: UIProgressView = UIProgressView()
     fileprivate let backgroundView: UIView = UIView()
+    public var extendedViewPercentHiddenForShowingTitle: CGFloat?
+    public var title: String?
     
     /// back button presses will be forwarded to this nav controller
     @objc public weak var delegate: UIViewController? {
@@ -170,6 +172,11 @@ public class NavigationBar: SetupView {
         _extendedViewPercentHidden = extendedViewPercentHidden
         setNeedsLayout()
         //print("nb: \(navigationBarPercentHidden) ev: \(extendedViewPercentHidden)")
+        if let extendedViewPercentHiddenForShowingTitle = self.extendedViewPercentHiddenForShowingTitle {
+            UIView.animate(withDuration: 0.2) {
+                self.delegate?.title = extendedViewPercentHidden >= extendedViewPercentHiddenForShowingTitle ? self.title : nil
+            }
+        }
         let changes = {
             self.layoutIfNeeded()
             additionalAnimations?()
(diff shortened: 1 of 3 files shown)


WMF Framework/NavigationBar.swift
```diff
@@ -170,6 +172,11 @@ public class NavigationBar: SetupView {
         _extendedViewPercentHidden = extendedViewPercentHidden
         setNeedsLayout()
         //print("nb: \(navigationBarPercentHidden) ev: \(extendedViewPercentHidden)")
+        if let extendedViewPercentHiddenForShowingTitle = self.extendedViewPercentHiddenForShowingTitle {
+            UIView.animate(withDuration: 0.2) {
+                self.delegate?.title = extendedViewPercentHidden >= extendedViewPercentHiddenForShowingTitle ? self.title : nil
```
montehurd: Could this go in the `changes` block below (which gets animated with same duration)? Would avoid potential situation where the two animation blocks could fight a bit...

nambatee: the title animation wouldn't be animated then, I can change it to use completion?

        let applyChanges = {
            let changes = {
                self.layoutIfNeeded()
                additionalAnimations?()
            }
            if animated {
                UIView.animate(withDuration: 0.2, animations: changes)
            } else {
                changes()
            }
        }
        if let extendedViewPercentHiddenForShowingTitle = self.extendedViewPercentHiddenForShowingTitle {
            UIView.animate(withDuration: 0.2, animations: {
                self.delegate?.title = extendedViewPercentHidden >= extendedViewPercentHiddenForShowingTitle ? self.title : nil
            }, completion: { (_) in
                applyChanges()
            })
        } else {
            applyChanges()
        }

montehurd: Ya that looks like that would keep the 2 animations from trying to happen simultaneously.

Can you check if the `self.layoutIfNeeded()` call in the `changes` block causes the `navigationBarPercentHidden` and `extendedViewPercentHidden` changes to not animate? At first glance it looks like only `additionalAnimations` changes would animate as a result...

montehurd: Ah sorry - in the case animated is true, it looks like the primary things being animated were the changes to `navigationBarPercentHidden` and `extendedViewPercentHidden`, and then there are `additionalAnimations`. 

But the call to `self.layoutIfNeeded()` looks like it could force the layout of the `navigationBarPercentHidden` and `extendedViewPercentHidden` changes instead of animating them.

nambatee: so are you saying that calling `self.layoutIfNeeded()` inside `UIView.animate` will not animate the layout of the `navigationBarPercentHidden` and `extendedViewPercentHidden` changes? 🤔

montehurd: ya i thought `layoutIfNeeded` forced immediate layout

montehurd: (Based on slack conversation should be good to go.)

PR #2172, opened 2018-04-04
Followup: As a user of reading lists, I would like to have a place in Settings to manage my reading list preferences

nambatee:
- [x] use temporary cache size to calculate how much space will be freed up on saved articles removal
- [x] clear image and articles saved on disk cache on saved articles removal


Wikipedia/Code/MWKDataStore.m
```diff
@@ -1561,6 +1561,16 @@ - (void)clearMemoryCache {
     [self.articlePreviewCache removeAllObjects];
 }
 
+- (void)clearImageAndArticlesSavedOnDiskCache {
```
montehurd: Could we name this a bit differently to convey that it only deletes cached article and image data for `unsaved` articles?
Maybe something like `clearCachesForUnsavedArticles`?

montehurd: (updated after chat confimation)

PR #2183, opened 2018-04-06
Pull to refresh fix

montehurd:
https://phabricator.wikimedia.org/T189171

Fixes the following bug:

On app clean start:

On iOS 10.3.1:
- tap "Saved" tab bar icon
- do pull to refresh on "All articles"
- tap "Reading lists"
- do pull to refresh
- switch back to "All articles"
- attempt pull to refresh again, which won't work

On iOS 11.2:
- tap "Saved" tab bar icon
- do pull to refresh on "All articles"
- tap "Reading lists"
- do pull to refresh, which won't work

### DIFF
--- a/Wikipedia/Code/SavedViewController.swift
+++ b/Wikipedia/Code/SavedViewController.swift
@@ -95,7 +95,6 @@ class SavedViewController: ViewController {
                 isSearchBarHidden = true
                 activeEditableCollection = readingListsViewController
             }
-            scrollView?.refreshControl = pullToRefresh
         }
     }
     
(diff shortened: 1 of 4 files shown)


Wikipedia/Code/SavedViewController.swift
```diff
@@ -174,7 +173,8 @@ class SavedViewController: ViewController {
         edgesForExtendedLayout = .all
         
         pullToRefresh.addTarget(self, action: #selector(pulledToRefresh), for: .valueChanged)
-        
+        scrollView?.refreshControl = pullToRefresh
+
```
nambatee: @montehurd if we set `refreshControl` of this `scrollView`, it'll only work in the All articles view because when the view loads, the `scrollView` of the `SavedViewController` is set to that of `SavedArticlesViewController`'s (see line 85). I missed that we can use the refresh control inherited from `ColumnarCollectionViewController`, we were already doing that in `SavedArticlesViewController` and `ReadingListsViewController` (see the `isRefreshControlEnabled` flag)

PR #2187, opened 2018-04-09
[BUG] Logging out causes sync error

nambatee:
https://phabricator.wikimedia.org/T191748

Also was causing & fixes https://phabricator.wikimedia.org/T190194 and https://phabricator.wikimedia.org/T189172

### DIFF
--- a/WMF Framework/ReadingListsSyncOperation.swift
+++ b/WMF Framework/ReadingListsSyncOperation.swift
@@ -153,9 +153,20 @@ internal class ReadingListsSyncOperation: ReadingListsOperation {
             self.finish()
         }
         
-        guard readingListsController.isLoggedIn else {
+        guard let authenticationDelegate = readingListsController.authenticationDelegate, authenticationDelegate.isUserLoggedInLocally() else {
             try localSyncOnly()
             return
+        }
+        
+        if !authenticationDelegate.isUserLoggedInRemotely() {
+            taskGroup.enter()
+            authenticationDelegate.attemptLogin({
+                taskGroup.leave()
+            }, failure: {
+                try? localSyncOnly()
+                return
+            })
+            taskGroup.wait()
         }
         
         // local only sync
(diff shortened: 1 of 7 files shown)


coreyfloyd:
Wow, this is fantastic that you found the cause of all these issues!

WMF Framework/ReadingListsSyncOperation.swift
```diff
@@ -147,7 +147,12 @@ internal class ReadingListsSyncOperation: ReadingListsOperation {
             try moc.save()
         }
         
-
+        guard readingListsController.isLoggedIn else {
```
coreyfloyd: 💥

Wikipedia/Code/WMFAppViewController.m
```diff
@@ -784,6 +784,8 @@ - (void)pauseApp {
     self.settingsViewController = nil;
 
     [self.dataStore.feedContentController stopContentSources];
+    
+    [[WMFAuthenticationManager sharedInstance] deleteLoginTokensAndBrowserCookies];
```
coreyfloyd: Why do this?

Just curious what the effect is

coreyfloyd: nm… you removed it entirely

Wikipedia/Code/WMFAuthenticationManager.swift
```diff
@@ -177,21 +177,9 @@ class WMFAuthenticationManager: NSObject {
      *  Logs out any authenticated user and clears out any associated cookies
      */
     @objc public func logout(completion: @escaping () -> Void = {}){
-        self.resetLocalUserLoginSettings()
-        completion()
-    }
-    
-    @objc public func deleteLoginTokensAndBrowserCookies() {
```
coreyfloyd: Why is it ok to kill this method? Does it not need to happen or does it happen in another way now?

nambatee: I moved it back where it was, to `logout`, so there's no need for a separate method anymore

coreyfloyd:
Looks good!

PR #2194, opened 2018-04-10
[Design bug] Missing alerts when turning reading list sync off

nambatee:
https://phabricator.wikimedia.org/T189064

Question: Right now, it seems like we have no way of knowing whether sync was explicitly disabled via user action or **never enabled**. All we know from the server is whether it's currently enabled or disabled for a given account (if it's disabled, we receive `readinglists-db-error-not-set-up` error from the server). We want to present one of the panels ("Sync disabled - Reading list syncing has been disabled for your Wikipedia account on another device. You can turn sync back on by updating your settings.") after sync was disabled for the account via user action. Since we don't know how sync was disabled for the account, we end up showing the panel even if sync was **never enabled**. 🤔

Solution: We can check if sync was ever enabled on device by checking if we sent a `setup` request to the API.


Wikipedia/Code/Panels.swift
```diff
@@ -23,28 +23,58 @@ class AddSavedArticlesToReadingListPanelViewController : ScrollableEducationPane
...
+            subheading = CommonStrings.keepSavedArticlesOnDeviceMessage
+            secondaryButtonTitle = CommonStrings.readingListDoNotKeepSubtitle
+        } else if trigger == .syncDisabled {
+            subheading = CommonStrings.keepSavedArticlesOnDeviceMessage + "\n" + WMFLocalizedString("reading-list-keep-sync-disabled-additional-subtitle", value: "Turning sync off will remove these articles from your account. If you remove them from your device they will not be recoverable by turning sync on again in the future.", comment: "Additional subtitle informing user that turning sync off will remove saved articles from their account.")
+            secondaryButtonTitle = WMFLocalizedString("reading-list-keep-sync-disabled-remove-article-button-title", value: "No, remove articles from device and my Wikipedia account", comment: "Title for button that removes save articles from device and Wikipedia account.")
+        }
+    }
+}
+
+class SyncDisabledPanel: ScrollableEducationPanelViewController {
```
montehurd: Would it be ok to keep following the naming convention I established in this file - so this would instead be named `SyncDisabledPanelViewController`?

Wikipedia/Code/ReadingListsController.swift
```diff
@@ -90,6 +90,7 @@ public enum ReadingListError: Error, Equatable {
 public class ReadingListsController: NSObject {
     @objc public static let readingListsServerDidConfirmSyncIsEnabledForAccountNotification = NSNotification.Name("WMFReadingListsServerDidConfirmSyncIsEnabledForAccount")
     @objc public static let readingListsServerDidConfirmSyncIsEnabledForAccountIsSyncEnabledKey = NSNotification.Name("isSyncEnabledForAccount")
+    @objc public static let readingListsServerDidConfirmSyncIsEnabledForAccountWasSyncEnabledOnDeviceKey = NSNotification.Name("wasSyncEnabledOnDevice")
```
montehurd: It's a little hard to tease these out. Usually we have one "is" or "was" per key and these have "...is...is..." and "is...was" respectively:

`readingListsServerDidConfirmSyncIsEnabledForAccountIsSyncEnabledKey`
`readingListsServerDidConfirmSyncIsEnabledForAccountWasSyncEnabledOnDeviceKey`

... any thoughts on distilling these down a bit?

montehurd: It's fine if these are actually named accurately too - just double-checking :)

nambatee: yeah, they are named accurately but if you can think of better names, I'll be happy to update these, in `readingListsServerDidConfirmSyncIsEnabledForAccountWasSyncEnabledOnDeviceKey`  there's `was` instead of `is` cause we want to know if sync **was** ever enabled on this device, in `readingListsServerDidConfirmSyncIsEnabledForAccountIsSyncEnabledKey ` it's `is`, instead of `was` cause it's supposed to reflect the current sync state for the account. I guess we could re-name `readingListsServerDidConfirmSyncIsEnabledForAccountIsSyncEnabledKey` to `readingListsServerDidConfirmSyncWasEnabledForAccountWasSyncEnabledKey` if you think that'd be better?

nambatee:
jenkins retest this please

Wikipedia/Code/ReadingListsController.swift
montehurd: Ah thanks! Yes, I think your "was...was" idea would make it better 👍

Wikipedia/Code/ScrollableEducationPanelViewController.swift
```diff
@@ -131,7 +131,8 @@ class ScrollableEducationPanelViewController: UIViewController, Themeable {
         
         reset()
         primaryButton.titleLabel?.wmf_configureToAutoAdjustFontSize()
-        secondaryButton.titleLabel?.wmf_configureToAutoAdjustFontSize(numberOfLines: 2)
+        secondaryButton.titleLabel?.numberOfLines = 2
+        secondaryButton?.titleLabel?.adjustsFontSizeToFitWidth = true
```
montehurd: Oh is `adjustsFontSizeToFitWidth` now working if `numberOfLines` is greater than 1? iirc it didn't on iOS 11.2 and before...

nambatee: 11.2
screen shot 2018-04-10 at 8 03 44 pm

10.3
screen shot 2018-04-10 at 8 05 08 pm

montehurd: One quick thing to check... 

In your screenshots I see the bottom button text wrapped to 2 lines, but did it change the font size to do so (ala `adjustsFontSizeToFitWidth`)? 

To check add a little more text to the `No, remove articles...` string such that it would push it to 3 lines and see if the text gets a little smaller.

montehurd: Thanks for checking. That last sentence was hurtful :(

PR #2196, opened 2018-04-10
Add missing strings files.

montehurd:


montehurd:
(Had chat approval from Natalia.)

PR #2205, opened 2018-04-12
Followup: [BUG] "Done" button is disabled when I try to edit details of an empty reading list

nambatee:
This followup is slightly bigger due to change in responsibilities. Before, `CollectionViewEditController` was responsible for managing state transitions for all classes implementing `CollectionViewEditControllerNavigationDelegate`. The (incorrect: https://phabricator.wikimedia.org/T191664#) assumption was that all VCs would want to transition from state A to state B. Now, we're letting VCs decide what state they want to transition to.


Wikipedia/Code/ReadingListDetailViewController.swift
```diff
@@ -548,6 +569,14 @@ extension ReadingListDetailViewController: ReadingListDetailExtendedViewControll
     func extendedViewController(_ extendedViewController: ReadingListDetailExtendedViewController, didBeginEditing textField: UITextField) {
         editController.isTextEditing = true
     }
+    
+    func extendedViewController(_ extendedViewController: ReadingListDetailExtendedViewController, titleTextFieldWillClear textField: UITextField) {
+        navigationItem.rightBarButtonItem?.isEnabled = false
+    }
+    
+    func extendedViewController(_ extendedViewController: ReadingListDetailExtendedViewController, titleTextFieldWillChangeCharacters textField: UITextField, replacementString string: String) {
+        navigationItem.rightBarButtonItem?.isEnabled = textField.text?.count == 1 && string.isEmpty ? false : true
```
montehurd: Do you recall if `isEmpty` accounts for whitespace? If not could we use `!string.wmf_hasNonWhitespaceText` instead of `isEmpty` here so it considers the field empty if the user enters just a bunch of spaces?

nambatee: 👍 fix pushed

PR #2217, opened 2018-04-14
Fix regression caused by #2214: incorrect editing state in reading list detail view

nambatee:


Wikipedia/Code/ReadingListDetailViewController.swift
```diff
@@ -157,7 +157,7 @@ class ReadingListDetailViewController: ColumnarCollectionViewController, Editabl
     // MARK: - Empty state
     
     override func isEmptyDidChange() {
-        guard readingList.countOfEntries == 0 else {
+        guard isEmpty == (readingList.countOfEntries == 0) else {
```
montehurd: Just curious, why `==` vs `||`?

nambatee: you can try https://phabricator.wikimedia.org/T192165 with `||`, also if you swipe to delete in the detail view, then tap `Edit`, the action pane won't open

nambatee: if the collection view goes from an empty state to a nonempty state, `isEmpty || readingList.countOfEntries == 0` (`false || false`)  will be false and we'll never update `editController's` `isCollectionViewEmpty`

montehurd: Oh interesting thank you! I wonder if we should put the contents of that guard into a `let` with a name which conveys what you just explained. If you can think of something that'd be cool. No stress if not.

montehurd: Oh I noticed one more strange related edge case where the empty view doesn't appear:

4 mov

Follow-on is fine if you prefer. Let me know :)

nambatee: Yeah, good point! Let me add that let and fix that bug in this PR :) thanks for checking it out!

PR #2224, opened 2018-04-17
Change reading lists to be sorted case insensitively

nambatee:
https://phabricator.wikimedia.org/T191653


Wikipedia/Code/ReadingListsViewController.swift
```diff
@@ -51,7 +51,7 @@ class ReadingListsViewController: ColumnarCollectionViewController, EditableColl
         }
         
         var sortDescriptors = baseSortDescriptors
-        sortDescriptors.append(NSSortDescriptor(keyPath: \ReadingList.canonicalName, ascending: true))
+        sortDescriptors.append(NSSortDescriptor(key: "canonicalName", ascending: true, selector: #selector(NSString.caseInsensitiveCompare)))
```
montehurd: Can we use `localizedCaseInsensitiveCompare` here? The docs for `caseInsensitiveCompare` says:

> When working with text that’s presented to the user, use the localizedCaseInsensitiveCompare(_:) method instead.

PR #2234, opened 2018-04-18
[BUG] Batch editing view shows on the left for RTL languages

nambatee:
https://phabricator.wikimedia.org/T192314#


montehurd:
seeing some oddness on 10.3.1 - i portrait the last 2 radio buttons are overlapping the text a bit:

screen shot 2018-04-18 at 2 45 24 pm

screen shot 2018-04-18 at 2 45 32 pm

here's iOS 11 for reference:

screen shot 2018-04-18 at 2 48 20 pm

screen shot 2018-04-18 at 2 48 11 pm

montehurd:
^ in the 1st and 3rd images above it looks like the margin on the left of the thumbnail is different on 10 vs 11 - may be a clue as to what the layout is considering the edge of the screen (which if different between OS versions could explain the radio button issue)...

PR #2236, opened 2018-04-18
Fix reading list title not appearing in reading list detail modal view

nambatee:


montehurd:
https://phabricator.wikimedia.org/T192193

montehurd:
Oh sorry the ticket says the header is what we want (vs the title). This part is missing:

screen shot 2018-04-18 at 3 59 21 pm

montehurd:
Per chat with Natalia we're merging this for now and considering a follow-on for showing the header VS the title.

PR #2237, opened 2018-04-18
Fix reading list swipe animating article count label in wrong direction on RTL device lang.

montehurd:
https://phabricator.wikimedia.org/T192510


montehurd:
Natalia's last PR made this one no longer needed.

PR #2240, opened 2018-04-19
[BUG] Batch edit selection indicator is not centered

nambatee:
https://phabricator.wikimedia.org/T192859


montehurd:
One thing to check...

On develop we're making room for the notch:
screen shot 2018-04-23 at 4 16 44 pm


On this PR it looks like it's no longer making the notch affordance:
screen shot 2018-04-23 at 4 14 53 pm

nambatee:
@montehurd on develop, it looks like we're making room for the notch but it's only because the indicator is not centered

not_centered

montehurd:
@nambatee 

iirc for the X it needs to be centered with respect to the safe layout margins, so we'd ignore the part to the left of the orange line below:

39457873-d2595642-4cbe-11e8-8529-ba78caf5f0db

...at least that's my reading of the HIG image...
screen shot 2018-04-30 at 6 48 33 pm

montehurd:
Not a big deal though. Would it be ok to just add a design review note for Carolyn to check landscape on X?

nambatee:
added WIP til questions answered https://phabricator.wikimedia.org/T192859#

PR #2151, opened 2018-03-30
Fix Xcode 9.3 / Swift 4.1 warnings

montehurd:
- [x] fix build errors and warnings
- [x] fix tests build errors and warnings
- [x] fix tests (a couple FB snapshot tests are failing) 

Question for Josh:
<strike>5.8 or 5.8.1?</strike>
Update: we're waiting until 5.8.2 to [update buildmini to use Xcode 9.3](https://phabricator.wikimedia.org/T192602) and to merge this PR.

(the tests are passing when ran using Xcode 9.3 / Swift 4.1)

PR #2242, opened 2018-04-20
Fix self before super init error

tiagomartinho:
Would it make sense to merge this while this [PR](https://github.com/wikimedia/wikipedia-ios/pull/2151) is being accepted?

So it's possible to use Swift 4.1 and Xcode 9.3 to develop in the meanwhile.


joewalsh:
Hi @tiagomartinho - we should be updating to Xcode 9.3 after 5.8.1 is released. It looks like this older PR addresses the same issue: https://github.com/wikimedia/wikipedia-ios/pull/2151/ 

Could you confirm the fix in the other PR looks good and we can close this PR? Thanks!

tiagomartinho:
Yes @joewalsh, this was meant as a temporary patch while the whole migration was done. 

I will close the PR. Thanks!

PR #2255, opened 2018-04-25
[BUG] Tapping on "Update synced reading lists" alerts the user that reading lists will be synced when they're logged out

nambatee:
https://phabricator.wikimedia.org/T192649

The `else` case on line `231` in `StorageAndSyncingSettingsViewController` is handled in #2256.


Wikipedia/Code/Panels.swift
```diff
@@ -193,6 +193,11 @@ extension UIViewController {
         }
         let panel = SyncEnabledPanelViewController(showCloseButton: true, primaryButtonTapHandler: primaryButtonTapHandler, secondaryButtonTapHandler: nil, dismissHandler: nil, theme: theme)
         let presenter = self.presentedViewController ?? self
+        if presenter is WMFThemeableNavigationController {
```
montehurd: Would it be possible to move this `if` to the place where the problematic call to `wmf_showSyncEnabledPanelOncePerLogin` is happening so it doesn't even dip into this func if we don't want it to? (No worries if not.)

Wikipedia/Code/Panels.swift
```diff
@@ -193,6 +193,11 @@ extension UIViewController {
...
+            guard presenter.presentedViewController == nil else {
+                return
```
montehurd: Do you think an assert makes sense here or is it something we expect to happen?

nambatee: We expect it to happen. "Sync enabled" panel will be presented when we get a response from the server that sync is enabled for the account. We can be presenting "Your reading lists will be synced in the background" alert at that time.

montehurd: If you expect it to happen it's fine.

PR #2256, opened 2018-04-25
[BUG] Tapping on "Update synced reading lists" alerts the user that reading lists will be synced when sync is disabled for their account

nambatee:
https://phabricator.wikimedia.org/T192650

Branched off #2255. #2255 has to be merged first.


Wikipedia/Code/Panels.swift
```diff
@@ -193,6 +198,11 @@ extension UIViewController {
         }
         let panel = SyncEnabledPanelViewController(showCloseButton: true, primaryButtonTapHandler: primaryButtonTapHandler, secondaryButtonTapHandler: nil, dismissHandler: nil, theme: theme)
         let presenter = self.presentedViewController ?? self
+        if presenter is WMFThemeableNavigationController {
+            guard presenter.presentedViewController == nil else {
+                return
```
montehurd: Could we use an assert here if we expect this should never happen? (had a similar question on the related PR)

nambatee: This is the same code 😄 responded in #2255

montehurd: haha "update branch before merging"... doh!

PR #2263, opened 2018-04-29
Share GIF

tiagomartinho:
When sharing an animated gif image from the gallery view, it only shares a still png image
https://phabricator.wikimedia.org/T139519

### DIFF
--- a/Wikipedia/Code/ShareActivityController.swift
+++ b/Wikipedia/Code/ShareActivityController.swift
@@ -207,8 +207,11 @@ class ShareActivityController: UIActivityViewController {
     
     @objc init(imageInfo: MWKImageInfo, imageDownload: ImageDownload) {
         var items = [Any]()
-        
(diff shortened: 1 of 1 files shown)


nambatee:
@tiagomartinho thank you for the PR! Looks good to me + tested on a device :) 👍

tiagomartinho:
Yes @nambatee! 

I also fixed the copy bug in the [NYTimes library](https://github.com/NYTimes/NYTPhotoViewer/pull/263) and in the [Wikimedia fork](https://github.com/wikimedia/NYTPhotoViewer/pull/1) but I couldn't understand why the Wikipedia app doesn't inherit the bug fix. I couldn't find any override of UIPasteboard or UIMenu​Controller in the iOS app.

joewalsh:
OK to test

tiagomartinho:
Thank you!

PR #2267, opened 2018-05-01
Add an in-line out of space alert to list items on reading list view and all articles view

nambatee:
https://phabricator.wikimedia.org/T191427

- adds a new `ReadingListAlertType`
- adds a new `isSavedToDisk` attribute to `WMFArticle`
- moves all alert configuration for `SavedArticlesCollectionViewCell.swift` to `configureAlert`


WMF Framework/Wikipedia.xcdatamodeld/Wikipedia.xcdatamodel/contents
```diff
@@ -79,6 +79,7 @@
         <attribute name="imageWidth" optional="YES" attributeType="Integer 32" defaultValueString="0" usesScalarValueType="NO" syncable="YES"/>
         <attribute name="isDownloaded" attributeType="Boolean" defaultValueString="NO" usesScalarValueType="YES" syncable="YES"/>
         <attribute name="isExcludedFromFeed" attributeType="Boolean" defaultValueString="NO" usesScalarValueType="YES" syncable="YES"/>
+        <attribute name="isSavedToDisk" attributeType="Boolean" defaultValueString="YES" usesScalarValueType="YES" syncable="YES"/>
```
montehurd: Seems there may be overlap between `isSavedToDisk` and `isDownloaded`. How do they differ?

nambatee: `isSavedToDisk` indicates if the save operations were successful. `isDownloaded` indicates if fetch was successful. `isDownloaded` will be set to `YES`, even if save to disk operations fail. If we want to modify the code to rely on `isDownloaded`, we'd still need something that tell us if the article is queued to be downloaded or if save to disk operations failed

montehurd: I see. Formerly I think I may have assumed `isDownloaded` was not set to true if the data could not be saved.

Wikipedia/Code/SavedArticlesFetcher.m
```diff
@@ -160,6 +160,19 @@ - (void)_update {
             self.updating = NO;
         }
     } else {
+        NSFetchRequest *notSavedToDiskRequest = [WMFArticle fetchRequest];
```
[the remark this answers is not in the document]
montehurd: b.t.w. If this gets too wonky given the structure of `SavedArticlesFetcher`, I think we could probably ask that this wait until we have time to pay down `SavedArticlesFetcher` tech debt in `5.8.2`...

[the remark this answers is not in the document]
nambatee: Moved out of the `else` case

WMF Framework/Wikipedia.xcdatamodeld/Wikipedia.xcdatamodel/contents
nambatee: If you comment out https://github.com/wikimedia/wikipedia-ios/blob/bf5086e63d9f7ae91084c2870c9540d247c2b591/Wikipedia/Code/MWKDataStore.m#L1127 and then try to save an article, its sections won't be saved but `isDownloaded` will be set to `YES`

Wikipedia/Code/MWKDataStore.m
```diff
@@ -1105,29 +1118,23 @@ - (void)saveArticle:(MWKArticle *)article {
     NSString *path = [self pathForArticle:article];
     NSDictionary *export = [article dataExport];
     NSError *error;
-    BOOL success = [self saveDictionary:export path:path name:@"Article.plist" error:&error];
-    if (!success) {
-        [self postArticleSaveToDiskDidFailNotification:article.url error:error];
-    }
+    [self saveDictionary:export path:path name:@"Article.plist" error:&error];
+    [self postArticleSaveToDiskDidFailNotification:article.url error:error];
```
montehurd: Consider changing the name of `postArticleSaveToDiskDidFailNotification` to reflect the changes made to it. (The name is no longer declarative of what it does.)

Wikipedia/Code/SavedArticlesFetcher.m
```diff
@@ -128,6 +128,20 @@ - (void)_update {
     self.updating = YES;
     NSAssert([NSThread isMainThread], @"Update must be called on the main thread");
     NSManagedObjectContext *moc = self.dataStore.viewContext;
+    
+    NSFetchRequest *notSavedToDiskRequest = [WMFArticle fetchRequest];
+    notSavedToDiskRequest.predicate = [NSPredicate predicateWithFormat:@"savedDate != NULL && isSavedToDisk != YES"];
```
montehurd: Should the predicate string have this at the end: ` && isDownloaded == YES`

Wikipedia/Code/SavedArticlesFetcher.m
```diff
@@ -128,6 +128,20 @@ - (void)_update {
...
+    notSavedToDiskRequest.fetchLimit = 1;
+    NSError *notSavedToDiskError = nil;
+    WMFArticle *articleToSaveToDisk = [[self.dataStore.viewContext executeFetchRequest:notSavedToDiskRequest error:&notSavedToDiskError] firstObject];
+    if (notSavedToDiskError) {
+        DDLogError(@"Error fetching article to save to disk: %@", notSavedToDiskError);
+    }
+    if (articleToSaveToDisk.URL) {
```
montehurd: Looks like this `if` only needs to happen if `notSavedToDiskError` is nil.

Wikipedia/Code/MWKDataStore.m
```diff
@@ -1086,10 +1086,22 @@ - (BOOL)saveString:(NSString *)string path:(NSString *)path name:(NSString *)nam
...
+- (void)postArticleSaveToDiskDidFailNotificationIfNeeded:(NSURL *)articleURL error:(NSError *)error {
     dispatch_async(dispatch_get_main_queue(), ^{
-        NSDictionary *userInfo = @{WMFArticleSaveToDiskDidFailErrorKey: error, WMFArticleSaveToDiskDidFailArticleURLKey: articleURL};
-        [NSNotificationCenter.defaultCenter postNotificationName:WMFArticleSaveToDiskDidFailNotification object:nil userInfo:userInfo];
+        WMFArticle *article = [self fetchArticleWithURL:articleURL];
+        if (error) {
+            NSDictionary *userInfo = @{WMFArticleSaveToDiskDidFailErrorKey: error, WMFArticleSaveToDiskDidFailArticleURLKey: articleURL};
+            [NSNotificationCenter.defaultCenter postNotificationName:WMFArticleSaveToDiskDidFailNotification object:nil userInfo:userInfo];
+            if (error.domain == NSCocoaErrorDomain && error.code == NSFileWriteOutOfSpaceError) {
+                article.isSavedToDisk = NO;
```
montehurd: It seems unusual that a notification posting method would mutate article.

Wikipedia/Code/SavedArticlesFetcher.m
nambatee: Wouldn't `articleToSaveToDisk` be nil if fetch failed? If you look at line 150 and 159, we're not checking if `fetchError` is `nil`, same on line 182 and 185, not checking if `downloadedFetchError` is `nil`?

Wikipedia/Code/SavedArticlesFetcher.m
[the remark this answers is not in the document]
montehurd: One thing to consider double-checking with this new location (if you haven't already) would be that it still works if space runs out while saving the last article.

Wikipedia/Code/SavedArticlesFetcher.m
montehurd: Yeah this whole fetcher beast is a bit of a Frankensteins monster :) If you're down it would be cool if L132-L143 were just pulled out into a method, then you can just return in the error case after DDLog.

montehurd: (ala Swifty `guard` pattern)

WMF Framework/Wikipedia.xcdatamodeld/Wikipedia.xcdatamodel/contents
joewalsh: That's a bug and these errors should bubble up to the `SavedArticlesFetcher`. Looking at addressing this in the other PR

Wikipedia/Code/SavedArticlesFetcher.m
```diff
@@ -128,6 +146,9 @@ - (void)_update {
     self.updating = YES;
     NSAssert([NSThread isMainThread], @"Update must be called on the main thread");
     NSManagedObjectContext *moc = self.dataStore.viewContext;
+    
+    [self trySavingArticlesToDiskIfNeeded];
```
joewalsh: I'd be concerned about adding a second fetch on the main thread here, would be good to profile and/or try to combine with the one below. Ideally it'd be nice to move the fetch below off the main thread as well, could be addressed with 5.8.2 tech debt.

PR #2275, opened 2018-05-02
Fetch & use displaytitle for search results, ensure URL creation is done with title

joewalsh:
- Add `displaytitle` to `ppprop` for search requests
- Update MWKSearchResult model version, add title, properly read displaytitle (falling back on title)
- Update all articleURL constructors that were using displaytitle to use the built in method on `MWKSearchResult` that uses title (`articleURLForSiteURL:`)

https://phabricator.wikimedia.org/T193105


montehurd:
Confirmed:
screen shot 2018-05-02 at 1 25 51 pm

joewalsh:
@montehurd good catch - added a question with some product implications here https://phabricator.wikimedia.org/T193101#4178112

```
Should we remove the HTML instead of falling back to title so we retain the capitalization?

Also, if we want to start showing italics, now would probably be a good time.
```

PR #2279, opened 2018-05-04
Use the same appInstallID for all event logging

nambatee:
https://phabricator.wikimedia.org/T189356

- removes `ReadingActionFunnel`
- removes references to `appInstallID` from individual funnels
- adds `appInstallID` to the event dictionary in `EventLoggingFunnel `
- removes `preprocessData` implementations if all they were doing was adding the funnel's `appInstallID `
- adds `wmf_appInstallID` to `NSUserDefaults+WMFExtensions` (get only)
- adds a `X-WMF-UUID` header to AF requests (in `wmf_applyAppRequestHeaders`)
- adds a `X-WMF-UUID` header to Session (`Session.swift`) requests


nambatee:
Retagging as 5.8.1

Wikipedia/Code/Session.swift
```diff
@@ -263,4 +265,11 @@ import Foundation
             completion(summaryResponses)
         }
     }
+    
+    private func addAppInstallIDIfNeeded(_ request: inout URLRequest) {
+        guard SessionSingleton.sharedInstance().shouldSendUsageReports, let appInstallID = UserDefaults.wmf_userDefaults().wmf_appInstallID else {
```
joewalsh: If this header needs to be set on all requests, it should be done through`URLSession.shared.configuration.httpAdditionalHeaders` - that way it covers the places in the code we're still using `URLSession.shared`  directly. There are also manually created `URLSession`s in `ImageController` and `EventLoggingService` that might need to be updated

Wikipedia/Code/Session.swift
```diff
@@ -32,15 +32,15 @@ import Foundation
...
+        let sharedSession = URLSession.shared
         guard SessionSingleton.sharedInstance().shouldSendUsageReports, let appInstallID = UserDefaults.wmf_userDefaults().wmf_appInstallID else {
-            return
+            return sharedSession
         }
-        session.configuration.httpAdditionalHeaders = ["X-WMF-UUID": appInstallID]
-    }
-
-    fileprivate let session = URLSession.shared
+        sharedSession.configuration.httpAdditionalHeaders = ["X-WMF-UUID": appInstallID]
```
nambatee: is this what you were thinking @joewalsh? 🤔

joewalsh: @nambatee yeah, but I think we need to update it if the `shouldSendUsageReports` preference changes

PR #2288, opened 2018-05-07
Keep event logging rate at 1 for WMFSearchFunnel

nambatee:
Keeping the rate for `WMFSearchFunnel` at 1 (child subclasses will get the rate from parent - https://github.com/wikimedia/wikipedia-ios/blob/2c59a14e4ffd07fab54ca6a4ac4794bf292bc84c/Wikipedia/Code/EventLoggingFunnel.m#L14) as requested by Chelsy.

### DIFF
--- a/Wikipedia/Code/WMFSearchFunnel.m
+++ b/Wikipedia/Code/WMFSearchFunnel.m
@@ -19,9 +19,6 @@ static NSString *const kSearchResultsCount = @"numberOfResults";
 
 - (instancetype)init {
     self = [super initWithSchema:kSchemaName version:kSchemaVersion];
-    if (self) {
-        self.rate = 100;
-    }
     return self;
 }
 


chelsyx:
Thanks @nambatee !
More context about this change: This is related to bug [T192520](https://phabricator.wikimedia.org/T192520). [L23 at WMFSearchFunnel.m](https://github.com/wikimedia/wikipedia-ios/blob/06491b0e26f09acd2006cf61cb40c1a595371a50/Wikipedia/Code/WMFSearchFunnel.m#L23) (`self.rate = 100;`) makes [L32 at EventLoggingFunnel.m](https://github.com/wikimedia/wikipedia-ios/blob/2c59a14e4ffd07fab54ca6a4ac4794bf292bc84c/Wikipedia/Code/EventLoggingFunnel.m#L32) (`if (self.rate == 1)`) failed, and thus no search event got sent to the beacon.

montehurd:
@nambatee @joewalsh since this is tagged `5.8.1` and the `buildmini` is still being updated, I'd say just run tests locally for this one.

nambatee:
Tested locally, merging

PR #2290, opened 2018-05-08
Fix Xcode 9.3 warnings followup

montehurd:
Followup to:
https://github.com/wikimedia/wikipedia-ios/pull/2151


montehurd:
@joewalsh This isn't exactly per your suggestions, but it simplifies things and is less of a diff from pre-Swift 4.1 than my original fix.

PR #2292, opened 2018-05-09
Fix trailing alignment warning

joewalsh:
I investigated this compiler warning because I was concerned it might be crashing on iOS 10, but it looks like it's just a misalignment.

Do not merge until after group discussion of the ticket:
https://phabricator.wikimedia.org/T194265

This introduces a behavior change as well, the close button won't scroll with the rest of the panel.


joewalsh:
Switching this to develop & 5.8.2 due to low prio

montehurd:
aaa mov


^ I checked scrolling behavior.

Now I remember why we scrolled the close button away: if we didn't we had this odd top margin (seen in the animation above) unless we constrained the top of the stackview to the very top, but doing so caused the close button to overlap the right side of the text when the text was scrolled up, and carolyn didn't want extra side margin.

joewalsh:
I'm fine either way, it was my assumption that the close button was supposed to stay visible, but it could just as easily be put inside the scroll view and scroll away again

carolynlimadeo:
In this case I think we can have it scroll off as tapping outside of the overlay will also dismiss the overlay.

montehurd:
Looks like the close button moved down. Otherwise lgtm.

| before  | after |
|screen shot 2018-05-09 at 4 56 39 pm|screen shot 2018-05-09 at 4 55 36 pm|

PR #2295, opened 2018-05-09
Use new pagelib reference collection bits.

montehurd:
https://phabricator.wikimedia.org/T194444

<strike>This doesn't include the pagelib version bump because the PR moving those bits to the page lib isn't merged at the moment. As soon as https://github.com/wikimedia/wikimedia-page-library/pull/140 is merged we can bump the pagelib version (as a new commit on this PR) and it should work.</strike>

<strike>Question: `5.8.2` or `5.9`? (kinda also depends on how quickly the pagelib PR clears CR)</strike>
Josh said ok for `5.8.2`


www/js/clickHandling.js
```diff
@@ -110,12 +110,31 @@ const sendMessageForImagePlaceholderWithTarget = innerPlaceholderSpan => {
...
 }
 
+/**
+ * Use "X", "Y", "Width" and "Height" keys so we can use CGRectMakeWithDictionaryRepresentation in
+ * native land to convert to CGRect.
+ * @param  {!ReferenceItem} referenceItem
+ * @return {void}
+ */
+const reformatReferenceItemRectToBridgeToCGRect = referenceItem => {
+  referenceItem.rect = {
```
montehurd: This was needed so the pagelib didn't have to know about iOS specific CGRect bridging details. The pagelib keeps the rect as a standard JS ClientRect.

montehurd:
jenkins retest this please

PR #2298, opened 2018-05-10
Remove pre-search debounce delay

montehurd:
https://phabricator.wikimedia.org/T192494

### DIFF
--- a/Wikipedia/Code/WMFSearchViewController.m
+++ b/Wikipedia/Code/WMFSearchViewController.m
@@ -319,34 +319,34 @@ static NSUInteger const kWMFMinResultsBeforeAutoFullTextSearch = 12;
 - (IBAction)textFieldDidChange {
     NSString *query = self.searchField.text;
 
-        DDLogDebug(@"Search field text changed to: %@", query);
+    DDLogDebug(@"Search field text changed to: %@", query);
 
-        BOOL isFieldEmpty = [query wmf_trim].length == 0;
+    BOOL isFieldEmpty = [query wmf_trim].length == 0;
 
-        /**
-         * This check is to avoid interpretting the "speech recognition in progress" blue spinner as
-         * actual text input. I could not find a clean way to detect this beyond subclassing the UITextField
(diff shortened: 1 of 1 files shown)


Wikipedia/Code/WMFSearchViewController.m
```diff
@@ -319,30 +319,8 @@ - (void)textFieldDidBeginEditing:(UITextField *)textField {
 - (IBAction)textFieldDidChange {
     NSString *query = self.searchField.text;
 
-    dispatchOnMainQueueAfterDelayInSeconds(0.4, ^{
```
montehurd: Ok to remove because when we get results we check if the results are for the search term (the success block captures the original search term):
https://github.com/wikimedia/wikipedia-ios/pull/2298/files#diff-6b0e15d84fb42eb1eea4b6ee96fe8b3bR419

montehurd: Ok to remove because when we get results we check if the results are for the search term (the success block captures the original search term):  
https://github.com/wikimedia/wikipedia-ios/pull/2298/files#diff-6b0e15d84fb42eb1eea4b6ee96fe8b3bR419

Wikipedia/Code/WMFSearchViewController.m
```diff
@@ -319,57 +319,34 @@ - (void)textFieldDidBeginEditing:(UITextField *)textField {
...
+        return;
+    }
 
-        [self setRecentSearchesHidden:YES animated:YES];
+    [self setRecentSearchesHidden:YES animated:YES];
 
-        DDLogDebug(@"Searching for %@ after delay.", query);
-        [self searchForSearchTerm:query wasSearchTermSuggested:NO];
-    });
+    DDLogDebug(@"Searching for %@ after delay.", query);
```
joewalsh: This log is no longer valid

montehurd: Fixed!

PR #2302, opened 2018-05-14
Add display title with HTML to WMFArticle

joewalsh:
- App fonts are now defined as `DynamicTextStyle`s that can be scaled for dynamic type and dynamically adjusted for different weights and different traits (italic). The underlying `UIFont` logic for doing this is somewhat messy so this abstracts that away
- WMFArticle keeps `displayTitle` (plaintext for certain areas of the UI and searching) and gains `displayTitleHTML`

https://phabricator.wikimedia.org/T193105


WMF Framework/UIFont+WMFDynamicType.swift
```diff
@@ -1,18 +1,60 @@
 import UIKit
 
-@objc public enum WMFFontFamily: Int {
+@objc(WMFFontFamily) public enum FontFamily: Int {
```
joewalsh: Font families no longer define weight or traits

PR #2303, opened 2018-05-15
Implement event logging for reading lists

nambatee:
https://phabricator.wikimedia.org/T190748

Schemas and flows: https://docs.google.com/presentation/d/1drI6nN3xQ5CMZrfQJkiMAYVI_5uQ1NypkYxg_qDXojk/edit?usp=sharing

- Enables new event logging
- Adds `ReadingListsFunnel`
- Renames old `LoginFunnel` to `WMFLoginFunnel`
- Adds `LoginFunnel`
- Adds `SettingsFunnel`
- Adds `UserHistoryFunnel`
- Adds `SessionsFunnel`


Wikipedia/Code/ArticleCollectionViewController.swift
```diff
@@ -8,7 +8,7 @@ protocol ArticleCollectionViewControllerDelegate: NSObjectProtocol {
 }
 
 @objc(WMFArticleCollectionViewController)
-class ArticleCollectionViewController: ColumnarCollectionViewController, ReadingListHintPresenter, EditableCollection {
+class ArticleCollectionViewController: ColumnarCollectionViewController, ReadingListHintPresenter, EditableCollection, EventLoggingEventValuesProviding, ReadingListsFunnelProvider {
```
montehurd: I'm guessing `provider` and `providing` mean the same thing? If so any thoughts on which one we could standardize on?

montehurd: Oh I guess a `provider` is an instance conforming to `providing`... never mind :)

WMF Framework/EventLoggingEventValuesProviding+EventLoggingCategory+EventLoggingLabel.swift
```diff
@@ -0,0 +1,60 @@
...
+    var standardEvent: Dictionary<String, Any> { get }
+}
+
+extension EventLoggingStandardEventProviding where Self: EventLoggingFunnel {
+    var standardEvent: Dictionary<String, Any> {
+        let appInstallID = wmf_appInstallID()
+        let timestamp = DateFormatter.wmf_iso8601().string(from: Date())
+        let sessionID = wmf_sessionID()
+        return ["app_install_id": appInstallID, "session_id": sessionID, "event_dt": timestamp];
+    }
```
nambatee: all new schemas require `app_install_id`, `session_id` and `event_dt`

Wikipedia/Code/KeychainCredentialsManager.swift
```diff
@@ -0,0 +1,20 @@
+public class KeychainCredentialsManager: NSObject {
+    private var keychainCredentials = WMFKeychainCredentials()
+    @objc public static let shared = KeychainCredentialsManager()
+    
+    @objc public var appInstallID: String? {
```
nambatee: If I wanted to put these inside `WMFAuthenticationManager`, calling `[WMFAuthenticationManager sharedInstance]` in `wmf_applyAppRequestHeaders` (https://github.com/wikimedia/wikipedia-ios/blob/ed91c31bf38a54d72f50540c3acb17e8685b7db6/Wikipedia/Code/AFHTTPRequestSerializer%2BWMFRequestHeaders.m#L12) would crash because `WMFAuthenticationManager` has a reference to `WMFAuthTokenFetcher`, which initializes `AFHTTPSessionManager`

Wikipedia/Code/LoginFunnel.swift
```diff
@@ -17,11 +17,6 @@
     }
     
     private func event(category: EventLoggingCategory, label: EventLoggingLabel?, action: Action, measure: Double? = nil) -> Dictionary<String, Any> {
-        guard category != .undefined else {
-            assertionFailure("category cannot be undefined")
-            return [:]
-        }
```
nambatee: The only `unknown` case that we want to catch will be caught here:

https://github.com/wikimedia/wikipedia-ios/blob/efb079dbddb15f54ded57c40908f91cb9561e355/Wikipedia/Code/ArticleCollectionViewController.swift#L84

PR #2310, opened 2018-05-16
Use PCS media endpoint for article media

joewalsh:
- Adds a network call to `WMFArticleFetcher` that fetches media from the PCS endpoint and passes it into `MWKArticle` for storage
- `MWKArticle` uses the new media if available, falls back on parsing if it isn't available
- Currently, the PCS media fetch is disabled due to changes pending on PCS ( https://gerrit.wikimedia.org/r/#/c/433563/ , https://phabricator.wikimedia.org/T195325 , https://gerrit.wikimedia.org/r/#/c/434143/ ). This can be re-enabled in a separate PR after the changes are merged.
- Fixes an issue where the image gallery was trying to immediately load the gallery width image instead of using the article width image as a placeholder, then loading the gallery width image. The user should no longer see a spinner when opening the gallery if the image is loaded already in the article view.


mdholloway:
Here's a PCS patch to provide the image descriptions in both HTML and plain text, in case you want to update for it: https://gerrit.wikimedia.org/r/#/c/434143/

I'll try to get this merged and deployed soon — maybe Tuesday at the latest?

joewalsh:
@mdholloway thanks! It looks like I mistakenly used caption instead of description. I'll update the PR to use description and accommodate the text-only version

mdholloway:
@joewalsh Oh, that's gross.  Thanks for the heads-up.  I Phabbed it: https://phabricator.wikimedia.org/T195845

PR #2317, opened 2018-05-22
Reading lists event logging follow-up

nambatee:
https://phabricator.wikimedia.org/T190748#

- [x] Update schema revisions
- [x] For updated schemas, send `measure` of type `Int` instead of `Double`
- [x] Log save via share sheet
- [x] For articles saved via 3d press in article view, update label to `out_link`
- [ ] For articles in "Read more" saved via 3d press in article view, update label to `read_more`
- [x] Log saving/unsaving main page with `main_page` label if it ever occurs
- [x] In article view, log saving a similar page with `similar_page` label and `article` category
- [ ] update `wiki` to be based on the wiki of the article affected ❓https://phabricator.wikimedia.org/T190748#
- [x] don't log unsave when removing articles from a non-default list
- [x] log `logout` only when user confirms logout
- [x] If the user doesn't opt in to send reports during onboarding and opts in in Settings later, send the snapshot when they opt in in Settings
- [x] set `measure_readinglist_itemcount` in MobileWikiAppiOSUserHistory to the count of unique saved articles

### DIFF
--- a/WMF Framework/EventLoggingStandardEventProviding+EventLoggingEventValuesProviding+EventLoggingCategory+EventLoggingLabel.swift
+++ b/WMF Framework/EventLoggingStandardEventProviding+EventLoggingEventValuesProviding+EventLoggingCategory+EventLoggingLabel.swift
@@ -87,6 +87,7 @@ public extension EventLoggingStandardEventProviding where Self: EventLoggingFunn
     case login
     case syncArticle
     case location
+    case mainPage
     case none
     
     public var value: String? {
@@ -125,6 +126,8 @@ public extension EventLoggingStandardEventProviding where Self: EventLoggingFunn
             return "sync_article" // Article storage and syncing settings screen
         case .location:
             return "location"
+        case .mainPage:
+            return "main_page"
         case .none:
             return nil
         }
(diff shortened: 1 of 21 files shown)


joewalsh:
> Can we know if the article 3d previewed in the article view is one of the articles in the Read more section? If the user uses TOC to navigate to Read more we can check if the selected item is TableOfContentsReadMoreItem but if they don't use TOC, we don't know

@nambatee it looks like the only thing we get is a URL in `- (BOOL)webView:(WKWebView *)webView shouldPreviewElement:(WKPreviewElementInfo *)elementInfo` - we could add a query parameter to the footer URLs that we look for and remove before opening the URL

WMF Framework/EventLoggingStandardEventProviding+EventLoggingEventValuesProviding+EventLoggingCategory+EventLoggingLabel.swift
```diff
@@ -1,6 +1,6 @@
-public protocol EventLoggingEventValuesProviding {
+@objc public protocol EventLoggingEventValuesProviding {
```
nambatee: Changed it to be available in objc to let `WMFExploreViewController` conform and provide the values. In `WMFArticleViewController` I'm checking if `articlePreviewingActionsDelegate` conforms to `EventLoggingEventValuesProviding` and log the provided values if it does. `WMFArticleViewController` handles peek & pop for a couple different controllers and I needed to log different values based on the origin of the peeked article.

PR #2319, opened 2018-05-22
Use latest build system

joewalsh:
Let's try this out for awhile to gauge if it's any faster for day to day development


montehurd:
@joewalsh 
I've been using it for the last couple weeks with no issues so far...

PR #2321, opened 2018-05-22
Add flag so we can tell if peeked item is a `Read more` item.

montehurd:
Adds a flag to the webVC which is used to record last time a `Read more` item was touched.

Then we simply can check in `webView:shouldPreviewElement:` if this happened recently - right now its checking if it happened within the last second.

This way was less invasive and didn't involve and pagelib or url hackery.

### DIFF
--- a/Wikipedia/Code/WMFArticleViewController.m
+++ b/Wikipedia/Code/WMFArticleViewController.m
@@ -1690,7 +1690,23 @@ static const CGFloat WMFArticleViewControllerTableOfContentsSectionUpdateScrollD
 
 #pragma mark - Article link and image peeking via WKUIDelegate
 
+- (BOOL)wasReadMoreItemTouchedRecently {
+    NSDate *lastReadMoreItemTouchStartDate = self.webViewController.lastReadMoreItemTouchStartDate;
+    if (lastReadMoreItemTouchStartDate != nil) {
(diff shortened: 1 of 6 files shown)


Wikipedia/Code/WMFArticleViewController.m
```diff
@@ -1690,7 +1690,23 @@ - (void)sectionEditorFinishedEditing:(SectionEditorViewController *)sectionEdito
...
+        }
+    }
+    return NO;
+}
+
 - (BOOL)webView:(WKWebView *)webView shouldPreviewElement:(WKPreviewElementInfo *)elementInfo {
+    BOOL isReadMoreItemBeingPeeked = [self wasReadMoreItemTouchedRecently];
+    
+// TODO: add event logging here (can now know if peeked item is a `Read more` item by checking `isReadMoreItemBeingPeeked`.
+NSLog(@"\nisReadMoreItemBeingPeeked = %d\n", isReadMoreItemBeingPeeked);
```
montehurd: @nambatee Here's the flag you can now check to see if a peeked item is a `Read more`.

montehurd:
Closing this for now - Joe has fix for the `?` encoding issue I ran into with the approach add a key/value to the `Read more` item urls.

PR #2327, opened 2018-05-23
Use local time with UTC offset in event logging timestamp

nambatee:


Wikipedia/Code/NSDateFormatter+WMFExtensions.m
```diff
@@ -19,18 +18,18 @@ + (NSDateFormatter *)wmf_iso8601Formatter {
     return iso8601Formatter;
 }
 
-+ (NSDateFormatter *)wmf_iso8601LocalizedFormatter {
-    static NSDateFormatter *iso8601LocalizedFormatter = nil;
++ (NSDateFormatter *)wmf_iso8601FormatterWithoutZone {
```
joewalsh: It seems like the intended result is a timestamp that looks like this:
`2018-05-24T07:26:57-04:00`

Apple added `NSISO8601DateFormatter` which can format in that style (RFC 3339). If you create a static `wmf_rfc3339LocalTimeZoneFormatter` with `timeZone` set to the local time zone and `formatOptions` set to `NSISO8601DateFormatWithInternetDateTime`, this should give you a string with local date & time with offset from UTC without having to do a two pass format.

I wouldn't bother updating `wmf_iso8601Formatter` formatter to `NSISO8601DateFormatter` yet, but it'd be good to use it for this new formatter

Wikipedia/Code/EventLoggingFunnel.h
```diff
@@ -118,6 +118,8 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (NSString *)wmf_sessionID;
 
+- (NSString *)timestamp;
```
joewalsh: if this is `@property (nonatomic, readonly) NSString *timestamp;` it'll allow you to use it as `.timestamp` instead of `.timestamp()` in Swift

PR #2330, opened 2018-05-24
Update pagelib

montehurd:
Allows us to know in `webView:shouldPreviewElement:` that the `elementInfo.linkURL` is a `Read more` URL - such urls will end in `?event_logging_label=read_more`.


montehurd:
Confirmed that `elementInfo.linkURL` in `webView:shouldPreviewElement:` has `?event_logging_label=read_more` when article `Read more` items are peeked.

PR #2331, opened 2018-05-24
Handle event logging label from pagelib

joewalsh:


Wikipedia/Code/NSURLComponents+WMFLinkParsing.m
```diff
@@ -105,4 +105,32 @@ - (NSString *)wmf_fragment {
     return [self.fragment precomposedStringWithCanonicalMapping];
 }
 
+- (nullable NSString *)wmf_eventLoggingLabel {
+    NSString *eventLoggingLabel = nil;
```
montehurd: Would `wmf_valueForQueryKey` be useful here? (and `wmf_urlWithoutQueryKey:` for removing it?)

PR #2332, opened 2018-05-25
Fixes issue with newlines in Wikidata descriptions that led to T195551

joewalsh:
In addition to apostrophes, escape newlines (caused the issue) and quotes (extra protection for future use) as well

https://phabricator.wikimedia.org/T195551


montehurd:
This lgtm - seems more resilient. 

I have pinged @berndsi and @mdholloway since iirc newlines shouldn't be in the descriptions. It's appearing in the desktop portal results as well:

screen shot 2018-05-25 at 8 44 08 am

Normally I'd just do a quick edit of the wikidata.org record for `San Francisco` but the description there doesn't match the one we're seeing so I don't know where it's coming from. Alternatively we can now override via the `{{SHORTDESC:` magic word (by adding one directly to the article wiki text). The article doesn't currently have a `{{SHORTDESC:` that I could see, so letting the services guys take a peek before adding one for quick fix.

montehurd:
BearND says it's likely coming from a template which is getting transcluded into the article (which is why I don't see a `{SHORTDESC:` when I look at the wiki text of the article itself.

PR #2335, opened 2018-05-25
Use new event logging label property to identify peeked read more items

nambatee:
In order to use the new event logging property in `WMFArticleViewController`, I would have to convert the `"read_more"` string to `EventLoggingLabel` (since Swift enum methods are not visible in objc, I can't go down that path) - one way to do it would be by passing a string to `ReadingListsFunnel` and letting it convert that to an enum. However, if I did that, I wouldn't be able to check if the `articlePreviewingActionsDelegate` conforms to `EventLoggingEventValuesProviding` and grab its label and category if it does, the way I am doing it right now. So I thought it might be better (as in easier to understand, easier to extend?, not safer) to give up on using Swift enums for category and label altogether.

Solution suggested by @joewalsh 🎉🥂:

defining `EventLoggingCategory` and `EventLoggingLabel`using `NS_TYPED_EXTENSIBLE_ENUM`


Wikipedia/Code/ArticleCollectionViewController.swift
```diff
@@ -85,8 +85,8 @@ class ArticleCollectionViewController: ColumnarCollectionViewController, Reading
         return .unknown
     }
     
-    var eventLoggingLabel: EventLoggingLabel {
-        return .none
+    var eventLoggingLabel: EventLoggingLabel? {
+        return nil
```
joewalsh: You could define `EventLoggingLabelNone` as nil if you want

Wikipedia/Code/ReadingListsFunnel.swift
```diff
@@ -16,12 +16,11 @@
     }
     
     private func event(category: EventLoggingCategory, label: EventLoggingLabel?, action: Action, measure: Int = 1) -> Dictionary<String, Any> {
-        let category = category.value
         let action = action.rawValue
         
         var event: [String: Any] = ["category": category, "action": action, "measure": measure, "primary_language": primaryLanguage(), "is_anon": isAnon]
-        if let labelValue = label?.value {
-            event["label"] = labelValue
+        if let label = label {
```
joewalsh: might be good to use `.rawValue` here and the other places to make it explicitly a `String`

Wikipedia/Code/WMFExploreViewController.m
```diff
@@ -2097,11 +2097,11 @@ - (void)logArticleSavedStateChange:(BOOL)wasArticleSaved saveButton:(WMFSaveButt
     }
 }
 
-- (EventLoggingCategory)eventLoggingCategory {
-    return EventLoggingCategoryFeed;
+- (NSString *)eventLoggingCategory {
+    return @"feed";
```
joewalsh: The changes to this file can be reverted now with the typedef

Wikipedia/Code/WMFArticleViewController.m
```diff
@@ -1877,17 +1889,17 @@ - (void)commitViewController:(UIViewController *)viewControllerToCommit {
...
     void (^logPreviewSaveIfNeeded)(void) = ^{
         BOOL providesEventValues = [weakArticlePreviewingActionsDelegate conformsToProtocol:@protocol(EventLoggingEventValuesProviding)];
         if (!providesEventValues) {
             return;
         }
         id<EventLoggingEventValuesProviding> eventLoggingValuesProvider = (id<EventLoggingEventValuesProviding>)weakArticlePreviewingActionsDelegate;
-        EventLoggingCategory eventLoggingCategory = [eventLoggingValuesProvider eventLoggingCategory];
-        EventLoggingLabel eventLoggingLabel = [eventLoggingValuesProvider eventLoggingLabel];
-        [[ReadingListsFunnel shared] logSaveWithCategory:eventLoggingCategory label:eventLoggingLabel measure:1 language:articleLanguage];
+        NSString *eventLoggingCategory = [eventLoggingValuesProvider eventLoggingCategory];
```
joewalsh: These can stay `EventLoggingCategory`

PR #2338, opened 2018-05-29
Use preferred variant for zh, default to zh-hans

joewalsh:
https://phabricator.wikimedia.org/T195884


montehurd:
Still doesn't seem to be getting the strings in the article lang:

aaaa mov

PR #2343, opened 2018-05-31
Allow multiple languages in the explore feed

joewalsh:
- Load explore feed content for all of the user's preferred languages

https://phabricator.wikimedia.org/T195972

### DIFF
--- a/WMF Framework/WMFExploreFeedContentController.m
+++ b/WMF Framework/WMFExploreFeedContentController.m
@@ -90,8 +90,7 @@ static const NSString *kvo_WMFExploreFeedContentController_operationQueue_operat
                                 feedContentSource,
                                 [[WMFRandomContentSource alloc] initWithSiteURL:siteURL],
                                 [[WMFAnnouncementsContentSource alloc] initWithSiteURL:siteURL],
-                                [[WMFOnThisDayContentSource alloc] initWithSiteURL:siteURL]
-                                ];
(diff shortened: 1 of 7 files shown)


WMF Framework/WMFExploreFeedContentController.m
```diff
@@ -75,23 +75,24 @@ - (WMFNearbyContentSource *)nearbyContentSource {
...
-            [[WMFMainPageContentSource alloc] initWithSiteURL:[self siteURL]],
-            [[WMFContinueReadingContentSource alloc] initWithUserDataStore:self.dataStore],
-            [[WMFNearbyContentSource alloc] initWithSiteURL:[self siteURL]
-                                                  dataStore:self.dataStore],
-            feedContentSource,
-            [[WMFRandomContentSource alloc] initWithSiteURL:[self siteURL]],
-            [[WMFAnnouncementsContentSource alloc] initWithSiteURL:[self siteURL]],
-            [[WMFOnThisDayContentSource alloc] initWithSiteURL:[self siteURL]]
-        ];
+        NSMutableArray *mutableContentSources = [NSMutableArray arrayWithCapacity:2 + self.siteURLs.count * 7];
```
montehurd: Wondering if we'll forget to adjust `7` here next time we add/remove a content source type...

joewalsh: There's not a huge penalty for the capacity being off in this case since this isn't performance critical or called frequently, we probably don't even need to set it here. Moreso just habit at this point

PR #2346, opened 2018-06-02
Add empty string protection for app install id and session id

joewalsh:
Best guess fix for https://phabricator.wikimedia.org/T196131 . Other possibilities are underlying issues with keychain access.


chelsyx:
Thanks @joewalsh ! 👍

PR #2348, opened 2018-06-02
Fix account messaging & logout status when in airplane mode

joewalsh:
https://phabricator.wikimedia.org/T196184

- Only fail re-auth on server-side failures, not `NSURLErrorDomain` failures
- Keep `loggedInUsername` if re-auth fails due to network connection so the user doesn't appear logged out

### DIFF
--- a/Wikipedia/Code/WMFAuthenticationManager.swift
+++ b/Wikipedia/Code/WMFAuthenticationManager.swift
@@ -142,12 +142,19 @@ public class WMFAuthenticationManager: NSObject {
             self.loggedInUsername = result.name
             userAlreadyLoggedInHandler(result)
         }, failure:{ error in
-            self.loggedInUsername = nil
-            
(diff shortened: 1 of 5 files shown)


Wikipedia/Code/WMFAuthenticationManager.swift
```diff
@@ -73,7 +73,7 @@ public class WMFAuthenticationManager: NSObject {
         return baseURL!
     }
     
-    @objc public func attemptLogin(_ completion: @escaping () -> Void = {}, failure: @escaping () -> Void = {}) {
+    @objc public func attemptLogin(_ completion: @escaping () -> Void = {}, failure: @escaping (_ error: Error) -> Void = {_ in }) {
```
nambatee: is `{}` vs `{ _ in }`a stylistic change or is `{}` incorrect?

PR #2349, opened 2018-06-03
Time profiler fixes

joewalsh:
- Fix language bundle caching bug (lookup was by language, cache was by path)
- Remove `performWithoutAnimation` inside of `SaveButton`'s `layoutSubviews` - @montehurd do you remember what this was fixing?
- Cache header icons
- Fix compiler warning for precision loss without cast

### DIFF
--- a/WMF Framework/Localization/WMFLocalization.m
+++ b/WMF Framework/Localization/WMFLocalization.m
@@ -57,7 +57,7 @@
         }
         bundle = [NSBundle bundleWithPath:path];
         if (bundle) {
-            bundles[path] = bundle;
+            bundles[language] = bundle;
         }
     }
     return bundle;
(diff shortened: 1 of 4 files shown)


montehurd:
>  @montehurd do you remember what this was fixing?

Checking...

montehurd:
@joewalsh 

iirc I was seeing weird jumpiness on the `Explore` feed `Save for later` buttons, but I just tried to repro and haven't been able to.

I added it here:
https://github.com/wikimedia/wikipedia-ios/pull/2208/files#diff-51538023d4c0e608efd2701ce8ccd1f8R55

In that diff you can see it we formerly had `UIView.performWithoutAnimation` around the `setTitle`, `setImage` and `layoutIfNeeded()`. 

I tried to repro both the way it was previously and without `UIView.performWithoutAnimation` and couldn't, so I say leave it out and if we see any odd `Save for later` behavior we can look at it again :)

PR #2352, opened 2018-06-04
Re-enable use of PCS for article media

joewalsh:


montehurd:
The media endpoint doesn't deliver image description language as of today. I added code for us to use it (once implemented) as part of this PR: 
https://github.com/wikimedia/wikipedia-ios/pull/2515/files#diff-9f7b151a9f56a0ff793cd54d9f88ae5eR354

I pinged @mdholloway about this and he created a ticket to add the same `description/lang` to the media endpoint that we get with the feed endpoint's image description.

mdholloway:
For the record, the mobileapps patch is deployed as of https://github.com/wikimedia/mediawiki-services-mobileapps-deploy/commit/4ef02e15c9072771aac0935a01f6721f9155c835.

montehurd:
Confirmed I'm now seeing `description/lang` for images which have the lang specified!

Example: first image for `Cat`:
https://en.wikipedia.org/api/rest_v1/page/media/Cat
<pre>
"description": {
     "html": "A montage of cat pictures using images from Wikimedia creators",
     "text": "A montage of cat pictures using images from Wikimedia creators",
     "lang": "en"
}
</pre>

montehurd:
@joewalsh I think we're probably good to resuscitate this PR now :)

PR #2354, opened 2018-06-05
Move values relevant to event logging into EventLoggingService, persist in event logging db

joewalsh:
https://phabricator.wikimedia.org/T196131

- Un-checking & re-checking "Send usage reports" in settings re-sets the app install id & days installed ( 360f835...4b0adcb ) as per https://phabricator.wikimedia.org/T196131#4261822
- Moves persistence of event logging related fields from user defaults and keychain into `EventLoggingService`
- Reverts `WMFKeychainCredentials` to 5.8.1 state
- Fixes potential mismatch of user agent version and event version if event is saved and then not submitted until the next version ( 7b6afcb )
- Fixes issue where `EventRecord` objects were being accessed outside of their `managedObjectContext`'s queue ( 44713a1 )


joewalsh:
WIP while making updates based on https://phabricator.wikimedia.org/T196131#4261822

Wikipedia/Code/WMFKeychainCredentials.swift
```diff
@@ -104,79 +79,35 @@ struct WMFKeychainCredentials {
...
-    private func tryGetString(forKey key: String) -> String? {
-        return try? string(forKey: key)
-    }
-    
-    private func tryGetValue(forKey key: String) -> Any? {
-        return try? value(forKey: key)
-    }
-    
-    private func string(forKey key: String) throws -> String {
-        let queryDictionary = matchQuery(forKey: key) as CFDictionary
```
montehurd: It's been a while so I don't recall details, but was using `matchQuery` here important?

iirc I added this [comment](https://github.com/wikimedia/wikipedia-ios/pull/2354/files#diff-0f72898e428341fe29ccb13e091ccfe4L9) as a reminder of why the internals were structured as they were, but also unsure if your change impacts what the comment refers to...

joewalsh: This is a revert on `WMFKeychainCredentials.swift` back to where it was before the addition of the event logging values. `matchQuery` was added as a convenience method and this reversion removes that convenience method and has it inline. It's possible we want to keep this convenience method, but I just brought over the file from the 5.8.1 release.

PR #2359, opened 2018-06-12
Write fastlane "description.txt" files for localized app store descriptions.

montehurd:
https://phabricator.wikimedia.org/T194688


montehurd:
Oops for some reason I thought this would cause the description files to be created at the time we run `fastlane build` but that doesn't seem to be the case...

montehurd:
Hmm it could be that I didn't include the updated `localization` binary...

montehurd:
@joewalsh 

I think I have it copying the binary over to `scripts` when we build `localization` target ( 0846b5a ). 

The binary ends up being a little bigger since it's copying the debug build binary vs the archive one... feel free to tweak as needed if this is a problem, but seems to work fine.

To confirm it's working I was able to make a tweak to `localization.swift` which added some debug text to the `description.txt` files strings, then I built `localization` target, then I built `Wikipedia` target and I saw the debug text appear in all the `description.txt` files.

PR #2374, opened 2018-06-28
Add global cards to Explore feed settings

nambatee:
- adds Continue reading and Because you read to Explore feed settings
- modifies Picture of the day to be non-language specific
- speeds up loading of the Explore feed settings view

To do:

- [x] Update copy

Will be addressed in other PRs (existing bugs that weren't introduced by this change):
- Reload animation
- Cut off footer label

PR #2378, opened 2018-06-29
Customize feed via customization buttons

nambatee:
Merge after #2374.

- Shows customization buttons for all the customizable content groups
- Shows action sheet with 3 options when customization button tapped


Wikipedia/Code/ExploreViewController.swift
```diff
@@ -532,6 +527,37 @@ extension ExploreViewController: ReadingListsAlertControllerDelegate {
     }
 }
 
+private extension WMFContentGroupKind {
+    var hideCardsActionTitle: String {
+        switch self {
+        case .featuredArticle:
+            return WMFLocalizedString("explore-feed-preferences-hide-featured-article-action-title", value: "Hide all Featured article cards", comment: "Title for action that allows users to hide all Featured article cards")
```
joewalsh: These strings could be a format - "Hide all %@ cards" with the existing titles (or new ones)

Wikipedia/Code/ExploreViewController.swift
```diff
@@ -411,12 +411,7 @@ class ExploreViewController: ColumnarCollectionViewController, ExploreCardViewCo
         cell.titleLabel.text = group.headerTitle
         cell.subtitleLabel.text = group.headerSubTitle
         cell.footerButton.setTitle(group.moreTitle, for: .normal)
-        switch group.contentGroupKind {
-        case .relatedPages, .locationPlaceholder:
-            cell.customizationButton.isHidden = false
-        default:
-            cell.customizationButton.isHidden = true
-        }
+        cell.customizationButton.isHidden = !group.contentGroupKind.isCustomizable
```
joewalsh: The customization button is missing on "Because you read" cards. Maybe because that kind `isGlobal` but not `isCustomizable`

PR #2380, opened 2018-06-30
Updates to the On this day card

montehurd:
https://phabricator.wikimedia.org/T198429

- [x] 2 events
- [x] gradient adjustments
- [x] timeline adjustments

screen shot 2018-06-29 at 5 13 04 pm

Question: ^ the spacing between the bottom of the `On this day` card and the top of the `Random article` card is too tight. It looks like this is a pre-existing issue, but am unsure if I've made it worse somehow...

### DIFF
--- a/Wikipedia/Code/ExploreCardViewController.swift
+++ b/Wikipedia/Code/ExploreCardViewController.swift
@@ -222,12 +222,12 @@ class ExploreCardViewController: PreviewingViewController, UICollectionViewDataS
         cell.selectionDelegate = self
     }
     
-    private func configureOnThisDayCell(_ cell: UICollectionViewCell, layoutOnly: Bool) {
-        guard let cell = cell as? OnThisDayExploreCollectionViewCell, let events = contentGroup?.contentPreview as? [WMFFeedOnThisDayEvent], events.count > 0 else {
+    private func configureOnThisDayCell(_ cell: UICollectionViewCell, forItemAt indexPath: IndexPath, layoutOnly: Bool) {
+        guard let cell = cell as? OnThisDayExploreCollectionViewCell, let events = contentGroup?.contentPreview as? [WMFFeedOnThisDayEvent], events.count > 0, events.indices.contains(indexPath.row) else {
             return
         }
-        let previousEvent: WMFFeedOnThisDayEvent? = events.count > 1 ? events[1] : events[0]
-        cell.configure(with: events[0], previousEvent: previousEvent, dataStore: dataStore, theme: theme, layoutOnly: layoutOnly)
+        let event = events[indexPath.row]
+        cell.configure(with: event, isFirst: events.first == event, isLast: events.last == event, dataStore: dataStore, theme: theme, layoutOnly: layoutOnly)
         cell.selectionDelegate = self
     }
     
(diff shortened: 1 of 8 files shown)


Wikipedia/Code/ExploreCardViewController.swift
```diff
@@ -222,12 +222,12 @@ class ExploreCardViewController: PreviewingViewController, UICollectionViewDataS
...
-    private func configureOnThisDayCell(_ cell: UICollectionViewCell, layoutOnly: Bool) {
-        guard let cell = cell as? OnThisDayExploreCollectionViewCell, let events = contentGroup?.contentPreview as? [WMFFeedOnThisDayEvent], events.count > 0 else {
+    private func configureOnThisDayCell(_ cell: UICollectionViewCell, forItemAt indexPath: IndexPath, layoutOnly: Bool) {
+        guard let cell = cell as? OnThisDayExploreCollectionViewCell, let events = contentGroup?.contentPreview as? [WMFFeedOnThisDayEvent], events.count > 0, events.indices.contains(indexPath.row) else {
             return
         }
-        let previousEvent: WMFFeedOnThisDayEvent? = events.count > 1 ? events[1] : events[0]
-        cell.configure(with: events[0], previousEvent: previousEvent, dataStore: dataStore, theme: theme, layoutOnly: layoutOnly)
+        let event = events[indexPath.row]
+        cell.configure(with: event, isFirst: events.first == event, isLast: events.last == event, dataStore: dataStore, theme: theme, layoutOnly: layoutOnly)
```
joewalsh: this can be quicker index == 0, index == events.count - 1 checks instead of equality checks on the event objects

joewalsh: or == events.indices.first, == events.indicies.last

nambatee:
simulator screen shot - iphone x - 2018-07-02 at 13 21 02
is the height off here? 🧐 (iPhone X, 11.4)

montehurd:
@joewalsh Do you think the calculation issue could be related to de-queued reusable cells not quite doing their size recalculations quite right? Per: https://phabricator.wikimedia.org/T198427

joewalsh:
@montehurd  @nambatee  the layout height issue is due to the layout cache re-using the same estimate for both cells

https://github.com/wikimedia/wikipedia-ios/blob/c9012fc5dc7d3044e2d65e9bc14e3e0dc5406b94/Wikipedia/Code/ExploreCardViewController.swift#L383

^ that should be updated to include the index in the key for the .event card type

Wikipedia/Code/ExploreCardViewController.swift
montehurd: Fixed.

Wikipedia/Code/OnThisDayExploreCollectionViewCell.swift
```diff
@@ -1,25 +1,30 @@
...
+            bottomGradientView.isHidden = !isLast
         }
         return super.sizeThatFits(size, apply: apply)
     }
 
     override open func setup() {
         super.setup()
-        addSubview(topGradientView)
-        addSubview(bottomGradientView)
+        clipsToBounds = true
```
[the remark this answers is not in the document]
montehurd: Oops, no. Initially I had the gradient view frames not set correctly, but I forgot to remove this after fixing that. Good catch!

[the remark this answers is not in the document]
montehurd: Fixed.

PR #2387, opened 2018-07-03
Turn off feed when last feed card is turned off

nambatee:
- Shows alert warning the user that feed will be turned off if they turn off all feed cards

TODO:

- [ ] Alert copy
- [ ] Switch to Search tab when Explore is replaced with Settings


nambatee:
ready for review, copy can be updated whenever

PR #2398, opened 2018-07-09
Initial implementation of large headers

nambatee:
* Adds bold headers to Explore, Saved, Places, History and Search
* Moves Settings bar button in Explore to the right
* Moves Settings' close button to the right
* Removes Search bar button from History

Workarounds:

* on iOS 11, bar buttons don't always show so right now I'm recreating them (lines 66 - 91 in `NavigationBar.swift`)

TODO:

* Create a new list button (right now the add button is on the right since adding that Create a new list button is a bigger effort)
* Theming


WMF Framework/Theme.swift
```diff
@@ -237,7 +237,7 @@ public class Theme: NSObject {
     }
 
     @objc public lazy var navigationBarBackgroundImage: UIImage = {
-        return UIImage.wmf_image(from: colors.paperBackground)
+        return #imageLiteral(resourceName: "transparent-pixel")
```
joewalsh: was this intentional or did I add this when we were considering going to the blur background?

nambatee: I think this happened when we were considering going to the blur bg

nambatee:
@montehurd I couldn't reproduce the Places artifact on a device, when you have time, could you see if it's happening on a device? 🧐

PR #2405, opened 2018-07-10
Feed theme updates

montehurd:
https://phabricator.wikimedia.org/T198146

- [x] hide card shadow on non-default themes
- [x] add card outline
- [x] update card content colors
- [x] ensure actual announcement card is correctly re-themed  
- [x] search bar theme colors

<strike>fix issues from @nambatee's screenshots</strike> (@carolynlimadeo is making separate ticket for these)


montehurd:
@nambatee Oh good question! I think so - @carolynlimadeo?

carolynlimadeo:
@montehurd @nambatee Yep, it'd be great to have these using paper instead of chrome. Monte, would some mocks in Zeplin help you?

WMF Framework/ExploreCardCollectionViewCell.swift
```diff
@@ -23,6 +23,12 @@ public class ExploreCardCollectionViewCell: CollectionViewCell, Themeable {
     
     static let overflowImage = UIImage(named: "overflow")
     
+    public var singlePixelDimension: CGFloat = 0.5
+    open override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        super.traitCollectionDidChange(previousTraitCollection)
+        singlePixelDimension = traitCollection.displayScale > 0 ? 1.0/traitCollection.displayScale : 0.5
```
joewalsh: You should ensure `displayScale > 0`, otherwise the pixel dimension could be negative or infinity

montehurd: hmm it looks like it's doing that...

joewalsh: 🤦‍♂️ my bad

PR #2414, opened 2018-07-12
Fix collection view crashes

joewalsh:


Wikipedia/Code/WMFTabBarUI.storyboard
```diff
@@ -94,7 +94,7 @@
         <!--History View Controller-->
         <scene sceneID="qtm-Om-dtd">
             <objects>
-                <viewController automaticallyAdjustsScrollViewInsets="NO" id="w0S-8T-SRU" customClass="WMFHistoryViewController" sceneMemberID="viewController">
```
joewalsh: ^ this was the crash culprit

PR #2418, opened 2018-07-12
Fix missing KVO contexts

joewalsh:
Unique contexts are necessary to ensure we're only handling the changes we're observing (and not any super or subclass observations). Also ensure `[super` is called to avoid KVO leaks


nambatee:
🙌🎊💐

PR #2422, opened 2018-07-13
Fix on this day detail header size

montehurd:
| Before  | After |
| screen shot 2018-07-12 at 5 29 14 pm | screen shot 2018-07-12 at 5 42 45 pm |


joewalsh:
@montehurd it looks like this was an underlying issue in the new collection view layout. The changes to constraints had no effect, it was just taking the estimated height and not recalculating as autolayout requested a new height. I opened #2424  with this fix.

montehurd:
@joewalsh oh sorry the storyboard changes were to fix the storyboard warning we get when you open the xib

PR #2445, opened 2018-07-17
Merge image scale transition code

joewalsh:
Merging the original work and making a few tweaks. Still not enabled, a lot of issues would need to be worked out.


joewalsh:
Reopening because the branch with the detail view modal change is at the point where a custom transition can be added, so it'd be good to merge this in.

PR #2446, opened 2018-07-17
Combine WMFTabBarUI, WMFAppViewController

joewalsh:
- Status bar appearance of view controllers is now automatically respected (was previously being blocked by the root view controller being WMFAppViewController)
- Simplifies initial app setup


montehurd:
follow-up lgtm too!

nambatee:
lgtm too

PR #2447, opened 2018-07-18
MobileWikiAppiOSUserHistory schema event logging updates

montehurd:
https://phabricator.wikimedia.org/T197194

Log new events for these schemas:

-----

[MobileWikiAppiOSUserHistory](https://meta.wikimedia.org/wiki/Schema:MobileWikiAppiOSUserHistory)
- [x] trend_notify
- [x] feed_disabled
- [x] feed_enabled_list
- [x] search_tab <strike>(blocked by [T199847](https://phabricator.wikimedia.org/T199847))</strike>

-----

<strike>MobileWikiAppiOSFeed</strike> 
^ Moving these to separate PR - other schema tweaks coming.


montehurd:
jenkins retest this please

Wikipedia/Code/UserHistoryFunnel.swift
```diff
@@ -43,6 +46,17 @@
         if let readingListCount = try? dataStore.viewContext.allReadingListsCount() {
             event["measure_readinglist_listcount"] = readingListCount
         }
+
+        var feedEnabledList = [String: Any]()
+        [WMFContentGroupKind.featuredArticle, WMFContentGroupKind.topRead, WMFContentGroupKind.onThisDay, WMFContentGroupKind.news, WMFContentGroupKind.relatedPages, WMFContentGroupKind.continueReading, WMFContentGroupKind.location, WMFContentGroupKind.random, WMFContentGroupKind.pictureOfTheDay].forEach({contentGroupKind in
```
nambatee: You could use static `customizableContentGroupKindNumbers` and `globalContentGroupKindNumbers` on `WMFFeedContentController` here, otherwise you'll be missing all the global groups + `WMFContentGroupKindLocationPlaceholder`

montehurd: Users can't disable logging WMFContentGroupKindLocationPlaceholder cells separately from WMFContentGroupKindLocation cards.

montehurd: (in `Settings > Explore feed`)

montehurd: though we can probably log the placeholder cards in `MobileWikiAppiOSFeed` vs this schema

nambatee: both content groups are disabled when either of them is disabled, the advantage of using the static arrays on `WMFExploreFeedContentController` instead of redeclaring it here would be not having to remember to update them in 2 places if anything changes

nambatee: (also, users can disable `WMFContentGroupKindLocationPlaceholder` directly from the feed)

Wikipedia/Code/UserHistoryFunnel.swift
```diff
@@ -89,3 +119,48 @@
...
 }
+
+private extension WMFContentGroupKind {
+    var offLanguageCodes: Set<String> {
+        let preferredLangCodes = MWKLanguageLinkController.sharedInstance().preferredLanguages.map{$0.languageCode}
+        return Set(preferredLangCodes).subtracting(languageCodes)
+    }
+    
+    // codes define by: https://meta.wikimedia.org/wiki/Schema:MobileWikiAppiOSUserHistory
+    var userHistorySchemaCode: String? {
```
montehurd: per our chat i made this optional

i also renamed this and the next var since they're specific to this funnel's schema (which is also why i moved it into this file and made the entire extension private)

PR #2454, opened 2018-07-19
Explore feed customization: Collapsed feed cards

nambatee:
https://phabricator.wikimedia.org/T198855

- Collapses explore feed card when user hides card or card type via overflow button

Question about scroll view during hiding: Right now, it's ok if I hide a single card or if I hide **all cards of type T** as long as all the other cards of type T are **below** the toggled card but if there's one card of type T **above** the toggled card, is the scrollView's behavior something I should change? 🤔
Repro: have all cards visible, scroll to yesterday's Featured article, hide all Featured article cards via overflow button on yesterday's Featured article

What to look for:
- Weird layout
- CV crashes
- Hiding all card types in the feed and then turning them back on in Settings


joewalsh:
@nambatee  re: scroll position, we might be able to fix that by adding something like
```
    public override func targetContentOffset(forProposedContentOffset proposedContentOffset: CGPoint) -> CGPoint {
        var superTarget = super.targetContentOffset(forProposedContentOffset: proposedContentOffset)
        superTarget.y += offsetDelta
        return superTarget
    }
```
to our collection view layout. Via https://developer.apple.com/videos/play/wwdc2014/232/

I think you would need to track the delta height of items & sections above the current screen. Not sure if this would work, but it's definitely called as it's making those changes.

Wikipedia/Code/ExploreViewController.swift
```diff
@@ -2,6 +2,8 @@ import UIKit
 import WMF
 
 class ExploreViewController: ColumnarCollectionViewController, ExploreCardViewControllerDelegate, UISearchBarDelegate, CollectionViewUpdaterDelegate, WMFSearchButtonProviding, ImageScaleTransitionProviding, DetailTransitionSourceProviding {
+
+    private var wantsDeleteInsertOnNexItemtUpdate: Bool = false
```
montehurd: small typo

Wikipedia/Code/ExploreViewController.swift
```diff
@@ -19,6 +21,29 @@ class ExploreViewController: ColumnarCollectionViewController, ExploreCardViewCo
...
         title = CommonStrings.exploreTabTitle
+
+        NotificationCenter.default.addObserver(self, selector: #selector(exploreFeedPreferencesDidSave(_:)), name: NSNotification.Name.WMFExploreFeedPreferencesDidSave, object: nil)
+    }
+
+    deinit {
+        NotificationCenter.default.removeObserver(self)
+    }
+
+    private func dismissCollapsedCards() {
```
joewalsh: @nambatee In retrospect, this might be a bit expensive for the foreground, and it'd be good to group all the updates in one batch to the UI. Could this kick off an operation on the feed content controller?

PR #2458, opened 2018-07-20
Add an extension on IndexPath to check if it's valid

nambatee:
Branched off #2454

f17125a is the relevant commit

It looks we often check if `indexPath` is valid in a given `fetchedResultsController`- would it make it easier to have an extension on `IndexPath` or do you feel like it would obscure things too much? 
Maybe it would make more sense to keep it in `ColumnarCollectionViewController`?


joewalsh:
@nambatee I think the extension is a good idea, however it might make more sense as an extension on `NSFetchedResultsController` because `NSFetchedResultsController` already depends on `IndexPath` and `IndexPath` has no knowledge of `NSFetchedResultsController`. Adding it on `NSFetchedResultsController` eliminates the circular dependency

PR #2477, opened 2018-07-24
As a user of search, I'd like to see fewer animations at the start of my search

nambatee:
https://phabricator.wikimedia.org/T199570


Wikipedia/Code/SearchViewController.swift
```diff
@@ -97,7 +97,9 @@ class SearchViewController: ArticleCollectionViewController, UISearchBarDelegate
         let start = Date()
         
         AFNetworkActivityIndicatorManager.shared().incrementActivityCount()
-        fakeProgressController.start()
+        DispatchQueue.main.asyncAfter(deadline: .now() + .milliseconds(1000)) {
```
joewalsh: If the user types a bunch, this will still call `start()` for every key. I thought the intent was to call `start()` 1s after the user has been idle. To do this, it might be best to integrate this into `FakeProgressController` as `cancelPreviousStartsAndStartAfter(_ after: TimeInterval)` and use `perform(` and `cancelPreviousPerformRequests(`

WMF Framework/FakeProgressController.swift
```diff
@@ -62,10 +64,10 @@ public class FakeProgressController: NSObject {
             if isProgressHidden {
                 NSObject.cancelPreviousPerformRequests(withTarget: self, selector: #selector(showProgress), object: nil)
                 NSObject.cancelPreviousPerformRequests(withTarget: self, selector: #selector(incrementProgress), object: nil)
-                perform(#selector(hideProgress), with: nil, afterDelay: 0.7)
+                perform(#selector(hideProgress), with: nil, afterDelay: minVisibleDuration)
             } else {
                 NSObject.cancelPreviousPerformRequests(withTarget: self, selector: #selector(hideProgress), object: nil)
-                showProgress()
+                perform(#selector(showProgress), with: self, afterDelay: delay)
```
joewalsh: This is missing a cancel of previous `showProgress` requests before kicking off the new one - as implemented `showProgress` calls would be made for every keystroke

PR #2479, opened 2018-07-24
Update the gradient on the Top read card's circled numbers in the Sepia theme to be easier to read

nambatee:
https://phabricator.wikimedia.org/T199490

### DIFF
--- a/Wikipedia/Code/ArticleCollectionViewCell+WMFFeedContentDisplaying.swift
+++ b/Wikipedia/Code/ArticleCollectionViewCell+WMFFeedContentDisplaying.swift
@@ -99,7 +99,7 @@ public extension RankedArticleCollectionViewCell {
     public override func configure(article: WMFArticle, displayType: WMFFeedDisplayType, index: Int, shouldShowSeparators: Bool = false, theme: Theme, layoutOnly: Bool) {
         rankView.rank = index + 1
         let percent = CGFloat(index + 1) / CGFloat(5)
-        rankView.tintColor = theme.colors.linkToAccent.color(at: percent)
+        rankView.tintColor = theme == .sepia ? UIColor.wmf_blue : theme.colors.linkToAccent.color(at: percent)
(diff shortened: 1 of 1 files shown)


Wikipedia/Code/ArticleCollectionViewCell+WMFFeedContentDisplaying.swift
```diff
@@ -99,7 +99,7 @@ public extension RankedArticleCollectionViewCell {
     public override func configure(article: WMFArticle, displayType: WMFFeedDisplayType, index: Int, shouldShowSeparators: Bool = false, theme: Theme, layoutOnly: Bool) {
         rankView.rank = index + 1
         let percent = CGFloat(index + 1) / CGFloat(5)
-        rankView.tintColor = theme.colors.linkToAccent.color(at: percent)
+        rankView.tintColor = theme == .sepia ? UIColor.wmf_blue : theme.colors.linkToAccent.color(at: percent)
```
joewalsh: Instead of having a hardcoded theme check, if you update the gradient to have the same start and end color, this class doesn't need to know about specific themes

nambatee: do you mean updating the start and end here https://github.com/wikimedia/wikipedia-ios/blob/41f3ff7c6e7bf327fef02bc9c3faad7d15a8560c/WMF%20Framework/Theme.swift#L214 by introducing new theme properties? If I modify `accent` on Sepia that will affect other places or am I missing something?

joewalsh: The gradient is currently only used on top read, so it could be renamed to `rankGradient`, and add new colors in the initializer for `rankGradientStart` and `rankGradientEnd`. Other color sets are start & end are the same as link & accent, sepia has both as link

joewalsh: Alternatively, the initializer could take a `rankGradient` parameter so it's one less variable to add

PR #2485, opened 2018-07-24
fix explore refresh spinner clipping

montehurd:
on explore if you pull to refresh by pulling very far down, pausing for a second, then releasing, when the top card moves back up it clips the spinner before the spinner disappears

(this is super minor - feel free to close if there are issues)


montehurd:
@joewalsh it does... will fix...

PR #2495, opened 2018-07-25
Update Places card spacing, fix invisible nav bar when adding articles to a new list

nambatee:
* Moves 'Enable location' button down
* Fixes invisible nav bar when adding articles to a new list via the 'Add to reading list' action on a share sheet


Wikipedia/Code/ReadingListsViewController.swift
```diff
@@ -294,6 +294,7 @@ class ReadingListsViewController: ColumnarCollectionViewController, EditableColl
 
     override func apply(theme: Theme) {
         super.apply(theme: theme)
+        view.backgroundColor = theme.colors.paperBackground
```
nambatee: this is for ->
screen shot 2018-07-25 at 9 02 39 am

PR #2498, opened 2018-07-25
Move collapsed card away from section headers

nambatee:
* Fixes layout for collapsed cells
* Fixes RTL


joewalsh:
@nambatee I don't think this should have to check it's frame origin. In `ExploreCardCollectionViewCell`:

```
} else if isCollapsed {
            if apply {
                let cardBackgroundViewFrame = CGRect(x: layoutMargins.left, y: contentView.frame.origin.y, width: widthMinusMargins, height: contentView.frame.height)
                cardBackgroundView.frame = cardBackgroundViewFrame
            }
        }
```
Any idea why that's `contentView.frame.origin.y` and not `layoutMargins.top`?

joewalsh:
same question for `contentView.frame.height` - the height should be defined by the height required by the undo label & button

nambatee:
@joewalsh `layoutMargins.top` results in uneven spacing

screen shot 2018-07-25 at 1 23 12 pm

joewalsh:
@nambatee yeah, I think the bigger issue is what I mentioned in the second comment - the height of the whole cell (and the background view) needs to be defined by the undo labels, not the other way around. It looks like the bottom margin isn't being added at the end either.

montehurd:
Looks like this may undo the [side margin consistency change](https://github.com/wikimedia/wikipedia-ios/pull/2488/commits/d61ce64e57d94927a1536641ee8205017de83a42) (made the undo cell use same internal side margins as other cards) from my pixel push session yesterday. 

screen shot 2018-07-25 at 12 49 46 pm

montehurd:
🔥🔥 it's perfect! 🔥🔥

WMF Framework/ExploreCardCollectionViewCell.swift
```diff
@@ -226,24 +225,30 @@ public class ExploreCardCollectionViewCell: CollectionViewCell, Themeable {
                 cardBackgroundView.frame = cardContentViewFrame.insetBy(dx: -singlePixelDimension, dy: -singlePixelDimension)
             }
             origin.y += cardContentViewFrame.layoutHeight(with: 20)
-        } else if isCollapsed {
+        }
+
+        if isCollapsed, !undoLabel.isHidden, !undoButton.isHidden {
+            let undoInsets: (x: CGFloat, y: CGFloat) = (x: 15, y: 16)
```
joewalsh: Could this be a `UIOffset`,  a`CGPoint`, or `UIEdgeInsets`?

WMF Framework/ExploreCardCollectionViewCell.swift
```diff
@@ -226,24 +225,30 @@ public class ExploreCardCollectionViewCell: CollectionViewCell, Themeable {
...
+            labelOrigin.x += undoInsets.x
+            labelOrigin.y += undoInsets.y
+
+            let undoLabelMaxWidth = widthMinusMargins - (widthMinusMargins * 0.25)
```
joewalsh: Might be good to pull `0.25` out into a local variable as `undoButtonMaxWidthPercentage` so it's only updated in one place for both calculations

WMF Framework/ExploreCardCollectionViewCell.swift
```diff
@@ -226,24 +225,30 @@ public class ExploreCardCollectionViewCell: CollectionViewCell, Themeable {
...
-            var undoLabelFrame = undoLabel.wmf_preferredFrame(at: undoLabelOrigin, maximumWidth: widthMinusMargins, minimumWidth: 84, horizontalAlignment: labelHorizontalAlignment, apply: false)
-            let halfHeight = round(0.5 * undoLabelFrame.height)
-            undoLabelFrame.origin.y -= halfHeight
-            undoLabel.frame = undoLabelFrame
-        }
 
-        if !undoButton.isHidden {
-            let undoButtonOrigin = CGPoint(x: origin.x - 8, y: undoLabel.frame.origin.y)
-            origin.y += undoButton.wmf_preferredHeight(at: undoButtonOrigin, maximumWidth: widthMinusMargins, horizontalAlignment: buttonHorizontalAlignment, spacing: 20, apply: apply)
+            origin.y += cardBackgroundViewFrame.layoutHeight(with: layoutMargins.bottom)
```
joewalsh: I think we want to add `layoutMargins.bottom` if it isn't collapsed as well. Right now the space is made up for with `let interItemSpacing: CGFloat = 32` in layout metrics, which should really be `17` and the `layoutMargins.bottom` is the remaining`15`

WMF Framework/ExploreCardCollectionViewCell.swift
```diff
@@ -226,24 +225,30 @@ public class ExploreCardCollectionViewCell: CollectionViewCell, Themeable {
...
+            labelOrigin.y += undoInsets.y
+
+            let undoLabelMaxWidth = widthMinusMargins - (widthMinusMargins * 0.25)
+            let undoLabelMinWidth = widthMinusMargins * 0.5
+            let undoLabelX = isRTL ? widthMinusMargins - undoLabelMaxWidth : labelOrigin.x
+            let undoLabelFrameHeight = undoLabel.wmf_preferredHeight(at: CGPoint(x: undoLabelX, y: labelOrigin.y), maximumWidth: undoLabelMaxWidth, minimumWidth: undoLabelMinWidth, horizontalAlignment: labelHorizontalAlignment, spacing: 0, apply: apply)
+
+            let undoButtonMaxWidth = widthMinusMargins * 0.25
+            let undoButtonX = isRTL ? labelOrigin.x : widthMinusMargins - undoButtonMaxWidth
+            let undoButtonFrameHeight = undoButton.wmf_preferredHeight(at: CGPoint(x: undoButtonX, y: labelOrigin.y), maximumWidth: undoButtonMaxWidth, horizontalAlignment: buttonHorizontalAlignment, spacing: 0, apply: apply)
```
joewalsh: It'd be nice if the undo button was vertically centered for longer text. Maybe minimum height should be `undoLabelFrameHeight`? 
simulator screen shot - iphone 8 - 2018-07-26 at 09 38 04

WMF Framework/ExploreCardCollectionViewCell.swift
```diff
@@ -226,24 +225,30 @@ public class ExploreCardCollectionViewCell: CollectionViewCell, Themeable {
...
+
+            let undoHeight = max(undoLabelFrameHeight, undoButtonFrameHeight)
+            let cardBackgroundViewHeight = ceil(origin.y) + undoHeight + undoInsets.y
```
joewalsh: What's the reason for `origin.y` here? Conceptually it seems like `cardBackgroundViewHeight` should be `undoHeight + 2*undoInsets.y`

WMF Framework/ExploreCardCollectionViewCell.swift
nambatee: what about 'layoutMargins.top'? If `interItemSpacing` becomes `17`, wouldn't we need to add `layoutMargins.top` to have `32` in spacing on both ends?

joewalsh: `interItemSpacing` is just between the items. If we want additional spacing above the first item or below the last item, we would need to add `sectionInset`. This naming convention matches `UICollectionViewFlowLayout`. My implementation in `ColumnarCollectionViewLayoutInfo` is a bit off at the moment but should probably be updated. For example, it's currently erroneously adding `interItemSpacing` below the last item

montehurd:
This PR fixed an RTL issue on this ticket: https://phabricator.wikimedia.org/T199569

PR #2507, opened 2018-07-26
Editing languages during onboarding crashes app

nambatee:
https://phabricator.wikimedia.org/T200431

could have been causing 
https://phabricator.wikimedia.org/T200433 as well


Wikipedia/Code/WMFAppViewController.m
```diff
@@ -414,7 +414,14 @@ - (void)updateExploreFeedPreferencesIfNecessary {
     MWKLanguageLink *mostRecentlyModifiedPreferredLanguage = languageLinkController.mostRecentlyModifiedPreferredLanguage;
     NSURL *siteURL = mostRecentlyModifiedPreferredLanguage.siteURL;
     BOOL appendedNewPreferredLanguage = [preferredLanguages containsObject:mostRecentlyModifiedPreferredLanguage];
-    [self.dataStore.feedContentController toggleContentForSiteURL:siteURL isOn:appendedNewPreferredLanguage updateFeed:NO];
+    if (self.isPresentingOnboarding) {
+        __typeof(self) __weak weakSelf = self;
+        self.exploreViewController.exploreFeedPreferencesUpdateOnFirstAppearance = ^{
+            [weakSelf.dataStore.feedContentController toggleContentForSiteURL:siteURL isOn:appendedNewPreferredLanguage waitForCallbackFromCoordinator:NO updateFeed:NO];
```
nambatee: this call was causing the crash, lmk if there's a better way to apply initial prefs to the feed (than passing a block like this and running it in `viewDidLoad` of `ExploreViewController`)

joewalsh: @nambatee it looks like if the user makes multiple language changes, this could get out of sync. Also, it would replace the block on explore with whatever latest change is made. Is it possible to make this sync based on the whole preferred language list instead of depending on `mostRecentlyModifiedPreferredLanguage`?

nambatee: actually I think this is not needed at all because with feed update we will create default preferences if they don't exist

PR #2513, opened 2018-07-27
Allow article image header to passthrough touches to the article view

joewalsh:
https://phabricator.wikimedia.org/T200501


nambatee:
it breaks peek & pop on the lead image for me

joewalsh:
@nambatee I'm not seeing an easy fix for the peek/pop issue. Might want to file that as a follow on fix

PR #2514, opened 2018-07-27
Cleanup AFHTTPSessionManager extensions

montehurd:
This is a follow-up to https://github.com/wikimedia/wikipedia-ios/pull/2524.

Cleanup `AFHTTPSessionManager` extensions.

- removes unused `retry` params (logic was removed a while ago)
- removes naming which indicates retries happen (logic was removed a while ago)
- removes unused methods
- re-names methods which use zero url to be explicit about doing so
- uses swiftier method signatures
- adds some assertions just in case


montehurd:
Oh good question - will check.

montehurd:
Per @joewalsh's suggestion I switched to `POST` vs `GET`. Was more invasive but more robust too.

PR #2519, opened 2018-07-30
Fix Explore feed preferences being off for visible content groups

nambatee:
Explore feed preferences are set when `exploreFeedPreferencesInManagedObjectContext` is called (https://github.com/wikimedia/wikipedia-ios/blob/d8703acf3eb45c804ae8d2d7224521fede412335/WMF%20Framework/WMFExploreFeedContentController.m#L392). It will be called when explore feed preferences are updated by the user (https://github.com/wikimedia/wikipedia-ios/blob/d8703acf3eb45c804ae8d2d7224521fede412335/WMF%20Framework/WMFExploreFeedContentController.m#L559) or when feed is updated and new preferences are applied (https://github.com/wikimedia/wikipedia-ios/blob/d8703acf3eb45c804ae8d2d7224521fede412335/WMF%20Framework/WMFExploreFeedContentController.m#L653).
In case neither of the above scenarios take place before we try to access the cached explore feed preferences, we should fall back on `defaultExploreFeedPreferences`.


nambatee:
@joewalsh it's possible but it will be fetched many times

joewalsh:
@nambatee that's OK as long as most of them are in the background - the view context one should only be fetched when the user goes into settings?

nambatee:
@joewalsh sounds good 👍

nambatee:
@joewalsh when you say `fetch the value`, do you mean fetching via `WMFAsyncBlockOperation` or fetching it off the context outside of the operation and using `NSLock` to protect access while fetch is in progress? With fetching via `WMFAsyncBlockOperation`, we'd have to use the retrieved value in the completion block which would require using `WMFTaskGroup` to wait for operation inside the completion block to be completed before we return, or is there another way to do this?

nambatee:
@joewalsh Alternatively, couldn't we keep `exploreFeedPreferences` and instead of falling back on defaults if it's nil, fetch from moc?

joewalsh:
@nambatee sorry, I missed these comments earlier. I meant fetching it from the `viewContext` when accessing from the main thread (for example, in the settings UI) and from the `feedImportContext` when accessing it from the `feedImportContext`'s private queue. 

You can fetch to initialize the value in the accessor, but there should be a assertion to ensure we're only ever performing that fetch on the main thread. There's also currently an issue keeping that cached value in sync. I noticed that when
```
[[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(viewContextDidSave:) name:NSManagedObjectContextDidSaveNotification object:self.dataStore.viewContext];
```  
is called, `self.dataStore.viewContext` is nil, so you're getting save notifications from every context.

It might make sense to figure out how to hook into `- (void)viewContextDidChange:(NSNotification *)note` on `MWKDataStore` to keep the prefs in sync. Alternatively, you could ensure that every time preferences are updated and successfully saved to the `feedImportContext`, `exploreFeedPreferences` is updated afterwards on the main thread and the`WMFExploreFeedPreferencesDidSaveNotification` is posted

nambatee:
@joewalsh thanks, that helps! 🙌 What if I set the observer in dataStore's setter https://github.com/wikimedia/wikipedia-ios/pull/2519/commits/b88a172d7238638bc448811022aa4fc47b567389?

joewalsh:
@nambatee that should work if you change to`NSManagedObjectContextDidSaveNotification` to `NSManagedObjectContextDidChangeNotification`- IIRC the `viewContext` is set up to only save when the app goes to background. I could be wrong, though

nambatee:
@joewalsh `viewContextDidSave` is called right after I save new preferences via settings or overflow menu 🧐 if I break or print in there

joewalsh:
@nambatee it looks I was remembering wrong - the view context is saved every time a background context is saved (`MWKDataStore backgroundContextDidSave:`)

WMF Framework/WMFExploreFeedContentController.m
```diff
@@ -70,6 +69,15 @@ - (void)setOperationQueue:(NSOperationQueue *)operationQueue {
 - (void)setDataStore:(MWKDataStore *)dataStore {
     _dataStore = dataStore;
     self.exploreFeedPreferencesUpdateCoordinator = [[ExploreFeedPreferencesUpdateCoordinator alloc] initWithFeedContentController:self];
+    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(viewContextDidSave:) name:NSManagedObjectContextDidSaveNotification object:self.dataStore.viewContext];
+}
+
+- (NSDictionary *)exploreFeedPreferences {
+    if (!_exploreFeedPreferences) {
+        assert([NSThread isMainThread]);
```
joewalsh: This assertion either needs to be moved up or this method (and the setter) needs to be made atomic. Otherwise you could be accessing this at the same time on two different threads.

joewalsh: If the assertion moves up, the setter would also need a main thread assertion

PR #2522, opened 2018-07-30
Apple Maps license

josve05a:
Should the license attribution URL link to the actual license, or Apple Maps Terms of Use? I'd say to the actual license.


joewalsh:
OK to test

PR #2529, opened 2018-07-31
Order explore feed by preferred language

joewalsh:
https://phabricator.wikimedia.org/T200050
https://phabricator.wikimedia.org/T200460

### DIFF
--- a/WMF Framework/WMFExploreFeedContentController.m
+++ b/WMF Framework/WMFExploreFeedContentController.m
@@ -201,11 +201,7 @@ NSString *const WMFNewExploreFeedPreferencesWereRejectedNotification = @"WMFNewE
                                         [moc performBlock:^{
                                             NSError *saveError = nil;
                                             if ([moc hasChanges]) {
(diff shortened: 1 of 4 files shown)


WMF Framework/WMFExploreFeedContentController.m
```diff
@@ -200,11 +201,7 @@ - (void)updateFeedSourcesWithDate:(nullable NSDate *)date userInitiated:(BOOL)wa
                                         [moc performBlock:^{
                                             NSError *saveError = nil;
                                             if ([moc hasChanges]) {
-                                                if (date) {
```
joewalsh: @nambatee do you remember why the `if (date)` switch is here?

nambatee: I think it's because originally we were applying preferences to the updated objects when feed was updated but then we saw that it wasn't always working (in some cases, we wanted to apply preferences to all objects on feed update) so this date check achieved that without always fetching all objects

PR #2542, opened 2018-08-02
Don't catch nil reading list entries in ReadingListDetailViewController

nambatee:
When article is updated, edit controller updates `availableActions` for all the swipeable cells:

https://github.com/wikimedia/wikipedia-ios/blob/4cd192c8c94eb049430b512a4793b510dcce44fc/WMF%20Framework/CollectionViewEditController.swift#L100

However, when an article is deleted (and cell disappears), we shouldn't be setting `availableActions` for that cell (deleting an article from a reading list will trigger assertion failure)

We could

1) Remove the deleted `indexPath` from `configuredCellsByIndexPath` before `articleWasUpdated` is triggered
2) In `articleWasUpdated` guard against changes to the article that involve reading lists but not savedDate?
3) Remove the assertion that catches nil entries

From what I see option 1 would be unreliable because I can't see how we can remove the deleted `indexPath` from `configuredCellsByIndexPath` before moc is saved, we can know when an article is about to be deleted but many things can go wrong in the meantime and we might not delete the article in the end

Option 2 might break other things?


joewalsh:
@nambatee I think it should be on the view controllers to re-call `configureSwipeableCell` if the underlying data changes. This way, the `CollectionViewEditController` is just concerned about handling editing state with the collection view, not the underlying data model changes

PR #2547, opened 2018-08-06
Bump feed refresh timeout interval to 60s

joewalsh:
With multiple languages being requested and less frequent failures, this should be bumped to 60 seconds


joewalsh:
OK to test

PR #2548, opened 2018-08-06
Use newestVisibleGroupOfKind to ensure only visible cards are selecte…

joewalsh:
…d for the featured article widget

Repro steps:

- Have multiple languages selected, load an explore feed day with all languages, verify featured article card is visible for all languages
- Turn off all but one of the languages
- View the featured article widget

Expected:
Featured article widget is in the only language left enabled

Actual:
Featured article widget is one of the disabled languages

The daily sort order is set lower due to the language not being in the list of preferred languages. This ensures only visible cards are selected for the featured article widget. This would also prevent hidden cards from appearing in the featured article widget.


joewalsh:
ok to test

PR #2551, opened 2018-08-07
Add MediaWiki language code mapping

joewalsh:
- Send `sr-ec` and `sr-el` instead of `sr-cyrl` and `sr-latn`
- Only send language code if the language isn't present in the mapping

https://phabricator.wikimedia.org/T198419


joewalsh:
retest this please

PR #2552, opened 2018-08-07
Implement feed event logging

nambatee:


nambatee:
retest this please

WMF Framework/WMFExploreFeedContentController.m
```diff
@@ -156,6 +156,7 @@ - (void)updateFeedSourcesUserInitiated:(BOOL)wasUserInitiated completion:(nullab
 
 - (void)updateFeedSourcesWithDate:(nullable NSDate *)date userInitiated:(BOOL)wasUserInitiated completion:(nullable dispatch_block_t)completion {
     WMFAssertMainThread(@"updateFeedSources: must be called on the main thread");
+    [FeedFunnel.shared logFeedRefreshed];
```
joewalsh: Does it matter here if the user refreshed it? I would have assumed it would only  be `if (wasUserInitiated)` so background updates aren't logged

nambatee: yeah, good point, maybe this should be in Explore's `refresh` and `scrollViewDidScroll` (Chelsy wanted to log feed update on scroll too)

Wikipedia/Code/ColumnarCollectionViewController.swift
```diff
@@ -257,6 +257,13 @@ class ColumnarCollectionViewController: ViewController, ColumnarCollectionViewLa
         previewingContext.sourceRect = view.convert(cell.bounds, from: cell)
         return indexPath
     }
+
+    // MARK: - Event logging utiities
+
+    var maxViewed: NSNumber {
+        let value = ((Double(collectionView.contentOffset.y) + Double(collectionView.bounds.height)) / Double(collectionView.contentSize.height)) * 100
```
joewalsh: If the user scrolls to the bottom (100%), then scrolls half way back up (50%) and selects an item to view, this will be 50% instead of 100%

joewalsh: You should be able to override `scrollViewDidScroll` as long as it still calls super. You might also want to ensure `contentSize.height.isGreaterThan` is greater than 0 and that you aren't returning percentages higher than 100

nambatee: why should it be 100% if they scrolled back up to 50%?

nambatee: nvm I missed `max` in `maxViewed` 😂

Wikipedia/Code/ArticleCollectionViewController.swift
```diff
@@ -101,6 +101,19 @@ class ArticleCollectionViewController: ColumnarCollectionViewController, Reading
...
     }
+
+    var eventLoggingIndex: NSNumber? {
+        guard let index = previewedIndexPath?.item else {
+            return nil
+        }
+        return NSNumber(value: index)
+    }
+
+    var contentGroup: WMFContentGroup?
```
joewalsh: If a property on this Core Data object is accessed after the object is deleted, it causes a crash. You could either observe core data changes and nil out this property if the content group is deleted (I don't think this makes sense in this case) OR just transfer the analytics properties you need as strings or some other non-core data class

Wikipedia/Code/ColumnarCollectionViewController.swift
```diff
@@ -260,9 +260,14 @@ class ColumnarCollectionViewController: ViewController, ColumnarCollectionViewLa
 
     // MARK: - Event logging utiities
 
-    var maxViewed: NSNumber {
-        let value = ((Double(collectionView.contentOffset.y) + Double(collectionView.bounds.height)) / Double(collectionView.contentSize.height)) * 100
-        return NSNumber(value: value)
+    var maxViewed: NSNumber = NSNumber(value: 0)
+
+    override func scrollViewDidScroll(_ scrollView: UIScrollView) {
+        guard collectionView.contentSize.height > 0 else {
```
joewalsh: missing a `super.scrollViewDidScroll` above the guard

nambatee:
LGTM

PR #2556, opened 2018-08-08
When the app is upgraded, we want to send a new snapshot to MobileWikiAppiOSUserHistory

nambatee:
https://phabricator.wikimedia.org/T200388

This would work on the 2nd upgrade - I don't think there's a way to dig out previous app version for current users? 🧐

### DIFF
--- a/Wikipedia/Code/AppDelegate.m
+++ b/Wikipedia/Code/AppDelegate.m
@@ -67,6 +67,10 @@ static NSTimeInterval const WMFBackgroundFetchInterval = 10800; // 3 Hours
 #pragma mark - UIApplicationDelegate
 
 - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
+    NSString *lastAppVersion = NSUserDefaults.wmf_userDefaults.wmf_lastAppVersion;
+    if (!lastAppVersion) {
+        NSUserDefaults.wmf_userDefaults.wmf_lastAppVersion = WikipediaAppUtils.appVersion;
+    }
 
 #if WMF_IS_NEW_EVENT_LOGGING_ENABLED
     [[WMFEventLoggingService sharedInstance] start];
--- a/Wikipedia/Code/AppDelegate.m
+++ b/Wikipedia/Code/AppDelegate.m
@@ -67,11 +67,6 @@ static NSTimeInterval const WMFBackgroundFetchInterval = 10800; // 3 Hours
 #pragma mark - UIApplicationDelegate
 
 - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
-    NSString *lastAppVersion = NSUserDefaults.wmf_userDefaults.wmf_lastAppVersion;
-    if (!lastAppVersion) {
-        NSUserDefaults.wmf_userDefaults.wmf_lastAppVersion = WikipediaAppUtils.appVersion;
-    }
-
 #if WMF_IS_NEW_EVENT_LOGGING_ENABLED
     [[WMFEventLoggingService sharedInstance] start];
 #endif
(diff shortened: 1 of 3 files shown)


Wikipedia/Code/AppDelegate.m
```diff
@@ -67,6 +67,10 @@ - (void)updateDynamicIconShortcutItems {
 #pragma mark - UIApplicationDelegate
 
 - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
+    NSString *lastAppVersion = NSUserDefaults.wmf_userDefaults.wmf_lastAppVersion;
+    if (!lastAppVersion) {
+        NSUserDefaults.wmf_userDefaults.wmf_lastAppVersion = WikipediaAppUtils.appVersion;
```
joewalsh: `appVersion` includes the build number - is it OK to send snapshots when a beta user upgrades between builds or should it only be on major version changes?

joewalsh:
retest this please

Wikipedia/Code/AppDelegate.m
```diff
@@ -67,6 +67,10 @@ - (void)updateDynamicIconShortcutItems {
 #pragma mark - UIApplicationDelegate
 
 - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
+    NSString *lastAppVersion = NSUserDefaults.wmf_userDefaults.wmf_lastAppVersion;
+    if (!lastAppVersion) {
```
joewalsh: This will only be updated once when `lastAppVersion` is `nil` - it needs an `isEqualToString:` check. You could do `if (appVersion && ![lastAppVersion isEqualToString:appVersion])` and that should cover nil & inequality cases

nambatee: Shouldn't it be updated after I log a new snapshot? If I check if `lastAppVersion == appVersion`  when app launches and set `lastAppVersion` to `appVersion`, I will lose `lastAppVersion` and the check in snapshot log will fail

joewalsh: Yeah that makes sense - or explicitly log the snapshot here and remove the check from the funnel

Wikipedia/Code/UserHistoryFunnel.swift
```diff
@@ -95,10 +96,14 @@ private typealias ContentGroupKindAndLoggingCode = (kind: WMFContentGroupKind, l
         guard isTarget else {
             return
         }
-        
+        guard let lastAppVersion = UserDefaults.wmf_userDefaults().wmf_lastAppVersion else {
+            assertionFailure("Last app version should be set by now")
```
joewalsh: could the initial set of `lastAppVersion` be moved from `AppDelegate` to here to keep it contained within one class?

nambatee: Not really, User history snapshot is logged at the end of the session so if we set `lastAppVersion` at that time, it will be the same as current `appVersion`. I can wrap it in `setLastAppVersion` method on `UserHistoryFunnel` and call it in `AppDelegate` if you think that's better

nambatee: but we could set it in `logStartingSnapshot`

joewalsh: I mean just in the `nil` case - rather than having an assertion here, you could set the initial value here (you could also potentially use this to log the initial upgrade from 5.8 to 6.0 by logging when the previous value was nil)

nambatee: we're gonna log the first snapshot no matter what in `logStartingSnapshot ` and then once we want to log another one, we should be good setting `lastAppVersion` in `logSnapshot` since it's only necessary for that check in there, no? 🤔

joewalsh: If you `logStartingSnapshot` without setting the version, then log on `nil` version in `logSnapshot`, won't it log twice for new users?  Setting the version in `logStartingSnapshot` when it reaches `log(event())` should prevent the double log

nambatee: yeah that's a good point 🤯, but we should be ok cause we're setting `wmf_lastAppVersion` to `WikipediaAppUtils.appVersion()` in `logged` so after the first log, `wmf_lastAppVersion` won't be nil

PR #2559, opened 2018-08-08
Fix crash on selecting "No recent searches yet" cell

nambatee:
Repro steps

1) Make sure you have no recent searches
2) Go to the Search tab
3) Tap on "No recent searches yet"

### DIFF
--- a/Wikipedia/Code/MWKList.m
+++ b/Wikipedia/Code/MWKList.m
@@ -191,6 +191,9 @@
 }
 
 - (id)objectInEntriesAtIndex:(NSUInteger)idx {
+    if (![self countOfEntries] || [self countOfEntries] < idx) {
+        return NULL;
+    }
     return [_mutableEntries objectAtIndex:idx];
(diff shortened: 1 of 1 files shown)


nambatee:
also, should selection be disabled if there are no recent searches?

joewalsh:
Yeah - it might be easier to use `func collectionView(_ collectionView: UICollectionView, shouldSelectItemAt indexPath: IndexPath) -> Bool {` and return false when it's empty

Wikipedia/Code/MWKList.m
```diff
@@ -191,6 +191,9 @@ - (NSUInteger)countOfEntries {
 }
 
 - (id)objectInEntriesAtIndex:(NSUInteger)idx {
+    if (![self countOfEntries] || [self countOfEntries] < idx) {
```
joewalsh: should be `<= idx`, `idx == countOfEntries` is out of range

PR #2560, opened 2018-08-08
Add event logging testing section to readme

montehurd:


README.md
```diff
@@ -47,6 +47,29 @@ We use [Carthage](https://github.com/Carthage/Carthage) to manage third-party na
 ## Testing
 The **Wikipedia** scheme is configured to execute the project's iOS unit tests, which can be run using the `Cmd+U` hotkey or the **Product->Test** menu bar action.
 
+### Event logging
+Note: Testing event logging requires `labs` access to `deployment-eventlog05.eqiad.wmflabs`
+
+To test event logging:
+- ensure event logging is enabled via `Gear icon > Send usage reports`
+- use debugging settings for `scheme` and `host` in `EventLoggingService.swift` as seen [here](https://github.com/wikimedia/wikipedia-ios/pull/2552/commits/0b0e848064118b984255b79a933cae7daf42d7ef#diff-39f3285546bbb773abe632f15a594ca5)
```
joewalsh: this could be set up as a separate target (similar to the beta cluster) rather than having to update scheme and host manually

montehurd: @joewalsh 

Looking at beta cluster bits it looks like we need a new: 
- scheme
- build configuration
- target
- macro
- `#if`/`#else`/`#endif`:
```
#if WMF_EVENT_LOGGING_DEV_DEBUG
    private static let scheme = "http"
    private static let host = "deployment.wikimedia.beta.wmflabs.org"
#else
    private static let scheme = "https"
    private static let host = "meta.wikimedia.org"
#endif
```

Seems bizarrely complicated... am I missing something?

PR #2561, opened 2018-08-08
Log snapshot after app upgrade when session starts, log feed impression after delay

nambatee:
- If user upgrades the app, snapshot is logged at the beginning of the session (not the end)
- Feed impressions are logged after 3s delay

### DIFF
--- a/Wikipedia/Code/AppDelegate.m
+++ b/Wikipedia/Code/AppDelegate.m
@@ -67,6 +67,7 @@ static NSTimeInterval const WMFBackgroundFetchInterval = 10800; // 3 Hours
 #pragma mark - UIApplicationDelegate
 
 - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
+    [UserHistoryFunnel.shared logSnapshot];
 #if WMF_IS_NEW_EVENT_LOGGING_ENABLED
     [[WMFEventLoggingService sharedInstance] start];
 #endif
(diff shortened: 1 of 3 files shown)


Wikipedia/Code/AppDelegate.m
```diff
@@ -67,6 +67,7 @@ - (void)updateDynamicIconShortcutItems {
 #pragma mark - UIApplicationDelegate
 
 - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
+    [UserHistoryFunnel.shared logSnapshot];
 #if WMF_IS_NEW_EVENT_LOGGING_ENABLED
     [[WMFEventLoggingService sharedInstance] start];
 #endif
```
joewalsh: It might make sense to move everything above this line into `WMFAppViewController` after all the migrations have run. In general, it's nice to not load up `didFinishLaunchingWithOptions:` - I think Andy put it here initially while testing and we never moved it where it should be

nambatee: do you mean everything below line `70` should go into `WMFAppViewController`'s `launchAppInWindow`, after `[self migrateIfNecessary]`? Or do you mean just starting event logging shouldn't be here?

joewalsh: Just ```[UserHistoryFunnel.shared logSnapshot];
#if WMF_IS_NEW_EVENT_LOGGING_ENABLED
    [[WMFEventLoggingService sharedInstance] start];
#endif```

joewalsh: and can be added below `[self checkRemoteAppConfigIfNecessary];`

joewalsh: err - actually scratch this for now, it can be a follow on if we get to it

PR #2567, opened 2018-08-09
Event logging dev debug scheme

montehurd:
- adds `Event Logging Dev Debug` scheme so you don't have to uncomment debugging host and url when testing event logging changes


montehurd:
I'm missing some critical bit here - the `WMF_EVENT_LOGGING_DEV_DEBUG` flag doesn't seem to be working...

joewalsh:
screen shot 2018-08-10 at 1 36 50 pm
@montehurd ^ i'm not seeing the scheme - is it marked as shared on your end?

montehurd:
@joewalsh oops good catch - fixed!

chelsyx:
Thanks for creating this debug scheme! I'm wondering whether it is better to make the same change to the beta scheme(?), so that the eventlogging data from beta testers would be sent to the backend beta clusters as well, instead of the production database.

joewalsh:
@chelsyx the same "Wikipedia" app binary we release to beta testers on Test Flight ends up in the app store so it'd probably be more complex and require a check a runtime. We do have separate "Wikipedia Beta" and "Wikipedia Alpha" apps that don't get released to the app store that we could switch to the backend clusters

chelsyx:
@joewalsh Got you. Please switch "Wikipedia Beta" and "Wikipedia Alpha" to the beta cluster and thank you!

Also, I guess for the same reason, it would be hard to add "beta" to the user agent string for the beta app released on Test Flight so that they can be identified easily, right?

joewalsh:
@chelsyx correct, for the same reason. We could move all of our internal testers & test flight testers over to the "Wikipedia Beta" app.

chelsyx:
@joewalsh That sounds great!

PR #2570, opened 2018-08-10
Improve eventlogging SearchFunnel on iOS app

nambatee:
https://phabricator.wikimedia.org/T195937

- [x] Rename the schema to MobileWikiAppiOSSearch
- [x] Add a source field to identify the source from which the search interface was invoked: the top of feed, the search tab, or from the article view
- [x] Add a langswitch action when users tap on a different language on the search interface
- [x] Add a position field to log the position of the search result that was clicked, from the list that was presented to the user
- [x] Add is_anon, primary_language, session_id fields to standardize it with other iOS eventlogging schemas
- [x] Rename other fields from camelCase to snake_case because our database are case insensitive


Wikipedia/Code/SearchViewController.swift
```diff
@@ -557,18 +557,39 @@ extension SearchViewController: CollectionViewHeaderDelegate {
...
+        guard let navigationController = navigationController, navigationController.viewControllers.count > 0 else {
+            return "unknown"
+        }
+        let viewControllers = navigationController.viewControllers
+        let viewControllersCount = viewControllers.count
+        if viewControllersCount == 1 {
+            return "search_tab"
+        }
+        let penultimateViewController = viewControllers[viewControllersCount - 2]
+        if penultimateViewController is WMFArticleViewController {
```
joewalsh: Might be nice to decouple this as a protocol - `AnalyticsSearchSourceProviding` so this doesn't have to check specific classes. Not a necessary fix, just an optional enhancement

PR #2576, opened 2018-08-13
As a multilingual user, I'd like to know that I can update my Explore feed language settings via the Languages section of the Settings screen

nambatee:
https://phabricator.wikimedia.org/T199843


montehurd:
I'm seeing localization file changes appear when I build this. Prob just need to add those to the PR...

nambatee:
@montehurd  #2577 should fix that

montehurd:
When I tap the lang icon on an article, the other articles langs seem to be missing:

| Before  | After |
|before mov|after mov|

montehurd:
One more small issue - in the gifs above the title used in the article language switching case was "Change language" but is now "Wikipedia languages". Disregard if this was intentional for the article lang picking case.

PR #2578, opened 2018-08-14
Update the Settings header to match other existing tab headers

nambatee:
https://phabricator.wikimedia.org/T200301


montehurd:
hmm i'm still seeing the old `Settings` header...

screen shot 2018-08-13 at 6 45 04 pm

PR #2579, opened 2018-08-14
Fix jerky tablet TOC toggle animation

montehurd:
https://phabricator.wikimedia.org/T201642

- hides the herky-jerky text reflow
- fine-tuned the fade, durations and delays with Carolyn


montehurd:
We may want to have Anthony double-check that this is an actual regression - I suspect it was this way previously... in which case we could bump the change to 6.0.1.

montehurd:
@jtminor said ok to wait for 6.0.1 :)

PR #2587, opened 2018-08-17
Fix various gallery image description issues

montehurd:
https://phabricator.wikimedia.org/T200268

- [x] fix gallery description truncation on landscape
- [x] make description scrollable
- [x] tap description to expand/collapse long descriptions
- [x] very subtle top gradient so close and share buttons don't completely disappear when over white parts of image

asdfasdf mov


Wikipedia/Code/WMFScreenHeightConstrainedGradientTextView.swift
```diff
@@ -0,0 +1,109 @@
...
+            0.0,
+            NSNumber(value: normalizedFadeHeight),          // upper stop
+            NSNumber(value: 1.0 - normalizedFadeHeight),    // lower stop
+            1.0
+        ]
+
+        // Maintain fixed mask location on scroll
+        CATransaction.begin()
+        CATransaction.setDisableActions(true)
+        gradientMask.frame = CGRect(x: 0, y: contentOffset.y, width: bounds.size.width, height: bounds.size.height)
```
joewalsh: It might be better to use the approach used in `RoundedCornerView` and update only when the size of the view changes without a `CATransaction`

Wikipedia/Code/WMFScreenHeightConstrainedGradientTextView.swift
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    override func awakeFromNib() {
+        super.awakeFromNib()
+        assert(contentCompressionResistancePriority(for: .vertical) == .required, "vertical contentCompressionResistancePriority must be `.required` for height to correctly account for text height.")
+    }
+    
+    override var intrinsicContentSize: CGSize {
+        var size = super.intrinsicContentSize
+        size.height = size.height.constrainedBetween(minHeight: minHeight, maxPercentOfScreenHeight: maxPercentOfScreenHeight)
```
joewalsh: The view shouldn't have to know about the screen size here - the View Controller should be controlling this aspect of min/max size for the gradient text view

Wikipedia/Code/WMFImageGalleryDetailOverlayView.m
```diff
@@ -73,16 +36,25 @@ - (IBAction)didTapInfoButton {
     }
 }
 
+- (IBAction)didTapImageDescription {
```
joewalsh: Could this be propagated up to the View Controller via a delegate so that this doesn't have to tell the superview to `layoutIfNeeded`?

Wikipedia/Code/WMFImageGalleryGradientViews.swift
```diff
@@ -0,0 +1,28 @@
+@objcMembers class WMFImageGalleryTopGradientView: WMFUnconfiguredGradientView {
+    public static let sharedStretchableImage = {
+        return WMFImageGalleryTopGradientView.init(frame: CGRect(x: 0, y: 0, width: 25, height: 44)).wmf_stretchableSnapshotImageWithAlphaChannel()
```
joewalsh: is the bar always `44` or is it `32` in landscape?

montehurd: that’s why I used a stretchable img

joewalsh: assuming it's a gradient that fades vertically, it should only be stretched horizontally, otherwise it gets distorted

Wikipedia/Code/WMFScreenHeightConstrainedGradientTextView.swift
montehurd: checking...

Wikipedia/Code/WMFImageGalleryGradientViews.swift
montehurd: hmm i didn't notice any distortion, but may be because i was on simulator and wasn't zoomed in to 1:1... but ya i recall gradient re-sampling producing banding...

Wikipedia/Code/WMFImageGalleryDetailOverlayView.m
montehurd: sure - i'll try to move the other `didTap` methods right above this one while i'm at it...

montehurd: fixed!

Wikipedia/Code/WMFScreenHeightConstrainedGradientTextView.swift
montehurd: This approach keeps controller bloat down, but ya I didn't like that it had knowledge of screen height...

Wikipedia/Code/WMFScreenHeightConstrainedGradientTextView.swift
montehurd: The size is usually constant on scroll. The rounded corner view example seems like it wouldn't quite work even if switched to check position. The transaction keeps the mask re-positioning animation-free (https://stackoverflow.com/a/23291097/135557).

Hmm...

joewalsh: I see now - I misunderstood and thought this was just adjusting for size and not content offset. Instead of repositioning the mask on scroll, could it be a view with a static mask that has this UITextView as a subview?

montehurd: Yes it could. I considered it but was trying to do it w/o extra view. The CATransaction doesn't seem to impact performance (though I haven't checked in instruments), but I'm totally fine with switching to separate view.

montehurd: fixed! i like it better now too :)

montehurd: no more having to fiddle with the scrolly bits!

Wikipedia/Code/WMFScreenHeightConstrainedGradientTextView.swift
montehurd: It now gets passed the available height.

montehurd:
@joewalsh i made those changes we chatted about. @carolynlimadeo said the visual "can minimize" indicator design will be landing some time this week - ok if i add that as a follow on? no worries either way.

montehurd:
jenkins retest this please

PR #2589, opened 2018-08-18
Add missing UIGraphicsEndImageContext call

montehurd:


joewalsh:
@montehurd I updated this to re-use an existing category

PR #2591, opened 2018-08-20
Fix Event logging crash

nambatee:
https://rink.hockeyapp.net/manage/apps/152731/app_versions/653/crash_reasons/233144328

### DIFF
--- a/WMF Framework/EventLoggingService.swift
+++ b/WMF Framework/EventLoggingService.swift
@@ -154,7 +154,7 @@ public class EventLoggingService : NSObject, URLSessionDelegate {
     public func start() {
         assert(Thread.isMainThread, "must be started on main thread")
 
-        let operation = AsyncBlockOperation.init { (operation) in
+        let operation = AsyncBlockOperation { (operation) in
             DispatchQueue.main.async {
                 self.urlSession = URLSession(configuration: self.urlSessionConfiguration, delegate: self, delegateQueue: nil)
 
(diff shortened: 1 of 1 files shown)


WMF Framework/EventLoggingService.swift
```diff
@@ -286,38 +288,32 @@ public class EventLoggingService : NSObject, URLSessionDelegate {
...
                     }
+                    eventRecords = try moc.fetch(fetch)
+                } catch let error {
+                    DDLogError(error.localizedDescription)
                 }
-                eventRecords = try moc.fetch(fetch)
-            } catch let error {
-                DDLogError(error.localizedDescription)
+
+                operation.finish()
```
joewalsh: This will finish early if there are event records. This might be cleaner with the event record declaration outside of the try catch, and the if statement at the end of the method instead of inside the defer as 
```
if eventRecords.count > 0 {
   self.postEvents(eventRecords, completion: {
    operation.finish()
  })
} else {
  operation.finish()
}

File: WMF Framework/EventLoggingService.swift [hunk 2]
```diff
@@ -286,38 +288,34 @@ public class EventLoggingService : NSObject, URLSessionDelegate {
...
                 var eventRecords: [EventRecord] = []
-                defer {
-                    if eventRecords.count > 0 {
-                        self.postEvents(eventRecords)
-                    } else {
-                        self.postLock.lock()
-                        self.posting = false
-                        self.postLock.unlock()
+                do {
+                    defer {
```
joewalsh: If code is added after this do/catch block, it'd be called after this `defer` statement. I didn't know this was the case, but tested it with a playground:

```
do {
    defer {
        print("defer")
    }
}

sleep(5)
print("ok")
```

Since there's no early return to necessitate the defer, it'd be clearer to add the contents of this `defer` block after the do/catch

PR #2594, opened 2018-08-21
[BUG] Title missing from reading list hint when tapping article "Read more" save buttons

nambatee:
https://phabricator.wikimedia.org/T198270

### DIFF
--- a/Wikipedia/Code/ReadingListHintViewController.swift
+++ b/Wikipedia/Code/ReadingListHintViewController.swift
@@ -13,7 +13,14 @@ class ReadingListHintViewController: UIViewController {
     }
     
     private var hintButtonTitle: String {
-        let articleTitle = article?.displayTitle ?? "article"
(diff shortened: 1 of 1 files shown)


Wikipedia/Code/ReadingListHintViewController.swift
```diff
@@ -13,7 +13,14 @@ class ReadingListHintViewController: UIViewController {
     }
     
     private var hintButtonTitle: String {
-        let articleTitle = article?.displayTitle ?? "article"
+        var articleTitle = "article"
```
joewalsh: It'd be nice if this wasn't hardcoded to "article" in English, but that's outside the scope of this PR

joewalsh: (should be nil or empty by default, then have a separate localized string that is "Add this article to a reading list?" that shows when the title is nil or empty)

PR #2595, opened 2018-08-21
Add generic string equivalent for reading list hint

joewalsh:


montehurd:
hehe `article` wasn't good enough ;)

PR #2601, opened 2018-08-22
Update the Settings header to match other existing tab headers

nambatee:
https://phabricator.wikimedia.org/T200301


Wikipedia/Code/WMFViewController.h
```diff
@@ -18,6 +18,10 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (nonatomic, readwrite) BOOL ignoresTopContentInset; // workaround for WKWebView ignoring touches inside of the content inset https://bugs.webkit.org/show_bug.cgi?id=134230
 
+@property (nonatomic, readwrite) BOOL definesScrollViewInsets;
```
joewalsh: You should be able to drop these properties (and the associated code) - it looks like the inset bug is because the settings table view is constrained to the layout margins guide top & bottom instead of the superview top & bottom. If you update those constraints and remove this, the existing code should work fine

montehurd:
This is pretty great!!!

PR #2611, opened 2018-08-24
Reinstate the feed layout cache with fixes for invalidation by content group and article

joewalsh:
Alleviates performance issues on older devices


nambatee:
looks like layout isn't adjusted when we collapse cells

simulator screen shot - iphone 8 - 2018-08-24 at 13 29 08

joewalsh:
@nambatee I saw that before but thought it was fixed by 223890c - are you still seeing it after pulling?

nambatee:
@joewalsh fixed 👍

PR #2615, opened 2018-08-27
Remove card shadows from older devices

joewalsh:


montehurd:
noticed one small regression - the size of the border around the cards doubled:

screen shot 2018-08-27 at 10 49 49 am

carolyn and i pixel pushed this a while ago so i'm hyper aware ;)

PR #2620, opened 2018-08-28
Fix panel button long translation text wrapping

montehurd:
https://phabricator.wikimedia.org/T202974

| Before  | After |
| screen shot 2018-08-28 at 2 54 39 am|screen shot 2018-08-28 at 2 49 08 am|


Wikipedia/Code/AutoLayoutSafeMultiLineButton.swift
```diff
@@ -0,0 +1,49 @@
...
+        guard let titleLabel = titleLabel else {
+            return super.sizeThatFits(size)
+        }
+        return sizeByAddingInsets(to: titleLabel.sizeThatFits(size))
+    }
+    
+    override func layoutSubviews() {
+        super.layoutSubviews()
+        guard let titleLabel = titleLabel else { return }
+        titleLabel.preferredMaxLayoutWidth = titleLabel.frame.width
```
joewalsh: Setting `preferredMaxLayoutWidth` likely affects the layout, so it should be done before `layoutSubviews` - I'm not sure these changes would be applied if there wasn't another layout pass after it was updated

This could be updated on bounds change by overriding bounds and updating it in the `didSet`. It would be dictated by the bounds width minus the `contentEdgeInsets`, `titleEdgeInsets`, the image width & image edge insets if an image is set (using `image(for:` not `imageView.image`). This way it's updated before the layout pass and doesn't rely on the `titleLabel.frame`

Wikipedia/Code/AutoLayoutSafeMultiLineButton.swift
```diff
@@ -0,0 +1,49 @@
...
+    }
+    
+    private func sizeByAddingInsets(to size: CGSize) -> CGSize {
+        return CGSize(
+            width: size.width + titleEdgeInsets.left + titleEdgeInsets.right,
```
joewalsh: This is missing `contentEdgeInsets` and also doesn't account for the image if it's set

Wikipedia/Code/AutoLayoutSafeMultiLineButton.swift
```diff
@@ -0,0 +1,49 @@
+
+// See multi-line button thread: https://stackoverflow.com/q/23845982/135557
+@IBDesignable
+class AutoLayoutSafeMultiLineButton: UIButton {
+    func multiLineSafeSetup () {
```
joewalsh: Could this be abstracted into a `SetupButton` class similar to `SetupView` so it's consistent & reusable?

Wikipedia/Code/AutoLayoutSafeMultiLineButton.swift
montehurd: oh cool will try!

montehurd: hmm so far no luck with this... but if i move the `preferredMaxLayoutWidth` before `super.layoutSubviews()` it works fine... would that be less likely to affect layout?

joewalsh: I'd be leery of updating anything that could trigger a `setNeedsLayout` inside of `layoutSubviews` - potentially triggers an infinite layout loop

montehurd: I added a recursion guard and `Changes welcome` label ;)

Wikipedia/Code/AutoLayoutSafeMultiLineButton.swift
montehurd: fixed!

montehurd:
@joewalsh 
only thing i notice that's a bit worse with `wmf_sizeThatFits` is layout if i set an image on the button: 

| No Image (pre and post `wmf_sizeThatFits`) | Image with pre-`wmf_sizeThatFits` commit  | Image with `wmf_sizeThatFits` |
| ------------- | ------------- | ------------- |
|screen shot 2018-08-28 at 4 47 38 pm|screen shot 2018-08-28 at 4 45 33 pm|screen shot 2018-08-28 at 4 43 12 pm|
|||^ mysterious additional top and bottom padding|

montehurd:
@joewalsh not sure if we care for now though since we're not using images in the one place we're needing to use this button class at the moment...

joewalsh:
simulator screen shot - iphone x - 2018-08-29 at 09 52 28

@montehurd  previous commit should fix the extra padding ^

PR #2622, opened 2018-08-28
Add max aspect ratio for side scrolling cell images

joewalsh:
https://phabricator.wikimedia.org/T202995

Before:
simulator screen shot - ipad air 2 - 2018-08-28 at 12 06 32

After:
simulator screen shot - ipad air 2 - 2018-08-28 at 13 29 32

### DIFF
--- a/Wikipedia/Code/SideScrollingCollectionViewCell.swift
+++ b/Wikipedia/Code/SideScrollingCollectionViewCell.swift
@@ -95,7 +95,6 @@ public class SideScrollingCollectionViewCell: CollectionViewCell, SubCellProtoco
     }
     
     public let imageViewHeight: CGFloat = 130
-    public let maxImageViewAspectRatio: CGFloat = 2.88
     public let spacing: CGFloat = 6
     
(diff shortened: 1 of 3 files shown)


nambatee:
lgtm

Wikipedia/Code/SideScrollingCollectionViewCell.swift
```diff
@@ -100,7 +104,8 @@ public class SideScrollingCollectionViewCell: CollectionViewCell, SubCellProtoco
         let widthToFit = size.width - layoutMargins.left - layoutMargins.right
         if !isImageViewHidden {
             if (apply) {
-                imageView.frame = CGRect(x: 0, y: 0, width: size.width, height: imageViewHeight)
+                let imageViewWidth = size.width - widthToFit > 50 ? widthToFit : size.width
```
montehurd: may be nice for future reference to use some sort of declarative naming for the frame bits here... something like `let saneAspectRatioFrame = `...

PR #2623, opened 2018-08-28
Fix batch edit toolbar's height

nambatee:
To avoid

simulator screen shot - iphone x - 2018-08-28 at 12 03 31


Wikipedia/Code/CollectionViewEditControllerNavigationDelegate+Extensions.swift
```diff
@@ -8,10 +8,10 @@ extension CollectionViewEditControllerNavigationDelegate where Self: UIViewContr
             batchEditToolbarViewController.view.translatesAutoresizingMaskIntoConstraints = false
             view.addSubview(batchEditToolbarViewController.view)
             batchEditToolbarViewController.didMove(toParentViewController: self)
-            
+            let height = max(view.safeAreaInsets.bottom, 60)
```
joewalsh: should this be `view.safeAreaInsets.bottom + 44`? seems like the height should never == safeAreaInsets.bottom

nambatee: in `Saved`, if we did `view.safeAreaInsets.bottom + 44` it would be `128`, which seems pretty tall 🤔 `view.safeAreaInsets.bottom` is `84` in the `Saved` tab

PR #2624, opened 2018-08-28
[BUG] Long "Saved > All articles" tab text truncated in middle

nambatee:
https://phabricator.wikimedia.org/T202950


nambatee:
modified this to be using AutoLayoutSafeMultiLineButton

PR #2629, opened 2018-08-29
fix reading list text fields jump when inserting text at beginning

montehurd:
https://phabricator.wikimedia.org/T202447

### DIFF
--- a/Wikipedia/Code/CreateReadingListViewController.swift
+++ b/Wikipedia/Code/CreateReadingListViewController.swift
@@ -110,11 +110,15 @@ class CreateReadingListViewController: WMFScrollViewController, UITextFieldDeleg
             descriptionTextField.returnKeyType = .default
         }
         if descriptionTextField.isFirstResponder {
+            let selectedRange = descriptionTextField.selectedTextRange
             descriptionTextField.resignFirstResponder()
             descriptionTextField.becomeFirstResponder()
+            descriptionTextField.selectedTextRange = selectedRange
         } else {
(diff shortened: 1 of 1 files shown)


Wikipedia/Code/CreateReadingListViewController.swift
```diff
@@ -110,11 +110,15 @@ class CreateReadingListViewController: WMFScrollViewController, UITextFieldDeleg
             descriptionTextField.returnKeyType = .default
         }
         if descriptionTextField.isFirstResponder {
+            let selectedRange = descriptionTextField.selectedTextRange
             descriptionTextField.resignFirstResponder()
```
joewalsh: I know this wasn't added in this PR, but I think this resign/become toggle is going to be prone to break in other unexpected ways down the line. Could we remove the functionality that changes the return key type altogether? Either make both of them `.next` and have it be a toggle back and forth between the fields or have them both be `.done` and show an error if they try to press done with no name entered. I think I'd lean towards them both being `.next` and the only way to create the list is with the button. If there's a vertical space issue hiding the button below the fields on smaller devices, maybe also have a "Done" as the right bar button item that is also enabled/disabled when the list can be created.

nambatee: This sounds like a question for @carolynlimadeo since originally we wanted users to be able to create lists via tapping the Create button or tapping return on the keyboard

carolynlimadeo: Thanks, @joewalsh, I agree that I'd prefer having them both be .next

PR #2631, opened 2018-08-30
Set reading lists limits

nambatee:


Wikipedia/Code/MWKDataStore.m
```diff
@@ -1596,34 +1596,64 @@ - (void)updateLocalConfigurationFromRemoteConfigurationWithCompletion:(nullable
...
+                                       if (!remoteConfigURL) {
+                                           combinedCompletion([NSError wmf_errorWithType:WMFErrorTypeInvalidRequestParameters userInfo:nil]);
+                                           return;
+                                       }
+                                       NSURLRequest *request = [NSURLRequest requestWithURL:remoteConfigURL];
+                                       if (!request) {
+                                           combinedCompletion([NSError wmf_errorWithType:WMFErrorTypeInvalidRequestParameters userInfo:nil]);
+                                           return;
+                                       }
+                                       [[[WMFSession shared] jsonDictionaryTaskWith:request
```
joewalsh: Could these requests be made in parallel instead of serially?

Wikipedia/Code/MWKDataStore.m
```diff
@@ -1583,6 +1583,31 @@ - (void)updateLocalConfigurationFromRemoteConfigurationWithCompletion:(nullable
...
+    if (!siteInfoRequest) {
+        combinedCompletion([NSError wmf_errorWithType:WMFErrorTypeInvalidRequestParameters userInfo:nil]);
+        return;
+    }
+    [[[WMFSession shared] jsonDictionaryTaskWith:siteInfoRequest
+                               completionHandler:^(NSDictionary<NSString *, id> *_Nullable siteInfo, NSURLResponse *_Nullable response, NSError *_Nullable error) {
+                                   dispatch_async(dispatch_get_main_queue(), ^{
+                                       if (error) {
+                                           DDLogError(@"Error fetching site info: %@", error);
+                                           combinedCompletion(error);
```
joewalsh: As implemented, `combinedCompletion` could be called multiple times, and is sometimes called before the site info request completes. It should be called once when everything is completed. This could be achieved with a `WMFTaskGroup` that waits for both network calls to complete, then calls `combinedCompletion`. An __block error variables could be used to hold errors from the network calls (or `WMFTaskGroup` could be extended to handle errors)

PR #2638, opened 2018-09-06
Html lead image spike

montehurd:


montehurd:
Closing per convo with @joewalsh regarding https://github.com/wikimedia/wikipedia-ios/pull/2641 which seems a much lower impact fix. One of the main complications with the HTML/CSS approach would be dealing with transition animation when loading articles from feed. Also some face detection issues with articles loaded from link taps.

PR #2643, opened 2018-09-07
Unify face detection cache internals and make its key size invariant

montehurd:
I noticed some face detection cache issues while working on the lead image html/css spike: https://github.com/wikimedia/wikipedia-ios/pull/2638

This pr:
- unifies the face detection cache which formerly handled image urls and MWKImages differently
- makes the face detection cache keys size invariant - face rects are stored in unit coords so no need to re-calculate face rect for, say, a 240px image if we already have a cached face rect for the same image at 640px
- removes unused/unneeded related bits

Confirmed `CIDetector featuresInImage` is no longer re-called when tapping on a feed item which has a face - say one from top read. Formerly it would get re-called when the full article loaded.


joewalsh:
@montehurd my only concern would be that if the detector doesn't find a face on a smaller variant, the app won't try to find the face on a larger variant. However, I wouldn't want to switch it to keep checking for faces if it caches an empty array. We could probably bring this in, monitor detection quality, and decide whether or not it needs further enhancement

montehurd:
@joewalsh 

> my only concern would be that if the detector doesn't find a face on a smaller variant, the app won't try to find the face on a larger variant

Yeah I was worried about that too. I poked around for a while it seemed that face detection was having good luck detecting faces on the smaller images we use for search & detail views, etc.

montehurd:
jenkins retest this please

PR #2650, opened 2018-09-12
Wikidata description editing back end - first pass

nambatee:
Biggest changes so far:

- `CSRFTokenOperation`
  - changed `completion`'s first parameter to be of type `Any` - right now `CSRFTokenOperation` is used by reading lists operations and wikidata operations - reading list operations are expected to complete with an array of `[String: Any]` while wikidata operations can be modeled to complete with `Decodable`
  - delegation - instead of hardcoding what happens after operation fetches a token, delegates can respond accordingly (so reading list operations can use `jsonDictionaryTask` and wikidata operations can use `jsonCodableTask`)
  - added `queryParameters` - before we created `queryParameters` with `csrf_token` as the key and the fetched token as the value, now we can pass params

- `Authentication`
  - stopped storing `host` in `KeychainCredentials` - now we have more than 1 host to log into
  - updated `logout` to log out of all available hosts

- `TokenFetcher`
  - retrying with login - if we get an anonymous token while the app claims that we're logged in, we'll try to log in and fetch again

All suggestions how to improve welcome 🧐


montehurd:
cool improvements! will do deeper review once my UI bits solidify a bit more.

one quick thing to double-check if you haven't already - ensure **article** editing still plays nice (expected user credited in article revision history) when logged in or anonymous

Wikipedia/Code/WMFAuthenticationManager.swift
```diff
@@ -183,19 +168,32 @@ public class WMFAuthenticationManager: NSObject {
...
-        }, failure: { (_, error) in
-            // ...but if "action=logout" fails we *still* want to clear local login settings, which still effectively logs the user out.
-            DDLogDebug("Failed to log out, delete login tokens and other browser cookies: \(error)")
+    @objc public func logout(completion: @escaping () -> Void = {}) {
+        let reset = {
+            DDLogDebug("Deleted login tokens and other browser cookies")
             self.resetLocalUserLoginSettings()
+        }
+        let taskGroup = WMFTaskGroup()
+        for loginSite in LoginSite.allCases {
```
joewalsh: On the web, it looks like logging out of either wikidata or any centralauth wikipedia site logs the user out of all projects - is there something else causing the stored cookies to stick around in the app that requires logging out of both sites?

Wikipedia/Code/WikidataDescriptionEditingController.swift
```diff
@@ -0,0 +1,126 @@
...
+    var errorDescription: String? {
+        return info
+    }
+}
+
+extension WikidataAPIResult {
+    var succeeded: Bool {
+        return success == 1
+    }
+}
```
joewalsh: ^ this is a nice setup for constants and results

Wikipedia/Code/WMFAuthenticationManager.swift
nambatee: good point, I assumed that I had to do it for each site, will fix 🙌

WMF Framework/ReadingListsAPIController.swift
```diff
@@ -546,3 +546,40 @@ class ReadingListsAPIController: NSObject {
         }
     }
 }
+
+extension ReadingListsAPIController: CSRFTokenOperationDelegate {
+    func CSRFTokenOperationDidFetchToken(_ operation: CSRFTokenOperation, token: WMFAuthToken, context: CSRFTokenOperationContext, completion: @escaping () -> Void) {
```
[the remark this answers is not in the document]
nambatee: good idea, will try 🙌

Wikipedia/Code/WikidataDescriptionEditingController.swift
```diff
@@ -0,0 +1,98 @@
...
+    public func isBlacklisted(_ languageCode: String) -> Bool {
+        guard !blacklistedLanguages.isEmpty else {
+            return false
+        }
+        return blacklistedLanguages.contains(languageCode)
+    }
+
+    @objc(publishNewWikidataDescription:forArticle:completion:)
+    public func publish(newWikidataDescription: String, for article: MWKArticle, completion: @escaping (_ error: Error?) -> Void) {
+        guard let title = article.displaytitle,
```
montehurd: `displaytitle` can contain html so i think we need `article.url.wmf_title` here...

noticed when trying to edit description of https://en.wikipedia.org/wiki/Ashurbanipal_(sculpture)
`displaytitle` : `<i>Ashurbanipal</i> (sculpture)`
`article.url.wmf_title` : `Ashurbanipal (sculpture)`

Wikipedia/Code/WMFAuthenticationManager.swift
```diff
@@ -127,8 +104,17 @@ public class WMFAuthenticationManager: NSObject {
      *  @param userAlreadyLoggedInHandler     The handler called if a user was found to already be logged in
      *  @param failure     The handler for any errors
      */
-    @objc public func loginWithSavedCredentials(success:@escaping WMFAccountLoginResultBlock, userAlreadyLoggedInHandler:@escaping WMFCurrentlyLoggedInUserBlock, failure:@escaping WMFErrorHandler){
+    public func loginWithSavedCredentials(_ loginURL: URL? = LoginSite.wikipedia.url, success: @escaping WMFAccountLoginResultBlock, userAlreadyLoggedInHandler: @escaping WMFCurrentlyLoggedInUserBlock, failure: @escaping WMFErrorHandler, completion: @escaping () -> Void = {}) {
         
+        defer {
+            completion()
```
joewalsh: Is the intent that this be called after whichever other block (success, alreadyLoggedIn, failure) is called?

nambatee: yes

joewalsh: The code inside `defer { }` will be executed when that function returns, it doesn’t take into account the various async calls. So as it is currently, the `completion` is called before  any of the other blocks are called

This might be a good opportunity to turn this into a method with a single completion block with a result type:

```
    enum LoginResult {
        case success(_: WMFAccountLoginResult)
        case alreadyLoggedIn(_: WMFCurrentlyLoggedInUser)
        case failure(_: Error)
    }

    ... completion: @escaping (LoginResult) -> Void ...
```

Wikipedia/Code/WMFAuthTokenFetcher.swift
```diff
@@ -67,9 +78,48 @@ public class WMFAuthTokenFetcher: NSObject {
...
+            self.fetchToken(ofType: context.tokenType, siteURL: context.siteURL, success: context.success, failure: context.failure)
+        }
+        WMFAuthenticationManager.sharedInstance.loginWithSavedCredentials(context.siteURL, success: { (_) in
+            fetchToken()
+        }, userAlreadyLoggedInHandler: { (_) in
+            fetchToken()
+        }, failure: { (error) in
+            context.failure(error)
+        }) {
+            self.isRetryingLogin = false
```
joewalsh: There might need to be some assurance that this is thread safe - maybe a main thread assert above and an async to the main thread here or a lock or an access queue

Wikipedia/Code/WMFAuthenticationManager.swift
nambatee: Thanks, I will update the completion part. Given that `WMFAuthenticationManager` relies on `failure` and `success` blocks in many places (methods that are called inside this method), it seems like moving to a single completion would be a bigger change - could it be submitted as a follow-on?

joewalsh: Yeah, fine to do as a follow on or not at all, just wanted to note it if it was an opportune time to consolidate

Wikipedia/Code/WMFAuthTokenFetcher.swift
```diff
@@ -111,14 +112,17 @@ public class WMFAuthTokenFetcher: NSObject {
...
         }
-        WMFAuthenticationManager.sharedInstance.loginWithSavedCredentials(context.siteURL, success: { (_) in
-            fetchToken()
-        }, userAlreadyLoggedInHandler: { (_) in
-            fetchToken()
-        }, failure: { (error) in
-            context.failure(error)
-        }) {
-            self.isRetryingLogin = false
+        DispatchQueue.main.async {
```
joewalsh: This asyncs the `loginWithSavedCredentials` call, but not the callbacks. `self.isRetryingLogin = false` in the callbacks below could still be on any thread

nambatee: https://github.com/wikimedia/wikipedia-ios/pull/2650/commits/ecd79d425d82a99d0f480777233423a8f67a0e19 is this what you mean? 🧐

joewalsh: @nambatee yes, but you can also delete the old `DispatchQueue.main.async` { above the entire method

PR #2659, opened 2018-09-24
Replace success, failure login blocks with single completion using LoginResult 

nambatee:
Branched off #2650, addresses https://github.com/wikimedia/wikipedia-ios/pull/2650#discussion_r219612237

Relevant commits: 10e6620 - 24040c8


Wikipedia/Code/WMFAuthenticationManager.swift
```diff
@@ -48,26 +48,40 @@ public class WMFAuthenticationManager: NSObject {
...
+                performCompletionOnTheMainThread()
+            case .alreadyLoggedIn(let result):
+                DDLogDebug("\n\nUser \(result.name) is already logged in.\n\n")
+                performCompletionOnTheMainThread()
+            case .failure(let error):
+                DDLogDebug("\n\nloginWithSavedCredentials failed with error \(error).\n\n")
+                performCompletionOnTheMainThread()
+                DispatchQueue.main.async {
+                    completion(.failure(error))
+                }
```
joewalsh: It looks like the completion is called twice here (once in `performCompletionOnTheMainThread` and again for the failure) - maybe `performCompletionOnTheMainThread` should be removed entirely and below the switch should be a single `DispatchQueue.main.async { completion(loginResult) }`

Wikipedia/Code/WMFAccountCreationViewController.swift
```diff
@@ -226,11 +223,14 @@ class WMFAccountCreationViewController: WMFScrollViewController, WMFCaptchaViewC
...
-            self.setViewControllerUserInteraction(enabled: true)
-            self.enableProgressiveButtonIfNecessary()
-            WMFAlertManager.sharedInstance.showErrorAlert(error as NSError, sticky: true, dismissPreviousAlerts: true, tapCallBack: nil)
-        })
+            case .failure(let error):
+                self.setViewControllerUserInteraction(enabled: true)
+                self.enableProgressiveButtonIfNecessary()
+                WMFAlertManager.sharedInstance.showErrorAlert(error as NSError, sticky: true, dismissPreviousAlerts: true, tapCallBack: nil)
+            default:
+                break
```
montehurd: would it be of any utility to add an assert for the default case here? or do we not care if result is ever `alreadyLoggedIn` here?

Wikipedia/Code/WMFAuthenticationManager.swift
```diff
@@ -48,26 +48,29 @@ public class WMFAuthenticationManager: NSObject {
...
+
+    public typealias LoginResultHandler = (LoginResult) -> Void
     
-    public func attemptLogin(_ loginURL: URL? = LoginSite.wikipedia.url, completion: @escaping () -> Void = {}, failure: @escaping (_ error: Error) -> Void = {_ in }) {
-        let performCompletionOnTheMainThread = {
-            DispatchQueue.main.async {
-                completion()
+    public func attemptLogin(_ loginURL: URL? = LoginSite.wikipedia.url, completion: @escaping LoginResultHandler) {
+        self.loginWithSavedCredentials(loginURL) { (loginResult) in
+            switch loginResult {
```
montehurd: much nicer!

Wikipedia/Code/WMFChangePasswordViewController.swift
```diff
@@ -101,23 +101,20 @@ class WMFChangePasswordViewController: WMFScrollViewController, Themeable {
         enableProgressiveButton(false)
         WMFAlertManager.sharedInstance.dismissAlert()
                 
-        guard let userName = userName else {
+        guard let userName = userName,
+            let password = passwordField.text,
+            let retypePassword = retypeField.text else {
```
montehurd: i tested changing a pwd and still works fine, but may want assert here just in case

Wikipedia/Code/WMFLoginViewController.swift
```diff
@@ -153,54 +153,60 @@ class WMFLoginViewController: WMFScrollViewController, UITextFieldDelegate, WMFC
...
-            
-            if let start = self.startDate {
-                LoginFunnel.shared.logSuccess(timeElapsed: fabs(start.timeIntervalSinceNow))
-            } else {
-                assertionFailure("startDate is nil; startDate is required to calculate timeElapsed")
-            }
-        
-        }, failure: { error in
+        guard let username = usernameField.text, let password = passwordField.text else {
+            return
```
montehurd: * same assert comment as above

Wikipedia/Code/WMFAccountCreationViewController.swift
montehurd: ah gotcha. would it be useful to assert in case someone changes that in the future?

montehurd:
@nambatee were you able to try 2FA with these changes? lgtm if so

nambatee:
testing 2FA @montehurd 🧐

PR #2660, opened 2018-09-24
Wikidata description API tweaks

nambatee:
* persists blacklisted languages as `WMFKeyValue`
* fixes `isWikidataDescriptionEditable` by adding `!`

### DIFF
--- a/Wikipedia/Code/WikidataDescriptionEditingController.swift
+++ b/Wikipedia/Code/WikidataDescriptionEditingController.swift
@@ -40,6 +40,7 @@ enum WikidataPublishingError: LocalizedError {
 
 @objc public final class WikidataDescriptionEditingController: NSObject {
     weak var viewContext: NSManagedObjectContext?
+    private let semaphore = DispatchSemaphore(value: 1)
 
     @objc public init(with viewContext: NSManagedObjectContext) {
         self.viewContext = viewContext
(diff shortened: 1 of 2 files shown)


Wikipedia/Code/WikidataDescriptionEditingController.swift
```diff
@@ -39,14 +39,44 @@ enum WikidataPublishingError: LocalizedError {
...
+                return fallback
+        }
+        return value
+    }
 
     @objc public func setBlacklistedLanguages(_ blacklistedLanguagesFromRemoteConfig: Array<String>) {
-        blacklistedLanguages = Set(blacklistedLanguagesFromRemoteConfig)
+        assert(Thread.isMainThread)
+        assert(viewContext != nil)
+        semaphore.wait()
```
joewalsh: If this & the above are always on the main thread, the semaphore shouldn't be needed

Wikipedia/Code/WikidataDescriptionEditingController.swift
```diff
@@ -39,14 +39,35 @@ enum WikidataPublishingError: LocalizedError {
...
+
+    private let BlacklistedLanguagesKey = "WMFWikidataDescriptionEditingBlacklistedLanguagesKey"
+    private var blacklistedLanguages: NSSet {
+        assert(Thread.isMainThread)
+        let fallback = NSSet(set: ["en"])
+        guard
+            let viewContext = viewContext,
+            let keyValue = viewContext.wmf_keyValue(forKey: BlacklistedLanguagesKey),
+            let value = keyValue.value as? NSSet else {
+                assertionFailure("Failed to retrieve WMFKeyValue for key \(BlacklistedLanguagesKey), returning fallback object \(fallback)")
```
joewalsh: Is this assertion failure necessary if there's a fallback? Since the value isn't set by a library migration, it won't be there until the initial remote config check completes

nambatee: yeah that's a good point 👍

PR #2663, opened 2018-09-26
Edit Notifications

nambatee:
Things to look out for:

- Thread safety
- Handling Core Data failures

Doing

- [x] Fetch and save notifications that appeared in the account after first authorized wikidata description edit
- [x] Stop polling after 24h
- [x] UI
- [x] Localizations

Merge after #2676 

Testing:

Prerequisites:
1. You have to make at least one authorized (logged in) Wikidata description edit

If you have any old `reverted` notifications in your account:
1. Open the affected article on the device
2. Comment out age validation on line 163-165 in `RemoteNotificationsModelController`

I if you don't have any old `reverted` notifications, you can use the test account in 1P


Wikipedia/Code/WikidataDescriptionEditingController.swift
```diff
@@ -98,7 +97,23 @@ enum WikidataPublishingError: LocalizedError {
...
+            }
+            guard let result = result else {
+                completion(WikidataPublishingError.apiResultNotParsedCorrectly)
+                return
+            }
+
+            completion(result.error)
+
+            // TODO: Check if request was sent with an authorized token instead of checking if user is logged in.
+            let isLoggedIn = WMFAuthenticationManager.sharedInstance.isLoggedIn(at: WMFAuthenticationManager.LoginSite.wikidata)
```
nambatee: Given this CSRFOperation setup, is there a way to know if the request was authorized with a non-anon token in this completion? Looks like the result is the same for anon & non-anon token so can't get it from there. I couldn't think of a good way so resorted to checking if user is logged in

WMF Framework/Remote Notifications/Model/RemoteNotificationsModelController.swift
```diff
@@ -0,0 +1,221 @@
...
+            abort() // TODO
+        }
+        let container = NSPersistentContainer(name: modelName, managedObjectModel: model)
+        let sharedAppContainerURL = FileManager.default.wmf_containerURL()
+        let remoteNotificationsStorageURL = sharedAppContainerURL.appendingPathComponent(modelName)
+        let description = NSPersistentStoreDescription(url: remoteNotificationsStorageURL)
+        container.persistentStoreDescriptions = [description]
+        container.loadPersistentStores { (storeDescription, error) in
+            if let error = error {
+                fatalError("Unexpected Core Data error occurred while loading persistent stores: \(error)") // TODO
```
nambatee: Not sure what to do in this case, I could post a notification and stop the controller if this fails

WMF Framework/Remote Notifications/Operations/RemoteNotificationsOperation.swift
```diff
@@ -0,0 +1,16 @@
...
+
+    init(with apiController: RemoteNotificationsAPIController, modelController: RemoteNotificationsModelController) {
+        self.apiController = apiController
+        self.modelController = modelController
+        self.managedObjectContext = modelController.managedObjectContext
+        super.init()
+    }
+
+    override var shouldExecute: Bool {
+        return WMFAuthenticationManager.sharedInstance.isLoggedIn
```
joewalsh: Could this be achieved by stopping the controller when the user isn't logged in? That way the operations are never created in first place, and it eliminates the need for the `shouldExecute` shortcut in `AsyncOperation`

Wikipedia/Code/WikidataDescriptionEditingController.swift
joewalsh: Could the completion block pass back the token as well? The token model object might have to be updated with a property that indicates whether it's anon or not

WMF Framework/Remote Notifications/Model/RemoteNotificationsModelController.swift
joewalsh: Ensuring the controller is stopped and stays stopped seems like a good plan. I would also suggest removing the fatalError so the app doesn't crash

WMF Framework/Remote Notifications/Model/RemoteNotificationsModelController.swift
```diff
@@ -0,0 +1,221 @@
...
+            } else {
+                print(storeDescription)
```
joewalsh: This should be `DDLogDebug`

Wikipedia/Code/WMFAppViewController.m
```diff
@@ -1927,6 +1975,113 @@ - (void)entriesLimitReachedWithNotification:(NSNotification *)notification {
...
+    for (RemoteNotification *editRevertedNotification in editRevertedNotifications) {
+        WMFArticle *article = [self.dataStore fetchArticleWithWikidataID:editRevertedNotification.affectedPageID];
+        if (article) {
+            // Hide notifications without article context
+            [filteredNotifications addObject:editRevertedNotification];
+            // We don't create notifications without ids so the objects that we're adding won't be nil
+            [filteredNotificationsIDs addObject:editRevertedNotification.id];
+            [articles addObject:article];
+            [articleKeys addObject:article.key];
+        }
```
nambatee: For our current use case, if we don't have article context for a given notification, my assumption is that we probably don't want to show it to the user - if that's correct, I'm not sure what to do with those notifications. Delete them and then re-save them when we fetch them from the server, check if there's article context again, repeat? Or maybe it'd be better to check for the article context when we fetch the notifications, instead of waiting until we're about to show them to the user

Wikipedia/Localizations/qqq.lproj/Localizable.strings
```diff
@@ -506,6 +506,16 @@
 "relogin-failed-subtitle" = "Subtitle for letting user know they are no longer logged in.";
 "relogin-failed-title" = "Title for letting user know they are no longer logged in.";
 "request-feedback-on-error" = "Displayed to beta users when they encounter an error we'd like feedback on";
+"reverted-edit-back-to-article-button-title" = "Title for button that allows the user to go back to the article they edited";
+"reverted-edit-back-to-history-button-title" = "Title for button that allows the user to go back to reading history";
+"reverted-edit-notification-read-more-action-title" = "Title on the 'Read more' action button";
+"reverted-edit-possible-reasons" = "List of possible reasons describing why an edit might have been reverted. $1 is replaced with a link to view editing guidelines";
+"reverted-edit-possible-reasons-subtitle" = "Subtitle leading to an explanation why an edit was reverted.";
+"reverted-edit-thanks-for-editing-title" = "Title thanking the user for contributing to Wikipedia";
+"reverted-edit-title" = "Title for the view explaning why an edit was reverted.";
```
montehurd: typo in `explaning`

Wikipedia/Localizations/qqq.lproj/Localizable.strings
```diff
@@ -506,6 +506,16 @@
...
+"reverted-edit-view-guidelines-text" = "Text for link for viewing editing guidelines";
+"reverted-edits-possible-reasons-subtitle" = "Subtitle leading to an explanation why an edit was reverted.";
+"reverted-edits-title" = "Title for the view explaning why edits were reverted.";
```
montehurd: same typo

WMF Framework/Remote Notifications/Model/RemoteNotification+CoreDataProperties.swift
```diff
@@ -0,0 +1,22 @@
...
+    @NSManaged public var affectedPageID: String?
+    @NSManaged public var categoryString: String?
+    @NSManaged public var date: NSDate?
+    @NSManaged public var id: String?
+    @NSManaged public var message: String?
+    @NSManaged public var typeString: String?
+    @NSManaged public var wiki: String?
+    @NSManaged public var wasRead: Bool
+    @NSManaged public var wasSeen: Bool
+    @NSManaged public var isExcluded: Bool
```
joewalsh: would it make sense to combine these three booleans into a state enum?

nambatee: yes!

WMF Framework/Remote Notifications/RemoteNotificationsAPIController.swift
```diff
@@ -0,0 +1,203 @@
...
+                assertionFailure()
+                completion(MarkReadError.unknown)
+                return
+            }
+            completion(result.error)
+        }
+    }
+
+    private func request<T: Decodable>(_ queryParameters: Query.Parameters?, method: Session.Request.Method = .get, completion: @escaping (T?, URLResponse?, Bool?, Error?) -> Void) {
+        let _ = Session.shared.requestWithCSRF(type: CSRFTokenJSONDecodableOperation.self, scheme: NotificationsAPI.scheme, host: NotificationsAPI.host, path: NotificationsAPI.path, method: method, queryParameters: queryParameters, bodyEncoding: .form, tokenContext: CSRFTokenOperation.TokenContext(tokenName: "token", tokenPlacement: .body, shouldPercentEncodeToken: true), completion: completion)
```
joewalsh: shouldn't the CSRF token only be required for mutating requests (POST, PUT, PATCH, etc)?

WMF Framework/Remote Notifications/Operations/RemoteNotificationsMarkAsReadOperation.swift
```diff
@@ -0,0 +1,21 @@
+class RemoteNotificationsMarkAsReadOperation: RemoteNotificationsOperation {
+    override func execute() {
+        DispatchQueue.main.async {
```
joewalsh: What's the reason for the main queue async here and in RemoteNotificationsFetchOperation?

nambatee: don't I need to kick off operations on private queues from the main queue?

WMF Framework/Remote Notifications/RemoteNotificationsAPIController.swift
nambatee: Getting all notifications requires a token

WMF Framework/Remote Notifications/Operations/RemoteNotificationsMarkAsReadOperation.swift
joewalsh: Not necessarily. We're doing it elsewhere where it triggers creation of a background moc from the viewContext, which we want to limit to occurring on the main thread. In this case, since the moc exists already, it should be unnecessary.

nambatee: Ok, makes sense 👍

WMF Framework/Remote Notifications/Model/RemoteNotification+CoreDataClass.swift
```diff
@@ -15,6 +15,18 @@ import CoreData
     }
 }
 
+public struct RemoteNotificationState: OptionSet {
+    public let rawValue: Int16
+
+    public static let wasSeen = RemoteNotificationState(rawValue: 1 << 0)
+    public static let wasRead = RemoteNotificationState(rawValue: 1 << 1)
+    public static let isExcluded = RemoteNotificationState(rawValue: 1 << 2)
+
```
nambatee: When user sees a notification, the state will be asynchronously set to **wasSeen** ( call A: https://github.com/wikimedia/wikipedia-ios/pull/2663/files#diff-82cb2edf849f93541737d9c5c5ff2cfeR1677), if user reads through, the state will be asynchronously set to **wasRead** (call B: https://github.com/wikimedia/wikipedia-ios/pull/2663/files#diff-82cb2edf849f93541737d9c5c5ff2cfeR1709 or https://github.com/wikimedia/wikipedia-ios/pull/2663/files#diff-82cb2edf849f93541737d9c5c5ff2cfeR1712 ). Given that both operations happen asynchronously, I think there could be a situation where call B finishes before call A, which would result in us not marking the notification as read on the server (and keeping It around for no reason)

joewalsh: do both changes happen on the private queue of the moc?

WMF Framework/Remote Notifications/RemoteNotificationsOperationsController.swift
```diff
@@ -122,3 +81,91 @@ class RemoteNotificationsOperationsController {
         }
     }
 }
+
+final class RemoteNotificationsOperationsTimeController {
+    weak var viewContext: NSManagedObjectContext?
```
nambatee: Is it better to keep this value on the `viewContext` or add `WMFKeyValue` to `RemoteNotifications.xcdatamodeld` (like in `EventLogging.xcdatamodeld`) and use the private context created by `RemoteNotificationsModelController`?

WMF Framework/Remote Notifications/Model/RemoteNotification+CoreDataClass.swift
nambatee: yes, inside `perform`

joewalsh: that should ensure the calls happen serially

WMF Framework/Remote Notifications/Model/RemoteNotification+CoreDataClass.swift
```diff
@@ -15,6 +15,18 @@ import CoreData
     }
 }
 
+public struct RemoteNotificationState: OptionSet {
```
joewalsh: could these be modeled as discrete states? it seems like if a notification is excluded, it'll never be seen or read. also, it seems like read implies that it was seen. The states I'm imagining are `ignored` -> `presented` -> `read`

WMF Framework/Remote Notifications/RemoteNotificationsAPIController.swift
joewalsh: https://www.mediawiki.org/w/api.php?notwikis=wikidatawiki%7Cenwiki&notlimit=max&action=query&notformat=model&meta=notifications&format=json&formatversion=2&notfilter=!read seems to work for me without a CSRF token

nambatee: maybe it's the cookies, will check 🧐

WMF Framework/Remote Notifications/RemoteNotificationsOperationsController.swift
joewalsh: It'd be better to use the private context

WMF Framework/Remote Notifications/RemoteNotificationsOperationsController.swift
```diff
@@ -0,0 +1,178 @@
+class RemoteNotificationsOperationsController {
+    private let apiController: RemoteNotificationsAPIController
+    private let modelController: RemoteNotificationsModelController?
+    private let deadlineController: RemoteNotificationsOperationsDeadlineController?
+
+    private let syncRepeatingTime: Int = 15
```
montehurd: Does this mean it's still checking every 15 seconds? Oh, related, should this polling be reachability aware?

montehurd: (on both counts can be follow-ups as needed)

PR #2676, opened 2018-10-11
Show wikidata editor for non-local descriptions

nambatee:
Description can come from 3 different sources:

- `local`
- `central`
- none (`descriptionsource` is not in the response, article has no description)

---

Here's what I'm understanding:

If a description comes from a local source, we want to show wikitext.
If a description comes from a central source, we want to show wikidata editor.
If there is no description source, we want to show wikidata editor?

So in summary, **unless** a description comes from a local source, we want to show a wikidata editor?

If the above is correct, what's the point of having a list of blacklisted languages? 🤔

---

Or maybe if there is no description source and the language is blacklisted, we want to show wikitext?

---

Sample articles:

*Greek language* on English has a **central** description
*Oceanography* on English has a **local** description
*Hlersu language* on English has **no** description (no source)

### DIFF
--- a/Wikipedia/Code/MWKArticle.m
+++ b/Wikipedia/Code/MWKArticle.m
@@ -30,6 +30,7 @@ static MWKArticleSchemaVersion const MWKArticleCurrentSchemaVersion = MWKArticle
 @property (readwrite, assign, nonatomic) BOOL editable;                   // required
 @property (readwrite, assign, nonatomic, getter=isMain) BOOL main;
 @property (readwrite, strong, nonatomic) NSNumber *revisionId;
+@property (readwrite, strong, nonatomic) NSNumber *descriptionSourceNumber; // optional
 
 @property (readwrite, nonatomic) NSInteger ns; //optional, defaults to 0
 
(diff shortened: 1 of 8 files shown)


Wikipedia/Code/MWKArticle.m
```diff
@@ -97,7 +98,7 @@ - (BOOL)isEqual:(id)object {
...
+    return WMF_EQUAL(self.url, isEqual:, other.url) && WMF_EQUAL(self.lastmodified, isEqualToDate:, other.lastmodified) && WMF_IS_EQUAL(self.lastmodifiedby, other.lastmodifiedby) && WMF_EQUAL(self.displaytitle, isEqualToString:, other.displaytitle) && WMF_EQUAL(self.protection, isEqual:, other.protection) && WMF_EQUAL(self.thumbnailURL, isEqualToString:, other.thumbnailURL) && WMF_EQUAL(self.imageURL, isEqualToString:, other.imageURL) && WMF_EQUAL(self.revisionId, isEqualToNumber:, other.revisionId) && self.articleId == other.articleId && self.languagecount == other.languagecount && self.isMain == other.isMain && self.sections.count == other.sections.count && self.descriptionSourceNumber == self.descriptionSourceNumber;
```
joewalsh: `==` in obj-c only compares the pointers, so this is true only if they're the exact same object, not the same number. Also, it's `self.` here on both sides instead of `self.` and `other.` It looks like `WMF_EQUAL(self.descriptionSourceNumber, isEqualToNumber:, other.descriptionSourceNumber)` handles nil cases and equality of the underlying numbers properly

montehurd:
https://phabricator.wikimedia.org/T206786

PR #2681, opened 2018-10-15
Support landscape on description edit intro

montehurd:
https://phabricator.wikimedia.org/T206788

@carolynlimadeo and i pixel pushed the landscape layout - came up with a side-by-side layout which plays reasonably well on the range of supported screen sizes. For example, below are screenshots on landscape on iPhone SE:

screen shot 2018-10-15 at 1 29 51 pm

screen shot 2018-10-15 at 1 30 36 pm

It's not perfect, but we're slightly constrained by some of the `containerizing` issues (which text should scroll vs which parts shouldn't at larger sizes etc) and design requirements (images on both panels having same top edge etc).


Wikipedia/Code/DescriptionWelcomeInitialViewController.swift
```diff
@@ -32,15 +32,7 @@ class DescriptionWelcomeInitialViewController: UIViewController, Themeable {
     }
     
     override var supportedInterfaceOrientations : UIInterfaceOrientationMask {
```
joewalsh: I think this property can be removed entirely now

montehurd: removed

PR #2684, opened 2018-10-16
Consolidate timers, periodic worker & background fetcher logic

joewalsh:
https://phabricator.wikimedia.org/T207086


WMF Framework/EventLoggingService.swift
```diff
@@ -12,10 +13,11 @@ public class EventLoggingService : NSObject, URLSessionDelegate {
         static let lastLoggedSnapshot = "WMFLastLoggedSnapshot"
         static let appInstallDate = "AppInstallDate"
         static let loggedDaysInstalled = "DailyLoggingStatsDaysInstalled"
+        static let lastSuccessfulPost = "LastSuccessfulPost"
     }
     
     private var pruningAge: TimeInterval = 60*60*24*30 // 30 days
-    private var sendImmediatelyOnWWANThreshhold: TimeInterval = 30
+    private var sendOnWWANThreshhold: TimeInterval = 24 * 60 * 60
```
montehurd:
```suggestion
    private var sendOnWWANThreshold: TimeInterval = 24 * 60 * 60
```

WMF Framework/EventLoggingService.swift
```diff
@@ -310,19 +164,30 @@ public class EventLoggingService : NSObject, URLSessionDelegate {
                 } catch let error {
                     DDLogError(error.localizedDescription)
                 }
+                
+                var wifiOnly = true
+                if let lastSuccessNumber = moc.wmf_keyValue(forKey: Key.lastSuccessfulPost)?.value as? NSNumber {
+                    let now = CFAbsoluteTimeGetCurrent()
+                    let interval = now - CFAbsoluteTime(lastSuccessNumber.doubleValue)
+                    if interval > self.sendOnWWANThreshhold {
```
montehurd:
```suggestion
                    if interval > self.sendOnWWANThreshold {
```

WMF Framework/EventLoggingService.swift
```diff
@@ -286,18 +127,39 @@ public class EventLoggingService : NSObject, URLSessionDelegate {
...
+    private func tryPostEvents(_ completion: (() -> Void)? = nil) {
         let operation = AsyncBlockOperation { (operation) in
             let moc = self.managedObjectContext
             moc.perform {
+                let pruneFetch = NSFetchRequest<NSFetchRequestResult>(entityName: "WMFEventRecord")
+                pruneFetch.returnsObjectsAsFaults = false
+                
+                let pruneDate = Date().addingTimeInterval(-(self.pruningAge)) as NSDate
+                pruneFetch.predicate = NSPredicate(format: "(recorded < %@) OR (posted != nil) OR (failed == TRUE)", pruneDate)
+                let delete = NSBatchDeleteRequest(fetchRequest: pruneFetch)
```
nambatee: What's the reason for choosing `NSBatchDeleteRequest` over `wmf_batchProcessObjects` (as used in `ReadingListsSyncOperation` here https://github.com/wikimedia/wikipedia-ios/blob/b726e6603fe274d8748206b2d82727be54f5e069/WMF%20Framework/ReadingListsSyncOperation.swift#L255)?

WMF Framework/EventLoggingService.swift
```diff
@@ -418,12 +289,16 @@ public class EventLoggingService : NSObject, URLSessionDelegate {
...
-
-            let task = urlSession.dataTask(with: request, completionHandler: { (_, response, error) in
+            let session = onlyWiFi ? self.session.wifiOnlyURLSession : self.session.defaultURLSession
+            let task = session.dataTask(with: request, completionHandler: { (_, response, error) in
                 guard error == nil,
                     let httpResponse = response as? HTTPURLResponse,
                     httpResponse.statusCode / 100 == 2 else {
-                        completion(EventLoggingError.generic)
+                        if let error = error as NSError?, error.domain == NSURLErrorDomain {
+                            completion(EventLoggingError.network)
```
nambatee: Would it make sense to make a wrapper around this dataTask method inside our `Session` so that we can check for these generic errors in one place? `WikidataFetcher` is also calling this method and defining its own generic errors

WMF Framework/EventLoggingService.swift
joewalsh: `NSBatchDeleteRequest` is more efficient as it deletes the items without populating them into memory. However, if I'm remembering correctly, it doesn't process object relationship changes or send object change notifications (if you have the deleted objects in memory already, they're invalid). In this case, since we're only ever fetching the events and not relying on change notifications, it's fine. For the reading lists, we want to make sure the related objects are updated and that we get change notifications. For what it's worth, this change is just moving the event prune code that was in existence, not a new implementation.

WMF Framework/EventLoggingService.swift
```diff
@@ -310,19 +172,30 @@ public class EventLoggingService : NSObject, URLSessionDelegate {
...
+                } else {
+                    operation.finish()
                 }
             }
         }
         operationQueue.addOperation(operation)
+        guard let completion = completion else {
+            return
+        }
+        operationQueue.addOperation(completion)
```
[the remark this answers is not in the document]
joewalsh: good catch, would be good to add a dependency

WMF Framework/EventLoggingService.swift
joewalsh: Down the line, yes, but not within the scope of this PR as this is currently called directly on the `URLSession`, not the `Session`

WMF Framework/EventLoggingService.swift
nambatee: Makes sense, thanks 👍

WMF Framework/EventLoggingService.swift
```diff
@@ -596,3 +471,17 @@ public class EventLoggingService : NSObject, URLSessionDelegate {
...
+extension EventLoggingService: PeriodicWorker {
+    public func doPeriodicWork(_ completion: @escaping () -> Void) {
+        tryPostEvents(completion)
+    }
+}
+
+extension EventLoggingService: BackgroundFetcher {
+    public func performBackgroundFetch(_ completion: @escaping (UIBackgroundFetchResult) -> Void) {
+        doPeriodicWork {
+            completion(.noData)
```
nambatee: I might not be understanding this right 🤔 but does this mean that `EventLoggingService` fetcher periodically tells the system that it didn't do anything (aka fetched no data) in the background? If a controller does nothing in the background, do we still want to make it a `BackgroundFetcher`?

joewalsh: It's sending up events (via it's own `doPeriodicWork` method), but reporting `.noData` because technically nothing was downloaded. We could report `.newData` if events were successfully sent, though maybe we shouldn't make this a background task at all if it's uploading instead of downloading. https://developer.apple.com/library/archive/documentation/iPhone/Conceptual/iPhoneOSProgrammingGuide/BackgroundExecution/BackgroundExecution.html#//apple_ref/doc/uid/TP40007072-CH4-SW56

PR #2688, opened 2018-10-17
Use new PeriodicWorker & BackgroundFetcher in remote notifications

nambatee:


WMF Framework/Remote Notifications/RemoteNotificationsOperationsController.swift
```diff
@@ -52,6 +56,9 @@ class RemoteNotificationsOperationsController: NSObject {
             stop()
             return
         }
+        guard UserDefaults.wmf.didMakeAuthorizedTitleDescriptionEdit else {
+            return
```
joewalsh: The `completion` should be called anywhere there's an early return (here and above)

WMF Framework/Remote Notifications/RemoteNotificationsOperationsController.swift
```diff
@@ -61,19 +68,17 @@ class RemoteNotificationsOperationsController: NSObject {
             }
             let markAsReadOperation = RemoteNotificationsMarkAsReadOperation(with: apiController, modelController: modelController)
             let fetchOperation = RemoteNotificationsFetchOperation(with: apiController, modelController: modelController)
+            let completionOperation = BlockOperation(block: completion)
             fetchOperation.addDependency(markAsReadOperation)
+            completionOperation.addDependency(fetchOperation)
             operationQueue.addOperation(markAsReadOperation)
             operationQueue.addOperation(fetchOperation)
+            operationQueue.addOperation(completionOperation)
```
joewalsh: also missing a call of the `completion` if it's after the deadline

Wikipedia/Code/WikidataDescriptionEditingController.swift
```diff
@@ -80,10 +81,7 @@ enum WikidataPublishingError: LocalizedError {
             completion(result.error)
 
             if let authorized = authorized, authorized, result.error == nil {
-                DispatchQueue.main.async {
-                    NotificationCenter.default.post(name: WikidataDescriptionEditingController.DidMakeAuthorizedWikidataDescriptionEditNotification, object: nil)
```
joewalsh: It'd be nice to consolidate the stored state for notifications, right now it's across `UserDefaults` and its own Core Data stack. Could this still post `.didMakeAuthorizedWikidataDescriptionEditNotification` and when that happens, the notifications controller writes the deadline date to its own Core Data stack? That way, the state is only in one place. The check for whether or not to make the request for new notifications would be if the date keyValue exists and is after the current date.

nambatee: That makes sense, I was thinking it might make sense to store it in defaults if in the future we wanted to show something based on whether the user made an authorized edit

PR #2699, opened 2018-10-19
Fix iOS 12 article search result vertical centering

montehurd:
https://phabricator.wikimedia.org/T207408

## Fixes search result vertical centering
| Before  | After |
|before 22 mov|after 22 mov|

## Fixes another iOS 12 reference highlight regression
- load an article with references visible w/o scrolling (ie refs near very top)
- scroll ever so slightly just enough so when you release the nav bar hides
- tap a reference (remember these should already be onscreen and tappable without scrolling)
- notice the ref highlight is not in the correct position after the ref is tapped

| Before  | After |
|before 2 mov|after 2 mov|


Wikipedia/Code/WebViewController.m
```diff
@@ -974,23 +978,24 @@ - (CGRect)windowCoordsReferenceGroupRect {
...
-
-        rect = CGRectOffset(rect, 0, yOffset);
+        rect = CGRectOffset(rect, 0, 1);
         rect = CGRectInset(rect, -1, -3);
         return rect;
     }
     return CGRectNull;
 }
 
+- (CGFloat)iOS12yOffsetHack {
```
montehurd: naming suggestions wecome! :)

PR #2702, opened 2018-10-19
Batch mark notifications as read requests

nambatee:
Split mark as read requests into chunks of 50 so that we don't exceed the limit on the server


WMF Framework/Remote Notifications/RemoteNotificationsAPIController.swift
```diff
@@ -218,3 +258,11 @@ extension RemoteNotificationsAPIController {
         return session.hasValidCentralAuthCookies(for: .mediawiki)
     }
 }
+
+private extension Array {
```
joewalsh: could this be a public extension? would be good to use elsewhere

WMF Framework/Remote Notifications/RemoteNotificationsAPIController.swift
```diff
@@ -116,25 +117,64 @@ struct RemoteNotificationsAPIController {
...
-
-        request(Query.markAsRead(notifications: notifications), method: .post) { (result: MarkReadResult?, _, _, error) in
-            if let error = error {
-                completion(error)
+        if notifications.count > maxNumberOfNotificationsPerRequest {
+            let split = notifications.chunked(into: maxNumberOfNotificationsPerRequest)
+            let group = DispatchGroup()
+            var markAsReadError: Error? = nil
+            var failed = 0
+            for notifications in split {
```
joewalsh: It looks like this could use `asyncMap`, or maybe a new variant `asyncCompactMap` to allow the error return value to be nil

WMF Framework/Remote Notifications/RemoteNotificationsAPIController.swift
```diff
@@ -116,25 +117,64 @@ struct RemoteNotificationsAPIController {
...
-                completion(MarkReadError.unknown)
-                return
+            group.notify(queue: DispatchQueue.global(qos: .default)) {
+                if failed > 1 {
+                    DDLogError("\(failed) of \(split.count) mark as read requests failed")
+                }
+                completion(markAsReadError)
+            }
+        } else {
+            request(Query.markAsRead(notifications: notifications), method: .post) { (result: MarkReadResult?, _, _, error) in
```
joewalsh: it'd simplify things to allow this to follow the same code path as above, it looks like `chunked` works regardless of whether or not the chunk size is larger than the array size

WMF Framework/Remote Notifications/RemoteNotificationsAPIController.swift
nambatee: sure 👍

WMF Framework/Remote Notifications/RemoteNotificationsAPIController.swift
nambatee: good point 👍

WMF Framework/Remote Notifications/RemoteNotificationsAPIController.swift
nambatee: good idea 👍

WMF Framework/Remote Notifications/RemoteNotificationsAPIController.swift
```diff
@@ -116,25 +118,33 @@ struct RemoteNotificationsAPIController {
...
+                    assertionFailure("Expected result; make sure MarkReadResult maps the expected result correctly")
+                    completion(MarkReadError.noResult)
+                    return
+                }
+                if let error = result.error {
+                    completion(error)
+                }
+                if !result.succeeded {
+                    completion(MarkReadError.unknown)
+                }
```
joewalsh: This is missing a completion for the non-error case. Could implement `asyncCompactMap` and add `completion(nil)`

WMF Framework/Remote Notifications/RemoteNotificationsAPIController.swift
```diff
@@ -116,25 +118,33 @@ struct RemoteNotificationsAPIController {
...
+                if let error = error {
+                    completion(error)
+                }
+                guard let result = result else {
+                    assertionFailure("Expected result; make sure MarkReadResult maps the expected result correctly")
+                    completion(MarkReadError.noResult)
+                    return
+                }
+                if let error = result.error {
+                    completion(error)
```
joewalsh: Missing a return here, completion could be called twice

PR #2711, opened 2018-10-22
Fix description editing cursor jump

montehurd:
was caused by whitespace normalization on the text view being called in its `textViewDidChange`

instead just normalized whitespace on save (otherwise we'd have to dynamically adjust cursor point based on how much leading whitespace was trimmed, which seemed silly)

https://phabricator.wikimedia.org/T207646

### DIFF
--- a/Wikipedia/Code/DescriptionEditViewController.swift
+++ b/Wikipedia/Code/DescriptionEditViewController.swift
@@ -182,7 +182,6 @@ class DescriptionEditViewController: WMFScrollViewController, Themeable, UITextV
         wmf_hideKeyboard()
         
         guard
-            let descriptionToSave = descriptionTextView.normalizedWhitespaceText(),
             let article = article,
             let dataStore = article.dataStore,
             let articleURL = article.url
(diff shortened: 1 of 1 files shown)


Wikipedia/Code/DescriptionEditViewController.swift
```diff
@@ -181,10 +181,12 @@ class DescriptionEditViewController: WMFScrollViewController, Themeable, UITextV
         enableProgressiveButton(false)
         wmf_hideKeyboard()
         
-        // Final trim to remove leading and trailing space
-        let descriptionToSave = descriptionTextView.text.trimmingCharacters(in: .whitespacesAndNewlines)
-        
-        guard let article = article, let dataStore = article.dataStore, let articleURL = article.url else {
+        guard
+            let descriptionToSave = descriptionTextView.normalizedWhitespaceText(),
```
joewalsh: If this is nil, do we want to show an error to the user?

montehurd: checking...

joewalsh: (nil or an empty string)

montehurd: fixing...

PR #2722, opened 2018-10-26
don't allow editing English wikidata descriptions

nambatee:


Wikipedia/Code/WikidataDescriptionEditingController.swift
```diff
@@ -67,7 +67,7 @@ enum WikidataPublishingError: LocalizedError {
     ///   - wiki: wiki of the page to be updated, e.g., "enwiki"
     ///   - completion: completion block called when operation is completed.
     private func publish(newWikidataDescription: String, from source: ArticleDescriptionSource, forPageWithTitle title: String, language: String, wiki: String, completion: @escaping (Error?) -> Void) {
-        guard source != .local else {
```
montehurd: Thoughts on leaving the `source != .local` in place and just adding an `&&` condition checking if language is in a set of blacklisted langs?

Same with the other change below?

Would be nice if ever another language decides they want to do what EN does - would just add that lang code to the blacklisted set...

PR #2726, opened 2018-10-29
Peek/Pop in Places, list view sometimes pops the wrong card/image

nambatee:
https://phabricator.wikimedia.org/T208098


joewalsh:
tests passed locally, merging

PR #2729, opened 2018-10-29
[BUG] Missing reading lists hint when article added to list via long press

nambatee:
https://phabricator.wikimedia.org/T207409

When `SaveButtonsController` notifies the delegate that article saved state changed, `ExploreViewController` is still presenting `AddArticlesToReadingListViewController` so `ReadingListHintController` fails to show the hint


montehurd:
Do you recall why we even have long press on feed card save for later buttons?

joewalsh:
@montehurd so that you can go straight to adding an article to a reading list instead of having to tap the popover

PR #2743, opened 2018-11-03
[BUG] Login status incorrect when changing password (or when cookies expire)

nambatee:
https://phabricator.wikimedia.org/T207382

Changes:

* post a notification when user is logged out as a result of receiving 401 response from the server
* show a "You have been logged out" panel when user is logged out as a result of receiving 401 response from the server
* show "You have been logged out" panel once per login/logout


Wikipedia/Code/Session.swift
```diff
@@ -200,8 +200,9 @@ import Foundation
         }
         switch httpResponse.statusCode {
         case 401:
-            WMFAuthenticationManager.sharedInstance.logout {
+            WMFAuthenticationManager.sharedInstance.logout(initiatedBy: .server) {
                 self.removeAllCookies()
+                NotificationCenter.default.post(name: Session.wasDeauthenticatedNotification, object: nil)
```
joewalsh: Should this notification move into the authentication manager so it's also posted when logout is initiated by `.app`?

Wikipedia/Code/WMFAuthenticationManager.swift
```diff
@@ -36,7 +36,11 @@ public class WMFAuthenticationManager: NSObject {
     /**
      *  The current logged in user. If nil, no user is logged in
      */
-    @objc dynamic private(set) var loggedInUsername: String? = nil
+    @objc dynamic private(set) var loggedInUsername: String? = nil {
+        didSet {
+            UserDefaults.wmf.didShowLoggedOutPanel = false
```
joewalsh: Should this check that the username isn't nil?

Wikipedia/Code/NSUserDefaults+WMFExtensions.swift
```diff
@@ -44,6 +44,8 @@ let WMFSearchLanguageKey = "WMFSearchLanguageKey"
 @objc public extension UserDefaults {
     @objc(WMFUserDefaultsKey) public class Key: NSObject {
         @objc static let defaultTabType = "WMFDefaultTabTypeKey"
+        @objc static let didShowLoggedOutPanel = "WMFDidShowLoggedOutPanelKey"
```
joewalsh: Could the persistence be handled by one property on `WMFAuthenticationManager`? The underlying storage could still be user defaults, but would be good to have `WMFAuthenticationManager` own the state instead of having the panel check it. When the user is logged out by `.app` or `.server`, the property `isUserUnintentionallyLoggedOut` is set to `true`. `WMFAppViewController` checks this when it gets the logged out notification and also on launch after login is attempted. When the user acknowledges the alert, `WMFAppViewController` calls `userDidAcknowledgeUnintentionalLogout()` on `WMFAuthenticationManager` which resets `isUserUnintentionallyLoggedOut` to false.

Wikipedia/Code/WMFAppViewController.m
```diff
@@ -882,7 +887,7 @@ - (void)finishResumingApp {
             if ([error.domain isEqualToString:NSURLErrorDomain]) {
                 return;
             }
-            [self wmf_showReloginFailedPanelIfNecessaryWithTheme:self.theme];
+            [self showLoggedOutPanelIfNeeded];
```
joewalsh: `[self showLoggedOutPanelIfNeeded]` should be called regardless of whether or not there's a network error. It also should be called inside of the success block, since there might be a successful non-login while the user is unaware of a previous logout.

joewalsh: Also, everything in the above completion block needs to be called as well now that the failure block is working. Everything could be combined into one shared `dispatch_block_t loginAttemptCompletion`

nambatee: Could you explain what you mean by "everything in the above completion block needs to be called as well now that the failure block is working"?

joewalsh: Everything in the success completion block - the one declared here `[[WMFAuthenticationManager sharedInstance] attemptLoginWithCompletion:^{`

joewalsh: Everything in the success block should be called regardless of whether or not the user is logged in. Since the success block was always being called before, it wasn't an issue. Now that failure is fixed and will be called sometimes, they should share a completion block that does what the success block was doing before, plus `[self showLoggedOutPanelIfNeeded]`

nambatee: But called where? Attempt login takes a completion block and a failure block, then attempt login transforms those blocks into AuthenticationResultHandler. The completion block is called unless it's a failure.

joewalsh: Everything in the prior success block needs to be called even if it's a failure, so it should be called in both the success and the failure blocks.
```
dispatch_block_t loginAttemptCompletion = ^{
        [self showLoggedOutPanelIfNeeded];
        [self checkRemoteAppConfigIfNecessary];
        if (!self.reachabilityNotifier) {
...
        }
        [self.periodicWorkerController start];
        [self.savedArticlesFetcher start];
        [self.reachabilityNotifier start];
        self.resumeComplete = YES;
    };

    [[WMFAuthenticationManager sharedInstance] attemptLoginWithCompletion:loginAttemptCompletion failure:^(NSError *error) { loginAttemptCompletion(); }];
```

Alternatively, you could rewrite `attemptLoginWithCompletion` to only have one completion block.

nambatee: Ok, it makes more sense to have one completion then.

PR #2751, opened 2018-11-06
Invisible reading lists after erasing

nambatee:
https://phabricator.wikimedia.org/T205203


Wikipedia/Code/StorageAndSyncingSettingsViewController.swift
```diff
@@ -144,17 +144,12 @@ class StorageAndSyncingSettingsViewController: SubSettingsViewController {
...
                 assertionFailure("dataStore is nil")
                 return
             }
-            self.dataStore?.clearCachesForUnsavedArticles()
-            if isSyncEnabled {
-                self.dataStore?.readingListsController.setSyncEnabled(true, shouldDeleteLocalLists: true, shouldDeleteRemoteLists: true)
-            } else {
-                self.dataStore?.readingListsController.setSyncEnabled(true, shouldDeleteLocalLists: true, shouldDeleteRemoteLists: true)
-                self.dataStore?.readingListsController.setSyncEnabled(false, shouldDeleteLocalLists: false, shouldDeleteRemoteLists: false)
-            }
```
joewalsh: This was originally to work around the lack of batch delete. On the branch for this PR, when I create 100 lists with 100 items each, there are 10,000 delete requests created. However, it should be possible to work around that by only deleting the lists from the server and not the entries.

Wikipedia/Code/ReadingListsController.swift
```diff
@@ -452,7 +452,35 @@ public typealias ReadingListsController = WMFReadingListsController
...
+
+        let oldSyncState = syncState
+        var newSyncState = oldSyncState
+
+        if isSyncEnabled {
+            // Since there is no batch delete on the server,
+            // we remove local and remote reading lists
+            // by disabling and then enabling the service.
+            // Otherwise, we'd have to delete everything via single requests.
+            newSyncState.insert(.needsRemoteDisable)
```
joewalsh: does this need a local clear as well?

nambatee: My thinking was -

1. Service is disabled remotely (reading lists are removed on the server)
2. Service is enabled remotely
3. Sync (get lists from the server, delete local lists if they're not on the server - local lists get deleted)

If I add a local clear, it would happen after sync, in `processLocalUpdates`. And at that time, there should be no lists to clear (because they were deleted in `createOrUpdate`)

nambatee:
retest this please

PR #2758, opened 2018-11-07
Fix event logging crash on willEnterForeground

joewalsh:
- Don't load UI or log a session until the app is active
  - Resume should only occur once, on cold launch. Occurs whether or not the user has launched the app and it becomes active.
  - App should become active every time it's launched by the user
  - On cold launch by user, resume might not finish before `didBecomeActive`, so there are some tasks in `performTasksThatShouldOccurAfterBecomeActiveAndResume`


nambatee:
Is there a repro case or a crash report in Hockey or Organizer to see how it went down?

joewalsh:
@nambatee third crash down in Xcode for 6.1.0 release `CoreData: _execute + 156`

PR #2764, opened 2018-11-10
Use new pagelib article title/descrip method.

montehurd:
<strike>Needs https://github.com/wikimedia/wikimedia-page-library/pull/166 to be merged 1st, then will have to bump page lib version to get the new `newEditLeadSectionHeader` method.</strike>

Reminder: can look at this b1597d7 reverted commit later (and parts of 
08efd01) if we want to enable title pronunciation icon. (@carolynlimadeo is leaning against doing so)


joewalsh:
It looks like this is breaking article rendering in the app: ```Promise was rejected with error: TypeError: undefined is not a function (near '...requirements.editTransform.newEditLeadSectionHeader...')```

montehurd:
@joewalsh ah thanks will double-check that

joewalsh:
page load breaks only after building the web components into the assets folder, it looks like they're out of sync with www in the current state of this pr

montehurd:
@joewalsh so far i haven't been able to reproduce the issue you saw. i'm not getting anything funky after running grunt to rebuild. quick things to check would be Clean Build Folder and ensuring the normal 'Wikipedia' scheme is selected (not the page lib debugging one).

i may be missing something though - lemme know :)

montehurd:
@joewalsh is there maybe a specific article you saw it break with? (i've clicked around for a while and haven't seen any issues so far)

montehurd:
oh interesting i am seeing a diff when i run grunt on develop! i think updating dependencies may have caused the grunt process's output to change slightly. checking...

montehurd:
oh weird now i'm not seeing any diff after grunt... hmm...

montehurd:
...and not seeing the page loading issue any more... 🤔

montehurd:
gonna try blasting `www/node_modules/` and rebuilding again...

montehurd:
still no diff, still works fine... gah!

montehurd:
ok i suspect what's happening was `www/node_modules` was out of date.

@joe can you try deleting that folder, then running 'make grunt' and see if you still see the page loading problem?

montehurd:
i'm gonna look at blasting `www/node_modules` as part of `make grunt`...

PR #2767, opened 2018-11-14
TOC selection scrolling offset fixes

montehurd:
https://phabricator.wikimedia.org/T209442

Fixes for TOC selection scrolling offsets: 
- iOS 12 when selecting non-root sections
- iOS 11 and 12 when nav bar hidden


WMF Framework/NavigationBar.swift
```diff
@@ -432,7 +432,11 @@ public class NavigationBar: SetupView, FakeProgressReceiving, FakeProgressDelega
     var extendedViewHeight: CGFloat {
         return extendedView.frame.size.height
     }
-    
+
+    @objc var backgroundViewY: CGFloat {
```
joewalsh: Could this be named `hiddenHeight` for consistency with `visibleHeight`? `backgroundView` is an implementation detail and private, so having this named `backgroundViewY` could be confusing. `hiddenHeight` could be set the same way as `visibleHeight` in `layoutSubviews`, equal to `barTransformHeight + extendedViewTransformHeight + underBarTransformHeight`. `totalTransform` is then `0 - hiddenHeight` and `self.navigationBar.hiddenHeight` is added to the calculation in `WebViewController`

montehurd: Sounds good 👍 Will update.

joewalsh: it should be positive - `hiddenHeight` == the total height that is hidden, similar to how `visibleHeight` == the total height that is visible.  It should be `barTransformHeight + extendedViewTransformHeight + underBarTransformHeight` and the transform is negative, `0 - hiddenHeight`

montehurd: Oops I think I should have had coffee before reading your initial comment ;) Will remedy that now hehe.

montehurd: Fixed!

Wikipedia/Code/WebViewController.m
```diff
@@ -789,7 +789,7 @@ - (void)scrollToFragment:(NSString *)fragment animated:(BOOL)animated {
         [self.webView getScrollViewRectForHtmlElementWithId:fragment
                                                  completion:^(CGRect rect) {
                                                      if (!CGRectIsNull(rect)) {
-                                                         [self.webView.scrollView wmf_safeSetContentOffset:CGPointMake(self.webView.scrollView.contentOffset.x, rect.origin.y)
+                                                         [self.webView.scrollView wmf_safeSetContentOffset:CGPointMake(self.webView.scrollView.contentOffset.x, rect.origin.y + [self iOS12yOffsetHack] - (0 - self.delegate.navigationBar.hiddenHeight))
```
joewalsh: `1 - (0 - x)` is the same as `1 + x`

montehurd: lol oops

montehurd: just a sec..

PR #2770, opened 2018-11-14
As an iOS app editor, when my edit is published I want to be taken to the changed section in the article

nambatee:
Merge after #2767

I created separate tasks for scrolling to the specific content offset (https://phabricator.wikimedia.org/T209509) and handling editing across multiple sections (https://phabricator.wikimedia.org/T209510)


Wikipedia/Code/WMFArticleViewController.m
```diff
@@ -1869,6 +1881,10 @@ - (void)sectionEditorFinishedEditing:(SectionEditorViewController *)sectionEdito
     self.skipFetchOnViewDidAppear = YES;
     [self dismissViewControllerAnimated:YES completion:NULL];
     if (didChange) {
+        __weak typeof(self) weakSelf = self;
+        self.articleContentLoadCompletion = ^{
+            [weakSelf.webViewController scrollToSection:sectionEditorViewController.section animated:YES];
```
joewalsh: This is OK without the corresponding strongify because the block above that asyncs to the main queue strongly captures self. If it were a weak self above, self could be deallocated between when the async call is made an the block is executed

PR #2780, opened 2018-11-19
Remove AFNetworking from WMFLocationSearchFetcher and WMFRandomArticleFetcher

nambatee:


WMF Framework/Configuration.swift
```diff
@@ -114,6 +115,17 @@ public class Configuration: NSObject {
         }
         return false
     }
+
+    @objc(mediawikiAPIURLForHost:)
+    public func mediawikiAPIURLForHost(_ host: String? = nil) -> URL? {
```
nambatee: should this live here or somewhere else?

joewalsh: here is fine

Wikipedia/Code/Session.swift
```diff
@@ -309,15 +314,16 @@ import Foundation
...
-            return
-        }
-        var request = URLRequest(url: url)
-        if ignoreCache {
-            request.cachePolicy = .reloadIgnoringLocalCacheData
+    @objc(getJSONDictionaryFromURL:withQueryParameters:bodyParameters:ignoreCache:completionHandler:)
+    public func getJSONDictionary(from url: URL?, with queryParameters: [String: Any]? = nil, bodyParameters: [String: Any]? = nil, ignoreCache: Bool, completionHandler: @escaping ([String: Any]?, HTTPURLResponse?, Error?) -> Swift.Void) {
+        guard
+            let url = url,
+            let request = requestWithURL(url: url, queryParameters: queryParameters, bodyParameters: bodyParameters, ignoreCache: ignoreCache)
```
nambatee: is it ok if this `getJSONDictionaryFromURL` uses our `request` wrapper instead of `URLRequest`? with the wrapper we're getting params, user agent etc
if not I'll add a separate method

Wikipedia/Code/WMFRandomArticleFetcher.m
```diff
@@ -21,37 +21,64 @@ @implementation WMFRandomArticleFetcher
...
+    }
 
-- (void)fetchRandomArticleWithSiteURL:(NSURL *)siteURL failure:(nonnull WMFErrorHandler)failure success:(nonnull WMFMWKSearchResultHandler)success {
+    NSURL *url = [WMFConfiguration.current mediawikiAPIURLForHost:siteURL.host];
     NSDictionary *params = [[self class] params];
 
-    [self.operationManager wmf_apiZeroSafeGETWithURL:siteURL
-        parameters:params
-        success:^(NSURLSessionDataTask *operation, NSArray *responseObject) {
-            [[MWNetworkActivityIndicatorManager sharedManager] pop];
```
nambatee: does this need this `pop`? there was no matching `[[MWNetworkActivityIndicatorManager sharedManager] push]` before

PR #2781, opened 2018-11-19
Use keyPath based willChangeValue methods instead of string based methods

joewalsh:


Wikipedia/Code/ReadingListsController.swift
```diff
@@ -516,10 +516,10 @@ public typealias ReadingListsController = WMFReadingListsController
...
             let sync = ReadingListsSyncOperation(readingListsController: self)
             addOperation(sync)
-            operationQueue.addOperation {
-                DispatchQueue.main.async {
-                    completion?()
-                }
+            if let completion = completion {
+                let completionBlockOp = BlockOperation(block: completion)
+                completionBlockOp.addDependency(sync)
+                operationQueue.addOperation(completion)
```
nambatee: is it ok if this completion doesn't happen on the main thread?

joewalsh: nope, good catch

PR #2783, opened 2018-11-20
Remove AFNetworking from WMFRandomArticleFetcher and WMFLocationSearchFetcher

nambatee:


WMF Framework/Configuration.swift
```diff
@@ -108,6 +108,17 @@ public class Configuration: NSObject {
         return builder.components(queryParameters: queryParameters)
     }
 
+    @objc(mobileMediaWikiAPIURLComponentsForHost:withQueryParameters:)
```
joewalsh: This can be removed - it was for Wikipedia Zero. The desktop URL can always be used.

PR #2784, opened 2018-11-22
Align color scheme with Safari reader mode

ishitatsuyuki:
These colors are extracted from screenshots manually (since there seems to be no source code on GitHub at least).

I found Safari's color's much more readable due to it's properly adjusted contrast.

### DIFF
--- a/Wikipedia/assets/wikimedia-page-library-transform.css
+++ b/Wikipedia/assets/wikimedia-page-library-transform.css
@@ -7,24 +7,24 @@ styles with only the things which need tweaking.
 
 /* baseline body */
 .pagelib_theme_dark body {
-  color: #f8f9fa;
-  background: #2e3136;
+  color: #d7d7d8;
+  background: #4a4a4c;
 }
 .pagelib_theme_sepia body {
-  color: #222;
-  background: #f8f1e3;
+  color: #4a331f;
+  background: #f7f1e4;
 }
 .pagelib_theme_black body {
-  color: #f8f9fa;
-  background: #000;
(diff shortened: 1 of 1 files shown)


carolynlimadeo:
@ishitatsuyuki, for the default reading theme we utilize the WMF Style Guide colors (https://design.wikimedia.org/style-guide/visual-style_colors.html) which maintain a WCAG conformance level of AA or AAA. The Dark and Black mode color palettes on the Wikipedia app also maintain either a AA or AAA conformance level. 

For Sepia mode the goal of the theme was to reduce blue light, therefore not all colors in this palette reach AA or AAA conformance. 

Was there a specific aspect of the current themes that you were having legibility issues with?

ishitatsuyuki:
The first thing I noticed is that link colors are poorly tuned. Blue is definitely not a good choice for the sepia theme, and the color for dark themes lack contrast over the excessively white text. 

Second, I think dark themes are designed to be used under low lighting environments which means a high contrast is not always the best. Therefore we may want to avoid pure white text here. 

Finally, I prefer the brown-ish text color for the sepia theme probably because it mimics the style of an old book. This may be opinion based though. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

joewalsh:
@ishitatsuyuki  thanks for your contribution and feedback, but I'm closing this PR to maintain conformance to the [style guide](https://design.wikimedia.org/style-guide/visual-style_colors.html) and [WCAG contrast levels](https://www.w3.org/TR/WCAG20/) as per @carolynlimadeo 's comments.

PR #2786, opened 2018-11-26
Remove proxy server, use WKURLSchemeHandler instead

joewalsh:


Wikipedia/Code/WMFURLSchemeHandler.m
```diff
@@ -0,0 +1,494 @@
...
+    if (error) {
+        [task didFailWithError:error];
+    } else if (response) {
+        [task didReceiveResponse:response];
+        if (data) {
+            [task didReceiveData:data];
+        }
+        [task didFinish];
+    } else {
+        [task didFinish];
```
montehurd: I saved a couple articles then cold started the app with no internet connection and see a crash here when I try to load the saved article.

PR #2788, opened 2018-11-26
fix missing TOC dash

montehurd:
https://phabricator.wikimedia.org/T209304


montehurd:
Do you think we should do the full list? : 
https://www.fileformat.info/format/w3c/htmlentity.htm

Oh, side note: we're currently converting the `&nbsp;` to a breaking space, but I think it's ok to continue doing so.

montehurd:
Can try performance test tomorrow...

PR #2792, opened 2018-11-28
mediawiki codemirror integration for editing

montehurd:
- [x] mechanism for fetching [mediawiki-extensions-CodeMirror](https://github.com/wikimedia/mediawiki-extensions-CodeMirror) bits (used `carthage` to grab latest branch: [wmf/1.33.0-wmf.6](https://github.com/wikimedia/mediawiki-extensions-CodeMirror/tree/wmf/1.33.0-wmf.6))
- [x] copy needed code mirror bits to `assets` (used same grunt mechanism we already use to copy files to `assets` - so if we ever want newer version we'll need to bump the `carthage` branch `wmf/1.33.0-wmf.6` to something newer, then re-run `make grunt`)
- [x] html file configured to use codemirror (will be used by editing interface)
- [x] wire up ^ in a wkwebview to editing VC (to display wiki text being edited)
- [x] wire up JS message to send back edited wiki text to native land so it can be previewed/saved
- [x] send messages to native land when cursor insertion point changes so native buttons can respond contextually - ie if a bolded wiki text is selected the bold button selected state can be activated
- [x] wire up methods for messaging from native to js land when buttons are tapped
- [ ] figure out how to handle selected ranges of text (vs just cursor with no text selected)
- [ ] figure out where to get and how to set lang specific config bits appropriately when editing other langs
- [ ] wire up codemirror `find` add on
- [ ] wire up codemirror theme files for closest match to our themes. Related note from chat with @carolynlimadeo: "_looks like all our cool buttons need to still be functional when not in `rich` editing mode (ie syntax highlighting with line numbers). i had just toggled code mirror off for this mode but looks like i'll need to do something to just tone down the text colors instead. (this is because the button wiring takes advantage of codemirror parsing). should be ok but i'll need to confirm the theming bits for this can work with highlighting in either state_".

### So far:
(halfway through i trigger a memory warning to toggle to plain wiki text mode)
fffff mov

### DIFF
--- a/Wikipedia/Code/SectionEditorWebViewConfiguration.swift
+++ b/Wikipedia/Code/SectionEditorWebViewConfiguration.swift
@@ -1,7 +1,10 @@
 import WebKit
 
 protocol SectionEditorWebViewSelectionChangedDelegate: NSObjectProtocol {
-    func turnOffAllButtonHighlights()
+    // Inside 'selectionChanged' the delegate should de-select all buttons.
+    func selectionChanged(isRangeSelected: Bool)
+
+    // Inside these the delegate should enable the respective button.
     func highlightBoldButton()
     func highlightItalicButton()
     func highlightReferenceButton()
@@ -17,6 +20,7 @@ protocol SectionEditorWebViewSelectionChangedDelegate: NSObjectProtocol {
 }
 
 private enum MessageNameConstants: String {
+    case selectionChanged
     case highlightTheseButtons
 }
 
@@ -55,6 +59,7 @@ class SectionEditorWebViewConfiguration: WKWebViewConfiguration, WKScriptMessage
         setURLSchemeHandler(WMFURLSchemeHandler.shared(), forURLScheme: WMFURLSchemeHandlerScheme)
         
         let contentController = WKUserContentController()
+        contentController.add(self, name: MessageNameConstants.selectionChanged.rawValue)
         contentController.add(self, name: MessageNameConstants.highlightTheseButtons.rawValue)
         userContentController = contentController
     }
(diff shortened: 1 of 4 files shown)


Wikipedia/Code/SectionEditorWebView+ExecCommand.swift
```diff
@@ -0,0 +1,100 @@
...
+    @objc func execCodeMirrorCommand(type: CodeMirrorExecCommandType, completionHandler: (SectionEditorWebViewCompletionWithResultBlock)? = nil) {
+        evaluateJavaScript("window.wmf.execCommand(window.wmf.ExecCommandType.\(string(for: type)));", completionHandler: completionHandler)
+    }
+}
+
+
+
+/*
+TODO:
+- use separate methods instead of the exec command approach above
```
montehurd: @nambatee head's-up here are initial stub methods which i'm gonna play with wiring up to the test web view's buttons

Wikipedia/Code/SectionEditorViewController.m
```diff
@@ -323,6 +323,7 @@ - (void)scrollTextViewSoCursorNotUnderKeyboard:(UITextView *)textView {
 - (void)didReceiveMemoryWarning {
     [super didReceiveMemoryWarning];
     // [self.webView toggleRichEditor];
+    // [self.webView becomeFirstResponder];
```
montehurd: @nambatee if you uncomment this, then on my buttons subclass if you select some text, then tap the text area to steal focus, then trigger mem warning, you should see the web view text be re-selected :)

Wikipedia/Code/SectionEditorWebViewConfiguration.swift
```diff
@@ -1,7 +1,10 @@
 import WebKit
 
 protocol SectionEditorWebViewSelectionChangedDelegate: NSObjectProtocol {
-    func turnOffAllButtonHighlights()
+    // Inside 'selectionChanged' the delegate should de-select all buttons.
+    func selectionChanged(isRangeSelected: Bool)
```
montehurd: @nambatee I changed the name of this one passing it the `isRangeSelected` value you needed.

PR #2794, opened 2018-11-29
Faster splash screen animation

joewalsh:
- 0.15 instead of 0.3
- Allow user interaction during cross fade


montehurd:
is this just when hiding the initial app splash panels?

joewalsh:
@montehurd yes

PR #2795, opened 2018-11-29
Commit deps

joewalsh:
https://phabricator.wikimedia.org/T210761


montehurd:
does this orphan the `./scripts/setup` file?

joewalsh:
@montehurd no, carthage is still required for building updated dependencies and it also sets up the linter

PR #2799, opened 2018-12-05
Fix fastlane metadata

joewalsh:
- Only export supported languages in the localization script
- Map our locales to valid app store locales
- Add `upload_app_store_metadata` lane
- Remove `release_notes` until there's a better mechanism for keeping them up-to-date

Note that submitting the metadata isn't currently working due to either an issue with fastlane or the app store (bad gateway error). This gets is closer to working if that's ever resolved.


fastlane/metadata/it/description.txt
```diff
@@ -0,0 +1,43 @@
+Explore your world, find a quick fact, or dive down a Wikipedia rabbit hole with the official Wikipedia app for iOS. With more than 40 million articles across nearly 300 languages, your favorite free online encyclopedia is at your fingertips.
```
montehurd: Just double-checking that the `EN` description should be in this `IT` file... same with `DA`...

PR #2801, opened 2018-12-08
update fastlane snapshot helper file

montehurd:
The screenshot helper file is out of date. Needed so the `kickoff-screenshots` script I wrote (on `appsci`) can run - fastlane refuses to run screenshot process if this file version falls behind.

The `kickoff-screenshots` script (on Desktop of `appsci`) does the following:
- starts screenshot process
- then starts simple python server on port 9090 on the screenshots folder
- when VPN'ed in screenshots can then be accessed at `http://appsci:9090/screenshots.html`

Once this is merged I'll have Automator (or some such) run the script - probably once a day or week depending on how long the full process takes.

### DIFF
--- a/WikipediaUITests/SnapshotHelper.swift
+++ b/WikipediaUITests/SnapshotHelper.swift
@@ -114,10 +114,14 @@ open class Snapshot: NSObject {
         } catch {
             print("Couldn't detect/set locale...")
         }
+        
         if locale.isEmpty {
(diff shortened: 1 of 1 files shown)


joewalsh:
The script should probably be periodically run by Jenkins instead of automator or cron. Jenkins might also have a means of storing/serving the test results.

PR #2808, opened 2018-12-11
Edit toolbar integration

nambatee:


Wikipedia/Code/SectionEditorWebView.swift
```diff
@@ -12,7 +12,7 @@ class SectionEditorWebView: WKWebView {
         super.init(frame: .zero, configuration: config)
         loadHTMLFromAssetsFile(codeMirrorIndexFileName, scrolledToFragment: nil)
         scrollView.keyboardDismissMode = .interactive
-        setKeyboardRequiresUserInteraction(false)
+        allowDisplayingKeyboardWithoutUserInteraction()
```
montehurd: No worries if you want to revert this :)

montehurd:
@joewalsh oh those changes will need to be made to the file in 'www/codemirror' too

Wikipedia/Code/SectionEditorWebViewConfiguration.swift
```diff
@@ -0,0 +1,119 @@
...
+class SectionEditorWebViewConfiguration: WKWebViewConfiguration, WKScriptMessageHandler {
+
+    override init() {
+        super.init()
+        setURLSchemeHandler(WMFURLSchemeHandler.shared(), forURLScheme: WMFURLSchemeHandlerScheme)
+        
+        let contentController = WKUserContentController()
+        contentController.add(self, name: MessageNameConstants.selectionChanged.rawValue)
+        contentController.add(self, name: MessageNameConstants.highlightTheseButtons.rawValue)
+        userContentController = contentController
```
joewalsh: Rather than having the configuration object handle all of the script messages, could these be broken up into separate objects that route the business logic and update the associated views? There's not a clear flow of state or separation of concerns with these notifications being observed and responded to in several places.

www/codemirror/codemirror-index.html
```diff
@@ -0,0 +1,465 @@
...
+  <link rel="stylesheet" type="text/css" href="resources/mode/mediawiki/mediawiki.css">
+  <link rel="stylesheet" type="text/css" href="codemirror-common.css">
+  <link rel="stylesheet" id="codemirror-theme" type="text/css" href="">
+
+  <script>
+    const applyTheme = (themeName) => {
+      document.getElementById('codemirror-theme').setAttribute('href', `codemirror-${themeName}.css`)  
+    }
+    document.addEventListener('DOMContentLoaded', () => {
+      const params = (new URL(window.location)).searchParams
```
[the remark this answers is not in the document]
berndsi: @joewalsh You could still add it but don't expect any effect from it unless the server side implements it. I would try to avoid varying the server response based on user settings if we can can. But I think we could still add client side changes based on it. We plan to allow the client to set the theme and platform anyways, and this could then just made part of that.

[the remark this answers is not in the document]
joewalsh: @berndsi yeah - I mean using query params to pass information to the client-side JS, not any variance in the server response. Thinking about it some more, I would rather have the native clients always go through the same JS interface if possible instead of having the initial config through query params and any subsequent adjustments through executing JS directly.

Wikipedia/assets/mediawiki-extensions-CodeMirror/codemirror-index.html
```diff
@@ -460,10 +460,61 @@
         toggleMarkup('strikethrough', '<s>', '</s>')
       },
       heading: (depth) => {
+        const selectionRange = getSelectionRange(editor)
```
montehurd: I'm gonna re-use and genericize some of this tomorrow...

Wikipedia/Code/ContextualHighlightEditToolbarView.swift
```diff
@@ -1,14 +1,13 @@
...
-    func contextualHighlightEditToolbarViewDidTapItalicToggleButton(_ contextualHighlightEditToolbarView: ContextualHighlightEditToolbarView, button: UIButton)
-    func contextualHighlightEditToolbarViewDidTapRemoveSelectionFormattingButton(_ contextualHighlightEditToolbarView: ContextualHighlightEditToolbarView, button: UIButton)
-    func contextualHighlightEditToolbarViewDidTapReferenceToggleButton(_ contextualHighlightEditToolbarView: ContextualHighlightEditToolbarView, button: UIButton)
-    func contextualHighlightEditToolbarViewDidTapAnchorToggleButton(_ contextualHighlightEditToolbarView: ContextualHighlightEditToolbarView, button: UIButton)
-    func contextualHighlightEditToolbarViewDidTapUnorderedListToggleButton(_ contextualHighlightEditToolbarView: ContextualHighlightEditToolbarView, button: UIButton)
-    func contextualHighlightEditToolbarViewDidTapOrderedListToggleButton(_ contextualHighlightEditToolbarView: ContextualHighlightEditToolbarView, button: UIButton)
+    func textFormattingTapped(sender: ContextualHighlightEditToolbarView)
```
nambatee: this was following the delegate naming convention: https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/CodingGuidelines/Articles/NamingMethods.html#//apple_ref/doc/uid/20001282-1001803-BCIDAIJE

Wikipedia/assets/mediawiki-extensions-CodeMirror/resources/lib/codemirror/lib/codemirror.js
```diff
@@ -9017,10 +9017,10 @@ ContentEditableInput.prototype.setUneditable = function (node) {
 };
 
 ContentEditableInput.prototype.onKeyPress = function (e) {
```
montehurd: @joewalsh 
Will probably want to fork the mediawiki code mirror bits ( https://github.com/wikimedia/wikipedia-ios/pull/2808/files#diff-60bb710c5144ead8f8cb73a22de579abR13 ) or this and the next commit will get overwritten when we run carthage update.

PR #2813, opened 2018-12-19
 Refactor SavedArticlesViewController and ReadingListDetailsViewController

salmanjamil:
https://phabricator.wikimedia.org/T209104

Most of the work is done, but I still have to figure out how to implement sorting based on `ReadingListEntry` creation date.  It would be great if someone can go through the changes quickly and provide some feedback. I have provided an overview of the changes below. Also, I am looking forward  for some suggestions around implementing Sorting on Reading Lists, the challenge here is that we want to have the ability to sort using a `ReadingListEntry` `creationDate` when we are viewing a specific reading list but the fetch results controller on the refactored view controller works on `WMFArticle` object. Since there is no direct reference from an article to a reading list entry so we can't create a sort descriptor for reading list entry creation date. I have a rough idea around adding a `readingListSortOrder` property on `WMFArticle` but I will definitely prefer something simpler that that. 

**Overview of Changes**

I created a new class called `ArticlesCollectionViewController` to implement the common functionality b/w `SavedArticlesViewController` and `ReadingListDetailsViewController`.  One major difference b/w the implementation of `SavedArticlesViewController` and `ReadingListDetailsViewController` was that 
`SavedArticlesViewController` was using `WMFArticle` as the element type of it's fetchedResultsController while `ReadingListDetailsViewController` was using `ReadingListEntry` as its fetchedResultsController element type. `ArticlesCollectionViewController` uses `WMFArticle` as its element type.  `SavedViewController` and `ReadingListDetailViewController` both acts as the container for `ArticlesCollectionViewController` now and also implements `ArticlesCollectionViewControllerDelegate` to customise different parts. `SavedArticlesViewController` has been deleted. 

The differences that I figured out b/w these two classes and how I handled those are listed below

- Both of these classes used a different base predicate, so `basePredicate` is now a property on `ArticlesCollectionViewController` and is passed during initialisation.

- Both had different `ReadingList` instance which is again passed during initialisation

- Batch Editing tool bar actions is configured using a property  and is passed during construction 

- Both had different logic to delete items. `SavedArticlesViewController` would remove an item from all the reading lists while `ReadingListDetailsViewController` would remove it only from it's associated list. This is handled using the delegate object `ArticlesCollectionViewControllerDelegate`.

- Also `SavedArticlesViewController` would ask for confirmation before deleting any article that was present in more than one reading list, this is also handled using the delegate object.

- `SavedArticlesViewController` would display tags for article's reading lists, this is also handled using the delegate object.

- Apart from that the delegate object sends out a couple of other notifications that conforming objects can use to query changes in data. (`ReadingListDetailsViewController` uses one of these methods to update the count of objects after an object is deleted from a `ReadingList`)


nambatee:
@salmanjamil thanks for the PR, great work!

I think you could rely on `ReadingListEntry` in both. To replicate the deletion behavior in what used to be `SavedArticlesViewController`, you could grab all the article keys from a given reading list and nullify the `savedDate` of the articles with those keys. In fact, if you look at `ReadingListsController` and the implementation of `@objc(unsaveArticles:inManagedObjectContext:) public func unsave(_ articles: [WMFArticle], in moc: NSManagedObjectContext)`, we're already kind of doing that:

        do {
            let keys = articles.compactMap { $0.key }
            let entryFetchRequest: NSFetchRequest<ReadingListEntry> = ReadingListEntry.fetchRequest()
            entryFetchRequest.predicate = NSPredicate(format: "articleKey IN %@", keys)
            let entries = try moc.fetch(entryFetchRequest)
            try markLocalDeletion(for: entries)
        } catch let error {
            DDLogError("Error removing article from default list: \(error)")

Once you get rid of `WMFArticle`, you could pass keys directly. You could have an open `delete` `func` in the parent class and override it to either `unsave` given entries (as shown below) or `delete` them from a reading list (as in `public func remove(articles: [WMFArticle], readingList: ReadingList) throws` of `ReadingListsController`). It might be easier to subclass instead of delegate.

Let me know if this makes sense and if there's anything that you think I missed recommending this approach, it's been a while since we made changes to those parts and you have a good grasp of what's going on there.

Thanks for the PR and looking forward to your future contributions!

salmanjamil:
So I ended up using inheritance to solve this issue. `ArticlesCollectionViewController` is now an Abstract class and generic over **T: NSManagedObject**.  As long as subclasses implement 
```
func article(at indexPath: IndexPath) -> WMFArticle {
        fatalError("must be implemented by subclasses")
    }
```
 meaningfully, they are free to chose whatever **T** they want for implementation. This has allowed me to create two subclasses `SavedArticlesCollectionViewController: ArticlesCollectionViewController<WMFArticle>` and  `ReadingListArticlesCollectionViewController: ArticlesCollectionViewController<ReadingListEntry>` and now I can chose readingEntry's `creationDate` as the bases of sorting in `ReadingListArticlesCollectionViewController` and article's `savedDate` in `ArticlesCollectionViewController`. Also I have moved the deletion functionality to concrete subclasses  instead of delegate object. 

There is one issue with this though, since `ArticlesCollectionViewController` is now generic over parameter **T** we can't implement any overrides in extensions. So I have to move all the override method implementations   to the class itself. 

The functionality is complete now, I have done some testing and it is looking fine so far. I plan to go through this once more and test this thoroughly again tomorrow. In the meanwhile if someone has time to review it and provide feedback that will be great.

nambatee:
@salmanjamil you're welcome, thanks for the update! If it makes things easier, `createdDate` of default `ReadingListEntry` should be more or less equivalent to `savedDate` of `WMFArticle` -

when you save an article locally
https://github.com/wikimedia/wikipedia-ios/blob/b726e6603fe274d8748206b2d82727be54f5e069/Wikipedia/Code/MWKSavedPageList.m#L163

you'll first hit this method
https://github.com/wikimedia/wikipedia-ios/blob/75753c01878d6b37f4662d2b90b4d16ee000870d/Wikipedia/Code/ReadingListsController.swift#L675

and in there, we'll set new `createdDate`
https://github.com/wikimedia/wikipedia-ios/blob/75753c01878d6b37f4662d2b90b4d16ee000870d/Wikipedia/Code/ReadingListsController.swift#L691

and right after that, we'll set `savedDate` of the article
https://github.com/wikimedia/wikipedia-ios/blob/75753c01878d6b37f4662d2b90b4d16ee000870d/Wikipedia/Code/ReadingListsController.swift#L705

when we fetch entries from the server, we set `createdDate` to `createdDate` fetched from the server here
https://github.com/wikimedia/wikipedia-ios/blob/ca4df05eaeab29aaa9c023aa8bb9ce59f5feb39e/WMF%20Framework/ReadingListEntry%2BJSON.swift#L7

and here we're setting `savedDate` to `createdDate` of the entry from the server: https://github.com/wikimedia/wikipedia-ios/blob/87ee1099cc9f2fb702b4393fee83ba4a5520b8a4/WMF%20Framework/ReadingListsSyncOperation.swift#L847

So I think `createdDate` and `savedDate` would work similarly in the context of sorting.
`WMFArticle` predates `ReadingListEntry`, originally we had to rely on `savedDate` to display saved articles in order but now it looks like it might not be necessary.

salmanjamil:
@nambatee thank you for taking out time to help me. Now that I think about it, sorting based on `createdDate` of `ReadingListEntry` for **default reading list**  seems like a pretty simple idea but somehow it didn't occurred to me earlier, so thank you again. I have updated the PR and It looks a lot simpler without any generics :) . I have done some testing again and things are looking fine.  Looking forward to your feedback

Wikipedia/Code/ArticlesCollectionViewController.swift
```diff
@@ -0,0 +1,486 @@
...
+            self.endRefreshing()
+        }
+    }
+    
+    open override func endRefreshing() {
+        let now = Date()
+        let timeInterval = 0.5 - now.timeIntervalSince(refreshStart)
+        DispatchQueue.main.asyncAfter(deadline: DispatchTime.now() + timeInterval, execute: {
+            self.collectionView.refreshControl?.endRefreshing()
+        })
```
nambatee: did you mean to add this override? looks like that's exactly what the implementation in `ColumnarCollectionViewController` is doing 😀

https://github.com/wikimedia/wikipedia-ios/blob/8be08a95ea4b11a8372e2c7eefb6589d5a8acfb0/Wikipedia/Code/ColumnarCollectionViewController.swift#L137

Wikipedia/Code/SavedViewController.swift
```diff
@@ -49,7 +49,9 @@ class SavedViewController: ViewController {
                 return
             }
             title = CommonStrings.savedTabTitle
-            savedArticlesViewController = SavedArticlesViewController(with: newValue)
+            let viewController = SavedArticlesCollectionViewController(with: newValue)
+            viewController.emptyViewType = .noSavedPages
```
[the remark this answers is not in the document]
salmanjamil: I think it's reasonable to set the emptyViewType in the initializer of this class and then use single line initialization here. I will update this

Wikipedia/Code/ArticlesCollectionViewController.swift
salmanjamil: Ah, great catch. I probably copied it from somewhere while refactoring and then forgot to get rid of this. Will remove this.

salmanjamil:
@nambatee thank you for the review. I have updated the PR to include your feedback.

Wikipedia/Code/ReadingListDetailViewController.swift
```diff
@@ -109,21 +104,10 @@ class ReadingListDetailViewController: ColumnarCollectionViewController, Editabl
         navigationBar.addExtendedNavigationBarView(extendedView)
     }
     
-    override func viewWillHaveFirstAppearance(_ animated: Bool) {
-        super.viewWillHaveFirstAppearance(animated)
-        setNavigationBarHidingEnabled(true)
```
nambatee: If you add `setNavigationBarHidingEnabled(true)` to `viewDidLoad`, you'll be able to make sure that the search bar (in this case, `extendedView`) is scrolled away unless it's active (like in `SavedViewController`). Also, feel free to get rid of `setNavigationBarHidingEnabled` and set `navigationBar.isExtendedViewHidingEnabled` directly if you want

Wikipedia/Code/ReadingListDetailViewController.swift
```diff
@@ -566,9 +190,7 @@ extension ReadingListDetailViewController: ReadingListDetailUnderBarViewControll
         title = name
     }
     
-    func readingListDetailUnderBarViewController(_ underBarViewController: ReadingListDetailUnderBarViewController, didBeginEditing textField: UITextField) {
-        editController.isTextEditing = true
```
nambatee: If you set `editController.isTextEditing` to `true` here, the top bar buttons will change to "Cancel" and "Done" when you begin editing a list's title, allowing user to decide if they want to keep the new title or discard changes -

img_b36b6f4f6a29-1

salmanjamil:
@nambatee thank you for catching those issues. I have updated the PR. I found another issue where I wasn't returning the edit button state for empty list properly. I have fixed that one also. While fixing that I found out the current implementation like in ReadingListDetailViewController
``` 
var shouldShowEditingButtonForEmtpyState: Bool {
   return !readingList.isDefault
}
```
From my **not so complete understanding** of this part I thought we only display non-default reading lists in `ReadingListDetailViewController`. If that's the case we can probably simply return `true` from this method. For the time being I have used the exact same implementation though since I wasn't sure.

nambatee:
@salmanjamil no problem, thanks for all the fixes. We display the default list among the other lists but it's hidden by default - you can enable it by going to Settings > Article storage and syncing > Show Saved reading list. After you toggle that setting, you should see the default reading list among other reading lists, at the very top. It should have a special tag that says "This list cannot be deleted".

salmanjamil:
@nambatee  I have merged `develop` into my branch to resolve the merge conflict in PR. Please have a look when you get time

Wikipedia/Code/ArticlesCollectionViewController.swift
```diff
@@ -0,0 +1,482 @@
...
+    
+    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        super.traitCollectionDidChange(previousTraitCollection)
+        cellLayoutEstimate = nil
+    }
+    
+    func configure(cell: SavedArticlesCollectionViewCell, for article: WMFArticle, at indexPath: IndexPath, layoutOnly: Bool) {
+        cell.isBatchEditing = editController.isBatchEditing
+        
+        guard let entry = readingList.entry(for: article) else {
```
joewalsh: For reading lists with a lot of entries, this search over the entire list of entries to get the entry for a given article is probably too slow to call for every cell configuration.

This function could take a `ReadingListEntry`, because it can be looked up quickly from the `fetchedResultsController`. Then the article can be looked up using `dataStore.fetchArticle(` (which is cached) as it is in `article(at:`

salmanjamil:
@joewalsh  thank you for your input. I wasn't aware of the caching going on and thought that searching in memory was better then performing a core data fetch :). I have updated the PR to include your suggestions. Please have a look when you get time

WMF Framework/ReadingList+CoreDataClass.swift
```diff
@@ -28,6 +28,12 @@ public class ReadingList: NSManagedObject {
         }
     }
     
+    public func entry(for article: WMFArticle) -> ReadingListEntry? {
```
joewalsh: It looks like this is unused now and can be removed

Wikipedia/Code/ArticlesCollectionViewController.swift
```diff
@@ -0,0 +1,487 @@
...
+    // MARK: - ColumnarCollectionViewLayoutDelegate
+    
+    override func collectionView(_ collectionView: UICollectionView, estimatedHeightForItemAt indexPath: IndexPath, forColumnWidth columnWidth: CGFloat) -> ColumnarCollectionViewLayoutHeightEstimate {
+        // The layout estimate can be re-used in this case becuause both labels are one line, meaning the cell
+        // size only varies with font size. The layout estimate is nil'd when the font size changes on trait collection change
+        if let estimate = cellLayoutEstimate {
+            return estimate
+        }
+        var estimate = ColumnarCollectionViewLayoutHeightEstimate(precalculated: false, height: 60)
+        guard let placeholderCell = layoutManager.placeholder(forCellWithReuseIdentifier: reuseIdentifier) as? SavedArticlesCollectionViewCell, let entry = fetchedResultsController?.object(at: indexPath) else {
```
joewalsh: `let entry = fetchedResultsController?.object(at: indexPath)` previously this was an `entry(at: IndexPath)` function that ensured `indexPath` was valid before requesting it from the `fetchedResultsController` in an ideal world, there's never an invalid index path due to updates from the FRC being handled properly, but in practice it's nice to have the extra insurance against a crash

joewalsh: This comment applies to the other `let entry = fetchedResultsController?.object(at: indexPath)` uses in the file as well

salmanjamil:
@joewalsh, I have pushed the suggested changes. Please have a look when you get time

PR #2818, opened 2019-01-05
Edit messaging refactor

nambatee:
- `SectionEditorViewController` owns `inputViewsController` that manages `webView`'s input views
-  `SectionEditorViewController` owns `messagingController` that receives script messages and delegates back to `SectionEditorViewController`, `SectionEditorViewController` tells->`inputViewsController` tells -> all the input views
- button types have ids to select them by tag instead of switching every time

let me know if you think there's a better way to do this


nambatee:
thanks for catching @montehurd, will fix after the merge if that's ok 👍

PR #2820, opened 2019-01-07
Add language config support, dark theme support

joewalsh:


Wikipedia/assets/mediawiki-extensions-CodeMirror/codemirror-index.html
```diff
@@ -387,10 +364,43 @@
       window.webkit.messageHandlers.highlightTheseButtons.postMessage(selectedButtons(doc))
     }
         
-    const setupCodemirror = () => {
+    const applyTheme = (themeName) => {
+      document.getElementById('codemirror-theme').setAttribute('href', `codemirror-${themeName}.css`)
```
montehurd: Would it be simpler to use the existing theming mechanism? 
`editor.setOption("theme", theme)`

https://github.com/codemirror/CodeMirror/blob/master/demo/theme.html#L165

https://codemirror.net/demo/theme.html#default

joewalsh: It doesn't look any simpler to me unless there would be some advantage with switching themes on the fly. We can evaluate switching to the built in theme handler when adding that functionality.

PR #2827, opened 2019-01-09
List item updates

montehurd:
For list buttons:
- toggles multi-line selections on and off
- deals with single line selections correctly if selection is zero length or more than zero chars

For Increase / decrease indent buttons:
- increases / decreases list items depth for single line or multi-line list item selections

I still need to disable the Increase / decrease indent buttons when a list item is not selected, but will finish adding that in follow-up after interview this afternoon.

I also am not currently adding "Bulleted list item" placeholder text the desktop version does (for zero length selection case) but the cursor is in place so user can add their text. Will follow up with Carolyn about this.


Wikipedia/assets/codemirror/codemirror-index.html
```diff
@@ -250,8 +235,10 @@
       return tagRanges.filter(isTagRangeComplete)
     }
 
-
-
+    const countOcurrences = (str, value) => {
+      const regExp = new RegExp(value, 'gi')
```
montehurd: This is gross and I'll refactor this.

www/codemirror/codemirror-index.html
```diff
@@ -539,8 +529,73 @@
       toggleMarkup(button, openingMarkup, closingMarkup)        
     }
 
-    const convertSelectedLinesToList = (char) => {
-      editor.replaceSelection(`${char} ${editor.getSelection().replace(/\n/gi, `\n${char} `)}\n${char} `)
+    const toggleListForSelectedLines = (char) => {
```
joewalsh: this should match the 2016 wikitext editor behavior and just toggle the list setting of the current line without adding a newline, regardless of selection size

montehurd: the 2016 wikitext editor behavior adds new lines in a few cases where the selection starts after the 1st character of the line. which case are you referring to?

PR #2828, opened 2019-01-09
Fix editing text size

nambatee:
To test

1. Tap on the leftmost button on the default toolbar ("A") or select text and tap on the rightmost button on the contextual toolbar ("...")
2. Tap on "Text size"


Wikipedia/assets/codemirror/codemirror-index.html
```diff
@@ -665,35 +665,39 @@
...
-          return
-        }
-        if (newSize === 'normal') {
-          if (selectionPayloadForButton('smallTextSize')) {
-            expandTagSelectionAndToggleMarkup('smallTextSize', '<small>', '</small>')  
-          }
-          if (selectionPayloadForButton('bigTextSize')) {
-            expandTagSelectionAndToggleMarkup('bigTextSize', '<big>', '</big>')  
-          }
+        const selectedButton = selectionPayloadForButton('textSize')
```
montehurd: It looks like the native side still needs to be updated to handle the size messages from JS land:
aaaaa mov

I double-checked and it looked the size message is making to native land correctly.

Checking out a pre-native-refactor commit (1da401c0b) the JS looks to have been behaving itself, so I'm curious if fixing the native issue above suffices. The reason I wonder this is these new JS changes broke one key thing the old JS structure was trying to do: I made an assumption a user would be more likely to intend the behavior in the first animation below than the second one. I'm curious if you think the same. (side note: the repeated calls to `selectionPayloadForButton` were intentional part of the old approach - though i'm sure they looked sloppy 🤪)

Before this PR (JS seems ok though we still had native interface switching back to `Text formatting` issue):
bbbbbbb mov

With this PR's changes:
ccccccccccc mov

nambatee: Thanks for testing and catching this!

When I was refactoring, I changed the `textSize` message coming from JS - instead of receiving `smallTextSize` and `bigTextSize` as separate messages, I wanted to receive a `textSize` message with information about the size inside `info` (just like what `heading` does with `depth`). However, I didn't update it here so whenever I was telling the text size to change, it was checking whether `smallTextSize` or `bigTextSize` were among the selected buttons, which was never the case anymore. So this is what this change should be fixing - checking whether `textSize` is selected and figuring out the size via `info.size`. I updated it so that each case checks for the `selectedButton`, the way you had it before. If you think there's a better way to do this, let me know or feel free to push an update

The native side changed to fix the pop visible in 1da401c

PR #2835, opened 2019-01-11
Disable indent/outdent buttons dynamically

montehurd:
- disables indent/outdent buttons dynamically (disabled for non-list item selections)
- adds explicit 'disableTheseButtons' message (instead of overloading the 'selectTheseButtons' message)
- switches button tint color back to primaryText (for better contrast with disabled buttons)
- removes some debugging cruft

sdfsfeeefewfe mov

### DIFF
--- a/Wikipedia/Code/SectionEditorWebViewMessagingController.swift
+++ b/Wikipedia/Code/SectionEditorWebViewMessagingController.swift
@@ -61,9 +61,7 @@ class SectionEditorWebViewMessagingController: NSObject, WKScriptMessageHandler 
 
         let ordered = info[Button.Info.ordered] as? Bool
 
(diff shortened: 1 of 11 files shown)


Wikipedia/Code/SectionEditorWebViewMessagingController.swift
```diff
@@ -318,7 +307,7 @@ extension SectionEditorWebViewMessagingController {
                     return 11
                 case .redo:
                     return 12
-                case .debug:
+                case .progress:
```
montehurd: i stuck `progress` here so we wouldn't have a gap (the number 13 wasn't wired up to any button tags so nicer to just re-use it for `progress`)

PR #2838, opened 2019-01-11
fix nav button flicker when using cursor buttons

montehurd:
<strike>this vastly reduces</strike> _fix_ the flicker by ensuring all messages to native land are sent <strike>with minimal amount of time between messages</strike> _at the same time_

<strike>looking at ways of tightening up these intervals even further, but will do as part of follow-on refactors</strike>


montehurd:
One alternate approach would be to bundle all 3 messages (`selectionChanged`, `highlightTheseButtons` and `disableTheseButtons`) as parts of a single message we send to native land. The only trick would be upon receipt we'd have to first process the `selectionChanged` message before the other 2, but I think this would probably eliminate the last tiny flicker I occasionally am still seeing. (But if this PR looks good in the mean time I'd probably prefer to merge it.)

PR #2841, opened 2019-01-14
Fix handling of redirects to title and fragment

joewalsh:
Started a fix for https://phabricator.wikimedia.org/T202923 because iOS has the same bug and was helping Android investigate. Wanted to commit the WIP but it shouldn't be pulled into 6.2.


montehurd:
On `testwiki > bird` in the `goat` section I have a few links I used to tests links which target sections. I added one which uses an article which redirects (`enwiki > UK`) and targets a section:

screen shot 2019-01-16 at 11 06 26 am

It works on desktop but doesn't work on `develop` or this PR. Unsure if it's the cross-wiki syntax that's not being handled. The other links in the screenshot behave correctly on both `develop` and this PR though, so I'm a little foggy on what this PR is fixing.

Side note: The repro steps on the phab ticket are out of date. The ticket repro steps also mention a link which is no longer linking to a section.

joewalsh:
Tagging this into 6.2 to help with/not conflict with #2853

PR #2844, opened 2019-01-16
Re-show keyboard after other input view is interactively dismissed

montehurd:
https://phabricator.wikimedia.org/T213335


montehurd:
Added `WIP` label since https://phabricator.wikimedia.org/T213450 is related - was thinking I'll tack those changes on here...

joewalsh:
@montehurd  this looks good, https://phabricator.wikimedia.org/T213450 might be a big change, so let's merge this and keep that separate

PR #2845, opened 2019-01-16
Fix keyboard observations not being removed

joewalsh:
`NotificationCenter.default.removeObserver(self)` was already present, but that doesn't remove block observers. Switched to the older observation style so that `NotificationCenter.default.removeObserver(self)` works as intended


Wikipedia/Code/SearchLanguagesBarViewController.swift
```diff
@@ -57,25 +57,9 @@ class SearchLanguagesBarViewController: UIViewController, WMFPreferredLanguagesV
...
         
-        NotificationCenter.default.addObserver(forName: NSNotification.Name.WMFAppLanguageDidChange, object: nil, queue: nil) { notification in
-            guard let langController = notification.object, let appLanguage = (langController as AnyObject).appLanguage else {
-                assertionFailure("Could not extract app language from WMFAppLanguageDidChangeNotification")
-                return
-            }
-            self.currentlySelectedSearchLanguage = appLanguage
-        }
+        NotificationCenter.default.addObserver(self, selector: #selector(appLanguageDidChange(_:)), name: NSNotification.Name.WMFAppLanguageDidChange, object: nil)
+        NotificationCenter.default.addObserver(self, selector: #selector(appLanguageDidChange(_:)), name: NSNotification.Name.WMFPreferredLanguagesDidChange, object: nil)
```
ScottPetit: Should this selector be `preferredLanguagesDidChange(_:)` instead of `appLanguageDidChange(_:)`

joewalsh: Yes, good catch!

PR #2847, opened 2019-01-16
Fix ES6 template literals crash

montehurd:
Fix crash seen when trying to edit 
`U.S._vice_president` section of `enwiki  > Political_career_of_John_C._Breckinridge`

### DIFF
--- a/Wikipedia/Code/SectionEditorViewController.swift
+++ b/Wikipedia/Code/SectionEditorViewController.swift
@@ -132,7 +132,11 @@ class SectionEditorViewController: UIViewController {
         // Can use ES6 backticks ` now instead of 'wmf_stringBySanitizingForJavaScript' with apostrophes.
         // Doing so means we *only* have to escape backticks instead of apostrophes, quotes and line breaks.
         // (May consider switching other native-to-JS messaging to do same later.)
-        let escapedWikitext = wikitext.replacingOccurrences(of: "`", with: "\\`", options: .literal, range: nil)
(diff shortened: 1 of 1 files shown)


Wikipedia/Code/SectionEditorViewController.swift
```diff
@@ -132,7 +132,11 @@ class SectionEditorViewController: UIViewController {
         // Can use ES6 backticks ` now instead of 'wmf_stringBySanitizingForJavaScript' with apostrophes.
         // Doing so means we *only* have to escape backticks instead of apostrophes, quotes and line breaks.
         // (May consider switching other native-to-JS messaging to do same later.)
-        let escapedWikitext = wikitext.replacingOccurrences(of: "`", with: "\\`", options: .literal, range: nil)
+        var escapedWikitext = wikitext.replacingOccurrences(of: "`", with: "\\`", options: .literal, range: nil)
+        // Handle ES6 template literals. Fixes crash on `${{Inflation|US|50|1858|r=2}}` string seen on revision 878776294 of: https://en.wikipedia.org/wiki/Political_career_of_John_C._Breckinridge#U.S._vice_president
+        escapedWikitext = escapedWikitext.replacingOccurrences(of: "{", with: "\\{", options: .literal, range: nil)
+        escapedWikitext = escapedWikitext.replacingOccurrences(of: "}", with: "\\}", options: .literal, range: nil)
+
```
joewalsh: Could these be consolidated to a single regex replacement similar to `wmf_stringBySanitizingForJavaScript`?

montehurd: ya was actually doing that right now in a playground :)

PR #2848, opened 2019-01-17
Keep input accessory view at bottom of screen when keyboard dismissed

montehurd:
https://phabricator.wikimedia.org/T213450

11111111 mov

In this case keyboard dismissal occurs when the editor web view resigns 1st responder.
When this happens the system goes up the responder chain looking for a `UIResponder` subclass which wants to become 1st responder.

Making `SectionEditorViewController` able to become 1st responder means it's the first such item in the responder chain encountered when the keyboard is dismissed.

Making its `inputAccessoryView` return `inputViewsController.inputAccessoryView` means the accessory view shown will be the one which was previously shown atop the keyboard.

-----

Looking at how find in page docks its accessory view at bottom when keyboard hidden was helpful:
	https://github.com/wikimedia/wikipedia-ios/pull/794#discussion_r72737202
	https://github.com/wikimedia/wikipedia-ios/pull/794#discussion_r72865675
	https://github.com/wikimedia/wikipedia-ios/pull/794#issuecomment-236402422

These were useful too:
	https://robots.thoughtbot.com/input-accessorizing-uiviewcontroller
	https://developer.apple.com/documentation/uikit/uiresponder/1621119-inputaccessoryview

-----

Question for @carolynlimadeo - this fix doesn't yet keep `Style` or `TextFormatting` header from completely disappearing as seen below. It's pretty easy to fix (just need to move these headers in their own `inputAccessoryView's` too) but wanted to double-check you wanted that 1st.

22222222 mov

### DIFF
--- a/Wikipedia/Code/SectionEditorInputViewsController.swift
+++ b/Wikipedia/Code/SectionEditorInputViewsController.swift
@@ -73,7 +73,7 @@ class SectionEditorInputViewsController: SectionEditorInputViewsSource, Themeabl
         }
     }
 
-    private var inputAccessoryView: UIView? {
+    public var inputAccessoryView: UIView? {
         guard let inputAccessoryViewType = inputAccessoryViewType else {
             return nil
         }
--- a/Wikipedia/Code/SectionEditorViewController.swift
+++ b/Wikipedia/Code/SectionEditorViewController.swift
@@ -164,6 +164,19 @@ class SectionEditorViewController: UIViewController {
         navigationController?.popViewController(animated: true)
         return true
     }
+    
+    // MARK: - Docking `inputAccessoryView` at bottom of screen when keyboard is hidden.
+
+    override var inputAccessoryView: UIView? {
+        // In conjunction with overriding `canBecomeFirstResponder` to be true, this override
+        // allows `inputViewsController.inputAccessoryView` to dock at bottom of screen when
+        // the keyboard is dismissed.
+        return inputViewsController.inputAccessoryView
+    }
+    
+    override var canBecomeFirstResponder: Bool {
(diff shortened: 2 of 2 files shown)


nambatee:
After keyboard is dismissed and toolbar gets docked at the bottom, when I tap on the leftmost icon on the default toolbar ("A"), nothing happens. I have to tap on the `webView` for the Text formatting input view to appear. And then when I dismiss it by tapping on the "X", the default toolbar won't be there, and if I scroll away the keyboard, no view will be visible at the bottom of the screen.

Repro steps:
1. Scroll to dismiss the keyboard
2. Tap "A" on the default toolbar

Expected:
Text formatting input view appears

Actual:
Nothing happens

Repro steps 2:

1. Scroll to dismiss the keyboard
2. Tap "A" on the default toolbar
3. Tap webView
4. Watch Text formatting input view appear right away
5. Tap X to close

Expected:
Previously visible toolbar appears above the keyboard

Actual:
There's no view above the keyboard

So when you dismiss the keyboard now, there will be no view at the bottom.

carolynlimadeo:
@montehurd, thanks for pinging me on this. 

Would it be possible to show the default toolbar when the keyboard is dismissed instead of the sub-view header? This might involve allowing the subview to scroll off and then triggering the toolbar to appear. 

I'm hesitant to keep the sub-view sticky as we're also hoping to have the sub-view be dismissed once it's off the screen.

montehurd:
@nambatee @joewalsh oops sorry I should have added WIP until I heard back from Carolyn.

montehurd:
@nambatee good catch - will see if those are still issues after trying Carolyn's suggestion.

PR #2851, opened 2019-01-17
Scroll to the changed section

nambatee:
https://phabricator.wikimedia.org/T209174

Background:

**TLDR**; When I try to scroll to the edited section, `scrollView.contentSize` is not yet reflecting the size of the article so we scroll to section 0.

When section editor finishes editing with changes, its delegate (`WMFArticleViewController`), sets `articleContentLoadCompletion` to scroll to the edited section:

https://github.com/wikimedia/wikipedia-ios/blob/268319db085afd7042d7a8e36441f44f3d0c70ad/Wikipedia/Code/WMFArticleViewController.m#L1885

`articleContentLoadCompletion` is called when JS posts a `articleContentLoaded` message, which should happen after footer is added:

https://github.com/wikimedia/wikipedia-ios/blob/80ad6edd02cf1999c5e1dacb4bdc24017f5619b2/Wikipedia/Code/WKWebView%2BWMFWebViewControllerJavascript.swift#L220

Unfortunately, when `articleContentLoadCompletion` asks `webView`'s `scrollView` to scroll to the edited section, `scrollView`'s `contentSize` is not yet reflecting its true size so we end up getting a warning about target offset exceeding contentSize and scrolling to 0:

https://github.com/wikimedia/wikipedia-ios/blob/40a90bec2b2e0631a3cc52cb3bd008646d57034c/Wikipedia/Code/UIScrollView%2BWMFContentOffsetUtils.m#L23

and

https://github.com/wikimedia/wikipedia-ios/blob/40a90bec2b2e0631a3cc52cb3bd008646d57034c/Wikipedia/Code/UIScrollView%2BWMFContentOffsetUtils.m#L33

`contentSize` is correct in `didLoadArticle` but since it's called before the content gets loaded, we can't find a scroll view rect for the html element:

https://github.com/wikimedia/wikipedia-ios/blob/268319db085afd7042d7a8e36441f44f3d0c70ad/Wikipedia/Code/WMFArticleViewController.m#L1547

https://github.com/wikimedia/wikipedia-ios/blob/3095d1df940262e6599201b884fed0bcc309c403/Wikipedia/Code/WebViewController.m#L793

Delaying scrolling to section is obviously not a reliable solution so is there something else you'd try?


montehurd:
I think setting the section fragment on the article url should work - was working yesterday on some other stuff I was testing.

PR #2858, opened 2019-01-19
Add find in page to section editor

nambatee:
Dependencies were pulled from `feature/search` branch on https://github.com/wikimedia/wikipedia-ios-codemirror. Cartfile will be updated to pull from `master` once https://github.com/wikimedia/wikipedia-ios-codemirror/pull/1 is merged

TODO:

- [x] apply theme
- [x] fix animation with hidden keyboard (open find in page, scroll away keyboard, tap X to show default toolbar, tap find in page)

https://phabricator.wikimedia.org/T214398


joewalsh:
Looking great - one minor bug I found is that if I search for `wii` in the lead section of today's featured article on enwiki `Mario & Sonic at the Olympic Games` it shows `/10` results but only cycles through 7. If this is a larger issue or somewhere on the Codemirror side it can be a follow-on.

Cartfile.resolved
```diff
@@ -9,4 +9,4 @@ github "luisobo/Nocilla" "0.11.0"
 github "wikimedia/NYTPhotoViewer" "88d2a1a7ac403439e310f55904b3768d257cccd9"
 github "wikimedia/Tweaks" "52db8948eaa85e0a52870237cf0ae2d50a55289a"
 github "wikimedia/hpple" "63951b4c33817321207bfa11ec850b22983cf03a"
-github "wikimedia/wikipedia-ios-codemirror" "77dce80e5da16b41ec2e2cd5e4feedb83297b29e"
+github "wikimedia/wikipedia-ios-codemirror" "24605fe5cab470928e915af5afe97c98f633a4f5"
```
joewalsh: If you run `carthage update wikipedia-ios-codemirror` it'll update codemirror without re-building the other deps. Just for future reference - it's ok to leave the rebuilt deps on this PR

nambatee:
@joewalsh thanks for catching that bug 👍 it should be fixed now. There's also a bug with match highlights persisting when going backwards (aka find previous match) - fix is here https://github.com/wikimedia/wikipedia-ios-codemirror/pull/2

PR #2860, opened 2019-01-20
Fix typo in README.md

revi:
Yes, this is so trivial but better fixed than doing nothing.


revi:
I skipped myself on Contributors list but can do so if you think that's better :P

nambatee:
@revi thanks for the fix and please add yourself to the Contributors list :) every contribution counts 👍

PR #2866, opened 2019-01-23
 Remove Wikipedia Zero, remove AFNetworking 

joewalsh:
https://phabricator.wikimedia.org/T197528


Wikipedia/Code/EditPreviewViewController.m
```diff
@@ -0,0 +1,100 @@
+#import "EditPreviewViewController.h"
```
nambatee: are this file + its header supposed to be here? looks like there's a `EditPreviewViewController.swift` below

Wikipedia/Code/EditSaveViewController.m
```diff
@@ -0,0 +1,579 @@
+#import "EditSaveViewController.h"
```
nambatee: same question as above - I don't see either of these old edit & preview files in Xcode but for some reason GitHub thinks they're coming in

joewalsh: fixed - these were converted to swift in the other PR

Wikipedia/Code/SectionEditorViewController.swift
```diff
@@ -145,10 +146,40 @@ class SectionEditorViewController: UIViewController {
         }
         let message = WMFLocalizedString("wikitext-downloading", value: "Loading content...", comment: "Alert text shown when obtaining latest revision of the section being edited")
         WMFAlertManager.sharedInstance.showAlert(message, sticky: true, dismissPreviousAlerts: true)
-        let manager = QueuesSingleton.sharedInstance()?.sectionWikiTextDownloadManager
-        QueuesSingleton.sharedInstance()?.sectionWikiTextDownloadManager.wmf_cancelAllTasks {
-            let _ = WikiTextSectionFetcher(andFetchWikiTextFor: section, with: manager, thenNotify: self)
+        sectionFetcher.fetch(section) { (result, error) in
+            if let error = error {
+                WMFAlertManager.sharedInstance.showErrorAlert(error as NSError, sticky: true, dismissPreviousAlerts: true)
```
nambatee: should this and the other `showAlert` be dispatched to the main thread?

joewalsh: yup, fixed

Wikipedia/Code/Session.swift
```diff
@@ -261,24 +267,24 @@ import Foundation
         return task
     }
 
-    public func jsonDecodableTask<T: Decodable>(with url: URL?, method: Session.Request.Method = .get, bodyParameters: Any? = nil, bodyEncoding: Session.Request.Encoding = .json, authorized: Bool? = nil, completionHandler: @escaping (_ result: T?, _ response: URLResponse?,_ authorized: Bool?,  _ error: Error?) -> Swift.Void) {
```
nambatee: Apple's example in the Type Constraint Syntax part of The Swift Programming Language (4.2) https://docs.swift.org/swift-book/LanguageGuide/Generics.html shows the `<T: Decodable>` syntax as an example, I think it's more readable to have it up front, especially if the signature is long

joewalsh: good point - fixed

Wikipedia/Code/WMFArticleFetcher.m
```diff
@@ -80,35 +63,25 @@ - (instancetype)initWithDataStore:(MWKDataStore *)dataStore {
 }
 
 - (void)dealloc {
-    [self.operationManager invalidateSessionCancelingTasks:YES];
-    [self.pageSummarySessionManager invalidateSessionCancelingTasks:YES];
+    [self cancelAllFetches];
```
nambatee: `WMFLegacyFetcher` calls `cancelAllFetches` in its `dealloc` so I guess this could be deleted?

joewalsh: yup! done

PR #2859, opened 2019-01-19
Update edit submission screen to use current look and feel

montehurd:
The animation below is WIP -  I'm working on the remaining (much simpler) UI changes to the last screen now.

https://phabricator.wikimedia.org/T209172

- [x] separate Edit Preview and Edit Save interfaces ensuring the following bits still work: 
    - [x] CAPTCHA *
    - [x] login/anon *
    - [x] abuse filter *
    - [x] themes *
    - [ ] more cleanup of the newly split Edit Preview and Edit Save code - first step was getting them functionally separated (time permitting may swift-ify some bits and de-horrible-ize some old cruft)
- [x] use new Edit Summary UI
- [ ] add Minor Edit UI and query param
- [ ] sync with Carolyn on captcha and abuse filter UI

\* So far so good, I've been testing these as I go to lessen breakage repair work.

**WIP animation**:
split mov

PR #2869, opened 2019-01-24
Update edit submission screen to use current look and feel, Part 2

montehurd:
Follow-up for https://github.com/wikimedia/wikipedia-ios/pull/2859

- [x] follow-up cleanups


Wikipedia/Code/EditPreviewViewController.swift
```diff
@@ -8,7 +8,7 @@ protocol EditPreviewViewControllerDelegate: NSObjectProtocol {
 class EditPreviewViewController: UIViewController, Themeable, UITextFieldDelegate, UIScrollViewDelegate, WMFOpenExternalLinkDelegate, WMFPreviewSectionLanguageInfoDelegate, WMFPreviewAnchorTapAlertDelegate {
     var section: MWKSection?
     var wikiText = ""
-    var funnel: EditFunnel?
+    private var funnel: EditFunnel?
```
montehurd: making these private should not have worked... looking at where these get set...

montehurd: i think some edit funnels may have not been wired up previously...

Wikipedia/Code/PreviewWebViewContainer.swift
```diff
@@ -10,70 +10,79 @@ import WMF
...
-            window.wmf.utilities.setLanguage('\(code)', '\(dir)', '\(uidir)');
-            """
+    private func href(from message: WKScriptMessage) -> String? {
+        guard message.name == "anchorClicked", let messageDict = message.body as? [String: Any], let href = messageDict["href"] as? String else {
+            return nil
         }
+        return href
+    }
+    
+    private func configuration() -> WKWebViewConfiguration {
```
joewalsh: This returns a new configuration every time, so it could be potentially confusing if someone tries to use it like a property. It seems like this could live in the web view lazy var.

Wikipedia/Code/AbuseFilterAlertView.swift
```diff
@@ -0,0 +1,68 @@
+
+enum AbuseFilterAlertType {
+    case warning
+    case disallow
+}
+
+class AbuseFilterAlertView: UIView, Themeable {
```
montehurd: Quick and dirty conversion of `AbuseFilterAlert.m` - guessing most of this will go away once Carolyn tweaks the old design.

Wikipedia/Code/PreviewWebViewContainer.swift
```diff
@@ -10,70 +10,79 @@ import WMF
...
+        let controller = WKUserContentController()
+        var earlyJSTransforms = ""
+        if let langInfo = previewSectionLanguageInfoDelegate.wmf_editedSectionLanguageInfo() {
+            earlyJSTransforms = earlyJSTransformsString(for: langInfo, isRTL: UIApplication.shared.wmf_isRTL)
+        }
+        controller.addUserScript(WKUserScript(source: earlyJSTransforms, injectionTime: .atDocumentEnd, forMainFrameOnly: true))
+        controller.addUserScript(WKUserScript(source: "window.wmf.themes.classifyElements(document)", injectionTime: .atDocumentEnd, forMainFrameOnly: true))
+        controller.add(WeakScriptMessageDelegate(delegate: self), name: "anchorClicked")
```
joewalsh: The link handling can likely be handled by setting the `webView`'s `navigationDelegate` instead of adding a click listener

Wikipedia/Code/PreviewWebViewContainer.swift
montehurd: Fixed!

Wikipedia/Code/EditSummaryViewController.swift
```diff
@@ -1,24 +1,33 @@
...
+class EditSummaryViewController: UIViewController, Themeable {
     static let maximumSummaryLength = 255
     
-    public var theme: Theme? {
+    public var theme: Theme = .standard {
         didSet {
-            guard let theme = theme else {
-                return
-            }
             apply(theme: theme)
```
joewalsh: The `Themeable` protocol defines `applyTheme` so that function should set this theme variable as it does in other VCs

Wikipedia/Code/PreviewWebViewContainer.swift
montehurd: Haha good catch!

Wikipedia/Code/EditSummaryViewController.swift
```diff
@@ -65,7 +72,7 @@ import UIKit
         delegate?.learnMoreButtonTapped(sender: sender)
     }
 
-    public func textFieldDidChange(textField: UITextField){
+    @objc public func textFieldDidChange(textField: UITextField){
         notifyDelegateOfSummaryChange()
```
nambatee: wouldn't using `textField(_ textField: UITextField, shouldChangeCharactersIn range: NSRange, replacementString string: String)` be enough to detect changes?

nambatee:
looks like it needs to be constrained to safe area
screen shot 2019-01-25 at 2 08 51 pm

montehurd:
@nambatee good catch! will fix...

Wikipedia/Code/EditSummaryViewController.swift
montehurd: Ah will check thx!

Wikipedia/Code/PreviewWebViewContainer.swift
montehurd: Fixed!

Wikipedia/Code/EditSummaryViewController.swift
montehurd: Fixed!

Wikipedia/Code/EditSummaryViewController.swift
montehurd: In this case `didChange` seems to be the better place to check that a change happened. If `shouldChange` was notified when the text was changed programmatically (when we tap one of the canned summary description buttons) I'd probably switch it so we didn't have to manually notify the delegate on button tap, but unfortunately neither `shouldChange` nor `didChange` get called when we change the text programmatically.

montehurd:
@nambatee fixed the layout guide issue from your screenshot

nambatee:
looks like `cornerRadius` and `masksToBounds` set on the `stackView` inside `EditSummaryViewController` cause console warnings

montehurd:
@nambatee good catch! fixed! must have set that accidentally when setting it on the buttons...

PR #2873, opened 2019-01-26
Update edit submission screen to use current look and feel, Part 3

montehurd:
More follow-ups for #2859 

- [x] add `minor edit` checkbox and wire it up
- [x] add `add to watchlist` checkbox and wire it up (works but is commented pending https://phabricator.wikimedia.org/T214749) 
- [x] update license string
- [x] pixel push font sizes
- [x] wire up `first edit published` panel


Wikipedia/Code/ClassNibView.swift
```diff
@@ -0,0 +1,9 @@
+
+class ClassNibView: UIView {
```
montehurd: makes it easier to use custom xib view classes in another xib or storyboard. makes it easier to break things up

montehurd:
strange... i delete a comment of mine above but it says i "deleted a comment from @joewalsh"...

Wikipedia/Code/EmbeddableView.swift
```diff
@@ -0,0 +1,9 @@
+/* Makes it easier to use custom xib view classes in another xib or storyboard. */
+class EmbeddableView: UIView {
+    override func awakeAfter(using aDecoder: NSCoder) -> Any? {
```
joewalsh: This is more confusing and prone to break than using one of the built-in storyboard/xib composition methods. I'd suggest embedding child view controllers or just having duplicate labels and switches for the edit option views.

Wikipedia/Code/PreviewLicenseView.swift
```diff
@@ -2,7 +2,7 @@ import Foundation
 import UIKit
 import WMF
 
-class PreviewLicenseView: UIView, Themeable {
+class PreviewLicenseView: EmbeddableView, Themeable {
```
joewalsh: I thought we talked about replacing this with the approach used on the announcement cell - it would just be a UITextView with a delegate to handle the link that was actually tapped

PR #2875, opened 2019-01-30
Add haptic feedback to Save for later & Reveal more edit toolbar button

nambatee:
* removes 2 unused methods from `DefaultEditToolbarView`
* adds light haptic feedback to edit toolbar button that reveals the 2nd set of tools
* adds light haptic feedback to Save for later


joewalsh:
Merging to bring the change into a beta build for design feedback

PR #2887, opened 2019-02-07
Fix wikidata description editing event logging bugs

nambatee:
https://phabricator.wikimedia.org/T210970

- [x] Value for wiki should be the same as the article's wiki
- [x] "action":"ready" should be sent once
- [x] add revision number to "action":"saved"
- [x] add the error message to "action":"error"

### DIFF
--- a/Wikipedia/Code/DescriptionEditViewController.swift
+++ b/Wikipedia/Code/DescriptionEditViewController.swift
@@ -222,7 +222,7 @@ class DescriptionEditViewController: WMFScrollViewController, Themeable, UITextV
                     }
                     return
                 }
-                self.editFunnel?.logWikidataDescriptionEditError(self.isEditingExistingDescription, language: language)
+                self.editFunnel?.logWikidataDescriptionEditError(self.isEditingExistingDescription, language: language, errorText: error.localizedDescription)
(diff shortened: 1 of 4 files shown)


Wikipedia/Code/DescriptionEditViewController.swift
```diff
@@ -216,14 +215,14 @@ class DescriptionEditViewController: WMFScrollViewController, Themeable, UITextV
...
-                    self.editFunnel?.logWikidataDescriptionEditSaved(self.isEditingExistingDescription)
+                    self.editFunnel?.logWikidataDescriptionEditSaved(self.isEditingExistingDescription, language: language, revID: self.article?.revisionId)
                     self.delegate?.descriptionEditViewControllerEditSucceeded(self)
                     self.dismiss(animated: true) {
                         presentingVC?.wmf_showDescriptionPublishedPanelViewController(theme: self.theme)
                     }
                     return
                 }
-                self.editFunnel?.logWikidataDescriptionEditError(self.isEditingExistingDescription)
+                self.editFunnel?.logWikidataDescriptionEditError(self.isEditingExistingDescription, language: language, errorText: error.localizedDescription)
```
joewalsh: Is this definitely what the schema calls for? I read it as only errors returned by the API, this could be a network error, JSON parse error, or any other app-side error. Using the localized string instead of a standard set of error code strings could also be an issue

PR #2888, opened 2019-02-07
Add "edit published" hint

nambatee:
https://phabricator.wikimedia.org/T215330

Since the edit hint behaves like the reading list hint, I wanted to consolidate the shared behavior in 2 types - 1 `UIViewController` responsible for the view presentation & responding to touch events and 1 `NSObject` responsible for controlling the hint's visibility.

More or less, `ReadingListHintViewController` became `HintViewController` and `ReadingListHintController` became `HintController`.

Since different hints have different prerequisites for toggling their visibility (reading list hint needs to know about `WMFArticle`), controllers customize their response to the toggle event by overriding the main toggle method.

The bigger change has to do with how hint controllers find out about having to toggle hint's visibility. Before, each VC that presented reading list hint held a reference to the hint controller and called its `didSave` method whenever an article was saved.
Now, the edit hint can appear anywhere in the app so a notification seemed like a better solution. 

However, the problem with that is figuring out which VC should adopt the hint VC 😬

TODO:
- [ ] Replace edit hint icon


Wikipedia/Code/HintController.swift
```diff
@@ -0,0 +1,196 @@
+@objc(WMFHintPresenting)
+protocol HintPresenting: AnyObject {
+    var willDragCompletion: (() -> Void)? { get set }
```
joewalsh: This is a bit unclear - could it be a ref to the `HintController` and on drag explicitly call `dismissHint` or new method `dismissHintDueToUserInteraction` if it needs to be differentiated?

Wikipedia/Code/HintViewController.swift
```diff
@@ -0,0 +1,101 @@
...
+        }
+        previousHeight = view.frame.size.height
+    }
+
+    open func configureSubviews() {
+
+    }
+
+    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        defaultLabel.font = UIFont.wmf_font(.mediumSubheadline, compatibleWithTraitCollection: traitCollection)
```
joewalsh: missing a call to super

Wikipedia/Code/HintController.swift
nambatee: sounds good 👍

PR #2895, opened 2019-02-08
Show "no internet connection" on search instead of top alert

montehurd:
https://phabricator.wikimedia.org/T200333


WMF Framework/UIScrollView+Limits.swift
```diff
@@ -26,9 +26,7 @@ extension UIScrollView {
     }
 
     @objc public func wmf_setContentInset(_ updatedContentInset: UIEdgeInsets, scrollIndicatorInsets updatedScrollIndicatorInsets: UIEdgeInsets, preserveContentOffset: Bool = true) -> Bool {
-        guard updatedContentInset != contentInset || updatedScrollIndicatorInsets != scrollIndicatorInsets else {
```
montehurd: Question: any issues with this being removed? Was preventing the empty view from being sized correctly...

joewalsh: Yes, it prevents duplicate content inset updates and duplicate calls to `scrollViewInsetsDidChange` - the empty view shouldn't have to rely on duplicated calls to either of those so it's likely a problem somewhere else

Wikipedia/Code/SearchViewController.swift
```diff
@@ -123,8 +123,11 @@ class SearchViewController: ArticleCollectionViewController, UISearchBarDelegate
                 guard searchTerm == self.searchBar.text else {
                     return
                 }
-                self.resultsViewController.wmf_showEmptyView(of: WMFEmptyViewType.noSearchResults, action: nil, theme: self.theme, frame: self.resultsViewController.view.bounds)
-                WMFAlertManager.sharedInstance.showErrorAlert(error as NSError, sticky: false, dismissPreviousAlerts: true, tapCallBack: nil)
+                
+                let emptyViewType: WMFEmptyViewType = (error as NSError).wmf_isNetworkConnectionError() ? .noInternetConnection : .noSearchResults
+
+                self.resultsViewController.wmf_showEmptyView(of: emptyViewType, action: nil, theme: self.theme, frame: self.resultsViewController.view.bounds)
```
joewalsh: Is this missing a corresponding `wmf_hideEmptyView` somewhere? Probably on search success?

montehurd: Seemed be resetting already... guessing it did so to ensure the "No results found" empty view was reset on search success. Will double-check.

WMF Framework/UIScrollView+Limits.swift
montehurd: Hmm also seeing the divider line under the search langs disappear after rotate... wonder if that's related...

Wikipedia/Code/SearchViewController.swift
joewalsh: Also, `self.resultsViewController.view.bounds` is likely the cause of the sizing issue. You could hook into `emptyViewType` and `updateEmptyState()` provided by the superclass `ColumnarCollectionViewController` and the frame should be updated correctly.

montehurd: oh cool will check...

montehurd: Doesn't work because of a `guard` in `updateEmptyState` so it never gets to `isEmptyDidChange`... but even if I call `isEmptyDidChange` directly the frame is wrong... looking at why now...

WMF Framework/UIScrollView+Limits.swift
montehurd: Looks like the divider disappearing is caused by `NavigationBarHider`'s `scrollViewDidScroll` setting `shadowAlpha` to 0 in its second `guard`. It seems like both lines in that `guard` which set `shadowAlpha` can be removed - fixes the bug without any side effects that I could see - I may be missing something...

joewalsh: It looks like shadow hiding should be off on the search view altogether - it's there for views like explore & history where the nav bar shadow disappears. Can confirm with Carolyn

Wikipedia/Code/SearchViewController.swift
joewalsh: Fixed empty state update to handle changes from one empty type to another and put calls to update after the results view controller reloads data.

montehurd:
@joe's commits lgtm (GH won't let me `approve` my own PR)

PR #2899, opened 2019-02-08
Can't view articles in certain languages

nambatee:
What's the reason behind constructing articleURLs instead of using the ones provided by the feed endpoint? We end up with invalid URLs for languages with compound language codes, for example Classical Chinese or Samogitian (see https://www.wikidata.org/wiki/Help:Wikimedia_language_codes/lists/all).
Not a recent change cause it's broken in the App Store build (6.1.4) too 

https://phabricator.wikimedia.org/T215634


montehurd:
wonder if the feed endpoint only added those urls after the code in question was already written?

joewalsh:
goes all the way back to the original implementation https://github.com/wikimedia/wikipedia-ios/commit/32203492ec296897d2cb557e7ec313cc4687852c#diff-ffebdfc14cb122f072a34e1bb48c2a8a

Wikipedia/Code/WMFFeedArticlePreview.m
```diff
@@ -44,13 +44,10 @@ + (NSValueTransformer *)thumbnailURLJSONTransformer {
...
                                               BOOL *success,
                                               NSError *__autoreleasing *error) {
-            NSString *lang = value[@"lang"];
-            NSString *normalizedTitle = value[@"normalizedtitle"];
-            NSURL *siteURL = [NSURL wmf_URLWithDefaultSiteAndlanguage:lang];
-            return [siteURL wmf_URLWithTitle:normalizedTitle];
+            return [NSURL URLWithString:value];
         }
         reverseBlock:^NSDictionary *(NSURL *articleURL,
                                      BOOL *success,
```
joewalsh: The reverse block needs to be updated to return the URL's absolute string instead of the dictionary. Also, `+ (NSUInteger)modelVersion` should be bumped to 4

nambatee:
Should be retested after https://gerrit.wikimedia.org/r/#/c/mediawiki/services/mobileapps/+/489329/ is merged

PR #2900, opened 2019-02-08
"No internet connection" panel for article refresh

montehurd:
Fixes for second mock on https://phabricator.wikimedia.org/T200333

- [x] new "no internet connection" panel for article refresh
- [x] DRY up "no internet connection" string
- [x] wire it up
- [x] don't show alert in this case


montehurd:
Side note:

I tried to keep the changes minimal, but we may want to look at refactoring `fetchArticleForce` and `fetchLatestVersionOfArticleWithURL` at some point...

PR #2901, opened 2019-02-08
Fix null localization string on russian articles

montehurd:
https://phabricator.wikimedia.org/T214860

- [x] fixes null substitution on footer `Available in X other languages` on Russian articles
- [x] uses plural syntax on `Edited X days ago` footer menu item string


montehurd:
May want to write test which checks for `{{plural...` any other plurals existing in translation files which don't exist in the base EN file... might look at this this weekend...

Nikerabbit:
Are you sure this works for languages which don't "use" plural like Chinese? I was expecting the second form to be prefixed with `1=`.

joewalsh:
@Nikerabbit for languages that don't use plurals, the translator can enter the universal form as the only entry in a `{{PLURAL` bracket or use no plural syntax and it should work on our end.

For this case, the second form can be prefixed with `1=`, but without the `1=` our script will also treat it as singular because it's the first of two entries without a prefix (first is singular, second is 'other').

Nikerabbit:
I'm talking about `{{PLURAL:%1$d|0=Edited today|Edited yesterday|Edited %1$d days ago}}`. For languages where nplurals = 1, the third form is ignored in my opinion. Hence they cannot make a difference between "yesterday" and "n days ago" unless `1=` is used, which is the reason I think it should be present in all languages. Another reason is that in some language singular form can cover more things than `n = 1` and the form "yesterday" could be inappropriately shown in those cases.

joewalsh:
@Nikerabbit makes sense, I will create a follow-up PR adding` 1=` for clarity in cases like `yesterday` and `last year`

PR #2907, opened 2019-02-11
Plural sytax test and fixes

montehurd:
https://github.com/wikimedia/wikipedia-ios/pull/2901#issuecomment-462029080

- [x] add test for plurals in a translation for EN key which doesn't yet specify plurals
- [x] ~~update strings for translations causing the test above to fail (either by adding plural syntax to EN string or removing it from translation string). I added the `HOLD` label until I hear back on an i18n ticket I filed regarding the last translation string which is causing this PR test to continue failing: https://phabricator.wikimedia.org/T215854.~~
- [x] per @joewalsh's https://github.com/wikimedia/wikipedia-ios/pull/2907#issuecomment-462729415 de-compose the terms & login strings moving the links into the strings and use `byAttributingHTML` to style links instead of old method
- [x] ~~update `testIncomingTranslationStringForHTML` to allow html if the EN strings has html~~
- [x] fix `localization.swift` `isEmpty` error

Side note:
The translate [wiki docs](https://translatewiki.net/wiki/FAQ#A_MediaWiki_message_needs_plural_support_in_my_language_but_the_English_original_doesn't_contain_the_magic_word_PLURAL) also specify that MediaWiki too requires plural syntax be added to the EN string before it's added to a translation string:
_"You can only use the PLURAL function for a variable in MediaWiki if the English source message already contains PLURAL for that variable."_


joewalsh:
@montehurd the confusion around`wikipedia-ios-wikitext-upload-save-terms-cc-by-sa-and-gfdl` could stem from the composition of multiple localized strings. [Other translations embed HTML link tags directly](https://translatewiki.net/w/i.php?title=Special:SearchTranslations&offset=25&query=%3Ca+href&language=en&group=mediawiki&grouppath=mediawiki). As we discussed before, this could be used in conjunction with `UITextView` and our `byAttributingHTML` method to handle links instead of having to compose license names into the string.

scripts/localization.swift
```diff
@@ -483,10 +485,14 @@ func importLocalizationsFromTWN(_ path: String) {
                 try? fm.removeItem(at: folderURL)
             }
             
+<<<<<<< HEAD
```
montehurd: Gah! Not sure how this snuck in here...

montehurd:
Hmm there are 2 tests failing in `WMFNotificationTests.m` but they seem to be failing on `develop` too...

scripts/localization.swift
```diff
@@ -442,18 +450,12 @@ func importLocalizationsFromTWN(_ path: String) {
                     strings[key] = nativeLocalization
                 }
             }
-            let stringsFilePath = "\(path)/Wikipedia/iOS Native Localizations/\(locale).lproj/Localizable.strings"
             if locale != "en" { // only write the english plurals, skip the main file
-                if !strings.isEmpty {
+                if strings.count > 0 {
```
joewalsh: this should stay `!strings.isEmpty`

scripts/localization.swift
```diff
@@ -483,10 +485,7 @@ func importLocalizationsFromTWN(_ path: String) {
                 try? fm.removeItem(at: folderURL)
             }
             
-            
-            let stringsDictFilePath = "\(path)/Wikipedia/iOS Native Localizations/\(locale).lproj/Localizable.stringsdict"
-            
-            if !stringsDict.isEmpty {
+            if stringsDict.count > 0 {
```
joewalsh: this should stay`!stringsDict.isEmpty`

Wikipedia/Code/EditSaveViewController.swift
```diff
@@ -132,23 +134,13 @@ class EditSaveViewController: WMFScrollViewController, Themeable, UITextFieldDel
...
-        licenseTitleLabel.attributedText = licenseTitleLabel.text?.attributedString(attributes: baseAttributes, substitutionStrings: [
-            Licenses.localizedSaveTermsTitle,
-            Licenses.localizedCCBYSA3Title,
-            Licenses.localizedGFDLTitle
-            ], substitutionAttributes: [linkAttributes, linkAttributes, linkAttributes])
+    private var licenseTitleTextViewAttributedString: NSAttributedString {
+        return WMFLocalizedStringWithDefaultValue("wikitext-upload-save-terms-and-licenses", nil, nil, "By publishing changes, you agree to the <a href=\"https://foundation.wikimedia.org/wiki/Special:MyLanguage/Terms_of_Use\">Terms of Use</a>, and you irrevocably agree to release your contribution under the <a href=\"https://creativecommons.org/licenses/by-sa/3.0/\">CC BY-SA 3.0</a> License and the <a href=\"https://www.gnu.org/licenses/fdl.html\">GFDL</a>. You agree that a hyperlink or URL is sufficient attribution under the Creative Commons license.", "Text for information about the Terms of Use and edit licenses.").byAttributingHTML(with: .caption2, matching: traitCollection)
```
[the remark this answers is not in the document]
montehurd: Ya I saw some links inside anchor tags on the "[Other translations embed HTML link tags directly](https://translatewiki.net/w/i.php?title=Special:SearchTranslations&offset=25&query=%3Ca+href&language=en&group=mediawiki&grouppath=mediawiki)" link you shared and forgot to ask for pages like these if there was benefit for translators being able to sub-in urls for translated pages which may exist in their lang... was on the fence...

Wikipedia/Code/EditSaveViewController.swift
```diff
@@ -120,8 +121,9 @@ class EditSaveViewController: WMFScrollViewController, Themeable, UITextFieldDel
...
         addToWatchlistButton.setTitle(WMFLocalizedStringWithDefaultValue("edit-watch-list-learn-more-text", nil, nil, "Learn more about watch lists", "Text for watch lists learn more button"), for: .normal)
         
-        licenseTitleLabel.text = WMFLocalizedStringWithDefaultValue("wikitext-upload-save-terms-cc-by-sa-and-gfdl", nil, nil, "By publishing changes, you agree to the %1$@, and you irrevocably agree to release your contribution under the %2$@ License and the %3$@. You agree that a hyperlink or URL is sufficient attribution under the Creative Commons license.", "Button text for information about the Terms of Use and edit licenses. Parameters:\n* %1$@ - 'Terms of Use' link ([[Wikimedia:Wikipedia-ios-wikitext-upload-save-terms-name]])\n* %2$@ - license name link 1\n* %3$@ - license name link 2")
-        licenseLoginLabel.text = CommonStrings.editAttribution
+        licenseTitleTextView.attributedText = licenseTitleTextViewAttributedString
+        licenseLoginTextView.attributedText = licenseLoginTextViewAttributedString
```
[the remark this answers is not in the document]
montehurd: 👍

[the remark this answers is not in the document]
montehurd: Fixed!

Wikipedia/Code/EditSaveViewController.swift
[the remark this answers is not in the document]
joewalsh: Also missed this before, but should be the Swift function `WMFLocalizedString` instead of `WMFLocalizedStringWithDefaultValue`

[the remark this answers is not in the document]
montehurd: Oh god I missed it again! Forgot to switch these when swift-ifying...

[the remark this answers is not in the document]
montehurd: Fixed!

Wikipedia/Localizations/en.lproj/Localizable.strings
```diff
@@ -679,8 +679,8 @@
 "wikitext-upload-captcha-needed" = "Need CAPTCHA verification.";
 "wikitext-upload-result-unknown" = "Unable to determine wikitext upload result.";
 "wikitext-upload-save" = "Publishing...";
-"wikitext-upload-save-anonymously-or-login" = "Edits will be attributed to the IP address of your device. If you $1Log in$2 you will have more privacy.";
+"wikitext-upload-save-anonymously-or-login" = "Edits will be attributed to the IP address of your device. If you $1 Log in $2 you will have more privacy.";
 "wikitext-upload-save-anonymously-warning" = "Edits will be attributed to the IP address of your device. If you $1 you will have more privacy.";
 "wikitext-upload-save-sign-in" = "Log in";
-"wikitext-upload-save-terms-and-licenses" = "By publishing changes, you agree to the $1Terms of Use$2, and you irrevocably agree to release your contribution under the $3CC BY-SA 3.0$4 License and the $5GFDL$6. You agree that a hyperlink or URL is sufficient attribution under the Creative Commons license.";
```
montehurd: I was a little worried about this `$3CC BY-SA 3.0$4` string potentially causing problems for translators - hence this commit adding space (and the next one trimming it back out). I don't feel strongly about the issue given that our test should fail if the wrong number of substitution strings come back for a translation... happy to revert if you think this is overkill.

montehurd:
test are passing locally 🤔

Wikipedia/Localizations/en.lproj/Localizable.strings
montehurd: Added “question” tag for the comment above...

joewalsh: I think it's good to have available as an option if there is confusion from translators but to start let's keep it without spaces. I rolled back those commits for now.

WikipediaUnitTests/Code/TWNStringsTests.m
```diff
@@ -118,6 +118,15 @@ + (NSRegularExpression *)reverseTWNTokenRegex {
     return reverseTWNTokenRegex;
 }
 
++ (NSRegularExpression *)reverseTWNTokenRegex {
+    static dispatch_once_t onceToken;
+    static NSRegularExpression *iosTokenRegex;
```
nambatee: is it intentional for `NSRegularExpression` to be `iosTokenRegex` and not `twnTokenRegex` here? same question for `reverseiOSTokenRegex` but `twnTokenRegex ` instead of `iosTokenRegex `

joewalsh: nope, fixed

PR #2908, opened 2019-02-12
Fix saved vc download status updating

joewalsh:
- Rename `ArticlesCollectionViewController` to `ReadingListEntryCollectionViewController` to eliminate confusion with `ArticleCollectionViewController`
- Observe article change notifications to update on download status changes, wikidata description changes, etc
- Fix method that was performing an uncached fetch for every cell

https://phabricator.wikimedia.org/T215703

### DIFF
--- a/Wikipedia/Code/ReadingListDetailViewController.swift
+++ b/Wikipedia/Code/ReadingListDetailViewController.swift
@@ -8,7 +8,7 @@ class ReadingListDetailViewController: ViewController {
     let dataStore: MWKDataStore
     let readingList: ReadingList
     
-    let articlesCollectionViewController: ArticlesCollectionViewController
+    let articlesCollectionViewController: ReadingListEntryCollectionViewController
     
     var updater: ArticleURLProviderEditControllerUpdater?
     private let readingListDetailUnderBarViewController: ReadingListDetailUnderBarViewController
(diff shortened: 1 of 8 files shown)


Wikipedia/Code/ReadingListDetailViewController.swift
```diff
@@ -8,7 +8,7 @@ class ReadingListDetailViewController: ViewController {
     let dataStore: MWKDataStore
     let readingList: ReadingList
     
-    let articlesCollectionViewController: ArticlesCollectionViewController
+    let articlesCollectionViewController: ReadingListEntryCollectionViewController
```
nambatee: given the new name, should `articlesCollectionViewController` be renamed to `readingListEntryCollectionViewController` here and in other places?

Wikipedia/Code/ReadingListEntryCollectionViewController.swift
```diff
@@ -230,6 +218,31 @@ class ArticlesCollectionViewController: ColumnarCollectionViewController, Editab
...
+        }
+        
+        let visibleIndexPathsWithChanges = collectionView.indexPathsForVisibleItems.filter { (indexPath) -> Bool in
+            guard let entry = entry(at: indexPath) else {
+                return false
+            }
+            return entry.articleKey == articleKey
+        }
+        
+        guard visibleIndexPathsWithChanges.count > 0 else {
```
nambatee: According to https://developer.apple.com/documentation/swift/array/1688398-isempty, it'd be better to check `isEmpty` than accessing `count` in this case -

> When you need to check whether your collection is empty, use the isEmpty property instead of checking that the count property is equal to zero. For collections that don’t conform to RandomAccessCollection, accessing the count property iterates through the elements of the collection.

joewalsh: In this case, an array of index paths is a `RandomAccessCollection` so it's still constant time, but it's good to get into the habit

PR #2909, opened 2019-02-12
Fix empty SFSafariViewController bug

joewalsh:
- Use namespace returned by search to prevent pushing non-articles to the article vc
- Fetch summary response for externally loaded articleURLs to verify they're articles before pushing them to the article vc
- Remove external url fallback from article vc

https://phabricator.wikimedia.org/T214690


Wikipedia/Code/WMFAppViewController.m
```diff
@@ -1202,8 +1201,28 @@ - (BOOL)processUserActivity:(NSUserActivity *)activity animated:(BOOL)animated c
...
+                    NSNumber *namespaceId = namespaceDictionary[@"id"];
+                    if (![namespaceId isKindOfClass:[NSNumber class]]) {
+                        bail();
+                        return;
+                    }
+                    if ([namespaceId integerValue] != 0) {
+                        bail();
+                        return;
+                    }
+                    [self showArticleForURL:URL animated:animated completion:done];
```
nambatee: wouldn't you want to `bail` if there's an error?

joewalsh: When there's an error, `response` will be nil and it'll bail at the first check

joewalsh: @nambatee I thought about this some more, the app should show the error in this case. If it's a network error or some other error, pushing into the SFSafariVC would just be confusing. I added this change.

joewalsh:
Holding pending more product discussion

joewalsh:
Tagging into 6.2.1 due to impact

PR #2910, opened 2019-02-12
Use isEmpty instead of count > 0 to avoid redundant iteration

nambatee:
https://developer.apple.com/documentation/swift/array/1688398-isempty

> When you need to check whether your collection is empty, use the isEmpty property instead of checking that the count property is equal to zero. For collections that don’t conform to RandomAccessCollection, accessing the count property iterates through the elements of the collection.


joewalsh:
Tagging into 6.2.1 as it's a refactor

montehurd:
good catch!

Wikipedia/Code/MWKSavedPageList.m
```diff
@@ -205,7 +205,7 @@ - (void)removeEntriesWithURLs:(NSArray<NSURL *> *)urls {
 }
 
 + (NSArray<NSDictionary *> *)savedEntryDataFromListWithUnknownSchema:(NSDictionary *)data {
-    return [data[MWKSavedPageExportedEntriesKey] wmf_reverseArray];
```
joewalsh: removed `wmf_reverseArray` because it's not used elsewhere and these migrations will be removed soon anyway

PR #2912, opened 2019-02-12
Can't view articles in certain languages

nambatee:
~Waiting for https://gerrit.wikimedia.org/r/#/c/mediawiki/services/mobileapps/+/489329/~

https://phabricator.wikimedia.org/T215634


joewalsh:
@nambatee with the fallback to the old method and the database key standardized does this still need to be held for the server-side change?

nambatee:
@joewalsh good point, should be good to go 🤔

PR #2918, opened 2019-02-13
Remove URLSessionTaskOperation, fix KVO crash

joewalsh:
Could have handled the cancelled state in the operation which was the cause of the leaky KVO. but it seems the operation for controlling max concurrent requests is entirely unnecessary since `URLSession` handles this independently

https://phabricator.wikimedia.org/T216037


Wikipedia/Code/Session.swift
```diff
@@ -280,10 +279,10 @@ import Foundation
                 completionHandler(nil, response, resultParsingError)
             }
         }) else {
+            completionHandler(nil, nil, RequestError.invalidParameters)
```
nambatee: should this also be added to `jsonCodableTask`? when it returns nil on line `256`

joewalsh: Yeah, good catch

PR #2923, opened 2019-02-14
Log article language in edit events

nambatee:


Wikipedia/Code/MWKSection.m
```diff
@@ -52,6 +53,8 @@ - (instancetype)initWithArticle:(MWKArticle *)article dict:(NSDictionary *)dict
         self.text = [self optionalString:@"text" dict:dict];
 
         self.mutableChildren = [NSMutableArray new];
+
+        self.articleLanguage = article.url.wmf_language;
```
joewalsh: does this code need to be standardized at all?

PR #2926, opened 2019-02-14
Fix edit anchor button state when nested inside template

montehurd:
https://phabricator.wikimedia.org/T216176


joewalsh:
@montehurd this PR has a target branch of develop, it should be 6.2. I'll merge and cherry-pick into 6.2.

PR #2938, opened 2019-02-22
Handle zero, one, few, many prefixes in translations

joewalsh:
This is in addition to the more commonly used numerical prefixes. Follow-on fix for https://phabricator.wikimedia.org/T214860

Also adds 1= prefixes to some translations for clarity. See comments on https://github.com/wikimedia/wikipedia-ios/pull/2901


Jenkinsfile
```diff
@@ -8,9 +8,11 @@ pipeline {
     stage('Test') {
       steps {
         sh '''rm -rf build/reports
+        export LANG=en_US.UTF-8
+        export LANGUAGE=en_US.UTF-8
+        export LC_ALL=en_US.UTF-8
         eval "$(rbenv init -)"
         bundle install
-        scripts/carthage_bootstrap
```
joewalsh: `carthage_bootstrap` is no longer necessary since the prebuilt dependencies are checked in

Jenkinsfile
```diff
@@ -8,9 +8,11 @@ pipeline {
     stage('Test') {
       steps {
         sh '''rm -rf build/reports
+        export LANG=en_US.UTF-8
+        export LANGUAGE=en_US.UTF-8
+        export LC_ALL=en_US.UTF-8
```
joewalsh: this fixes an issue where some output (including new diagnostic output from the localization script as a result of this pr) would cause fastlane to crash while running the build

PR #2940, opened 2019-02-25
Update Ruby & bundler to latest versions

joewalsh:
If you're using `rbenv` update locally by running the setup instructions [starting here](https://github.com/wikimedia/wikipedia-ios/blob/develop/docs/ci.md#install-the-required-ruby-version).

```
cd /your/local/path/to/this/repo
rbenv install -s
gem install bundler
bundle install
```


montehurd:
Updated everything locally per the PR description and lgtm.

PR #2948, opened 2019-02-27
Presenting reading themes popover from Editor

tonisevener:
https://phabricator.wikimedia.org/T213208


Wikipedia/Code/ReadingThemesControlsPresenter.swift
```diff
@@ -0,0 +1,130 @@
...
+    var wkWebView: WKWebView { get }
+    var readingThemesControlsToolbarItem: UIBarButtonItem { get }
+    var passthroughNavBar: Bool { get }
+}
+
+extension ReadingThemesControlsPresenterProtocol {
+    
+    var fontSizeMultipliers: [Int] {
+
+        return [WMFFontSizeMultiplier.extraSmall.rawValue,
```
tonisevener: Do I need to keep the FBTweakValues from the old Obj-C article code?

Wikipedia/Code/ReadingThemesControlsPresenter.swift
```diff
@@ -0,0 +1,130 @@
...
+    }
+    
+    @objc init(readingThemesControlsViewController: ReadingThemesControlsViewController, wkWebView: WKWebView, readingThemesControlsToolbarItem: UIBarButtonItem) {
+        self.readingThemesControlsViewController = readingThemesControlsViewController
+        self.wkWebView = wkWebView
+        self.readingThemesControlsToolbarItem = readingThemesControlsToolbarItem
+        super.init()
+    }
+    
+    @objc func objCShowReadingThemesControlsPopup(on viewController: UIViewController, theme: Theme) {
```
tonisevener: I'm not crazy about these wrapper methods - definitely open to suggestions on this.

Wikipedia/Code/ReadingThemesControlsPresenter.swift
joewalsh: Nope, we haven't been using `FBTweaks`

Wikipedia/Code/ReadingThemesControlsPresenter.swift
joewalsh: I think this is a fine workaround for now. The `mobile-html` work will likely involve re-writing `WMFArticleViewController` in Swift which would eliminate the need for this

Wikipedia/Code/ReadingThemesControlsPresenter.swift
```diff
@@ -0,0 +1,130 @@
...
+//  ReadingThemesControlsPresenter.swift
+//  Wikipedia
+//
+//  Created by Toni Sevener on 2/27/19.
+//  Copyright © 2019 Wikimedia Foundation. All rights reserved.
+//
+
+import Foundation
+
+protocol ReadingThemesControlsPresenterProtocol: WMFReadingThemesControlsViewControllerDelegate, UIPopoverPresentationControllerDelegate {
```
nambatee: if you want to have a shorter name for the protocol, you could use `ReadingThemesControlsPresenting`, after https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/CodingGuidelines/Articles/NamingBasics.html

> Most protocols group related methods that aren’t associated with any class in particular. This type of protocol should be named so that the protocol won’t be confused with a class. A common convention is to use a gerund (“...ing”) form

tonisevener: @nambatee oh yeah that's better, I'll do that. thx!

Wikipedia/assets/codemirror/codemirror-black-none.css
```diff
@@ -0,0 +1,95 @@
+html, body, .CodeMirror {
```
joewalsh: @tonisevener it should be possible to have a single `codemirror-syntax-highlighting-off.css` that has all of the css classes with non-default text colors set to `color: inherit`. This separate css could be linked below the `codemirror-theme` css and toggled on and off with the href switching between `"codemirror-syntax-highlighting-off.css"` and `""`

joewalsh: Also, just to clarify, this isn't a required suggestion since the approach here works and affords more flexibility, just mentioning in case it's helpful

tonisevener: @joewalsh ah no I like that, inherit was what I was missing. I wasn't sure how to turn off syntax coloring but still keep the base theme color without quadrupling up on css files. I'll change, thanks!

www/codemirror/codemirror-index.html
```diff
@@ -371,6 +371,17 @@
...
+  
+    const toggleSyntaxColors = () => {
+        var name = document.getElementById('codemirror-theme').getAttribute('href')
+        
+        var newName = name.includes("-none") ? name.replace("-none", "") : name.substring(0, name.length - 4) + "-none.css"
+        document.getElementById('codemirror-theme').setAttribute('href', newName)
+    }
+  
+    const scaleBodyText = (textSizeAdjustment) => {
+      document.querySelector('body').style['font-size'] = (textSizeAdjustment/100.0) + 'em'
```
montehurd: One thing to maybe try would be using `-webkit-text-size-adjust` vs `font-size`. It seemed to scale things proportionally a little better for articles. Not a big deal, but may be interesting to compare the two...

Wikipedia/Code/ReadingThemesControlsViewController.xib
```diff
@@ -256,6 +261,38 @@
...
+                                    <constraints>
+                                        <constraint firstAttribute="height" priority="999" constant="1" id="1me-Am-mAC"/>
+                                    </constraints>
+                                </view>
+                                <view contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="Kkd-Mq-to7">
+                                    <rect key="frame" x="0.0" y="240" width="290" height="52"/>
+                                    <subviews>
+                                        <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" verticalHuggingPriority="251" text="Syntax Highlighting" textAlignment="natural" lineBreakMode="tailTruncation" baselineAdjustment="alignBaselines" adjustsFontSizeToFit="NO" translatesAutoresizingMaskIntoConstraints="NO" id="uMd-Oo-cjP">
+                                            <rect key="frame" x="10" y="15" width="211" height="24"/>
+                                            <fontDescription key="fontDescription" style="UICTFontTextStyleTitle2"/>
```
tonisevener: Note I used the system Title 2 here, but let me know if there's a preferred way to do text styles.

nambatee: You could override `traitCollectionDidChange` and then use `wmf_font`:

```
    open override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
        super.traitCollectionDidChange(previousTraitCollection)
        syntaxHighlightingLabel.font = UIFont.wmf_font(.[font type], compatibleWithTraitCollection: traitCollection)
    }
```

Wikipedia/Code/ReadingThemesControlsViewController.swift
```diff
@@ -1,15 +1,17 @@
...
 
-@objc public protocol WMFReadingThemesControlsViewControllerDelegate {
+public protocol WMFReadingThemesControlsViewControllerDelegate: class {
     
     func fontSizeSliderValueChangedInController(_ controller: ReadingThemesControlsViewController, value: Int)
+    func toggleSyntaxHighlighting(_ controller: ReadingThemesControlsViewController)
 }
 
 @objc(WMFReadingThemesControlsViewController)
 open class ReadingThemesControlsViewController: UIViewController {
```
nambatee: should this be `open` or would `public` suffice?

Wikipedia/Code/SectionEditorViewController.swift
```diff
@@ -250,6 +255,14 @@ extension SectionEditorViewController: SectionEditorNavigationItemControllerDele
     func sectionEditorNavigationItemController(_ sectionEditorNavigationItemController: SectionEditorNavigationItemController, didTapRedoButton redoButton: UIBarButtonItem) {
         messagingController.redo()
     }
+    
+    func sectionEditorNavigationItemController(_ sectionEditorNavigationItemController: SectionEditorNavigationItemController, didTapReadingThemesControlsButton readingThemesControlsButton: UIBarButtonItem) {
+        
+        messagingController.blur()
```
montehurd: `webView.resignFirstResponder()` may work here w/o need for new new js `blur()`

nambatee:
is it intended behavior that the keyboard stays visible if find in page is active?

tonisevener:
> is it intended behavior that the keyboard stays visible if find in page is active?

Oh yes, sorry forgot to mention, that's a bug. According to the future find & replace design (https://phabricator.wikimedia.org/T215955) the user won't be able to select the reading themes button in that mode, so I didn't think it was worth chasing down at this point if the user will soon be prevented from getting in that state.

tonisevener:
A few more bugs I've noticed -

Popover arrow seems to lose it's anchor sometimes and displays center-aligned instead of pointing at the themes button.
CSS swapping performance can drag if it's a longer section.
Keyboard & toolbars animating away need improvements. When on a formatting submenu it switches to an unthemed keyboard before animating away.

montehurd:
> Keyboard & toolbars animating away need improvements. When on a formatting submenu it switches to an unthemed keyboard before animating away.

fingers-crossed this helps 🤞...  
https://github.com/wikimedia/wikipedia-ios/pull/2948#pullrequestreview-214190657

Wikipedia/Code/ReadingThemesControlsViewController.swift
tonisevener: @nambatee good point - actually I don't even think public is necessary since it's not a part of the WMF framework...I have removed, let me know what you think.

Wikipedia/Code/ReadingThemesControlsViewController.xib
tonisevener: 👍 done!

Wikipedia/Code/SectionEditorViewController.swift
tonisevener: 👍 done, and it fixed that bug too, thanks!

montehurd: 🕺🎉

Wikipedia/Code/ReadingThemesControlsViewController.swift
nambatee: awesome! 🙌

PR #2949, opened 2019-02-28
Fix reading list sync

nambatee:
`jsonDictionaryTask` is never executed without calling `resume` on it first

To reproduce, set a breakpoint here (it won't break) or try to sync

https://github.com/wikimedia/wikipedia-ios/blob/6e7bc47f6509fc19a8c5a6e1a1929d60636bdf09/WMF%20Framework/ReadingListsAPIController.swift#L153


Wikipedia/Code/Session.swift
```diff
@@ -169,7 +169,9 @@ import Foundation
         guard let request = request(with: url, method: method, bodyParameters: bodyParameters, bodyEncoding: bodyEncoding) else {
             return nil
         }
-        return jsonDictionaryTask(with: request, completionHandler: completionHandler)
+        let task = jsonDictionaryTask(with: request, completionHandler: completionHandler)
+        task.resume()
+        return task
```
joewalsh: The original intent was to have the methods named `jsonDictionaryTask` just return tasks without calling `resume()` so they could be tracked before resuming in the caller. As it is here, the completion might be called before it's tracked in the `ReadingListsAPIController`, so the resume should be there after the `self.track`

nambatee: Wouldn't this also apply to `jsonDecodableTaskWithDecodableError`?

joewalsh: Yup, good catch. I have that that wrong there as well

PR #2951, opened 2019-03-01
Fix keyboard behavior in section editor

joewalsh:
If `preventDefault()` is called on mouse, touch, or pasteboard events, the iOS keyboard doesn't update to reflect the current state. For example, if the user taps to place the cursor and `preventDefault()` is called on the resulting `"touchend"` event, the keyboard doesn't update to reflect the new cursor position. This causes [T217075](https://phabricator.wikimedia.org/T217075). 

This PR expands the `preventDefaultOnKeyPress` option added to fix the initial symptoms of this issue where the shift key stuck on while typing. The new option, `preventDefault`, disables the mouse, key, touch, and pasteboard event handlers installed by CodeMirror. [The event handlers that update CodeMirror's state from the DOM](https://github.com/wikimedia/wikipedia-ios-codemirror/blob/1b2b8cd0fa2e42583d4a8f92e216114260da40b2/resources/lib/codemirror/lib/codemirror.js#L8710...L8727) remain, and it seems that's why everything continues to work as expected.

I've verified that the editing toolbars, cut/copy/paste, and undo/redo work as expected. I added the `hold` tag pending further discussion as this is an invasive change.


joewalsh:
After testing these changes some more, there's broken behavior around scrolling to the text cursor. Repro steps are - scroll to the bottom of a section > scroll to hide the keyboard > tap to insert the cursor

PR #2953, opened 2019-03-01
Wikidata description welcome shown after every edit

nambatee:
https://phabricator.wikimedia.org/T217070

Initially, I modified the notification posted from `WikidataDescriptionEditingController` to be a generic `didPublishNotification` with `authorized` flag as the object.
However, when `AppViewController` is notified, `DescriptionEditViewController` is still being presented so the hint is not shown.

I considered

1) Dismissing `DescriptionEditViewController` in `AppViewController` before the hint appears - if I call `dismissViewController` and toggle hint in the `completion`, `completion` won't be called because `DescriptionEditViewController` will dismiss itself before `AppViewController` dismisses it
2) Moving the notification from `WikidataDescriptionEditingController` to `DescriptionEditViewController` and passing `authorized` in the completion block - `RemoteNotificationsOperationsController` is in `WMF` so it couldn't observe the notification if its name was in `DescriptionEditViewController `

### DIFF
--- a/Wikipedia/Code/WMFAppViewController.m
+++ b/Wikipedia/Code/WMFAppViewController.m
@@ -224,7 +224,7 @@ static const NSString *kvo_SavedArticlesFetcher_progress = @"kvo_SavedArticlesFe
 
     [[NSNotificationCenter defaultCenter] addObserver:self
                                              selector:@selector(dscriptionEditWasPublished:)
-                                                 name:[WikidataDescriptionEditingController didPublishNotification]
+                                                 name:[DescriptionEditViewController didPublishNotification]
                                                object:nil];
 
     self.readingListsAlertController = [[WMFReadingListsAlertController alloc] init];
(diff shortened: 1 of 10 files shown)


Wikipedia/Code/WMFAppViewController.m
```diff
@@ -507,13 +512,20 @@ - (void)articleWasUpdated:(NSNotification *)note {
...
 
-- (void)editPublished:(NSNotification *)note {
+- (void)editWasPublished:(NSNotification *)note {
     if (![NSUserDefaults.wmf wmf_didShowFirstEditPublishedPanel]) {
         return;
     }
     [self toggleHint:self.editHintController context:nil];
 }
 
+- (void)dscriptionEditWasPublished:(NSNotification *)note {
```
joewalsh: Typo here and in `@selector(dscriptionEditWasPublished:)`

PR #2957, opened 2019-03-06
Added IconBarButtonItem for disabled tint control

tonisevener:
https://phabricator.wikimedia.org/T216097

UIBarButtonItem needs a custom UIButton view to control disabled tintColor, so I created an IconBarButtonItem subclass to be able to hook into isEnabled.
Most of the other bar button icons on ArticleVC could potentially be disabled (not just language) so converted those over to IconBarButtonItems as well.

For the disabled color, I'm using Theme's disabledLink. It was only used in 1 other place, on a pageIndicatorTintColor so instead I made a separate color for pageIndicatorTintColor & pageIndicatorTintColorCurrent so I could be free to tweak the disabledLink values without affecting other colors.

final black
final dark


Wikipedia/Code/WMFArticleViewController.m
```diff
@@ -531,69 +531,68 @@ - (void)updateToolbarItemEnabledState {
...
 }
 
-- (UIBarButtonItem *)shareToolbarItem {
+- (IconBarButtonItem *)shareToolbarItem {
     if (!_shareToolbarItem) {
-        _shareToolbarItem = [[UIBarButtonItem alloc] initWithBarButtonSystemItem:UIBarButtonSystemItemAction
-                                                                          target:self
-                                                                          action:@selector(shareArticle)];
+        _shareToolbarItem = [[IconBarButtonItem alloc] initWithIconName: @"share" target: self action:@selector(shareArticle) for: UIControlEventTouchUpInside];
+        _shareToolbarItem.accessibilityLabel = WMFCommonStrings.accessibilityShareTitle;
```
joewalsh: Good catch, thanks for fixing the missing accessibility label!

tonisevener: @joewalsh thanks but it was still accessible before for me, I guess because it was a system button :)

PR #2958, opened 2019-03-07
Select text to edit

montehurd:
https://phabricator.wikimedia.org/T215448

Allows user to highlight a word or range of text and tap "Edit" to show either the title description editor or the section editor:
 
color mov

### High-level description of approach:
- user selects text and taps `Edit`
- js extracts the selection and some adjacent words to either side of the selection. section id is also determined as is whether the selection is part of a title description
- this info is sent to native land
- native land checks the info received and shows title description editor or section editor as needed
- if the section editor is shown native land relays the selection text and adjacent words to the code mirror instance 
- regex tries to locate the correct occurrence of the selected text by using it and a few words to either side of it. regex tries its best to work around differences between article text and wikitext - this is the hard part.
- if can't be found tries with fewer words 
- if can't be found tries with more permissive (but slower and less accurate) regex
- if found the selection range is highlighted

### Know issues 😱:

👎 wikitext template markup can differ drastically from template output - sometimes they output text (which is selectable) which appears nowhere in the template wikitext

👎 page lib transforms can complicate locating selections:
- 1st paragraph relocation: makes things harder because the paragraph is no longer in same relative location as it is in the wikitext (can _sometimes_ make adjacent words useless)
- collapsed table transform: adds text to things such as collapsed tables headers and footers, which is also selectable but not represented anywhere in wikitext
- footer transform: same issue with outputting selectable text not represented anywhere in wikitext

# Edge cases

The edge case screenshots below are from:
- enwiki > Aitkin County, Minnesota
- enwiki > Al-Mu'tasim
- enwiki > John Adams
- testwiki > Bird7

## 😀 Handled edge cases:
Getting examples like these to work led me to approach I used.

| Article selection  | Wikitext selection |
| screen shot 2019-03-06 at 9 27 27 pm  | screen shot 2019-03-07 at 2 38 23 pm | 
| screen shot 2019-03-06 at 8 24 07 pm  | screen shot 2019-03-07 at 2 41 09 pm | 
| screen shot 2019-03-06 at 5 19 40 pm  |screen shot 2019-03-07 at 2 44 51 pm| 
| screen shot 2019-03-06 at 4 15 56 pm | screen shot 2019-03-07 at 2 45 05 pm|
| screen shot 2019-03-06 at 3 42 37 pm | screen shot 2019-03-07 at 2 45 24 pm |
| screen shot 2019-03-06 at 2 52 43 pm | screen shot 2019-03-07 at 2 45 36 pm |
| screen shot 2019-03-06 at 1 26 51 pm | screen shot 2019-03-07 at 2 45 48 pm |
| screen shot 2019-03-06 at 1 20 17 pm | screen shot 2019-03-07 at 2 46 06 pm|
| screen shot 2019-03-04 at 11 28 12 pm | screen shot 2019-03-07 at 2 46 25 pm|
| screen shot 2019-03-04 at 10 22 55 pm | screen shot 2019-03-07 at 2 46 50 pm |
| screen shot 2019-03-04 at 10 22 26 pm | screen shot 2019-03-07 at 2 47 19 pm|
| screen shot 2019-03-06 at 10 29 18 pm | screen shot 2019-03-07 at 2 47 39 pm |
|screen shot 2019-03-07 at 2 52 58 pm|screen shot 2019-03-07 at 2 53 04 pm|
|screen shot 2019-03-07 at 2 55 04 pm|screen shot 2019-03-07 at 2 55 10 pm|
|screen shot 2019-03-07 at 2 56 13 pm|screen shot 2019-03-07 at 2 56 18 pm|
|screen shot 2019-03-07 at 2 56 26 pm|screen shot 2019-03-07 at 2 56 33 pm|
|screen shot 2019-03-07 at 2 56 42 pm|screen shot 2019-03-07 at 2 56 47 pm|
|screen shot 2019-03-07 at 2 56 58 pm|screen shot 2019-03-07 at 2 57 03 pm|
|screen shot 2019-03-07 at 3 00 27 pm|screen shot 2019-03-07 at 3 00 33 pm|
|screen shot 2019-03-07 at 3 49 31 pm|screen shot 2019-03-07 at 3 47 34 pm|
|screen shot 2019-03-07 at 4 37 44 pm|screen shot 2019-03-07 at 4 37 51 pm|

## 🥴 Unhandled edge cases:

| Article selection  | Wikitext selection |
|screen shot 2019-03-07 at 3 13 37 pm|screen shot 2019-03-07 at 3 13 44 pm|

2 issues above: The first is that in wikitext the infobox appears above the selection, whereas in the article we've moved the first paragraph above the infobox. This normally wouldn't be a problem without the second issue - which is the pronunciation guide immediately after the selection comes from a wikitext template that has markup that differs pretty significantly from what it outputs. This makes the regex which tries to disambiguate using some words after the selection fail. The end result of these 2 issues is an incorrect infobox occurrence is highlighted.

| | |
| screen shot 2019-03-07 at 3 31 40 pm | screen shot 2019-03-07 at 3 31 46 pm|

1 issue above: The selected word is output by the `{{distinguish` template, which as you can see has markup which doesn't even include the word `Not` - so there's nothing to highlight.

| | |
| screen shot 2019-03-07 at 3 42 38 pm|screen shot 2019-03-07 at 3 42 43 pm|

1 issue above: The selected word is output by a page library transform which isn't produced by wikitext so there's nothing to highlight.

| | |
|screen shot 2019-03-07 at 4 43 50 pm|screen shot 2019-03-07 at 4 44 00 pm|

2 issues above: Some of the selection is output by the `{{convert` template, which adds a comma to the number 1995. Second issue is a few other `{{convert` usages come right after this, further complicating disambiguating the text the user selected.


Wikipedia/Code/WebViewController.h
```diff
@@ -77,6 +77,7 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)webViewController:(WebViewController *)controller didTapOnLinkForArticleURL:(NSURL *)url;
 - (void)webViewController:(WebViewController *)controller didSelectText:(NSString *)text;
 - (void)webViewController:(WebViewController *)controller didTapShareWithSelectedText:(NSString *)text;
+- (void)webViewControllerDidTapEditMenuItem:(WebViewController *)controller;
```
montehurd: i need to rename this...

Wikipedia/Code/WKWebView+EditSelectionJavascript.swift
```diff
@@ -0,0 +1,63 @@
...
+                
+                completionHandler(selectedTextEditInfo, nil)
+                return
+            }
+            DDLogError("Error when evaluating javascript on fetch and transform: \(error)")
+        }
+    }
+}
+
+@objcMembers class SelectedAndAdjacentText: NSObject  {
```
nambatee: does this need to be exposed to objc and inherit from `NSObject`? it's only used in Swift

Wikipedia/assets/codemirror/codemirror-editTextSelection.js
```diff
@@ -0,0 +1,126 @@
...
+  }
+}
+
+const wikitextRangeForSelectedAndAdjacentText = (selectedAndAdjacentText, wikitext) => {
+  const regex = selectedAndAdjacentText.regexForLocatingSelectedTextInWikitext(wikitext)
+  if (regex === null) {
+    return null
+  }
+  const match = wikitext.match(regex)
+  const matchedWikitextBeforeSelection = match[1]
```
nambatee: should we check if `match` is not null before we try to retrieve values?

Wikipedia/assets/codemirror/codemirror-editTextSelection.js
```diff
@@ -0,0 +1,126 @@
...
+  return wikitextRange
+}
+
+const getWikitextRangeToSelect = (wikitextBeforeSelection, wikitextSelection) => {
+  const wikitextBeforeSelectionLines = wikitextBeforeSelection.split('\n')
+  const startLine = wikitextBeforeSelectionLines.length - 1 
+  const startCh = wikitextBeforeSelectionLines.pop().length
+
+  const wikitextSelectionLines = wikitextSelection.split('\n')
+  const endLine = startLine + wikitextSelectionLines.length - 1
```
nambatee: should we check if these strings are not `undefined` before we call `split`?

Wikipedia/assets/codemirror/codemirror-editTextSelection.js
```diff
@@ -0,0 +1,126 @@
...
+  
+  return {from, to}
+}
+
+const scrollToAndHighlightRange = (range, codemirror) => {
+  let marker = null
+  // Temporarily set selection so we can use existing `scrollCursorIntoViewIfNeeded` method to bring the selection on-screen.
+  codemirror.setSelection(range.from, range.to)
+  setTimeout(() => {
+    scrollCursorIntoViewIfNeeded(true)
```
nambatee: would it be appropriate to use `window.requestAnimationFrame` here instead of `setTimeout` or is this a different use case?

Wikipedia/Code/WMFArticleViewController.m
```diff
@@ -1651,6 +1651,22 @@ - (void)webViewController:(WebViewController *)controller didTapShareWithSelecte
...
+    @weakify(self);
+    [self.webViewController.webView wmf_getSelectedTextEditInfoWithCompletionHandler:^(SelectedTextEditInfo* selectedTextEditInfo, NSError *error) {
+        @strongify(self);
+        if (error) {
+            return;
+        }
+        if (selectedTextEditInfo.isSelectedTextInTitleDescription) {
+            [self showTitleDescriptionEditor];
+        } else {
+            MWKSection *section = self.article.sections[selectedTextEditInfo.sectionID];
```
tonisevener: might want to confirm the sections count before accessing

Wikipedia/assets/codemirror/codemirror-editTextSelection.js
```diff
@@ -0,0 +1,126 @@
...
+  let from = {line: startLine, ch: startCh}
+  let to = {line: endLine, ch: endCh}
+  
+  return {from, to}
+}
+
+const scrollToAndHighlightRange = (range, codemirror) => {
+  let marker = null
+  // Temporarily set selection so we can use existing `scrollCursorIntoViewIfNeeded` method to bring the selection on-screen.
+  codemirror.setSelection(range.from, range.to)
```
tonisevener: since you mention it's temporary, do you need to unset this somewhere? Only reason I ask is the highlighted menu appears (starts with Bold, Italic, etc) rather than the default menu from the Zeplin screenshot, I was thinking this might be the culprit.

Wikipedia/Code/WKWebView+EditSelectionJavascript.swift
montehurd: good catch! fixed!

Wikipedia/assets/codemirror/codemirror-editTextSelection.js
montehurd: for sure! fixed!

Wikipedia/assets/codemirror/codemirror-editTextSelection.js
montehurd: just tested - now that we're checking the `match` for null it looks like these strings are never `undefined` here - they can be zero length, but that's sometimes expected and doesn't cause an issue with `split`, so... good to go i think :)

Wikipedia/assets/codemirror/codemirror-editTextSelection.js
montehurd: good question! just checked - and the `removeAllRanges` call is a good use case for `window.requestAnimationFrame`

montehurd: fixed!

montehurd: based on @tonisevener's comments i was able to refactor this part be able to scroll to the location without using `setSelection` and `scrollCursorIntoView` - there's now a `scrollRangeIntoViewIfNeeded` which for this use case works better and fixes this 
https://github.com/wikimedia/wikipedia-ios/pull/2958#pullrequestreview-212389499 too.

Wikipedia/Code/SectionEditorViewController.swift
```diff
@@ -137,6 +138,11 @@ class SectionEditorViewController: UIViewController {
     
     func setWikitextToWebView(_ wikitext: String, completionHandler: ((Error?) -> Void)? = nil) {
         messagingController.setWikitext(wikitext, completionHandler: completionHandler)
+
+        guard let selectedTextEditInfo = selectedTextEditInfo else {
+            return
+        }
+        messagingController.highlightAndScrollToText(for: selectedTextEditInfo)
```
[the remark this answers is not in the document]
montehurd: great catch! fixed!

Wikipedia/assets/codemirror/codemirror-editTextSelection.js
montehurd: oh wow i totally missed that! good catch!

fixed! 

now using `scrollRangeIntoViewIfNeeded` so no more need to highjack `setSelection` and `scrollCursorIntoView` here

Wikipedia/Code/WMFArticleViewController.m
montehurd: hopefully there's no way to edit text if there's no section having that text... but can't hurt to check :)

fixed!

Wikipedia/assets/preview.html
```diff
@@ -12,6 +12,8 @@
       A.edit-page {
         display: none;
       }
+      .section-heading {display: block !important;}
+      .mw-editsection {display: none !important;}
```
nambatee: I couldn't find any references to `.mw-editsection` in the DOM - is it used elsewhere?

montehurd: On edit preview on non-lead section - it shows a giant [Edit] link which we used to hide somehow...

PR #2964, opened 2019-03-12
Clear markup button

montehurd:
https://phabricator.wikimedia.org/T215954

### Basic functionality:
rrrrr mov

### Range splitting:
(a range splitting example is at start of animation)
ccccc mov

### Deep nesting, multi-line and range splitting:
fefefefefe mov

### More nesting:
nesting 2 mov

- [x] clears selection markup
- [x] works with multi-line selections
- [x] works with tag-based and non-tag markup (new range determination bits can be used to vastly simplify button state determination and “(un)wrapping” later)
- [x] handles moving markup around partial selections
- [x] handles start or end of selection in middle of markup text 
- [x] handles start or end of selection on edge of markup text 
- [x] removals grouped as single `undo` item
- [x] button is disabled if no markup can be removed
- [x] button is disabled if zero length selection
- [x] button is disabled if selection encompasses reference
- [x] button is disabled if selection encompasses template
- [x] edge case where markup needs to be added (vs cases above which involve moving or removal) - see `Scenario 2e` in the ticket

### Approach
Background: 

Codemirror wikitext tokenization spreads information about logical wikitext items across several tokens in a way that makes them difficult to treat as logical objects to be further processed/compared/filtered etc.

This PR provides a method `markupItemsForLineTokens` which gathers information for logical wikitext items from a spread of Codemirror tokens and models them as individual `MarkupItem`s which can be more easily used.

`MarkupItem`s have inner and outer ranges:
```
MarkupItem
     innerRange: ItemRange
     outerRange: ItemRange
```
`ItemRange`s have start and end locations:
```
ItemRange
     startLocation: ItemLocation
     endLocation: ItemLocation
```
`ItemLocation`s have line and ch `Ints`:
```
ItemLocation
     line: Int
     ch: Int
```

This PR uses `MarkupItem`s to enable wikitext format clearing:

- A `clearFormatting` method is exposed which performs the actual clearing logic:
```
const clearFormatting = (codeMirror) => {
  codeMirror.operation(() => {
    relocateOrRemoveExistingMarkupForSelectionRange(codeMirror, false)
    splitMarkupAroundSelectionRange(codeMirror, false)
    pruneWhitespaceOnlyMarkupItems(codeMirror)
  })
}
```
- The `relocateOrRemoveExistingMarkupForSelectionRange` method handles cases where existing wikitext markup (such as `<sup>` or `'''`) needs to be moved around the selection (but not added).
- The `splitMarkupAroundSelectionRange` method handles cases where clearing the selection of markup means we'd have to add markup around the selection.
- A `canClearFormatting` method is also provided and use for button state.


Wikipedia/assets/codemirror/codemirror-range-determination-bundle.js
```diff
@@ -0,0 +1,736 @@
...
+    markupRangesToAddAfterSelection.push(item.openingMarkupRange())
+  })
+
+  let accumulatedLeftMarkup = getTextFromRanges(codeMirror, markupRangesToAddBeforeSelection)
+  let accumulatedRightMarkup = getTextFromRanges(codeMirror, markupRangesToAddAfterSelection)
+
+  // Work-around for cases which will result in empty markup items such as `<u></u>`.
+  // Injects text inside such markup items so they can be identified and pruned. 
+  // This also prevents a codemirror bug where it incorrectly tokenizes empty markup items.
+  const removeMeMarker = 'R!E@M#O$V%E'
```
joewalsh: I'd be OK with the empty markup case if it means removing this workaround - not sure it's worth the added complexity and potential for leaving this string in the wikitext.

I'm noticing that if I have a bold, underlined selection:
Screen Shot 2019-03-21 at 6 31 33 AM

Then tap clear markup, I end up with:
Screen Shot 2019-03-21 at 6 31 44 AM

Same problem with bold & italics:
Screen Shot 2019-03-21 at 6 36 05 AM

joewalsh: Not as critical, but I'm noticing extra steps in the undo stack - If I have this selection:
Screen Shot 2019-03-21 at 6 46 32 AM

Then clear formatting, it takes two undo steps to get back to the initial state.

montehurd: good catch will check!

montehurd: Fixed both! 

The first issue was a result of an underlying code mirror wiki text parsing bug where it incorrectly tokenized the closing markup for certain italic-nested-in-bold cases.

I was able to [detect and correct](https://github.com/wikimedia/wikipedia-ios/pull/2964/commits/fccba146ea9a44e59fe4f9c93987c5f0a298dc47) for these cases within the non-tag range determination code. Nice 'cause it means our `MarkupItem` will have correct ranges even when the underlying code mirror tokens don't.

Second issue with undo stack getting too many items is fixed here https://github.com/wikimedia/wikipedia-ios/pull/2964/commits/e93c51433f555a05fffa5c8609761f54d8af1271

Working on nixing the `R!E@M#O$V%E` marker now...

montehurd: Got rid of `R!E@M#O$V%E` :)

joewalsh:
Selecting the text of a header:
Simulator Screen Shot - iPhone X - 2019-03-25 at 09 52 15
Then tapping clear formatting results in more formatting being added:
Simulator Screen Shot - iPhone X - 2019-03-25 at 09 52 30
Verified on sim & device

tonisevener:
I think this might fail scenario 3b: 

`User highlights a portion of a template or a string within a template that has a formatting tag associated with it`
`Expected result: Clear formatting icon is set to disabled`

IMG_0014

montehurd:
Ah good catch! Fixing... (Flight’s not ‘till 4pm.)

montehurd:
@carolynlimadeo regarding @tonisevener's comment https://github.com/wikimedia/wikipedia-ios/pull/2964#issuecomment-476238918 , i initially fixed it but then realized it's nice to be able to clear the markup inside templates - such as the `[[nuclear weapons]]` link in the selection in Toni's screenshot... what do u think? Leave it so users can do this or should the button be grayed out in these cases too?

carolynlimadeo:
Hi Monte, sorry I didn't catch this comment before your flight. I'm leaning towards Toni's initial fix. As nice as it would be to strip out formatting within a template, I can't help but feel like this gets us into weird territory. Let's disable the clear formatting option anywhere inside of a template. Thanks!

montehurd:
@carolynlimadeo - to be clear, presently it strips markup (like link, bold etc) inside internal template selection but leaves the internal template markup itself untouched. But ya fine with disabling altogether.

PR #2965, opened 2019-03-13
Switch related pages and random to RESTBase endpoints

nambatee:
https://phabricator.wikimedia.org/T199456


WMF Framework/RelatedSearchFetcher.swift
```diff
@@ -18,14 +18,6 @@ final class RelatedSearchFetcher: Fetcher {
             let thumbnail: Image?
             let coordinates: Coordinates?
 
-            var isList: Bool {
-                return description?.contains("Wikimedia list article") ?? false
-            }
-
-            var isDisambiguation: Bool {
-                return description?.contains("disambiguation page") ?? false
```
joewalsh: Rather than further the use of these english-specific checks, it's probably better to remove them from `MWKSearchResult`.  They were only used to improve the quality of random unused elsewhere. I updated the random fetcher to filter on `isDisambiguation` and removed the `isList` sort. I'd argue that it's OK for random to show lists, but worth discussing.

WMF Framework/RelatedSearchFetcher.swift
```diff
@@ -60,7 +60,7 @@ final class RelatedSearchFetcher: Fetcher {
     @objc func fetchRelatedArticles(forArticleWithURL articleURL: URL?, resultLimit: Int = RelatedSearchFetcher.MaxResultLimit, completion: @escaping (Error?, [MWKSearchResult]?) -> Void) {
         guard
             let articleURL = articleURL,
-            let articleTitle = articleURL.wmf_titleWithUnderscores
+            let articleTitle = articleURL.wmf_titleWithUnderscores?.addingPercentEncoding(withAllowedCharacters: .wmf_articleTitlePathComponentAllowed)
```
joewalsh: Maybe `addingPercentEncoding` should move into `mutating func replacePercentEncodedPathWithPathComponents(_ pathComponents: [String]?) {` to prevent this mistake in the future

WMF Framework/RelatedSearchFetcher.swift
nambatee: There's also this endpoint we could start using at some point -> https://en.wikipedia.org/api/rest_v1/#!/Page_content/get_page_random_format

WMF Framework/RelatedSearchFetcher.swift
nambatee: Good idea 👍

WMF Framework/RelatedSearchFetcher.swift
montehurd: @joewalsh hmm... unsure why it's better to decrease the quality of the English results...

joewalsh: @montehurd I like list articles. From my perspective, it's an increase in quality. Ultimately I'm fine with whatever's decided after consulting with the rest of the team - I've already implemented the switch to the upstream random endpoint that keeps similar heuristics: https://github.com/wikimedia/wikipedia-ios/tree/bug/random

montehurd: I'm fine if Carolyn and Josh want them back but they were not a good experience imo. Mind if we loop them in before changing this?

joewalsh: Don't mind at all, that's what I was suggesting

WMF Framework/RelatedSearchFetcher.swift
joewalsh: Moving this introduces issues with other path components, so parking it for now

PR #2969, opened 2019-03-15
Editing on boarding: As a new editor, I'd like for there to be an introduction to Wikitext the first time I visit the editor

nambatee:
https://phabricator.wikimedia.org/T212789

Some of the views are set up for animation follow-ons and future consolidation - https://phabricator.wikimedia.org/T218379


joewalsh:
Simulator Screen Shot - iPhone SE - 2019-03-15 at 08 08 55
^ last commit fixes this. When the W icon popover was presented behind editing, the stack view wouldn't switch to horizontal for some reason. Repro steps were reset SE simulator > clean install > view article > tap edit pencil before W Icon popover appears > rotate to landscape

Wikipedia/Code/WelcomePageViewController.swift
```diff
@@ -0,0 +1,199 @@
...
+        guard viewIfLoaded != nil else {
+            self.theme = theme
+            return
+        }
+        allViewControllers.forEach { ($0 as? Themeable)?.apply(theme: theme) }
+        view.backgroundColor = theme.colors.midBackground
+        pageControl?.pageIndicatorTintColor = theme.colors.pageIndicator
+        pageControl?.currentPageIndicatorTintColor = theme.colors.pageIndicatorCurrent
+        nextButton.tintColor = theme.colors.link
+        skipButton.tintColor = UIColor(0xA2A9B1)
```
tonisevener: maybe this should be a theme color?

Wikipedia/Code/WelcomeContainerViewController.swift
```diff
@@ -0,0 +1,71 @@
...
+        if let backgroundAnimationView = dataSource.backgroundAnimationView {
+            addChild(WelcomeAnimationViewController(position: .background, animationView: backgroundAnimationView), to: topBackgroundContainerView)
+        } else {
+            topBackgroundContainerView.isHidden = true
+        }
+        addChild(dataSource.panelViewController, to: bottomContainerView)
+        apply(theme: theme)
+    }
+    
+    private func addChild(_ viewController: UIViewController?, to view: UIView) {
```
tonisevener: I notice there's a few similar chunks of this code, we could move adding a child VC to a UIViewController extension at some point

Wikipedia/Code/WelcomePageViewController.swift
nambatee: good point 👍

nambatee: although maybe when we consolidate all the onboarding stuff, we could get away with just having one `public static wmf_someShadeOfGray` since this button has the same colors in all the themes right now 🧐

Wikipedia/Code/WelcomeContainerViewController.swift
nambatee: we have `@objc public func wmf_add(childController: UIViewController!, andConstrainToEdgesOfContainerView containerView: UIView!)` that I forgot about 🙈

PR #2970, opened 2019-03-15
Find & replace in Editor

tonisevener:
https://phabricator.wikimedia.org/T215955


Wikipedia/assets/codemirror/resources/addon/search/search.js
```diff
@@ -208,16 +208,72 @@
     });}
```
tonisevener: https://github.com/wikimedia/wikipedia-ios-codemirror/pull/6

Wikipedia/Code/SectionEditorViewController.swift
```diff
@@ -73,6 +80,36 @@ class SectionEditorViewController: UIViewController {
...
+    private func setupFocusNavigationView() {
+        #warning("todo: localize")
+        
+        focusNavigationView.isHidden = true
+        focusNavigationView.configure(text: "Find and replace", traitCollection: traitCollection)
+        focusNavigationView.delegate = self
+        focusNavigationView.apply(theme: theme)
+        
+        focusNavigationView.translatesAutoresizingMaskIntoConstraints = false
+        navigationController?.navigationBar.addSubview(focusNavigationView)
```
tonisevener: This seems gross to me and I'm not sure if it'll be accessible yet. I'm open to other ideas on it. In terms of responsibility this code might make more sense living in WMFThemeableNavigationController?

tonisevener:
This is ready to review - I will make subtasks to T215955 for localizing / accessibility stuff.

Wikipedia/Code/FindAndReplaceKeyboardBar.swift
```diff
@@ -0,0 +1,345 @@
...
+        currentMatchLabel.text = labelText
+    }
+    
+    func updatePreviousNextButtonsState(total: UInt) {
+        previousButton.isEnabled = total >= 2
+        nextButton.isEnabled = total >= 2
+    }
+    
+    func updateReplaceLabelState() {
+        #warning("todo: localize")
```
joewalsh: Could this (and the other `#warning`s) be updated to comments so that we keep the project at zero warnings? It's easier to see if new compiler warnings are added while developing that way

tonisevener: 👍 done!

Wikipedia/Code/SectionEditorViewController.swift
joewalsh: Could this tell the navigation controller to hide the navigation bar and then display the focus navigation view at the top of this view controller's view? That might be an easier way to address the responsibility issue

Wikipedia/Code/SectionEditorViewController.swift
```diff
@@ -75,6 +82,36 @@ class SectionEditorViewController: UIViewController {
...
+        
+        focusNavigationView.setNeedsLayout()
+        focusNavigationView.layoutIfNeeded()
+        
+        navigationController?.navigationBar.leadingAnchor.constraint(equalTo: focusNavigationView.leadingAnchor).isActive = true
+        navigationController?.navigationBar.trailingAnchor.constraint(equalTo: focusNavigationView.trailingAnchor).isActive = true
+        focusNavigationViewBottomConstraint = navigationController?.navigationBar.bottomAnchor.constraint(equalTo: focusNavigationView.bottomAnchor, constant: focusNavigationView.frame.height)
+        focusNavigationViewBottomConstraint.isActive = true
```
joewalsh: you might be able to use  `wmf_addSubview(_, withConstraintsToEdgesWithInsets:)` here

Wikipedia/Code/SectionEditorViewController.swift
joewalsh: Alternatively, this could become a subclass of `ViewController` and utilize the included `NavigationBar` - `focusNavigationView` could be the `underBarView` and it has the ability to hide/show individual components with `setNavigationBarPercentHidden(_ navigationBarPercentHidden: CGFloat, underBarViewPercentHidden: CGFloat, extendedViewPercentHidden: CGFloat, topSpacingPercentHidden: CGFloat...`

joewalsh: Either way, not necessary for this PR, can be a follow-on

tonisevener: @joewalsh I went with your first suggestion (in https://github.com/wikimedia/wikipedia-ios/pull/2987). The ViewController superclass would move some of the theming and keyboard handling events around, which seems a little risky at this point.

PR #2975, opened 2019-03-15
Fix unhelpful password reset error messages

nambatee:
https://phabricator.wikimedia.org/T218416

### DIFF
--- a/Wikipedia/Code/WMFAuthenticationManager.swift
+++ b/Wikipedia/Code/WMFAuthenticationManager.swift
@@ -17,14 +17,14 @@ public class WMFAuthenticationManager: Fetcher {
         public var errorDescription: String? {
             switch self {
             default:
-                return CommonStrings.genericErrorDescription
+                return WMFLocalizedString("error-generic-description", value: "An unexpected error occurred", comment: "Generic error message for when the error isn't recoverable by the user.")
             }
         }
         
         public var recoverySuggestion: String? {
             switch self {
             default:
-                 return CommonStrings.genericErrorRecoverySuggestion
+                 return WMFLocalizedString("error-generic-recovery-suggestion", value: "Please try again later", comment: "Generic recovery suggestion for when the error isn't recoverable by the user.")
             }
         }
     }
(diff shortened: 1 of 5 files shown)


Wikipedia/Code/WMFPasswordResetter.swift
```diff
@@ -2,14 +2,25 @@
...
+        case .accountError(let code, let message) where WMFPasswordResetterError.supportedAccountErrorCodes.contains(code):
+            return message
+        default:
+            return CommonStrings.genericErrorDescription
         }
     }
+
+    private static let supportedAccountErrorCodes = [
+        "ratelimited",
+        "noemail"
```
nambatee: should we be selective here or allow all errors to be shown? I thought html could creep in like here `https://phabricator.wikimedia.org/T217186` but not sure if that would ever be the case

montehurd: Even if it has html/wikitext it's prob good to show... hopefully these continue to be improved upstream...

joewalsh: I think we can allow all errors to be shown if they're parsed properly - It looks like it would need to be parsed initially with `action=parse` then we could convert to an `NSAttributedString` from HTML with links

montehurd: Oh... so if the error string has wikitext hit the api to convert to html?

joewalsh: Yeah - but that could be handled separate from this PR. I think for this PR just pass whatever error message MW is sending back, then the fix for https://phabricator.wikimedia.org/T217186  will handle the issue of wikitext and html in errors

Wikipedia/Code/WMFAuthenticationManager.swift
```diff
@@ -17,14 +17,14 @@ public class WMFAuthenticationManager: Fetcher {
...
-                return CommonStrings.genericErrorDescription
+                return WMFLocalizedString("error-generic-description", value: "An unexpected error occurred", comment: "Generic error message for when the error isn't recoverable by the user.")
             }
         }
         
         public var recoverySuggestion: String? {
             switch self {
             default:
-                 return CommonStrings.genericErrorRecoverySuggestion
+                 return WMFLocalizedString("error-generic-recovery-suggestion", value: "Please try again later", comment: "Generic recovery suggestion for when the error isn't recoverable by the user.")
```
joewalsh: Out of the scope of this PR, but seems like either we should unify on the description including the recovery suggestion (as it is in `RequestError`) or splitting the description or recovery suggestion everywhere (as it is here).

PR #2978, opened 2019-03-18
Wikitext scrolled out of view when exiting out of Find in page

nambatee:
https://phabricator.wikimedia.org/T218480

Let me know if there's a better way to do this - ideally, we would keep the current scroll position instead of scrolling to last cursor position and then scrolling back to the match


montehurd:
Looks like something isn't getting reset between searches:

rrrrrrr mov

Note that it works fine if you tap *above*:
wwefwefwfw mov

Suspect it could have something to do with https://github.com/wikimedia/wikipedia-ios-codemirror/pull/6/files#diff-437e391dc6434e078c5157b0c10cf0d9R264

Wikipedia/assets/codemirror/codemirror-index.html
```diff
@@ -819,6 +819,23 @@
       focus: () => {
         editor.focus()
       },
+      focusWithoutScroll: () => {
+        const scrollY = window.scrollY
+        editor.focus()
```
montehurd: Might consider using this instead of `editor.focus()`...
<pre>
editor.display.input.focus({
     preventScroll: true
})
</pre>
... it bypasses a bunch of codemirror code that doesn't need to be called in this case (i think) and when webkit implements `preventScroll` we'd get it for free automatically.

If that sounds good may want a comment like this as a reminder for us later:

``
'preventScroll' is used here so when webkit starts supporting it we'll no longer have a jump. see:
https://html.spec.whatwg.org/multipage/interaction.html#dom-focusoptions-preventscroll
At that time we'll no longer need the `window.scrollTo` call below.
``

Wikipedia/assets/codemirror/codemirror-index.html
```diff
@@ -819,6 +819,23 @@
...
+        const state = editor.state.search
+        if (!state) {
+          return
+        }
+        const from = state.posFrom
+        const to = state.posTo
+        if (!from || !to) {
+          return
+        }
+        editor.setSelection(from, to)
```
montehurd: May want the `scroll` param with `setSelection` here just in case code mirror is doing some internal scroll-y bits...
``editor.setSelection(from, to, {scroll: false})``

montehurd:
Oh man... if this didn't need to set focus we could just send the same message to native land that I used in `scrollRangeIntoViewIfNeeded` 😭

montehurd:
No better way than your approach (my comments above are just tweaks) jumps out at me though.

Wikipedia/assets/codemirror/codemirror-index.html
montehurd: Could also try `editor.setCursor(from)` here... might prevent the "cut copy paste" menu from appearing... though `editor.setSelection(from, null, {scroll: false})` might work for that too...

haha we have many options... none of them good 😂

Wikipedia/assets/codemirror/codemirror-index.html
nambatee: Calling `focus` on `input` will always go through CodeMirror because of https://github.com/wikimedia/wikipedia-ios/blob/develop/Wikipedia/assets/codemirror/resources/lib/codemirror/lib/codemirror.js#L8853

It could take options and pass that to `this.div.focus()` but given that it doesn't do anything right now, I think it's not worth messing with `codemirror.js`. What do you think?

nambatee:
@montehurd thanks for catching the reset part, I'll address that after Toni's Find & Replace follow-up work gets merged to avoid conflicts

nambatee:
@carolynlimadeo

when user closes Find & Replace, should we select the last match or set the cursor at its beginning?

a) selection

sel

b) cursor

curs

Wikipedia/assets/codemirror/codemirror-index.html
montehurd: ya prob not worth it 🧐

carolynlimadeo:
@nambatee let's do the last match that was highlighted in the Find and Replace view.

PR #2980, opened 2019-03-19
Find & replace bug fixes

tonisevener:
https://phabricator.wikimedia.org/T218444

Logic to highlight next match after cursor instead of first match in doc + workarounds for additional bugs.

https://github.com/wikimedia/wikipedia-ios-codemirror/pull/7


Wikipedia/assets/codemirror/resources/addon/search/search.js
```diff
@@ -30,6 +34,15 @@
         var match = query.exec(stream.string);
         if (match && match.index == stream.pos) {
           stream.pos += match[0].length || 1;
+
+          var fromCursor = cm.getCursor('from')
+          if (stream.lineOracle.line > fromCursor.line || (stream.lineOracle.line == fromCursor.line && stream.start >= fromCursor.ch)) {
```
tonisevener: This is the main logic for determining if this match is the first one after the cursor and saving the index in a state property. They were just available values I saw to me in a breakpoint, not determined from any documentation...but it seems to work pretty well. It's worth noting that our find & replace implementation doesn't lean on setting the selection to highlight the match, so these are workarounds for that existing implementation. Seems like the way our native menus react are based on selection so I didn't want to hijack that to match the codemirror find/replace demo.

Wikipedia/assets/codemirror/resources/addon/search/search.js
```diff
@@ -255,16 +283,20 @@
...
             return;
           }
         }
-        
+
         state.posFrom = cursor.from(); state.posTo = cursor.to();
-        focusOnMatch(state)
+        focusOnMatch(state);
         if (shouldReplace) {
+          cm.isReplacing = true;
```
tonisevener: isReplacing is to get around another bug - intermittently (after looping back around from the end) codemirror would fire a cursorActivity event and our native side which would clear out find/replace (assuming the user tapped the editor).

Wikipedia/assets/codemirror/resources/addon/search/search.js
```diff
@@ -30,6 +34,15 @@
         var match = query.exec(stream.string);
         if (match && match.index == stream.pos) {
           stream.pos += match[0].length || 1;
+
+          var fromCursor = cm.getCursor('from')
+          if (stream.lineOracle.line > fromCursor.line || (stream.lineOracle.line == fromCursor.line && stream.start >= fromCursor.ch)) {
+              if (state.initialFocusedMatchIndex == -1) {
```
nambatee: would it make make sense to make it the first check to avoid checking `stream.lineOracle.line > fromCursor.line...` if `initialFocusedMatchIndex` is not -1?

Wikipedia/assets/codemirror/resources/addon/search/search.js
```diff
@@ -82,7 +95,7 @@
       state.queryText = query;
       state.query = parseQuery(query);
       cm.removeOverlay(state.overlay, queryCaseInsensitive(state.query));
-      state.overlay = searchOverlay(state.query, queryCaseInsensitive(state.query));
+      state.overlay = searchOverlay(cm, state, state.query, queryCaseInsensitive(state.query));
```
nambatee: now that `searchOverlay` is taking the entire `state`, it could get `query` itself instead of taking it as a separate parameter

Wikipedia/assets/codemirror/resources/addon/search/search.js
tonisevener: Yep agreed, done!

Wikipedia/assets/codemirror/resources/addon/search/search.js
tonisevener: Yep good catch, this is done too.

Wikipedia/assets/codemirror/resources/addon/search/search.js
```diff
@@ -35,11 +36,11 @@
...
-              if (state.initialFocusedMatchIndex == -1) {
-                state.initialFocusedMatchIndex = loopMatchPositionIndex;
-              }
+          if (state.initialFocusedMatchIndex == -1) {
+            var fromCursor = cm.getCursor('from')
+            if (stream.lineOracle.line > fromCursor.line || (stream.lineOracle.line == fromCursor.line && stream.start >= fromCursor.ch)) {
+              state.initialFocusedMatchIndex = loopMatchPositionIndex;
+             }
           }
           loopMatchPositionIndex++;
```
nambatee: I think `loopMatchPositionIndex++` could go inside the first if too

PR #2982, opened 2019-03-20
Fix WebKit exception crash

nambatee:
Potentially https://phabricator.wikimedia.org/T217936

The most frequent 6.2.0 WebKit crash happens on this line:

https://github.com/wikimedia/wikipedia-ios/blob/f5c4dc84cdadb53719ee855816edf5ba58dd920e/Wikipedia/Code/WMFURLSchemeHandler.m#L76

[Documentation for `didFinish`](https://developer.apple.com/documentation/webkit/wkurlschemetask/2890839-didreceiveresponse?language=objc) says that an exception can occur if:
a) you try to finish the task before sending a response
b) you try to mark the task as completed after it has already been completed
c) **you try to call this method after the handler has been told to stop loading this task**

Since we always send a response before we mark the task as finished (a) and we don't mark it as complete more than once (b), it seems like the exception might be caused by calling `didFinish` after the task was stopped (c).

When [`stopURLSchemeTask`](https://github.com/wikimedia/wikipedia-ios/blob/f5c4dc84cdadb53719ee855816edf5ba58dd920e/Wikipedia/Code/WMFURLSchemeHandler.m#L442) is called, we mark the task as inactive via [`dispatch_barrier_async`](https://github.com/wikimedia/wikipedia-ios/blob/f5c4dc84cdadb53719ee855816edf5ba58dd920e/Wikipedia/Code/WMFURLSchemeHandler.m#L444).
Since [`dispatch_barrier_async`](https://developer.apple.com/documentation/dispatch/1452797-dispatch_barrier_async) waits for all the other blocks on the queue to finish executing before it executes, if [`stopURLSchemeTask`](https://github.com/wikimedia/wikipedia-ios/blob/f5c4dc84cdadb53719ee855816edf5ba58dd920e/Wikipedia/Code/WMFURLSchemeHandler.m#L442) is called [after we check whether task was active or while we're checking whether the task is active](https://github.com/wikimedia/wikipedia-ios/blob/f5c4dc84cdadb53719ee855816edf5ba58dd920e/Wikipedia/Code/WMFURLSchemeHandler.m#L63), we could end up calling any of the [`WKURLSchemeTask`](https://developer.apple.com/documentation/webkit/wkurlschemetask?language=objc) methods on a task that was stopped.

If the above is correct, would the lowest impact fix for now be checking if the task if still active right before calling any of the [`WKURLSchemeTask`](https://developer.apple.com/documentation/webkit/wkurlschemetask?language=objc) methods?


joewalsh:
Part of the problem is that `webView:stopURLSchemeTask:` will return before the task is removed from active tasks because it's using `_async` instead of `_sync`:
```
- (void)webView:(nonnull WKWebView *)webView stopURLSchemeTask:(nonnull id<WKURLSchemeTask>)urlSchemeTask {
    DDLogDebug(@"stopURLSchemeTask %@", urlSchemeTask);
    dispatch_barrier_async(self.activeTaskQueue, ^{
        [self.activeTasks removeObject:urlSchemeTask];
    });
}
```

It should be safe to assume that WebKit won't throw an exception if the return from `webView:stopURLSchemeTask:` is blocked until any in-flight callbacks finish. To achieve that, all the interaction with `task` in `finishTask:withResponse:data:error:` could be wrapped in `dispatch_sync`. This way it's ensured that if a task is finishing while `webView:stopURLSchemeTask:` is called, `webView:stopURLSchemeTask:` won't return before the callbacks are called. The removal from `activeTaskQueue` should still be inside of a `barrier_async`:
```
- (void)finishTask:(id<WKURLSchemeTask>)task withResponse:(nullable NSURLResponse *)response data:(nullable NSData *)data error:(nullable NSError *)error {
    dispatch_sync(self.activeTaskQueue, ^{
        if (![self.activeTasks containsObject:task]) {
            return;
        }
        if (error) {
            [task didFailWithError:error];
        } else if (response) {
            [task didReceiveResponse:response];
            if (data) {
                [task didReceiveData:data];
            }
            [task didFinish];
        } else {
            [task didFailWithError:[WMFFetcher unexpectedResponseError]];
        }
        dispatch_barrier_async(self.activeTaskQueue, ^{
            [self.activeTasks removeObject:task];
        });
    });
}
```

nambatee:
Makes sense! 👍 I thought the documentation implied that "after this method is called" (in relation to `stopURLSchemeTask`) meant "as soon as WebKit calls this method" instead of "as soon as this method actually returns". So the task is only marked as stopped after the method returns, correct? 🧐

joewalsh:
>the task is only marked as stopped after the method returns

I'm assuming that's the case, otherwise there's not really anything we can do. If they mark it as stopped internally and reject any other callbacks on task before calling or while calling `stopURLSchemeTask`, there will always be the potential for a race condition

joewalsh:
This looks like the relevant code on their end: https://svn.webkit.org/repository/webkit/trunk/Source/WebKit/UIProcess/Cocoa/WebURLSchemeHandlerCocoa.mm

nambatee:
no more changed coming in, ready for review

PR #2983, opened 2019-03-20
Switch themes and syntax highlighting using classes

joewalsh:
Follow up due to https://phabricator.wikimedia.org/T218303#5040430

Before this PR,  longer layouts (red bars) were triggered on theme change and when switching syntax highlighting on:
Screen Shot 2019-03-20 at 12 24 46 PM

After this change, it's quicker:
Screen Shot 2019-03-20 at 1 37 19 PM

Benchmarks taken on an iPhone X, "History" section of the SpaceX article

### DIFF
--- a/Wikipedia/assets/codemirror/codemirror-index.html
+++ b/Wikipedia/assets/codemirror/codemirror-index.html
@@ -9,9 +9,11 @@
   <script src="resources/mode/mediawiki/mediawiki.js"></script>
   <link rel="stylesheet" type="text/css" href="resources/mode/mediawiki/mediawiki.css">
   <link rel="stylesheet" type="text/css" href="codemirror-common.css">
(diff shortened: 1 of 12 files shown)


Wikipedia/assets/codemirror/codemirror-index.html
```diff
@@ -372,13 +374,16 @@
     }
         
     const applyTheme = (themeName) => {
-      document.getElementById('codemirror-theme').setAttribute('href', `codemirror-${themeName}.css`)  
+      document.body.className = `cm-mw-${themeName}`
```
montehurd: Would this be a good time to check if the built-in them switching mechanism is more performant? 
`editor.setOption("theme", theme)`

joewalsh: This is the same approach as `setOption("theme", theme)` but also handles setting the background color of the body to prevent any color flash on startup

PR #2987, opened 2019-03-20
Localized & accessible find & replace

tonisevener:
https://phabricator.wikimedia.org/T218425


Wikipedia/Localizations/en.lproj/Localizable.strings
```diff
@@ -328,7 +330,13 @@
 "field-token-title" = "Verification code";
 "field-username-placeholder" = "enter username";
 "field-username-title" = "Username";
-"find-in-page-number-matches" = "$1 / $2";
+"find-clear-button-accessibility-label" = "Clear find textfield";
```
montehurd: I think `textfield` should maybe be separate words. Same with `find-textfield-accessibility-label`.

tonisevener: actually I noticed VoiceOver says "textfield" anyways for the component so it said it twice. I removed in a later commit :) Let me know if you notice any others, I think all the remaining textfields are in comments.

WMF Framework/CommonStrings.swift
```diff
@@ -168,4 +168,5 @@ public class CommonStrings: NSObject {
     
     public static let genericErrorDescription = WMFLocalizedString("fetcher-error-generic", value: "Something went wrong. Please try again later.", comment: "Error shown to the user for generic errors with no clear recovery steps for the user.")
     
+    public static let findAndReplaceHeaderTitle = WMFLocalizedString("find-replace-header", value: "Find and replace", comment: "Find and replace header title.")
```
nambatee: would it make sense to keep this string in `SectionEditorViewController`? looks like it's only used in that file

Wikipedia/Localizations/en.lproj/Localizable.strings
montehurd: oh good catch!

Wikipedia/Code/FindAndReplaceKeyboardBar.swift
```diff
@@ -289,17 +304,24 @@ private extension FindAndReplaceKeyboardBar {
         nextButton.isEnabled = total >= 2
     }
     
+    func replaceTypeText() -> String {
```
nambatee: if it makes it easier, this could be a computed `text` property inside `ReplaceType` and then you could do `replaceType.text` or `accessibilityTex`, but either way is 👍

Wikipedia/Code/SectionEditorViewController.swift
```diff
@@ -84,32 +84,36 @@ class SectionEditorViewController: UIViewController {
...
+
+        view.addSubview(focusNavigationView)
         
-        navigationController?.navigationBar.leadingAnchor.constraint(equalTo: focusNavigationView.leadingAnchor).isActive = true
-        navigationController?.navigationBar.trailingAnchor.constraint(equalTo: focusNavigationView.trailingAnchor).isActive = true
-        focusNavigationViewBottomConstraint = navigationController?.navigationBar.bottomAnchor.constraint(equalTo: focusNavigationView.bottomAnchor, constant: focusNavigationView.frame.height)
-        focusNavigationViewBottomConstraint.isActive = true
+        view.leadingAnchor.constraint(equalTo: focusNavigationView.leadingAnchor).isActive = true
+        view.trailingAnchor.constraint(equalTo: focusNavigationView.trailingAnchor).isActive = true
+        view.safeAreaLayoutGuide.topAnchor.constraint(equalTo: focusNavigationView.topAnchor).isActive = true
```
nambatee: if these can be activated in one pass via `activateConstraints`, it might be more efficient according to https://developer.apple.com/documentation/uikit/nslayoutconstraint/1526955-activate

> Typically, using this method is more efficient than activating each constraint individually.

Wikipedia/Code/SectionEditorViewController.swift
```diff
@@ -152,7 +156,12 @@ class SectionEditorViewController: UIViewController {
         webView.inputViewsSource = inputViewsController
         
         webView.translatesAutoresizingMaskIntoConstraints = false
-        view.wmf_addSubviewWithConstraintsToEdges(webView)
+        view.addSubview(webView)
+        view.leadingAnchor.constraint(equalTo: webView.leadingAnchor).isActive = true
+        view.trailingAnchor.constraint(equalTo: webView.trailingAnchor).isActive = true
+        webViewTopConstraint = view.safeAreaLayoutGuide.topAnchor.constraint(equalTo: webView.topAnchor)
+        webViewTopConstraint.isActive = true
+        view.bottomAnchor.constraint(equalTo: webView.bottomAnchor).isActive = true
```
nambatee: same as above with `activateConstraints` unless these need to be set 1 by 1

WMF Framework/CommonStrings.swift
tonisevener: yep you're right, done!

Wikipedia/Code/FindAndReplaceKeyboardBar.swift
tonisevener: good idea, done!

Wikipedia/Code/SectionEditorViewController.swift
tonisevener: another good idea, also looks a bit cleaner IMO. done!

Wikipedia/Code/FindAndReplaceKeyboardBar.swift
```diff
@@ -289,17 +304,24 @@ private extension FindAndReplaceKeyboardBar {
...
-        //TODO: Localize
         let count = replaceTextField.text?.count ?? 0
-        replacePlaceholderLabel.text = "Replace with..."
-        
-        var replaceTypeText: String
-        switch replaceType {
-        case .replaceSingle: replaceTypeText = "Replace"
-        case .replaceAll: replaceTypeText = "Replace all"
-        }
+        replacePlaceholderLabel.text = WMFLocalizedString("replace-textfield-placeholder", value: "Replace with...", comment: "Placeholder text seen in replace textfield before textfield is focused.")
```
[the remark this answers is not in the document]
tonisevener: done!

Wikipedia/Code/FocusNavigationView.swift
```diff
@@ -6,21 +6,33 @@ protocol FocusNavigationViewDelegate: class {
...
+    func configure(headerText: String? = nil, headerAccessibilityText: String? = nil, closeButtonAccessibilityText: String? = nil, traitCollection: UITraitCollection) {
+       
+        if let headerText = headerText {
+            titleLabel.text = headerText
+        }
         
-        if let text = text {
-            titleLabel.text = text
+        if let headerAccessibilityText = headerAccessibilityText {
+            titleLabel.accessibilityLabel = headerAccessibilityText
```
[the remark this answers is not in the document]
tonisevener: agreed, done!

Wikipedia/Code/FindAndReplaceKeyboardBar.swift
```diff
@@ -86,7 +86,9 @@ final class FindAndReplaceKeyboardBar: UIInputView {
         hideUndoRedoIcons()
         previousButton.isEnabled = false
         nextButton.isEnabled = false
-        closeButton.accessibilityLabel = CommonStrings.closeButtonAccessibilityLabel
+
```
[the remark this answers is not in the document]
tonisevener: @joewalsh hm strange, it's moving in order on my iPad - what device were you testing on?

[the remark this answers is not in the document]
joewalsh: iPhone 6S - I'll double check

PR #2989, opened 2019-03-20
Find & Replace - undo to the end of history clears out all text

nambatee:
https://phabricator.wikimedia.org/T218696

When undoing, if the cursor doesn't change its position, `cursorActivity` won't fire and `sendNativeMessages` won't be called. As a result, we won't get an update about button states.

Calling `sendNativeMessages` on `change` does mean that we'll get more updates than we need


nambatee:
this could also fix https://phabricator.wikimedia.org/T218698

www/codemirror/codemirror-index.html
```diff
@@ -407,8 +407,18 @@
           }, 25)
         }
       })
+      editor.on('change', (doc) => {
+        if (!doc.isReplacing) {
```
nambatee: this could be consolidated into 1 method but not sure what to call it 🤔
```
        if (!doc.isReplacing) {
          sendNativeMessages(doc)

          clearTimeout(cursorActivityScrollTimer)
          cursorActivityScrollTimer = setTimeout(() => {
            scrollCursorIntoViewIfNeeded()        
          }, 25)
        }
```

Wikipedia/assets/codemirror/codemirror-index.html
```diff
@@ -407,8 +407,18 @@
           }, 25)
         }
       })
+      editor.on('change', (doc) => {
```
montehurd: hmm i think onchange might be too expensive as it will happen every time a character is typed... 🤔 can we just call sendNativeMessages once when the find/replace interface is dismissed?

nambatee: it's called then but it needs to be called in between undos

nambatee: maybe I'll just call it after each undo

montehurd: oops wrong bug 🤣

in that case may want to just call it after the undo call?

Wikipedia/assets/codemirror/codemirror-index.html
```diff
@@ -397,28 +397,20 @@
...
 
-          clearTimeout(cursorActivityScrollTimer)
-          cursorActivityScrollTimer = setTimeout(() => {
-            scrollCursorIntoViewIfNeeded()        
-          }, 25)
-        }
-      })
-      editor.on('change', (doc) => {
-        if (!doc.isReplacing) {
+    const notifyNativeAboutChangesAndScrollIfNeeded = (doc) => {
```
nambatee: name suggestions welcome

montehurd: sounds about right!

PR #2990, opened 2019-03-21
RandomArticleWidget

kemayo:
This is a first pass of a Random Article widget. It's somewhat naïve, and may well require at least some sort of throttling so it doesn't attempt to reload a new article every single time it's viewed.

### DIFF
--- a/RandomArticleWidget/RandomArticleWidget.swift
+++ b/RandomArticleWidget/RandomArticleWidget.swift
@@ -1,0 +1,164 @@
+//
+//  RandomArticleWidget.swift
+//  RandomArticleWidget
+//
+//  Created by David Lynch on 2/5/19.
+//  Copyright © 2019 Wikimedia Foundation. All rights reserved.
+//
+
+import UIKit
+import NotificationCenter
+import WMF
+
+class RandomArticleWidget: UIViewController, NCWidgetProviding {
+    let randomArticleFetcher = RandomArticleFetcher()
+    let articlePreviewFetcher = WMFArticlePreviewFetcher()
+    let collapsedArticleView = ArticleRightAlignedImageCollectionViewCell()
+    let expandedArticleView = ArticleFullWidthImageCollectionViewCell()
+
+    var isExpanded = true
+    
+    var isEmptyViewHidden = true {
+        didSet {
+            collapsedArticleView.isHidden = !isEmptyViewHidden
+            expandedArticleView.isHidden = !isEmptyViewHidden
+        }
+    }
+    
+    var article: WMFArticle?
+//    var currentArticleKey: String?
+    
+    var dataStore: MWKDataStore? {
+        return SessionSingleton.sharedInstance()?.dataStore
+    }
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        // Do any additional setup after loading the view from its nib.
+        
+        view.translatesAutoresizingMaskIntoConstraints = false
+        
+        let tapGR = UITapGestureRecognizer(target: self, action: #selector(handleTapGesture(_:)))
+        view.addGestureRecognizer(tapGR)
+        
+        collapsedArticleView.frame = view.bounds
+        view.addSubview(collapsedArticleView)
+        
+        expandedArticleView.saveButton.addTarget(self, action: #selector(saveButtonPressed), for: .touchUpInside)
+        expandedArticleView.frame = view.bounds
+        view.addSubview(expandedArticleView)
+        
+        extensionContext?.widgetLargestAvailableDisplayMode = .expanded
+    }
+        
+    func widgetPerformUpdate(completionHandler: (@escaping (NCUpdateResult) -> Void)) {
+        // Perform any setup necessary in order to update the view.
+        
+        // If an error is encountered, use NCUpdateResult.Failed
+        // If there's no update required, use NCUpdateResult.NoData
+        // If there's an update, use NCUpdateResult.NewData
+        
+        defer {
+            updateView()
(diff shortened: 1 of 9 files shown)


RandomArticleWidget/RandomArticleWidget.swift
```diff
@@ -0,0 +1,164 @@
...
+        }
+        
+        let siteURL = SessionSingleton.sharedInstance().currentArticleSiteURL
+        self.randomArticleFetcher.fetchRandomArticle(withSiteURL: siteURL!, completion: { (error, result, summary) in
+            if (error != nil) {
+                completionHandler(.failed)
+                return
+            }
+            
+            self.articlePreviewFetcher.fetchArticlePreviewResults(forArticleURLs: [result!], siteURL: siteURL!, completion: { (searchResults) in
```
joewalsh: The random article fetcher returns the `summary` so the `articlePreviewFetcher` can be removed - the `WMFArticle` can be created with:
```
            DispatchQueue.main.async {
                guard
                    let articleURL = articleURL,
                    let summary = summary,
                    let article = self.dataStore?.viewContext.fetchOrCreateArticle(with: articleURL)
                else {
                    completionHandler(.failed)
                    return
                }

                article.update(withSummary: summary)
                ...
             }  
```

RandomArticleWidget/RandomArticleWidget.swift
```diff
@@ -0,0 +1,164 @@
...
+    func widgetPerformUpdate(completionHandler: (@escaping (NCUpdateResult) -> Void)) {
+        // Perform any setup necessary in order to update the view.
+        
+        // If an error is encountered, use NCUpdateResult.Failed
+        // If there's no update required, use NCUpdateResult.NoData
+        // If there's an update, use NCUpdateResult.NewData
+        
+        defer {
+            updateView()
+        }
```
joewalsh: This `defer` block will be called at the end of this function but before the random article is fetched - it should probably move into the return block of `fetchRandomArticle`

RandomArticleWidget/RandomArticleWidget.swift
```diff
@@ -0,0 +1,164 @@
...
+        
+        let siteURL = SessionSingleton.sharedInstance().currentArticleSiteURL
+        self.randomArticleFetcher.fetchRandomArticle(withSiteURL: siteURL!, completion: { (error, result, summary) in
```
joewalsh: For something like this it's better to `guard let siteURL = SessionSingleton.sharedInstance().currentArticleSiteURL else { ...failure case... return }` to prevent the force unwrap `!` which will crash if `siteURL` is nil

RandomArticleWidget/RandomArticleWidget.swift
```diff
@@ -0,0 +1,164 @@
+//
+//  RandomArticleWidget.swift
+//  RandomArticleWidget
+//
+//  Created by David Lynch on 2/5/19.
+//  Copyright © 2019 Wikimedia Foundation. All rights reserved.
+//
```
joewalsh: These comment headers can be deleted

RandomArticleWidget/RandomArticleWidget.swift
kemayo: Ah, interesting. I know I'd tried to pass that into the `fetchOrCreateArticle updatedWith` bit and run into a type mismatch there because of it being `[String: Any]`. I think I just assumed that all the `update` methods would thus require a `MWKSearchResult`, and so didn't see this was an option.

joewalsh: Yeah, the problem here is how unclear those parts of the codebase are. We're in the process of clarifying those methods to make it clear that summary is the preferred method as part of https://phabricator.wikimedia.org/T217411 .

Wikipedia/Code/NSDictionary+WMFCommonParams.m
```diff
@@ -25,7 +25,7 @@ + (instancetype)wmf_titlePreviewRequestParametersWithExtractLength:(NSUInteger)e
                                          //@"any", @"pilicense",
                                          imageWidth, @"pithumbsize",
                                          // revision
-                                         @(1), @"rrvlimit",
+                                         //@(1), @"rrvlimit",
```
kemayo: I commented this out because it was causing `"Unrecognized parameter: rrvlimit."` (on generated API urls like [this](https://en.wikipedia.org/w/api.php?prop=description%7Cpageimages%7Cpageprops%7Crevisions%7Cextracts%7Ccoordinates&ppprop=ns%7Cdisplaytitle&rrvlimit=1&piprop=thumbnail&pithumbsize=640&titles=Truxillic%20acid&exchars=525&pilimit=1&rvprop=ids&continue=&exintro=1&action=query&format=json&exlimit=1&explaintext=)), and the fetcher code treated its presence as a complete-failure rather than just a warning. Then I saw an earlier commit (6daf9e58a976e07b4f1313ddace8f2db019caba9) which had commented out the parameter elsewhere, and so...

I could probably remove this from the pull request, given that 088dbc88bf76baea69f274df3b7feedf28740b65 removed the path that fetched the article preview anyway.

joewalsh: It's OK either way,`WMFArticlePreviewFetcher` will be removed entirely, it's only used in one other place. I added it to the list for https://phabricator.wikimedia.org/T217411

joewalsh:
Closing as old-style widgets are deprecated in iOS 14. Could be worth re-writing this for the new widget system in iOS 14

PR #3003, opened 2019-03-26
Box briefly appears around recent search cell text if text itself is tapped

nambatee:
https://phabricator.wikimedia.org/T216221


joewalsh:
This is low impact enough we could probably pull it into 6.2.1, but let's confirm at planning

PR #3005, opened 2019-03-26
Fix clear formatting on headers

joewalsh:
As per discussion in weekly planning, this PR makes the clear formatting butting clear header markup when any part of a header is selected


tonisevener:
A couple of things I'm noticing - 

Article "Diamonds Are Forever"  - the first paragraph has remove markup disabled. Maybe related to that template embedded earlier in the line but formatting after the template should be removable.

Simulator Screen Shot - iPhone X - 2019-03-26 at 14 30 32

I see some characters removed when removing formatting. I tested on italics, bold, and links -
out

joewalsh:
@tonisevener the first issue I'm seeing on develop as well so I can file that as a follow-on issue. The second issue is definitely a regression caused by this PR and I've pushed a fix

joewalsh:
@tonisevener it looks like the first issue was caused by the template changes, so I reverted that and made disabling the clear formatting button in templates a follow on task: https://phabricator.wikimedia.org/T219321

PR #3008, opened 2019-03-28
Add animations to editing welcome screens

nambatee:
https://phabricator.wikimedia.org/T218381


tonisevener:
👏Looks great! I also tested a few simulators, only thing I found is if I rotate at just the right moment the papers get thrown off until you rotate again. It doesn't happen on the other animations. Maybe you could add one last check to move them to the destination origin after rotation if they don't seem to be in the expected place. Seems like an edge case though, not a big deal if there's not an easy way to fix.

rotation

nambatee:
thanks for catching that @tonisevener! can you reproduce it with the latest change?

nambatee:
also - updated an asset for the 3rd screen ("Cite reliable sources"), the old one had unnecessary shadow

tonisevener:
@nambatee looks good to me now! 👍

PR #3011, opened 2019-03-29
Remove Article Summary fetching from session, make separate Summary fetcher

joewalsh:
One of the bullet points in https://phabricator.wikimedia.org/T217411 

Ran benchmarking to confirm that summary parsing and updating is roughly the same speed, but reverted it it as I'm not sure we need to run it with every test pass. While going from JSON data --> codable ArticleSummary is 2x slower than JSON data --> [String: Any], the full process of parsing and updating the `WMFArticle` object is roughly the same.

Hold tag is pending 6.2.1 branch split so that develop can continue with 6.3 work.


joewalsh:
6.2.1 branch is split, merging

PR #3012, opened 2019-04-01
[Design bug] After using highlight to edit, the cursor is placed after the highlighted word

nambatee:
https://phabricator.wikimedia.org/T219521

### DIFF
--- a/www/codemirror/codemirror-editTextSelection.js
+++ b/www/codemirror/codemirror-editTextSelection.js
@@ -111,15 +111,6 @@ const getWikitextRangeToSelect = (wikitextBeforeSelection, wikitextSelection) =>
 
 const scrollToAndHighlightRange = (range, codemirror) => {
   codemirror.setSelection(range.from, range.to, {scroll: false})
-
-  let marker = null
-  marker = codemirror.markText(range.from, range.to, {
-    css: 'background-color: rgba(255, 204, 51, 0.4)', // Can use 'className' (vs 'css') if needed.
-    clearOnEnter: true,
-    inclusiveLeft: true,
-    inclusiveRight: true
-  })
-
   setTimeout(() => { // Slight pause needed to ensure keyboard height is accounted for.
     scrollRangeIntoViewIfNeeded(range.from, range.to)
   }, 250)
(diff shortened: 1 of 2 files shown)


Wikipedia/Code/SectionEditorViewController.swift
```diff
@@ -346,31 +341,26 @@ extension SectionEditorViewController: UIScrollViewDelegate {
...
+                self.needsSelectLastSelection = true
+                vc.theme = self.theme
+                vc.section = self.section
+                vc.wikitext = wikitext
+                vc.delegate = self
+                vc.funnel = self.editFunnel
+                self.navigationController?.pushViewController(vc, animated: true)
             }
-        } else {
-            let message = WMFLocalizedString("wikitext-preview-changes-none", value: "No changes were made to be previewed.", comment: "Alert text shown if no changes were made to be previewed.")
```
joewalsh: Seems like this should stay and at least do a string comparison to the original wikitext to verify that something was changed - currently it allows the user to go through the process and fails silently

nambatee: the `progressButton` should be disabled if no changes were made - what steps do you take to go through and fail silently?

www/codemirror/codemirror-editTextSelection.js
```diff
@@ -110,7 +110,7 @@ const getWikitextRangeToSelect = (wikitextBeforeSelection, wikitextSelection) =>
 }
 
 const scrollToAndHighlightRange = (range, codemirror) => {
-  codemirror.setCursor(range.to)
+  codemirror.setSelection(range.from, range.to, {scroll: false})
 
   let marker = null
   marker = codemirror.markText(range.from, range.to, {
```
joewalsh: `markText` can be removed since the text is highlighted by the selection now

nambatee: makes sense to me but given https://phabricator.wikimedia.org/T215448#5012000 - maybe we can remove it after confirming with Carolyn? I added a comment to the ticket https://phabricator.wikimedia.org/T219521#5075653

PR #3013, opened 2019-04-01
Check if wikitext was changed

nambatee:


Wikipedia/Code/SectionEditorViewController.swift
```diff
@@ -366,16 +364,16 @@ extension SectionEditorViewController: SectionEditorNavigationItemControllerDele
...
             }
-        } else {
-            let message = WMFLocalizedString("wikitext-preview-changes-none", value: "No changes were made to be previewed.", comment: "Alert text shown if no changes were made to be previewed.")
-            WMFAlertManager.sharedInstance.showAlert(message, sticky: false, dismissPreviousAlerts: true)
         }
     }
     
     func sectionEditorNavigationItemController(_ sectionEditorNavigationItemController: SectionEditorNavigationItemController, didTapCloseButton closeButton: UIBarButtonItem) {
-        delegate?.sectionEditorDidFinishEditing(self, withChanges: false)
+        delegate?.sectionEditorDidFinishEditing(self, withChanges: changesMade)
```
joewalsh: This will always be `false` unless the user taps "Next" before tapping close - would it make sense to `getWikitext` here and check if there were changes?

PR #3015, opened 2019-04-01
Add "Clear formatting" button to TextFormattingTableViewController

joewalsh:
https://phabricator.wikimedia.org/T215954#5069873


Wikipedia/Code/TextFormattingTableViewController.swift
```diff
@@ -65,6 +69,21 @@ class TextFormattingTableViewController: TextFormattingProvidingTableViewControl
...
+    
+    private func didSelectClearFormatting() {
+        guard clearFormatting.content.isEnabled else {
+            return
+        }
+        delegate?.textFormattingProvidingDidTapClearFormatting()
+    }
+    
+    private lazy var clearFormatting: Item = {
+        let content = Content(type: .destructiveAction, title: WMFLocalizedString("edit-text-clear-formatting", value: "Clear formatting", comment: "Title for the button that clears formatting from the selected range"), detailText: nil, customView: nil)
```
tonisevener: actually might want to have an accessibility label for this to indicate enabled or disabled. Right now in a disabled state VoiceOver just says "Clear formatting"

joewalsh: Good point! Fixed

PR #3016, opened 2019-04-01
As a user of the Find and Replace feature, I want an easy way to see that replace is working

nambatee:
https://phabricator.wikimedia.org/T219666

https://github.com/wikimedia/wikipedia-ios-codemirror/pull/11


Wikipedia/assets/codemirror/resources/addon/search/search.js
```diff
@@ -129,7 +129,23 @@
...
+      if (state.replacedMarkers) {
+        state.replacedMarkers.push(marker);
+      } else {
+        state.replacedMarkers = [marker];
+      }
+    }
+
+    function clearReplaced(state) {
+      if (state.replacedMarkers) {
+        state.replacedMarkers.forEach((marker) => { marker.clear() });
```
tonisevener: Might want to set replacedMarkers to [] here too, otherwise it'll build up.

nambatee: fixed

Wikipedia/assets/codemirror/resources/addon/search/search.js
```diff
@@ -301,14 +320,15 @@
...
           cm.isReplacing = false;
         }
       }
-      var doReplace = function(match) {
+      var doReplace = function(cm, state) {
         cursor.replace(replaceText);
+        markReplacedText(cm, cursor);
         advance(false);
-        var forceIncrement = replaceText.includes(query) || replaceText.lowercase == query.lowercase;
+        var forceIncrement = replaceText.includes(query) || replaceText.toLowerCase() === query.toLowerCase();
```
nambatee: Fixes wrong focus after single replace -
ezgif-4-953ae6f693aa

tonisevener: ahh good catch, thanks

PR #3026, opened 2019-04-05
Turning off Featured article cards in the feed causes Featured article widget to go blank

nambatee:
https://phabricator.wikimedia.org/T220252

### DIFF
--- a/FeaturedArticleWidget/FeaturedArticleWidget.swift
+++ b/FeaturedArticleWidget/FeaturedArticleWidget.swift
@@ -42,7 +42,7 @@ class FeaturedArticleWidget: UIViewController, NCWidgetProviding {
     }
     
     var article: WMFArticle? {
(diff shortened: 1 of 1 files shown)


FeaturedArticleWidget/FeaturedArticleWidget.swift
```diff
@@ -42,7 +42,7 @@ class FeaturedArticleWidget: UIViewController, NCWidgetProviding {
     }
     
     var article: WMFArticle? {
-        guard let featuredContentGroup = dataStore?.viewContext.newestVisibleGroup(of: .featuredArticle),
+        guard let featuredContentGroup = dataStore?.viewContext.newestGroup(of: .featuredArticle),
```
joewalsh: The filter was on newest visible in order to not show a card in the widget that was individually hidden in the feed by the user. Also, if a user has multiple languages and hides the featured cards of all but one, it would show the non-hidden language. Maybe this could fallback to `newestGroup` only if `newestVisibleGroup` is nil

PR #3027, opened 2019-04-06
Core Data performance

nambatee:
https://phabricator.wikimedia.org/T220605

It seems like linking all the contexts directly to the coordinator is slightly faster (~7%) than our current setup (main context directly linked to the coordinator, background contexts linked to the main context). However, the main problem with this approach is the lack of `NSManagedObjectContextDidSaveNotification` emitted by the main context when changes from background contexts get merged. We rely on those notifications in 2 places:
1) https://github.com/wikimedia/wikipedia-ios/blob/908bd09ae9e8c0907488512b149d5ce049c5a48d/Wikipedia/Code/MWKDataStore.m#L297 and
2) https://github.com/wikimedia/wikipedia-ios/blob/8a36013b3fafeb51558b444dd0de0814962c8fcb/WMF%20Framework/WMFExploreFeedContentController.m#L321
While applying Explore feed preferences in the second case could happen when view context changes, I'm sure not sure about the first case and other future cases that might call for `NSManagedObjectContextDidSaveNotification`.

Apart from this setup, I tested 2 other scenarios:
1) Current setup but without saving changes to the store via `backgroundContextDidSave` - ~4% worse than out current setup, fetches are slower.
2) Current setup but with an additional private context linking directly to the coordinator with the main context as its child - ~%1 worse than our current setup, fetches are slower.

Question:

Is it possible to rely on `NSManagedObjectContextDidChangeNotification` alone?


joewalsh:
For scenario 1. trying to merge from `NSManagedObjectContextDidChangeNotification` might be a problem because it's syncing the changes across processes (between the widgets and the app) and on `DidChange` the changes aren't in the persistent store. Could the data store and feed content controller listen for `DidSave` from any managed object context that's linked to the persistent store? Maybe data store handles listening to all relevant contexts, synchronizes for thread safety, and emits a separate notification that the persistent store did change.

Wikipedia/Code/MWKDataStore.m
```diff
@@ -467,10 +464,24 @@ - (void)migrateToQuadKeyLocationIfNecessaryWithCompletion:(nonnull void (^)(NSEr
 
 #pragma mark - Background Contexts
 
+- (void)backgroundContextDidSave:(NSNotification *)note {
+    dispatch_async(dispatch_get_main_queue(), ^{
```
joewalsh: It might be that this can't be async and should not allow this method to return until the information is read from the notification. Also, it'd be ideal if this didn't happen on the main thread and had access limited by a semaphore.

WMF Framework/WMFExploreFeedContentController.m
```diff
@@ -72,7 +72,7 @@ - (void)setOperationQueue:(NSOperationQueue *)operationQueue {
 - (void)setDataStore:(MWKDataStore *)dataStore {
     _dataStore = dataStore;
     self.exploreFeedPreferencesUpdateCoordinator = [[ExploreFeedPreferencesUpdateCoordinator alloc] initWithFeedContentController:self];
-    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(viewContextDidSave:) name:NSManagedObjectContextDidSaveNotification object:self.dataStore.viewContext];
+    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(feedImportContextDidSave:) name:WMFFeedImportContextDidSave object:nil];
```
joewalsh: Since the view context can also write changes to the PSC, it should be observed as well

Wikipedia/Code/MWKDataStore.m
```diff
@@ -480,23 +491,13 @@ - (void)performBackgroundCoreDataOperationOnATemporaryContext:(nonnull void (^)(
...
-        }
-    }];
-}
-
 - (NSManagedObjectContext *)feedImportContext {
     WMFAssertMainThread(@"feedImportContext must be created on the main thread");
     if (!_feedImportContext) {
         _feedImportContext = [[NSManagedObjectContext alloc] initWithConcurrencyType:NSPrivateQueueConcurrencyType];
-        _feedImportContext.parentContext = _viewContext;
+        _feedImportContext.persistentStoreCoordinator = _persistentStoreCoordinator;
```
joewalsh: Does this also need `automaticallyMergesChangesFromParent`?

Wikipedia/Code/MWKDataStore.m
```diff
@@ -467,10 +464,24 @@ - (void)migrateToQuadKeyLocationIfNecessaryWithCompletion:(nonnull void (^)(NSEr
...
+        [[NSNotificationCenter defaultCenter] postNotificationName:WMFFeedImportContextDidSave object:note.object userInfo:note.userInfo];
+        [self handleCrossProcessChangesFromContextDidSaveNotification:note];
+    });
+}
+
 - (void)performBackgroundCoreDataOperationOnATemporaryContext:(nonnull void (^)(NSManagedObjectContext *moc))mocBlock {
     WMFAssertMainThread(@"Background Core Data operations must be started from the main thread.");
     NSManagedObjectContext *backgroundContext = [[NSManagedObjectContext alloc] initWithConcurrencyType:NSPrivateQueueConcurrencyType];
-    backgroundContext.parentContext = _viewContext;
+    backgroundContext.persistentStoreCoordinator = _persistentStoreCoordinator;
```
joewalsh: Does this also need `automaticallyMergesChangesFromParent`?

nambatee:
1 known issue which could make this setup unviable has to do with too many `change` notifications being emitted, which causes too many reloads when toggling all cards of 1 type in the feed. I think it has to do with WMFContentGroup objects being refreshed when changes are merged from the store into viewContext via internal `mergeChanges(fromContextDidSave:)` (https://developer.apple.com/documentation/coredata/nsmanagedobjectcontext/1506606-mergechanges)

nambatee:
ready for review

PR #3036, opened 2019-04-11
Small wikitext search refactors

montehurd:
Joe's fix https://github.com/wikimedia/wikipedia-ios/pull/3046 is better than my bc48b5e commit below (which I reverted), but the remaining cleanups (934ed07 and da92997) would be nice to keep.

<strike>
https://phabricator.wikimedia.org/T219389

Coalesces search kickoff until typing has stopped for more that .4 seconds.

**Question:** this seems to alleviate some of the delay - though it's not as fast as article find-in-page. Was thinking if this suffices it would be the lowest impact approach. Still looking into it, but wondering if others think this is good enough for now...

(will update `wikipedia-ios-codemirror` pending question above) 

</strike>


joewalsh:
I don't think it's worth adding a delay. Although it's better when the user types quickly, it's still unclear that anything is happening, and it adds some time before the request actually starts, degrading the experience on shorter articles. If anything, a low impact fix for now would be some sort of spinner or progress indicator so that it's clear that their request is being processed and that find in page isn't just broken

PR #3038, opened 2019-04-15
Tech: Session & SchemeHandler changes for mobile-html

tonisevener:
https://phabricator.wikimedia.org/T217411 - "Bring over updated Session & SchemeHandler from the lite project"

Brought over Callback and SessionDelegate from lite project into Session.swift.
Converted & refactored most of WMFURLSchemeHandler to Swift.


Wikipedia/Code/WMFArticleJSONCompilationHelper.m
```diff
@@ -0,0 +1,104 @@
...
+#import <WMF/NSString+WMFHTMLParsing.h>
+#import <WMF/WMFImageTag.h>
+#import <WMF/WMFImageTag+TargetImageWidthURL.h>
+#import <WMF/NSURL+WMFSchemeHandler.h>
+#import <WMF/MWKSection.h>
+#import <WMF/MWKSectionList.h>
+
+@implementation WMFArticleJSONCompilationHelper
+
++ (nullable NSData *)jsonDataForArticle: (MWKArticle *)article withImageWidth: (NSInteger)imageWidth {
```
tonisevener: I figured this code will probably go away once we move to mobile-html, so I didn't see much point in converting it to Swift. It's from WMFURLSchemeHandler, I just moved it over into a separate helper class.

Wikipedia/Code/WMFURLSchemeHandler.m
```diff
@@ -1,439 +0,0 @@
...
-#pragma mark - BaseURL (for testing only)
-
-- (NSURL *)baseURL {
-    return [[self baseURLComponents] URL];
-}
-
-- (void)webView:(nonnull WKWebView *)webView startURLSchemeTask:(nonnull id<WKURLSchemeTask>)urlSchemeTask {
-    WMFAssertMainThread(@"startURLSchemeTask assumed to be called on the main thread");
-
-    [self.activeTasks addObject:urlSchemeTask];
```
tonisevener: SchemeHandler from the lite project keeps track of the URLSessionTasks returned from Session, but WMFURLSchemeHandler kept track of the URLSchemeTasks. I didn't keep URLSchemeTask tracking when changing to Swift but let me know if that should be preserved.

joewalsh: As long as you can guarantee that no other calls are made on the `WKURLSchemeTask` after `stop urlSchemeTask` is called, it will be fine. Without tracking the `WKURLSchemeTask` you'll need to check for the `URLSessionTask` in the `Session.Callback` blocks. Also, we discovered in https://github.com/wikimedia/wikipedia-ios/pull/2982 that it looks like WebKit is ensuring thread safety on `WKURLSchemeTask` by using the main thread, so there will need to be `DspatchQueue.main.async`s inside of the callback blocks. At that point `tasks` should be access controlled by the main thread as well, eliminating the need for the private `SchemeHandlerQueue`

Wikipedia/Code/SchemeHandler/SchemeHandler.swift
```diff
@@ -0,0 +1,241 @@
...
+            }
+            urlSchemeTask.didFinish()
+            self.activeSchemeTasks.remove(urlSchemeTask)
+        }
+        
+        activeSchemeTasks.add(urlSchemeTask)
+        
+        switch baseComponent {
+        case FileHandler.basePath:
+            fileHandler.handle(pathComponents: pathComponents, requestURL: requestURL, completion: localCompletionBlock)
```
joewalsh: Since this reads from the filesystem, It'd be good to move it (and this entire switch statement) off of the main queue. My guess is the URLCache in the default handler might hit the filesystem as well. Callbacks to the task would have to bounce back to main.

Wikipedia/Code/SchemeHandler/SchemeHandler+BaseSubHandler.swift
```diff
@@ -0,0 +1,29 @@
+import Foundation
+
+extension SchemeHandler {
+    class BaseSubHandler {
```
joewalsh: It's nice to have these separated out 👍

Wikipedia/Code/SchemeHandler/SchemeHandler.swift
tonisevener: agh good point, will do

tonisevener:
I notice sometimes I see a WebKit error "Could not find specified service" in the console, could be related to my changes. Will look into that next.

joewalsh:
It looks like "Could not find specified service" is happening on `develop` as well, Apple staffer in [this thread](https://forums.developer.apple.com/thread/112095) seems wholly unconcerned if there's no strange behavior other than the log message

Wikipedia/Code/SchemeHandler/SchemeHandler+ArticleSectionHandler.swift
```diff
@@ -0,0 +1,73 @@
...
+        required init(cacheDelegate: ArticleSectionHandlerCacheDelegate) {
+            self.cacheDelegate = cacheDelegate
+        }
+        
+        override class var basePath: String? {
+            return "articleSectionData"
+        }
+        private var basePath: String? {
+            return ArticleSectionHandler.basePath
+        }
```
nambatee: is the private `basePath` returning the overridden `basePath` needed? It compiles for me when I take it out (the private one)

tonisevener: good catch, no it's not. I think at some point I was using it in an instance method but missed cleaning this out.

PR #3045, opened 2019-04-17
Update to current long term stable node version (10.15.3) 

joewalsh:
The node version used by the app should be the same as the node version used by the page library. This way, with node version management set up properly, debugging will work using `npm link`. This is on hold pending merge of https://github.com/wikimedia/wikimedia-page-library/pull/187.

- Upgrade node to 10.15.13
- `npm audit fix` vulnerable dependencies
- Update incorrect codemirror dependency version
- Update to re-built version of page library from https://github.com/wikimedia/wikimedia-page-library/pull/187.


joewalsh:
@montehurd yeah, that's expected and the reason for the hold on the PR. Ran the update in cf3fcd4 to verify that the app was working with the changes

montehurd:
Opps didn't read the description closely enough 😂

montehurd:
Changes to local page lib are not showing up when using `npm link` and the `Page Library Development` scheme. I tried using `nvm` to set the version but still no luck...

montehurd:
It think it may be `nvm` related? Now no longer works even on `develop`...

PR #3046, opened 2019-04-18
Combine slow search commands into a single operation in the section editor

joewalsh:
Hold on merge until the upstream change is made and brought in. Test & code review here, then I'll make the change upstream if this is approved.

Combining clearing overlays and adding overlays into one operation reduces the number of forced layout passes from two to one, roughly halving the search time:
Screen Shot 2019-04-18 at 8 35 41 AM

It's still slow on long sections. The long layout pass comes when CodeMirror recalculates the height of the content. Given that the highlight doesn't affect the height of any lines (maybe it could if the styling was different), it seems like the ideal fix would be to prevent the recalculation from happening at all in this case. However, it looks like that would be a more invasive change.


montehurd:
> The long layout pass comes when CodeMirror recalculates the height of the content.

That was the part I couldn't recall when we chatted in 1:1 😄

Wikipedia/assets/codemirror/resources/addon/search/search.js
```diff
@@ -372,9 +370,27 @@
       });
     }
   
-    CodeMirror.commands.find = function(cm) {clearSearch(cm); doSearch(cm);};
-    CodeMirror.commands.findNext = function(cm) {clearFocusedMatches(cm); doSearch(cm, false, {next: true});};
-    CodeMirror.commands.findPrev = function(cm) {clearFocusedMatches(cm); doSearch(cm, true, {prev: true});};
+    CodeMirror.commands.find = function(cm) { 
+      cm.operation(function () {
```
montehurd: I tried something similar to this but it threw the match count off. Seems to work with your version 👍

montehurd:
I re-purposed my debouncing PR to only include a couple small cleanups I had made while looking at the search performance issue: https://github.com/wikimedia/wikipedia-ios/pull/3036

montehurd:
https://github.com/wikimedia/wikipedia-ios/pull/3046

PR #3049, opened 2019-04-22
add ability to close app and reopen to same place

nambatee:
https://phabricator.wikimedia.org/T219658


Wikipedia/Code/WMFAppViewController.m
```diff
@@ -901,7 +901,9 @@ - (void)finishResumingApp {
                                                    completion:^{
                                                    }];
     } else if (locationAuthorized != [defaults wmf_locationAuthorized]) {
-        [self.dataStore.feedContentController updateNearbyForce:NO completion:NULL];
+        [self.dataStore.feedContentController updateContentSource:[WMFNearbyContentSource class] force:NO completion:NULL];
+    } else if (!NSUserDefaults.wmf.shouldShowLastReadArticleOnResume) {
```
joewalsh: Seems like this should be a separate `if` instead of an `else if` since it should be able to happen whether or not nearby was updated due to a location authorization change

Wikipedia/Code/WMFAppViewController.m
```diff
@@ -901,7 +901,9 @@ - (void)finishResumingApp {
                                                    completion:^{
                                                    }];
     } else if (locationAuthorized != [defaults wmf_locationAuthorized]) {
```
joewalsh: To preserve the behavior where location and continue reading aren't updated if there's a full update, could be split here as

```
} else {
  if (location) {
  }
  
  if (shouldShowOnResume) {
  }
}
```

tonisevener:
Looks good, here's a few things I noticed but they could be unrelated or existing:

I notice is when the app opens up again (after terminating on an article) to the article and section, the table of contents has the top section highlighted, regardless of which section it's scrolled to upon launch. Once you start scrolling the highlight updates to the right section.

I also see a pretty significant delay in loading the explore feed on a fresh install. It seems to kick in and load eventually once I do other things (like search for an article, enable location in Places, etc.). Not sure if that's just me, I'm testing on my iPad.

When I go to an article, change the device date up a day, then relaunch, the continued reading card is under "Yesterday" further down rather than at the top per the ticket.

nambatee:
I ended up checking if `WMFArticleViewController`'s url has a fragment and resetting TOC in `viewDidScroll` if if it does and TOC is loaded. I'm resetting that flag as soon as the reset happen so it shouldn't affect the way TOC behaves otherwise

nambatee:
asked about the sticky card https://phabricator.wikimedia.org/T219658#5135795

Wikipedia/Code/WMFArticleViewController.h
```diff
@@ -43,6 +43,11 @@ extern NSString *const WMFEditPublishedNotification;
                          dataStore:(MWKDataStore *)dataStore
                              theme:(WMFTheme *)theme;
 
+//- (instancetype)initWithArticleURL:(NSURL *)url
+//                         dataStore:(MWKDataStore *)dataStore
+//             restoreScrollPosition:(BOOL)restoreScrollPosition
+//                             theme:(WMFTheme *)theme;
```
joewalsh: Could this be removed?

Wikipedia/Code/AppDelegate.m
```diff
@@ -99,6 +99,7 @@ - (void)applicationWillEnterForeground:(UIApplication *)application {
 - (void)applicationDidBecomeActive:(UIApplication *)application {
     // Restart any tasks that were paused (or not yet started) while the application was inactive. If the application was previously in the background, optionally refresh the user interface.
     [[NSUserDefaults wmf] wmf_setAppBecomeActiveDate:[NSDate date]];
+    NSUserDefaults.wmf.shouldShowLastReadArticleOnResume = [self shouldShowLastReadArticleOnResume];
```
joewalsh: The app should only need to show last article on the resume on a full quit & relaunch - `didBecomeActive` is called on every re-launch of the app even if it wasn't fully terminated. This should probably move to `didFinishLaunching`

Wikipedia/Code/WMFAppViewController.m
```diff
@@ -1486,6 +1491,10 @@ - (void)showExplore {
 #pragma mark - Last Read Article
 
 - (BOOL)shouldShowLastReadArticleOnLaunch {
+    if (!NSUserDefaults.wmf.shouldShowLastReadArticleOnResume) {
+        return NO;
+    }
+
     NSURL *lastRead = [[NSUserDefaults wmf] wmf_openArticleURL];
```
joewalsh: Out of scope for this PR but since this is user data it probably shouldn't be in user defaults. `WMFArticle` could have an `isOpen` or `navigationStackIndex` property. The other added benefit there would be the ability to restore the full navigation stack rather than just the top article.

Wikipedia/Code/WMFContinueReadingContentSource.m
```diff
@@ -35,11 +35,13 @@ - (void)stopUpdating {
 
 - (void)loadNewContentInManagedObjectContext:(NSManagedObjectContext *)moc force:(BOOL)force completion:(nullable dispatch_block_t)completion {
     NSURL *lastRead = [[NSUserDefaults wmf] wmf_openArticleURL] ?: self.userDataStore.historyList.mostRecentEntry.URL;
+    if (!lastRead) {
+        return;
```
joewalsh: Missing a call to the completion here - causing the hang on launch @tonisevener described

Wikipedia/Code/WMFAppViewController.m
nambatee: Makes sense, it was there before so I didn't touch it

joewalsh: Filed separately https://phabricator.wikimedia.org/T221859

Wikipedia/Code/AppDelegate.m
```diff
@@ -115,6 +116,17 @@ - (void)resumeAppIfNecessary {
     }
 }
 
+- (BOOL)shouldShowLastReadArticleOnResume {
+    NSDate *resignActiveDate = [[NSUserDefaults wmf] wmf_appResignActiveDate];
+    if (!resignActiveDate) {
+        return NO;
+    }
+    NSDate *becomeActiveDate = [[NSUserDefaults wmf] wmf_appBecomeActiveDate];
+    NSDate *cutoffDate = [[NSCalendar wmf_utcGregorianCalendar] nextDateAfterDate:resignActiveDate matchingHour:5 minute:0 second:0 options:NSCalendarMatchStrictly];
```
montehurd: At some point we may want to round up the various constant values throughout the app (like this `hour` setting here) and consolidate them into a `constants` plist or some other structure... maybe a nested Swift enum?

PR #3050, opened 2019-04-22
login page gives "invalid password" message when pw is correct but username is wrong

nambatee:
https://phabricator.wikimedia.org/T219559

hold til https://phabricator.wikimedia.org/T219559#5128702 is resolved


nambatee:
ready for review

PR #3051, opened 2019-04-22
As an editor I would like to be able to have a fuller preview of my links from the preview view

nambatee:
https://phabricator.wikimedia.org/T213753

- [x] constraints


nambatee:
ready for review

PR #3055, opened 2019-04-24
Fix section editor focus on entry

joewalsh:
https://phabricator.wikimedia.org/T221605


nambatee:
Works for me 👍 . Occasionally when using highlight to edit, it appears to be selected twice

ezgif-2-43fbcfe4192e

PR #3056, opened 2019-04-24
Address code review comments for media insert PR

nambatee:
https://github.com/wikimedia/wikipedia-ios/pull/3048


Wikipedia/Code/MediaWizardController.swift
```diff
@@ -97,12 +97,12 @@ final class MediaWizardController: NSObject {
             didCancelSearch()
             return
         }
-        if !isFirstSearch {
+        if isFirstSearch {
             progressController.delay = 0
```
nambatee: I was meant to set the delay to 0 on the first search cause it can be slow

Wikipedia/Localizations/en.lproj/Localizable.strings
```diff
@@ -393,6 +393,7 @@
 "insert-media-image-preview-select-image-action-title" = "Select image";
 "insert-media-image-size-setting-custom" = "Custom";
 "insert-media-image-size-setting-default" = "Default";
+"insert-media-image-size-settings-px-unit-name" = "px";
```
montehurd: In css is 'px' ever not 'px' in non-EN usage?

nambatee: This is a display name here -

Simulator Screen Shot - iPhone Xʀ - 2019-04-24 at 17 18 59

montehurd: ooh never mind i think i see how it's being used

montehurd: hehe dueling comments 😂

Wikipedia/Code/InsertMediaSettingsTableViewController.swift
```diff
@@ -96,6 +96,8 @@ final class InsertMediaSettingsTableViewController: UITableViewController {
                     }
                 }
 
+                static var unitName = WMFLocalizedString("insert-media-image-size-settings-px-unit-name", value: "px", comment: "Image size unit name, abbreviation for 'pixels'")
```
joewalsh: Should be a format string with a plural in case any languages need to adjust pluralization or would have the word in front of the number

nambatee: "px" is a separate label and the number is in its own text field so that the user can see that whatever they input, it's always pixels - I think the only way to make it work as a format string would be having all of the text ("220 px") as text field's text? But then the user could think that they can/should input other units like "em" etc? I could take numbers only when I create the wikitext but then it could look like a bug when they see "px" in the wikitext but they input something else? 😬 not sure how to best handle this case
Screen Shot 2019-04-24 at 5 35 26 PM

joewalsh: Makes sense, there's not an easy way to have it as a text field, it'd have to be a separate screen for entry or maybe the unit moves to the Width & Height labels to be something like Width (pixels). Let's leave it as is for now (there are examples of `px` alone as a string on translatewiki) and discuss at task sync.

PR #3062, opened 2019-04-26
Media insert: scroll away the "Selected image" view when user scrolls down the results list

nambatee:
https://phabricator.wikimedia.org/T221800


Wikipedia/Code/SectionEditorViewController.swift
```diff
@@ -576,24 +576,24 @@ extension SectionEditorViewController: SectionEditorWebViewMessagingControllerSc
...
-extension SectionEditorViewController: MediaWizardControllerDelegate {
-    func mediaWizardController(_ mediaWizardController: MediaWizardController, didPrepareViewController viewController: UIViewController) {
-        present(viewController, animated: true)
-    }
-
-    func mediaWizardController(_ mediaWizardController: MediaWizardController, didTapCloseButton button: UIBarButtonItem) {
-        self.mediaWizardController = nil
+extension SectionEditorViewController: InsertMediaViewControllerDelegate {
+    func insertMediaViewController(_ insertMediaViewController: InsertMediaViewController, didTapCloseButton button: UIBarButtonItem) {
+        self.insertMediaViewController = nil
```
tonisevener: You could get rid of this property now that you're using a view controller, the nav stack should keep it from deallocating.

nambatee: thanks, good catch

tonisevener:
I notice I'm not able to go back after selecting an image and tapping Next, the navigation bar hides. I'm testing on an iPhone 7.

nambatee:
@tonisevener I couldn't reproduce it but pushed a change that might fix it, let me know if you still see it

tonisevener:
@nambatee yep that worked 👍

PR #3064, opened 2019-04-26
(Continue reading) Fix accessing viewContext from a background thread

nambatee:
`historyList.mostRecentEntry` uses `viewContext`, `loadNewContentInManagedObjectContext` is called from a background thread


Wikipedia/Code/MWKHistoryList.m
```diff
@@ -84,6 +84,12 @@ - (nullable WMFArticle *)mostRecentEntry {
     return [[self.dataStore.viewContext executeFetchRequest:request error:nil] firstObject];
 }
 
+- (nullable WMFArticle *)mostRecentEntryInManagedObjectContext:(NSManagedObjectContext *)context {
```
joewalsh: The method above could now call this method with `self.dataStore.viewContext` but not a required change

PR #3065, opened 2019-04-26
Fix views affected by Swift 5 compiler bug

nambatee:
https://phabricator.wikimedia.org/T221906
https://phabricator.wikimedia.org/T221683
https://phabricator.wikimedia.org/T220694

~~I couldn't reproduce this consistently so this might not be the ultimate fix. It looked like the row height was off - all the rows were 44. It would make sense cause `heightForRowAt` can be called before the cells are available, causing custom cells to have the default height of 44. That was preventing me from changing the theme.~~

~~Let me know if you've seen it behave differently (as in if you were unable to change the theme while the table view looked ok).~~

### DIFF
--- a/Wikipedia/Code/AppearanceSettingsViewController.swift
+++ b/Wikipedia/Code/AppearanceSettingsViewController.swift
@@ -198,7 +198,7 @@ final class AppearanceSettingsViewController: SubSettingsViewController {
         } else if let spacerViewItem = sections[indexPath.section].items[indexPath.item] as? AppearanceSettingsSpacerViewItem {
             return spacerViewItem.spacing
         }
-        return UITableView.automaticDimension
+        return tableView.rowHeight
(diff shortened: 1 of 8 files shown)


joewalsh:
Table view looks fine for me- it doesn't repro on the debug build in the simulator but does repro on the release build

joewalsh:
Ready for review - something changed in Swift 5. On release builds, subclasses of classes that conform to an obj-c protocol and are adding a method that should conform to that protocol but not explicitly overriding a superclass implementation are compiled in such a way that they aren't called. Adding `@objc` fixed the issue. Might not be a complete understanding of what's going on, so happy to hear other info if anyone has it

nambatee:
https://bugs.swift.org/browse/SR-10257

nambatee:
lgtm

PR #3066, opened 2019-04-29
Initial talk page work

tonisevener:


tonisevener:
Not much to see on the UI side of things but please let me know if I'm off track with any of this so far. I will be switching to integrating the talk page endpoint & list screen UI now.

WMF Framework/Wikipedia.xcdatamodeld/Wikipedia 2.xcdatamodel/contents
```diff
@@ -0,0 +1,209 @@
...
+        <relationship name="discussions" toMany="YES" deletionRule="Cascade" ordered="YES" destinationEntity="TalkPageDiscussion" inverseName="talkPage" inverseEntity="TalkPageDiscussion" syncable="YES"/>
+    </entity>
+    <entity name="TalkPageDiscussion" representedClassName="TalkPageDiscussion" syncable="YES" codeGenerationType="class">
+        <attribute name="title" attributeType="String" syncable="YES"/>
+        <relationship name="items" toMany="YES" deletionRule="Cascade" ordered="YES" destinationEntity="TalkPageDiscussionItem" inverseName="discussion" inverseEntity="TalkPageDiscussionItem" syncable="YES"/>
+        <relationship name="talkPage" maxCount="1" deletionRule="Deny" destinationEntity="TalkPage" inverseName="discussions" inverseEntity="TalkPage" syncable="YES"/>
+    </entity>
+    <entity name="TalkPageDiscussionItem" representedClassName="TalkPageDiscussionItem" syncable="YES" codeGenerationType="class">
+        <attribute name="depth" attributeType="Integer 16" defaultValueString="0" usesScalarValueType="YES" syncable="YES"/>
+        <attribute name="text" attributeType="String" syncable="YES"/>
```
nambatee: You could also keep the text in a file and store a reference as an attribute on `TalkPageDiscussionItem`, in some cases there could be a lot of text

Wikipedia/Code/TalkPageLocalHandler.swift
```diff
@@ -0,0 +1,97 @@
...
+    func updateExistingTalkPage(existingTalkPage: TalkPage, with networkTalkPage: NetworkTalkPage) -> TalkPage? {
+        existingTalkPage.name = networkTalkPage.name
+        existingTalkPage.revisionId = networkTalkPage.revisionId
+        
+        guard let newDiscussions: [TalkPageDiscussion] = createTalkPageDiscussions(with: networkTalkPage) else {
+            return nil
+        }
+        
+        //first delete old discussions
+        //todo: is there a better way to do this? simply reassigning discussions is not enough
```
nambatee: If I'm understanding this correctly, you could compare your existing discussions with the discussions that are coming from the server and figure out what changed (what was deleted, what was added, what was updated) and then process those specific changes - we do something like that for reading lists -https://github.com/wikimedia/wikipedia-ios/blob/908bd09ae9e8c0907488512b149d5ce049c5a48d/WMF%20Framework/ReadingListsSyncOperation.swift#L244

specifically here's where the updates happen - https://github.com/wikimedia/wikipedia-ios/blob/908bd09ae9e8c0907488512b149d5ce049c5a48d/WMF%20Framework/ReadingListsSyncOperation.swift#L290

WMF Framework/Wikipedia.xcdatamodeld/Wikipedia 2.xcdatamodel/contents
tonisevener: Good to know, I was unsure if it would be overkill so I didn't want to tackle it initially - I'll look into that after I flesh out the UI more.

Wikipedia/Code/TalkPageLocalHandler.swift
tonisevener: Oh yeah that would be better. Right now I'm just throwing out all discussions and recreating them if the revision ID has changed. The question was more around delete rules - the old discussions still hang around in the DB even after I reassign the talk page to a new set of discussions.  This was a note to look into a way to have that delete cascade on that assignment without explicitly looping through and deleting them all. But agreed being more selective about what I delete would be a better way to go.

Wikipedia/Code/ReplyListItemCollectionViewCell.swift
```diff
@@ -0,0 +1,115 @@
...
+    
+    func configure(title: String, depth: UInt) {
+        
+        self.depth = depth
+        depthMarker.isHidden = depth < 1
+        
+        let font = UIFont.wmf_font(.body, compatibleWithTraitCollection: traitCollection)
+        let boldFont = UIFont.wmf_font(.semiboldBody, compatibleWithTraitCollection: traitCollection)
+        
+        //todo: we seem to be losing <b> tags (possibly more). Need to look into this.
```
montehurd: Is this still happening? As a starting point for adding list item parsing I'm using...
> .byAttributingHTML(with: .caption2, matching: traitCollection)

... which seems to play nice with bold. 

Screen Shot 2019-05-13 at 11 39 29 AM

The params it passes to `wmf_attributedStringFromHTML` may help?

WMF Framework/Wikipedia.xcdatamodeld/Wikipedia 2.xcdatamodel/contents
tonisevener: FYI planning to do this separately as a part of https://phabricator.wikimedia.org/T222721.

tonisevener:
Not feature complete but I think this is far enough along to review for merging into develop.

Wikipedia/Code/ReplyListItemCollectionViewCell.swift
tonisevener: Should be fixed now, I was accidentally overwriting the attributed string.

PR #3068, opened 2019-05-01
Design review updates: Link previews

nambatee:
https://phabricator.wikimedia.org/T213753#5150195

Sizing question:

The VC that I'm presenting is in a xib (`EditPreviewInternalLinkViewController.xib`). When I lay out my subviews, I rely on the VC's view's bounds to figure out how much space I have available:

https://github.com/wikimedia/wikipedia-ios/blob/908bd09ae9e8c0907488512b149d5ce049c5a48d/Wikipedia/Code/ArticlePeekPreviewViewController.swift#L47

But at that time, the bounds are wrong so my `ArticleFullWidthImageCollectionViewCell` ends up thinking that it has more available width to lay out its labels than it actually does. I thought I could set `preferredContentSize` on the VC before presenting it and that would let it know what size it should shoot for but that doesn't affect the bounds. 😬  Is there a good way to let the VC how big its bounds should be before it's presented? 

To demonstrate the issue - here, I think, the title label thought that it had more width available so it told the cell that its height would be only ~30 when in fact it's more than that

Simulator Screen Shot - iPhone X - 2019-05-01 at 16 45 08

### DIFF
--- a/Wikipedia/Code/ArticlePeekPreviewViewController.swift
+++ b/Wikipedia/Code/ArticlePeekPreviewViewController.swift
@@ -33,6 +33,12 @@ class ArticlePeekPreviewViewController: UIViewController, Peekable {
         updateView(with: article)
     }
     
+    public func updatePreferredContentSize(for contentWidth: CGFloat) {
+        var updatedContentSize = expandedArticleView.sizeThatFits(CGSize(width: contentWidth, height: UIView.noIntrinsicMetric), apply: true)
+        updatedContentSize.width = contentWidth // extra protection to ensure this stays == width
+        preferredContentSize = updatedContentSize
+    }
+    
     fileprivate func updateView(with article: WMFArticle) {
         expandedArticleView.configure(article: article, displayType: .pageWithPreview, index: 0, theme: theme, layoutOnly: false)
         expandedArticleView.isSaveButtonHidden = true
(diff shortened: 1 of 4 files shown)


joewalsh:
> But at that time, the bounds are wrong so my ArticleFullWidthImageCollectionViewCell ends up thinking that it has more available width to lay out its labels than it actually does. I thought I could set preferredContentSize on the VC before presenting it and that would let it know what size it should shoot for but that doesn't affect the bounds. 😬 Is there a good way to let the VC how big its bounds should be before it's presented?

I'm not sure the `preferredContentSize` should continue to depend on the `bounds.width` (because setting it might affect the `bounds.width`) but given that it's been like this for a while and the `preferredContentSize.width` should always equal the `bounds.width`, it's probably OK to leave it. 

With `preferredContentSize` continuing to depend on `bounds.width`, `preferredContentSize` needs to be updated when the `bounds` updates. It would make sense to update it in one of the methods that signals the `bounds` have changed like  `viewWillTransition(to size` or `viewDidLayoutSubviews`. It looks like adding the update to `viewDidLayoutSubviews` fixes the issue so I made that commit - feel free to revert if that's not fixing what you were seeing.

joewalsh:
before: 
Simulator Screen Shot - iPhone 5s - 2019-05-02 at 08 19 07
after:
Simulator Screen Shot - iPhone 5s - 2019-05-02 at 08 17 45

PR #3069, opened 2019-05-01
Design review updates: Insert media

nambatee:
- [x] Make image info overlay tappable
- [x] Update 'More information' image
- [x] Make the text area placeholder text flush with the rest of the text on the screen
- [x] Update font size for 'Uploaded image', 'Caption' and 'Alternative text'
- [x] Make the image title in Settings a button, push to Commons
- [x] Increase the height of Advanced settings cells
~- [ ] Fix cursor disappearing / web view losing focus after dismissing the wizard or inserting an image~ fixed in 12.2
- [x]  Enable toolbar icon as long as the cursor is not placed within other syntax (eg. formatting, link, template) and break to a new line if it's inserted somewhere in the middle of a line


Wikipedia/assets/codemirror/codemirror-index.html
```diff
@@ -358,8 +358,10 @@
         buttonsToDisable.push(buttonPayload('clearFormatting'))
       }
 
-      const tokenType = doc.getTokenTypeAt(doc.getCursor())
-      if ((tokenType && tokenType.length > 0) || selectedButtons.length > 0) {
+      const cursorPosition = doc.getCursor()
+      const token = doc.getTokenAt(cursorPosition)
+      const isAtLineEnd = doc.getLine(cursorPosition.line).length === cursorPosition.ch
+      if (((token.type && token.type.length > 0) || selectedButtons.length > 0) && !isAtLineEnd) {
```
nambatee: Let me know if there's a better way to check if a cursor is placed inside wikitext. In `selectedButtons`, we store only the buttons that we care about in the UI so checking that array is not enough. When the cursor is at the end of the line, `getTokenAt` will return the adjacent token so checking token alone is not enough - that's why there's an additional line check.

nambatee:
ready for review

Wikipedia/Code/Advanced Settings/InsertMediaAdvancedSettingsViewController.swift
```diff
@@ -142,6 +141,10 @@ extension InsertMediaAdvancedSettingsViewController: UITableViewDataSource {
         cell.textLabel?.textColor = cell.isUserInteractionEnabled ? theme.colors.primaryText : theme.colors.secondaryText
         cell.detailTextLabel?.textColor = theme.colors.secondaryText
     }
+
+    func tableView(_ tableView: UITableView, heightForRowAt indexPath: IndexPath) -> CGFloat {
+        return 50
```
joewalsh: Simulator Screen Shot - iPhone X - 2019-05-07 at 13 47 26
^ breaks for larger dynamic type font sizes

Could try adjusting the cell margins? Otherwise I'd push back as I'm not sure it's worth a custom cell for the extra height.

nambatee: thanks for catching, I tried margins but they seem to have no effect

nambatee: not sure why but when setting `layoutMargins` or `directionalLayoutMargins` on `cell` in `cellForRowAt`, only `right` or `trailing` is applied, `top`, `left`/`leading`, `bottom` are ignored.

@carolynlimadeo would you be okay keeping the cell height as is? Right now, we're using a built-in cell, to increase the height we'd have to create a custom component

tonisevener:
One thing I notice - if I select one image, then a different one and tap "Next" fast enough, the old image name shows up on the Media settings screen and gets inserted. We might need to disable Next again while the image is fetching if they select another one.

Wikipedia/assets/codemirror/codemirror-index.html
joewalsh: It looks like the range determination bundle could be used for this. I modified it a bit to detect ranges of template argument names (prior to that it would allow media insertion in the middle of a template argument name) and it seems to behave mostly as expected. One case that doesn't work is a line with an unterminated template on it like `{{Infobox coin` in today's featured article. Feel free to revert or update if there's something I'm missing.

montehurd:
@joewalsh oops left my comment before your last commit... re-checking...

montehurd:
(I updated my comment above... could also see the `outer` for template arg maybe just encompassing the param name...)

montehurd:
Looks like there may also be more than one `MarkupItem` for a given template arg... Vs one with inner and outer range set accordingly.

PR #3071, opened 2019-05-03
Add GlobalPreferencesController and GlobalPreferencesFetcher

joewalsh:
data model for https://phabricator.wikimedia.org/T220549

### DIFF
--- a/WMF Framework/Fetcher.swift
+++ b/WMF Framework/Fetcher.swift
@@ -198,8 +198,3 @@ public class Token: NSObject {
         self.isAuthorized = value != "+\\"
     }
 }
-
-public enum Result<Success, Error> {
-    case success(Success)
-    case failure(Error)
-}
--- a/WMF Framework/Fetcher.swift
+++ b/WMF Framework/Fetcher.swift
@@ -5,7 +5,6 @@ import UIKit
 
 // TODO: Standardize on returning CancellationKey or URLSessionTask
 // TODO: Centralize cancellation and remove other cancellation implementations (ReadingListsAPI)
-// TODO: Remove CSRFTokenOperation and create other new methods here for token generation and use
 // TODO: Think about utilizing a request buildler instead of so many separate functions
 // TODO: Utilize Result type where possible (Swift only)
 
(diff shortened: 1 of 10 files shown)


WMF Framework/Fetcher.swift
```diff
@@ -200,7 +199,7 @@ public class Token: NSObject {
     }
 }
 
-public enum FetcherResult<Success, Error> {
+public enum Result<Success, Error> {
```
joewalsh: Should we move this to a separate file and use it everywhere we want a result type?

montehurd: That sounds cool! 😎

tonisevener: Can we just use the Swift 5 Result type?

joewalsh: good point - forgot that the Result type landed with Swift 5. Will remove this one.

joewalsh:
setting WIP to work out global/local preference questions

Wikipedia/Code/WMFAppViewController.m
```diff
@@ -252,18 +252,21 @@ - (NSURL *)siteURL {
...
     [self.periodicWorkerController add:[WMFEventLoggingService sharedInstance]];
+    [self.periodicWorkerController add:self.dataStore.globalPreferencesController];
 
     self.backgroundFetcherController = [[WMFBackgroundFetcherController alloc] init];
     self.backgroundFetcherController.delegate = self;
     [self.backgroundFetcherController add:self.dataStore.readingListsController];
     [self.backgroundFetcherController add:self.dataStore.remoteNotificationsController];
     [self.backgroundFetcherController add:(id<WMFBackgroundFetcher>)self.dataStore.feedContentController];
     [self.backgroundFetcherController add:[WMFEventLoggingService sharedInstance]];
+    [self.periodicWorkerController add:self.dataStore.globalPreferencesController];
```
tonisevener: should be backgroundFetcherController :)

joewalsh: thanks! fixed

PR #3072, opened 2019-05-06
Fix missing reading list entry titles

joewalsh:
https://phabricator.wikimedia.org/T222486
https://phabricator.wikimedia.org/T222487

I broke this here https://github.com/wikimedia/wikipedia-ios/pull/3011 

Since it's only broken 6.3 which hasn't gone to beta, I don't think we need a migration to fix existing DBs. I added extra protection against this mistake in the future by "deprecating" `displayTitleHTMLString` which will cause a build warning if it's used elsewhere.


WMF Framework/WMFArticle+Extensions.swift
```diff
@@ -1,16 +1,6 @@
 import Foundation
 
 extension WMFArticle {
-    @objc public var displayTitleHTML: String {
```
joewalsh: Had to move this to obj-c to use `clang diagnostic ignored "-Wdeprecated-declarations"` - looks like there's no equivalent in Swift yet

PR #3076, opened 2019-05-14
Insert link: As an editor on the iOS app I would like to have an easy way to insert links into an article

nambatee:
https://phabricator.wikimedia.org/T213979


Wikipedia/assets/codemirror/codemirror-index.html
```diff
@@ -1073,7 +1258,62 @@
...
+            }
+          }
+        } else { // new link
+          if (label) {
+            const newPageWithLabel = `[[${page}|${label}]]`
+            editor.replaceRange(newPageWithLabel, link.pagePosition.from, link.pagePosition.to)
+          } else {
+            const newPage = `[[${page}]]`
+            editor.replaceRange(newPage, link.pagePosition.from, link.pagePosition.to)
+          }
```
joewalsh: This new link block doesn't seem to be working. `link.pagePosition` is undefined. Should these be `replaceSelection` instead of `replaceRange`?

joewalsh: Repro steps are select unlinked text (like "home"), tap the link button, tap an article

joewalsh:
For new links, it seems like the selected text should be preserved as the link label. There are two parts to this depending on whether or not the article title matches (case insensitive) the selection. If you select the text `home` (lowercased) and add the link for `Home` (capitalized) the link should be `[[home]]`. If I select the text `Home Improvement` and add the link for `Home Improvement (TV series)` the link should be `[[Home Improvement (TV series)|Home Improvement]]`. It might be fine to use the second case for both (so the first case would be `[[Home|home]]`) but we should confirm with @carolynlimadeo

Wikipedia/Code/SearchViewController.swift
```diff
@@ -65,7 +65,20 @@ class SearchViewController: ArticleCollectionViewController, UISearchBarDelegate
...
+    var shouldSetTitleViewWhenRecentSearchesAreDisabled: Bool = true
+
+    var shouldShowCancelButton: Bool = true
+    var delegatesSelection: Bool = false {
+        didSet {
+            resultsViewController.delegatesSelection = delegatesSelection
+        }
+    }
+
+    var showLanguageBar: Bool?
```
joewalsh: I ran into an issue with multiple languages where I was editing an article in a different language but I could only add links from my default language. It makes sense to have the language bar hidden but there should probably be a way to set the site to be searched. This way, `EditLinkViewController` and `InsertLinkViewController` could pass in the site of the article being edited so that the search is performed on that site.

Wikipedia/assets/codemirror/codemirror-index.html
nambatee: `replaceSelection` won't work for 0 length selection

Wikipedia/Code/SearchViewController.swift
nambatee: Good point, thanks!

nambatee: Something like https://github.com/wikimedia/wikipedia-ios/pull/3076/files/78d95ac0b070a1ce03afb5e1f08fa0e96adf8af1..4bf211294acc9dd4aa01b3d19a62df800e3b2ec0?

joewalsh: yup! looks good

carolynlimadeo:
@joewalsh good point, yes we'll want to keep the case that the user had set in the text, and it would be helpful to auto generate the pipe and label text when the article title is longer / different from the highlighted text in the editing view (re: home improvement...)

nambatee:
updated to include label with new links if the title is different from the highlighted text - https://github.com/wikimedia/wikipedia-ios/pull/3076/commits/108e4c239c1fd1079ef901ffb080f1803cc0b4d4

Wikipedia/Code/EditLinkViewController.swift
```diff
@@ -0,0 +1,203 @@
...
+        closeButton.tintColor = theme.colors.primaryText
+        doneButton.tintColor = theme.colors.link
+        displayTextView.textColor = theme.colors.primaryText
+        activityIndicatorView.style = theme.isDark ? .white : .gray
+    }
+}
+
+extension EditLinkViewController: ArticleCollectionViewControllerDelegate {
+    func articleCollectionViewController(_ articleCollectionViewController: ArticleCollectionViewController, didSelectArticleWith articleURL: URL, at indexPath: IndexPath) {
+        self.articleURL = articleURL
```
joewalsh: Feels like this should pop the search VC as well - it was a bit confusing to have to select "Back" to confirm the change

Wikipedia/Code/ViewController.swift
```diff
@@ -7,6 +7,10 @@ class ViewController: PreviewingViewController, Themeable, NavigationBarHiderDel
     init() {
         super.init(nibName: nil, bundle: nil)
     }
+
+    override init(nibName nibNameOrNil: String?, bundle nibBundleOrNil: Bundle?) {
```
tonisevener: You can delete this.

Wikipedia/Code/EditLinkViewController.swift
```diff
@@ -0,0 +1,204 @@
...
+    @IBOutlet private weak var scrollViewTopConstraint: NSLayoutConstraint!
+    @IBOutlet private weak var displayTextLabel: UILabel!
+    @IBOutlet private weak var displayTextView: UITextView!
+    @IBOutlet private weak var displayTextViewHeightConstraint: NSLayoutConstraint!
+    @IBOutlet private weak var linkTargetLabel: UILabel!
+    @IBOutlet private weak var linkTargetContainerView: UIView!
+    @IBOutlet private weak var linkTargetContainerViewHeightConstraint: NSLayoutConstraint!
+    @IBOutlet private weak var activityIndicatorView: UIActivityIndicatorView!
+    @IBOutlet private weak var removeLinkButton: AutoLayoutSafeMultiLineButton!
+    @IBOutlet private var separatorViews: [UIView] = []
```
tonisevener: Thought these had to be IBOutletCollections 🤔 seems to work fine without though.

Wikipedia/Code/ViewController.swift
nambatee: I'm using that init in `EditLinkViewController`, without it I'll get an `Ambiguous reference to member 'init()'` error

Wikipedia/Code/EditLinkViewController.swift
nambatee: Looks like `@IBOutletCollection` is not supported in Swift? 🤔 Trying to declare `@IBOutletCollection` produces and error and dragging from the xib, `Outlet Collection` shows up as `@IBOutlet`

nambatee:
@carolynlimadeo if we have `[[Autumn leaf color|autumn leaves]]` and the user taps the _Remove link_ button, should we end up with `autumn leaves` or `Autumn leaf color|autumn leaves`? Tapping on the _remove wikitext_ toolbar button would leave us with `Autumn leaf color|autumn leaves`

Wikipedia/Code/ViewController.swift
tonisevener: ah ok my mistake, just thought it was a missed unused override

Wikipedia/Code/EditLinkViewController.swift
tonisevener: 🤦‍♀️ ha no problem - a difference that I'm just now noticing 5 years later.

carolynlimadeo:
@nambatee if they remove the link can we leave them with the label text only?

nambatee:
* pushed an update that leaves the label when link is removed https://github.com/wikimedia/wikipedia-ios/pull/3076/commits/cb08c1a96bdb44ac1f03ac1e357848126eafe826
* pushed a change that fixes links not being recognized if the cursor is touching the outer edge of the markup - let me know if there's a better way to do this - https://github.com/wikimedia/wikipedia-ios/pull/3076/commits/07f2d0814122080f5917cc4e2bcbe5ea6eb1e6bc
Simulator Screen Shot - iPhone SE - 2019-05-15 at 13 02 15

montehurd:
Hmm unsure the cursor touching an outer edge of an element should behave as if it's inside the element. The reason is you can have two elements right next to each other and there's no way to know what the user intends if the cursor is between them (touching two edges simultaneously) ie:
> [[anchor one]]*CURSOR*[[anchor two]]

montehurd:
Then again...doing this edge behavior if exactly one edge is touched means there's no ambiguity... sorry if you're already checking this - haven't deep dived 👀 into the diff yet.

nambatee:
@montehurd I asked about that behavior (cursor touching the outer edge) here https://phabricator.wikimedia.org/T213979#5148656
and Carolyn confirmed that's what we wanted.

The case you're describing is being handled - right now, the link button is disabled.

montehurd:
Oh cool! Ya disabling the button if the cursor is on two outer edges at once sounds good to me :)

PR #3078, opened 2019-05-14
Remove shadow from the Toolbar on Reading themes (dark, black, sepia)

nambatee:
https://phabricator.wikimedia.org/T216551


joewalsh:
On the ticket, it looks like the navbar and toolbar colors changed as well - @carolynlimadeo was that intentional or no?

joewalsh:
Looks good, going to merge and any color changes can be done as a follow on

PR #3080, opened 2019-05-14
Remove shadow from the Toolbar input views on Reading themes (dark, black, sepia), update inputAccessoryBackground

nambatee:
- remove shadow from edit input views for non-default themes
- update `inputAccessoryBackground` color


Wikipedia/Code/TextFormattingInputViewController.swift
```diff
@@ -110,7 +103,17 @@ extension TextFormattingInputViewController: Themeable {
             return
         }
         view.backgroundColor = theme.colors.inputAccessoryBackground
-        view.layer.shadowColor = theme.colors.shadow.cgColor
+        if theme == Theme.standard {
```
joewalsh: I missed this in the last review, but instead of checking equality to a specific theme type, could this check a new property on the theme like `theme.hasInputAccessoryShadow`? In general the `theme` instance should be updated to contain any information needed for visual customization so that individual views don't have knowledge of specific theme instances, just properties of theme objects

nambatee: Sure, that's a good idea

PR #3081, opened 2019-05-14
Remove GlobalPreferencesController & Fetcher

joewalsh:
Follow up from decision on https://phabricator.wikimedia.org/T220549


WMF Framework/Configuration.swift
```diff
@@ -92,7 +92,7 @@ public class Configuration: NSObject {
         }
     }
     
-    func mediaWikiAPIURLComponentsBuilder(for host: String? = nil) -> APIURLComponentsBuilder {
+    func mediaWikiAPIURLComponentsBuilderForHost(_ host: String? = nil) -> APIURLComponentsBuilder {
```
joewalsh: Reverted this naming change to match the rest of the file

PR #3088, opened 2019-05-21
Replace direct calls to 'layoutSubviews'

montehurd:
Per docs it should not be called directly.


montehurd:
It looks like things are behaving correctly after this change, but please do test 😁

tonisevener:
I found 1 layout-y bug but I can repro in develop so it's unrelated - I'll make a ticket for it.

PR #3090, opened 2019-05-22
Select to edit performance fix

montehurd:
https://phabricator.wikimedia.org/T221608

Fixes variety of small select-to-edit bugs and speeds it up.

### Before:
select - before mov

### After:
select after mov


Wikipedia/assets/codemirror/codemirror-editTextSelection.js
```diff
@@ -6,92 +6,59 @@ class SelectedAndAdjacentText {
...
+        score: thisMatchScore,
+        matchedWikitextBeforeSelection: {start: 0, end: match.index},
+        matchedWikitextSelection: {start: match.index, end: selectedTextRegex.lastIndex}
+      }    
+    }
   }
-  const match = wikitext.match(regex)
-  if (match === null) {
+
+  if (bestScoredMatch === undefined) {
```
joewalsh: `null === undefined` is `false` so this fails when `bestScoredMatch` remains `null`

Wikipedia/assets/codemirror/codemirror-editTextSelection.js
```diff
@@ -6,92 +6,59 @@ class SelectedAndAdjacentText {
...
+  return wordsBeforeScore + wordsAfterScore
 }
 
 const wikitextRangeForSelectedAndAdjacentText = (selectedAndAdjacentText, wikitext) => {
-  const regex = selectedAndAdjacentText.regexForLocatingSelectedTextInWikitext(wikitext)
-  if (regex === null) {
-    return null
+  const wordsBeforeFromHTML = getNumberOfWordsFromEndOfString(selectedAndAdjacentText.textBeforeSelectedText, adjacentComparisonWordCount)
+  const wordsAfterFromHTML = getNumberOfWordsFromBeginningOfString(selectedAndAdjacentText.textAfterSelectedText, adjacentComparisonWordCount)
+  const selectedTextRegexPattern = `(${selectedAndAdjacentText.selectedText.replace(/\s+/g, '\\W*(?:(?:[<{][^>]*[}>])*)\\W*')})`
```
joewalsh: Could the in-between-words regex be updated to be more permissive of text within links? I had some luck with `((\\[\\[[^\\]\\|]+\\|)|{{[^}]*}}|<[^>]*>|\\W)+` but unsure if that has any other consequences. The case I was looking at was selecting "is a grammatically correct sentence in American English" in [this article](https://en.wikipedia.org/wiki/Buffalo_buffalo_Buffalo_buffalo_buffalo_buffalo_Buffalo_buffalo)

montehurd: ah good catch! will give it a spin!

Wikipedia/assets/codemirror/codemirror-editTextSelection.js
montehurd: doh! 😜

montehurd: fixed!

Wikipedia/assets/codemirror/codemirror-editTextSelection.js
montehurd: works great! fixed!

PR #3096, opened 2019-05-23
Bouncing down on reading list details screen is jittery

nambatee:
https://phabricator.wikimedia.org/T224127

### DIFF
--- a/WMF Framework/CollectionViewEditController.swift
+++ b/WMF Framework/CollectionViewEditController.swift
@@ -576,7 +576,7 @@ public class CollectionViewEditController: NSObject, UIGestureRecognizerDelegate
     
     public var isTextEditing: Bool = false {
         didSet {
-            editingState = .editing
+            editingState = isTextEditing ? .editing : .done
         }
     }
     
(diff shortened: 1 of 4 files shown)


nambatee:
original jitter goes away but there's still some if you try to scroll up while any of the fields are in focus, still looking at a fix

joewalsh:
When using it, it seemed like text editing should end when scrolling or tapping through on an article, which would also solve the remaining scroll jitter. I've made those changes but feel free to revert if that's not the intended behavior. CC @carolynlimadeo - this is on the reading list detail view where you can edit the reading list's title and description. Is it OK to have any editing commit when the user scrolls the list of articles or taps through on an article?

carolynlimadeo:
@joewalsh having the editing view close on scroll might be a problem if there's a large number of articles but if this would only happen when there is 1 page or fewer I think it should be okay. Assuming you're saying that scroll or navigating away is equivalent to tapping cancel?

joewalsh:
@carolynlimadeo I have it wired up as equivalent to tapping done but it could be equivalent to tapping cancel if you think that'd be better. It would happen no matter how many articles are in the list - similar to the behavior on text fields where any user scroll dismisses the keyboard

carolynlimadeo:
@joewalsh sorry I think I'm confused. If I have 40 articles in a reading list, then tap on edit and scroll, I am still able to scroll to the bottom of the list of articles without executing the edit, correct?

If this is the case, I'm not sure I understand what scrolling action would trigger the bulk edit view to close. My apologies for not understanding.

joewalsh:
Following up after screen share with @carolynlimadeo  - confusion was around title/description editing vs batch article editing. OK to move forward with committing the title or description edit (equivalent to tapping done) on user scroll or tap on an article.

nambatee:
lgtm

PR #3099, opened 2019-05-24
make "Please be kind" view an inputAccessoryView

nambatee:
ezgif-3-901b7e66e0fa


Wikipedia/Code/TalkPageTopicNewViewController.swift
```diff
@@ -279,7 +223,7 @@ private extension TalkPageTopicNewViewController {
         bodyContainerVerticalPaddingConstraints.forEach { contentFittingBodyContainerHeight += $0.constant  }
         
         var availableVerticalScreenSpace = talkPageScrollView.frame.height - bodyContainerOrigin.y
-        availableVerticalScreenSpace = availableVerticalScreenSpace - finePrintContainerView.frame.height - beKindContainerView.frame.height
+        availableVerticalScreenSpace = availableVerticalScreenSpace - finePrintContainerView.frame.height - beKindInputAccessoryView!.frame.height
```
joewalsh: Might be helpful to make `beKindInputAccessoryView` a lazy, non optional `var` to avoid having to force unwrap here

joewalsh: Saw it was changes welcome so I made this change, feel free to revert

PR #3107, opened 2019-05-29
add isRead property to track read status of talk page topics

joewalsh:
https://phabricator.wikimedia.org/T222726

### DIFF
--- a/Wikipedia/Code/TalkPageTopicListViewController.swift
+++ b/Wikipedia/Code/TalkPageTopicListViewController.swift
@@ -241,11 +241,12 @@ private extension TalkPageTopicListViewController {
     }
     
     func configure(cell: TalkPageTopicCell, at indexPath: IndexPath) {
-        guard let title = fetchedResultsController.object(at: indexPath).title else {
+        let topic = fetchedResultsController.object(at: indexPath)
+        guard let title = topic.title else {
             return
         }
         
-        cell.configure(title: title)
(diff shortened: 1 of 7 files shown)


Wikipedia/Code/TalkPageTopicListViewController.swift
```diff
@@ -241,11 +241,12 @@ private extension TalkPageTopicListViewController {
...
     
     func configure(cell: TalkPageTopicCell, at indexPath: IndexPath) {
-        guard let title = fetchedResultsController.object(at: indexPath).title else {
+        let topic = fetchedResultsController.object(at: indexPath)
+        guard let title = topic.title else {
             return
         }
         
-        cell.configure(title: title)
+        cell.configure(title: title, isRead: topic.isRead)
```
nambatee: Another call to this configure method happens when the cell is updated -

https://github.com/wikimedia/wikipedia-ios/blob/unread/Wikipedia/Code/TalkPageReplyListViewController.swift#L263

should `isRead` bet set the same way there instead of defaulting to `false`?

joewalsh: That one is for `TalkPageReply`s which IIRC we aren't supposed to handle unread state for yet, only `TalkPageTopic`s

joewalsh:
@tonisevener your feedback sounds good - I want to utilize the changes from the other PR #3108 in the updates so I'm going to merge this one with that

PR #3108, opened 2019-05-29
Track read status of talk page topics, Core Data thread safety fixes

joewalsh:
https://phabricator.wikimedia.org/T222726

bc843fe ensures `NSManagedObject`s are only accessed within the `NSManagedObjectContext` perform block they were fetched in so they aren't accessed on any other queue and can't be deleted before being re-accessed on the same queue. Also removes some of the dependence on the `viewContext`, allowing for backgrounding of update operations in the future. Open to suggestions for different approaches.

### DIFF
--- a/Wikipedia/Code/TalkPagesController.swift
+++ b/Wikipedia/Code/TalkPagesController.swift
@@ -36,10 +36,10 @@ class TalkPageController {
         return type.displayTitle(for: title, titleIncludesPrefix: titleIncludesPrefix)
     }
     
-    required init(fetcher: TalkPageFetcher = TalkPageFetcher(), articleRevisionFetcher: WMFArticleRevisionFetcher = WMFArticleRevisionFetcher(), dataStore: MWKDataStore, title: String, host: String, languageCode: String, titleIncludesPrefix: Bool, type: TalkPageType) {
+    required init(fetcher: TalkPageFetcher = TalkPageFetcher(), articleRevisionFetcher: WMFArticleRevisionFetcher = WMFArticleRevisionFetcher(), moc: NSManagedObjectContext, title: String, host: String, languageCode: String, titleIncludesPrefix: Bool, type: TalkPageType) {
         self.fetcher = fetcher
         self.articleRevisionFetcher = articleRevisionFetcher
-        self.moc = dataStore.viewContext
+        self.moc = moc
         self.title = title
         self.host = host
         self.languageCode = languageCode
@@ -57,9 +57,7 @@ class TalkPageController {
             do {
                 guard let localTalkPage = try self.moc.talkPage(for: taskURL) else {
                     self.createTalkPage(with: urlTitle, taskURL: taskURL, in: self.moc, completion: { (result) in
-                        DispatchQueue.main.async {
-                            completion?(result)
-                        }
+                        completion?(result)
                     })
                     return
                 }
@@ -70,23 +68,17 @@ class TalkPageController {
                     case .success(let lastRevisionID):
                         //if latest revision ID is the same return local talk page. else forward revision ID onto talk page fetcher
                         if localRevisionID == lastRevisionID {
-                            DispatchQueue.main.async {
-                                completion?(.success(localObjectID))
-                            }
+                            completion?(.success(localObjectID))
                         } else {
                             self.fetchAndUpdateLocalTalkPage(with: localObjectID, revisionID: lastRevisionID, completion: completion)
                         }
                     case .failure(let error):
-                        DispatchQueue.main.async {
-                            completion?(.failure(error))
-                        }
+                        completion?(.failure(error))
                     }
                 }
                 
             } catch {
-                DispatchQueue.main.async {
-                    completion?(.failure(TalkPageError.fetchLocalTalkPageFailure))
-                }
+                completion?(.failure(TalkPageError.fetchLocalTalkPageFailure))
                 return
             }
         }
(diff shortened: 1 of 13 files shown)


Wikipedia/Code/TalkPagesController.swift
```diff
@@ -54,59 +48,49 @@ class TalkPageController {
...
         guard let urlTitle = type.urlTitle(for: title, titleIncludesPrefix: titleIncludesPrefix),
             let taskURL = fetcher.taskURL(for: urlTitle, host: host) else {
             completion?(.failure(TalkPageError.createTaskURLFailure))
             return
         }
         
-        var localTalkPage: TalkPage?
-        localHandler.dataStore.viewContext.performAndWait {
+        let moc = dataStore.viewContext
+        moc.perform {
```
nambatee: I thought you could skip the `perform` block when accessing the main context from the main queue - is that incorrect or is the assumption that this could be called from a non-main queue? 🤔

joewalsh: You're correct. The intent here was to reduce dependence on the moc being the `viewContext` but this perform could be removed

Wikipedia/Code/TalkPagesController.swift
```diff
@@ -147,10 +131,9 @@ class TalkPageController {
         }
         
         taskGroup.waitInBackground {
-            
-            self.localHandler.dataStore.viewContext.perform {
+            moc.perform {
                 if talkPageDoesNotExist {
-                    if let newLocalTalkPage = self.localHandler.createEmptyTalkPage(with: taskURL, languageCode: self.languageCode, displayTitle: self.displayTitle) {
+                    if let newLocalTalkPage = moc.createEmptyTalkPage(with: taskURL, languageCode: self.languageCode, displayTitle: self.displayTitle) {
                         completion?(.success(newLocalTalkPage))
```
tonisevener: Maybe these completion calls within the waitInBackground should be wrapped up in a Dispatch.main.async?

Wikipedia/Code/TalkPagesController.swift
```diff
@@ -194,7 +176,7 @@ class TalkPageController {
                     return
                 }
                 
-                self.fetchAndUpdate(localTalkPage: talkPage, revisionID: newRevisionID, completion: { (result) in
+                self.fetchAndUpdateLocalTalkPage(with: talkPageObjectID, revisionID: newRevisionID, completion: { (result) in
                     switch result {
                     case .success:
                         completion(.success(.success))
```
tonisevener: This and line 184 also might need Dispatch.main.async.

Wikipedia/Code/TalkPagesController.swift
```diff
@@ -229,15 +217,7 @@ class TalkPageController {
...
-                        completion(.success(.topicMissingTalkPageRelationship))
-                    }
-                    return
-                }
-                
-                self.fetchAndUpdate(localTalkPage: talkPage, revisionID: newRevisionID, completion: { (result) in
+                self.fetchAndUpdateLocalTalkPage(with: talkPageObjectID, revisionID: newRevisionID, completion: { (result) in
                     switch result {
                     case .success:
                         completion(.success(.success))
```
tonisevener: This and line 225 also might need Dispatch.main.async.

Wikipedia/Code/TalkPagesController.swift
```diff
@@ -299,21 +279,34 @@ private extension TalkPageController {
...
+                            DispatchQueue.main.async {
+                                completion?(.failure(TalkPageError.fetchLocalTalkPageFailure))
+                            }
+                            return
+                        }
+                        switch result {
+                        case .success(let networkTalkPage):
+                            assert(networkTalkPage.revisionId != nil, "Expecting network talk page to have a revision ID here so it can pass it into the local talk page.")
+                            if let updatedLocalTalkPage = moc.updateTalkPage(localTalkPage, with: networkTalkPage) {
+                                completion?(.success(updatedLocalTalkPage))
```
tonisevener: Also might need a DispatchQueue.main.async around these.

Wikipedia/Code/TalkPagesController.swift
```diff
@@ -299,21 +279,34 @@ private extension TalkPageController {
         fetcher.fetchTalkPage(urlTitle: urlTitle, displayTitle: displayTitle, host: host, languageCode: languageCode, revisionID: revisionID, completion: completion)
     }
     
-    func fetchAndUpdate(localTalkPage: TalkPage, revisionID: Int, completion: ((Result<TalkPage, Error>) -> Void)? = nil) {
-        
+    func fetchAndUpdateLocalTalkPage(with moid: NSManagedObjectID, revisionID: Int, completion: ((Result<TalkPage, Error>) -> Void)? = nil) {
         fetchTalkPage(revisionID: revisionID) { (result) in
             DispatchQueue.main.async {
```
tonisevener: This DispatchQueue.main.async we could probably remove since perform switches you below anyways.

Wikipedia/Code/TalkPagesController.swift
nambatee: looks like `waitInBackgroundWithCompletion` notifies on the main queue by default - https://github.com/wikimedia/wikipedia-ios/blob/dad1b8ba65bbdb4fe29c8b1620a9ef4052078f9b/Wikipedia/Code/WMFTaskGroup.m#L48

Wikipedia/Code/TalkPagesController.swift
joewalsh: This one's for ensuring the `viewContext` is accessed from the main thread

Wikipedia/Code/TalkPagesController.swift
nambatee: missed that this was inside a `waitInBackground` and `perform`, disregard 👆

Wikipedia/Code/TalkPagesController.swift
tonisevener: aah I see 👍

PR #3109, opened 2019-05-29
User talk is not showing in History

nambatee:
https://phabricator.wikimedia.org/T224037

- added a `ns` attribute to `WMFArticle`, defaults to 0 (main)
- added `PageNamespace` enum - right now it includes 8 cases - changes welcome if you think it should include more or less cases
- modified `addPageToHistoryWithURL` to return the `WMFArticle` added to History
- added a call to `addPageToHistoryWithURL` when user taps on a user talk in search results


WMF Framework/PageNamespace.swift
```diff
@@ -0,0 +1,32 @@
+// https://en.wikipedia.org/wiki/Wikipedia:Namespace
+public enum PageNamespace: Int {
```
tonisevener: I also noticed there's a MWKTitleNamespace enum in the project, doesn't look like it's used though (and in obj-c), maybe we can clean it out?

joewalsh:
I updated the `ArticleSummary` response object to include the `Namespace` - this should ensure we get the correct namespace in cases when the full article isn't requested. I also noticed that user talk pages were appearing as continue reading cards for me, I updated the continue reading content source to exclude items that are excluded from the feed and made sure `isExcludedFromFeed` was updated everywhere the `ns` is set

PR #3110, opened 2019-05-29
talk page - added superscript and subscript handling

tonisevener:
https://phabricator.wikimedia.org/T222729 - superscript and subscript support

Endpoint currently seems to strip it out but here is native handling for it.

`let newTitle = "Testing with superscript<sup>test</sup> and subscript<sub>test</sub>."`

Simulator Screen Shot - iPhone X - 2019-05-29 at 16 27 34

`let newTitle = "Testing with superscript<sup><a href=\"http://www.google.com\">test</a></sup> and subscript<a href=\"http://www.google.com\"><sub>test</sub></a>."`

Simulator Screen Shot - iPhone X - 2019-05-29 at 16 29 12


montehurd:
cool! i can see about not stripping them in the endpoint!

nambatee:
nice catch 👍

PR #3112, opened 2019-05-31
Track usage of the "open here to edit" feature

nambatee:
https://phabricator.wikimedia.org/T218933


nambatee:
@chelsyx should the `ready` and `preview` events be logged only once or every time the user navigates to screens associated with those events?

For example, when I start editing an article, `ready` will be logged, then if I navigate to the Preview screen and go back to the Editing screen, should `ready` be logged again?

chelsyx:
> @chelsyx should the `ready` and `preview` events be logged only once or every time the user navigates to screens associated with those events?
> For example, when I start editing an article, `ready` will be logged, then if I navigate to the Preview screen and go back to the Editing screen, should `ready` be logged again?

We want to log these events only once.

> In some cases, when the user highlights text to edit, we won't be able to find that text in the editor and instead we'll show an alert (_The text that you selected could not be located_) - would you want to log that as an error with some special code?

Ideally, we want to log the unlocalized codes of this error like https://phabricator.wikimedia.org/T217554. But since T217554 is not fixed yet, and tracking this error is not super important to us, I'd suggest you log it in a way that's easier for you.

nambatee:
@chelsyx That's not a server error so we don't get an error - we would have to create a custom code, something that would make it identifiable to you. The easiest thing for me is not to log it at all.

chelsyx:
> @chelsyx That's not a server error so we don't get an error - we would have to create a custom code, something that would make it identifiable to you. The easiest thing for me is not to log it at all.

Let's log this error as "non-editable" then. For example:
```
event = {
    action = error;
    anon = 1;
    app_install_id = "ABCDE-12345-ABCDE-12345";
    client_dt = "2019-05-05T20:38:57-07:00";
    session_token = "ABCDE-12345-ABCDE-12345";
    source = highlight;
    errorText = "non-editable";
};
```

nambatee:
Sounds good, thanks @chelsyx!

Wikipedia/Code/EditFunnel.m
```diff
@@ -1,122 +0,0 @@
...
-    [self log:@{kActionKey: @"start"} language:language] ;
-}
-
-- (void)logPreview:(NSString *)language {
-    [self log:@{kActionKey: @"preview"} language:language];
-}
-
-- (void)logEditSummaryTap:(NSString *)editSummaryTapped language:(NSString *)language {
-    [self log:@{kActionKey: @"editSummaryTap",
-                @"editSummaryTapped": editSummaryTapped ? editSummaryTapped : @""} language:language];
```
montehurd: haha glad to be rid of this edit summary “tap” and “tapped” business 😂

Wikipedia/Code/EditFunnel.swift
```diff
@@ -0,0 +1,219 @@
...
+
+    private func logStart(source: EditFunnelSource, wikidataDescriptionEdit: WikidataDescriptionEdit? = nil, revision: Int, language: String) {
+        // session token should be regenerated at every 'start' event
+        sessionToken = singleUseUUID()
+        log(event(action: .start, source: source, sessionToken: sessionToken, wikidataDescriptionEdit: wikidataDescriptionEdit, revision: revision), language: language)
+    }
+
+    // MARK: Onboarding
+
+    @objc public func logOnboardingPresentationInitiatedBy(source: EditFunnelSource, revision: Int, language: String) {
```
montehurd: ‘logOnboardingPresentation(initiatedBy source: EditFunnelSource, revision: Int, language: String)’ may be a bit swiftier 😂 (other methods too?) not a big deal though

nambatee: like `logOnboardingPresentation(initiatedBy source: EditFunnelSource, revision: Int, language: String)` or do you have something else in mind?

PR #3121, opened 2019-06-10
Fix revealing the selected image when searching in Media Insert

nambatee:
https://phabricator.wikimedia.org/T221800#5150096

The proposed solution is to make the search in Media Insert behave more like the search in the Search tab - instead of exiting out of search by scrolling, users can tap "Cancel" or "Done" (keyboard)

Before

before

After

after


nambatee:
@carolynlimadeo would you rather keep the old behavior (before) or go with the new one (after)? 🤔

carolynlimadeo:
@nambatee the new version looks great! Let's go with that, thanks!

PR #3125, opened 2019-06-12
Talk pages - handle additional links

tonisevener:
https://phabricator.wikimedia.org/T222728

Loading state after ~2 seconds or so will be handled in https://phabricator.wikimedia.org/T225545.

https://github.com/wikimedia/wikipedia-ios/pull/3131 depends on this PR & is set to be merged into this branch.


Wikipedia/Code/TalkPageContainerViewController.swift
```diff
@@ -191,6 +192,121 @@ private extension TalkPageContainerViewController {
...
+        
+        var absoluteUrl: URL?
+        
+        if let firstPathComponent = url.pathComponents.first,
+            firstPathComponent == ".",
+            url.host == nil,
+            url.scheme == nil,
+            let siteURLHost = siteURL.host,
+            let siteURLScheme = siteURL.scheme {
+            absoluteUrl = URL(string: "\(siteURLScheme)://\(siteURLHost)/wiki/")
```
joewalsh: There's a `wmf_URL(withPath:` extension on URL that might be of use here. If that doesn't fit, is there a reason not to use `URLComponents` rather than assembling the string manually?

Wikipedia/Code/TalkPageContainerViewController.swift
```diff
@@ -191,6 +192,121 @@ private extension TalkPageContainerViewController {
...
+            
+            guard let self = self,
+            let loadingViewController = loadingViewController else {
+                return
+            }
+            
+            self.toggleLinkDeterminationState(loadingViewController: loadingViewController, shouldDisable: false)
+            
+            if error != nil {
+                self.openURLInSafari(url: absoluteURL)
```
joewalsh: If this is a network error, might be better to show it instead of bumping to Safari

Wikipedia/Code/TalkPageContainerViewController.swift
```diff
@@ -191,6 +192,121 @@ private extension TalkPageContainerViewController {
...
+                let languageCode = absoluteURL.wmf_language,
+                let siteURL = MWKLanguageLinkController.sharedInstance().language(forLanguageCode: languageCode)?.siteURL() {
+                
+                switch namespace {
+                case .userTalk:
+                    let lastPathComponent = url.lastPathComponent
+                    self.pushTalkPage(title: lastPathComponent, siteURL: siteURL)
+                case .user:
+                    self.showUserActionSheet(siteURL: siteURL, absoluteURL: absoluteURL)
+                default:
```
joewalsh: Should there be a case above `default` for `.main` to show an article in an article view controller instead of bumping to Safari?

Wikipedia/Code/TalkPageContainerViewController.swift
tonisevener: @joewalsh nope no reason, either of those sound better, thanks!

Wikipedia/Code/TalkPageContainerViewController.swift
tonisevener: @joewalsh true - I had noticed global user pages were returning 404s from the summary endpoint, which had me thinking maybe it would be good to give Safari a second chance to resolve it in case there are other cases like that. The 404 case gets caught in line [306](https://github.com/wikimedia/wikipedia-ios/blob/2b1aacaacd900c7db257fe656aa6a6722985155d/Wikipedia/Code/TalkPageContainerViewController.swift#L306) though, so I can easily change this without breaking that.

Wikipedia/Code/TalkPageContainerViewController.swift
tonisevener: @joewalsh agreed!

Wikipedia/Code/TalkPageContainerViewController.swift
tonisevener: @joewalsh used wmf_URL so feels a little better but cleaner might be to pass along a baseURL into wmf_attributedStringFromHTMLWithFont rather than resolving to absolute here. I explored that a little bit but ran into issues because the endpoint seems to prefix relatives with "./", checking for that and removing that path component within wmf_attributedStringFromHTMLWithFont felt risky at this point.

PR #3129, opened 2019-06-12
Default to auto-signing talk page discussions

nambatee:
https://phabricator.wikimedia.org/T220549#5254658

### DIFF
--- a/Wikipedia/Code/NSUserDefaults+WMFExtensions.swift
+++ b/Wikipedia/Code/NSUserDefaults+WMFExtensions.swift
@@ -533,7 +533,7 @@ let WMFSearchLanguageKey = "WMFSearchLanguageKey"
 
     var autoSignTalkPageDiscussions: Bool {
         get {
-            return bool(forKey: UserDefaults.Key.autoSignTalkPageDiscussions)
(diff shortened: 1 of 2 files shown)


Wikipedia/Code/NSUserDefaults+WMFExtensions.swift
```diff
@@ -533,7 +533,7 @@ let WMFSearchLanguageKey = "WMFSearchLanguageKey"
 
     var autoSignTalkPageDiscussions: Bool {
         get {
-            return bool(forKey: UserDefaults.Key.autoSignTalkPageDiscussions)
+            return (value(forKey: UserDefaults.Key.autoSignTalkPageDiscussions) as? Bool) ?? true
```
tonisevener: Might be better to set a default in [registerDefaults](https://github.com/wikimedia/wikipedia-ios/blob/f1fbf36090dea6ddf31426337e2eac67b287ac0e/Wikipedia/Code/AppDelegate.m#L29) so you can keep the bool(forKey) here and any automatic conversion it does under the hood for you.

nambatee: done

PR #3133, opened 2019-06-14
talk page - show local data if fetch fails

tonisevener:
I noticed a flash if the talk page endpoint fails and the user has local data to show. I think we should give users the chance to browse their local discussions/replies even if the endpoint fails and disable topic creation/replying. I didn't cover this in the [error state ticket](https://phabricator.wikimedia.org/T222725) so this is more of a proposition. I also refactored into a view state enum. Just a personal preference, it is more code but felt a little clearer.

Before if user has local data:
before3

After if user has local data:
after4

After if user does not have local data (unchanged from before):
after5


tonisevener:
@carolynlimadeo let me know what you think of this extra error state.

carolynlimadeo:
@tonisevener this looks good, can users pull to refresh to try and load the page again?

tonisevener:
> @tonisevener this looks good, can users pull to refresh to try and load the page again?

Not at the moment but we can spin off another task for adding one. Would you want pull to refresh on error ("unable to load talk page"), empty state & reply thread screen too?

carolynlimadeo:
@tonisevener, pull to refresh on the empty state and reply thread screen would be great, that way if a user fixes their connection issue we can possibly go a head and allow them to post after refreshing. Thanks!

tonisevener:
@carolynlimadeo done, added here - https://phabricator.wikimedia.org/T225942

Wikipedia/Code/TalkPageContainerViewController.swift
```diff
@@ -1,6 +1,106 @@
...
+            return true
+        case .fetchFinishedResultData, .fetchFinishedResultEmpty, .linkFinished, .linkFailure:
+            return false
+        }
+    }
+}
+
+extension TalkPageContainerViewState: Equatable {
+    
+    public static func ==(lhs: TalkPageContainerViewState, rhs:TalkPageContainerViewState) -> Bool {
```
nambatee: I trimmed it by switching on both sides, feel free to revert

Wikipedia/Code/TalkPageContainerViewController.swift
```diff
@@ -38,9 +90,51 @@ class TalkPageContainerViewController: ViewController, HintPresenting {
...
+                fakeProgressController.start()
+                navigationItem.rightBarButtonItem?.isEnabled = false
+            case .fetchInitialResultData:
+                navigationItem.rightBarButtonItem?.isEnabled = false
+                hideEmptyView()
+            case .fetchInitialResultEmpty:
+                navigationItem.rightBarButtonItem?.isEnabled = false
+                wmf_showEmptyView(of: .emptyTalkPage, theme: self.theme, frame: self.view.bounds)
+            case .fetchFinishedResultData:
+                fakeProgressController.stop()
```
nambatee: there's also a `finish()` method on the `fakeProgressController` that sets the progress to `1` before hiding the bar, might be useful for the happy states https://github.com/wikimedia/wikipedia-ios/blob/a1782a596d0cc21925ed24222e59bb53e6967517/WMF%20Framework/FakeProgressController.swift#L71

PR #3134, opened 2019-06-14
talk page - CoreData housekeeping

tonisevener:
Delete talk pages > 50

### DIFF
--- a/WMF Framework/Wikipedia.xcdatamodeld/Wikipedia 2.xcdatamodel/contents
+++ b/WMF Framework/Wikipedia.xcdatamodeld/Wikipedia 2.xcdatamodel/contents
@@ -70,6 +70,7 @@
         </fetchIndex>
     </entity>
     <entity name="TalkPage" representedClassName="TalkPage" syncable="YES" codeGenerationType="class">
+        <attribute name="dateAccessed" attributeType="Date" defaultDateTimeInterval="582233760" usesScalarValueType="NO" syncable="YES"/>
         <attribute name="displayTitle" attributeType="String" syncable="YES"/>
         <attribute name="key" attributeType="String" syncable="YES"/>
         <attribute name="revisionId" optional="YES" attributeType="Integer 64" minValueString="0" usesScalarValueType="NO" syncable="YES"/>
@@ -215,7 +216,7 @@
     <elements>
         <element name="ReadingList" positionX="-54" positionY="135" width="128" height="300"/>
         <element name="ReadingListEntry" positionX="-27" positionY="153" width="128" height="180"/>
-        <element name="TalkPage" positionX="-45" positionY="135" width="128" height="105"/>
+        <element name="TalkPage" positionX="-45" positionY="135" width="128" height="120"/>
         <element name="TalkPageReply" positionX="-27" positionY="153" width="128" height="120"/>
         <element name="TalkPageTopic" positionX="-36" positionY="144" width="128" height="180"/>
         <element name="TalkPageTopicContent" positionX="-36" positionY="135" width="128" height="90"/>
--- a/Wikipedia/Code/TalkPageLocalHandler.swift
+++ b/Wikipedia/Code/TalkPageLocalHandler.swift
@@ -28,7 +28,6 @@ extension NSManagedObjectContext {
         let talkPage = TalkPage(context: self)
         talkPage.key = url.wmf_talkPageDatabaseKey
         talkPage.displayTitle = displayTitle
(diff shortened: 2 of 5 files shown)


Wikipedia/Code/TalkPageLocalHandler.swift
```diff
@@ -124,7 +135,7 @@ extension NSManagedObjectContext {
     func removeUnlinkedTalkPageTopicContent() throws {
         let request: NSFetchRequest<TalkPageTopicContent> = TalkPageTopicContent.fetchRequest()
         request.predicate = NSPredicate(format: "topics.@count == 0")
-        let results = try request.execute()
+        let results = try fetch(request)
```
tonisevener: before this change I was getting a "Cannot fetch without an NSManagedObjectContext in scope}" console error and TalkPageTopicContent objects were not clearing out in housekeeping.

WMF Framework/Wikipedia.xcdatamodeld/Wikipedia 2.xcdatamodel/contents
```diff
@@ -70,6 +70,7 @@
         </fetchIndex>
     </entity>
     <entity name="TalkPage" representedClassName="TalkPage" syncable="YES" codeGenerationType="class">
+        <attribute name="dateAccessed" attributeType="Date" defaultDateTimeInterval="582233760" usesScalarValueType="NO" syncable="YES"/>
```
tonisevener: This defaults to the date/time I created the attribute (06/14/2019), let me know if there's a better one.

Wikipedia/Code/WMFDatabaseHouseKeeper.swift
```diff
@@ -6,9 +6,31 @@ import Foundation
...
+     
+    */
+    private func deleteStaleTalkPages(_ moc: NSManagedObjectContext) throws {
+        let request: NSFetchRequest<TalkPage> = TalkPage.fetchRequest()
+        request.sortDescriptors = [NSSortDescriptor(key: "dateAccessed", ascending: false)]
+        request.fetchOffset = 50
+        
+        let oldTalkPages = try moc.fetch(request)
+        for talkPage in oldTalkPages {
+            moc.delete(talkPage)
```
nambatee: might be a good use case for a batch delete https://developer.apple.com/library/archive/featuredarticles/CoreData_Batch_Guide/BatchDeletes/BatchDeletes.html

Wikipedia/Code/TalkPageLocalHandler.swift
```diff
@@ -20,14 +20,23 @@ extension NSManagedObjectContext {
         fetchRequest.fetchLimit = 1
         fetchRequest.predicate = NSPredicate(format: "key == %@", databaseKey)
         
-        return try fetch(fetchRequest).first
+        let talkPage = try fetch(fetchRequest).first
+        talkPage?.dateAccessed = Date()
```
joewalsh: I'm a bit concerned that the method for fetching a talk page also mutates it and saves the context - could I move this into a separate function that's explicitly called when the user views the talk page or is there some other consideration that I'm missing?

joewalsh: I think I misread "accessed" to mean "viewed by the user" - is the intent to have this property updated anytime the object is accessed even by non-user task?

tonisevener: @joewalsh true, agreed, non-user task shouldn't update this in my opinion. I really meant viewed by the user & wanted to keep track of what talk pages they seem to be interested in lately and update that date. Then later I sort by that date so we are deleting talk pages that they don't seem to be visiting anymore. To keep it simpler we could probably only timestamp this after it's fetched in the view controller so we know it was instigated by the user.

joewalsh: @tonisevener makes sense. I pushed a change, but feel free to alter or revert if you had other ideas

tonisevener:
Updated PR feedback -

1. I removed dateAccessed timestamp from TalkPageLocalHandler, I think only setting it in TalkPageContainerVC should be enough.
2. Added batch delete in `deleteStaleTalkPages` & `removeUnlinkedTalkPageTopicContent` and am merging the changes in case the user has that talk page in memory. I didn't bother with reacting on the UI-side to a talk page deleting. I don't really see how a talk page currently on screen could become the 50th talk page sorted by access date and delete out from under them in housekeeping, so it seems unnecessary. But, I'm happy to handle if you all think otherwise.

PR #3135, opened 2019-06-17
Preserve navigation stack when preserving last opened article

nambatee:
https://phabricator.wikimedia.org/T221859

Open questions

- Is there a better way to structure saving/restoring nav state?
- Should the controller be attached to dataStore  and save last open article URL as a separate WMFKeyValue to avoid UserDefaults?

update housekeeping after https://github.com/wikimedia/wikipedia-ios/pull/3134 is merged


Wikipedia/Code/WMFDatabaseHouseKeeper.swift
```diff
@@ -132,7 +133,19 @@ import Foundation
             let referencedKeysPredicate = NSPredicate(format: "!(key IN %@)", referencedArticleKeys)
             articlesToDeletePredicate = NSCompoundPredicate(andPredicateWithSubpredicates:[articlesToDeletePredicate,referencedKeysPredicate])
         }
-        
+
+        let oldNavigationStackArticlesToDelete: NSPredicate?
+        if let preservedArticleKeys = navigationStateController.allPreservedArticleKeys(in: moc) {
```
joewalsh: It seems like `preservedArticleKeys` could be added to `referencedArticleKeys` above to eliminate the need for this additional predicate

joewalsh:
> Should the controller be attached to dataStore and save last open article URL as a separate WMFKeyValue to avoid UserDefaults?

@nambatee I made an attempt at that on the [remove_open_article_url_from_user_default](https://github.com/wikimedia/wikipedia-ios/compare/feature/T221859...remove_open_article_url_from_wmf_defaults?expand=1) branch, let me know if it's in line with what you were thinking

Wikipedia/Code/WMFAppViewController.m
```diff
@@ -865,8 +866,8 @@ - (void)resumeApp:(dispatch_block_t)completion {
                            completion:^(BOOL didProcess) {
                                done();
                            }];
-        } else if ([self shouldShowLastReadArticleOnLaunch]) {
-            [self showLastReadArticleAnimated:NO];
+        } else if (NSUserDefaults.wmf.shouldRestoreNavigationStackOnResume) {
+            [self.navigationStateController restoreNavigationStateFor:self.navigationController in:self.dataStore.viewContext];
```
joewalsh: Since this is asynchronous, the explore feed is showing before the navigation stack is restored. Could this either be made synchronous or could the splash screen remain up until this finishes?

Also, is the login call in the state controller necessary if it's happening again in finishResumingApp?

Wikipedia/Code/WMFDatabaseHouseKeeper.swift
nambatee: `viewedDate` won't be `NULL` for the articles from the stack

Wikipedia/Code/WMFAppViewController.m
nambatee: `finishResumingApp` is happening after restoring so there's no guarantee that the user gets logged in before we attempt to show Account VC

Wikipedia/Code/WMFDatabaseHouseKeeper.swift
joewalsh: That should be fine. We only want to delete an article if the `viewedDate` is `NULL` AND it's not referenced by the navigation stack or the explore feed. If it's the case now that all articles returned by `allPreservedArticleKeys` have `viewedDate` as non-null, they're already covered by this predicate and we could remove the consideration entirely. Including `allPreservedArticleKeys` keys in  `referencedArticleKeys` as with 9a9c6b would handle any future state when articles are on the stack with a null `viewedDate` - maybe some other view that has an article that isn't persisted to history, but I'm fine with removing that as well.

nambatee: So previously, we were not saving any of the articles fetched in Article VC to disk and now we're saving all of them, and we want to keep most of them? I was thinking that since we're saving all the articles, we would want to be more aggressive about removing them (in the nav stack case)

Wikipedia/Code/WMFAppViewController.m
joewalsh: It seems the Settings VC and Account VC should be able to be shown with ambiguous login state (spinner or otherwise), then respond to login state changes, either by showing the correct info or an error that they aren't actually logged in. This way the app launch and state restoration isn't dependent on a network call. I've noticed this in the past with the app, if I navigate quickly enough to settings, it shows me as logged out even though the app eventually logs me in.

Wikipedia/Code/WMFDatabaseHouseKeeper.swift
joewalsh: OK - I see where I misunderstood. Currently deletion from Core Data and from disk are linked because only saved articles are stored on disk and need to be deleted from disk. In the change you made, it deletes older articles from Core Data in order to delete them from disk. However, deleting them from Core Data would also remove them from history. It seems like there needs to be an additional consideration to remove these articles from disk while keeping them in Core Data. Do I have that right?

nambatee: Yup that's it 👍 I missed the history deletion part. 😬  So would it make sense to append the urls of the articles where

`"viewedDate != NULL && savedDate == NULL && placesSortOrder == 0 && isExcludedFromFeed == FALSE"` AND `key NOT IN the union of referenced keys & preserved keys`

without deleting them in Core Data

or we could remove the `viewedDate` condition from the predicate and then guard against `viewedDate == nil` before core data deletion

joewalsh: I did an initial implementation and the problem I'm seeing is that since the Core Data state remains the same for articles not in the nav stack but still in history, the same set gets added to the disk removal queue every time the app quits. This would be problematic for users with a lot of articles in history as the app would have to iterate through and try to delete all of them every time the app exits. We could use `isDownloaded` to signal whether or not the article is cached and set it to false when returning it with the list of URLs. Currently `isDownloaded` is only used by the saved articles fetcher and includes images so it's set to false to trigger a re-download of images when a new revision is fetched (maybe this is the stuck downloading bug?). We could work around this by adding a separate method to trigger an image update. The other caveat is that this is all going to be re-worked for `mobile-html` so I'm not sure how much effort it's worth now

joewalsh: It seems like a separate `isCached` property is clearer, keeping the `isDownloaded` property for the saved articles fetcher. I tried this approach in b254d22

Wikipedia/Code/WMFAppViewController.m
nambatee: I think the explore feed might be showing before the nav stack's restored because Explore is the first VC that's being restored. I took out the `dispatch_async` here https://github.com/wikimedia/wikipedia-ios/pull/3135/commits/bf97cd236b23123dc2d5aabbee0fe6712705f576 assuming that `resumeApp` will always be called from the main thread - is that what you were thinking?

nambatee: Settings can handle login ambiguity but there are 2 separate VCs for logged in / logged out state so it'd require merging those 2 into 1 VC

nambatee: I think merging them and not being dependent on that call would be ideal but might take more time

joewalsh: That main queue async can remain as it doesn't affect the issue I'm seeing. What I'm seeing is the splash screen is removed, the stack restoration is kicked off, the stack restoration makes a network call to verify I'm logged in before restoring the stack, because of that async network call the explore feed shows since the splash screen was removed, then the whole stack is restored when the network call returns. When logged out, it seems to work as expected because there's no async network call to verify the user is logged in.

joewalsh: If merging and updating the VCs would take too long, a quicker option would be to only attempt login when the account VC is on the stack and ensure the splash screen stays visible until the login attempt callback is called asynchronously. This would ensure most app launches won't be slowed down by the login call and also ensure the explore screen doesn't show briefly before the restored nav stack when the login call needs to be made

Wikipedia/Code/AppDelegate.m
```diff
@@ -81,7 +81,7 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
 
     [NSUserDefaults wmf_migrateToWMFGroupUserDefaultsIfNecessary];
     [[NSUserDefaults wmf] wmf_migrateFontSizeMultiplier];
-    NSUserDefaults.wmf.shouldShowLastReadArticleOnResume = [self shouldShowLastReadArticleOnResume];
+    NSUserDefaults.wmf.shouldRestoreNavigationStackOnResume = [self shouldRestoreNavigationStackOnResume];
```
tonisevener: I might just be misremembering the original ask on these dates but at this point seems like wmf_appBecomeActiveDate isn't set for this session yet so it reflects the last persisted wmf_appBecomeActiveDate from their last session. You won't get an accurate date to verify that it is launched 5 hours later since they last resigned.

Wikipedia/Code/TalkPageContainerViewController.swift
```diff
@@ -4,12 +4,12 @@ import UIKit
...
-    private let talkPageTitle: String
-    private let siteURL: URL
-    private let type: TalkPageType
+    let talkPageTitle: String
+    let siteURL: URL
+    let type: TalkPageType
     private let dataStore: MWKDataStore
     private let controller: TalkPageController
-    private let talkPageSemanticContentAttribute: UISemanticContentAttribute
+    let talkPageSemanticContentAttribute: UISemanticContentAttribute
```
tonisevener: Doesn't seem like this is used outside anywhere, I'll test to see if it affects anything. Might need to add to the Info struct.

nambatee: it was needed to create `TalkPageReplyListViewController` initially but now I'm using `pushToReplyThread` so doesn't need to be public anyone 👍

Wikipedia/Code/AppDelegate.m
nambatee: great catch, fixed!

WMF Framework/NavigationState.swift
```diff
@@ -0,0 +1,95 @@
...
+            guard let kind = kind else {
+                return nil
+            }
+            self.kind = kind
+            self.presentation = presentation
+            self.info = info
+            self.children = children
+        }
+        
+        public enum Kind: Int, Codable {
```
tonisevener: > Is there a better way to structure saving/restoring nav state?

One thought is you could try splitting up the info struct into kind-specific info structs, then pass them in as associated values to this Kind enum. That way you don't have to deal with all the optionals and can guarantee it's set up with all the info it needs to restore itself if the correct kind is set. Not sure if that will give you grief with Codable though.

nambatee: associated values won't work with Codable 😞 I saw workarounds (https://www.objc.io/blog/2018/01/23/codable-enums/) but I thought it wasn't worth the manual encoding

Wikipedia/Code/WMFAppViewController.m
```diff
@@ -855,23 +855,30 @@ - (void)resumeApp:(dispatch_block_t)completion {
...
                            completion:^(BOOL didProcess) {
                                done();
                            }];
-        } else if ([self shouldShowLastReadArticleOnLaunch]) {
-            [self showLastReadArticleAnimated:NO];
+        } else if (NSUserDefaults.wmf.shouldRestoreNavigationStackOnResume) {
+            [self.navigationStateController restoreNavigationStateFor:self.navigationController in:self.dataStore.viewContext with:self.theme completion:^{
+                [self hideSplashViewAnimated:!didShowOnboarding];
+            }];
             done();
```
joewalsh: `done();` should probably move into the completion as well - would ensure two login requests aren't kicked off at the same time

PR #3136, opened 2019-06-17
Add Giphy sticker pack to app package

nambatee:
https://phabricator.wikimedia.org/T225653

I added the stickers that matched the size of the current stickers (408 x 408) and left the `Accessibility` field blank because it defaults to "Same as name".

For stickers with the same names, I added "animated" to the new ones ("Citation needed animated") and for the Tea icons, I added "dark skin tone" and "light skin tone" following https://emojipedia.org/search/?q=clap - let me know if the names should change @carolynlimadeo 

The new stickers are below the old stickers - should that change?

First time adding stickers so might be missing something


carolynlimadeo:
@nambatee this sounds great! Names look good! Thanks so much!

nambatee:
could I get a review on this?

PR #3150, opened 2019-06-25
Talk Pages Design QA fixes

tonisevener:
https://phabricator.wikimedia.org/T215928#5266987

- [x] Background color not full bleed on 'Please be kind' toast on iPhone X
- [x] Navigating to a user talk page from history opens in the article mode
- [x] Unable to scroll on messages with very long titles
- [x] Scroll issue on iPhone X when a user has 11 messages


nambatee:
lgtm 👍 I'm seeing a different offset issue on X with 11 messages but it's not caused by the new changes so we can fix it in a separate PR

ezgif-3-64799e8620c5

PR #3153, opened 2019-06-26
Talk Page Topics become hidden after rotation

nambatee:
https://phabricator.wikimedia.org/T226658
https://github.com/wikimedia/wikipedia-ios/pull/3150#issuecomment-505971636

Reproducible on iPhone X with 11 topics ^ - for some reason, the nav bar's frame doesn't get recalculated the same way it does in other cases (portrait vs landscape) - instead of getting the right frame somewhere before `updateScrollViewInsets` is called in `ViewController`, it gets it after all the calls `updateScrollViewInsets` are done

also, updated trailing & leading constraints for `TalkPageHeaderView` to relate to safe area instead of superview so that the header is not covered by the notch - I set `20` as the constant, feel free to update


Wikipedia/Code/TalkPageTopicListViewController.swift
```diff
@@ -68,9 +68,13 @@ class TalkPageTopicListViewController: ColumnarCollectionViewController {
         completedActivityType = nil
     }
     
-    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
-        super.traitCollectionDidChange(previousTraitCollection)
-        cellLayoutEstimate = nil
```
joewalsh: was it intentional to remove `cellLayoutEstimate = nil`?

nambatee: no, thanks for catching!

PR #3154, opened 2019-06-26
talk page - fix iPad popover crashes

tonisevener:
https://phabricator.wikimedia.org/T226653
https://phabricator.wikimedia.org/T226654


Wikipedia/Code/UITextView+Extensions.swift
```diff
@@ -0,0 +1,16 @@
...
+extension UITextView {
+    func frame(of textRange: NSRange) -> CGRect? {
+        selectedRange = textRange
+        
+        var rect: CGRect?
+        if let uiTextRange = selectedTextRange {
+            rect = firstRect(for: uiTextRange)
+        }
+        
+        selectedRange = NSRange(location: 0, length: 0)
```
joewalsh: Would it make sense to store the previously selected range in case this is used in a case where selection is available?

tonisevener: @joewalsh agreed, done!

PR #3155, opened 2019-06-26
Fix continue reading layout, ensure URLs are unique

joewalsh:
The content group URL for continue reading was the same no matter which article was displayed. Since this URL is used as the layout cache key, it would cause layout issues when the continue reading article changed.

https://phabricator.wikimedia.org/T219046

### DIFF
--- a/Wikipedia/Code/WMFContinueReadingContentSource.m
+++ b/Wikipedia/Code/WMFContinueReadingContentSource.m
@@ -34,7 +34,7 @@ static NSTimeInterval const WMFTimeBeforeDisplayingLastReadArticle = 60 * 60 * 2
 }
 
 - (void)loadNewContentInManagedObjectContext:(NSManagedObjectContext *)moc force:(BOOL)force completion:(nullable dispatch_block_t)completion {
-    NSURL *lastRead = [self.userDataStore.historyList mostRecentEntryInManagedObjectContext:moc].URL;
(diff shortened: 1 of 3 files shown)


Wikipedia/Code/WMFContinueReadingContentSource.m
```diff
@@ -34,7 +34,7 @@ - (void)stopUpdating {
 }
 
 - (void)loadNewContentInManagedObjectContext:(NSManagedObjectContext *)moc force:(BOOL)force completion:(nullable dispatch_block_t)completion {
-    NSURL *lastRead = [moc wmf_openArticleURL] ?: [self.userDataStore.historyList mostRecentEntryInManagedObjectContext:moc].URL;
+    NSURL *lastRead = [self.userDataStore.historyList mostRecentEntryInManagedObjectContext:moc].URL;
```
nambatee: should we get rid of `wmf_openArticleURL` completely and use `mostRecentEntryInManagedObjectContext` in other places? 🤔 https://github.com/wikimedia/wikipedia-ios/blob/27cd80b3aacdc36cbc3d541006337f318a4185d1/Wikipedia/Code/WMFAppViewController.m#L1505

joewalsh: I think there's still a useful differentiation between `wmf_openArticleURL` and `mostRecentEntryInManagedObjectContext` - the former is an article that's open and being actively read, the latter is just the most recently read article, regardless of whether or not it's open. Either way, I think I should revert the removal of `wmf_openArticleURL` here and we can discuss that separately

PR #3156, opened 2019-06-27
Talk page - fix inset bug when returning from state restoration

tonisevener:
https://phabricator.wikimedia.org/T226732

For some reason the navigation bar maxY doesn't include the safe area height when returning from state restoration. Not sure why but low risk fix is to relayout the navigation bar & recalculate the scroll view insets on appearance when coming from state restoration.

Before:
before

After:
after


joewalsh:
This might be a more widespread problem. It seems like anytime the app triggers an update of scroll view insets the navigation bar should get a `layoutIfNeeded` call to ensure the heights are accurate. I tried that approach in the [layoutIfNeeded branch](https://github.com/wikimedia/wikipedia-ios/compare/layoutIfNeeded?expand=1) and it seems to fix this issue for me, let me know if you you see the same

tonisevener:
> This might be a more widespread problem. It seems like anytime the app triggers an update of scroll view insets the navigation bar should get a `layoutIfNeeded` call to ensure the heights are accurate. I tried that approach in the [layoutIfNeeded branch](https://github.com/wikimedia/wikipedia-ios/compare/layoutIfNeeded?expand=1) and it seems to fix this issue for me, let me know if you you see the same

Still seeing the bug on that branch :/ I also added a setNeedsLayout on the preceding line but that didn't work either.

PR #3161, opened 2019-06-28
Fetcher - pass along correct error object from result + tests

tonisevener:
- Fixes saved article sync loop when article is deleted on server

https://phabricator.wikimedia.org/T225822


WikipediaUnitTests/Code/ArticleFetcherTests.m
```diff
@@ -39,6 +39,51 @@ - (void)tearDown {
     return [[NSProcessInfo processInfo] wmf_isTravis] ? @[] : [super testInvocations];
 }
 
+- (void)testMissingTitleReturnsExpectedError {
```
joewalsh: 🎉

PR #3163, opened 2019-06-28
Handle redirects from summary responses

joewalsh:
- Update a `WMFArticle`'s key if it differs from what is returned by `/page/summary` in the case that the article was redirected
- Merge duplicate `WMFArticle`s if an article exists already for the redirected key
- Gracefully handle the duplicate article state in `SavedArticlesFetcher` to prevent infinite download loops (even though it now shouldn't get into this state, probably worth having the precaution)
- For redirected titles, remove the old title and add the new title to the remote reading list

https://phabricator.wikimedia.org/T227020


joewalsh:
One outstanding issue is that this change causes duplicate ‘WMFArticle’ objects with the same key - need a way to merge articles in this case or not create the duplicates in the first place

joewalsh:
This is ready for review into 6.3.1

WMF Framework/ReadingListsSyncOperation.swift
```diff
@@ -819,10 +814,25 @@ internal class ReadingListsSyncOperation: ReadingListsOperation {
                     return
                 }
                 
-                guard let entry = NSEntityDescription.insertNewObject(forEntityName: "ReadingListEntry", into: moc) as? ReadingListEntry else {
+                var entry = ReadingListEntry(context: moc)
```
nambatee: If this is created and the guard below returns, could this end up being saved as an empty (= all attributes would be nil) object? 🤔

joewalsh: good catch, fixed

WMF Framework/SummaryExtensions.swift
```diff
@@ -1,4 +1,58 @@
 extension WMFArticle {
+    func merge(_ article: WMFArticle) {
+        guard article.objectID != objectID else {
+            return
+        }
+        // merge important keys not set by the summary
+        let keysToMerge = ["savedDate", "placesSortOrder", "pageViews"]
+        // ensure these values still exist with an assertion. is there a way to use these directly?
+        assert([\WMFArticle.savedDate, \WMFArticle.placesSortOrder, \WMFArticle.pageViews].count == keysToMerge.count)
```
nambatee: looks like `let keysToMerge = [#keyPath(WMFArticle.savedDate), #keyPath(WMFArticle.placesSortOrder), #keyPath(WMFArticle.pageViews)]` might work, compiler flags members that don't exist

PR #3170, opened 2019-07-08
Universal link support for User:Talk

nambatee:
https://phabricator.wikimedia.org/T227117


WMF Framework/PageInfoFetcher.swift
```diff
@@ -0,0 +1,43 @@
+@objc(WMFPageInfoFetcher)
+public final class PageInfoFetcher: Fetcher {
```
joewalsh: Is there a reason the namespace returned by `/page/summary` via the `ArticleSummaryController` or `ArticleSummaryFetcher` couldn't be used?

Wikipedia/Code/NSUserActivity+WMFExtensions.m
```diff
@@ -241,7 +241,22 @@ - (WMFUserActivityType)wmf_type {
...
+                WMFTaskGroup *group = [WMFTaskGroup new];
+                [group enter];
+                [SessionSingleton.sharedInstance.dataStore.pageInfoFetcher fetchNamespaceValueForArticleWith:[self wmf_articleURL]
+                                                                                                  completion:^(NSNumber *_Nullable ns) {
+                                                                                                      if (ns.integerValue == PageNamespaceUserTalk) {
+                                                                                                          type = WMFUserActivityTypeUserTalk;
+                                                                                                      }
+                                                                                                      [group leave];
+                                                                                                  }];
+                [group wait];
```
joewalsh: This `wait` will block the main thread for a network request when this method is called in  `canProcessUserActivity`. This can result in the [app being terminated if the network call takes too long](https://developer.apple.com/library/archive/qa/qa1693/_index.html). One alternative option would be to have AppVC make this call asynchronously while keeping the splash screen up or showing some other progress indicator. Another option would be to push the ArticleVC with this URL and have it display the talk page as a child view controller when the summary response comes back with namespace == 3 (similar to how external URLs are handled in `- (void)showExternalURL:(NSURL *)externalURL {`)

WMF Framework/PageInfoFetcher.swift
nambatee: It's a heavier call that fetches more data that's needed for this case

Wikipedia/Code/NSUserActivity+WMFExtensions.m
nambatee: If AppVC is making an asynchronous call, wouldn't `- (BOOL)application:(UIApplication *)application continueUserActivity:(NSUserActivity *)userActivity restorationHandler:(void (^)(NSArray<id<UIUserActivityRestoring>> *__nullable restorableObjects))restorationHandler` still have to hang on the main thread to know what to return?

joewalsh: It can return true there like it did before based on `wmf_isWikiResource` then figure out whether it's a talk page asynchronously

WMF Framework/PageInfoFetcher.swift
joewalsh: It's likely the summary will be requested later anyway, requesting it here and having it cached by the networking layer for that later call is likely lighter weight overall. Also, `/page/summary` is cached server-side as well, I don't think the action API is

Quick unscientific benchmark shows that the two requests take about the same amount of time:
```
/Users/wmf/source/wikipedia-ios % for i in {1..10}; curl -s -w "%{time_total}\n" -o /dev/null -H "Accept: application/json; charset=utf-8; profile=\"https://www.mediawiki.org/wiki/Specs/Summary/1.1.2\"" -H "Accept-Encoding: gzip" https://en.wikipedia.org/api/rest_v1/page/summary/Barack_Obama
0.147791
0.136603
0.158135
0.156236
0.145728
0.156962
0.156832
0.158531
0.145259
0.196570
/Users/wmf/source/wikipedia-ios % for i in {1..10}; curl -s -w "%{time_total}\n" -o /dev/null -H "Accept: application/json; charset=utf-8;" -H "Accept-Encoding: gzip" "https://en.wikipedia.org/w/api.php?action=query&format=json&titles=Barack_Obama"
0.198058
0.186314
0.187906
0.192104
0.189743
0.184776
0.176015
0.196304
0.185752
0.187870
```

PR #3176, opened 2019-07-10
Show empty view when there's no Featured article (widget)

nambatee:
https://phabricator.wikimedia.org/T227137#5317900


FeaturedArticleWidget/FeaturedArticleWidget.swift
```diff
@@ -2,13 +2,35 @@ import UIKit
...
 import WMF
 
+private final class EmptyView: SetupView {
+    private let label = UILabel()
+
+    var theme: Theme = .widget
+
+    override func setup() {
+        super.setup()
+        label.text = WMFLocalizedString("featured-article-empty-title", value: "No Featured article available today", comment: "Title that displays when featured article is not available")
```
joewalsh: It looks like this isn't imported by the localization script... investigating

nambatee: thanks 🙌

PR #3178, opened 2019-07-12
Talk Page pull to refresh

tonisevener:
https://phabricator.wikimedia.org/T225942

Added pull to refresh functionality on talk page discussion list, reply thread, empty and error states.

Also moved share toolbar from the discussion list VC into the container VC. Code fits better there IMO and gives the user the ability to go to the wiki page in Safari in a server error situation.

Server error:
server_error

Empty state:
empty_state

### DIFF
--- a/Wikipedia/Code/EmptyRefreshingViewController.swift
+++ b/Wikipedia/Code/EmptyRefreshingViewController.swift
@@ -62,9 +62,5 @@ class EmptyRefreshingViewController: ViewController {
         view.backgroundColor = theme.colors.paperBackground
         scrollView?.backgroundColor = theme.colors.paperBackground
         refreshControl.tintColor = theme.colors.refreshControlTint
-        if let emptyView = emptyView {
-            EmptyRefreshingViewController.wmf_applyTheme(to: emptyView, with: theme)
-        }
-        
     }
 }
--- a/Wikipedia/Code/EmptyRefreshingViewController.swift
+++ b/Wikipedia/Code/EmptyRefreshingViewController.swift
@@ -1,6 +1,6 @@
 import Foundation
 
-protocol EmptyRefreshingViewControllerDelegate: class {
+protocol EmptyRefreshingViewControllerDelegate: AnyObject {
     func triggeredRefresh(refreshCompletion: @escaping () -> Void)
 }
 
--- a/Wikipedia/Code/EmptyRefreshingViewController.swift
+++ b/Wikipedia/Code/EmptyRefreshingViewController.swift
@@ -55,6 +55,10 @@ class EmptyRefreshingViewController: ViewController {
     override func apply(theme: Theme) {
         super.apply(theme: theme)
         
+        guard viewIfLoaded != nil else {
+            return
+        }
+        
         view.backgroundColor = theme.colors.paperBackground
         scrollView?.backgroundColor = theme.colors.paperBackground
         refreshControl.tintColor = theme.colors.refreshControlTint
--- a/Wikipedia/Code/EmptyRefreshingViewController.swift
+++ b/Wikipedia/Code/EmptyRefreshingViewController.swift
@@ -1,0 +1,66 @@
+import Foundation
+
+protocol EmptyRefreshingViewControllerDelegate: class {
+    func triggeredRefresh(refreshCompletion: @escaping () -> Void)
+}
+
+class EmptyRefreshingViewController: ViewController {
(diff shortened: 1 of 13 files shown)


Wikipedia/Code/EmptyRefreshingViewController.swift
```diff
@@ -0,0 +1,66 @@
...
+            self?.refreshControl.endRefreshing()
+        })
+    }
+    
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+    
+    override func apply(theme: Theme) {
+        super.apply(theme: theme)
```
nambatee: should this guard against view not being loaded like the parent does? 
```
        self.theme = theme
        guard viewIfLoaded != nil else {
            return
        }
```

Wikipedia/Code/EmptyRefreshingViewController.swift
```diff
@@ -0,0 +1,66 @@
+import Foundation
+
+protocol EmptyRefreshingViewControllerDelegate: class {
```
nambatee: `class` is supposed to be deprecated in favor of `AnyObject` in future versions of Swift https://forums.swift.org/t/class-only-protocols-class-vs-anyobject/11507

Wikipedia/Code/EmptyRefreshingViewController.swift
```diff
@@ -0,0 +1,66 @@
...
+    }
+    
+    override func apply(theme: Theme) {
+        super.apply(theme: theme)
+        
+        view.backgroundColor = theme.colors.paperBackground
+        scrollView?.backgroundColor = theme.colors.paperBackground
+        refreshControl.tintColor = theme.colors.refreshControlTint
+        if let emptyView = emptyView {
+            EmptyRefreshingViewController.wmf_applyTheme(to: emptyView, with: theme)
```
nambatee: Is this needed if `applyTheme` is called in `emptyViewOfType`? https://github.com/wikimedia/wikipedia-ios/blob/c5a287839aa3e4221f269b8fcea79d167746cceb/Wikipedia/Code/UIViewController%2BWMFEmptyView.m#L58

Wikipedia/Code/UIViewController+WMFEmptyView.h
```diff
@@ -36,6 +36,10 @@ typedef NS_ENUM(NSUInteger, WMFEmptyViewType) {
 - (void)wmf_applyThemeToEmptyView:(WMFTheme *)theme;
 @property (nonatomic, readonly, nullable) WMFEmptyView *wmf_emptyView;
 
++ (nullable WMFEmptyView *)emptyViewOfType:(WMFEmptyViewType)type action:(nullable SEL)action theme:(WMFTheme *)theme frame:(CGRect)frame;
```
nambatee: looks like this one is not used anywhere

Wikipedia/Code/EmptyRefreshingViewController.swift
tonisevener: @nambatee yeah, true, it works fine when I remove it. I was thinking it would be good for apply theme to work independent of that empty view instantiation. It would come into play if we were to ever add the theming switch button to the screen and they switched while the empty state was up. But I'm open to removing it, or at the very least figuring out why the compiler complains when try to have emptyView(apply: theme). I had to go through into the Objective-C class method to get around it. I'm probably missing something obvious with that one.

Wikipedia/Code/EmptyRefreshingViewController.swift
tonisevener: @nambatee good to know, done!

Wikipedia/Code/EmptyRefreshingViewController.swift
tonisevener: @nambatee done, added this on all the talk page related classes I missed this on 😬. It caused a couple of bugs that I fixed as a part of the PR feedback commit too.

Wikipedia/Code/EmptyRefreshingViewController.swift
tonisevener: @nambatee done ^

Wikipedia/Code/UIViewController+WMFEmptyView.h
tonisevener: @nambatee removed

PR #3188, opened 2019-07-25
talk page accessibility

tonisevener:
https://phabricator.wikimedia.org/T222730


Wikipedia/Code/TalkPageContainerViewController.swift
```diff
@@ -606,12 +619,31 @@ extension TalkPageContainerViewController {
     }
 
     private func showNoInternetConnectionAlertOrOtherWarning(from error: Error, noInternetConnectionAlertMessage: String = CommonStrings.noInternetConnection) {
+
         if (error as NSError).wmf_isNetworkConnectionError() {
-            WMFAlertManager.sharedInstance.showErrorAlertWithMessage(noInternetConnectionAlertMessage, sticky: true, dismissPreviousAlerts: true)
+            
+            if UIAccessibility.isVoiceOverRunning {
+                UIAccessibility.post(notification: UIAccessibility.Notification.announcement, argument: noInternetConnectionAlertMessage)
```
joewalsh: Would it be worth moving this upstream into WMFAlertManager for all alerts?

tonisevener: @joewalsh yeah maybe, though I think we ought to spin that off into another task since it would take some testing. I found some of the announcements I made would get interrupted without extra handling in that context so it could take a bit more work than just touching alert manager.

joewalsh: sounds good!

PR #3190, opened 2019-07-26
Removed "dv" Divehi language from unsupportedLanguagesCodes

simmunity:
* After messing with the www files where Divehi was already
  present, I reverted those file changes and removed "dv"
  from unsuppoted... and tested with dv.wikipedia.org
  and font support seems to work fine with iOS 12.2


joewalsh:
Added hold tag as this should wait for merge until after 6.3.1 is branched off

PR #3196, opened 2019-08-01
[OTRS] Images that open in article view cannot be deleted from history

nambatee:
https://phabricator.wikimedia.org/T229141

Per documentation, `stringByRemovingPercentEncoding` that we're using to construct `wmf_databaseKeyURL` returns _nil if the receiver contains an invalid percent-encoding sequence_ (https://developer.apple.com/documentation/foundation/nsstring/1409569-stringbyremovingpercentencoding?language=objc).

Some titles include "%" characters that are not part of a percent-encoded sequence ("LED_Filament_CloseUp_at_5_%_power" or "%s") - calling `stringByRemovingPercentEncoding` will result in a `nil` return. In practice, `wmf_databaseKey` will be incorrect and we won't be able to construct a correct article url to either view the article in ArticleVC or interact with it (toggle saved state, delete).

For articles like "%s" that are in the main namespace and can come up in search results, we could modify `wmf_databaseKeyURL` to include a fallback for `path` if `stringByRemovingPercentEncoding` fails.

Test case
1. Search for "%"
2. Select "%s"
3. Go to History
4. Attempt to view "%s" or delete it

### DIFF
--- a/Wikipedia/Code/WebViewController.m
+++ b/Wikipedia/Code/WebViewController.m
@@ -223,14 +223,7 @@ typedef NS_ENUM(NSUInteger, WMFFindInPageScrollDirection) {
                                                    url = [NSURL wmf_URLWithSiteURL:self.article.url escapedDenormalizedInternalLink:href];
                                                }
                                                url = [url wmf_urlByPrependingSchemeIfSchemeless];
-                                               // If stringByRemovingPercentEncoding succeeds, we will be able to create a correct wmf_databaseKeyURL
-                                               // and use it to create an article url. If it fails, wmf_databaseKeyURL will be incorrect and we'll fail to
-                                               // create a correct article url.
-                                               if ([url.path stringByRemovingPercentEncoding]) {
-                                                   [(self).delegate webViewController:(self) didTapOnLinkForArticleURL:url];
-                                               } else if (url) {
-                                                   [self wmf_openExternalUrl:url];
-                                               }
+                                               [(self).delegate webViewController:(self) didTapOnLinkForArticleURL:url];
                                            } else {
                                                // A standard external link, either explicitly http(s) or left protocol-relative on web meaning http(s)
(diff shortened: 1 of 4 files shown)


Wikipedia/Code/WebViewController.m
```diff
@@ -223,7 +223,13 @@ - (void)handleLinkClickedScriptMessage:(NSDictionary *)messageDict {
                                                    url = [NSURL wmf_URLWithSiteURL:self.article.url escapedDenormalizedInternalLink:href];
                                                }
                                                url = [url wmf_urlByPrependingSchemeIfSchemeless];
-                                               [(self).delegate webViewController:(self) didTapOnLinkForArticleURL:url];
+                                               [self.article.dataStore.articleSummaryController.fetcher fetchSummaryForArticleWithKey:[url wmf_databaseKey] priority:NSURLSessionTaskPriorityHigh completion:^(WMFArticleSummary * _Nullable summary, NSURLResponse * _Nullable response, NSError * _Nullable error) {
```
joewalsh: I'm concerned about introducing this delay without also addressing https://phabricator.wikimedia.org/T225545 - could this be handled the same way we currently handle a `WMFFetcher.unexpectedResponseError` and show an external url (line 1319). The namespace could be checked there - might require more info in the callback from the fetcher

joewalsh: Alternatively, this could be handled separately either with T225545 or in a separate PR, the fix for `stringByRemovingPercentEncoding` looks good

nambatee: Something like this https://github.com/wikimedia/wikipedia-ios/pull/3196/commits/8b19351722c3021e671e5ac0d5f2c97d53f441bc?

joewalsh: Yup! Looks good

PR #3198, opened 2019-08-02
Fix 2FA

joewalsh:
- More permissive check for ID values
- Remove force unwrapped optionals

https://phabricator.wikimedia.org/T228588


Wikipedia/Code/WMFAccountLogin.swift
```diff
@@ -102,15 +102,25 @@ public class WMFAccountLogin: Fetcher {
                     status == "UI",
                     let requests = clientlogin["requests"] as? [AnyObject]
                 {
-                    if let passwordAuthRequest = requests.first(where:{$0["id"]! as! String == "MediaWiki\\Auth\\PasswordAuthenticationRequest"}),
```
montehurd: oh god... why did i do that? 🤦‍♂️

PR #3199, opened 2019-08-06
dismiss findAndReplaceView when user taps X on find and replace header

eholshouser:
https://phabricator.wikimedia.org/T229523


eholshouser:
Good catch @tonisevener, I've pushed another commit that should resolve it for that use case as well.

PR #3203, opened 2019-08-09
Fix reading list creation crash

nambatee:
https://phabricator.wikimedia.org/T230167


Wikipedia/Code/ColumnarCollectionViewController.swift
```diff
@@ -176,11 +176,12 @@ class ColumnarCollectionViewController: ViewController, ColumnarCollectionViewLa
         return frame
     }
 
+    open var emptyViewTarget: AnyObject?
```
joewalsh: should this be weak to prevent a retain cycle?

joewalsh:
I think this is worth pulling into 6.2.3

PR #3207, opened 2019-08-19
Fix Core Data crash when holding on to db during suspension

nambatee:
This is the top crash in 6.3.2 and one of the top crashes in 6.3.1, 6.3 and possibly other versions. It happens in the background, during suspension: https://github.com/wikimedia/wikipedia-ios/blob/46c32220d4dcee06e02b908633b5bbb386a8be96/Wikipedia/Code/ReadingListsController.swift#L785.

The exception type is `SIGKILL` which means that the process was terminated by the system. The exception code `0xdead10cc` indicates that an application has been terminated by the OS because it held on to a file lock or sqlite database lock during suspension. https://developer.apple.com/library/archive/technotes/tn2151/_index.html

The recommendation is to request additional background time - we're doing that for housekeeping, library and feed updates.

Let me know if you think this is caused by something else and/or could be fixed in a better way.

### DIFF
--- a/Wikipedia/Code/MWKDataStore.m
+++ b/Wikipedia/Code/MWKDataStore.m
@@ -60,6 +60,8 @@ static NSString *const MWKImageInfoFilename = @"ImageInfo.plist";
 
 @property (nonatomic, strong) NSURL *containerURL;
 
+@property (readwrite, nonatomic) BOOL failedToUpdateLocalConfig;
+
 @end
 
 @implementation MWKDataStore
@@ -1648,7 +1650,12 @@ static uint64_t bundleHash() {
                                         }
                                         NSDictionary *generalProps = [siteInfo valueForKeyPath:@"query.general"];
                                         NSDictionary *readingListsConfig = generalProps[@"readinglists-config"];
-                                        [self updateReadingListsLimits:readingListsConfig];
+                                        if (self.isLocalConfigUpdateAllowed) {
+                                            [self updateReadingListsLimits:readingListsConfig];
+                                            self.failedToUpdateLocalConfig = NO;
+                                        } else {
+                                            self.failedToUpdateLocalConfig = YES;
+                                        }
                                         [taskGroup leave];
                                     });
                                 }];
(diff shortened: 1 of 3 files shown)


Wikipedia/Code/MWKDataStore.m
```diff
@@ -1664,7 +1671,12 @@ - (void)updateLocalConfigurationFromRemoteConfigurationWithCompletion:(nullable
                                             [taskGroup leave];
                                             return;
                                         }
-                                        [self updateLocalConfigurationFromRemoteConfiguration:remoteConfigurationDictionary];
+                                        if (self.isLocalConfigUpdateAllowed) {
+                                            [self updateLocalConfigurationFromRemoteConfiguration:remoteConfigurationDictionary];
+                                            self.failedToUpdateLocalConfig = NO;
+                                        } else {
+                                            self.failedToUpdateLocalConfig = YES;
```
tonisevener: @nambatee I feel like maybe you should keep track of separate failed flags here from the updateReadingListsLimits call above. Is there a chance that completion could flip failedToUpdateLocalConfig again before it's checked?

Wikipedia/Code/WMFAppViewController.m
```diff
@@ -1587,16 +1615,22 @@ - (void)checkRemoteAppConfigIfNecessary {
...
-    } else {
+    BOOL shouldCheckRemoteConfig = now - lastCheckTime >= WMFRemoteAppConfigCheckInterval || self.dataStore.failedToUpdateLocalConfig;
+    if (!shouldCheckRemoteConfig) {
         self.checkingRemoteConfig = NO;
+        return;
     }
+    self.dataStore.isLocalConfigUpdateAllowed = YES;
+    [self startRemoteConfigCheckBackgroundTask:^{
+        self.dataStore.isLocalConfigUpdateAllowed = NO;
+        [self endRemoteConfigCheckBackgroundTask];
```
tonisevener: @nambatee I think endRemoteConfigCheckBackgroundTask needs to be called in the updateLocalConfigurationFromRemoteConfigurationWithCompletion completion block. Instead here would be the place to cancel any tasks/operations because the background expiration has hit.

nambatee: Good point, I think it should be called in both places - per documentation (https://developer.apple.com/documentation/uikit/uiapplication/1623051-beginbackgroundtask), (in the handler) _Failure to end the task explicitly will result in the termination of the app_

tonisevener: Maybe this is fine actually, you could need it here to prevent the app from terminating. But do you also need it in updateLocalConfigurationFromRemoteConfigurationWithCompletion? Seems like you'd need to flag the system your operation is complete rather than wait for expiration but I could be getting it wrong. I was looking at how SavedArticlesFetcher works, `beginBackgroundTaskWithName:@"SavedArticlesFetch" ` expiration handler cancels the tasks and endBackgroundTask is called elsewhere.

tonisevener: @nambatee yeah, both places sounds good

Wikipedia/Code/MWKDataStore.m
nambatee: Good point, it can definitely flip

PR #3208, opened 2019-08-20
Change -webkit-text-size-adjust to font-size

tonisevener:
https://phabricator.wikimedia.org/T229497

I wasn't able to repro in iOS12 but Safari in iPadOS seems to not respond at all to -webkit-text-size-adjust (iPhone works fine). Tried setting the web view's [preferredContentMode](https://developer.apple.com/documentation/webkit/wkwebpagepreferences/3194426-preferredcontentmode?language=objc) to .mobile but no luck. Reworking to use font-size fixes it.


Wikipedia/Code/WKWebView+LoadAssetsHtml.swift
```diff
@@ -90,7 +90,8 @@ extension WKWebView {
         return """
             <style type='text/css'>
                 body {
-                    -webkit-text-size-adjust: \(fontSize)%;
+                    font-size = \(fontSize.intValue)%;
```
tonisevener: Unfortunately now this seems to be ignored. Still digging into why that is.

montehurd:
iirc this doesn't scale things like padding and margins proportionally... will double-check later today...

tonisevener:
> iirc this doesn't scale things like padding and margins (and images?) proportionally... will double-check later today...

Yeah, I didn't do any handling for padding and margins, preserving -webkit-text-size-adjust would be the best way to go. I filed a radar on this not working in iPadOS.

montehurd:
I found the old phab ticket!

These comments cover the relevant bits:

https://phabricator.wikimedia.org/T146690#2675532

https://phabricator.wikimedia.org/T146690#2675948

https://phabricator.wikimedia.org/T146690#2685703

montehurd:
I'll poke around a bit more but based on this https://phabricator.wikimedia.org/T146690#2675532 (there are a variety of unwanted side-effects to using `font-size`) I think we should wait for the upstream fix again...

joewalsh:
These changes look pretty much identical to iOS 12 with `-webkit-text-size-adjust` on the extreme small and large text end:

Simulator Screen Shot - iPad Pro (12 9-inch) (3rd generation) - 2019-09-03 at 17 03 49
Simulator Screen Shot - iPad Pro (12 9-inch) (3rd generation) - 2019-09-03 at 16 58 29

Simulator Screen Shot - iPad Pro (12 9-inch) (3rd generation) - 2019-09-03 at 17 03 45
Simulator Screen Shot - iPad Pro (12 9-inch) (3rd generation) - 2019-09-03 at 16 58 27


Given how late it is in the iOS 13 dev cycle, if there's not specific issues we can find with this implementation we should move forward with using `font-size` instead of `webkit-text-size-adjust`.

PR #3211, opened 2019-08-23
Update app's theme when system theme is updated

nambatee:
https://phabricator.wikimedia.org/T227123

* When system theme changes, we're notified about the change in multiple different methods - `traitCollectionDidChange(_:)`, `updateViewConstraints()`, `viewWillLayoutSubviews()` and `viewDidLayoutSubviews()` - I chose to use `traitCollectionDidChange(_:)`
* When we're notified of the change, I update* our theme if 1) the new system theme is different from the previous theme 2) the new app theme would be different from the one that's currently set
 *I don't update the theme right away, instead I broadcast our own `Theme.didChangeNotification` (we do the same thing when the theme change is initiated by the user in Settings or via Article reading controls) and VCs adjust their themes

Known issues:
* When the app goes into the background, it gets hit with system notifications for some reason - for example, if the current theme is set to dark, as you background the app, you'll see a notification about a system theme change - as if the theme changed from dark to default and then from default back to dark. Break in `adjustTheme`, after the `guard`, to reproduce.
* When I background the app, go to Settings to change the theme, then look at the app without making it active, all the native views are already themed (even though `traitCollectionDidChange` wasn't called). The same is not true for webViews - the theme gets fully applied only when the app becomes active again.


nambatee:
Updated the base branch and the description, ready for review

joewalsh:
`WMFAppViewController` is able to walk the VC hierarchy and apply the theme to all themeable view controllers (in `applyTheme: toNavigationControllers:` called by `applyTheme:`). This way, if `applyTheme` is called on `WMFAppViewController`, the whole app gets themed and only one place needs to handle listening for the trait collection change. It looks like since `WMFThemeableNavigationController` is the rootVC it's the only VC that always gets the trait collection change. It could pass that theme along to `WMFAppViewController` or the theme theme logic from `WMFAppViewController` could move into `WMFThemeableNavigationController`. I have a working proof of concept on the `trait_theme` branch.

It seems like this should also be a separate theme setting - either a separate "match system" theme in the list or something that automatically switches to the black theme from whatever the selected theme is. Probably worth discussing with the rest of the team.

nambatee:
Good idea, closing this in favor of [trait_theme](https://github.com/wikimedia/wikipedia-ios/commit/25b979e935da2794f0449fdc52144b06c7f0de7d). A couple things that are missing - adjusting theme on launch and making the initial onboarding themeable.

PR #3216, opened 2019-08-28
Link loading animation

tonisevener:
https://phabricator.wikimedia.org/T225545

Still lots of bugs left but making a WIP PR to get some early thoughts. Fixes should hopefully be smaller commits now.

- Made a container view controller class to handle link fetching & routing or embedding
- Shifted code in WMFArticleViewController a bit to allow for standalone fetching apart from viewWillAppear.
- Surfacing cancellable key from ArticleSummaryFetcher to be able to cancel task from cancel button on animation
- Wrapping Talk Pages and WMFArticleViewControllers in container view controller.

Remaining + known bugs:

 [x] Pushing to external safari page freezes talk page when you tap done to go back
 [x] Spotlight searching wikipedia article freezes when deeplinking to app
 [x] Status bar disappears on black theme article view and talk page
 [x] Some article paths do not embed in container, which breaks linking (like tapping article in explore cell)
 [x] Various code cleanup needed (todos, error handling, naming)
 [x] Support dynamic type on loading screen
 [x] Accessibility / VO of loading screen
 [x] Improve blank pause while html loads in article view. (why is nav bar hidden, show that + progress bar)
 [x] Check retain cycles
 [x] Clean up theming, guard viewIsLoaded in theming methods. Test with article theme change.
 [x] Talk page cancel on animation weirdly pushes on embed Safari
 [x] “Cancelled” banner pops up sometimes on article view at odd times
 [x] State restoration web view doesn't load on 3rd articleVC in stack


tonisevener:
Ready for review. Still need to dig into a new retain cycle on WMFArticleViewController but that's the last thing for me to fix.

joewalsh:
When tapping on an article in the explore feed, the article's image no longer scales to transition to the article view controller. The new container `LoadingFlowController` doesn't conform to `ImageScaleTransitionProviding`. In cases like explore where we're definitely presenting an article, could `WMFArticleViewController` be used directly? If that's not a good option, could `LoadingFlowController` conform to `ImageScaleTransitionProviding` through immediate display of the `WMFArticleViewController`?

WMF Framework/Fetcher.swift
```diff
@@ -111,6 +111,19 @@ open class Fetcher: NSObject {
...
     }
     
+    @discardableResult public func performMobileAppsServicesGET<T: Decodable>(for URL: URL?, pathComponents: [String], priority: Float = URLSessionTask.defaultPriority, cancellationKey: CancellationKey? = nil, completionHandler: @escaping (_ result: T?, _ response: URLResponse?,  _ error: Error?) -> Swift.Void) -> CancellationKey? {
+        
+        //The accept profile is case sensitive https://gerrit.wikimedia.org/r/#/c/356429/
+        let headers = ["Accept": "application/json; charset=utf-8; profile=\"https://www.mediawiki.org/wiki/Specs/Summary/1.1.2\""]
+        let taskURL = configuration.wikipediaMobileAppsServicesAPIURLComponentsForHost(URL?.host, appending: pathComponents).url
+        let key = cancellationKey ?? UUID().uuidString
+        let task = session.jsonDecodableTask(with: taskURL, headers: headers, priority: priority) { (result: T?, response: URLResponse?, error: Error?) in
+            completionHandler(result, response, error)
```
nambatee: Should this task be untracked in `jsonDecodableTask`'s completion?

Wikipedia/Code/FakeProgressLoading.swift
```diff
@@ -0,0 +1,14 @@
...
+
+protocol FakeProgressLoading {
+    var fakeProgressController: FakeProgressController { get }
+}
+
+extension TalkPageContainerViewController: FakeProgressLoading {
+    
+}
+
+extension TalkPageReplyListViewController: FakeProgressLoading {
```
nambatee: Is there a reason the empty conformances are declared here and not in the files?

Wikipedia/Code/LoadingAnimationViewController.swift
```diff
@@ -0,0 +1,90 @@
...
+    
+
+    @IBAction func tappedCancel(_ sender: UIButton) {
+        cancelBlock?()
+    }
+    
+    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        super.traitCollectionDidChange(previousTraitCollection)
+        cancelButton.titleLabel?.font = UIFont.wmf_font(.semiboldTitle3, compatibleWithTraitCollection: traitCollection)
+        loadingPageLabel.font = UIFont.wmf_font(.boldHeadline, compatibleWithTraitCollection: traitCollection)
```
nambatee: The fonts should also be set in `viewDidLoad` in case `traitCollectionDidChange` doesn't get called on iOS 13 initially (https://useyourloaf.com/blog/predicting-size-classes-in-ios-13/)

Wikipedia/Code/LoadingFlowController.swift
```diff
@@ -0,0 +1,368 @@
...
+        loadingAnimationViewController.apply(theme: theme)
+        flowChild.apply(theme: theme)
+        flowChild.customNavAnimationHandler?.apply(theme: theme)
+        wmf_applyTheme(toEmptyView: theme)
+    }
+}
+
+//MARK: Error Handling
+
+extension NSError {
```
nambatee: Was this meant to be an internal extension or could it be private?

tonisevener:
> When tapping on an article in the explore feed, the article's image no longer scales to transition to the article view controller. The new container `LoadingFlowController` doesn't conform to `ImageScaleTransitionProviding`. In cases like explore where we're definitely presenting an article, could `WMFArticleViewController` be used directly? If that's not a good option, could `LoadingFlowController` conform to `ImageScaleTransitionProviding` through immediate display of the `WMFArticleViewController`?

@joewalsh This is fixed now. LoadingFlowController conforms to ImageScaleTransitionProviding, though it just forwards it's calls down into is child if it conforms to ImageScaleTransitionProviding. The header view logic felt too specific to WMFArticleVC for me to want to move that up.

I also did some cleanup of LoadingFlowController. Now for all cases it immediately embeds it's child in viewDidLoad. It felt needlessly complicated to handle some paths embedding right away (i.e. from Explore to retain the transition) and other paths first waiting to fetch, then seeing if it's external, then embedding if not. Now LoadingFlowController is only a link tap listener that fetches / displays the loading animation / pushes on the appropriate next view controller.

WMF Framework/Fetcher.swift
tonisevener: @nambatee This is fixed, great catch!

Wikipedia/Code/FakeProgressLoading.swift
tonisevener: @nambatee nope no reason, I have moved these to the files.

Wikipedia/Code/LoadingAnimationViewController.swift
tonisevener: @nambatee Done

Wikipedia/Code/LoadingFlowController.swift
tonisevener: @nambatee Good point, I've made this private.

PR #3221, opened 2019-08-29
Fix iPad TOC scroll regression

montehurd:
https://phabricator.wikimedia.org/T231610

Just a quick restore of the related lines.

# Before
( TOC is static )
toc before mov

# After
( TOC tracks )
toc after mov


tonisevener:
@montehurd I notice if I arrive from a fragment link the web view scrolls to the right place but the TOC stays highlighted on the first section. I tried fixing it [here](https://github.com/wikimedia/wikipedia-ios/commit/54dbe7ff26361dd277555d8b4ee33fd1b1199622) - it's a little better but is highlighting the section directly before rather than the linked section so maybe there's something off with the offset logic or there's a better different fix.

montehurd:
@tonisevener Good find! I noticed a related bug which is fixed by adding a slight delay before scrolling to the fragment section. Will see if it helps the issue you're seeing and also see if there's a bit less hacky fix than the delay...

montehurd:
Moving fragment fixes to https://phabricator.wikimedia.org/T232022

PR #3223, opened 2019-08-30
iOS 13 and scale transforms for appearing and disappearing cells don't mix

joewalsh:
Previously, the scale transform didn't affect the `bounds.size` of the cell. On iOS 13 it does and breaks the layout. Remove the scale transform for now.


montehurd:
Seems like every other version of iOS they fiddle with these size/scale bits in a way we have to work around 😂

PR #3225, opened 2019-09-03
Match system theme for widgets and the news notification

joewalsh:
Part of https://phabricator.wikimedia.org/T227123

Before:
IMG_0017

After:
IMG_0018


ContinueReadingWidget/WMFTodayContinueReadingWidgetViewController.swift
```diff
@@ -2,7 +2,7 @@ import UIKit
 import NotificationCenter
 import WMF
 
-class WMFTodayContinueReadingWidgetViewController: UIViewController, NCWidgetProviding {
+class WMFTodayContinueReadingWidgetViewController: UIViewController, NCWidgetProviding, Themeable {
```
tonisevener: Seems like you could make this subclass ExtensionViewController instead.

TopReadWidget/WMFTodayTopReadWidgetViewController.swift
```diff
@@ -130,10 +128,39 @@ class WMFTodayTopReadWidgetViewController: UIViewController, NCWidgetProviding {
         perform(#selector(updateView), with: nil, afterDelay: 0.1)
     }
     
+    func updateThemeFromTraitCollection() {
```
tonisevener: Any reason for not using super's updateThemeFromTraitCollection? Looks like the same implementation.

joewalsh: forgot to delete this 😅

ContinueReadingWidget/WMFTodayContinueReadingWidgetViewController.swift
joewalsh: agreed - missed this class when I refactored

WMF Framework/ExtensionViewController.swift
```diff
@@ -0,0 +1,30 @@
+import UIKit
+
+open class ExtensionViewController: UIViewController, Themeable {
+    public final var theme: Theme?
```
nambatee: Would it make sense to set the default theme here instead of having every widget VC fall back on its own?

Wikipedia/Code/WMFArticlePreviewViewController.swift
```diff
@@ -67,4 +67,16 @@ open class WMFArticlePreviewViewController: UIViewController {
             self.view.layoutIfNeeded()
         }
     }
+    
+    public func apply(theme: Theme) {
+        guard viewIfLoaded != nil else {
+            return
+        }
+        titleTextColor = theme.colors.primaryText
```
nambatee: Shouldn't `WMFArticlePreviewViewController` have its own `theme` property that gets set if `apply:theme` is called before view loads (with `apply:theme` being called in `viewDidLoad` for those cases)?

TopReadWidget/WMFTodayTopReadWidgetViewController.swift
```diff
@@ -284,9 +286,7 @@ class WMFTodayTopReadWidgetViewController: UIViewController, NCWidgetProviding {
         stackViewHeightConstraint.isActive = true
         
         stackViewHeightConstraint.constant = size.height
-        
-        view.layoutIfNeeded()
```
nambatee: Was the removal of this `layoutIfNeeded` call intentional?

joewalsh: Initially yes but the refactor makes the removal no longer necessary

PR #3227, opened 2019-09-03
Match system theme

nambatee:
https://phabricator.wikimedia.org/T227123

See #3211 for context


Wikipedia/Code/WMFThemeableNavigationController.m
```diff
@@ -95,4 +100,31 @@ - (void)hideSplashViewAnimated:(BOOL)animated {
...
+- (void)debounceTraitCollectionThemeUpdate {
+    if (@available(iOS 12.0, *)) {
+        [NSObject cancelPreviousPerformRequestsWithTarget:self selector:@selector(_debounceTraitCollectionThemeUpdate) object:nil];
+        [self performSelector:@selector(_debounceTraitCollectionThemeUpdate) withObject:nil afterDelay:0.3];
+    }
+}
+
+- (void)traitCollectionDidChange:(UITraitCollection *)previousTraitCollection {
+    [super traitCollectionDidChange:previousTraitCollection];
+    [self debounceTraitCollectionThemeUpdate];
```
tonisevener: Premature optimization but since WMFThemeableNavigationController is used throughout the app, maybe we should have a property flag for whether it responds to theme changes or not? Only 1 instance will have a WMFApplicationViewController right, seems like a waste for every WMFThemeableNavigationController to respond with this logic.

joewalsh: Makes sense - maybe a delegate for receiving theme changes or trait collection changes that `WMFAppViewController` could set when it creates the root `WMFThemeableNavigationController` and it only responds here if that delegate property is set.

joewalsh: Made this change - still WIP pending discussion of https://phabricator.wikimedia.org/T227123#5461131

joewalsh:
Ready for review - merging #3225 first and updating will clean up the diff

Wikipedia/Code/WMFWelcomePageViewController.swift
```diff
@@ -12,10 +12,30 @@ public protocol WMFWelcomeNavigationDelegate: class{
...
+        self.theme = theme
+        guard viewIfLoaded != nil else {
+            return
+        }
+        nextButton.setTitleColor(theme.colors.link, for: .normal)
+        nextButton.setTitleColor(theme.colors.disabledText, for: .disabled)
+        nextButton.setTitleColor(theme.colors.link, for: .highlighted)
+        pageControl?.pageIndicatorTintColor = theme.colors.pageIndicator
+        pageControl?.currentPageIndicatorTintColor = theme.colors.pageIndicatorCurrent
+        skipButton.setTitleColor(theme.colors.unselected, for: .normal)
```
nambatee: Accessing `pageControl` (walking through the subviews to get the first subview that is of type `UIPageControl`) here (view loaded but not appeared) seems to be affecting the Skip button (not getting touch events), still trying to figure out why

nambatee: pushed a fix https://github.com/wikimedia/wikipedia-ios/pull/3227/commits/09922424256acb8c9ff0811b2ca75bc9a2dc25f5

tonisevener:
Made some tweaks for this one -

- App onboarding - language cells & picker and section editor onboarding weren't responding to system dark mode change
- made constant for "standard" and "default"
- hiding Light option & Default subtitle on <iOS13 in settings

montehurd:
ship🚢it🛳it⚓️it⛴it📦it!

PR #3229, opened 2019-09-03
iOS 13 [Bug] Move setup code from traitCollectionDidChange

nambatee:
https://phabricator.wikimedia.org/T230132

https://developer.apple.com/documentation/ios_ipados_release_notes/ios_ipados_13_beta_6_release_notes

> The traitCollectionDidChange(_:) method is only called when the value of a trait changes. Importantly, because the trait collection is now initialized to a prediction of the ultimate traits in the destination hierarchy, when the initial predicted traits match the ultimate traits in the hierarchy, traitCollectionDidChange(_:) will not be called when the trait environment is added to the hierarchy. Because traitCollectionDidChange(_:) is intended to be an invalidation callback to notify you that one or more traits changed, audit your existing implementations of this method, as well as the UIContentContainer method willTransition(to:with:), for places where you might have been relying on it to trigger initial setup. The best place to lazily perform work that uses the traitCollection is inside one of the layoutSubviews methods discussed above, but remember that these layout methods are called any time layout occurs so be sure to avoid repeating work when you don’t need to.


nambatee:
that appears to be correct - for some reason, I thought that the `traitCollection` wouldn't be correct at that time, but if that's the case then we should get a `didChange` call so setup & viewDidLoad should work

PR #3230, opened 2019-09-04
Tapping link with fragment scrolls to wrong area in article on iOS 13

montehurd:
https://phabricator.wikimedia.org/T232022

~~Should probably wait until https://github.com/wikimedia/wikipedia-ios/pull/3221 is merged.~~ Merged!

Fixes these somewhat related issues:
- Article sometimes jumping to wrong place (mostly on iOS 13 - but also observed on 11 and 12) for a fragment
- TOC sometimes highlighting previous section after fragment jump

# Before (iPad Mini 4 iOS 13)
ipad bug mov

# After
ipad after mov


montehurd:
Note: there's another [TOC issue](https://phabricator.wikimedia.org/T232141) which when fixed should make this commit https://github.com/wikimedia/wikipedia-ios/pull/3230/commits/fdf11fb6af6a32fb6cdcf6f3590e233231db9c64 (from this PR) unnecessary.

Wikipedia/Code/WebViewController.m
```diff
@@ -773,13 +773,16 @@ - (void)scrollToFragment:(NSString *)fragment animated:(BOOL)animated completion
...
                                                      if (!CGRectIsNull(rect)) {
-                                                         [self.webView.scrollView wmf_safeSetContentOffset:CGPointMake(self.webView.scrollView.contentOffset.x, rect.origin.y + [self.webView iOS12yOffsetHack] + self.delegate.navigationBar.hiddenHeight)
-                                                                                                  animated:animated
-                                                                                                completion:^(BOOL finished){
-                                                                                                    if (completion) {
-                                                                                                        completion();
-                                                                                                    }
-                                                                                                }];
+                                                         dispatchOnMainQueueAfterDelayInSeconds(0.1, ^{
```
joewalsh: the repro steps in the GIF seem fixed without the delay. What's the reason for it?

montehurd: iirc it occurs intermittently. have seen it first time using the link, but also when tapping back then using the link again

montehurd: frequency varied by device too - saw much more often on iPad mini 4

montehurd: (actual device)

montehurd: just updated the mini to beta 2 - lemme re-test...

montehurd: here it is on 13.1 beta 2 on the mini 4 w/o the delay:

delay mov

Wikipedia/Code/WKWebView+ElementLocation.m
```diff
@@ -54,8 +54,8 @@ - (void)getScreenRectForHtmlElementFromJavascriptString:(NSString *)javascriptSt
            }];
 }
```
montehurd: This is so we're doing apples-to-apples now that https://github.com/wikimedia/wikipedia-ios/pull/3230/commits/b296e400077f0d68f6e9f638ef8b10e0e797717f is using `self.delegate.navigationBar.visibleHeight`

joewalsh:
Closing in favor of #3251, #3252 and any other follow-ons

PR #3233, opened 2019-09-05
Restore Explore content offset

nambatee:
Follow-up for app language change / state restoration.

https://phabricator.wikimedia.org/T229540

- If the app goes into the background in 1 orientation and is restored in another - the layout will be different and the user won't be back in the 'same' spot. In those cases, we could either be okay with that or put users back at the top of the feed, cc @carolynlimadeo
- Since there's only 1 VC that needs its content offset adjusted, I'm using the explicit type instead of a protocol


joewalsh:
> If the app goes into the background in 1 orientation and is restored in another - the layout will be different and the user won't be back in the 'same' spot. In those cases, we could either be okay with that or put users back at the top of the feed

Would it be better to store the first fully visible `WMFContentGroup`s `key` and scroll to that? Would also handle the case where new content is loaded and the offset is now different

nambatee:
Great idea! Testing now

nambatee:
Pushed an update with content groups

PR #3234, opened 2019-09-05
fix editor dismissing upon background in iPadOS

tonisevener:
& fix UITableViewAlertForLayoutOutsideViewHierarchy console warning

https://phabricator.wikimedia.org/T229494

I couldn't repro the freezing anymore but editor would dismiss if you had table of contents open in article > go to editor > background for 5-10 seconds. This is because oddly only in iPadOS willTransitionToTraitCollection in WMFArticleViewController is called twice upon background - the size classes switch to compact then switch back to what they were before. We try presenting table of contents again in this situation assuming the user is split screening which seems to automatically dismiss the editor. Fix is to make sure WMFArticleViewController isn't presenting anything like editor before attempting to present the table of contents.

### DIFF
--- a/Wikipedia/Code/WMFTableOfContentsViewController.swift
+++ b/Wikipedia/Code/WMFTableOfContentsViewController.swift
@@ -224,8 +224,8 @@ open class WMFTableOfContentsViewController: UIViewController, UITableViewDelega
         tableOfContentsFunnel.logOpen()
     }
     
-    open override func viewDidDisappear(_ animated: Bool) {
-        super.viewDidDisappear(animated)
+    open override func viewWillDisappear(_ animated: Bool) {
+        super.viewWillDisappear(animated)
         deselectAllRows()
     }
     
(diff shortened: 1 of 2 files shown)


Wikipedia/Code/WMFTableOfContentsViewController.swift
```diff
@@ -224,8 +224,8 @@ open class WMFTableOfContentsViewController: UIViewController, UITableViewDelega
         tableOfContentsFunnel.logOpen()
     }
     
-    open override func viewDidDisappear(_ animated: Bool) {
-        super.viewDidDisappear(animated)
+    open override func viewWillDisappear(_ animated: Bool) {
+        super.viewWillDisappear(animated)
         deselectAllRows()
```
nambatee: Is it necessary to deselect all rows at all? If I background the app and come back, I'm scrolled to the right section but the selected section in the TOC is off

montehurd: hmm.. wonder if my tweak to [TOC fragment logic](https://github.com/wikimedia/wikipedia-ios/pull/3230) affects this? will check

montehurd: i may know a fix for the TOC issue @nambatee noticed... checking...

tonisevener: Sounds good, this was just to get rid of that console error I was seeing, didn't affect the editor presentation dismissal upon backgrounding.

montehurd: TOC selection reproducible on 6.3. Filing separate ticket with repro steps.

PR #3239, opened 2019-09-06
Fix image dimming regression

nambatee:
Image dimming is enabled only for the dark themes


nambatee:
closing since this is intentional now

PR #3250, opened 2019-09-11
Fix misplaced alert messages

joewalsh:
Switch presentation style to over full screen for anything presented in a themeable navigation controller

https://phabricator.wikimedia.org/T232604


joewalsh:
Hold pending discussion on https://phabricator.wikimedia.org/T232604#5483193

PR #3253, opened 2019-09-11
Ensure scroll stops when scrolling to a new section

joewalsh:
Helps with fixing https://phabricator.wikimedia.org/T232141


joewalsh:
This doesn't seem to fully fix the issue, and it's getting to be higher impact than is worth introducing at this point in 6.4. This should be revisited in 6.5 since it doesn't appear to be a regression or a new issue with iOS 13.

PR #3256, opened 2019-09-11
Fix blank content flash on web content scroll

joewalsh:
https://phabricator.wikimedia.org/T232689

This was broken prior to the recent changes to `wmf_safeSetContentOffset` so it's not a regression.

before (6.3.2): https://vimeo.com/359421009

after: https://vimeo.com/359420997


Wikipedia/Code/WebViewController.m
```diff
@@ -514,6 +518,41 @@ - (void)moveFindInPageSelectedMatchIndexInDirection:(WMFFindInPageScrollDirectio
...
+    Setting scrollView.contentOffset inside of an animation block
+    results in a broken animation https://phabricator.wikimedia.org/T232689
+    Calling [scrollView setContentOffset:offset animated:YES] inside
+    of an animation block fixes the animation but doesn't guarantee
+    the content offset will be updated when the animation's completion
+    block is called.
+    It appears the only reliable way to get a callback after the default
+    animation is to use scrollViewDidEndScrollingAnimation
+    */
+    [self.scrollViewAnimationCompletions addObject:completion];
```
tonisevener: I think we should pass done here instead - it'll crash if completion is nil.

Wikipedia/Code/WebViewController.m
```diff
@@ -539,11 +578,11 @@ - (void)scrollToAndFocusOnSelectedMatch {
...
                                                  CGPoint centeredOffset = CGPointMake(self.webView.scrollView.contentOffset.x, newOffsetY);
-                                                 [self.webView.scrollView wmf_safeSetContentOffset:centeredOffset
-                                                                                          animated:YES
-                                                                                        completion:^(BOOL done) {
-                                                                                            self.disableMinimizeFindInPage = NO;
-                                                                                        }];
+                                                 [self scrollToOffset:centeredOffset
+                                                             animated:YES
+                                                           completion:^{
+                                                               self.disableMinimizeFindInPage = NO;
```
tonisevener: Not sure why but there's now a find in page bug - it automatically dismisses after typing a couple of characters.

Wikipedia/Code/WebViewController.m
```diff
@@ -967,11 +1006,11 @@ - (void)scrollTappedReferenceUpIfNecessaryWithReferencePageViewController:(WMFRe
...
             CGPoint centeredOffset = CGPointMake(self.webView.scrollView.contentOffset.x, newOffsetY);
-            [self.webView.scrollView wmf_safeSetContentOffset:centeredOffset
-                                                     animated:YES
-                                                   completion:^(BOOL finished) {
-                                                       controller.backgroundView.clearRect = CGRectOffset(windowCoordsRefGroupRect, 0, delta);
-                                                   }];
+            [self scrollToOffset:centeredOffset
+                        animated:YES
+                      completion:^{
+                          controller.backgroundView.clearRect = CGRectOffset(windowCoordsRefGroupRect, 0, delta);
```
tonisevener: I can repro this in develop as well - the clearRect is off on iPhone (tested in iOS13). I can file a separate bug if necessary.

IMG_0012

Wikipedia/Code/WebViewController.m
joewalsh: good catch! fixed

Wikipedia/Code/WebViewController.m
joewalsh: If that repros on develop it'd be better to file separately. Thanks!

Wikipedia/Code/WebViewController.m
joewalsh: Another good catch - looks like the issue is caused by the completions stacking up and all being called on the first animation callback. This should wait until the last completion is called before setting `self.disableMinimizeFindInPage = NO;` Will fix

joewalsh: It looks like `disableMinimizeFindInPage` was there to filter out non-user scroll events from `scrollViewDidScroll` instead of using `scrollViewWillBeginDragging`. Neither of those are necessary when using the built in method on `UIScrollView`. Fixed

joewalsh:
Removing hold as this can be pulled into develop for 6.5 or 6.4.1 (if required)

PR #3263, opened 2019-09-13
Open Insert Media image urls in Safari

nambatee:


Wikipedia/Code/InsertMediaSearchResultsCollectionViewController.swift
```diff
@@ -276,7 +276,7 @@ extension InsertMediaSearchResultsCollectionViewController: UIViewControllerPrev
             self.delegate?.insertMediaSearchResultsCollectionViewControllerDidSelect(self, searchResult: searchResult)
         }
         previewingViewController.moreInformationAction = { url in
-            self.present(SFSafariViewController(url: url), animated: true)
+            self.wmf_openExternalUrl(url, useSafari: true)
```
joewalsh: I think these should be `useSafari: false` if in the future we'd like these to open in the app instead of Safari if that option becomes available again. Alternatively we could just remove the `useSafari` param for now because it has no effect

joewalsh: Going to merge this to roll a build - we can clean up the interface as a follow-on

PR #3266, opened 2019-09-13
Workaround for iOS 13 constraint breakage

montehurd:
https://phabricator.wikimedia.org/T232073

### DIFF
--- a/WMF Framework/NavigationBar.swift
+++ b/WMF Framework/NavigationBar.swift
@@ -18,11 +18,15 @@ public class NavigationBar: SetupView, FakeProgressReceiving, FakeProgressDelega
     public var underBarViewPercentHiddenForShowingTitle: CGFloat?
     public var title: String?
 
-    public init() {
-        super.init(frame: CGRect(x: 0, y: 0, width: 320, height: 44))
+    convenience public init() {
+        self.init(frame: CGRect(x: 0, y: 0, width: 320, height: 44))
         titleBar.frame = bounds
     }
 
+    private override init(frame: CGRect) {
+        super.init(frame: frame)
+    }
+    
     required init?(coder aDecoder: NSCoder) {
         fatalError("init(coder:) has not been implemented")
     }
--- a/WMF Framework/NavigationBar.swift
+++ b/WMF Framework/NavigationBar.swift
@@ -24,6 +24,7 @@ public class NavigationBar: SetupView, FakeProgressReceiving, FakeProgressDelega
 
     override init(frame: CGRect) {
         super.init(frame: frame)
+        assert(frame != .zero, "Non-zero frame required to prevent iOS 13 constraint breakage")
         titleBar.frame = bounds
     }
     
--- a/WMF Framework/NavigationBar.swift
+++ b/WMF Framework/NavigationBar.swift
@@ -8,7 +8,7 @@
 @objc(WMFNavigationBar)
 public class NavigationBar: SetupView, FakeProgressReceiving, FakeProgressDelegate {
     fileprivate let statusBarUnderlay: UIView =  UIView()
-    fileprivate let titleBar: UIToolbar = UIToolbar(frame: .infinite) // HAX: `.infinite` is workaround for iOS 13 constraint breakage
+    fileprivate let titleBar: UIToolbar = UIToolbar()
     fileprivate let bar: UINavigationBar = UINavigationBar()
     fileprivate let underBarView: UIView = UIView() // this is always visible below the navigation bar
     fileprivate let extendedView: UIView = UIView()
@@ -17,7 +17,17 @@ public class NavigationBar: SetupView, FakeProgressReceiving, FakeProgressDelega
     fileprivate let backgroundView: UIView = UIView()
     public var underBarViewPercentHiddenForShowingTitle: CGFloat?
     public var title: String?
-    
+
+    public override init(frame: CGRect) {
+        super.init(frame: frame)
+        assert(frame != .zero, "Frame should be non-zero")
+        titleBar.frame = bounds
+    }
+
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
     public var isAdjustingHidingFromContentInsetChangesEnabled: Bool = true
     
     public var isShadowHidingEnabled: Bool = false // turn on/off shadow alpha adjusment
--- a/WMF Framework/NavigationBar.swift
+++ b/WMF Framework/NavigationBar.swift
@@ -8,7 +8,7 @@
 @objc(WMFNavigationBar)
 public class NavigationBar: SetupView, FakeProgressReceiving, FakeProgressDelegate {
     fileprivate let statusBarUnderlay: UIView =  UIView()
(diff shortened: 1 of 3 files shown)


montehurd:
I don't like it, but it fixes the breakage and I think lowering the priority should be fairly low-impact...

WMF Framework/NavigationBar.swift
```diff
@@ -125,6 +125,17 @@ public class NavigationBar: SetupView, FakeProgressReceiving, FakeProgressDelega
...
+        
+        
+        // HAX: workaround for iOS 13 constraint breakage warnings
+        if #available(iOS 13, *) {
+            guard let contentView = titleBar.subviews.first, String(describing: type(of: contentView)) == "_UIToolbarContentView" else {
+                return
+            }
+            contentView.constraints
+                .filter { $0.priority == .required && $0.firstAttribute == .leading || $0.firstAttribute == .trailing }
+                .forEach { $0.priority = .defaultHigh }
```
joewalsh: The end result at runtime is likely the same since the warning is about it breaking the constraint. This just makes that constraint breakage OK by lowering the priority. Given that it's only silencing the warning, I don't think this is worth it. It's only really worth fixing if there's a way to alter the setup we can control without having to dig into the internals of `UIToolbar`

montehurd: The constraints don't break with this because of the way the priorities are compared, but I hear ya. It's not a good fix.

joewalsh: I'm defining a constraint 'break' as a constraint not being fully satisfied - the priorities are there to determine which constraint that can stay unsatisfied (or 'broken') when there's a conflict

montehurd: Ah, I see. I was thinking more literally. One of the things I like about SwiftUI is, iirc, it won't produce output with any unsatisfiable constraints, so in some distant future we won't be using  non-`.required` priorities (or eventually any😜 manual constraints) all over the place, as we do now.

montehurd: I for one welcome our reactive declarative one-way data flow overlords 😂

montehurd:
Noticed this works too:

```
if #available(iOS 13, *) {
    guard let contentView = titleBar.subviews.first, String(describing: type(of: contentView)) == "_UIToolbarContentView", let sv = contentView.superview else {
        return
    }
    contentView.translatesAutoresizingMaskIntoConstraints = false
    sv.wmf_addConstraintsToEdgesOfView(contentView)
}
```

... still gross 🤮

montehurd:
This works too:

```
if #available(iOS 13, *) {
    guard let contentView = titleBar.subviews.first, String(describing: type(of: contentView)) == "_UIToolbarContentView" else {
        return
    }
    contentView.frame = CGRect.init(x: 0, y: 0, width: Int.max, height: Int.max)
}
```

... but still using the nested container view. Tried giving the superviews a non-zero size but no luck so far - may have missed one... re-checking...

WMF Framework/NavigationBar.swift
```diff
@@ -8,7 +8,7 @@
 @objc(WMFNavigationBar)
 public class NavigationBar: SetupView, FakeProgressReceiving, FakeProgressDelegate {
     fileprivate let statusBarUnderlay: UIView =  UIView()
-    fileprivate let titleBar: UIToolbar = UIToolbar()
+    fileprivate let titleBar: UIToolbar = UIToolbar(frame: .infinite) // HAX: `.infinite` is workaround for iOS 13 constraint breakage
```
joewalsh: Setting this to a value that isn't valid for view frames like `CGRect.infinite` could lead to other unexpected behavior down the line. A fix that'd be more in line with expected behavior for views would be to have `NavigationBar` set all the subviews to the initial bounds in `init(frame:` and provide an arbitrary, valid default frame for `NavigationBar` like `CGRect(x: 0, y: 0, width: 100, height: 44)`

montehurd: Good point. `CGRect.zero` was causing problems so it's probably safer to assume `CGRect.infinite` could have issues too.

Wikipedia/Code/WMFViewController.m
```diff
@@ -25,7 +25,7 @@ @implementation WMFViewController
 
 - (void)setup {
     self.theme = [WMFTheme standard];
-    self.navigationBar = [[WMFNavigationBar alloc] initWithFrame:CGRectZero];
+    self.navigationBar = [[WMFNavigationBar alloc] initWithFrame:CGRectMake(0, 0, 320, 44)];
```
joewalsh: Rather than have to manually set an arbitrary frame in two places, it might be simpler to have the standard `init` call `initWithFrame` with the default value

montehurd: Ah, can do!

montehurd: @joewalsh done!

WMF Framework/NavigationBar.swift
```diff
@@ -17,7 +17,16 @@ public class NavigationBar: SetupView, FakeProgressReceiving, FakeProgressDelega
     fileprivate let backgroundView: UIView = UIView()
     public var underBarViewPercentHiddenForShowingTitle: CGFloat?
     public var title: String?
-    
+
+    public init() {
+        super.init(frame: CGRect(x: 0, y: 0, width: 320, height: 44))
+        titleBar.frame = bounds
```
joewalsh: This won't be set if `init(frame:` is called on `NavigationBar`. Since `init(frame:` is the designated initializer for `UIView`, `init` should be a convenience `init` that calls `self.init(frame:` with the default value

montehurd: haha was just updating that 😂

WMF Framework/NavigationBar.swift
```diff
@@ -18,11 +18,15 @@ public class NavigationBar: SetupView, FakeProgressReceiving, FakeProgressDelega
...
     public var title: String?
 
-    public init() {
-        super.init(frame: CGRect(x: 0, y: 0, width: 320, height: 44))
+    convenience public init() {
+        self.init(frame: CGRect(x: 0, y: 0, width: 320, height: 44))
         titleBar.frame = bounds
     }
 
+    private override init(frame: CGRect) {
```
montehurd: I don't recall overriding an `init` like this just to make it private... are there any potential issues with this?

joewalsh: It goes against the idea of a designated initializer and the documentation for UIView. `It is recommended that you implement this method. You can also implement custom initialization methods in addition to, or instead of, this method.` - in this setup, that means our `init` is a convenience init and `init(frame:` is still the designated initializer and should be public. All necessary setup should be done in `init(frame:`

montehurd: 🤦‍♂️

montehurd: (☕️ is helping 😂)

montehurd: f.f.r. i had some fog on swift convenience initializer specifics - these helped:
https://stackoverflow.com/a/36612507
https://stackoverflow.com/a/32108404

PR #3272, opened 2019-09-19
Fix TOC highlighting wrong section sometimes

montehurd:
https://phabricator.wikimedia.org/T232141

Extracted from #3230

-----

To debug set `WMFArticleViewControllerTableOfContentsSectionUpdateScrollDistance` to `1` and add this to `misc.less` and run `make grunt`:

```
div[id^="section_heading_and_content_block_"] {
  border: 1px solid red !important;
}
```

^ makes it easier to test that the selected TOC item changes (when you scroll the article on iPad when TOC is visible) exactly when the top of the section wrapper intersects the top of the screen. Be sure to test for case where nav bar is visible and hidden!

-----

Also fixes TOC highlight after tapping `geography section` link on `testwiki > bird` (bug would sometimes not manifest on the first tap, but only after popping back to `bird` article and tapping the link again). This [delay](https://github.com/wikimedia/wikipedia-ios/compare/T232141?expand=1#diff-38568396230724f82d0bf5ec47cec4d8R810) was (still! 😱) the only fix I could find for this.

-----

Trivia (private API's so can't use):

`iOS12yOffsetHack` can be deleted if you add this to a Swift `WKWebView` extension and call it from the web VC's `viewDidAppear`:

```
setValue(scrollView.contentInset, forKey: "_obscuredInsets")
setValue(true, forKey: "_haveSetObscuredInsets")
```

[source](https://stackoverflow.com/a/56387968/135557)


Wikipedia/Code/WebViewController.m
```diff
@@ -825,9 +828,19 @@ - (void)accessibilityCursorToSection:(MWKSection *)section {
     [self.webView wmf_accessibilityCursorToFragment:section.anchor];
 }
 
+- (CGFloat) insetTop {
+    // HAX: something odd going on with iOS 11...
+    CGFloat insetTop = self.delegate.navigationBar.subviews.firstObject.frame.origin.y;
```
montehurd: Better way to do this? Could maybe add another var to go with `visibleHeight`?

Wikipedia/Code/WebViewController.h
```diff
@@ -90,6 +90,7 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)webViewController:(WebViewController *)controller didTapFooterReadMoreSaveForLaterForArticleURL:(NSURL *)url didSave:(BOOL)didSave;
 - (void)webViewController:(WebViewController *)controller didTapAddTitleDescriptionForArticle:(MWKArticle *)article;
 - (void)webViewController:(WebViewController *)controller didScrollToSection:(MWKSection *)section;
+- (void)webViewController:(WebViewController *)controller didScrollToFragment:(NSString *)fragment;
```
montehurd: Likely safer not to assume every fragment corresponds to a section.

Wikipedia/Code/WMFArticleViewController.m
```diff
@@ -1586,6 +1586,11 @@ - (void)webViewController:(WebViewController *)controller didScrollToSection:(MW
     self.visibleSectionAnchor = section.anchor;
 }
 
+- (void)webViewController:(WebViewController *)controller didScrollToFragment:(NSString *)fragment {
+    self.visibleSectionAnchor = fragment;
+    [self updateTOCHighlightIfNecessaryWithScrollView:controller.webView.scrollView force:true];
```
joewalsh: Any reason for not selecting the correct section based on the fragment directly? ```
     if (!fragment) {
        return;
    }
    MWKSection *section = [self.article.sections sectionWithFragment:fragment];
    [self selectAndScrollToTableOfContentsItemForSection:section animated:NO]; 
``` with this change and the delay removed, it seems to fix the issue on click through and on the color article.

Comment (joewalsh) [reply Wikipedia/Code/WMFArticleViewController.m]: I pushed this change because I saw this PR is tagged with changes welcome but feel free to revert or modify if there was something I missed.

File: Wikipedia/Code/WMFArticleViewController.m [hunk 4]
```diff
@@ -189,7 +189,7 @@ - (instancetype)initWithArticleURL:(NSURL *)url
     if (self) {
         NSString *fragment = [url fragment];
         if (![[fragment stringByTrimmingCharactersInSet:[NSCharacterSet whitespaceAndNewlineCharacterSet]] isEqualToString:@""]) {
-            self.initialFragment = [fragment stringByRemovingPercentEncoding];
+            self.initialAnchor = [fragment stringByRemovingPercentEncoding];
```
joewalsh: `fragment` makes sense in the context of a URL, for our use case I renamed everything to anchor to better correspond with [wiki naming conventions](https://meta.wikimedia.org/wiki/Help:Link#Anchors) and `MWKSection`.

Wikipedia/Code/WMFArticleViewController.m
montehurd: Not all fragments are sections.

montehurd: At least, it seemed safer not to assume so...

joewalsh: If that's the case, it might make sense to fall back on the old method if the section returned is `nil`

montehurd:
Now getting an exception in `selectAndScrollToItem` when I tap the `geography section` link on `testwiki > bird`. Confirmed it's a new issue with the latest commits... looking into it...

montehurd:
Seeing the exception on iOS 13 Xs Max simulator.

Wikipedia/Code/WMFArticleViewController.m
```diff
@@ -1534,6 +1534,20 @@ - (void)showReadingThemesControlsPopup {
     [self.readingThemesControlsPresenter objCShowReadingThemesControlsPopupOn:self theme:self.theme];
 }
 
+#pragma mark - Scrolling to anchors
+
+- (void)scrollToAnchor:(nullable NSString *)anchor animated:(BOOL)animated completion:(nullable dispatch_block_t)completion {
+    self.updateTableOfContentsSectionOnScrollEnabled = NO;
+    [self.webViewController scrollToAnchor:anchor
```
montehurd: @joewalsh It seems the problem is related to this... if the `scrollToAnchor` call here is commented out the exception goes away but the article doesn't load at all (when tapping the `geography section` link on `testwiki > bird`). So I guess it makes sense we're seeing the exception since it's trying to scroll when the article isn't loaded?

montehurd: Opps it's completion block calls `showWebView` so that's why it wasn't loading... hmm...

montehurd: f.t.r i was typing too fast above - i meant the cause of the assertion not `exception` 🤦‍♂️

montehurd:
@tonisevener thanks! that did it 👏

aehlke:
Reportedly Apple does not reject for `_obscuredInsets` usage

PR #3274, opened 2019-09-20
Revert "Add `ku` strings"

joewalsh:
Reverts wikimedia/wikipedia-ios#3273

Two separate build machines for me running 10.14.6 and Xcode 11 GM seed 2 don't add these files @montehurd what config were you using that was adding them? It could be related to the macOS beta.


montehurd:
Using the `Wikipedia` build scheme. On Catalina beta... 🤔

nambatee:
I get these too (Catalina beta)

joewalsh:
Apple likely added `ku-latn` to the available languages for Catalina so the localization script began treating them as valid via:

```
let locales: Set<String> =  {
    var identifiers = Locale.availableIdentifiers
    if let filenames = try? FileManager.default.contentsOfDirectory(atPath: "\(path)/Wikipedia/iOS Native Localizations") {
        let additional = filenames.compactMap { $0.components(separatedBy: ".").first?.lowercased() }
        identifiers += additional
    }
    return Set<String>(identifiers)
}()

func localeIsAvailable(_ locale: String) -> Bool {
    let prefix = locale.components(separatedBy: "-").first ?? locale
    return locales.contains(prefix)
}
```
The script is assuming macOS's `Locale.availableIdentifiers` matches iOS's, not sure if that's true. Since Catalina is still in beta and the build machine is still on Mojave, let's move forward with this PR and add `ku-latn` after Catalina goes GM.

PR #3275, opened 2019-09-20
Fix iOS 13 bugs in Link & Image wizards

nambatee:
- reload data in list views when theme is applied
- make sure the Link wizard is presented modally to match the Media wizard

### DIFF
--- a/Wikipedia/Code/SectionEditorViewController.swift
+++ b/Wikipedia/Code/SectionEditorViewController.swift
@@ -668,12 +668,18 @@ extension SectionEditorViewController: SectionEditorInputViewsControllerDelegate
                 }
                 editLinkViewController.delegate = self
                 let navigationController = WMFThemeableNavigationController(rootViewController: editLinkViewController, theme: self.theme)
+                if #available(iOS 13.0, *) {
+                    navigationController.modalPresentationStyle = .automatic
+                }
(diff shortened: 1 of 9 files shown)


Wikipedia/Code/SectionEditorViewController.swift
```diff
@@ -668,12 +668,18 @@ extension SectionEditorViewController: SectionEditorInputViewsControllerDelegate
...
+                    navigationController.modalPresentationStyle = .automatic
+                }
                 navigationController.isNavigationBarHidden = true
                 self.present(navigationController, animated: true)
             } else {
                 let insertLinkViewController = InsertLinkViewController(link: link, siteURL: siteURL, dataStore: dataStore)
                 insertLinkViewController.delegate = self
                 let navigationController = WMFThemeableNavigationController(rootViewController: insertLinkViewController, theme: self.theme)
+                if #available(iOS 13.0, *) {
+                    navigationController.modalPresentationStyle = .automatic
```
joewalsh: Is there a chance of either of these being affected by https://phabricator.wikimedia.org/T232604 ?

nambatee: Yes, articles syncing or losing connection could cause the alerts to show up

PR #3285, opened 2019-10-02
Fix iOS12 bug that can sometimes cause dark article but rest of app light

montehurd:
Found while working on: https://phabricator.wikimedia.org/T234255

theme 480 mov


Wikipedia/Code/NSUserDefaults+WMFExtensions.swift
```diff
@@ -144,7 +144,7 @@ let WMFSearchLanguageKey = "WMFSearchLanguageKey"
         let name = string(forKey: WMFAppThemeName)
         guard name != nil, name != Theme.defaultThemeName else {
             if #available(iOSApplicationExtension 12.0, *) {
-                return traitCollection.userInterfaceStyle == .light ? .light : Theme.black.withDimmingEnabled(wmf_isImageDimmingEnabled)
+                return traitCollection.userInterfaceStyle == .dark ? Theme.black.withDimmingEnabled(wmf_isImageDimmingEnabled) : .light
```
joewalsh: Is the issue that we're checking the trait collection too early so `userInterfaceStyle` is `.undefined` or is it always undefined in some cases on iOS 12?

montehurd: Once it gets into this state it can continue to happen. That is, if I do pull-to-refresh, which fixes it, then quit the app, then restart it, it happens again...

joewalsh: Right, but is the reason it's getting into that state because we're reading `userInterfaceStyle` too early? If that's the case it's a wider problem similar to #3284 - view controllers would need to wait until the first `viewWillLayoutSubviews` to apply the correct theme based on `userInterfaceStyle`. That was Natalia's interpretation of the documentation in #3229 and I had assumed otherwise but it's looking like I was incorrect. This should probably default to `.light` either way, but the root cause might be informative for [https://phabricator.wikimedia.org/T234255](T234255).

montehurd: This is interesting! On develop w/iOS 12 there's a single frame (or so) on app launch which is dark - seems like this is related to @joewalsh's comment...

dark mov

montehurd: ^ this was on clean install

nambatee: On launch `traitCollection` is `.unspecified` and right now (not in your PR) we default to showing the black theme. Then we get another `traitCollectionDidChange` call with the correct theme and that's how we get the correct theme. iOS seems to be constantly taking snapshots to be able to update the app previews when the theme changes, that's why we get those silent trait change calls in the bg

montehurd: Also, as long as you don't dismiss the welcome screens it happens every time.

montehurd: Looks like b2dde0e fixes it too, though as @joewalsh mentioned accessing `userInterfaceStyle` too early may be the actual cause. Thankfully this is a bit easier to repro than the article loading steps I outline in the PR description gif.

montehurd: @nambatee those 2 replies above were made before I saw your reply - they're followups to the gif I posted 😂

montehurd: Regarding ...
> On launch traitCollection is .unspecified and right now (not in your PR) we default to showing the black theme. 

... pre iOS 13 is there a case where we'd want `.unspecified` to map to dark?

montehurd: I haven't noticed any unexpected side-effects of mapping `.unspecified` to light so far... was assuming that once subsequent `traitCollectionDidChange` calls happen the theme would be set correctly - so far haven't noticed any issues with my change aside from fixing those 2 issues from the gifs. What I haven't seen yet is quite how this issue could cause the sepia/dark combo Andrea reported...

nambatee: Is there a reason why `#available` is set to `12.0` and not `13.0`? If it was, we could avoid checking `traitCollection` pre 13.0 altogether

montehurd: Hmm good question 🤔 I was also about to just try building previous version, setting theme to sepia, then trying my orig repro steps, but I unfortunately have to run. Adding `changes welcome` in case you solve it before I return...

joewalsh: > Is there a reason why #available is set to 12.0 and not 13.0? If it was, we could avoid checking traitCollection pre 13.0 altogether
Good point - `userInterfaceStyle` became available in 12.0 but we don't need it there, this should switch to 13.0

I verified in Xcode 11.2 beta that on 13.2 the trait collection is correct at the time of `viewDidLoad` so the switch to `viewDidLayoutSubviews` doesn't seem warranted if we switch the availability check

PR #3294, opened 2019-10-09
Fix some incorrectly positioned reference highlights

montehurd:
https://phabricator.wikimedia.org/T233472

Fixes this issue seen on iOS 12 & up:

highlight mov


montehurd:
Found way to simplify this a bit... pushing commit after re-testing...

montehurd:
Tested fix on iOS 11, 12 & 13 on various phone and tablet simulators.

PR #3295, opened 2019-10-12
Page history refresh

nambatee:
https://phabricator.wikimedia.org/T228783

- [x] Fix spacing between sections
- [ ]  Add empty view for history
- [x] Localize strings
- [x] RTL
- [x]  Fix UIToolbar RTL
- [x] Accessibility
- [x] iPad


Wikipedia/Code/PageHistoryFilterCountsViewController.swift
```diff
@@ -0,0 +1,130 @@
+import UIKit
+
+protocol PageHistoryFilterCountsViewControllerDelegate: AnyObject {
+    func pageHistoryFilterCountsViewControllerDidDetermineIfFilterCountsAreAvailable(areFilterCountsAvailable: Bool)
```
tonisevener: Probably could shrink this name down, IMO I don't think the full view controller type name is needed since it's in the protocol name. Maybe something like 
`func didDetermineFilterCountsAvailability(_ available: Bool, viewController: PageHistoryFilterCountsViewController)`

Wikipedia/Code/PageHistoryViewController.swift
```diff
@@ -0,0 +1,547 @@
...
+                }
+                resetComparisonSelectionButtons()
+                navigationController?.setToolbarHidden(true, animated: true)
+            case .editing:
+                navigationItem.rightBarButtonItem = cancelComparisonButton
+                collectionView.allowsMultipleSelection = true
+                forEachVisibleCell { $1.setEditing(true) }
+                compareToolbarButton.isEnabled = false
+                NSLayoutConstraint.activate([
+                    firstComparisonSelectionButton.widthAnchor.constraint(equalToConstant: 90),
```
tonisevener: I haven't confirmed but might want to make sure you remove these constraints when in .idle so you don't have multiple width constraints added.

Wikipedia/Code/PageHistoryViewController.swift
```diff
@@ -0,0 +1,547 @@
...
+                ])
+                setToolbarItems([UIBarButtonItem(customView: firstComparisonSelectionButton), UIBarButtonItem.wmf_barButtonItem(ofFixedWidth: 10), UIBarButtonItem(customView: secondComparisonSelectionButton), UIBarButtonItem(barButtonSystemItem: .flexibleSpace, target: nil, action: nil), UIBarButtonItem(barButtonSystemItem: .flexibleSpace, target: nil, action: nil),  compareToolbarButton], animated: true)
+                navigationController?.setToolbarHidden(false, animated: true)
+            }
+            collectionView.collectionViewLayout.invalidateLayout()
+            navigationItem.rightBarButtonItem?.tintColor = theme.colors.link
+        }
+    }
+
+    private lazy var compareToolbarButton = UIBarButtonItem(title: "Compare", style: .plain, target: self, action: #selector(showDiff(_:)))
```
tonisevener: "Compare" needs to be localized

Wikipedia/Code/HintViewController.swift
```diff
@@ -114,9 +117,11 @@ extension HintViewController: Themeable {
         guard viewIfLoaded != nil else {
             return
         }
-        view.backgroundColor = theme.colors.hintBackground
-        defaultLabel?.textColor = theme.colors.link
-        confirmationLabel?.textColor = theme.colors.link
-        confirmationAccessoryButton.tintColor = theme.colors.link
+
+        view.backgroundColor = backgroundColor ?? theme.colors.hintBackground
```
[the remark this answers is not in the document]
nambatee: I use it when the same view can have different colors coming from `Theme` - for example, in `WMFEmptyView`, one of the colors can be swapped for a different `Theme` color: https://github.com/wikimedia/wikipedia-ios/blob/54ea08c2aaf85c39e383a6532cf2a046a5ef3347/Wikipedia/Code/WMFEmptyView.m#L261

Here it wouldn't work because the colors don't have to be `Theme` colors - for example, the background color set by `PageHIstoryHintController` is `UIColor.wmf_lightRed`. If it changes, it would make sense to use key paths

Wikipedia/Code/PageHistoryCollectionViewCell.swift
```diff
@@ -0,0 +1,299 @@
...
+    private let minorImageView = UIImageView()
+    private let commentLabel = UILabel()
+    private let authorButton = AlignedImageButton()
+    private let selectView = BatchEditSelectView()
+    private let spacing: CGFloat = 3
+    private var theme = Theme.standard
+
+    var time: String?
+
+    var displayTime: String? {
```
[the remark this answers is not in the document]
nambatee: It's used in that its `didSet` sets the text of `timeLabel` and requests a layout update

Wikipedia/Code/PageHistoryFilterCountsViewController.swift
nambatee: That could work. I always refer to https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/CodingGuidelines/Articles/NamingMethods.html#//apple_ref/doc/uid/20001282-1001803-BCIDAIJE for naming (_Start the name by identifying the class of the object that’s sending the message:_) and in the past I found it confusing when delegate methods didn't include the sender in the name

Wikipedia/Code/PageHistoryViewController.swift
nambatee: definitely 👍

Wikipedia/Code/PageHistoryCollectionViewCell.swift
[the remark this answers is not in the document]
nambatee: just noticed I was setting `time` instead of `displayTime` here - good catch! should be fixed now

Wikipedia/Code/PageHistoryFilterCountCollectionViewCell.swift
```diff
@@ -0,0 +1,57 @@
...
+class PageHistoryFilterCountCollectionViewCell: UICollectionViewCell {
+    @IBOutlet private weak var titleLabel: UILabel!
+    @IBOutlet private weak var imageView: UIImageView!
+    @IBOutlet private weak var imageLabel: UILabel!
+
+    @IBOutlet private weak var rightSeparator: UIView!
+
+    override func awakeFromNib() {
+        super.awakeFromNib()
+        updateFonts()
```
joewalsh: Due to the [traitCollection changes in iOS 13 ](https://developer.apple.com/documentation/ios_ipados_release_notes/ios_13_release_notes) this initial `updateFonts` might be off. I wouldn't worry about it if we're not seeing any issues though

WMF Framework/ColumnarCollectionViewLayout.swift
```diff
@@ -165,6 +165,12 @@ public class ColumnarCollectionViewLayout: UICollectionViewLayout {
         context.boundsDidChange = true
         return context
     }
+
+    public func invalidationContextForDataChange() -> UICollectionViewLayoutInvalidationContext {
```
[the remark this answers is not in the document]
nambatee: yes, thanks!

Wikipedia/Code/DiffHeaderSummaryView.swift
```diff
@@ -0,0 +1,72 @@
...
+    }
+    
+    
+
+    func update(_ viewModel: DiffHeaderEditSummaryViewModel) {
+
+        headingLabel.text = viewModel.heading
+        
+        if viewModel.isMinor {
+            tagLabel.text =  "m"
```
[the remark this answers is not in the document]
nambatee: good question, I remember we talked about it with @carolynlimadeo and I think that she confirmed all the minor edits would be marked with "m", irrespective of localization, cc @tonisevener @carolynlimadeo

Wikipedia/Code/PageHistoryComparisonSelectionViewController.swift
```diff
@@ -0,0 +1,113 @@
+import UIKit
+
+protocol PageHistoryComparisonSelectionViewControllerDelegate: AnyObject {
+    func pageHistoryComparisonSelectionViewController(_ pageHistoryComparisonSelectionViewController: PageHistoryComparisonSelectionViewController, didTapSelectionButton button: UIButton)
+    func pageHistoryComparisonSelectionViewController(_ pageHistoryComparisonSelectionViewController: PageHistoryComparisonSelectionViewController, didTapCompareButton button: UIButton)
```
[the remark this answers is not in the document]
nambatee: Makes sense, I agree with the enum approach. I'm not sure about the encapsulation part - the platform delegates pass subviews when it makes sense, for example in https://developer.apple.com/documentation/uikit/uitableviewdelegate/1614883-tableview

Wikipedia/Code/PageHistoryFilterCountsViewController.swift
```diff
@@ -0,0 +1,150 @@
...
+            counts = []
+            defer {
+                collectionView.reloadData()
+                activityIndicator.stopAnimating()
+            }
+            guard let editCounts = editCountsGroupedByType else {
+                return
+            }
+            if case let userEdits?? = editCounts[.userEdits] {
+                counts.append(Count(title: WMFLocalizedString("page-history-user-edits", value: "user edits", comment: "Text for view that shows many edits were made by logged-in users"), image: UIImage(named: "user-edit")!, count: userEdits))
```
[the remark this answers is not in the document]
nambatee: updated in dc2b500e80d922f30eeaf19c2fd63bb8ad814a7d, although to me, it's the same as an implicitly unwrapped `IBOutlet`

Wikipedia/Code/PageHistoryViewController.swift
```diff
@@ -0,0 +1,755 @@
...
+
+        pageHistoryFetcher.fetchPageCreationDate(for: pageTitle, pageURL: pageURL) { result in
+            switch result {
+            case .failure(let error):
+                self.showNoInternetConnectionAlertOrOtherWarning(from: error)
+            case .success(let firstEditDate):
+                self.pageHistoryFetcher.fetchEditCounts(.edits, for: self.pageTitle, pageURL: self.pageURL) { result in
+                    switch result {
+                    case .failure(let error):
+                        self.showNoInternetConnectionAlertOrOtherWarning(from: error)
```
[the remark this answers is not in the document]
nambatee: yes, good catch! I need to rethink failures in fetchEditCounts cause there are multiple requests

Wikipedia/Code/PageHistoryViewController.swift
```diff
@@ -0,0 +1,755 @@
...
+
+        layoutManager.register(PageHistoryCollectionViewCell.self, forCellWithReuseIdentifier: PageHistoryCollectionViewCell.identifier, addPlaceholder: true)
+        collectionView.dataSource = self
+        view.wmf_addSubviewWithConstraintsToEdges(collectionView)
+
+        setupComparisonSelectionViewController()
+
+        apply(theme: theme)
+
+        pageHistoryFetcher.fetchPageCreationDate(for: pageTitle, pageURL: pageURL) { result in
```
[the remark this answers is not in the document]
nambatee: fixed in 9d50436cdf2f739ac7fbdd4d25592f23a8d636f6

Wikipedia/Code/PageHistoryFetcher.swift
```diff
@@ -85,6 +95,144 @@ open class PageHistoryFetcher: WMFLegacyFetcher {
...
+                var allEdits = [NSNumber]()
+                defer {
+                    completion(.success(allEdits))
+                }
+                guard
+                    let items = editMetrics?.items,
+                    let firstItem = items.first,
+                    let results = firstItem.results
+                else {
+                    return
```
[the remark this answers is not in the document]
nambatee: great catch, thanks!

Wikipedia/Code/PageHistoryViewController.swift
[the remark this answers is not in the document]
nambatee: OK, so I return with a failure if I don't get any of the counts and if I managed to catch at least 1 error - https://github.com/wikimedia/wikipedia-ios/pull/3295/commits/0e5ec43501ff1186b31c2d1419600b00d0e7fbb8

In the UI, I'm assuming that I don't want to show an error if I can get at least one count.
In the fetcher, I end up relying on the most recent error, which might not be correct (in theory, calls can fail with different errors). I could return a dictionary of errors grouped by count types or I could split the requests and have the "client" VC match them all together, but even then, I would still want to show only 1 error, and only if I didn't get any counts. 😬

Let me know if you have a different idea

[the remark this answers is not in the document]
tonisevener: You could maybe return a more general error instead of the specific error from that endpoint to avoid the overhead of keeping track of them all - would RequestError.unexpectedResponse work as general enough? Or RequestError.unknown and I think the CommonStrings.genericErrorDescription would get displayed.

Wikipedia/Code/PageHistoryFetcher.swift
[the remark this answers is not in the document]
tonisevener: Would both completion blocks would get called here because of the defer? I tried testing in playgrounds but it's being buggy.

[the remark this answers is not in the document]
nambatee: yeah, defer isn't necessary anymore now that all the paths are covered

Wikipedia/Code/PageHistoryComparisonSelectionViewController.swift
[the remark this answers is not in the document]
nambatee: Pushed an update https://github.com/wikimedia/wikipedia-ios/pull/3295/commits/de10697e64c323e7a75cc12c1d934baaf2d0d998

Wikipedia/Code/DiffContainerViewController.swift
```diff
@@ -0,0 +1,495 @@
...
+    }
+    
+    override func apply(theme: Theme) {
+        
+        guard isViewLoaded else {
+            return
+        }
+        
+        super.apply(theme: theme)
+
```
nambatee: shouldn't super be called first so that `self.theme` can be set regardless of whether the view is loaded?

Wikipedia/Code/DiffContainerViewController.swift
```diff
@@ -0,0 +1,495 @@
...
+            setupScrollingEmptyViewControllerIfNeeded()
+            switch type {
+            case .compare:
+                scrollingEmptyViewController?.type = .diffCompare
+            case .single:
+                scrollingEmptyViewController?.type = .diffSingle
+            }
+            scrollingEmptyViewController?.view.isHidden = false
+            diffListViewController?.view.isHidden = true
+            wmf_hideEmptyView()
```
nambatee: for the `empty` state, shouldn't this trigger showing the empty view via a call to `wmf_showEmptyViewOfType`?

Wikipedia/Code/DiffContainerViewController.swift
```diff
@@ -0,0 +1,495 @@
...
+            setupScrollingEmptyViewControllerIfNeeded()
+            scrollingEmptyViewController?.type = .diffError
+            scrollingEmptyViewController?.view.isHidden = false
+            diffListViewController?.view.isHidden = true
+        case .data:
+            fakeProgressController.stop()
+            scrollingEmptyViewController?.view.isHidden = true
+            diffListViewController?.view.isHidden = false
+            wmf_hideEmptyView()
+            break
```
nambatee: does this `break` have any effect?

Wikipedia/Code/DiffContainerViewController.swift
```diff
@@ -0,0 +1,495 @@
...
+        
+        switch type {
+        case .compare(let articleTitle):
+            
+            let newTitleViewModel = DiffHeaderViewModel.generateTitleViewModelForCompare(articleTitle: articleTitle, counts: counts)
+            headerViewModel.title = newTitleViewModel
+            headerTitleView?.update(newTitleViewModel)
+        case .single:
+            assertionFailure("Should not call this method for the compare type.")
+            return
```
nambatee: I think this `return` might be redundant

Wikipedia/Code/DiffHeaderExtendedView.swift
```diff
@@ -0,0 +1,117 @@
...
+    @IBOutlet var summaryView: DiffHeaderSummaryView!
+    @IBOutlet var editorView: DiffHeaderEditorView!
+    @IBOutlet var compareView: DiffHeaderCompareView!
+    @IBOutlet var divViews: [UIView]!
+    @IBOutlet var editorDivView: UIView!
+    @IBOutlet var compareDivView: UIView!
+    
+    private var viewModel: DiffHeaderViewModel?
+    
+    var delegate: DiffHeaderActionDelegate? {
```
nambatee: needs `weak`

Wikipedia/Code/DiffListViewController.swift
```diff
@@ -0,0 +1,322 @@
...
+            
+            if updateWidthsOnLayoutSubviews {
+                
+                //possibly overkill...simple main thread way to do this is updateListViewModels(listViewModel: self.dataSource, updateType: updateType) followed by the main thread lines 78-83. This does seem very slightly faster for rotation though, and could yield more gains as diffs get larger. More improvements could be size caching & putting layoutSubviewsHeightCalculationsSerialQueue instead into an NSOperation to be cancelled if another viewDidLayoutSubviews is called. No need to continue calculating sizes if collection view frame has changed again.
+                //todo: clean up - move this and updateListViewModel methods into separate class, DiffListSizeCalculator or something
+                let updateType = ListUpdateType.layoutUpdate(collectionViewWidth: self.collectionView.frame.width, traitCollection: self.traitCollection)
+                
+                //actually not sure if this serial queue is needed or simply calling on the main thread (also serial) is the same. this also seems faster than without though.
+                layoutSubviewsHeightCalculationsSerialQueue.async {
```
nambatee: I thought queues with qos set to interactive `userInteractive` would run the tasks on the main thread but that might not be true according to https://stackoverflow.com/a/49658842/4574147

Wikipedia/Code/DiffListViewController.swift
```diff
@@ -0,0 +1,322 @@
...
+        }
+    }
+    
+    override func apply(theme: Theme) {
+        
+        guard isViewLoaded else {
+            return
+        }
+        
+        super.apply(theme: theme)
```
nambatee: https://github.com/wikimedia/wikipedia-ios/pull/3295#discussion_r343917094

Wikipedia/Code/DiffToolbarView.swift
```diff
@@ -0,0 +1,13 @@
+import UIKit
+
+class DiffToolbarView: UIView {
```
nambatee: could this be deleted?

Wikipedia/Code/EmptyViewController.swift
```diff
@@ -0,0 +1,114 @@
...
+}
+
+class EmptyViewController: UIViewController {
+    private let refreshControl = UIRefreshControl()
+    @IBOutlet var scrollView: UIScrollView!
+    @IBOutlet var emptyContainerView: UIView!
+    private var emptyView: WMFEmptyView? = nil
+    var canRefresh: Bool = false
+    weak var delegate: EmptyViewControllerDelegate?
+    var theme: Theme = .light
```
nambatee: could be set to `.standard` instead of `.light` in case `standard` changes

Wikipedia/Code/DiffContainerViewController.swift
tonisevener: good catch, fixed

Wikipedia/Code/DiffContainerViewController.swift
tonisevener: Good catch, I removed these hide calls, it was old code from a different attempt. I'm using EmptyViewController for all of it - I created it for pull to refresh on empty talk pages and now leaning on it for diff. It's headers get so large with dynamic type that the empty states are cut off without scrolling.

Wikipedia/Code/DiffContainerViewController.swift
tonisevener: nope, removed

Wikipedia/Code/DiffContainerViewController.swift
tonisevener: agreed, removed

Wikipedia/Code/DiffHeaderExtendedView.swift
tonisevener: Fixed - since this was a computed property and just passes the assignment into its children I thought maybe it wasn't necessary. But I also thought I would get a compile failure after adding weak but I'm not seeing that now 🤷‍♀️

Wikipedia/Code/DiffToolbarView.swift
tonisevener: Yep, deleted.

nambatee:
LGTM

Wikipedia/Code/DiffListViewController.swift
tonisevener: Yeah, I definitely see the code within both queues running on non-main threads. The main thread checker never fires (this resizes all the heights upon rotation) so I guess it's happy. I figured `userInteractive` fit better here than `userInitiated` since it needs to happen during a rotation, but I haven't measured the difference between the two.

PR #3297, opened 2019-10-16
Explicitly stating the setup script

acagastya:
Adding a note about running the script from the project directory, and not the scripts directory.


montehurd:
Good catch! 👍

PR #3301, opened 2019-10-18
Attempt re-login on 401 response

montehurd:
https://phabricator.wikimedia.org/T229264


Wikipedia/Code/Session.swift
```diff
@@ -221,16 +221,24 @@ import Foundation
...
     */
     private func handleResponse(_ response: URLResponse?) {
         guard let response = response, let httpResponse = response as? HTTPURLResponse else {
             return
         }
         switch httpResponse.statusCode {
         case 401:
-            WMFAuthenticationManager.sharedInstance.logout(initiatedBy: .server) {
-                self.removeAllCookies()
+            WMFAuthenticationManager.sharedInstance.attemptLogin { (loginResult) in
```
joewalsh: If this `attemptLogin` call returns a `401` will it get stuck infinitely attempting to login?

montehurd: that's a good point 😂

montehurd: Though I suppose the previous `logout` call had the same potential issue? But I guess it would be uncommon to have `unauthorized` `logout` response...

montehurd: I pushed a WIP to my fork which passes a flag through to session...
https://github.com/wikimedia/wikipedia-ios/commit/f09d0a5c6b53f7f7aac4933114dc19ebc288c62a
...but it feels all sorts of horrible given how many intermediate methods I had to change to relay the value :(

Also, since I don't have time tonight to properly test it, adding `changes welcome`.

montehurd: Sorry I ran out of time before week off!

montehurd: Per the comment above this commit e6f1307 should prevent the `401` recursion 🤞

Wikipedia/Code/WMFAccountLogin.swift
```diff
@@ -43,7 +43,7 @@ public class WMFAccountLoginResult: NSObject {
 }
 
 public class WMFAccountLogin: Fetcher {
-    public func login(username: String, password: String, retypePassword: String?, oathToken: String?, captchaID: String?, captchaWord: String?, siteURL: URL, success: @escaping WMFAccountLoginResultBlock, failure: @escaping WMFErrorHandler){
+    public func login(username: String, password: String, retypePassword: String?, oathToken: String?, captchaID: String?, captchaWord: String?, siteURL: URL, reattemptOn401Response: Bool = true, success: @escaping WMFAccountLoginResultBlock, failure: @escaping WMFErrorHandler){
```
joewalsh: It seems that any login call shouldn't attempt re-login - if it failed with 401 once there's no reason to expect it'd succeed on a second try. Is there something I'm missing?

montehurd:
@joewalsh I reverted 599f7e1 and tried a different approach... seems less invasive even though it's still kinda gross...

joewalsh:
@montehurd  I'd suggest going back to the previous approach - it was more explicit and didn't require inspecting request data. It also makes it easier to exclude other requests in the future.

montehurd:
@joewalsh fair enough 👍

montehurd:
(feel free to update this PR if it’s needed before tuesday)

WMF Framework/Fetcher.swift
```diff
@@ -68,8 +68,8 @@ open class Fetcher: NSObject {
         }
     }
     
-    @objc(performTokenizedMediaWikiAPIPOSTWithTokenType:toURL:withBodyParameters:cancellationKey:completionHandler:)
-    @discardableResult public func performTokenizedMediaWikiAPIPOST(tokenType: TokenType = .csrf, to URL: URL?, with bodyParameters: [String: String]?, cancellationKey: CancellationKey? = nil, completionHandler: @escaping ([String: Any]?, HTTPURLResponse?, Error?) -> Swift.Void) -> CancellationKey? {
+    @objc(performTokenizedMediaWikiAPIPOSTWithTokenType:toURL:withBodyParameters:cancellationKey:reattemptLoginOn401Response:completionHandler:)
+    @discardableResult public func performTokenizedMediaWikiAPIPOST(tokenType: TokenType = .csrf, to URL: URL?, with bodyParameters: [String: String]?, cancellationKey: CancellationKey? = nil, reattemptLoginOn401Response: Bool = true, completionHandler: @escaping ([String: Any]?, HTTPURLResponse?, Error?) -> Swift.Void) -> CancellationKey? {
```
nambatee: This is called by `logout` with `reattemptLoginOn401Response` set to `true` (https://github.com/wikimedia/wikipedia-ios/blob/8e114376e0f926a2299745d7b648ee459596c0e0/Wikipedia/Code/WMFAuthenticationManager.swift#L209) - is it intentional? When I log out via the app UI, I wouldn't expect a login attempt

montehurd: oh good catch will check!

PR #3302, opened 2019-10-20
Add missing rtl languages: azb, lrc, nqo

amire80:


tonisevener:
@amire80 thanks for the PR!

PR #3304, opened 2019-10-23
Fix link previewing on iOS 13

joewalsh:
- Remove style that was preventing context menus from showing
- Add new delegate methods, wrap old actions for new setup

https://phabricator.wikimedia.org/T235918


tonisevener:
I notice sizing gets wonky in landscape. Seems to be only on the articles that have a header image.

IMG_0017

PR #3310, opened 2019-10-31
context menu crash - do not allow completionHandler to be called twice

tonisevener:


joewalsh:
I'm confused about how this PR shows a diff - b6e025de578f1f68939db3125c987f9aeeeb64b3 got merged and I see the `return` locally on `6.4.1`. Am I missing something?

tonisevener:
@joewalsh sorry I didn't make the connection that it was the same crash fix. Strange I see it too now in 6.4.1, not sure why this doesn't show conflicts. Closing.

joewalsh:
@tonisevener no problem, thanks for being proactive about fixing the crash even if it turned out to be the same. It looks like this is showing a diff because `6.4.1` wasn't merged into this PR after #3309 was merged into `6.4.1`. GH doesn't seem to do the out-of-date check on non-default branches

PR #3319, opened 2019-11-07
Define in article announcement user experience for iOS app

nambatee:
https://phabricator.wikimedia.org/T235539

cards

Testing:

1. Check out https://gerrit.wikimedia.org/r/#/c/mediawiki/services/wikifeeds/+/545071/, run locally
2. Set scheme to Local
3. Set the correct cookie - https://github.com/wikimedia/wikipedia-ios/blob/3080578435ccd75fb504e2278ee1893811fa259f/WMF%20Framework/WMFAnnouncementsFetcher.m#L36
I tested `US`(feed) and `GB` (in article)


WMF Framework/WMFContentGroup+Extensions.m
```diff
@@ -637,13 +637,11 @@ - (nullable WMFContentGroup *)contentGroupForURL:(NSURL *)URL {
 }
 
 - (nullable WMFContentGroup *)newestGroupOfKind:(WMFContentGroupKind)kind requireIsVisible:(BOOL)isVisibleRequired {
```
joewalsh: it looks like `isVisibleRequired` is no longer necessary or mistakenly ignored

nambatee: mistakenly ignored, fixed in 8c87c4649e5e879809b9717f2fdef18d1c7df68f

joewalsh:
The theory with 5e985b0 was that removing the extra predicate from the explore feed fetch would help with fetch time but after testing there's not really a user perceptible difference. I used an iPhone X with the Core Data instrument in Instruments. Fetches were consistently ~50ms regardless of whether or not the extra predicate was there. Might be a more pronounced difference on older devices and multi-language feeds (or if we ever allow more feed history). Feel free to revert if you'd like or if there's something I missed about using `isVisible` in this way

PR #3324, opened 2019-11-11
Fix AlignedImageButton for RTL & iOS 13

nambatee:
Before
RTL device, 13.2.2 (Save button)

Simulator Screen Shot - iPhone 8 Plus - 2019-11-11 at 12 33 11

After
RTL device, 13.2.2 (Save button)
Simulator Screen Shot - iPhone 8 Plus - 2019-11-11 at 12 39 13

Nothing should change pre iOS 13


joewalsh:
Both issues seem to remain fixed when reverting the RTL/LTR switch:
Simulator Screen Shot - iPhone 11 Pro - 2019-11-12 at 16 28 23
 I committed that change but feel free to revert or update if I'm missing something

PR #3328, opened 2019-11-13
Fix typo

jhsoby:
Thanks was set to » Thanks was se**n**t to

Reported by @Papuass in https://translatewiki.net/wiki/Thread:Support/Reporting_typo_in_source


montehurd:
Oh good catch!

Looks like you'll need to change it here too:
https://github.com/wikimedia/wikipedia-ios/blob/develop/Wikipedia/Code/DiffContainerViewController.swift#L672

(`Wikipedia/Localizations/en.lproj/Localizable.strings` is auto-generated upon project build from LocalizedStrings calls like the one seen on that line)

jhsoby:
{{done}}

PR #3336, opened 2019-11-14
Diff single revision - up / down arrows

tonisevener:
https://phabricator.wikimedia.org/T237416


Wikipedia/Code/DiffContainerViewController.swift
```diff
@@ -691,3 +880,127 @@ class RevisionAuthorThanksErrorHintVC: HintViewController {
...
+        if var newViewControllers = navigationController?.viewControllers {
+            newViewControllers.removeLast()
+            newViewControllers.append(viewController)
+            navigationController?.setViewControllers(newViewControllers, animated: true)
+        }
+    }
+    
+    func tappedPrevious() {
+        
+        self.navigationController?.delegate = self
```
joewalsh: Setting the navigation controller's delegate here (and nil'ing it out elsewhere) breaks the transitions handled by `WMFAppViewController` (explore feed, search) as well as some other functionality it handles as the nav delegate. It seems that since `WMFAppViewController` sets up the nav controller it should remain the delegate. Could the prev/next transitions be added to `ViewControllerTransitionsController` which `WMFAppViewController` uses to set up transitions? Feel free to refactor or adjust `ViewControllerTransitionsController` if you see a better way for it to detect which transition to use

Wikipedia/Code/DiffRevisionAnimator.swift
```diff
@@ -0,0 +1,66 @@
...
+    }
+    
+    func animateTransition(using transitionContext: UIViewControllerContextTransitioning) {
+
+        let container = transitionContext.containerView
+
+        guard let fromView = transitionContext.view(forKey: UITransitionContextViewKey.from) else { return }
+        guard let toView = transitionContext.view(forKey: UITransitionContextViewKey.to) else { return }
+        guard let toVC = (transitionContext.viewController(forKey: UITransitionContextViewControllerKey.to) as? DiffRevisionAnimating) else { return }
+
```
joewalsh: Should these early returns complete the transition?

joewalsh:
The transition looks great! Added a few code review questions inline

Wikipedia/Code/DiffController.swift
```diff
@@ -61,56 +65,40 @@ class DiffController {
...
+                }
             }
-            
-            completion(.success(lastRevisionId))
-            
         }
+        
+        //failing that try fetching revision from API
+        guard let articleTitle = (articleTitle as NSString).wmf_normalizedPageTitle() else {
+            return
```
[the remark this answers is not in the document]
tonisevener: fixed

Wikipedia/Code/DiffContainerViewController.swift
tonisevener: Great catch! I totally missed that - this is integrated into `ViewControllerTransitionsController` now, I think that fixes the issues you were seeing.

Wikipedia/Code/DiffRevisionAnimator.swift
tonisevener: agreed, fixed

PR #3337, opened 2019-11-15
Diffs design updates

nambatee:
- [x] Font size for the body text is still looking a bit small, could we match the font size of the body text of feed cards in the explore feed (looks like 15pt to me)
- [x] If there is no edit summary information, the edit summary section does not need to be shown
- [x] Would it be possible to change the Back link label to History or the page that was navigated to before opening the individual revision?
- [x] Would it be possible for the grey subhead above the revision title (time stamp) to use the N (hours, days, weeks, months, years) ago format?
- [x] HR between edit summary and edit information appears to be darker than the HRs above/below these sections. It would be great to match the HRs above and below these sections.
- [x] Tapping on 'Paragraph moved' doesn't seem to consistently move you to where the paragraph was moved to (only appears to work sometimes)
- [ ] Would it be possible to put 2pts of white space between any moved paragraph elements, as a way of showing that they are separate actions?
- [ ] Currently the moved paragraphs are a bit confusing if no other changes were made. I think we’re going to need to show a line before and after each move and where it was moved to for context


tonisevener:
@nambatee I made some commits to restore the progress bar and the div line between title and summary were missing, let me know if that's ok.

Also added improvements to for moved paragraph scrolling - I needed extra logic to scroll to the item within a change cell, not just to the change cell, since those cells can become so tall. I also am matching the context cell font size & line spacing with the change cell.

nambatee:
That sounds good @tonisevener, thanks for the fixes and follow-up work!

PR #3340, opened 2019-11-18
History & Diff - handle new diff section contract + remaining punch list items

tonisevener:
- Updated DiffController to handle new compare endpoint output structure as discussed [here](https://phabricator.wikimedia.org/T237445#5646931).
- DiffController was getting a bit large so refactored the transform pieces into DiffTransformer.
- Added spacer view to consecutive move items in DiffChangeCell per [punch list](https://phabricator.wikimedia.org/T237314) remaining items.
- Showing navbar at all times on History so user can easily get to Compare button (see [here](https://phabricator.wikimedia.org/T237314#5644350)).
- Made minor flag flow inline on diff summary header view and diff compare header view. (also [here](https://phabricator.wikimedia.org/T237314#5644350)).


Wikipedia/Code/DiffController.swift
```diff
@@ -111,343 +109,53 @@ class DiffController {
     }
     
     private func fetchDiff(fromRevisionId: Int, toRevisionId: Int, theme: Theme, traitCollection: UITraitCollection, completion: @escaping ((Result<[DiffListGroupViewModel], Error>) -> Void)) {
+
+//        let queue = DispatchQueue.global(qos: .userInitiated)
```
tonisevener: Wikidiff2 still needs to be deployed to output diff item offsets. Uncomment here to see it working with an embedded test response that is fully populated.

Wikipedia/Code/DiffController.swift
```diff
@@ -28,20 +23,23 @@ class DiffController {
...
         self.diffFetcher = diffFetcher
         self.revisionFetcher = revisionFetcher
         self.globalUserInfoFetcher = globalUserInfoFetcher
         self.diffThanker = diffThanker
         self.articleTitle = articleTitle
         self.siteURL = siteURL
         self.type = type
+        
+        if let transformer = transformer {
+            self.transformer = transformer
```
nambatee: it looks like `transformer` is never passed as an argument - is it necessary to account for that case or could it be always created here?

tonisevener: @nambatee not necessary as it is now, was just trying to make it injectable for future unit testing with a mock transformer. I can remove until I have actual unit tests in place, pretty sure most of these injections aren't needed yet.

PR #3344, opened 2019-11-20
Diff - handle first revision in history

tonisevener:
https://phabricator.wikimedia.org/T237414

Branched off of https://github.com/wikimedia/wikipedia-ios/pull/3340, review after that is merged.

### DIFF
--- a/Wikipedia/Code/DiffContainerViewController.swift
+++ b/Wikipedia/Code/DiffContainerViewController.swift
@@ -211,6 +211,13 @@ private extension DiffContainerViewController {
         midSetup()
         containerViewModel.state = .loading
         
+        //different handling for populating view models on first revision in history
+        if let firstRevision = firstRevision,
+            toModel.revisionID == firstRevision.revisionID {
+            completeSetup()
+            return
+        }
+        
         diffController.fetchRevision(sourceRevision: toModel, direction: .previous) { (result) in
             switch result {
             case .success(let revision):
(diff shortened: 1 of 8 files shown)


Wikipedia/Code/DiffContainerViewController.swift
```diff
@@ -316,7 +355,8 @@ private extension DiffContainerViewController {
                         self.prevModel = NextPrevModel(from: prevFromModel, to: prevToModel)
                     }
                 }
-            case .failure:
+            case .failure(let error):
+                print(error)
```
joewalsh: Should be a `DDLogError`

Wikipedia/Code/DiffFetcher.swift
```diff
@@ -40,6 +41,57 @@ class DiffFetcher: Fetcher {
...
     
+    func fetchWikitext(siteURL: URL, revisionId: Int, completion: @escaping (Result<String, Error>) -> Void) {
+        
+        let params: [String: AnyObject] = [
+            "action": "query" as AnyObject,
+            "prop": "revisions" as AnyObject,
+            "revids": "\(revisionId)" as AnyObject,
+            "rvprop": "content" as AnyObject,
+            "format": "json" as AnyObject
+        ]
```
joewalsh: `performMediaWikiAPIGET` takes `[String: Any]` so these casts can be avoided. It looks like this was modeled after `PageHistoryFetcher` which takes `[String: AnyObject]` because it's a subclass of the obj-c `WMFLegacyFetcher`. It might be possible to update that to a `Fetcher` but that's outside the scope of this PR

tonisevener:
> Looks good overall. I had a few very minor comments so I just made the changes. Feel free to revert or adjust if there was something I missed

Thanks! Looks good.

PR #3345, opened 2019-11-20
Namespace and mainpage methods

montehurd:
`WikipediaURLTranslations` provides static lookup methods:

- `func commonNamespace(for namespaceString: String, in languageCode: String) -> WikipediaURLCommonNamespace?`

- `func mainpage(in languageCode: String) -> String?`

The lookup json:
- has pruned out ambiguous namespaces (i.e. those whose numbers may mean different things on different language wikis - specifically `104, 105, 106, 107, 110, 111`). Note: I used the term `CommonNamespace` to remind us that some ambiguous namespaces are not represented, but I don't feel strongly about it.
- has added [namespacealiases](https://www.mediawiki.org/wiki/API:Siteinfo#Parameters) - discovered that these are used a lot in some languages, such as Chinese

Update:
Also adds a script for updating the namespace translation files.


Wikipedia/Code/WikipediaURLTranslations.swift
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    static func mainpage(in languageCode: String) -> String? {
+        return WikipediaURLTranslations.sharedLookupTable[languageCode]?.mainpage
+    }
+}
+
+struct WikipediaURLLanguageCodeTranslations:Codable {
+    let namespace:Dictionary<String, WikipediaURLCommonNamespace>
+    let mainpage:String
```
joewalsh: I'm generally not a stickler for the more nuanced style questions, but all of the Swift documentation has a space between the `:` and the type

Wikipedia/Code/WikipediaURLTranslations.swift
```diff
@@ -0,0 +1,93 @@
...
+/*
+ TODO:
+ - re-export the json with all strings lower-cased then update the "commonNamespace" func to force "namespaceString" to lower-case
+ - replace spaces in the json with "_" too - also replace them in "namespaceString"
+*/
+
+struct WikipediaURLTranslations:Codable {
+    private var languagecode:Dictionary<String, WikipediaURLLanguageCodeTranslations> = Dictionary()
+    private static let sharedLookupTable = WikipediaURLTranslations.init(fileName: "wikipedia-namespaces").languagecode
+    private init(fileName : String) {
```
joewalsh: Same comment re: style - all of the Swift doc doesn't have a space between the variable name and the `:`.  Also missing a space between the`:` and `Codable` above. Maybe it's time for a Swift linter

joewalsh:
One more thought - the JSON file takes over a tenth of a second to parse on iPhone X - it might make sense to split out the languages into separate files and lazily build a single in-memory dictionary as languages are requested

PR #3347, opened 2019-11-20
Update WMFAppViewController and NSUserActivity to support in-app link parsing and routing

joewalsh:
Combined fixes for
https://phabricator.wikimedia.org/T228789
and
https://phabricator.wikimedia.org/T234977

Goals
- Handle URL activity type determination and parsing of important parameters in `Configuration` to allow for config-specific determination
- Provide a single routing point for unknown URLs -`NSUserActivity` through `WMFAppViewController`. External URLs already route through here and unknown URLs tapped within the app are now passed using `WMFNavigateToActivityNotification`
- `SinglePageWebViewController` for in-app links (anything on a known Wiki-domain)
- Parse relevant info for History & Diffs into `Router.Destination` and show the appropriate VCs


tonisevener:
My pieces are ready for review.

Wikipedia/Code/wikipedia-namespaces/test.json
```diff
@@ -0,0 +1,49 @@
+{
+  "languagecode": {
+    "en": {
```
montehurd: i think this "en" should be "test" as well?

montehurd: PR to update the test file when the others are updated:
https://github.com/wikimedia/wikipedia-ios/pull/3354

WMF Framework/Router.swift
```diff
@@ -0,0 +1,165 @@
...
+    unowned let configuration: Configuration
+    required init(configuration: Configuration) {
+        self.configuration = configuration
+    }
+    
+    // From https://github.com/wikimedia/mediawiki-title
+    private let namespaceRegex = try! NSRegularExpression(pattern: "^(.+?)_*:_*(.*)$")
+    private let mobilediffRegexCompare = try! NSRegularExpression(pattern: "^mobilediff/([0-9]+)\\.\\.\\.([0-9]+)", options: .caseInsensitive)
+    private let mobilediffRegexSingle = try! NSRegularExpression(pattern: "^mobilediff/([0-9]+)", options: .caseInsensitive)
+    private let historyRegex = try! NSRegularExpression(pattern: "^history/(.*)", options: .caseInsensitive)
```
tonisevener: This may be too broad, I'm trying to capture the article title in something like Special:History/Auguste_Rodin.

WMF Framework/Router.swift
```diff
@@ -0,0 +1,165 @@
...
+    
+     internal func destinationForWikiResourceURL(_ url: URL) -> Destination? {
+        
+        if configuration.isWResource(url) {
```
tonisevener: Figured out why I had a hard time testing - I think we will need to update our [association](https://en.wikipedia.org/apple-app-site-association) file in order for these /w/ path [links](https://phabricator.wikimedia.org/T228789) to intercept. 😕

joewalsh: Good catch! We can edit that file in the [operations/mediawiki-config repo](https://gerrit.wikimedia.org/r/plugins/gitiles/operations/mediawiki-config/+/refs/heads/master/docroot/wikipedia.org/apple-app-site-association). My guess is we'll release this with support then update the file sometime after enough users have upgraded. I'll confirm there's no other reason those URLs were excluded

joewalsh:
Going to merge this as it's getting unwieldy - feel free to post-review and any follow up work will be in more manageable PRs

WMF Framework/Router.swift
```diff
@@ -0,0 +1,162 @@
...
+        case userTalk(_: URL)
+        case search(_: URL, term: String?)
+    }
+    
+    unowned let configuration: Configuration
+    required init(configuration: Configuration) {
+        self.configuration = configuration
+    }
+    
+    private let mobilediffRegexCompare = try! NSRegularExpression(pattern: "^mobilediff/([0-9]+)\\.\\.\\.([0-9]+)", options: .caseInsensitive)
```
montehurd: Could maybe use Swift 5 raw strings ( https://www.hackingwithswift.com/articles/162/how-to-use-raw-strings-in-swift ) now so we don't have to do escaping?

Wikipedia/Code/WikipediaURLTranslations.swift
```diff
@@ -1,98 +1,41 @@
 
 struct WikipediaURLTranslations {
```
montehurd: Nice changes!

scripts/generate_namespace_files.js
```diff
@@ -12,7 +12,15 @@ const fetcher = (url, handler) => {
     res.on('data', (chunk) => {
       data += chunk
     })
-    res.on('end', () => handler(null, JSON.parse(data)))
+   
+    res.on('end', () => {
+      try {
+        handler(null, JSON.parse(data))
+      } catch (err) {
```
montehurd: Good catch! ;)

PR #3354, opened 2019-11-21
Update namespace script to duplicate English file for test wiki

montehurd:
When we run the `generate_namespace_files.js` script we'll get an updated `test.json` too.


montehurd:
closing this since it has a bug (the func i added mutates its last param), and @joewalsh is making related changes

PR #3356, opened 2019-11-22
Increase single diff next/prev button touch targets

montehurd:
https://phabricator.wikimedia.org/T237416#5680007

Ran into this issue with the side margin (hence the -8 leading constraint constant in the storyboard):
https://stackoverflow.com/q/6021138


montehurd:
oops forgot to fix the `!`... fixing...

joewalsh:
Could you update this to just be the button size fix and file the margin issue as a follow on?

montehurd:
The problem is that margin issue moves the next/prev buttons to the right a bit, but if we're ok with that sure 👍

joewalsh:
I'm ok with that for now - is it possible the issue was using the same `marginSpacing` twice in the same toolbar? If you separate it into separate `leadingMarginSpacing` and `trailingMarginSpacing` bar button items does the issue go away?

montehurd:
@joewalsh No, unfortunately :(

montehurd:
Removed!

montehurd:
iirc this [PR](https://github.com/wikimedia/wikipedia-ios/pull/3300/files) may have been same issue?

montehurd:
just noticed the edge insets hack on the other toolbar buttons - seeing about using same trick to move the next/prev...

montehurd:
turns out that's just shifting the icon images vertically a bit...

montehurd:
(technically could use to shift the image horizontally, but that doesn't reposition the touch target obviously)

PR #3367, opened 2019-11-25
Allow diff revision cells to be treated as atomic accessibility links

montehurd:

# Before
before mov

# After
after mov


montehurd:
Will re-open with updated branch name.

PR #3368, opened 2019-11-26
Add event logging for in-article fundraising impressions

joewalsh:
https://phabricator.wikimedia.org/T235447#5691685


joewalsh:
Removing the hold as the questions are answered on the phab ticket

PR #3370, opened 2019-11-26
Allow diff revision cells to be treated as atomic accessibility links

montehurd:

# Before
before mov

# After
after mov


Wikipedia/Code/PageHistoryCollectionViewCell.swift
```diff
@@ -16,6 +16,9 @@ class PageHistoryCollectionViewCell: CollectionViewCell {
     var displayTime: String? {
         didSet {
             timeLabel.text = displayTime
+            if let displayTime = displayTime {
```
joewalsh: Accessibility label is nullable so I'm not sure this check is necessary. It could lead to a state where the accessibility label doesn't match the actual label when the value is nil'd out

Wikipedia/Code/PageHistoryCollectionViewCell.swift
```diff
@@ -52,8 +55,19 @@ class PageHistoryCollectionViewCell: CollectionViewCell {
...
+        let isMinorAccessibilityString = isMinor ? WMFLocalizedString("page-history-revision-minor-edit-accessibility-label", value: "Minor edit", comment: "Accessibility label text used if edit was minor") : ""
+        accessibilityLabel = [timeLabel.accessibilityLabel, authorButton.accessibilityLabel, sizeDiffLabel.accessibilityLabel, isMinorAccessibilityString, commentLabel.accessibilityLabel]
+            .compactMap { $0 }
+            .filter { $0.wmf_hasNonWhitespaceText }
+            .joined(separator: ", ") // Comma adds slight voice-over pause.
+    }
+    
     var comment: String? {
         didSet {
+            if let comment = comment, comment.wmf_hasNonWhitespaceText {
```
joewalsh: Same comment as above - if it needs to be nil'd out for whitespace add an `else { comment.accessoryLabel = nil }`

Wikipedia/Code/PageHistoryCollectionViewCell.swift
```diff
@@ -52,8 +55,19 @@ class PageHistoryCollectionViewCell: CollectionViewCell {
         }
     }
 
+    func updateAccessibilityLabel() {
+        let isMinorAccessibilityString = isMinor ? WMFLocalizedString("page-history-revision-minor-edit-accessibility-label", value: "Minor edit", comment: "Accessibility label text used if edit was minor") : ""
+        accessibilityLabel = [timeLabel.accessibilityLabel, authorButton.accessibilityLabel, sizeDiffLabel.accessibilityLabel, isMinorAccessibilityString, commentLabel.accessibilityLabel]
+            .compactMap { $0 }
+            .filter { $0.wmf_hasNonWhitespaceText }
```
joewalsh: Filter and compact map can be combined here into a single step

Wikipedia/Code/PageHistoryCollectionViewCell.swift
montehurd: oh good catch!

Wikipedia/Code/PageHistoryCollectionViewCell.swift
montehurd: fixed!

Wikipedia/Code/PageHistoryCollectionViewCell.swift
montehurd: True. Seemed cleaner this way than unwrapping optional in the filter... I don't feel strongly about it though. Also I may have missed a simpler approach 🤔

montehurd: Ah didn't think to move the logic into compactMap vs filter 😂...

montehurd: oh wait... seems similarly less clean...

Wikipedia/Code/PageHistoryCountsViewController.swift
```diff
@@ -80,6 +80,11 @@ class PageHistoryCountsViewController: UIViewController {
 
         filterCountsViewController.delegate = self
         wmf_add(childController: filterCountsViewController, andConstrainToEdgesOfContainerView: filterCountsContainerView)
+
+        sparklineView.isAccessibilityElement = true
```
montehurd: I don't feel strongly about this, but some of the accessibility related info I read reminded me voice-over is used by non-blind users (low vision etc) - and I thought a bit of context might be nice for the graph...

Wikipedia/Code/PageHistoryCollectionViewCell.swift
montehurd: 🤔

joewalsh: Clean is subjective. I'd find it cleaner this way:
```
        .compactMap({ (maybeString) -> String? in
            guard
                let string = maybeString,
                string.wmf_hasNonWhitespaceText
            else {
                    return nil
            }
            return string
        })
```
Mainly because the types are explicit and it's easier to set breakpoints. In this case it doesn't really matter because you could inspect the entire list, but if it was a larger array it would be helpful to easily break on the nil condition and inspect what went wrong. Same goes for perf- the existing code iterates twice over the array, but again that doesn't really matter because the array has so few items.

Long story short, I don't feel strongly enough to hold up this PR.

montehurd: That's what I tried and thought was unjustifiably verbose given performance wasn't a concern and it was unlikely to be buggy, but perhaps I should have pulled it out into a function or declared it something like:

```
        let maybeNonWhitespaceString = { (maybeString: String?) -> String? in
            guard
                let string = maybeString,
                string.wmf_hasNonWhitespaceText
            else {
                    return nil
            }
            return string
        }
```

...then it's self-documenting at the point of use it reads nice and declaratively...

```            
.compactMap(maybeNonWhitespaceString)
```

...yet if issues are ever encountered you can dip into`maybeNonWhitespaceString` and set breakpoints, etc 🤔

montehurd: https://github.com/wikimedia/wikipedia-ios/pull/3375

PR #3375, opened 2019-11-27
Make 'maybe non-whitespace' string logic re-usable

montehurd:
For easy re-use by `compactMap`.

https://github.com/wikimedia/wikipedia-ios/pull/3370#discussion_r351227356

### DIFF
--- a/WMF Framework/HasText.swift
+++ b/WMF Framework/HasText.swift
@@ -6,6 +6,16 @@ extension String {
     public var wmf_hasNonWhitespaceText: Bool {
         return wmf_hasText && !self.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty
     }
+
+    public static func wmf_maybeNonWhitespaceString(maybeString: String?) -> String? {
+        guard
+            let string = maybeString,
+            string.wmf_hasNonWhitespaceText
+            else {
+                return nil
+        }
+        return string
+    }
 }
 
 extension NSAttributedString {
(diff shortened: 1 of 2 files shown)


montehurd:
( I don't feel strongly about this - just thought it was cool )

WMF Framework/HasText.swift
```diff
@@ -6,6 +6,16 @@ extension String {
     public var wmf_hasNonWhitespaceText: Bool {
         return wmf_hasText && !self.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty
     }
+
+    public static func wmf_maybeNonWhitespaceString(maybeString: String?) -> String? {
```
joewalsh: I realize it's also in the existing method above, but I believe the `wmf_` prefixes aren't necessary with Swift extensions like they were with Obj-C categories due to namespacing. In this case you could `@objc(wmf_maybe...)`

Wikipedia/Code/PageHistoryCollectionViewCell.swift
```diff
@@ -60,8 +60,7 @@ class PageHistoryCollectionViewCell: CollectionViewCell {
     func updateAccessibilityLabel() {
         let isMinorAccessibilityString = isMinor ? WMFLocalizedString("page-history-revision-minor-edit-accessibility-label", value: "Minor edit", comment: "Accessibility label text used if edit was minor") : ""
         accessibilityLabel = [timeLabel.accessibilityLabel, authorButton.accessibilityLabel, sizeDiffLabel.accessibilityLabel, isMinorAccessibilityString, commentLabel.accessibilityLabel]
-            .compactMap { $0 }
-            .filter { $0.wmf_hasNonWhitespaceText }
+            .compactMap(String.wmf_maybeNonWhitespaceString)
```
joewalsh: I'm not sure I see the utility of having this as an extension at the moment given it's only used here. I think I'd leave it inline for now. If we start using this pattern more for generating accessibility labels, I could see the argument for creating an extension to combine all of these steps (removing empty strings from the array and reducing into a comma separated string)

montehurd: no worries - will close for now

PR #3376, opened 2019-11-27
Remove unused code

montehurd:
Ran this script to see possibly unused code:

https://github.com/PaulTaykalo/swift-scripts/blob/master/unused.rb

The script, as noted [here](https://github.com/PaulTaykalo/swift-scripts#known-issues), produces false positives. I think I weeded these out but may have missed something... Also, I can break this up into separate commits if needed. Can also close and circle back later if this isn't a good time.


joewalsh:
> The script, as noted here, produces false positives

I think given that, it'd be better to wait on this. It also looks like there might be some things unused that are due to bugs (like the empty saved places strings)

PR #3381, opened 2019-12-06
Follow-up diff accessibility fixes

montehurd:
496bd29 and 4b95106 fixes issues flagged by QA [here](https://phabricator.wikimedia.org/T239179#5707207) (preserving the header scrolling the various diff `point:inside:with` overrides are currently allowing):

Also adds accessibility grouping for history filter cells and history compare item views - and adds a static method to make such groupings easier.


montehurd:
Added `WIP` - there's a bit of DRYing and addressing the same issue raised here: https://github.com/wikimedia/wikipedia-ios/pull/3370#pullrequestreview-322921628

Wikipedia/Code/UIAccessibility+Grouping.swift
```diff
@@ -0,0 +1,9 @@
+
+extension UIAccessibility {
```
montehurd: UIAccessibility is an informal protocol, and iirc its declaration only exists in Swift, so we can't use this static method in Obj-C, but thought this might not be a blocker? Also thought about hanging the method off of an Array extension (where the elements were optional strings), but that seemed a bit too leaky an abstraction...

Wikipedia/Code/DiffHeaderCompareItemView.swift
```diff
@@ -76,6 +76,13 @@ class DiffHeaderCompareItemView: UIView {
         updateFonts(with: traitCollection)
 
         self.viewModel = viewModel
+        updateAccessibilityLabel(viewModel: viewModel)
+    }
+    
+    func updateAccessibilityLabel(viewModel: DiffHeaderCompareItemViewModel) {
+        let isMinorAccessibilityString = viewModel.isMinor ? CommonStrings.minorEditTitle : ""
+        let authorString = String.localizedStringWithFormat(CommonStrings.authorTitle, viewModel.username ?? CommonStrings.unknownTitle)
+        accessibilityLabel = UIAccessibility.groupedAccessibilityLabel(for: [headingLabel.text, timestampLabel.text, authorString, isMinorAccessibilityString, summaryLabel.text])
```
tonisevener: Note that the timestampLabel is a link that takes you to the individual revision screen and the usernameLabel is a link that takes you to the user page in Safari. Seems like with this grouping I can't reach either anymore. Might be a good use case for custom actions (see [here](https://developer.apple.com/videos/play/wwdc2019/250/)) to let them navigate to the revision or user page after the group is highlighted.

montehurd: oh no! good catch will check...

montehurd: @tonisevener 

thanks for the custom actions tip! i looked into it and it seemed a bit easier for now to just group the `to/from` title with the timestamp...

Screen Shot 2019-12-09 at 2 57 34 PM

... and the user with the revision msg...

Screen Shot 2019-12-09 at 2 57 51 PM

does that look ok? it's the b144339 commit...

tonisevener: @montehurd cool, I think grouping the title & timestamp works well. Grouping user & summary reads a little weird because VO reads the summary (which could be long), says "link", then if you activate takes you to a user page which might be confusing. Only other feedback is the ordering - right now it goes "From / timestamp" > "To / timestamp" > "left username / summary" > "right username / summary", so might need to tweak the element ordering there so the left username / summary is read after "From / timestamp".

montehurd: doh! forgot to check the ordering 😂

montehurd: fixed!

montehurd: i hear ya though on the user/summary grouping - it's a little bit of a trade-off between grouping on one hand and having too many small touch targets - the user is a bit tiny. i think we're likely ok though even though the grouping can make this logical item a little long sometimes - the reason is i think most people who rely on voice over have the reading speed set super fast.

PR #3383, opened 2019-12-07
Fix editing in Ukrainian

montehurd:
https://phabricator.wikimedia.org/T239632

The first commit re-runs the fetcher so it's easier see that, after the fetcher is changed (second commit), nothing changes except the `uk` file the next time the fetcher runs (third commit).


codemirror-config/main.swift
```diff
@@ -19,6 +19,21 @@ let languagesJSONPathComponents = pathComponents + ["Wikipedia", "assets", "lang
 let languagesJSONPath = languagesJSONPathComponents.joined(separator: "/")
 let languagesJSONData = try! Data(contentsOf: URL(fileURLWithPath: languagesJSONPath))
 let languages = try! JSONDecoder().decode([LanguageJSON].self, from: languagesJSONData)
+let jsonExtractionRegex = try! NSRegularExpression(pattern: #"(?:mw\.config\.set\()(.*?)(?:\);\n*\}\);)"#, options: [.dotMatchesLineSeparators])
```
montehurd: The bug was caused by a line break - the `\n*` fixes it.

montehurd:
experiment i used to kick of a node app in a build phase then send it json (from swift) to validate in js-land:
https://github.com/wikimedia/wikipedia-ios/compare/develop...montehurd:swift-to-node-build-script

( the json gets used by js and swift's json validation was too strict )

PR #3385, opened 2019-12-09
Fix iPad content insets

joewalsh:
- Fix content inset when iPad keyboard is in the split-combined state
- Fix content inset when WKWebView's scroll view adjusts the inset for the keyboard even when it's disabled 

https://phabricator.wikimedia.org/T213446


joewalsh:
Added fixes for the bottom inset calculation which was no longer correct in the section editor after the other fixes. Repro steps > open the section editor > tap the cursor anywhere in the last few lines of a section > observe the cursor scrolling behind the keyboard

PR #3390, opened 2019-12-11
Fix article selected text context menu "Edit" action not selecting text in the editor

joewalsh:
Update (unfortunately still necessary) method swizzling for iOS 13

https://phabricator.wikimedia.org/T240335

### DIFF
--- a/Wikipedia/Code/WKWebViewWithSettableInputViews.swift
+++ b/Wikipedia/Code/WKWebViewWithSettableInputViews.swift
@@ -3,10 +3,10 @@ import WebKit
 class WKWebViewWithSettableInputViews: WKWebView {
     private var storedInputView: UIView?
     private var storedInputAccessoryView: UIView?
-    
+
     override init(frame: CGRect, configuration: WKWebViewConfiguration) {
         super.init(frame: frame, configuration: configuration)
-        setKeyboardRequiresUserInteraction(false)
+        WKWebViewWithSettableInputViews.overrideUserInteractionRequirementIfNecessary()
         overrideNestedContentViewGetters()
     }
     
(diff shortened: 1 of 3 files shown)


Wikipedia/Code/WKWebViewWithSettableInputViews.swift
```diff
@@ -89,24 +89,53 @@ class WKWebViewWithSettableInputViews: WKWebView {
...
+        let sel_13_0: Selector = sel_getUid("_elementDidFocus:userIsInteracting:blurPreviousNode:activityStateChanges:userObject:")
+
+        if let method = class_getInstanceMethod(WKContentView, sel_11_3) {
+            let originalImp: IMP = method_getImplementation(method)
+            let original: NewClosureType = unsafeBitCast(originalImp, to: NewClosureType.self)
+            let block : @convention(block) (Any, UnsafeRawPointer, Bool, Bool, Bool, Any?) -> Void = { (me, arg0, arg1, arg2, arg3, arg4) in
+                original(me, sel_11_3, arg0, !value, arg2, arg3, arg4)
+            }
+            let imp: IMP = imp_implementationWithBlock(block)
+            method_setImplementation(method, imp)
```
montehurd: would adding a `return` after these `method_setImplementation` calls and reversing the order to 13.0, 12.2, 11.3 be a bit more efficient?

Wikipedia/Code/WKWebViewWithSettableInputViews.swift
```diff
@@ -89,24 +89,53 @@ class WKWebViewWithSettableInputViews: WKWebView {
...
         guard let WKContentView: AnyClass = NSClassFromString("WKContentView") else {
-            DDLogError("Could not get class")
+            print("keyboardDisplayRequiresUserAction extension: Cannot find the WKContentView class")
             return
         }
-        let sel = sel_getUid("_startAssistingNode:userIsInteracting:blurPreviousNode:changingActivityState:userObject:")
-        guard let method = class_getInstanceMethod(WKContentView, sel) else {
-            DDLogError("Could not get method")
-            return
+        // For iOS 11.3, *
```
montehurd: good catch on the methods signatures!

Wikipedia/Code/WKWebViewWithSettableInputViews.swift
joewalsh: It's possible multiple are implemented and return-ing would shortcut that, but I haven't checked. Given this is only called once and there doesn't seem to be any performance degradation, I'm ok with leaving it as is

Wikipedia/Code/WKWebViewWithSettableInputViews.swift
joewalsh: I can't take credit, it's a stack overflow answer linked above

montehurd: i did notice this https://stackoverflow.com/a/55344531 example DRY's the swift up a bit - might be nice

montehurd: but i don't feel strongly about it :)

PR #3391, opened 2019-12-11
Fix image gallery dismissal by no longer wrapping it in a UINavigationController

joewalsh:
Work around the link-from-gallery issue (the original reason for wrapping the gallery) by presenting a `UINavigationController` from the gallery controller

### DIFF
--- a/Wikipedia/Code/ViewControllerRouter.swift
+++ b/Wikipedia/Code/ViewControllerRouter.swift
@@ -9,29 +9,27 @@ class ViewControllerRouter: NSObject {
         self.appViewController = appViewController
         self.router = router
     }
-    
-    
-    @objc(routeURL:completion:)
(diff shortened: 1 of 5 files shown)


Wikipedia/Code/ViewControllerRouter.swift
```diff
@@ -17,6 +17,20 @@ class ViewControllerRouter: NSObject {
             completion()
             return false
         }
+        
+        let theme = appViewController.theme
+        
+        let present = { (viewController: UIViewController) in
```
tonisevener: maybe rename to something like presentOrPush? I feel like it's unclear that this could possibly push.

joewalsh: Good idea, fixed!

PR #3395, opened 2019-12-17
mobile html progress

tonisevener:
WIP on mobile html

Testing steps:

1.  Fresh install
2. Tap featured article
3.  Confirm it loads. Tap link. Confirm it pushes.
4. Tap bookmark icon at bottom
5. Go back to Explore, enter airplane mode. Tap featured article again. This is pulling from URLCache.
6. Go to settings, clear cache, go back to Explore. Tap featured article again. This is pulling from Core Data.

Delete steps:
1. Tap featured article
2. Tap bookmark icon
3. Enter airplane mode. Go to Settings and clear cache.
4. Confirm tapping featured article still loads article.
5. Toggle off bookmark icon, go back to Explore
6. Confirm tapping featured article again no longer loads.


Wikipedia/assets/pcs-html-converter/index.html
```diff
@@ -0,0 +1,17 @@
+<!DOCTYPE html>
```
montehurd: this file will be changing in upcoming commits

WMF Framework/CacheDBWriterHelper.swift
```diff
@@ -0,0 +1,118 @@
...
+            guard let item = try moc.fetch(fetchRequest).first else {
+                return nil
+            }
+            return item
+        } catch let error {
+            fatalError(error.localizedDescription)
+        }
+    }
+    
+    static func inMemoryCacheItem(with key: String, in moc: NSManagedObjectContext) -> PersistentCacheItem? {
```
joewalsh: Is there an advantage to this over just running the fetch?

Wikipedia/Code/ArticleFetcher.swift
```diff
@@ -0,0 +1,177 @@
...
+    let scale: String
+    
+    enum CodingKeys: String, CodingKey {
+        case urlString = "src"
+        case scale
+    }
+}
+
+fileprivate enum MediaListItemType: String, Codable {
+    case image
```
joewalsh: Is there a way to add a catch-all here?

joewalsh: Tried an approach here: https://github.com/wikimedia/wikipedia-ios/pull/3408/commits/eb5def903cb40c7ef45690769c78a74ce1d9a02f

WMF Framework/CacheDBWriterHelper.swift
tonisevener: @joewalsh just an optimization I had read about, this could potentially speed things up by avoiding the fetch request and thus the hit to the database if the object is likely already in memory. One change I made recently is try to limit references to PersistentCacheItem & PersistentCacheGroup only to the dbWriters, I thought the separation was nice but that means I'm passing only the keys around and potentially fetching twice in the same flow.  So one flow being:

1. dbWriter creates (or fetches if they already exist) CacheItems
2. fileWriter to fetches & save files from CacheItem keys 
3. dbWriter fetches CacheItems from keys and marks isDownloaded=true

Step 3 could avoid hitting the database because those CacheItems are hopefully already in memory from step 1. I didn't performance test to confirm it's needed though, just thought of in the moment.

joewalsh: @tonisevener I added some logging to measure the time - it looks like after enough `NSManagedObjects` are in memory (after saving) it's about the same speed as the fetch and both options are quick enough. I'd say it's OK to remove the optimization for now and we can revisit (maybe with a constant time lookup) if performance becomes a concern

tonisevener: 👍 sounds good

Wikipedia/Code/ArticleFetcher.swift
tonisevener: @joewalsh will do, good catch

WMF Framework/CacheDBWriterHelper.swift
tonisevener: This has been removed now.

PR #3399, opened 2020-01-06
Header labels are thrown off on rotation

ibaldado:
This is a work-in progress attempting to fix a bug where the title label in collection view headers can be misaligned after device rotation. It fixes the bug, but I’m not very confident in the solution because 1) I still don’t fully understand the reasons the bug happens and 2) I’m not sure if the solution is performant and also how it doesn’t cause draw() and layoutSubviews() to infinitely call each other.

   When it comes to the bug, I found some strange behavior. After `CollectionViewHeader.sizeThatFits()` and `wmf_preferredHeight` are called, the titleLabel’s position (frame.origin.x) is adjusted and correctly set . However, on-screen, the titleLabel will still be at the location it was at pre-rotation. Just to re-iterate, at this point, the frame is correct, but the screen is not reflecting that. 

 When you scroll the collection view up, and the titleLabel starts to leave the screen, the titleLabel redraws and moves to the correct position (the position that its frame was already set at, just not drawn at) 

 Since the frame was already correctly set, I’m trying to find a way to trigger it to get redrawn or relayouted to reflect it. 

 I’d be very interested to get some feedback or help understanding how the frame is being set but not being reflected, and how this solution could be improved to avoid redundant calls to layout and draw.

### DIFF
--- a/Wikipedia/Code/CollectionViewHeader.swift
+++ b/Wikipedia/Code/CollectionViewHeader.swift
@@ -90,6 +90,11 @@ class CollectionViewHeader: SizeThatFitsReusableView {
         button.titleLabel?.font = UIFont.wmf_font(buttonTextStyle, compatibleWithTraitCollection: traitCollection)
     }
     
+    override func layoutMarginsDidChange() {
+        super.layoutMarginsDidChange()
+        setNeedsLayout()
+    }
+    
     override func sizeThatFits(_ size: CGSize, apply: Bool) -> CGSize {
         let additionalMargins: UIEdgeInsets
         switch style {


joewalsh:
@ibaldado I'm seeing the frame of the visible header set to the incorrect `frame.x` value in `sizeThatFits` after rotation back to portrait and the correct value set to offscreen headers. In this example 66 is incorrect for "Today" and 6 is correct for "August 1, 2019":
```
(lldb) po titleLabel.text
▿ Optional<String>
  - some : "Today"
(lldb) po origin.x
66.0
(lldb) po titleLabel.text
▿ Optional<String>
  - some : "August 1, 2019"
(lldb) po origin.x
6.0
````
It looks like the header isn't properly responding to `layoutMargin` changes. The title's x position depends on the `layoutMargin` and it's not updating the layout when it changes. Similar to the approach in `CollectionViewCell`, this should be fixable by calling `setNeedsLayout()` in `layoutMarginsDidChange()`

**Comment by ibaldado** [comment]
Ahh, that makes more sense. Thanks for that direction. I did the correct change and overrode the initial change to keep the git history a little cleaner.

Just for my own understanding, can you tell me more about how you were inspecting the frames? In CollectionViewHeader. sizeThatFits(), the last time it was being called, I was seeing the header view's frame.origin.x to be 0, and titleLabel's frame.origin.x to be the expected value.

Where in the method were you putting your breakpoint? And also at what point in the app or view cycle were you inspecting it? Just trying to understand what I am missing.

Any insight is appreciated. Thanks!

**Comment by joewalsh** [comment]
> Where in the method were you putting your breakpoint? And also at what point in the app or view cycle were you inspecting it? Just trying to understand what I am missing.

I put the breakpoint on this line
```
        origin.y += titleLabel.wmf_preferredHeight(at: origin, maximumWidth:widthMinusMargins, horizontalAlignment: labelHorizontalAlignment, spacing: 0, apply: apply)
```
And would log only with `apply == true`. I would enter the page history screen and rotate to landscape and back to inspect the values when the bug occurs.

Here's a screenshot of the breakpoint setup:
Screen Shot 2020-01-07 at 12 41 05 PM

Hope this helps. Thanks for the pull request!

PR #3408, opened 2020-01-15
Minor fixes for SavedArticlesFetcher.swift

joewalsh:
Confirmed that this seems to be working (testing with the save buttons in the feed and the updated article view save button in the feature/mobile-html-ui branch)

### DIFF
--- a/Wikipedia/Code/SavedArticlesFetcher.swift
+++ b/Wikipedia/Code/SavedArticlesFetcher.swift
@@ -156,6 +156,7 @@ private extension SavedArticlesFetcher {
                 }
             }) { (groupResult) in
                 DispatchQueue.main.async {
+                    self.isUpdating = false
                     switch groupResult {
                     case .success(let itemKeys):
                         print("🥶group completion: \(articleKey), itemKeyCount: \(itemKeys.count)")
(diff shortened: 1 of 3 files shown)


Wikipedia/Code/SavedArticlesFetcher.swift
```diff
@@ -76,7 +76,7 @@ private extension SavedArticlesFetcher {
     }
     
     @objc func articleWasUpdated(_ note: Notification) {
-        
+        update()
     }
     
     @objc func syncDidFinish(_ note: Notification) {
```
tonisevener: @joewalsh I think I missed the update() call in syncDidFinish() as well (see LegacySavedArticlesFetcher.m)

joewalsh: @tonisevener  good catch - fixed!

Wikipedia/Code/SavedArticlesFetcher.swift
```diff
@@ -196,51 +199,61 @@ private extension SavedArticlesFetcher {
...
-                    switch groupResult {
-                    case .success:
+                    DispatchQueue.main.async {
+                        switch groupResult {
+                        case .success:
+                            print("🙈success in groupCompletion of \(articleKey)")
+                            self.didRemoveArticle(with: articleKey)
+                            self.updateFetchesInProcessCount()
+                        case .failure:
+                            print("🙈failure in groupCompletion of \(articleKey)")
```
tonisevener: @joewalsh we might need a call here to self.updateFetchesInProcessCount()

tonisevener: eh on second thought maybe not. If it fails to remove, isDownloaded will remain true thus the in process count won't change.

PR #3409, opened 2020-01-15
Initial mobile-html UI pass

joewalsh:
- Pass necessary params to page content service setup function
- Add lead image and lead image transition
- Add W and search buttons
- Add toolbar, secondToolbar to `ViewController` from `WMFViewController`
- Add theme button and wire up theming and text size
- Update save button

Currently targeted to `feature/mobile-html` to clean up the diff, but should be retargeted to `develop` after that's merged


Wikipedia/Code/ViewController.swift
```diff
@@ -331,9 +343,143 @@ class ViewController: PreviewingViewController, NavigationBarHiderDelegate {
...
+    
+    lazy var secondToolbarVisibleConstraint: NSLayoutConstraint = {
+        return toolbar.topAnchor.constraint(equalTo: secondToolbar.bottomAnchor, constant: ViewController.secondToolbarSpacing)
+    }()
+    
+    lazy var secondToolbarHiddenConstraint: NSLayoutConstraint = {
+        return secondToolbar.topAnchor.constraint(equalTo: toolbar.topAnchor)
+    }()
+    
+    func setupToolbars() {
```
tonisevener: Good to see these consolidated to one place - I'm a little wary of every ViewController owning and having two in their hierarchy since this is called on viewDidLoad(). Maybe we can at least have a showsToolbars boolean to bypass for subclasses that don't need it?

Wikipedia/Code/ArticleContainerViewController.swift
```diff
@@ -18,26 +18,154 @@ class ArticleContainerViewController: ViewController {
         case loading
         case data
     }
-  
-    private var webViewController: ArticleWebViewController?
```
tonisevener: I'd be fine with changing the name of this class to ArticleViewController, doesn't seem like the we need the distinction anymore.

Wikipedia/Code/ArticleContainerViewController.swift
```diff
@@ -18,26 +18,154 @@ class ArticleContainerViewController: ViewController {
...
+    }
+    
+    func loadLeadImage(with leadImageURL: URL) {
+        leadImageHeightConstraint.constant = leadImageHeight
+        leadImageView.wmf_setImage(with: leadImageURL, detectFaces: true, onGPU: true, failure: { (error) in
+            DDLogError("Error loading lead image: \(error)")
+            self.showWebView()
+        }) {
+            self.layoutLeadImage()
+            self.showWebView()
```
tonisevener: Seems like showWebView could be called twice - once when setup() completes and once when lead image loading completes. That might be intentional but just wanted to point out.

Wikipedia/Code/ArticleToolbarController.swift
```diff
@@ -0,0 +1,88 @@
...
+    func toggleSave(from controller: ArticleToolbarController, shouldSave: Bool)
+    func saveButtonWasLongPressed(from controller: ArticleToolbarController)
+    func showThemePopover(from controller: ArticleToolbarController)
+}
+
+class ArticleToolbarController: Themeable {
+    weak var delegate: ArticleToolbarHandling?
+    
+    let toolbar: UIToolbar
+    private var isSaved: Bool = false //tonitodo: better state handling here
```
tonisevener: We could probably clean this out now and have all state flow through setSavedState. This was just a quick way for me to toggle before for testing.

Wikipedia/Code/ViewController.swift
joewalsh: Good idea - I adjusted it so that VCs that want toolbars need to enable them by calling an `enableToolbars()` func

Wikipedia/Code/ArticleContainerViewController.swift
joewalsh: Sounds good - I'll make this change in mobile-html-ui-2 to avoid merge conflicts

Wikipedia/Code/ArticleContainerViewController.swift
joewalsh: It was intended (so it's shown when either the image or content are showing) and there's a check at the beginning of the method where it should bail the second time. However, thinking about it now, this was all a workaround for issues with how the old article view worked so it looks like it can be removed for now and we can revisit if the transition breaks again or we want to add the cross fade back in

PR #3411, opened 2020-01-16
Rework random and help VCs to work with the new Article view

joewalsh:
- Rename `ArticleContainerViewController` to `ArticleViewController`
- Random becomes a subclass of `ArticleViewController`
- Help becomes a subclass of `SinglePageWebViewController` - we lose the ToC but I think that's OK given that it's now displayed as mobile web
- Remove other uses of `WMFLegacyArticleViewController`

https://phabricator.wikimedia.org/T240033


Wikipedia/Code/ArticleURLListViewController.swift
```diff
@@ -108,19 +118,6 @@ extension ArticleURLListViewController {
     }
 }
 
-// MARK: - WMFArticlePreviewingActionsDelegate
-extension ArticleURLListViewController {
-    override func shareArticlePreviewActionSelected(withArticleController articleController: WMFLegacyArticleViewController, shareActivityController: UIActivityViewController) {
-        FeedFunnel.shared.logFeedDetailShareTapped(for: feedFunnelContext, index: previewedIndexPath?.item)
```
tonisevener: I noticed this line went away in the move - is it needed?

joewalsh: Good catch - that is needed! Fixed

PR #3412, opened 2020-01-17
Converter updates

montehurd:
https://phabricator.wikimedia.org/T240031

- Adds a `isConversionFromMobileViewNeeded` flag to WMFArticle defaulting all existing records for saved articles to have this value set to `true`. New saved articles will default to `false`.
- Implements a `MobileviewToMobileHTMLConverter` and a `MobileViewToMobileHTMLMigrationController`
- The `migrateMobileviewToMobileHTMLIfNecessary` method should also probably be used by normal article loading logic to ensure any needed migration has occurred in case the user taps a saved article, while offline, before that particular saved article has been migrated: https://phabricator.wikimedia.org/T244621


Wikipedia/Code/WMFAppViewController.m
```diff
@@ -994,6 +994,18 @@ - (void)pauseApp {
         [self.dataStore removeArticlesWithURLsFromCache:articleURLsToRemoveFromDisk];
     }
 
+    // TODO: instead of checking `savedPageList` just query the data store for articles with the isConversionFromMobileviewNeeded flag true?
```
joewalsh: Yes, and it should operate in batches, most likely only one article at a time. With this setup, it looks like the `WKWebView` in `MobileviewToMobileHTMLConverter` often doesn't load the page when the app is backgrounded and waits til the next foregrounding. This delay causes the migrations in the buffer to remain incomplete until the next launch. Given this, it's likely this queue would be better off in `SavedArticlesFetcher`(which is really a `SavedArticlesController` in our current naming convention) and operate in a similar manner as the download queue - fetch and convert one at a time while the app is foregrounded.

Wikipedia/Code/MWKDataStore.m
```diff
@@ -518,13 +537,35 @@ - (void)performLibraryUpdates:(dispatch_block_t)completion {
...
         done();
     }];
 }
 
+- (void)markAllDownloadedArticlesInManagedObjectContextAsNeedingConversionFromMobileview:(NSManagedObjectContext *)moc {
+    NSFetchRequest *request = [WMFArticle fetchRequest];
+    request.predicate = [NSPredicate predicateWithFormat:@"isDownloaded == YES"];
+    request.propertiesToFetch = @[];
+    NSError *fetchError = nil;
+    NSArray *downloadedArticles = [moc executeFetchRequest:request error:&fetchError];
```
joewalsh: This fetch should be batched similar to the migrations below to prevent an excessively large fetch result for users with thousands of saved articles

Wikipedia/Code/MWKArticle+LegacyMobileview.swift
```diff
@@ -161,14 +158,17 @@ class MobileviewToMobileHTMLConverter : NSObject, WKNavigationDelegate {
...
+        conversionBuffer.removeAll()
     }
 }
 
-extension MobileviewToMobileHTMLConverter {
-    func convertMobileviewSavedDataToMobileHTML(article: MWKArticle, articleCacheController: ArticleCacheController) {
+@objc public extension MobileviewToMobileHTMLConverter {
+    func convertMobileviewSavedDataToMobileHTML(article: MWKArticle, completionHandler: ((Any?, Error?) -> Void)? = nil) {
         guard let articleURL = article.url else {
             assertionFailure("Article url not available")
```
joewalsh: The new completion handler should be called in each early return

Wikipedia/Code/WMFArticle+LegacyMobileview.swift
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        guard let articleCacheController = dataStore.articleCacheControllerWrapper.cacheController as? ArticleCacheController else {
+            completionHandler(MigrateMobileviewToMobileHTMLIfNecessaryError.noArticleCacheController)
+            return
+        }
+        
+        let mwkArticle = dataStore.article(with: articleURL)
+
+        dataStore.mobileviewConverter.convertMobileviewSavedDataToMobileHTML(article: mwkArticle) { (result, error) in
```
joewalsh: It seems like this conversion could be entirely handled by `MWKDataStore` rather than having the model object `WMFArticle` know about the data store. Is there an advantage to this approach that I'm missing?

montehurd: You're likely right :) Will check...

montehurd:
@joewalsh Before 9389e98 the articles were loading if you selected them from "Saved" after the conversion, but while still offline. (They looked a bit funny and could take a bit to render because the bundled `mobile-html-offline-resources` weren't in place.) Only just now noticed that they're not loading since 9389e98 - sorry I didn't have time to track down why :(

joewalsh:
@montehurd no problem - it looks like it's loading on 9389e98 in the Apps Labs configuration but not on the production config. There's a mismatch in the desktop article URL calculated by the cache with the prod config. I'll fix that in a follow-on.

PR #3417, opened 2020-01-24
mobile-html - etag functionality

tonisevener:
This needs to be tested in Production with some workarounds because the mobile-html-offline-resources endpoint is failing with 404.

1. unwrap `mobileHTMLURL.prodHost` and insert in place of `stagingHost` [here](https://github.com/wikimedia/wikipedia-ios/blob/feature/mobile-html-more-caching/WMF%20Framework/ArticleURLConverter.swift#L23).
2. Use `Configuration.current` [here](https://github.com/wikimedia/wikipedia-ios/blob/feature/mobile-html-more-caching/WMF%20Framework/ArticleURLConverter.swift#L64).
3. Comment out lines 116-120 [here](https://github.com/wikimedia/wikipedia-ios/blob/feature/mobile-html-more-caching/Wikipedia/Code/ArticleCacheDBWriter.swift#L116) and remove `mobileHtmlOfflineResourceItemKeys` from `mustHaveKeys` [here](https://github.com/wikimedia/wikipedia-ios/blob/feature/mobile-html-more-caching/Wikipedia/Code/ArticleCacheDBWriter.swift#L122)
4. Replace `self.didFailToFetchArticle(with: articleKey, error: error)` with `self.didFetchArticle(with: articleKey)` [here](https://github.com/wikimedia/wikipedia-ios/blob/feature/mobile-html-more-caching/Wikipedia/Code/SavedArticlesFetcher.swift#L168).

Here's some testing steps (I was watching specifically for the mobile-html endpoint):

1. Open an article, Charles shows hitting network at 200.
2. Open same article, Charles returns 304 and shows quickly because it's pulled from URLCache.
3. Bookmark article, Charles makes requests showing 200 to persist (might be ideal if this were 304 and it pulled from what was already in the URLCache instead...sometimes I see a 304 here actually? seems intermittent).
4. Go to Settings, tap clear URL cache.
5. Go back to article. Article still shows quickly because it's pulling from Persistent cache now. Charles shows 304 again.
6. Update article from elsewhere, go back to article in app. Article refetches and shows latest changes, you see 200 in Charles.
7. Go back, enter airplane mode. article still shows.
8. Clear cache in settings while in airplane mode, go back to article. article still shows bc it's pulling from persistent cache. (formatting might be off because it 404'ed when trying to fetch resources in Prod but that should fix itself once the 404 goes away)
9. Go to an article, tap save, background, foreground, terminate. relaunch, confirm you land on article page and NO mobile-html calls went out (because we are forcing from cache when restoring state upon launch for quick article viewing).


Wikipedia/Code/SchemeHandler/SchemeHandler.swift
```diff
@@ -107,37 +112,47 @@ extension SchemeHandler: WKURLSchemeHandler {
...
+                    return
             }
+            
             // IMPORTANT: Ensure the urlSchemeTask is not strongly captured by this block operation
             // Otherwise it will sometimes be deallocated on a non-main thread, causing a crash https://phabricator.wikimedia.org/T224113
             let op = BlockOperation { [weak urlSchemeTask] in
-                if let cachedResponse = self.articleCacheController?.recentCachedURLResponse(for: defaultURL) {
+                
+                //forceCache will be true for ArticleViewControllers when coming from Navigation State Controller. In this case stale data is fine.
+                if self.forceCache,
```
joewalsh: Is this alright without `forceCache` here and on the `ArticleViewController`? I think we should be always OK with the cached data unless I'm missing something

tonisevener: @joewalsh should be, the default functionality (when this flag is false) will attempt to pull from cache first if the response header rules indicate that. The mobile-html response header for `cache-control `returns `max-age=0` and `must-revalidate`, so it first validates it with the etag and if it sees 304 then pulls from cache (URLCache then fallback to PersistentCache). This `forceCache` flag bypasses this step and goes straight to pulling from cache if we know there's anything there. So it won't bother even validating to check for a 304 and will see stale data if edits have been made since then. For now this is only used when we restore their last read article upon launch.

I realize this flag sortof duplicates a `NSURLRequest.CachePolicy.returnCacheDataElseLoad` policy, but if I went with that policy I think I would have had to subclass `URLCache` and hit this PersistentCache system within it for the PersistentCache fallback functionality (this is how the lite project did it in with `PermanentlyPersistableURLCache`). Instead this flows through `CacheController.cachedURLResponse(for request: URLRequest)` which decides between URLCache or PersistentCache. Just felt a little more targeted to me than having every call that goes through URLCache (calls that have nothing to do with saved articles) potentially touch PersistentCache under the hood.

joewalsh: @tonisevener makes sense, thanks for the explanation! I see now that it's helpful to have this flag to avoid the pre-request check and agree it's more targeted this way

PR #3418, opened 2020-01-24
Make webView background clear

mdavison:
Prevents flash of white during loading when in dark mode
https://phabricator.wikimedia.org/T190847

It looks like the issue is just the default white background showing through before the web view with the dark background gets loaded. I just made the web view background transparent and that seems to have fixed the issue. I confirmed the background is still white while the web view is loading when the device is in light mode.

Also, I don't know if this is expected, but the following tests are failing for me on the develop branch:

`SecurityTests` -> `testFileProtection()` (line 28: "Unable to get fileProtection value")
`UITests` -> `testRecordAppScreenshots()` (I am using the configurations listed in `configurations_to_test_on_pull` in the Fastfile)


joewalsh:
It looks like the tests are passing on the build server and for me locally using an iPhone X sim running iOS 12. I'll see if I can replicate the problem  you're seeing - which config from `configurations_to_test_on_pull` are you using?

mdavison:
@joewalsh Thanks for the feedback!
I'm getting the failures on both configs:
`device: 'iPhone X', os: '12.4'`
 `device: 'iPad Air 2', os: '11.4'`
I'm using Xcode 11.3

mdavison:
@joewalsh Hmm... just setting the scroll view background to black doesn't seem to be working. I can only fix it by setting the web view background to transparent.
(See also: https://stackoverflow.com/questions/27655930/how-can-i-give-wkwebview-a-colored-background)
I could set the background in the loading state, like this?

```
private var state: ViewState = .loading {
    didSet {
        switch state {
        case .unknown:
            fakeProgressController.stop()
            webView.isOpaque = true
        case .loading:
            fakeProgressController.start()
            webView.isOpaque = false
        case .data:
            fakeProgressController.stop()
            webView.isOpaque = true
        }
    }
}
```

joewalsh:
@mdavison in `ArticleViewController`'s `apply(theme:)` function the correct background color for the theme is set to the `view` and the `webView` - it should also be set to `webView.scrollView`. At line 203:
```
        view.backgroundColor = theme.colors.paperBackground
        webView.backgroundColor = theme.colors.paperBackground
```
I'm looking into the testing issues now, I'll let you know if I can or can't repro. Thanks!

mdavison:
@joewalsh Right 👍. I can go ahead and set that for completeness but it doesn't seem to make a difference as far as the flash of white that appears during loading.

joewalsh:
@mdavison my bad! You're right, that doesn't fix it. It looks like it's happening intermittently for me and I thought making the change fixed it. If the clear background fixes it, it'd be good to keep your original fix and remove the lines that change the webView's background color from `apply(theme: Theme)`

PR #3425, opened 2020-02-04
mobile-html: Prototype of native references view

joewalsh:
Initial implementation as starting point for discussion with @carolynlimadeo on how to handle references

- Native reference list VC when tapping "View references" in the footer
- Port over existing reference sheet/popover presentation to display when tapping reference links in the article


joewalsh:
Adding WIP to add the reference panel work to this PR as well

PR #3429, opened 2020-02-06
mobile-html variant handling

tonisevener:
First pass at variant handling. WIP, still testing and cleaning up.

- PersistentCacheItem itemKey will now have a variant string identifier appended to the end to keep uniqueness.
- Added variantGroupKey to PersistentCacheItem to search against and find other variants for fallbacks.
- Image caching now only downloads one size, chosen depending on UIScreen scale.
- Pulling from image cache falls back to other variant size if needed.
- Pulling from article cache now considers appended language variant to itemKey, with a fallback to other variants, so we could potentially cache multiple variants now. WMFArticle keys have not been changed.


tonisevener:
Going to rework considering session headers as discussed in sync.

PR #3439, opened 2020-02-12
Consolidate user defaults

joewalsh:
The old article view was using `NSUserDefaults.standardUserDefaults` in some places and `NSUserDefaults.wmf` in other places. To avoid this confusion in the future, this PR migrates all the user defaults into `standardUserDefaults`. The `wmf` extension was used to ensure the defaults could be read by the widget extensions. The extensions no longer rely on `UserDefaults` so the custom `wmf` suite can be migrated to `standardUserDefaults` and removed.


Wikipedia/Code/MWKDataStore.m
```diff
@@ -433,7 +433,7 @@ - (void)performUpdatesFromLibraryVersion:(NSUInteger)currentLibraryVersion inMan
     }
 
     if (currentLibraryVersion < 8) {
-        NSUserDefaults *ud = [NSUserDefaults wmf];
+        NSUserDefaults *ud = [NSUserDefaults standardUserDefaults];
```
tonisevener: @joewalsh do you think we should move the library version 10 migration up to before this point?

joewalsh: Good catch! I changed this to keep using the wmf defaults since it's before the conversion. I'd be a bit uncomfortable putting the migrations out of order but let me know if there's something I'm missing about that approach

tonisevener: @joewalsh cool yeah agreed, feels better to keep the ordering.

PR #3440, opened 2020-02-12
Caching rework

tonisevener:
Reworked caching to lean more on URLRequests and their headers for Session to pick up & consider for offline and variant fallbacks. This hopefully completes the caching setup. Some extra related questions I had were:

- The custom request headers (Persistent-Cache-Item-Key, etc. ) that Session references for caching do show in Charles and are sent to the server which is not ideal. I can strip them out right before setting up the task, but I will need to map it to the original requests with the headers intact for when URLSession calls back with an error or a 304. Doable but wanted to make sure that's necessary before doing that work.
- I have noticed in Prod even after updating a test article I will receive a 304 from the server if I pass in an old If-None-Match. Mobile web also doesn't seem to update with the newest changes so I think this is a server-side issue.

A couple of remaining bugs I still see:
- The lead image doesn't show when launching in airplane mode and looking at a saved article
- Media-list is returning from cache when attempting to transition to the gallery, but imageInfoFetcher.fetchGalleryInfo is not cached.


WMF Framework/ImageCacheHeaderProvider.swift
```diff
@@ -0,0 +1,49 @@
...
+        header[Session.Header.persistentCacheItemKey] = itemKey
+        
+        if let variant = variantForURL(url) {
+            header[Session.Header.persistentCacheItemVariant] = variant
+        }
+        
+        header[Session.Header.persistentCacheItemType] = "Image"
+        header[Session.Header.persistentCacheForceCache] = String(forceCache)
+        //The accept profile is case sensitive https://gerrit.wikimedia.org/r/#/c/356429/
+        header["Accept"] = "application/json; charset=utf-8; profile=\"https://www.mediawiki.org/wiki/Specs/Summary/1.1.2\""
```
joewalsh: This header isn't necessary for image requests - it tells the server we want UTF8 encoded JSON and the specific spec to use for the endpoint. Looking closely, I think I've used this incorrectly elsewhere, it should only go to the summary endpoint

WMF Framework/ImageCacheHeaderProvider.swift
```diff
@@ -0,0 +1,49 @@
...
+        return sizePrefix == NSNotFound ? nil : String(sizePrefix)
+    }
+
+    func uniqueFileNameForURL(_ url: URL) -> String? {
+        guard let itemKey = itemKeyForURL(url),
+            let variant = variantForURL(url) else {
+                return nil
+        }
+        
+        return "\(itemKey)__\(variant)".precomposedStringWithCanonicalMapping
```
joewalsh: Same filename comment as above, but it sounds like the separate `ImageCache` might be going away anyway

Wikipedia/Code/Session.swift
```diff
@@ -432,6 +469,10 @@ class SessionDelegate: NSObject, URLSessionDelegate, URLSessionDataDelegate {
     }
     
     func urlSession(_ session: URLSession, dataTask: URLSessionDataTask, didReceive response: URLResponse, completionHandler: @escaping (URLSession.ResponseDisposition) -> Void) {
+        
+        //tonitodo: if response is not modified (304) and urlRequest header indicates there's a persistentCacheItemKey associated, pull from cache (try URLCache first, then PersistentCache) here and callback with that data.
+        //if cached data not there, check header Persistent-Cache-Fallback-Type. if image use variant image fallback logic and return that cached response. otherwise use standard fallback logic which is look up any other item response with the same item key but different variant.
+
```
joewalsh: 👍

Wikipedia/Code/Session.swift
```diff
@@ -210,11 +212,18 @@ import Foundation
     }
     
     public func dataTask(with request: URLRequest, callback: Callback) -> URLSessionTask {
+        
+        //tonitodo: may try a wrapper method here of some sort for SchemeHandler, if request header forceCache is true, return from cache and avoid urlsession task resume altogether. else kickoff the session task.
```
joewalsh: I might be missing something, but could the presence or lack of presence of the cache key header be an indicator of whether or not to check the persistent cache instead of a separate `forceCache` header?

WMF Framework/ArticleCacheHeaderProvider.swift
```diff
@@ -0,0 +1,75 @@
...
+                if firstPart == language {
+                    
+                    //return entire variant
+                    return languageCode
+                }
+            }
+        }
+        
+        //maybe tonitodo? default variant here. If url.wmf_language supports variants (not sure if we want to hardcode variant-able languages or not but if so use that, could also try pulling a previous cached urlresponse header and check Vary value.
+
```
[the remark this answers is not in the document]
joewalsh: I cleaned up the `Locale` extensions and added a method for getting the user's preferred language variant for a given URL in 1c8c1305f0d2f37eb437fb9b63458c412cfd39e7. Let me know what you think!

Wikipedia/Code/Session.swift
tonisevener: @joewalsh yeah, this was for handling that case where they restore from NavigationStateController on launch - don't even try to fetch anything first so they see the article as quickly as possible. Standard handling if the cache key exists in the header would be to still try to fetch then fallback to cache if there's an error or a 304, so they'd still see a loading bar. I think rather than a forceCache flag better would be assign NSURLRequest.CachePolicy here to .returnCacheDataElseLoad and check for that here, falling back to PersistentCache if URLCache doesn't exist.

WMF Framework/ArticleCacheHeaderProvider.swift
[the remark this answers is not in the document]
tonisevener: @joewalsh awesome thanks! yeah I was more wondering if we should encode a default variant if it **is** a language that supports variants but for whatever reason we can't glean an applicable preferred variant from the user's language settings. Not sure if that's even a possibility though. We could just leave variant blank here and in Session.swift just return whatever response returns first from the cache for that itemKey if urlRequest has no variant.

Wikipedia/Code/Session.swift
joewalsh: Sounds good - I forgot about that case (again)!

PR #3443, opened 2020-02-14
mobile-html: Implement find in page, fix other bugs

joewalsh:
- Implement find in page for mobile-html using the existing views and JavaScript
- Fix articles not appearing at the right place in history. `viewedDateWithoutTime` was not being updated. Removes `MWKHistoryList` and replaces it with extensions on `NSManagedObjectContext` and `WMFArticle`
- Fix for redirected articles. Repro steps: find a link in an article that leads to a redirect and tap through. I tested with enwiki > Ken Miles > 24 Hours of Daytona results section > 1966 link in the table
- Fix issues found by the static analyzer

https://phabricator.wikimedia.org/T240024


WMF Framework/NSManagedObjectContext+History.swift
```diff
@@ -0,0 +1,53 @@
...
+    }
+    
+    public func addToReadHistory() throws {
+        viewedDate = Date()
+        updateViewedDateWithoutTime()
+        try managedObjectContext?.save()
+    }
+    
+    fileprivate func removeFromReadHistoryWithoutSaving() {
+        viewedDate = nil
```
tonisevener: @joewalsh I notice the old code would set wasSignificantlyViewed = NO here. Is that still needed?

joewalsh: @tonisevener good catch, fixed! I also restored the `significantlyViewedTimer` to the `articleViewController`

PR #3450, opened 2020-02-23
Load gallery in offline mode

tonisevener:
Loading gallery in offline mode with current setup. This might change back to leaning on WMFImageController once migration is hashed out.


tonisevener:
> Looks good! I ran into an issue where article downloads would fail after the first download since the shared endpoints started returning 304. I pushed a fix but feel free to revert or adjust if that case should be handled differently. The other fix is for a bug I introduced in the article view, saved articles weren't being properly added to the default reading list.

@joewalsh great catch, thanks! I added another fix - the button state wasn't responding and downloads weren't kicking off until I set isSaved within the reading list controller methods, feel free to change if you feel it belongs elsewhere. Also I noticed the default reading list wasn't being created, changing `fetchOrCreateDefaultReadingList` to use `wmf_fetchOrCreate` fixed it.

joewalsh:
> the button state wasn't responding and downloads weren't kicking off until I set isSaved within the reading list controller methods, feel free to change if you feel it belongs elsewhere. Also I noticed the default reading list wasn't being created, changing fetchOrCreateDefaultReadingList to use wmf_fetchOrCreate fixed it.

@tonisevener it looks like both problems were stemming from the lack of default list - since they weren't being added to the default list, the saved date wasn't being set. Funny enough I fixed the default list issue as well on the other pr 86e684a but due to the severity of the issue I should have made a PR for just that. Merging this now

PR #3454, opened 2020-02-25
Restore article context menu previewing and 3d touch

joewalsh:
Handles https://phabricator.wikimedia.org/T242063 and the last item of https://phabricator.wikimedia.org/T240023


Wikipedia/Code/ArticleViewController+LinkPreviewing.swift
```diff
@@ -0,0 +1,234 @@
...
+        let bail = {
+            completionHandler(nullConfig)
+        }
+        
+        guard let linkURL = elementInfo.linkURL else {
+            bail()
+            return
+        }
+        
+        getPeekViewControllerAsync(for: linkURL) { (peekParentVC) in
```
tonisevener: @joewalsh since this is async, do you think we also need a timeout here?

joewalsh: Good catch, fixed!

PR #3457, opened 2020-02-27
Use old Cache database, migrate bundled resources

tonisevener:
This wraps up https://phabricator.wikimedia.org/T242059 & https://phabricator.wikimedia.org/T244620

- Updates for caching classes to use the old Cache database so we don't have to deal with migrating across databases.
- Added a custom migration to change the variant field to use a String for article language variants and am changing the name of the old cache file directory to not include the word "Image". 
- Added bundled mobile-html resources (minus site-specific CSS) for offline migration.
- Added delta logic from ImageController into ImageCacheController. This was an easier lift for me than restructuring things to use ImageController, but I think good next steps would be to make some high-level integration tests then refactor so that Fetchers can cache their response or return from cache or fallback in case of errors/304s, and have SchemeHandler call the appropriate Fetcher to get data.
- Still see some bugs but I'd rather we fix those in a followup since this PR is getting large (Chinese header image not showing in offline migration, gallery not showing in offline migration, odd migration output like vertical spaces around images, once I saw references displayed first, etc.)

Next steps are https://phabricator.wikimedia.org/T244621 and https://phabricator.wikimedia.org/T246722.


WMF Framework/CacheProviderHelper.swift
```diff
@@ -13,12 +13,26 @@ final class CacheProviderHelper {
...
         
-        guard let responseData = FileManager.default.contents(atPath: CacheFileWriterHelper.fileURL(for: responseFileName).path),
-            let responseHeaderData = FileManager.default.contents(atPath: CacheFileWriterHelper.fileURL(for: responseHeaderFileName).path) else {
+        guard let responseData = FileManager.default.contents(atPath: CacheFileWriterHelper.fileURL(for: responseFileName).path) else {
             return nil
         }
         
-        //let mimeType = FileManager.default.getValueForExtendedFileAttributeNamed(WMFExtendedFileAttributeNameMIMEType, forFileAtPath: responseFileName)
+        //missing response header file could happen as a result of migrating saved articles for mobile-html response and images.
+        //artifically returning a header here in this case, though maybe better would be to create a response header file during migration.
```
joewalsh: One advantage of that approach would be that you could use reading the mime type from the file as the check that the file exists before setting `isDownloaded == true`

WMF Framework/ArticleCacheController.swift
```diff
@@ -56,9 +64,51 @@ public final class ArticleCacheController: CacheController {
...
+    }
+    
+    private func cacheBundledResourcesIfNeeded(desktopArticleURL: URL, completionHandler: @escaping ((Error?) -> Void)) { //articleURL should be desktopURL
+        
+        guard let articleDBWriter = dbWriter as? ArticleCacheDBWriter else {
+            completionHandler(CacheFromMigrationError.invalidDBWriterType)
+            return
+        }
+        
+        //tonitodo: this will bundle those shared across siteURLs over and over again. need to bundle shared resources first as a separate thing, then bundle site-specific resource.
```
joewalsh: Might be able to skip the site specific CSS depending on how bad migrated articles look without it

Wikipedia/assets/pcs-html-converter/siteCSS.css
```diff
@@ -0,0 +1 @@
...e,.mw-parser-output .hlist li dt:first-child:before,.mw-parser-output .hlist li li:first-child:before{content:" (";font-weight:normal}.mw-parser-output .hlist dd dd:last-child:after,.mw-parser-output .hlist dd dt:last-child:after,.mw-parser-output .hlist dd li:last-child:after,.mw-parser-output .hlist dt dd:last-child:after,.mw-parser-output .hlist dt dt:last-child:after,.mw-parser-output .hlist dt li:last-child:after,.mw-parser-output .hlist li dd:last-child:after,.mw-parser-output .hlist li dt:last-child:after,.mw-parser-output .hlist li li:last-child:after{content:") ";font-weight:normal}.mw-parser-output .hlist ol{counter-reset:listitem}.mw-parser-output .hlist ol > li{counter-increment:listitem}.mw-parser-output .hlist ol > li:before{content:" " counter(listitem) " ";white-space:nowrap}.mw-parser-output .hlist dd ol > li:first-child:before,.mw-parser-output .hlist dt ol > li:first-child:before,.mw-parser-output .hlist li ol > li:first-child:before{content:" (" counter(listitem) " "}@media print{.navbox,.hatnote,.ambox{display:none}}.navigation-full-screen #mw-mf-page-left{width:100% !important;right:0 !important;opacity:1;display:block !important;visibility:visible !important}
```
joewalsh: This varies by site so we'd have to bundle one for every language. Same comment as above though, depending on how the resulting migration looks it might be OK without it

Wikipedia/Code/MWKDataStore.m
```diff
@@ -459,6 +457,12 @@ - (void)performUpdatesFromLibraryVersion:(NSUInteger)currentLibraryVersion inMan
             DDLogError(@"Error saving during migration: %@", migrationError);
             return;
         }
+        
+        [self moveImageControllerCacheFolderWithError:&migrationError];
```
tonisevener: @joewalsh - do we increment the library version for each type of migration, i.e. multiple per release? Or should we add this change and the standard user defaults change into version 9?

joewalsh: @tonisevener depends - usually if something has gone to public beta we'll increment the library version to ensure everyone's data gets updated. Since this is just in internal beta, I was OK with leaving it with the previous migration (and having internal beta testers re-install the app)

WMF Framework/ArticleCacheController.swift
tonisevener: @joewalsh this looked the same to me without it so I removed the site-specific CSS.

Wikipedia/Code/UIImageView+WMFImageFetchingInternal.m
```diff
@@ -174,7 +174,7 @@ - (void)wmf_setImage:(UIImage *)image
 }
 
 - (void)wmf_setImage:(UIImage *)image
-       animatedImage:(nullable FLAnimatedImage *)animatedImage
+       animatedImage:(FLAnimatedImage *)animatedImage
```
joewalsh: Should this and the one above still be marked nullable?

tonisevener: @joewalsh it probably doesn't hurt, I was undoing the work I did in https://github.com/wikimedia/wikipedia-ios/pull/3449/files since fetching the lead image is closer to the old way now.

PR #3461, opened 2020-03-04
mobile-html: Migrate article if load fails and migration is required

joewalsh:
- Migrate article if load fails and migration is required https://phabricator.wikimedia.org/T244621
- Ensure reload from pull to refresh doesn't use cached data https://phabricator.wikimedia.org/T242062

### DIFF
--- a/Wikipedia/Code/ArticleViewController.swift
+++ b/Wikipedia/Code/ArticleViewController.swift
@@ -294,16 +294,20 @@ class ArticleViewController: ViewController {
         }
     }
     
-    func loadPage() {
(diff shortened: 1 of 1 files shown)


tonisevener:
@joewalsh made a small tweak that seems to work after testing ~10 articles migrating in offline mode. I noticed didFailProvisionalNavigation gets called but didFailNavigation does not. Also the self.article.isConversionFromMobileViewNeeded flag was causing an early exit after successful migration.

joewalsh:
@tonisevener thanks for testing this and applying the fix!

PR #3464, opened 2020-03-05
Update migration assets

joewalsh:
Updates the migration code to the latest version

- Fixes a bug where image captions would have a lot of whitespace below
- Fixes edit pencils
- Fixes missing footer while offline
- Fixes migration on prod configuration


joewalsh:
Marking WIP because references are also broken on the resulting page - https://gerrit.wikimedia.org/r/c/mediawiki/services/mobileapps/+/577252 will fix and I'll update the assets again after that's merged

PR #3466, opened 2020-03-06
Saved articles fetcher error handling

joewalsh:
- Retry after `30 ^ retryCount` seconds, maxing out at four weeks
- Allow users to retry failed downloads manually, this will reset the backoff

First part of https://phabricator.wikimedia.org/T243573 (minus the housekeeping part)


Wikipedia/Code/SavedArticlesFetcher.swift
```diff
@@ -255,14 +261,54 @@ private extension SavedArticlesFetcher {
...
+            handleFailure(with: article, error: error)
         }
     }
     
-    func isErrorRecoverableOnRetry(_ error: Error) -> Bool {
-        // TODO: handle different errors differently - set isDownloaded = true on items that need to be skipped (errors that won't recover on retry)
-        return true
+    func handleFailure(with article: WMFArticle, error: Error) {
+        var underlyingError: Error = error
+        if let writerError = error as? ArticleCacheDBWriterError {
```
tonisevener: @joewalsh note there's also CacheFileWriter that could spit out errors too, if you wanted to save those specific errors. I think now it would probably default to .apiFailed if it's not specifically an NSFileWriteOutOfSpaceError, which could be enough.

joewalsh: @tonisevener thanks for the heads up! Seems like `.apiFailed` is enough for now but might be nice to split out in the future

PR #3467, opened 2020-03-06
T246722 - Cache new resources on saved article load

tonisevener:
https://phabricator.wikimedia.org/T246722

Also PR Feedback from https://github.com/wikimedia/wikipedia-ios/pull/3463 & using cache policies instead of forceCache & allowCache.


tonisevener:
@joewalsh removal of old urls is ready to be reviewed. Thanks!

PR #3469, opened 2020-03-10
Show empty view on article load error

joewalsh:
Match behavior of 6.5.1 and show the `. articleDidNotLoad` empty view when the article load fails

### DIFF
--- a/Wikipedia/Code/ArticleViewController.swift
+++ b/Wikipedia/Code/ArticleViewController.swift
@@ -758,9 +758,9 @@ private extension ArticleViewController {
             switch result {
             case .success(let user):
                 self.setupPageContentServiceJavaScriptInterface(with: user?.groups ?? [])
-            case .failure(let error):
+            case .failure:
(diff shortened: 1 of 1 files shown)


Wikipedia/Code/ArticleViewController.swift
```diff
@@ -816,28 +816,49 @@ extension ArticleViewController: ImageScaleTransitionProviding {
...
+        loadPage()
+    }
+}
+
+extension ArticleViewController: WKNavigationDelegate {
+    func webView(_ webView: WKWebView, didFailProvisionalNavigation navigation: WKNavigation!, withError error: Error) {
+        articleLoadDidFail(with: error)
+    }
+    
+    func webView(_ webView: WKWebView, didFail navigation: WKNavigation!, withError error: Error) {
```
joewalsh: Added this back in case some load errors manifest here instead of `didFailProvisionalNavigation`

PR #3472, opened 2020-03-11
Switch to NSSecureUnarchiveFromData for transformable properties on everything but WMFContentGroup and WMFContent

joewalsh:
This silences the warnings in the console but it's unclear why it's not complaining about `WMFContentGroup` and `WMFContent`. Those are harder to migrate because the transformable objects are `Mantle` models that don't conform to `NSSecureCoding`. Ideally, if those properties do need to be updated, they would get migrated to codable structs and we could drop `Mantle` entirely.

I verified that when installing from the app store, changing some preferences that affect `WMFLibraryValue`s, then running this new version the `WMFLibraryValue`s are preserved.

https://phabricator.wikimedia.org/T231987


WMF Framework/WMFContentGroup+Extensions.m
```diff
@@ -332,7 +332,7 @@ - (void)updateContentPreviewWithContent:(id)content {
         case WMFContentGroupKindNews:
         case WMFContentGroupKindUnknown:
         default: {
-            id<NSCoding> firstObject = contentArray.firstObject ?: content;
+            id<NSSecureCoding> firstObject = contentArray.firstObject ?: content;
```
tonisevener: Seems contentArray could contain non-conforming Mantle objects, WMFFeedTopReadArticlePreview for example.

WMF Framework/WMFContentGroup+Extensions.m
```diff
@@ -242,7 +242,7 @@ - (void)setSiteURL:(nullable NSURL *)siteURL {
     self.siteURLString = siteURL.wmf_databaseKey;
 }
 
-- (void)setFullContentObject:(NSObject<NSCoding> *)fullContentObject {
+- (void)setFullContentObject:(NSObject<NSSecureCoding> *)fullContentObject {
```
tonisevener: Here seems like NSArrays generally get passed in, but they could contain those Mantle objects within like WMFFeedOnThisDayEvent.

Wikipedia/Code/PlacesViewController.swift
```diff
@@ -1164,7 +1164,7 @@ class PlacesViewController: ViewController, UISearchBarDelegate, ArticlePopoverV
                 keyValue.key = search.key
                 keyValue.group = currentSearchHistoryGroup()
                 keyValue.date = Date()
-                keyValue.value = search.dictionaryValue as NSCoding
+                keyValue.value = search.dictionaryValue as NSSecureCoding
```
tonisevener: Also looks like this dictionaryValue could have an inner mantle object, MWKSearchResult.

WMF Framework/WMFContentGroup+Extensions.m
joewalsh: Good catch! This should still be NSCoding as the transformer hasn't been changed in the data model yet. I initially went down that path but didn't back out all the changes.

WMF Framework/WMFContentGroup+Extensions.m
joewalsh: Also a good catch, same issue as the earlier one

joewalsh:
Okay, looking closer, `NSSecureUnarchiveFromDataTransformer` is only available on iOS 12 and newer, so we should hold off on this work until we drop iOS 11. It seems like we should do at least one release of the app with `mobile-html` on iOS 11 to help with deprecating the `MobileView` API but looping in @jtminor to make the call.

PR #3478, opened 2020-03-11
Language fixes

tonisevener:
https://phabricator.wikimedia.org/T240017

- Now checking `"Vary":"Accept-Language"` in response header for populating variant in the DB.
- Also funneling new URLRequest generation methods through old Session method so default `accept-language` headers are populated.
- Added fix to update any variant of an article or image info request - it doesn't have to be that particular language variant. To test this piece: 

1. Prefer Chinese Simplified language in iOS Settings
2. Save zh.wiki article.
3. Go to airplane mode, change iOS Settings to Chinese Traditional. Load article, confirm old article version shows (fallback still works).
4. Leave airplane mode, load article, confirm you see traditional variant content.
5. Re-enter airplane mode, load article, confirm you see traditional variant content. Simplified was updated with Traditional content in step 4.

### DIFF
--- a/Wikipedia/Code/Session.swift
+++ b/Wikipedia/Code/Session.swift
@@ -238,8 +238,7 @@ import Foundation
                 }
             }
             
-            if let _ = error,
-            request.cachePolicy != .reloadIgnoringLocalCacheData {
+            if let _ = error {
                 
                 if let cachedResponse = self.defaultPermanentCache.cachedResponse(for: request) {
                     completionHandler(cachedResponse.data, cachedResponse.response, nil)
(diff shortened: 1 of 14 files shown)


Wikipedia/Code/Session.swift
```diff
@@ -238,8 +238,7 @@ import Foundation
                 }
             }
             
-            if let _ = error,
-            request.cachePolicy != .reloadIgnoringLocalCacheData {
```
tonisevener: This fixes the regression issue mentioned in https://github.com/wikimedia/wikipedia-ios/pull/3479.

Wikipedia/Code/Session.swift
```diff
@@ -667,8 +677,7 @@ class SessionDelegate: NSObject, URLSessionDelegate, URLSessionDataDelegate {
         if let error = error as NSError? {
             if error.domain != NSURLErrorDomain || error.code != NSURLErrorCancelled {
                 
-                if let request = task.currentRequest,
-                request.cachePolicy != .reloadIgnoringLocalCacheData,
```
tonisevener: Also fixes the regression issue mentioned in https://github.com/wikimedia/wikipedia-ios/pull/3479.

tonisevener:
> Looks good! It seemed like it was opportune to just pass the HTTPURLResponse down to save having to pull the MIMEType and variesOnLanguage separately. Also would be helpful if there's anything else from the response that we need to add to the `cacheItem` in the future. I did this in [24f1a10](https://github.com/wikimedia/wikipedia-ios/commit/24f1a108f46051a4db10a2b4da13423befe9fab1) but feel free to revert if it's not in line with what you were thinking or if I'm missing something

Good idea! Looks good to me 👍

PR #3479, opened 2020-03-11
Cache policy fixes

tonisevener:
Review https://github.com/wikimedia/wikipedia-ios/pull/3478 first.

PR for discussion - added a custom cache policy in  `a9a1365` &  `6b0b008` that is tacked onto the URLRequest header. This gives us the pull to refresh error banner for the user in airplane mode without the previous regression issue of all saved articles showing a failed state.

### DIFF
--- a/Wikipedia/Code/Session.swift
+++ b/Wikipedia/Code/Session.swift
@@ -1,7 +1,18 @@
 import Foundation
 
-public extension URLRequest.CachePolicy {
-    static let noPersistentCacheOnError: UInt = 99
+public enum WMFCachePolicy {
+    case foundation(URLRequest.CachePolicy)
+    case noPersistentCacheOnError
+    
+    var rawValue: UInt {
+        
+        switch self {
+        case .foundation(let cachePolicy):
+            return cachePolicy.rawValue
+        case .noPersistentCacheOnError:
+            return 99
+        }
+    }
 }
 
 @objc(WMFSession) public class Session: NSObject {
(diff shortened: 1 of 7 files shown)


Wikipedia/Code/Session.swift
```diff
@@ -1,7 +1,7 @@
 import Foundation
 
 public extension URLRequest.CachePolicy {
-    static let noReallyForceRemoteOnly: UInt = 99
+    static let noPersistentCacheOnError: UInt = 99
```
joewalsh: Would it help to have our own `CachePolicy` enum that has a case to wrap an underlying `URLRequest.CachePolicy`?
```
enum CachePolicy {
    case foundation(URLRequest.CachePolicy)
    case noPersistentCacheOnError
}
```

Wikipedia/Code/Session.swift
```diff
@@ -697,3 +699,15 @@ class SessionDelegate: NSObject, URLSessionDelegate, URLSessionDataDelegate {
         callback.success()
     }
 }
+
+fileprivate extension URLRequest {
+    var prefersPersistentCacheOverError: Bool {
+        if let customCachePolicyValue = allHTTPHeaderFields?[URLRequest.customCachePolicyHeaderKey],
```
joewalsh: Might be handy to implement the setter here too

Wikipedia/Code/Session.swift
tonisevener: @joewalsh agreed, that feels much cleaner. This is done!

Wikipedia/Code/Session.swift
tonisevener: 👍 Done!

PR #3482, opened 2020-03-13
Fix scroll to anchor for collapsed sections

joewalsh:
This is a two part client/server fix. `prepareForScrollToSections` needs to wait until the next paint to get the correct rect of the element.

[Server-side patch](https://gerrit.wikimedia.org/r/c/mediawiki/services/mobileapps/+/579537)

[T247434](https://phabricator.wikimedia.org/T247434)


Wikipedia/Code/ArticleWebMessagingController.swift
```diff
@@ -86,26 +86,18 @@ class ArticleWebMessagingController: NSObject {
         webView?.evaluateJavaScript(js)
     }
     
-    func prepareForScroll(to anchor: String, highlight: Bool, completion: @escaping (Result<CGRect, Error>) -> Void) {
+    func prepareForScroll(to anchor: String, highlight: Bool, completion: @escaping (Result<Bool, Error>) -> Void) {
```
joewalsh: This is just the callback for the initial JS execution, we don't get the actual value until the interaction handler event fires later. I initially tried `Result<Void, Error>` but the compiler wasn't happy - open to other suggestions here

tonisevener: I found https://stackoverflow.com/a/57005869, it seems like .success(()) works. I handled it this exact same way in other areas so I pushed a commit fixing these.

joewalsh: Awesome, thank you!

PR #3489, opened 2020-03-18
Re-implement article announcement functionality

joewalsh:
https://phabricator.wikimedia.org/T247781


Wikipedia/Code/ArticleViewController+Announcements.swift
```diff
@@ -0,0 +1,45 @@
...
+                try self.dataStore.viewContext.save()
+            } catch let saveError {
+                DDLogError("Error saving after marking article announcement as dismissed: \(saveError)")
+            }
+        }
+        let context = FeedFunnelContext(contentGroup)
+        FeedFunnel.shared.logFeedImpression(for: context)
+        wmf_showAnnouncementPanel(announcement: announcement, primaryButtonTapHandler: { (sender) in
+            self.navigate(to: actionURL, useSafari: true)
+            // dismiss handler is called
```
tonisevener: @joewalsh in `wmf_showAnnouncementPanel`, `dismissHandler` is only called in the footer link action, so I think you do need to explicitly call `dismiss()` here and in the secondary tap.

joewalsh: I noticed when debugging that these were called twice. I verified now that the other blocks call `self.dismiss()` and the `dismissHandler` is called on `viewDidDisappear`. Thanks for looking out for that though. I will add a comment in the other file as well.

joewalsh: Added in the next PR - 802a56bb1

PR #3497, opened 2020-03-25
Comment out legacy article data deletion after conversion

tonisevener:


Wikipedia/Code/MWKDataStore+LegacyMobileview.swift
```diff
@@ -32,20 +32,22 @@ extension MWKDataStore {
...
         mobileviewConverter.convertMobileviewSavedDataToMobileHTML(articleURL: articleURL, article: legacyArticle) { (result, error) in
-            let blastMobileviewSavedDataFolder = {
-                // Remove old mobileview saved data folder for this article
-                do {
-                    try FileManager.default.removeItem(atPath: self.path(forArticleURL: articleURL))
-                } catch {
-                    DDLogError("Could not remove mobileview folder for articleURL: \(articleURL)")
-                }
-            }
+            //TODO: uncomment after 6.6 Beta releases if we decide to delete old data incrementally
```
tonisevener: I still think it would be worth keeping the incremental deletion when we re-enable, in case a user is on the cusp of maxing out their phone storage.

PR #3502, opened 2020-03-27
Debug helper: Perform full refresh on pull-to-refresh for debug builds

joewalsh:
Helps with testing locally against the Page Content Service - all css and js are reloaded. For release builds, we keep the same behavior of reloading with a different cache policy.


mcleinman:
Thank you for this!

PR #3503, opened 2020-03-30
add missing completion handler check to prevent double calls

tonisevener:
Fixes crash at these steps:

1. Load article
2. Force press an image later in the article
3. App crashes


joewalsh:
Good catch!

PR #3505, opened 2020-03-31
Fix bottom "view in browser" link, and remove unneeded license code

mcleinman:
Part of bug https://phabricator.wikimedia.org/T248511

Fixes bottom "view in browser" link to again work, and remove unneeded license code

This was falling through in part because it ended up in the `default` clause of the switch statement. Since the `switch` is on an enum, I made it exhaustive so we can remove the `default` statement. (And thus, if in the future we add a new action to the enum, the app won't compile unless we handle it in this block of code.) To do this, I found the final two cases that were not being handled and added them - `.properties` and `.readMoreTitlesRetrieved`. I don't 100% know what these two actions should represent - **is it possible we also have a bug with them, in the same way we did with "view in browser" and the CC license?**

_While this is a partner PR of [this PCS PR](https://gerrit.wikimedia.org/r/c/mediawiki/services/mobileapps/+/584742), the two PRs can be merged (and go live) in either order. I tested it out, and as expected there are no crashes no matter which one goes live first._


joewalsh:
> To do this, I found the final two cases that were not being handled and added them - .properties and .readMoreTitlesRetrieved. I don't 100% know what these two actions should represent - is it possible we also have a bug with them, in the same way we did with "view in browser" and the CC license?

Thanks for double checking about these! The `.properties` event isn't sent by PCS and can be removed. The `.readMoreTitlesRetrieved` event is sent by PCS but there's no use for it at the moment. Previously, there were save buttons on the read more items, so we'd have to check if the user had those article saved, and update the state of the saved button to reflect that. Since those save buttons were removed, it can also be removed. Android isn't using it either, so I'll file a task to remove it from PCS.

tonisevener:
👍🎉

mcleinman:
I also stopped previews for "view article in browser" as it was showing a preview for the article "N/A", though it comes with the tradeoff of not showing a preview for the actual article "N/A" if that is the intention.

PR #3507, opened 2020-03-31
Clean up headers being sent to the Page Content Service

joewalsh:
- Update the request for mobile-html to send `text/html` instead of `application/json` for `Accept`. Should the server ever respect this header, the WebView would have a bad time with JSON.
- Remove the summary endpoint profile. This doesn't seem to be respected by RESTBase and should only apply to the summary endpoint, not all RESTBase requests.
- Consolidate and rename methods for clarity


joewalsh:
@tonisevener I updated the other requests I was seeing with application/json - it looked like the css and js requests through the SchemeHandler were going through with the right headers but let me know if you're seeing something different. Thanks!

Wikipedia/Code/ArticleCacheDBWriter.swift
```diff
@@ -68,7 +68,9 @@ final class ArticleCacheDBWriter: ArticleCacheResourceDBWriting {
                 
                 //append mobile-html-offline-resource URLRequests
                 for url in urls.offlineResourcesURLs {
-                    guard let urlRequest = self.articleFetcher.urlRequest(from: url) else {
+                    // We don't know or care what Content-Type these are
+                    let acceptWhateverHeaders = ["Accept": "*/*"]
```
staykids, line 72: Super minor: may be worth naming this variable either `acceptAnyType` (to match the [specific directive](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept) itself) or just `headers` (to [match here](https://github.com/wikimedia/wikipedia-ios/blob/d416b8378e0977c95b1d643ee2dd00a09f99f34a/WMF%20Framework/ImageCacheController.swift#L99) and elsewhere in this PR) for clarity and consistency.

joewalsh: Good idea, done! It might be worth consolidating these eventually, but for now hopefully the naming clarification helps.

PR #3509, opened 2020-03-31
Article performance tests

tonisevener:
https://phabricator.wikimedia.org/T247558

I'm a bit stuck on how to finish these up so I'm putting this up to gather feedback and thoughts while I move onto something else. Some questions & thoughts I have are:

1. Both tests fluctuate quite a bit which cause failures. `testArticleSetupTime` always has a large first run hit compared to the others which I haven't found a workaround for yet. At this point it seems too flaky to commit.
2. If we decide to hit the network as a way of testing the server load time, we should probably move this into beta cluster tests so these aren't constantly executed in Prod. Also would give us an opportunity to catch server-side slowdowns before they move to Prod.
3. On that note, should we even be hitting the server from app-side tests? I could stub calls out with Nocilla and only test the app code performance.
4. Testing 3D touch loading could be more fleshed out, as this part only tests the menu config piece and not the actual display of the summary on screen. It also will always max out 1 second due to the timeout completion handling so I'm not sure this would detect a big jump anyways.
5. Worst case scenario if these can't be more consistently automated, maybe we keep them around as a separate set of tests that we occasionally run manually as a part of the standard release performance ticket Phab task.


joewalsh:
This is great! Answers to questions inline below

>If we decide to hit the network as a way of testing the server load time, we should probably move this into beta cluster tests so these aren't constantly executed in Prod. Also would give us an opportunity to catch server-side slowdowns before they move to Prod.

I agree, anything that hits the network or depends on services out of our control should be in the beta cluster tests. Fortunately, I don't think we need those tests in this case. The Page Content Service has [performance tests](https://github.com/wikimedia/mobileapps/blob/master/test/lib/mobile/MobileHTML.test.js#L11) that should catch processing time regressions before they're merged to master. It'd be worth expanding those tests with more cases, currently the only fixture is the enwiki United States article. 

We're usually not dependent on that processing time on prod since the output is cached in RESTBase until the article changes or we manually clear the cache. The overall response time is [monitored in grafana](https://grafana.wikimedia.org/d/000000183/mobileapps?orgId=1), so we should be able to catch slowdowns there.

>On that note, should we even be hitting the server from app-side tests? I could stub calls out with Nocilla and only test the app code performance.

I'm in favor of removing the network from the equation for our tests for the reasons mentioned above. We would need to update the fixtures periodically so it's worth investing some time into making sure that's an easy-ish process. A fastlane lane similar to [record_visual_tests](https://github.com/wikimedia/wikipedia-ios/blob/develop/fastlane/Fastfile#L122) would be ideal.

tonisevener:
This has been folded into https://github.com/wikimedia/wikipedia-ios/pull/3549 and cleaned up.

PR #3515, opened 2020-04-03
Fix resolution of relative hrefs for titles with forward slashes

joewalsh:
Fixes [T249117](https://phabricator.wikimedia.org/T249117)

When attempting to resolve a relative URL on a page that has a title with forward slashes, `URLComponents` would treat the forward slash like it should, which is as a path component. Since we know that `articleURL` is an `articleURL`, we can percent encode everything after `/wiki/` and utilize that to resolve the relative href.

### DIFF
--- a/Wikipedia/Code/ArticleWebMessagingController.swift
+++ b/Wikipedia/Code/ArticleWebMessagingController.swift
@@ -40,7 +40,7 @@ class ArticleWebMessagingController: NSObject {
     }
     
     func addFooter(articleURL: URL, restAPIBaseURL: URL, menuItems: [PageContentService.Footer.Menu.Item], lastModified: Date?) {
-        guard let title = articleURL.wmf_title else {
+        guard let title = articleURL.percentEncodedPageTitleForPathComponents else {
             return
         }
         var editedDaysAgo: Int?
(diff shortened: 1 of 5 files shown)


Wikipedia/Code/ArticleWebMessagingController.swift
```diff
@@ -40,7 +40,7 @@ class ArticleWebMessagingController: NSObject {
     }
     
     func addFooter(articleURL: URL, restAPIBaseURL: URL, menuItems: [PageContentService.Footer.Menu.Item], lastModified: Date?) {
-        guard let title = articleURL.wmf_title else {
+        guard let title = articleURL.percentEncodedPageTitleForPathComponents else {
```
tonisevener, line 43: @joewalsh separate bug but may be related enough to fix in this PR - I notice footers are broken on articles with a '/' in the title. I see it in develop and this branch. I can make a separate task if that's preferable.

joewalsh: @tonisevener 🤔I saw that as well and thought I verified that this change fixed it, but I'll double check. Which article are you using for testing?

joewalsh: Looks like it's working for me on this branch with enwiki articles /dev/null, /Drive, and W:/2016Album/

tonisevener: @joewalsh I'm seeing a missing footer on G/O Media.

PR #3518, opened 2020-04-06
Display native survey prompt

tonisevener:
https://phabricator.wikimedia.org/T248510

Fetching and displaying new announcement in wikifeeds patch [here](https://gerrit.wikimedia.org/r/#/c/mediawiki/services/wikifeeds/+/586472/).

**General approach:**
We are leaning on the announcements endpoint and `WMFAnnouncementsContentSource.m` (which typically persists to Core Data), as well as the `AnnouncementPanelViewController` to display this survey prompt. Requirements for this are to not persist and only consider in online mode, so I am cleaning out this particular survey announcement in the same [place](https://github.com/wikimedia/wikipedia-ios/pull/3518/files#diff-792d3594554b3be9fe4ab2ec4b45e34cR135) where we are cleaning out the fundraising 2019 announcements, after saving the announcements in-memory via `SurveyAnnouncementsController`. I also had to do a few tweaks to the UI so the panel would look okay with this minimal amount of information and add the ability to know what action caused an announcement to dismiss for the `UserDefaults` flag persistence.

**To test:**
You'll need to pull the wikifeeds patch [here](https://gerrit.wikimedia.org/r/#/c/mediawiki/services/wikifeeds/+/586472/). Start wikifeeds and pcs locally, then use (local pcs & announcements) scheme when running. If you don't want to run pcs locally, you can remove -DWMF_LOCAL_PAGE_CONTENT_SERVICE from the LocalDebug build configuration under Custom Flags in the Build Settings.


WMF Framework/SurveyAnnouncementsController.swift
```diff
@@ -0,0 +1,77 @@
...
+            let nonTemplateActionURLString = announcement.actionURLString?.replacingOccurrences(of: "{{articleTitle}}", with: parameterizedArticleTitle)
+            guard let actionURLString = nonTemplateActionURLString,
+            let actionURL = URL(string: actionURLString) else {
+                continue
+            }
+                
+            let now = Date()
+            
+            //do not show if user has already seen and answered for this campaign
+            if let _ = UserDefaults.standard.object(forKey: identifier) as? Bool {
```
staykids, line 62: Minor – feel free to ignore. Instead of the unused result assignment, could use `guard` like the rest of function does. Makes the control flow through the function slightly easier to reason.  

```swift
guard UserDefaults.standard.object(forKey: identifier) == nil else {
    continue
}
```

tonisevener: @staykids agreed, done!

tonisevener:
This is ready now for final review.

tonisevener:
Added one bug fix to this in [b16f52e](https://github.com/wikimedia/wikipedia-ios/pull/3518/commits/b16f52e8db9c0e9486fe450e5bbcc1a387c178d3)  - it wasn't working for '2019–20 coronavirus pandemic' without this.

PR #3521, opened 2020-04-08
Run response header save on background thread

tonisevener:
https://phabricator.wikimedia.org/T249764

I couldn't reproduce this but there's a complaint of a crash on launch in Testflight. Looking at the stack trace it seems like saving the response header for images in CacheItemMigrationPolicy took too long, and the system kicked in and terminated the app. Going to try moving this piece to a background thread.


WMF Framework/CacheItemMigrationPolicy.swift
```diff
@@ -27,25 +27,25 @@ class CacheItemMigrationPolicy: NSEntityMigrationPolicy {
...
-                        ]
-                    CacheFileWriterHelper.saveResponseHeader(headerFields: headerFields, toNewFileName: headerFileName) { (result) in
-                        switch result {
-                        case .success, .exists:
-                            isDownloaded = true
-                        case .failure:
-                            break
+                //artifically create and save image response header
+                //perform on another thread in case this takes a while
+                DispatchQueue.global(qos: .background).async {
```
joewalsh, line 39: This could perform too many file operations concurrently since there's not a good built-in concurrency limit on dispatch queues. Also, the db migration itself happening on the main thread could block long enough to cause the system to kill the app. I committed an update that moves the whole migration to a background queue, but feel free to revert if I'm missing something.

Wikipedia/Code/MWKDataStore.m
```diff
@@ -453,29 +453,34 @@ - (void)performUpdatesFromLibraryVersion:(NSUInteger)currentLibraryVersion inMan
...
     if (currentUserLibraryVersion >= WMFCurrentLibraryVersion) {
-        if (completion) {
-            completion();
-        }
+        combinedCompletion();
         return;
     }
     [self performBackgroundCoreDataOperationOnATemporaryContext:^(NSManagedObjectContext *moc) {
-        dispatch_block_t done = ^{
-            dispatch_async(dispatch_get_main_queue(), completion);
```
joewalsh, line 475: This is intentionally removed - the caller has an async to the main queue in the completion block it passes.

WMF Framework/CacheItemMigrationPolicy.swift
tonisevener: @joewalsh thanks for catching these! Changes look good to me 👍

PR #3522, opened 2020-04-08
Remove mimeType logic

tonisevener:
https://phabricator.wikimedia.org/T247856

Since we are saving the entire response header, we no longer need to save mimeType separately. This will also get around these assertion failures we were seeing during large syncs.


joewalsh:
Looks good! I added some documentation to indicate that the remaining extended file attribute functionality is deprecated and can be removed when the migration is removed.

PR #3527, opened 2020-04-10
Change order of VoiceOver reading in edit mode

mcleinman:
Phab: T249354

This is somewhat heavy handed for a slight change - it reads the bulk edit option buttons *before* the actual list of items. I'm open to the idea that the slight improvement isn't worth this hacky code. Or suggestions on how to improve this.


staykids:
Minor coding style note – may be worth removing the explicit `self.` references here and instead just rely on Swift's implicit use of `self`. So for example `self.accessibilityElements = []` → `accessibilityElements = []` etc.

mcleinman:
> Minor coding style note – may be worth removing the explicit `self.` references here and instead just rely on Swift's implicit use of `self`. So for example `self.accessibilityElements = []` → `accessibilityElements = []` etc.

Happy to have my personal style beaten out of me on this one... 😄 I personally appreciate being explicit - I like when skimming code to see what is scoped more locally. That said, I'm happy to get on board if our team's convention is to drop the `self`.

staykids:
I agree re: team consensus – I totally defer to Toni and Joe + the team as a whole as to what our convention here should be!

To offer my two cents – given that this project's roots are in Objective-C, I can understand the mixed usage through the project. In my opinion though, it would be valuable if our new Swift code was as idiomatic to the language as possible (same for any new Objective-C code). Having consistent, cohesive, and idiomatic implementations could also potentially help encourage other new contributors to the project. 

https://github.com/wikimedia/wikipedia-ios/blob/3f90b3a0af247a8f36da554a1e283c3eb92c543a/Wikipedia/Code/SavedViewController.swift#L304
In the specific case of this function `didChangeEditingState`, these new occurrences of explicit `self` are inconsistent with the rest of the function itself (e.g. `navigationItem` above the changes is implicit) and the rest of [SavedViewController.swift](https://github.com/wikimedia/wikipedia-ios/blob/3f90b3a0af247a8f36da554a1e283c3eb92c543a/Wikipedia/Code/SavedViewController.swift) (with the exception of [self.theme](https://github.com/wikimedia/wikipedia-ios/blob/3f90b3a0af247a8f36da554a1e283c3eb92c543a/Wikipedia/Code/SavedViewController.swift#L301)).

I'd be interested to hear more of your perspective! In my experience on this topic, being explicit with `self` when not required has often led to a reduction in quick parsability by cost of increase in text noise. https://docs.swift.org/swift-book/LanguageGuide/Methods.html under `The self Property` section provides a little background. To extend on their example, I would also say that in addition to only using `self` when explicitly disambiguating, it also helps quickly identify retaining closures. So to some extent, writing idiomatic Swift and relying on `self` only when explicitly needed can be also be used kind of like a quick visual leak/code smell check. Kinda cool.

(Sorry for hijacking your PR 🤪 - thanks for letting me write up my thoughts about this)

mcleinman:
Looks like we're also currently chatting about this elsewhere (😄) but in short: I'd agree that I find it slightly easier to parse areas of a codebase _that I know well_ without `self`. (That said, I'm pretty good at ignoring `self` syntactic sugar when reading code, but I may have overdeveloped that skill because I'm used to writing `self` as well.) When I find `self` especially helpful to read is when I'm skimming part of a codebase that I'm not that familiar with - especially if it's in larger functions / blocks (the ones that should actually probably actually be broken down into multiple functions) or within blocks with lots of variables. I find it easier to parse where various variables are coming from if we're using `self` - rather than seeing if they were declared within that block of code.

Wikipedia/Code/SavedViewController.swift
```diff
@@ -310,6 +310,14 @@ extension SavedViewController: CollectionViewEditControllerNavigationDelegate {
         let editingStates: [EditingState] = [.swiping, .open, .editing]
         let isEditing = editingStates.contains(newEditingState)
         actionButton.isEnabled = !isEditing
+        if (newEditingState == .open),
+            let batchEditToolbar = self.savedArticlesViewController?.editController.batchEditToolbarView,
+            let contentView = self.containerView,
+            let appTabBar = self.parent?.view.subviews.first(where: {($0 as? UITabBar) != nil}) {
```
tonisevener, line 316: For this one, I would rather reference `appTabBar` a bit more strongly. Maybe via a delegate that has a `tabBar` property that `AppViewController` conforms to. Reason being should we ever present `SavedViewController` in a different context, say without a parent that has a tab bar, this would fail silently. Whereas if `SavedViewController` must be instantiated with a delegate that has a tab bar, the developer would know right away that something is up, i.e. this VC has an assumption that it can point to a tab bar.

mcleinman: That sounds much better to me as well. 

As I alluded to in the PR, do you think adding a new delegate to the codebase and setting it here is worth the minor improvement in VoiceOver support? This PR allows the edit view of the saved article VC to be read in a more logical order by VoiceOver (when swiping through items). If we can't do it easily, should we just close it out and leave it as is? I'm not sure how much code this minor improvement is worth. My inclination is that if we can't do it easily, it's not worth doing. But I'm very open to other thoughts around importance.

tonisevener: I say forge ahead, I think a slight VoiceOver improvement is always better than nothing and I don't think the code changes will be too heavy at this point. If it feels like it's sending you down a big rabbit hole we can talk about maybe punting until after the audit since that'll probably lead to additional changes anyways.

mcleinman: Sounds good. I updated this. I put the protocol in a new file - let me know if we should tuck it elsewhere instead. Thanks!

Wikipedia/Code/TabBarDelegate.swift
```diff
@@ -0,0 +1,13 @@
+//
+//  TabBarDelegate.swift
+//  Wikipedia
+//
+//  Created by Matt Cl on 4/13/20.
+//  Copyright © 2020 Wikimedia Foundation. All rights reserved.
+//
```
tonisevener, line 7: We take off our headings for new files in this project. I've just developed the habit of removing them each time but I think @joewalsh had found an automated way a while back. I also found [this](https://oleb.net/blog/2017/07/xcode-9-text-macros/) but it results in a single '//' header, so not ideal but better than nothing.

Wikipedia/Code/TabBarDelegate.swift
```diff
@@ -0,0 +1,13 @@
...
+
+import Foundation
+
+@objc public protocol AppTabBarDelegate {
```
tonisevener, line 11: I know the codebase isn't consistent with this but it's good to add a WMF prefix for obj-c:

```
@objc(WMFAppTabBarDelegate)
protocol AppTabBarDelegate {
```

Also I think you can remove the `public` and I would rename the file to AppTabBarDelegate.swift.

mcleinman: Thanks for those - they've now been updated. As you could probably guess, I renamed the class after I made the file and forgot to update it. Thanks for catching that.

tonisevener: I do it all the time 😁 LGTM, thanks!

PR #3530, opened 2020-04-14
add template macro for empty header on new files

tonisevener:
Reduces default header down to `//`. Throwing this up here so we can test & decide as a team if we want it.


mcleinman:
If we can't find a way to actually remove it, could we make that line `// Please manually remove this line from new files. (And let us know if you find a way to automate the process!).`? That way we're probably more likely to remember to remove it.

(I'd even be happy with an addendum of `If this line is found in a PR, the submitter owes you a drink.`, but that is probably too much for something we're committing to the project. 😄 )

staykids:
I don't have deep experience with git hooks (or `IDETemplateMacros.plist`), but I wonder if this paired with some kind of pre-commit git hook could solve this issue for us. If the header was something like `// Wikimedia Foundation` and the pre-commit script stripped whatever that particular unique string is, maybe it'd work generally pretty safely?

tonisevener:
@staykids @mcleinman thanks for the feedback! yeah I like the idea of some sort of unique header, then strip it out with a git hook. Will try that when I get a chance.

tonisevener:
I ended up doing a bit of a hybrid of both suggestions. Template has been changed to `~~~**DELETE THIS HEADER**~~~` and added a pre-commit hook to fail if it detects this string in any file. Optimization would be if it detects it in only the staged files versus the whole project directory, but this doesn't seem to take too long to run on my machine. I think it might be possible to actually change the file but I wasn't sure if it was worth spending the extra time getting that right at this point.

You will need to re-run the `scripts/setup` script from the project directory to to get the new pre-commit scripts copied over, then you can test by adding a new file and trying to commit with the header.

Screen Shot 2020-05-14 at 12 35 30 PM

scripts/setup_git_hooks
```diff
@@ -1,8 +1,6 @@
 #!/bin/bash
 # sets up pre commit hooks
```
tonisevener, line 2: It seems like this script was never used - not sure why - but now it is.

staykids: One sort of related thing with `clang_format_git_diff` now being used in `setup_git_hooks`: if the engineer's Homebrew configuration is non-standard (such that it isn't symlink installations into `/usr/local/bin`), `clang_format_git_diff` will fail unexpectedly. Frankly I'm not even positive if that's a configurable Homebrew thing or not, but might be safer to replace `/usr/local/bin/clang-format` with just `clang-format` as the `clang_format_all` script does.

PR #3531, opened 2020-04-15
Restore ability to edit title descriptions

joewalsh:
- Ensure the article's associated Wikidata ID is updated when the article summary is fetched. This occurs when the article content is loaded.
- Only show title description editor if there's an associated Wikidata ID - otherwise we have no place to put the description

https://phabricator.wikimedia.org/T250269

### DIFF
--- a/Wikipedia/Code/DescriptionEditViewController.swift
+++ b/Wikipedia/Code/DescriptionEditViewController.swift
@@ -32,14 +32,16 @@ import WMF
     // These would be better as let's and a required initializer but it's not an opportune time to ditch the storyboard
     // Convert these to non-force unwrapped if there's some way to ditch the storyboard or provide an initializer with the storyboard
     var article: WMFArticle!
+    var wikidataID: String!
     var articleURL: URL!
     var descriptionSource: ArticleDescriptionSource!
     var isAddingNewTitleDescription: Bool!
     var dataStore: MWKDataStore!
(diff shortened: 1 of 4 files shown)


Wikipedia/Code/DescriptionEditViewController.swift
```diff
@@ -32,14 +32,16 @@ import WMF
     // These would be better as let's and a required initializer but it's not an opportune time to ditch the storyboard
     // Convert these to non-force unwrapped if there's some way to ditch the storyboard or provide an initializer with the storyboard
     var article: WMFArticle!
+    var wikidataID: String!
     var articleURL: URL!
     var descriptionSource: ArticleDescriptionSource!
     var isAddingNewTitleDescription: Bool!
     var dataStore: MWKDataStore!
-    static func with(articleURL: URL, article: WMFArticle, descriptionSource: ArticleDescriptionSource, dataStore: MWKDataStore, theme: Theme) -> DescriptionEditViewController {
+    static func with(articleURL: URL, wikidataID: String, article: WMFArticle, descriptionSource: ArticleDescriptionSource, dataStore: MWKDataStore, theme: Theme) -> DescriptionEditViewController {
```
joewalsh, line 40: I opted to make `wikidataID` separate in this initializer instead of using the property on `article` to ensure the VC has everything it needs to make the edit at initialization time and to allow callers to ensure they have all the required info before initializing. Alternatively, this could check for `article.wikidataID` and throw an error, open to either approach.

PR #3536, opened 2020-04-16
Address visual issues with article pull to refresh

staykids:
https://phabricator.wikimedia.org/T249626

It was somewhat intentional to keep this a band-aid rather than truly address the underlying issue (to avoid more [unintended consequences](https://github.com/wikimedia/wikipedia-ios/pull/3526)  with the app's most critical feature). Given the timebox, this is the cleanest solution I've come up with so far that addresses the issue 90%* of the time. There's thankfully no gross additional `UIScrollView` as I mentioned in one possible solution, but to be clear, there's still some gross stuff here 🙃.

From my perspective so far, this bug stems from two related items: 
1. `WKWebView` itself doesn't play nicely with a `UIRefreshControl` on its `UIScrollView`
2. Our additional manipulation of `WKWebView`'s expectations of that scroll view probably doesn't help

Areas I'm still investigating:
* *the last 10% - there are still times when I can temporarily coax it into a jittery state. I have a few hunches. 
* Confirm this still plays nice under offline device conditions.

Feel free to let me know if it seems like I'm missing anything obvious!


Wikipedia/Code/ArticleViewController.swift
```diff
@@ -786,6 +819,14 @@ private extension ArticleViewController {
...
         webView.uiDelegate = self
         webView.navigationDelegate = self
+
+        webView.addSubview(refreshOverlay)
+        NSLayoutConstraint.activate([
+            refreshOverlay.topAnchor.constraint(equalTo: webView.topAnchor),
+            refreshOverlay.bottomAnchor.constraint(equalTo: webView.bottomAnchor),
+            refreshOverlay.leftAnchor.constraint(equalTo: webView.leftAnchor),
+            refreshOverlay.rightAnchor.constraint(equalTo: webView.rightAnchor),
+        ])
```
joewalsh, line 829: There's a `wmf_addSubviewWithConstraintsToEdges` extension on `UIView` that might be helpful here

staykids: 👍🏽 thanks, will update.

tonisevener:
Looking much better! Just wanted to mention since I'm not sure that this requirement is documented anywhere - in offline mode we currently show a "No internet connection" banner when they pull to refresh. I think it'd be nice to preserve that so the user doesn't mistakenly think they are seeing the latest.

out

staykids:
> Looking much better! Just wanted to mention since I'm not sure that this requirement is documented anywhere - in offline mode we currently show a "No internet connection" banner when they pull to refresh. I think it'd be nice to preserve that so the user doesn't mistakenly think they are seeing the latest.

@tonisevener 👍🏽. Do you happen to have rough repro steps to expose that "No internet connection " banner? Whether exploring an article or viewing a saved one, signed in or not, I'm not able to repro that banner on the current `master`'s HEAD when pulling to refresh.

staykids:
Alright – may this PR forever improve the article refreshing experience at least around 90% of the time!

I'll document the other improvement/exploration areas on the ticket and open a new branch/PR for that continued work.

PR #3537, opened 2020-04-17
Show loading animation if library migration takes too long

joewalsh:
https://phabricator.wikimedia.org/T249865

- Restore loading animation VC from article loading work
- Move splash screen transition into a separate VC
- Transition from the splash screen to the loading animation if loading takes longer than a given duration (currently 1s)
- Ensure the loading animation shows for a minimum duration (currently .6 seconds, .3 seconds of that are it crossfading in)

https://phabricator.wikimedia.org/T249865


Wikipedia/Code/WMFThemeableNavigationController.m
```diff
@@ -65,53 +65,37 @@ - (void)applyTheme:(WMFTheme *)theme {
...
-
-- (void)showSplashViewIfNotShowing {
-    if (!self.splashView.hidden) {
+    if (self.splashScreenViewController) {
         return;
     }
-    [self showSplashView];
+    WMFSplashScreenViewController *splashVC = [[WMFSplashScreenViewController alloc] init];
+    // Explicit appearance transitions need to be used here because UINavigationController overrides
+    // a lot of behaviors when adding the VC as a child and causes layout issues for our use case.
```
tonisevener, line 74: Hmm not really sold on this, feels like a sign that we shouldn't bother if nav view controller doesn't play nicely with a child view controller. It also feels weird for all WMFThemeableNavigationControllers to have the ability to show/hide a splash view if only one instance is ever going to use it. I know AppVC is huge but still seems like something that only it should have the logic to do. I'm open to leaving as-is since that's existing but wanted to say my two cents.

Wikipedia/Code/SplashScreenViewController.swift
```diff
@@ -0,0 +1,99 @@
...
+    
+    func setupSplashView() {
+        view.wmf_addSubviewWithConstraintsToEdges(splashView)
+        let wordmark = UIImage(named: "splashscreen-wordmark")
+        let wordmarkView = UIImageView(image: wordmark)
+        wordmarkView.translatesAutoresizingMaskIntoConstraints = false
+        splashView.addSubview(wordmarkView)
+        let centerXConstraint = splashView.centerXAnchor.constraint(equalTo: wordmarkView.centerXAnchor)
+        let centerYConstraint = splashView.centerYAnchor.constraint(equalTo: wordmarkView.centerYAnchor, constant: 12)
+        splashView.addConstraints([centerXConstraint, centerYConstraint].compactMap { $0 })
```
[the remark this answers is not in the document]
joewalsh: Nope, fixed

Wikipedia/Code/WMFThemeableNavigationController.m
joewalsh: >Hmm not really sold on this, feels like a sign that we shouldn't bother if nav view controller doesn't play nicely with a child view controller. 

It seems the built-in parent/child VC methods are more for convenience rather than a requirement. Given these appearance transition methods aren't deprecated, I feel OK about using them.

>  It also feels weird for all WMFThemeableNavigationControllers to have the ability to show/hide a splash view if only one instance is ever going to use it. 

Would it help to move this functionality to a separate `RootNavigationController` class?

> I know AppVC is huge but still seems like something that only it should have the logic to do. I'm open to leaving as-is since that's existing but wanted to say my two cents.

Unfortunately AppVC is poorly named as it isn't the root VC of the app. It's the root VC of a `WMFThemeableNavigationController` that is the root VC of the app. As a result, the appVC would have to present the splash screen modally off of its navigation controller anyway to cover any pushed VCs. I tried this approach, but the transition wasn't smooth, it didn't seem like we could control the cross fade duration, and it potentially conflicts with the modal presentation of onboarding.

tonisevener: > Would it help to move this functionality to a separate RootNavigationController class?

That sounds good, I think that would cover it all for me. And good point on those methods not being deprecated so should be fine. On your modal point, I can see how that would mess things up. I was initially more thinking have the view still live on the navigationController's view, but AppVC owns the logic that adds the splash view, swaps it out with the loading animation, etc. But since loading animation is already in a view controller it'd be more rework and now that I'm typing out-loud feels like moving backwards to cram more stuff into AppVC.

joewalsh: I moved the functionality into a separate `RootNavigationController`. Long term it'd be good to explore turning that into a plain view controller that has the `WMFThemeableNavigationController` and the `SplashScreenViewController` as a child view controllers, but for now given we're late in this release it makes sense to keep it as close to the old VC structure as we can.

tonisevener:
@joewalsh looking good! only thing I can think of is that the loadingVC should be themed, from what I see if the system is in dark mode it still shows it as light. I tried doing it in https://github.com/wikimedia/wikipedia-ios/commit/9337387d30a0cddf16f7cd44cedc4ef0ea9b48a6 but I didn't like it enough to commit on this PR.

joewalsh:
> @joewalsh looking good! only thing I can think of is that the loadingVC should be themed, from what I see if the system is in dark mode it still shows it as light. I tried doing it in 9337387 but I didn't like it enough to commit on this PR.

Good catch! I brought it in with a minor adjustment to utilize the theme from the navigation controller

tonisevener:
@joewalsh added one tweak - the animation would freeze when returning from background. I made a tweak to start it up again on `didBecomeActive`.

PR #3538, opened 2020-04-17
Fix CI issues, rename builds to match new conventions, update documentation

joewalsh:
- Previously, we had separate git tag prefixes for `beta`, `beta_app`, and `alpha`. Since we'll be rolling the staging and prod apps now at the same time, only make one tag per release with the `betas/` tag prefix.
- Updates the target names to match the new conventions. Beta -> Staging, Alpha -> Experimental
- Adds documentation about different app targets
- Removes unused targets, info.plist files, build settings, and fastlane lanes


fastlane/Fastfile
```diff
@@ -221,67 +213,52 @@ platform :ios do
     sh "xcodebuild -exportArchive -exportOptionsPlist ExportOptions.plist -archivePath \"#{build_dir}/#{product_name}.xcarchive\" -exportPath \"#{build_dir}\" -allowProvisioningUpdates"
   end
   
-  desc "Deploys both the production and beta apps"
+  desc "Pushes both the production and staging apps to test flight and tags the release. Only releases to internal testers."
```
staykids, line 216: Minor: "test flight" →"TestFlight"

fastlane/README.md
```diff
@@ -105,62 +100,42 @@ Build the app for distibution
 ```
 fastlane ios deploy
 ```
-Deploys both the production and beta apps
-### ios push_beta
-```
-fastlane ios push_beta
+Pushes both the production and staging apps to test flight and tags the release. Only releases to internal testers.
```
staykids, line 103: Minor: "test flight" →"TestFlight"

README.md
```diff
@@ -43,6 +43,36 @@ Covered in the [process document](docs/process.md).
 ### Testing
 The **Wikipedia** scheme is configured to execute the project's iOS unit tests, which can be run using the `Cmd+U` hotkey or the **Product->Test** menu bar action. Screenshot tests will fail unless you are running on one of the configurations defined by `configurations_to_test_on_pull` in `fastlane/Fastfile`.
 
+### Targets
+#### Wikipedia
+Points to production servers.
+#### Staging and Wikipedia (apps staging environment)
```
tonisevener, line 49: Rename to just Staging to match scheme name?

tonisevener: Actually nevermind, I thought it was referencing one scheme but I see now that it's both.

joewalsh: It seems like now we can get rid of `Wikipedia (apps staging environment)` since it's close enough to `Staging`, so I removed it.

joewalsh:
> One thing is maybe rename the UserTestingRelease config to just UserTesting to match the other release config naming patterns.

Done

> Also mainly a comment for thought/discussion since it's a bigger thing but I wonder if we ought to have Staging point to as many staging environments that we have available (beta cluster, etc). It might then feel like a true staging build rather than mostly-prod-but-one-part-is-staging. And we might catch issues earlier if we'll be testing this build often.

It's worth exploring - I'm not sure the app is set up to properly work with the beta cluster. We'd also need to ensure that the mobileapps staging service we use in that build was also pointing to the beta cluster. For now, I think it's useful to isolate the different staging environments so we're only changing one variable.

joewalsh:
This PR is a good candidate to test out a squash and merge.

README.md
```diff
@@ -43,6 +43,36 @@ Covered in the [process document](docs/process.md).
...
+Points to the [mobileapps](https://gerrit.wikimedia.org/r/q/project:mediawiki%252Fservices%252Fmobileapps) and [wikifeeds](https://gerrit.wikimedia.org/r/q/project:mediawiki%252Fservices%252Fwikifeeds) repos running locally.
+#### MediaWiki Beta Cluster
+Points to the [MediaWiki beta cluster](https://www.mediawiki.org/wiki/Beta_Cluster) for MediaWiki API calls and production for everything else.
+#### RTL
+Launches the app in an RTL locale using the -AppleLocale argument.
+#### Experimental
+For one off builds, can point to whatever is needed for the given experiment.
+#### User Testing
+For user testing. Has an alternate configuration so that it can be delivered ad hoc.
+#### Event Loggig Dev Debug
```
staykids, line 61: Sorry, I missed this: "loggig"

Wikipedia.xcodeproj/xcshareddata/xcschemes/RTL.xcscheme
```diff
@@ -352,7 +352,7 @@
       <CommandLineArguments>
          <CommandLineArgument
             argument = "-AppleLocale he_IL"
-            isEnabled = "NO">
+            isEnabled = "YES">
          </CommandLineArgument>
          <CommandLineArgument
             argument = "-AppleLanguages (he)"
```
tonisevener, line 358: isEnabled = "YES" on AppleLanguages too?

PR #3542, opened 2020-04-20
remove some unnecessary DEBUG checks

tonisevener:


staykids:
@tonisevener What's your feeling about the `DEBUG` conditional here? https://github.com/wikimedia/wikipedia-ios/blob/5e051dc18d4b2ed0e6261af526e30340cc53195c/Wikipedia/Code/ExploreViewController.swift#L616-L623

I personally worry when they are used to _silently_ change code paths while performing _normal_ user interactions (like in this case of a possibly inadvertent shake and the previous article pull to refresh case) because it complicates debugging issues between debug engineering builds and release/TestFlight builds in potentially non-obvious ways. IMO if we're not using the "tweaks" flags anymore, having something like an alert popup in the case of a standard user interaction having a `DEBUG` conditional path would 1) help make these paths quickly/explicitly identifiable to engineering and 2) have no effect on our end users (for the future, not suggesting we add that into this now.)

tonisevener:
> @tonisevener What's your feeling about the `DEBUG` conditional here?
> 
> https://github.com/wikimedia/wikipedia-ios/blob/5e051dc18d4b2ed0e6261af526e30340cc53195c/Wikipedia/Code/ExploreViewController.swift#L616-L623
> 
> I personally worry when they are used to _silently_ change code paths while performing _normal_ user interactions (like in this case of a possibly inadvertent shake and the previous article pull to refresh case) because it complicates debugging issues between debug engineering builds and release/TestFlight builds in potentially non-obvious ways. IMO if we're not using the "tweaks" flags anymore, having something like an alert popup in the case of a standard user interaction having a `DEBUG` conditional path would 1) help make these paths quickly/explicitly identifiable to engineering and 2) have no effect on our end users (for the future, not suggesting we add that into this now.)



> @tonisevener What's your feeling about the `DEBUG` conditional here?
> 
> https://github.com/wikimedia/wikipedia-ios/blob/5e051dc18d4b2ed0e6261af526e30340cc53195c/Wikipedia/Code/ExploreViewController.swift#L616-L623
> 
> I personally worry when they are used to _silently_ change code paths while performing _normal_ user interactions (like in this case of a possibly inadvertent shake and the previous article pull to refresh case) because it complicates debugging issues between debug engineering builds and release/TestFlight builds in potentially non-obvious ways. IMO if we're not using the "tweaks" flags anymore, having something like an alert popup in the case of a standard user interaction having a `DEBUG` conditional path would 1) help make these paths quickly/explicitly identifiable to engineering and 2) have no effect on our end users (for the future, not suggesting we add that into this now.)

@staykids yeah I know what you mean - this one I ignored I guess because I was almost certain we don't have any real functionality tied to that gesture in the app and shaking seems pretty deliberate (though if I did trigger it by accident and my feed reshuffled I would have wasted time thinking it was a bug). Looking at Tweaks again, maybe we could keep it around, clean out anything no longer useful and add stuff like this - debug code that we don't want to lose - under some sort of "Debug Help" category? If it's something fairly tied to a certain environment, we could default the tweaks value to ON in that environment (i.e. local scheme for article pull to refresh).

PR #3544, opened 2020-04-21
Make compatibility adjustments to link URLs from mobile-html

joewalsh:
Fixes the Dean Scream link. Links from the service might be `ogg` audio, which can't be read by iOS. Use the existing functionality we had for this (translated to Swift) to convert those links.

https://phabricator.wikimedia.org/T250105


WMF Framework/Configuration.swift
```diff
@@ -95,6 +95,13 @@ public class Configuration: NSObject {
         static let wikimedia = "wikimedia.org"
         static let metaWiki = "meta.wikimedia.org"
         static let wikimediafoundation = "wikimediafoundation.org"
+        static let uploads = "upload.wikimedia.org"
+    }
+    
+    public struct File {
+        public struct Extensions {
+            public static let oggAudio = Set<String>(["ogg", "oga"])
```
staykids, line 103: Is the intent behind making this a `Set` instead of an `Array` just for optimization sake? While I can see how `Set`'s technically the _more_ correct of the two data structures to use here, I wonder if there's value in it just being a plain ole `Array` to avoid unintended confusion as to why it requires the characteristics of a `Set` and to remove cruft from the definition. Since the data is well-bounded, accidentally having "ogg" or "oga" listed more than once isn't going to affect the time complexity or functionality of `contains(...)` in any meaningful way. Plus, we don't get a compiler warning or anything for accidentally defining duplicate `Set` items.

Feel free to ignore, was mostly just curious if I was missing something about the intent.

joewalsh: I made it a set for optimization and to signal that it's meant for lookup. It's entirely possible the set is slower than the array with only two elements. I'm fine with changing it to an array.

joewalsh: I wasn't really happy with having it in that struct anyway, and it's unlikely it'll be used elsewhere. I moved it into a private extension closer to where it's needed and switched it to an equality check.

joewalsh:
> Is this still working for you? Unless I'm missing something, tapping the "Yeah" link itself pushes a new link that never resolves (using Simulator on the production target). It looks like it's triggering a PCS link action, and I can't seem to trigger the getPronunciationAction path in ArticleWebMessagingController to test your related changes. Tapping the speaker icon pushes a new link as well, but it resolves. Same behavior in production 6.5.1.

Thanks for testing this and finding this issue. A couple factors were at play. First, it looks like "Wikipedia" hosted links (like the Dean scream) were never handled. Some files are hosted on Commons, some are hosted on the Wikipedias themselves, the Dean scream is hosted on English Wikipedia. I updated the link compatibility adjustment to handle these links and added tests for both cases. Second, in the updated code, the routing for pronunciation links was different than non-pronunciation audio links. I updated that logic to all run through the standard link handling and the router, so that if there's a link directly to an audio file, the app shows the `AVPlayerViewController`. Third, the speaker icon links to the file page for the audio instead of the file itself, which is probably expected to go to the file page, but I'll confirm whether or not this is the case.

Given all this, I've removed the low impact tag and we should discuss with the team whether we still want to pull this into 6.6 or wait for 6.6.1.

WikipediaUnitTests/Code/URLParsingAndRoutingTests.swift
```diff
@@ -94,4 +94,25 @@ class URLParsingAndRoutingTests: XCTestCase {
         XCTAssertEqual(url.resolvingRelativeWikiHref(".//dev/null")?.absoluteString, "https://en.wikipedia.org/wiki//dev/null")
     }
     
+    /// Ensure the 'transcoded' path component is inserted after the language path component for audio hosted on Wikipedia
+    func testWikipediaCompatibilityAdjustments() {
+        let url = URL(string: "https://upload.wikimedia.org/wikipedia/en/3/3f/DeanScream.ogg")!
```
staykids, line 99: Oh goodness, I've now officially listened to this scream more times than a human should.

PR #3546, opened 2020-04-21
Allow edit preview to reflect local PCS changes

mcleinman:
Splitting the request between local PCS and production, to allow us to see local mobilehtml changes, but still get the wikitext-to-html from the server (as that is not part of local PCS).

To test - get an edit preview page on both production and local PCS builds of the app. (Ensure you have local PCS running first.) Both should look identical.


mcleinman:
Putting this on hold for the PCS-side fix to go up, so you don't need to pull that down to test the entire piece.

mcleinman:
Taking off hold as PCS has been merged

mcleinman:
Back on hold, awaiting merge of PCS change that updates the name of the flag. (This branch has already been updated to new standard.)

mcleinman:
PCS part is merged, and this is again ready for review/merge.

Wikipedia/Code/EditPreviewViewController.swift
```diff
@@ -128,11 +128,54 @@ class EditPreviewViewController: ViewController, WMFPreviewSectionLanguageInfoDe
...
         do {
-            let request = try fetcher.mobileHTMLPreviewRequest(articleURL: articleURL, wikitext: wikitext)
+            #if WMF_LOCAL_PAGE_CONTENT_SERVICE
+            // If on local PCS, we need to split this call. On the server, wikitext-to-mobilehtml just puts together two other
+            // calls - wikitext-to-html, and html-to-mobilehtml. Since we have html-to-mobilehtml in local PCS but not the first call, if
+            // we're making PCS edits to mobilehtml we need this code in order to view them. This block splits the call on the local PCS scheme
+            // (similar to what the server does) routing the wikitext-to-html call to production, and html-to-mobilehtml to local PCS.
+            let htmlRequest = try fetcher.wikitextToHTMLRequest(articleURL: articleURL, wikitext: wikitext)
+            let htmlRequestCompletionHandler = { (data: Data?, response: URLResponse?,  error: Error?) in
```
tonisevener, line 140: I think it'd be nice if we buried the chaining and dataTask details of this into ArticleFetcher. Something like a function signature of 
```
struct SplitResult {
    let html: String
    let response: URLResponse
}
func fetchSplitMobileHTMLCall(articleURL: URL, wikitext: String, completion: (Result<SplitResult, Error>) -> Void)
```

Would give a cleaner call site in the view controller:

```
#if WMF_LOCAL_PAGE_CONTENT_SERVICE
fetcher.fetchSplitMobileHTMLCall(articleURL: articleURL, wikitext: wikitext) { result in
     DispatchQueue.main.async { 
            switch result {
                  case .success(let result):
                         self.previewWebViewContainer.webView.loadHTMLString(result.html, baseURL: result.response?.url)
                  case .failure:
                         self.showGenericError()
             } 
     }
}
#else
...
```

We can still keep the comment on line 135 here explaining it further, but it'd be nice to visually be able to easily skip over the details from the view controller.

Wikipedia/Code/EditPreviewViewController.swift
```diff
@@ -128,11 +128,54 @@ class EditPreviewViewController: ViewController, WMFPreviewSectionLanguageInfoDe
...
+                } catch {
+                    DispatchQueue.main.async {
+                        self.showGenericError()
+                    }
+                }
+            }
+            let htmlTask = fetcher.session.dataTask(with: htmlRequest, completionHandler: htmlRequestCompletionHandler)
+            htmlTask?.resume()
+            #else
+            let request = try fetcher.wikitextToMobileHTMLPreviewRequest(articleURL: articleURL, wikitext: wikitext)
```
tonisevener, line 176: I see this in the master so it's unrelated, but in the staging scheme this POST is returning a 404. Maybe worth looking into now while we're there. But I'm also happy to make a task if you think we should tackle it separately.

mcleinman: It's the exact same issue we were having - thanks for pointing that out. Staging PCS is about the same as local PCS, which is to say - in this case, you can't complete `/wikitext/to/mobilehtml/`. I updated the preprocessor flags to handle staging as well, and also updated the comments.

Wikipedia/Code/ArticleFetcher.swift
```diff
@@ -146,6 +146,50 @@ final public class ArticleFetcher: Fetcher, CacheFetching {
...
+        guard
+            let articleTitle = articleURL.wmf_title,
+            let percentEncodedTitle = articleTitle.percentEncodedPageTitleForPathComponents,
+            let url = configuration.pageContentServiceAPIURLComponentsForHost(articleURL.host, appending: ["transform", "html", "to", "mobile-html", percentEncodedTitle]).url
+        else {
+            throw RequestError.invalidParameters
+        }
+        
+        var request = URLRequest(url: url)
+        request.setValue("text/html; charset=utf-8", forHTTPHeaderField: "Content-Type")
```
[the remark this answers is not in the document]
mcleinman: Updated this. It seems that we should probably be using that wonderful method in a few other places in this class - I had copied a similar function. Is that worth making a Phab ticket for?

Wikipedia/Code/EditPreviewViewController.swift
mcleinman: I did this all with completion blocks - let me know what you think. And please double check the `@escaping` part of it all. I made the `PreviewEditVC`'s completion block use `weak self`, but I always get mildly nervous and want more eyes when I'm doing something that could easily lead to retain cycles.

PR #3548, opened 2020-04-27
Stop background task when no more articles exist

mcleinman:
Fixes warning being printed to console:
```[BackgroundTask] Background Task 10 ("MobileviewToMobileHTMLConverter"), was created over 30 seconds ago. In applications running in the background, this creates a risk of termination. Remember to call UIApplication.endBackgroundTask(_:) for your task in a timely manner to avoid this.```

It seems like we were terminating the task if anything went wrong, but not in the happy path. This line stops the task, and the message is no longer shown.


mcleinman:
@tonisevener mentioned this one directly - let me know if I should spin up a phab task for it.

PR #3549, opened 2020-04-28
Added some article performance and caching tests

tonisevener:
https://phabricator.wikimedia.org/T246147
https://phabricator.wikimedia.org/T247558
https://phabricator.wikimedia.org/T251566

Added some performance and cache tests for the new mobile-html work. It did require some tweaks in the release code to make it more testable and to fix the server error cache fallback bug, so please give those changes a close look 🙏 .

Remaining todo from the [performances PR feedback](https://github.com/wikimedia/wikipedia-ios/pull/3509#issuecomment-607198742) is that there's no easy way to update the new fixture files. I have added a [task](https://phabricator.wikimedia.org/T251482) for this for us to tackle later.


WikipediaUnitTests/Code/ArticlePerformanceTests.swift
```diff
@@ -0,0 +1,229 @@
...
+            let wikiSourceLogoImageData = wmf_bundle().wmf_data(fromContentsOfFile:"51pxWikisourceLogo.svg", ofType:"webp"),
+            let wikiBooksLogoImageData = wmf_bundle().wmf_data(fromContentsOfFile:"54pxWikibooksLogo.svg", ofType:"webp"),
+            let wikiNewsLogoImageData = wmf_bundle().wmf_data(fromContentsOfFile:"54pxWikinewsLogo.svg", ofType:"webp"),
+            let genericDogImageData = wmf_bundle().wmf_data(fromContentsOfFile:"640pxDogMorphologicalVariation", ofType:"png"),
+            let imageRegex = try? NSRegularExpression(pattern: "https://upload.wikimedia.org/wikipedia/commons/thumb.*", options: []) else {
+            assertionFailure("Error setting up fixtures.")
+            return
+        }
+
+        let _ = stubRequest("GET", "https://en.wikipedia.org/api/rest_v1/page/summary/Dog" as NSString)
```
staykids, line 68: I don't have experience with Nocilla, but just a quick note. If you're finding the `let _ =` unused assignment pattern noisy in these Swift test classes and don't want to do anything gross with Nocilla's `stubRequest` to add Objective-C's unused result attribute to suppress warnings, could be worth adding a simple facade-ish method for the Swift test classes to use. Cleans up the call sites and retains the desired behavior.

Maybe something like:
```swift
@discardableResult
func discardableStubRequest(_ method: String, _ url: LSMatcheable) -> LSStubRequestDSL {
    return stubRequest(method, url)
}
```

tonisevener: Great point! I gave this a super quick attempt but it wasn't as simple due to the extra chaining behind most of the stub methods. I'll give it another try tomorrow, but I added the changes welcome label just in case you wanted to take a look too. Nocilla gave me some trouble this round and I notice it hasn't been updated in a long time, so it may be worth looking into other options.

tonisevener:
These are ready for final review.

Wikipedia/Code/Session.swift
```diff
@@ -647,16 +649,31 @@ class SessionDelegate: NSObject, URLSessionDelegate, URLSessionDataDelegate {
     
     func urlSession(_ session: URLSession, dataTask: URLSessionDataTask, didReceive response: URLResponse, completionHandler: @escaping (URLSession.ResponseDisposition) -> Void) {
         
-        if let httpResponse = response as? HTTPURLResponse,
-            httpResponse.statusCode == 304 {
+        if let httpResponse = response as? HTTPURLResponse {
```
staykids, line 652: Is this the relevant chunk for resolving the banner issue you were mentioning in task sync? If so, is there a related phab ticket (or original repro steps) to help test the change? Apologies, but I didn't see it in the description linked tickets.

tonisevener: Sorry about that! Yes, this is the area of the fix. I added a phab task here - https://phabricator.wikimedia.org/T251566.

Wikipedia/Code/ArticleViewController.swift
```diff
@@ -67,7 +66,13 @@ class ArticleViewController: ViewController, HintPresenting {
         return rc
     }()
     
-    @objc init?(articleURL: URL, dataStore: MWKDataStore, theme: Theme) {
+    @objc convenience init?(articleURL: URL, dataStore: MWKDataStore, theme: Theme) {
```
staykids, line 69: As far as I can tell, production uses of `SchemeHandler` are always via its shared class instance. I wonder if there's value in removing this convenience initializer and just being explicit for the Objective-C calls (suggestion here: https://github.com/wikimedia/wikipedia-ios/commit/5730f5bb08b59032a927d75b75fb0bdaea7e2b01). This way, regardless of if it's via dependency injection for unit testing or by production, there's always only one initialization path into `ArticleViewController`.

Only a suggestion – I can see it either way if you feel more strongly about this approach.

WikipediaUnitTests/ArticleTestHelpers.swift
```diff
@@ -0,0 +1,450 @@
...
+        
+        //save headers in file system
+        let htmlHeaders: [String: String] = ["Content-Type": "text/html; charset=utf-8;"]
+        
+        CacheFileWriterHelper.saveResponseHeader(headerFields: htmlHeaders, toNewFileName: uniqueHTMLHeaderFileName) { (result) in
+            switch result {
+            case .success, .exists:
+                break
+            case .failure:
+                break
```
mcleinman, line 295: It looks like there is an `XTCFail` above when the call fails. Out of curiosity, is there a reason we don't want the same behavior here?

WikipediaUnitTests/ArticleTestHelpers.swift
```diff
@@ -0,0 +1,450 @@
...
+        
+        let imageHeaders: [String: String] = ["Content-Type": "image/jpeg",
+                                              "Content-Length": String((fixtureData.image as NSData).length)]
+        
+        CacheFileWriterHelper.saveResponseHeader(headerFields: imageHeaders, toNewFileName: uniqueImageHeaderFileName) { (result) in
+            switch result {
+            case .success, .exists:
+                break
+            case .failure:
+                break
```
mcleinman, line 307: same as above question

WikipediaUnitTests/ArticleTestHelpers.swift
```diff
@@ -0,0 +1,450 @@
...
+        
+        let cssHeaders: [String: String] = ["Content-Type": "text/css; charset=utf-8;",
+                                            "Content-Length": String((fixtureData.cachedCSS as NSData).length)]
+        
+        CacheFileWriterHelper.saveResponseHeader(headerFields: cssHeaders, toNewFileName: uniqueCSSHeaderFileName) { (result) in
+            switch result {
+            case .success, .exists:
+                break
+            case .failure:
+                break
```
mcleinman, line 319: same as above question

WikipediaUnitTests/ArticleTestHelpers.swift
```diff
@@ -0,0 +1,450 @@
...
+        //save headers in file system
+        let htmlHeaders: [String: String] = ["Content-Type": "text/html; charset=utf-8;",
+                                             "Vary": "Accept-Language, Accept-Encoding"]
+        
+        CacheFileWriterHelper.saveResponseHeader(headerFields: htmlHeaders, toNewFileName: uniqueHTMLHeaderFileName) { (result) in
+            switch result {
+            case .success, .exists:
+                break
+            case .failure:
+                break
```
mcleinman, line 422: same as above here, and two below -

WMF Framework/CacheFileWriterHelper.swift
```diff
@@ -88,6 +88,11 @@ final class CacheFileWriterHelper {
         replaceFileWithData(headerData, fileName: fileName, completion: completion)
     }
     
+    static func dataForFileName(_ fileName: String) -> Data? {
```
mcleinman, line 91: Is this used anywhere? Perhaps I'm missing it.

WMF Framework/CacheController.swift
```diff
@@ -10,7 +10,15 @@ public enum CacheControllerError: Error {
 
 public class CacheController {
     
+    //set for testing
+    public static var temporaryCacheURL: URL? = nil
```
staykids, line 14: `CacheController`'s public interface gets a bit muddied in app targets if it's not more explicit that `temporaryCacheURL` is only for test targets. What do you think about using our `TEST` flag to ensure a path to it isn't possible in app targets (like https://github.com/wikimedia/wikipedia-ios/commit/47fed0b4e1c502b20199e82bb92a0beb2532c1d0)?

WikipediaUnitTests/Code/ArticlePerformanceTests.swift
```diff
@@ -0,0 +1,122 @@
...
+            measurableArticleVC.initialLoadCompletion = {
+                setupExpectation.fulfill()
+                UIApplication.shared.keyWindow?.rootViewController = nil
+                dataStore.clearTemporaryCache()
+            }
+            
+            UIApplication.shared.keyWindow?.rootViewController = measurableArticleVC
+            let _ = measurableArticleVC.view
+        
+            wait(for: [setupExpectation], timeout: 10)
```
mcleinman, line 62: This implies a 10 second wait, right? I know we talked a bit about how long this should wait for... but 10 seconds seems high, if that is what it is.

Wikipedia/Code/ArticleViewController+ArticleWebMessageHandling.swift
```diff
@@ -54,7 +54,7 @@ extension ArticleViewController: ArticleWebMessageHandling {
         startSurveyAnnouncementTimer()
     }
     
-    func handlePCSDidFinishFinalSetup() {
+    @objc func handlePCSDidFinishFinalSetup() {
```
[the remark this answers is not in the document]
staykids: Because `handlePCSDidFinishFinalSetup()` get defined in a Swift extension (to `ArticleViewController`), I believe it's a Swift compiler requirement.

Wikipedia/Code/ArticleViewController.swift
tonisevener: That sounds like a good change - this is done.

WMF Framework/CacheController.swift
tonisevener: Also done (good catch and I need to remember to use that macro more).

Wikipedia/Code/ArticleViewController+ArticleWebMessageHandling.swift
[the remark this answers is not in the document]
tonisevener: yeah I think this was a auto-fix that the compiler told me to do, I have to do this to override it in the unit tests.

WikipediaUnitTests/Manual Tests/ArticleCacheReadingManualTests.swift
```diff
@@ -0,0 +1,121 @@
...
+    
+    override func tearDown() {
+        super.tearDown()
+        
+        URLCache.shared.removeAllCachedResponses()
+    }
+    
+    func testBasicNetworkNoConnectionWithCachedArticle() {
+
+        //FLIP DEVICE TO AIRPLANE MODE BEFORE RUNNING THIS
```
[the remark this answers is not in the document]
tonisevener: This doesn't fail gracefully without airplane mode unfortunately, there's an exception and it halts. I tried digging into using our reachability stuff to run an XCFail if it detects reachability but it wasn't working well and feels maybe a little overkill.

For now I have a dorky XCFail in the beginning of these tests to remind the user that this needs to be in airplane mode and to uncomment the failure once that's done. Keep in mind that I have these segmented off into a manual tests file that are not set run by default, so our build server shouldn't be affected by these. But I agree it's not ideal, I'll keep looking for a better solution.

WikipediaUnitTests/ArticleTestHelpers.swift
tonisevener: Don't think there was a reason for this, probably a miss and a a copy-paste propagation. These should be fixed now.

WMF Framework/CacheFileWriterHelper.swift
tonisevener: Nope! Good catch, I used it in a since-deleted test that I could never get to run with Nocilla. This should be gone now.

Wikipedia/Code/ArticleViewController+ArticleWebMessageHandling.swift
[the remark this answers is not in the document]
mcleinman: Sounds good!

WikipediaUnitTests/Code/ArticlePerformanceTests.swift
tonisevener: Agreed, good catch. I've bumped them down to 3 seconds.

WikipediaUnitTests/Manual Tests/ArticleCacheReadingManualTests.swift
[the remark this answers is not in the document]
mcleinman: Sounds great - seems like you've put a good amount of time into it.

WikipediaUnitTests/Code/ArticlePerformanceTests.swift
```diff
@@ -0,0 +1,122 @@
...
+            let displayExpectation = expectation(description: "Waiting for MeasurableArticlePeekPreviewViewController displayCompletion call")
+            
+            peekVC.displayCompletion = {
+                displayExpectation.fulfill()
+                UIApplication.shared.keyWindow?.rootViewController = nil
+                dataStore.clearTemporaryCache()
+            }
+            
+            UIApplication.shared.keyWindow?.rootViewController = peekVC
+            let _ = peekVC.view
```
staykids, line 117: Same question as `let _ = basicVC.view` – is this required?

WikipediaUnitTests/Code/ArticleViewControllerTests.swift
```diff
@@ -0,0 +1,58 @@
...
+        
+        measurableArticleVC.initialLoadCompletion = {
+            setupExpectation.fulfill()
+            XCTAssert(schemeHandler.accessed, "SchemeHandler was not accessed during article load.")
+            UIApplication.shared.keyWindow?.rootViewController = nil
+            dataStore.clearTemporaryCache()
+        }
+            
+        UIApplication.shared.keyWindow?.rootViewController = measurableArticleVC
+        let _ = measurableArticleVC.view
```
staykids, line 53: Same question as `let _ = basicVC.view` – is this required?

tonisevener:
By the way I fixed a new crash in the latest beta build that this PR introduces [here](https://github.com/wikimedia/wikipedia-ios/pull/3549/commits/a41638eff9edeb7059d4fc9cc7483c12fc9b7768). I could repro by 3D pressing the [Dean Scream "Yeah! link](https://phabricator.wikimedia.org/T250105).

staykids:
> By the way I fixed a new crash in the latest beta build that this PR introduces [here](https://github.com/wikimedia/wikipedia-ios/pull/3549/commits/a41638eff9edeb7059d4fc9cc7483c12fc9b7768). I could repro by 3D pressing the [Dean Scream "Yeah! link](https://phabricator.wikimedia.org/T250105).

Oh no, I thought I'd never have to hear that scream again 😅. Nice catch from the beta – I'd only been 3D touching links, references, and images, but a good reminder for me to be exhaustive with the force presses.

Your fix here also seems to resolve the case I missed of 3D touching external Safari links crashing (like URLs in footer references).

Wikipedia/Code/ArticleViewController+LinkPreviewing.swift
```diff
@@ -186,18 +175,48 @@ extension ArticleViewController: WKUIDelegate {
...
+    // MARK: Context menus
+    
+    @available(iOS 13.0, *)
+    func webView(_ webView: WKWebView, contextMenuConfigurationForElement elementInfo: WKContextMenuElementInfo, completionHandler: @escaping (UIContextMenuConfiguration?) -> Void) {
+        
+        let nullConfig = UIContextMenuConfiguration(identifier: nil, previewProvider: { () -> UIViewController? in
+            return nil
+        }) { (_) -> UIMenu? in
+            return nil
+        }
```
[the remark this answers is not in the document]
tonisevener: No reason that I can see - I updated this and seems to work fine in my testing.

WikipediaUnitTests/Code/ArticleCacheReadingTests.swift
```diff
@@ -0,0 +1,284 @@
...
+                XCTAssertEqual(trimmedCSS, "body {background-color: red;}", "Unexpected basic HTML content")
+            case "app://upload.wikimedia.org/wikipedia/en/thumb/a/a4/Flag_of_the_United_States.svg/960px-Flag_of_the_United_States.svg.png":
+                imageExpectation.fulfill()
+            default:
+                break
+            }
+        }
+        
+        UIApplication.shared.keyWindow?.rootViewController = basicVC
+        let _ = basicVC.view
```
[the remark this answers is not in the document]
tonisevener: I thought this was needed to properly trigger `viewDidLoad()` from unit tests but it could be my information is woefully out of date. I removed this in a commit and unit tests seem to run fine.

PR #3550, opened 2020-04-29
Improve VoiceOver for Table of Contents

mcleinman:
This does two things for VoiceOver:
- Reads the selected section as "Selected"
- Gives depth of header information.

We get the main header level for free via Apple's built in accessibility (in English, it reads "Heading"), but need to role our own subheader translations. This is my first time adding a new translation string in the app (I've been doing PCS-side localizable elements so far), so please let me know if any of this looks wrong.

Bug: T251318

### DIFF
--- a/Wikipedia/Code/TableOfContentsCell.swift
+++ b/Wikipedia/Code/TableOfContentsCell.swift
@@ -11,6 +11,11 @@ class TableOfContentsCell: UITableViewCell {
         didSet {
             titleLabelTopConstraint.constant = titleIndentationLevel == 0 ? 19 : 11;
             indentationConstraint.constant =  indentationWidth * CGFloat(1 + titleIndentationLevel)
+            if (titleIndentationLevel == 0) {
+                accessibilityTraits = .header
+            } else {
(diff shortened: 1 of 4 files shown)


Wikipedia/Code/TableOfContentsCell.swift
```diff
@@ -11,6 +11,11 @@ class TableOfContentsCell: UITableViewCell {
         didSet {
             titleLabelTopConstraint.constant = titleIndentationLevel == 0 ? 19 : 11;
             indentationConstraint.constant =  indentationWidth * CGFloat(1 + titleIndentationLevel)
+            if (titleIndentationLevel == 0) {
+                accessibilityTraits = .header
+            } else {
+                accessibilityValue = String.localizedStringWithFormat(WMFLocalizedString("table-of-contents-subheading-label", value:"Subheading %1$d", comment:"VoiceOver label to indicate level of subheading in table of contents. %1$d is replaced by the level of subheading."), titleIndentationLevel)
```
joewalsh, line 17: Is it possible to set `UIAccessibilityTextAttributeHeadingLevel` on the attributed string for the title or does that only work for input views?

PR #3551, opened 2020-04-29
Remove wmf_addHeightDeterminingSubviewWithConstraintsToEdges and wmf_addHeightDetermining(childController:

joewalsh:
It was added in https://github.com/wikimedia/wikipedia-ios/pull/1216 for the captcha views but it appears it's no longer necessary, captcha views are appropriately sized. Verified on iOS 11 and iOS 13. Setting the milestone to 6.6.1 and adding the hold label as it's a bit late in the release to pull this into 6.6.


staykids:
Removing the `hold` per today's standup.

WMF Framework/UIView+Constraints.swift
```diff
@@ -1,24 +1,6 @@
...
-        containerViewHeight?.isActive = false
-    }
-    
-    /// Configures the view to expand vertically dynamically to encompass the height of the subview even if the subview's height changes as a result of AutoLayout changes. Details: http://stackoverflow.com/a/35431534/135557
-    @objc func wmf_addHeightDeterminingSubviewWithConstraintsToEdges(_ subview: UIView, belowSubview: UIView? = nil) {
-        wmf_addHeightDeterminingSubview(subview, withConstraintsToEdgesWithInsets: .zero, belowSubview: belowSubview)
-    }
-    
-    @objc func wmf_addHeightDeterminingSubview(_ subview: UIView, withConstraintsToEdgesWithInsets insets: UIEdgeInsets, priority: UILayoutPriority = .required, belowSubview: UIView? = nil) {
-        wmf_deactivateHeightConstraint()
```
mcleinman, line 17: Do we lose anything by losing this line and the next line?

joewalsh: Nope - the height constraint deactivation was a workaround for a specific case that seemed to just be a misunderstanding of how to use constraints in interface builder. `translatesAutoresizingMaskIntoConstraints = false` is still called in wmf_addConstraintsToEdgesOfView

mcleinman: 👍

tonisevener:
I tested in iPhone 8 Plus 12.2 simulator and on device for iPadOS, only thing I noticed is the LoginViewController has changed slightly (notice the login button):

master:
old-landscape
old-portrait

PR:
new-landscape
new-portrait

joewalsh:
> I tested in iPhone 8 Plus 12.2 simulator and on device for iPadOS, only thing I noticed is the LoginViewController has changed slightly

Thanks for testing and finding this issue! It looks like that's the original constraint that caused this workaround to be written. I updated it to a placeholder constraint and it seems to work properly for me now on iPhone 8 Plus 12.4 sim and 11.4 sim.

PR #3554, opened 2020-05-01
Add `ArticleSurveyTimerController` for survey timer management

staykids:
https://phabricator.wikimedia.org/T248510#6098622

Incomplete and not particularly happy with my approach here, but at least wanted to have something for you to look at by the morning. Feel free to unassign me and pick it up from here or I can continue on it in the morning.

1. Pause/resume survey timer when app's backgrounded/foregrounded
2. Start survey timer in `viewWillAppear` if PCS is loaded to handle case of navigating back to article

Just stubbed a `surveyAnnouncementResult` to test timer appearance stuff. Please don't merge without removing those TODOs and testing against actual survey announcement stack – there may be stale state issues depending on how/when `surveyAnnouncementResult` is mutated. There's also a separate issue with the background/foreground notifications being triggered multiple times, but didn't want to touch anything there in case there were unintended consequences.

### DIFF
--- a/Wikipedia.xcodeproj/project.pbxproj
+++ b/Wikipedia.xcodeproj/project.pbxproj
@@ -9,6 +9,10 @@
 /* Begin PBXBuildFile section */
 		00841DE524477805003CF74A /* AppTabBarDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = FFE891452445150B0058B642 /* AppTabBarDelegate.swift */; };
 		00841DE724477806003CF74A /* AppTabBarDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = FFE891452445150B0058B642 /* AppTabBarDelegate.swift */; };
+		00A7946B245CA4E60063BA18 /* ArticleSurveyTimerController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 00A7946A245CA4E60063BA18 /* ArticleSurveyTimerController.swift */; };
+		00A7946C245CA4E60063BA18 /* ArticleSurveyTimerController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 00A7946A245CA4E60063BA18 /* ArticleSurveyTimerController.swift */; };
+		00A7946D245CA4E60063BA18 /* ArticleSurveyTimerController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 00A7946A245CA4E60063BA18 /* ArticleSurveyTimerController.swift */; };
+		00A7946E245CA4E60063BA18 /* ArticleSurveyTimerController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 00A7946A245CA4E60063BA18 /* ArticleSurveyTimerController.swift */; };
 		041EFC371996A1F800B2CB28 /* MapKit.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 041EFC361996A1F800B2CB28 /* MapKit.framework */; };
 		0E281A331DC263DE00FA1AB1 /* WMFLegacyReference.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0E281A321DC263DE00FA1AB1 /* WMFLegacyReference.swift */; };
 		0E36C2271AE0B59D00C58CFF /* Images.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = D4991453181D51DE00E6073C /* Images.xcassets */; };
@@ -3194,6 +3198,7 @@
 /* End PBXCopyFilesBuildPhase section */
 
 /* Begin PBXFileReference section */
+		00A7946A245CA4E60063BA18 /* ArticleSurveyTimerController.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ArticleSurveyTimerController.swift; sourceTree = "<group>"; usesTabs = 0; };
 		040E5C4E184566F4007AFE6F /* CoreData.framework */ = {isa = PBXFileReference; lastKnownFileType = wrapper.framework; name = CoreData.framework; path = System/Library/Frameworks/CoreData.framework; sourceTree = SDKROOT; };
 		041EFC361996A1F800B2CB28 /* MapKit.framework */ = {isa = PBXFileReference; lastKnownFileType = wrapper.framework; name = MapKit.framework; path = System/Library/Frameworks/MapKit.framework; sourceTree = SDKROOT; };
 		0E10C4FD1C81046300CEB5C2 /* Wikipedia.entitlements */ = {isa = PBXFileReference; lastKnownFileType = text.plist.entitlements; path = Wikipedia.entitlements; sourceTree = "<group>"; };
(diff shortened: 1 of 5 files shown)


tonisevener:
Thanks! I added one commit - if I pushed on another (non-survey) article from survey article, backgrounding / foregrounding would turn on the survey timer from the previous article on the stack - and ultimately the survey popup would show on the non-survey article. So I added some hairy code to disable pausing and resuming if the article isn't on screen. I haven't noticed the notifications being triggered multiple times but I'll keep testing.

staykids:
Feel free to revert if I'm missing something, but since it seems we're always interested in the most up to date state of `surveyAnnouncementResult`, I made it a read only computed property and removed the repeated `calculateSurveyAnnouncementResult()` assignment code. I'm ignoring all the other gross stuff we're doing in this in favor of us having something working for our build today 😅.

tonisevener:
> Feel free to revert if I'm missing something, but since it seems we're always interested in the most up to date state of `surveyAnnouncementResult`, I made it a read only computed property and removed the repeated `calculateSurveyAnnouncementResult()` assignment code. I'm ignoring all the other gross stuff we're doing in this in favor of us having something working for our build today 😅.

I agree, good call.

staykids:
Added the `Hold` label – best to wait on further review of this until I get a chance to make some of the updates Toni and I talked about.

staykids:
I think this is approaching a complete solution and ready for review again. @tonisevener I'm so psyched you suggested and prototyped separating the survey timer into its own controller – so much nicer with `ArticleViewController` feeling less spaghetti-ish. Here was the expected behavior criteria I followed:

1. When navigating to an article with a survey, start the timer.
2. When navigating away from an article with a survey, stop the timer.
3. When navigating back to an article with a survey, start the timer again fresh if it hasn't been seen yet (ignoring any remaining time when stopped from step 2).
4. When backgrounding app on an article with a running survey timer, save remaining timer time interval and "pause" the timer.
5. When foregrounding app on an article with survey timer, if the remaining time interval from step 4 is still in memory, resume timer with that remaining time interval.

Using in-app mocked `SurveyAnnouncementResult`s with a short timer interval, here are the navigation steps for the test scenarios I followed (may help for PR review):

1. Explore Feed > Article with Survey – survey presented
2. Explore Feed > Article without Survey – no survey
3. Article with Survey > Article without Survey – no survey on new article
4. Article with Survey > Article with Survey – _current_ article survey presented, not original article
5. Article with Survey > Article without Surver > Back to Original Article – original survey presented
6. Article without Survey > Article with Surver > Back to Original Article – no survey
7. Article with Survey that hasn't yet presented > Backgrounded > Foregrounded - survey presented after remaining time passes
8. Article without Survey > Backgrounded > Foregrounded - no survey

I'd still strongly recommend confirming this against the actual announcement stack to ensure there are no state mutation issues that create unexpected behavior.

tonisevener:
Looks good to me - I tested your cases and some of my own against a local announcement stack and it seems to work well. Thanks for taking this one on!

Wikipedia.xcodeproj/project.pbxproj
```diff
@@ -9,6 +9,10 @@
 /* Begin PBXBuildFile section */
 		00841DE524477805003CF74A /* AppTabBarDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = FFE891452445150B0058B642 /* AppTabBarDelegate.swift */; };
 		00841DE724477806003CF74A /* AppTabBarDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = FFE891452445150B0058B642 /* AppTabBarDelegate.swift */; };
+		00A7946B245CA4E60063BA18 /* ArticleSurveyTimerController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 00A7946A245CA4E60063BA18 /* ArticleSurveyTimerController.swift */; };
```
mcleinman, line 12: this line is repeated 4 times. while I can't say for sure that's bad, that seems non-typical, right?

Wikipedia/Code/ArticleSurveyTimerController.swift
```diff
@@ -0,0 +1,125 @@
...
+final class ArticleSurveyTimerController {
+
+    // MARK: - Public Properties
+
+    var timerFireBlock: ((SurveyAnnouncementsController.SurveyAnnouncementResult) -> Void)?
+
+    // MARK: - Private Properties
+
+    private let articleURL: URL
+    private let surveyController: SurveyAnnouncementsController
```
mcleinman, line 13: Do we need this? We pass in `SurveyAnnoucementsController.shared` into the initializer - doesn't that mean we could just call `SurveyAnnoucementsController.shared` in this class instead?

Wikipedia/Code/ArticleSurveyTimerController.swift
```diff
@@ -0,0 +1,125 @@
...
+    // MARK: - Lifecycle
+
+    init(articleURL: URL, surveyController: SurveyAnnouncementsController) {
+        self.articleURL = articleURL
+        self.surveyController = surveyController
+    }
+
+    // MARK: - Public
+
+    func articleContentDidLoad() {
```
mcleinman, line 41: Not a bug, just extra code - 

We don't check that `surveyAnnouncementResult` is non-nil here and we don't set `shouldPauseOnBackground` as `true` - as we do in `viewWillAppear` below. This is ultimately ok, because these things happen in `startTimer`. But given that it happens in `startTimer`, do we need these two things in `viewWillAppear`? In `viewWillAppear`, these two things happen twice - once in `viewWillAppear`, and once in `startTimer`.

Wikipedia.xcodeproj/project.pbxproj
staykids: These identifiers (`00A7946B245CA4E60063BA18` etc.) are actually not identical – easy to miss because they look so similar! I'm not an Xcode PBX file format expert, but if I'm not mistaken the similar looking lines are because of `ArticleSurveyTimerController`'s target membership in 4 app targets (e.g. Wikipedia, Staging, etc.), so this repetition is expected.

tonisevener: It shows up because we have it selected for 4 different targets - Wikipedia, Staging, Experimental, and User Testing.

Wikipedia/Code/ArticleSurveyTimerController.swift
tonisevener: yeah, I have a bad habit of setting things up for easier unit testing but not following through on making the tests themselves. But with this method we could inject our own flavor of SurveyAnnouncementsController that hard codes announcements. But I still don't expect we'd be able to do this without further decoupling, i.e. having this be a protocol that `SurveyAnnouncementsController` conforms to or marking `SurveyAnnouncementsController` as not-final so we can subclass with our own announcements. I'm fine with taking out dependency injection until we actually use it.

Wikipedia.xcodeproj/project.pbxproj
mcleinman: Clearly I didn't look at those strings closely enough.

Wikipedia/Code/ArticleSurveyTimerController.swift
staykids: Thanks for pointing this out - I just noticed I forgot to change: https://github.com/wikimedia/wikipedia-ios/blob/aced47d626cb3765c7e587168ae5ac7b66e5912d/Wikipedia/Code/ArticleSurveyTimerController.swift#L28 to use `surveyController`, so I'll do that now.

@tonisevener I think maintaining the classical dependency injection here, even if we're unable to immediately follow through on tests, more clearly establishes the responsibilities  and expectations when using this (or any) controller. Overusing singletons definitely contributes to unclear code paths and potentially confusing global mutable state.

Unless either of you feel very strongly that we should remove it, I think it's stronger in the long run to leave the dependency injection here as-is.

mcleinman: I was just coming back to point out the `.shared` still in there.

mcleinman: Once that's updated, I'm good to merge.

staykids: 👍 just removed that singleton usage and also removed some extraneous indentation. Feel free to merge when you're comfortable with it!

Wikipedia/Code/ArticleSurveyTimerController.swift
staykids: Apologies for not addressing this one earlier. Great point about potentially extraneous code. There are definitely some defensive programming strategies here, which I think are in large part helpful given the global mutability within `surveyAnnouncementResult` and async nature of web requests. The `viewWillAppear` and `viewWillDisappear` assignments in particular work together to mitigate issues surrounding 5, 6, and 7 from https://github.com/wikimedia/wikipedia-ios/pull/3554#issuecomment-624346601. Otherwise you run into unexpected timer behavior navigating the article stack and backgrounding/foregrounding the app.

I'd be happy to continue looking into optimizations here, but I wouldn't feel comfortable with us committing/merging changes to the timer start/stop/pause/resume logic unless we all had an opportunity to do another full round of tests with the updated logic. For that reason, if you're comfortable with it, may be worth merging and getting functional in-use feedback from the beta users.

mcleinman: Per elsewhere, I'm definitely comfortable merging it. Will do that shortly. 

That makes sense about being defensive around all this. We may also (at some point in the future) want to add the extra defensive check to `articleContentDidLoad` - that's the only one that doesn't have it.

PR #3555, opened 2020-05-01
 Update PCS conversion assets, ensure legacy user data is removed

joewalsh:
I tested this with a 6.5.1 library by launching offline, and verified that the article content appears. Tapping through on an image while in this offline state doesn't work but IIRC that's expected as long as the article content is there and the images are inline in the article.


tonisevener:
I think we should restore the incremental article data deletion lines [in this file](https://github.com/wikimedia/wikipedia-ios/blob/c8fc020e96104a9687e902e8ffe6b7c5c48d4918/Wikipedia/Code/MWKDataStore%2BLegacyMobileview.swift#L35). Seems like it would keep the data footprint steady while a user is migrating a giant library.

joewalsh:
> I think we should restore the incremental article data deletion lines [in this file](https://github.com/wikimedia/wikipedia-ios/blob/c8fc020e96104a9687e902e8ffe6b7c5c48d4918/Wikipedia/Code/MWKDataStore%2BLegacyMobileview.swift#L35). Seems like it would keep the data footprint steady while a user is migrating a giant library.

I agree and I missed those so thanks for catching! Fixed with the latest commit

mcleinman:
@joewalsh can you confirm the rough test steps here - 

This is what we were talking about this morning, I believe. Converting offline articles and deleting the old articles. This should work when offline, right? So...
1) Have a few articles saved on 6.5.1
2) Run 6.6 (in offline mode).
3) Ensure articles are converted and you can read the saved articles. 
4) This should be tested both with opening to an article page (the current article should be readable), as well as the Explore feed.

Is this correct?

joewalsh:
> @joewalsh can you confirm the rough test steps here -
> 
> This is what we were talking about this morning, I believe. Converting offline articles and deleting the old articles. This should work when offline, right? So...
> 
> 1. Have a few articles saved on 6.5.1
> 2. Run 6.6 (in offline mode).
> 3. Ensure articles are converted and you can read the saved articles.
> 4. This should be tested both with opening to an article page (the current article should be readable), as well as the Explore feed.
> 
> Is this correct?

1-3 are correct. I'm not sure I understand what you mean by  4- only saved articles should work offline, not articles from the explore feed

mcleinman:
> > @joewalsh can you confirm the rough test steps here -
> > This is what we were talking about this morning, I believe. Converting offline articles and deleting the old articles. This should work when offline, right? So...
> > 
> > 1. Have a few articles saved on 6.5.1
> > 2. Run 6.6 (in offline mode).
> > 3. Ensure articles are converted and you can read the saved articles.
> > 4. This should be tested both with opening to an article page (the current article should be readable), as well as the Explore feed.
> > 
> > Is this correct?
> 
> 1-3 are correct. I'm not sure I understand what you mean by 4- only saved articles should work offline, not articles from the explore feed

It looks like we run `migrateMobileviewToMobileHTMLIfNecessary` in two places - `SavedArticleFetcher,` as well as `ArticleViewController`'s `articleLoadDidFail`. I was guessing - perhaps incorrectly - that this is about reloading a current page. Perhaps not - is this just for a situation where a given saved article wasn't already reloaded from the other route?

tonisevener:
> > > @joewalsh can you confirm the rough test steps here -
> > > This is what we were talking about this morning, I believe. Converting offline articles and deleting the old articles. This should work when offline, right? So...
> > > 
> > > 1. Have a few articles saved on 6.5.1
> > > 2. Run 6.6 (in offline mode).
> > > 3. Ensure articles are converted and you can read the saved articles.
> > > 4. This should be tested both with opening to an article page (the current article should be readable), as well as the Explore feed.
> > > 
> > > Is this correct?
> > 
> > 
> > 1-3 are correct. I'm not sure I understand what you mean by 4- only saved articles should work offline, not articles from the explore feed
> 
> It looks like we run `migrateMobileviewToMobileHTMLIfNecessary` in two places - `SavedArticleFetcher,` as well as `ArticleViewController`'s `articleLoadDidFail`. I was guessing - perhaps incorrectly - that this is about reloading a current page. Perhaps not - is this just for a situation where a given saved article wasn't already reloaded from the other route?

@mcleinman Yep, `SavedArticlesFetcher` is an ongoing job that converts saved articles one by one. If the user preemptively visits a saved article before it's been converted, the call in `ArticleViewController`'s `articleLoadDidFail` catches it and converts it on-the-fly. This way we don't have to force the user to wait until all of their articles have converted before being able to read an article while offline. You should be able to trigger it if you launch the app in offline mode and visit a saved article at the end of the list. If SavedArticlesFetcher hasn't reached it yet to convert, the ArticleViewController call will catch it.

tonisevener:
> I get a whole lot (like dozens/hundreds) of this line in the log when I do the first run of 6.6 (on this branch):
> 
> ```
> 2020-05-01 15:08:37.053600-0700 Wikipedia[29703:4879748] [Process] kill() returned unexpected error 1
> 2020-05-01 15:08:41.000683-0700 Wikipedia[29703:4879748] [Process] kill() returned unexpected error 1
> 2020-05-01 15:08:41.000931-0700 Wikipedia[29703:4879748] [Process] kill() returned unexpected error 1
> ```
> 
> It doesn't seem to affect the app's run, and it's not there on the second run. Is this something worth investigating? (It's possible it's unrelated to this PR, and could be some other part of the app update process.)

Yeah I saw that for a while, I think it was a WebKit bug in some iOS13 releases (see [here](https://stackoverflow.com/questions/58321114/why-i-get-the-console-warning-process-kill-returned-unexpected-error-1-when). Apparently it's been fixed in 13.4 and I have been seeing it less in testing. Try running in a 13.4 simulator and see if they're gone. Also I think these show up on the first run for you because we're running a hidden web view to convert these articles, once it's done converting they stop, though I think you'll start seeing them again as soon as you tap into an article.

mcleinman:
> > I get a whole lot (like dozens/hundreds) of this line in the log when I do the first run of 6.6 (on this branch):
> > ```
> > 2020-05-01 15:08:37.053600-0700 Wikipedia[29703:4879748] [Process] kill() returned unexpected error 1
> > 2020-05-01 15:08:41.000683-0700 Wikipedia[29703:4879748] [Process] kill() returned unexpected error 1
> > 2020-05-01 15:08:41.000931-0700 Wikipedia[29703:4879748] [Process] kill() returned unexpected error 1
> > ```
> > 
> > 
> > It doesn't seem to affect the app's run, and it's not there on the second run. Is this something worth investigating? (It's possible it's unrelated to this PR, and could be some other part of the app update process.)
> 
> Yeah I saw that for a while, I think it was a WebKit bug in some iOS13 releases (see [here](https://stackoverflow.com/questions/58321114/why-i-get-the-console-warning-process-kill-returned-unexpected-error-1-when)). Apparently it's been fixed in 13.4 and I have been seeing it less in testing. Try running in a 13.4 simulator and see if they're gone. Also I think these show up on the first run for you because we're running a hidden web view to convert these articles, once it's done converting they stop, though I think you'll start seeing them again as soon as you tap into an article.

That is correct. Thank you! I'm going to merge this.

PR #3561, opened 2020-05-05
Match functionality of JavaScript's encodeURIComponent() for encoding URL path components

joewalsh:
https://phabricator.wikimedia.org/T251919


WMF Framework/NSCharacterSet+WMFLinkParsing.m
```diff
@@ -7,7 +7,7 @@ + (NSCharacterSet *)wmf_URLArticleTitlePathComponentAllowedCharacterSet {
     static NSCharacterSet *wmf_URLArticleTitleAllowedCharacterSet;
     dispatch_once(&onceToken, ^{
         NSMutableCharacterSet *pathAllowedCharacterSet = [[NSCharacterSet URLPathAllowedCharacterSet] mutableCopy];
-        [pathAllowedCharacterSet removeCharactersInString:@"/."];
```
tonisevener, line 10: It's a bit hard to track down what all could be affected but on first glance it seems like continue reading and related pages content group URLs that are generated hit this. Seems like something like this snippet from `WMFContinueReadingContentSource.m` could create a different `continueReadingURL` than what is persisted, in which case this fetch would start to return nothing.

```
NSURL *continueReadingURL = [WMFContentGroup continueReadingContentGroupURLForArticleURL:lastRead];
[moc fetchOrCreateGroupForURL:continueReadingURL
        ofKind:WMFContentGroupKindContinueReading
        forDate:userData.viewedDate
        withSiteURL:nil
        associatedContent:nil
        customizationBlock:^(WMFContentGroup *_Nonnull group) {
                group.contentPreview = lastRead;
        }
];
```

joewalsh: Good catch - for continue reading the old content sources are removed but for related pages, this change could cause a duplicate entry in the feed for articles with + in the title. I also found that the talk page (and anywhere else we construct a URL by setting wmf_title) now have the + encoded, but that is working fine.

joewalsh: For the duplicate related page issue, I'd rather take the risk that users with a related page card for an article with a + in the title will see a duplicate related page card instead of adding a migration to re-encode content group keys that will slow down migration for all users, but it's worth flagging the tradeoff to the rest of the team to weigh in.

mcleinman: @joewalsh to clarify, this would just be in the transition period (based on what they have in the existing app version, and what this would provide)? So users wouldn't be seeing duplicates in perpetuity? (This is my first time looking at that corner of the codebase.)

joewalsh: @mcleinman just in the transition period

joewalsh:
> I think `-` may have the same problem. `B-Line (Sydney)` also doesn't it's title description update upon refresh. That sound right? If so, I'm happy to make the updates.

This might be the bug fixed by #3564 because I'm seeing the right description there now, but aside from that I put in a request to verify which characters actually need to be encoded here: https://phabricator.wikimedia.org/T249284#6113649

PR #3562, opened 2020-05-05
Change "edit preview" view trigger to server-side PCS

mcleinman:
We had written some code that allowed the iOS-side setup to strip out editing elements. With additional editing elements needing manipulation for "edit preview" mode, we moved the work to the server part of PCS. This PR allows for that change - removing the local PCS trigger, and sending the proper flag to server-side PCS mobileHTML setup.

The flag is sent in two different requests because of the work to get Edit Previews working w/ mobile HTML on local PCS. We split the request on local PCS, and thus the flag gets sent twice - once for production, and once for PCS.

This interacts w/ other branches:
_PCS_ - [A patch](https://gerrit.wikimedia.org/r/c/mediawiki/services/mobileapps/+/594294) is currently up for branch `mattc_edit_regressions`. Until it is merged, you'll need to check that branch out to see this PR's work.
_Mediawiki_ - We need RESTbase to propagate the `isEditPreview` header from `wikitext/to/mobile-html` to `html/to/mobile-html` in order for this to work on production. I'll be filing that separately.

To test, check out that PCS branch, then try editing any section. You shouldn't see any edit pencil icons nor an article header. No sections should be collapsible. You should see "Preview of references" before the (uncollapsed) list of references in the section. You may note some missing references in that list - there is a [Phab ticket](https://phabricator.wikimedia.org/T251830) for that bug.


mcleinman:
Putting on hold - currently evolving discussion about best PCS strategy for this.

Wikipedia/Code/ActionHandlerScript.swift
```diff
@@ -26,7 +26,6 @@ final class PageContentService   {
             var textSizeAdjustmentPercentage: String // string like '125%'
             
             var userGroups: [String]
-            var areEditButtonsHidden: Bool
```
tonisevener, line 29: Just a heads up I made this [task](https://phabricator.wikimedia.org/T252218) to update the bundled files one more time since it seems like the output will be changing once the patch is merged. If I do a search for `areEditButtonsHidden` in the project I see that it's still in pcs.js. These bundled files are referenced and copied into our caching system when the user migrates old saved articles, to allow them to read saved articles when they launch 6.6 in offline mode. We can do this work separate from this PR though.

mcleinman: Sounds good. It's probably good to keep them in sync. Though unless you can edit in offline mode, these changes won't affect the offline experience.

Wikipedia/Code/ArticleFetcher.swift
```diff
@@ -166,21 +173,24 @@ final public class ArticleFetcher: Fetcher, CacheFetching {
...
             throw RequestError.invalidParameters
         }
-        return session.request(with: url, method: .post, bodyParameters: html, bodyEncoding: .html)
+
+        let headers = ["output-mode": mobileHTMLOutput.rawValue]
+        return session.request(with: url, method: .post, bodyParameters: html, bodyEncoding: .html, headers: headers)
     }
 
-    public func splitWikitextToMobileHTMLString(articleURL: URL, wikitext: String, completion: @escaping ((String?, URL?) -> Void)) throws {
+    public func wikitextToMobileHTMLRequest(articleURL: URL, wikitext: String, mobileHTMLOutput: MobileHTMLType = .contentAndReferences, completion: @escaping ((String?, URL?) -> Void)) throws {
```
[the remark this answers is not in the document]
mcleinman: I was trying to make it in line with the other two I added, but that is  a really good point. I updated it to `fetchMobileHTMLFromWikitext`. Another option would be `wikitextToMobileHTMLFetcher`. I'm also open to others.

Wikipedia/Code/ArticleFetcher.swift
```diff
@@ -166,21 +173,24 @@ final public class ArticleFetcher: Fetcher, CacheFetching {
...
         guard
             let articleTitle = articleURL.wmf_title,
             let percentEncodedTitle = articleTitle.percentEncodedPageTitleForPathComponents,
             let url = configuration.pageContentServiceAPIURLComponentsForHost(articleURL.host, appending: ["transform", "html", "to", "mobile-html", percentEncodedTitle]).url
         else {
             throw RequestError.invalidParameters
         }
-        return session.request(with: url, method: .post, bodyParameters: html, bodyEncoding: .html)
+
+        let headers = ["output-mode": mobileHTMLOutput.rawValue]
```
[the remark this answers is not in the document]
mcleinman: Good call, done!

Wikipedia/Code/ActionHandlerScript.swift
tonisevener: @mcleinman true, nope you can't edit in offline mode.

mcleinman:
@staykids This branch was a ways behind `master` - I just updated it, and I believe this no longer occurs. Please let me know if you still see it.

mcleinman:
@staykids - but I should have merged in `6.6` :facepalm: hang tight

mcleinman:
Take 2: @staykids This branch was a ways behind `6.6` - I just updated it, and I believe this no longer occurs. Please let me know if you still see it.

mcleinman:
> Nice - psyched to see our editing regressions getting ironed out.
> 
> > Mediawiki - We need RESTbase to propagate the `isEditPreview` header from `wikitext/to/mobile-html` to `html/to/mobile-html` in order for this to work on production. I'll be filing that separately.
> 
> Do you know if this work got merged? Testing the production app target, seems like probably not.

Not yet, it's getting filed shortly. That one requires the updated PCS patch to be merged, otherwise the RESTbase tests fail. That PCS one is approved but not merged.

PR #3564, opened 2020-05-05
Refresh the page on edit of description or section

joewalsh:
Fixes https://phabricator.wikimedia.org/T251936

When the user edits the article and there's a new revision ID, the app can bypass the cache by explicitly loading that revision. When the user edits the Wikidata description, the app still needs to poll for an ETag change. This approach is based on the discussion on [T252146](https://phabricator.wikimedia.org/T252146).


Wikipedia/Code/ArticleFetcher.swift
```diff
@@ -281,6 +281,54 @@ final public class ArticleFetcher: Fetcher, CacheFetching {
...
+            defer {
+                self.untrack(taskFor: key)
+            }
+            if let error = error {
+                completion(.failure(error))
+                return
+            }
+            guard
+                let httpURLResponse = response as? HTTPURLResponse,
+                httpURLResponse.statusCode == 200,
```
mcleinman, line 309: Why check for 200 here? If we get an error it will be caught above. 

It seems like the first three items in this guard are good safety checks - but if they resolve to `false` we probably want to treat them differently than the final `updatedETag != eTag`. The first three should never be false (I think), so if any of them are we might want to failure out of this loop. Let's say there is some weird error, and the server is returning a response w/ no error but including a 500 - the user will be caught in an infinite loop (until the timeout, that is).

Wikipedia/Code/ArticleViewController.swift
```diff
@@ -242,6 +246,8 @@ class ArticleViewController: ViewController, HintPresenting {
             switch state {
             case .initial:
                 break
+            case .reloading:
+                fallthrough
```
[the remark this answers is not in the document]
joewalsh: The `fallthrough` is here so that `.reloading` is treated like `.loading` - in both cases we want to call `fakeProgressController.start()`

Wikipedia/Code/ArticleFetcher.swift
joewalsh: Good question - the `If-None-Match` header is set above meaning that we should see a response code of 304 (not modified) until the eTag changes. However, there was a bug in this PR causing a cached response to be returned with a 200 status code and the old ETag. I pushed a change to fix that problem by utilizing the `.reloadIgnoringLocalCacheData` cache policy. I also updated the status code handling to fail on any code other than 200 and 304 that don't cause an error to be returned by the networking stack.

mcleinman: Looks great!

staykids:
> I'm consistently getting an error due to time out. Let me know if you can't replicate.

Seems like same for me.

joewalsh:
> > I'm consistently getting an error due to time out. Let me know if you can't replicate.
> 
> Seems like same for me.

Looks like an upstream issue

joewalsh:
> > > I'm consistently getting an error due to time out. Let me know if you can't replicate.
> > 
> > 
> > Seems like same for me.
> 
> Looks like an upstream issue

Filed https://phabricator.wikimedia.org/T252142

Wikipedia/Code/ArticleFetcher.swift
```diff
@@ -281,6 +281,79 @@ final public class ArticleFetcher: Fetcher, CacheFetching {
...
+            return nil
+        }
+        let requestURL: URL
+        do {
+            requestURL = try mobileHTMLURL(articleURL: articleURL)
+        } catch let error {
+            completion(.failure(error))
+            return nil
+        }
+        let start = CFAbsoluteTimeGetCurrent()
```
staykids, line 297: Just out of curiosity: is the intention behind using Foundation's time utility here over something higher level like `Date` just for performance characteristic improvements? I.e. avoiding a slow heavyweight `Date`. If I'm not mistaken, this layer of time utility is still affected by the system clock, same as `Date` is. So there might be funny business if the user manipulates the clock during the initial 7 second timeout, or the clock is changed under them by the system/DST/NTP. From what I understand `mach_absolute_time` provides a system clock agnostic way to performantly and precisely handle duration. Though again to be clear, I'm not suggesting we change this or use that – in practice, this seems fine. Just curious.

Wikipedia/Code/ArticleFetcher.swift
```diff
@@ -281,6 +281,79 @@ final public class ArticleFetcher: Fetcher, CacheFetching {
...
+                return
+            }
+            
+            let retry = {
+                dispatchOnMainQueueAfterDelayInSeconds(0.25) {
+                    let end = CFAbsoluteTimeGetCurrent()
+                    let duration = end - start
+                    self.waitForMobileHTMLChange(articleURL: articleURL, eTag: eTag, timeout: timeout - duration, cancellationKey: key, completion: completion)
+                }
+
```
staykids, line 321:
```suggestion                
```

Wikipedia/Code/ArticleFetcher.swift
tonisevener: Infinite loop potential is a bit scary here, we could also alternatively (or in addition to this) try a `maximumIterationCount` and force them out if they've reached beyond a certain number there.

joewalsh: > Just out of curiosity: is the intention behind using Foundation's time utility here over something higher level like `Date` just for performance characteristic improvements? I.e. avoiding a slow heavyweight `Date`. If I'm not mistaken, this layer of time utility is still affected by the system clock, same as `Date` is. So there might be funny business if the user manipulates the clock during the initial 7 second timeout, or the clock is changed under them by the system/DST/NTP. From what I understand `mach_absolute_time` provides a system clock agnostic way to performantly and precisely handle duration. Though again to be clear, I'm not suggesting we change this or use that – in practice, this seems fine. Just curious.

No specific intention here, thanks for flagging this as the system time changing could cause an issue. I did some digging and it looks like `DispatchTime` is a good somewhat high-level system-time-zone-agnostic option. I've updated the PR to use `DispatchTime`

staykids: > Infinite loop potential is a bit scary here, we could also alternatively (or in addition to this) try a `maximumIterationCount` and force them out if they've reached beyond a certain number there.

@tonisevener I very well could be missing something, but if `DispatchTime` is unaffected by the system clock (that is, its value _always_ increases in "tick" units), then we're guaranteed that `duration` is always non-negative and recursive calls into `waitForMobileHTMLChange` are always with timeout values _less than_ the originally seeded value (7 seconds). Given that, seems like the `guard timeout > 0` will always cover us and obviate the need for a `maximumIterationCount`, because it's essentially functionally equivalent to one. What do you think?

joewalsh: After thinking about this some more, I updated this to use `maxAttempts` instead of `timeout`. It allowed for an easier way to make each attempt wait longer in-between instead of making a request every quarter second. Also, given that `Double`s can do strange things sometimes, even though in theory the timeout should be safe, I agree with Toni's sentiment and feel better about having the loop breaking condition being an integer check.

tonisevener: Thanks, it's working well for me. @staykids agreed, I think I would have approved either approach - `DispatchTime` felt much safer than `CFAbsoluteTimeGetCurrent ` (great catch on that).

Wikipedia/Code/ArticleViewController.swift
```diff
@@ -551,17 +561,42 @@ class ArticleViewController: ViewController, HintPresenting {
...
+        }
+    }
 
     internal func performWebViewRefresh() {
         #if WMF_LOCAL_PAGE_CONTENT_SERVICE // on local PCS builds, reload everything including JS and CSS
-            webView.reloadFromOrigin()
+            self.webView.reloadFromOrigin()
         #else // on release builds, just reload the page with a different cache policy
-            loadPage(cachePolicy: .noPersistentCacheOnError)
+            self.loadPage(cachePolicy: .noPersistentCacheOnError)
```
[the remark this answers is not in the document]
joewalsh: Good catch, fixed

staykids:
Hmm - it's seemed to return to the state of constantly throwing the error banner. In particular it seems to be hitting the `guard maxAttempts > 0` case most of the time. Tracking the entity tag through the process, it either most of  the time remains the same through the retries and fails or changes on one of the last attempts. Makes me a bit worried if it's some nondeterministic amount of time more than what we have now that's required to solve this.

Here's a GIF testing on a Test Wiki article. I've confirmed similar behavior on an orphaned English article:

edit_error

joewalsh:
> Hmm - it's seemed to return to the state of constantly throwing the error banner. In particular it seems to be hitting the `guard maxAttempts > 0` case most of the time. Tracking the entity tag through the process, it either most of the time remains the same through the retries and fails or changes on one of the last attempts. Makes me a bit worried if it's some nondeterministic amount of time more than what we have now that's required to solve this.
> 
> Here's a GIF testing on a Test Wiki article. I've confirmed similar behavior on an orphaned English article:
> 
> edit_error

I updated the wait time increase to be exponential and it seems to be working for me (albeit slowly). I also updated the messaging to be clearer in the failure case.

mcleinman:
I'm going to remove myself as a reviewer from this, unless you need me as a second person. The others are more up to date on the changes on this, I believe.

Wikipedia/Code/ArticleViewController.swift
```diff
@@ -551,17 +562,48 @@ class ArticleViewController: ViewController, HintPresenting {
...
+        }
+    }
 
-    internal func performWebViewRefresh() {
+    internal func performWebViewRefresh(_ revisionID: UInt64? = nil) {
         #if WMF_LOCAL_PAGE_CONTENT_SERVICE // on local PCS builds, reload everything including JS and CSS
             webView.reloadFromOrigin()
         #else // on release builds, just reload the page with a different cache policy
-            loadPage(cachePolicy: .noPersistentCacheOnError)
+            loadPage(cachePolicy: .noPersistentCacheOnError, revisionID: revisionID)
```
tonisevener, line 606: I notice that reloading this page with the revisionID does not update the underlying saved page content. If you change to airplane mode after a content update, clear out URLCache.shared in the app settings and return, you will see the old content. My guess is because the additional revision ID in the path generates a different database key, so it doesn't see the saved article as being the same article to update. I'm not sure that this is a big enough deal to fix at this point in the release cycle because I confirmed a simple pull to refresh or revisiting the page with a connection updates the content with the latest.

Wikipedia/Code/ArticleViewController.swift
```diff
@@ -551,17 +562,48 @@ class ArticleViewController: ViewController, HintPresenting {
...
+        // If no revisionID was provided, wait for the ETag to change
+        guard let eTag = currentETag else {
+            performWebViewRefresh()
+            return
+        }
+        fetcher.waitForMobileHTMLChange(articleURL: articleURL, eTag: eTag, maxAttempts: 5) { (result) in
+            DispatchQueue.main.async {
+                switch result {
+                case .failure(let error):
+                    self.showError(error)
```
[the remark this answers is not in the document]
joewalsh: Good catch - I fixed this by making the message sticky

Wikipedia/Code/ArticleViewController.swift
joewalsh: I made an attempt at fixing this with 0ad47a7 but wasn't able to fully test it yet - we can roll it back if you feel it's too invasive and too late for this kind of change.

tonisevener: @joewalsh yeah, I think I would rather push this particular tweak to 6.6.1 to give us more testing time on it. I can make a Phab task on it here in a bit.

tonisevener: Here's the phab [task](https://phabricator.wikimedia.org/T252435) for this bug.

Wikipedia/Code/ArticleViewController.swift
```diff
@@ -551,17 +562,48 @@ class ArticleViewController: ViewController, HintPresenting {
...
+        // If no revisionID was provided, wait for the ETag to change
+        guard let eTag = currentETag else {
+            performWebViewRefresh()
+            return
+        }
+        fetcher.waitForMobileHTMLChange(articleURL: articleURL, eTag: eTag, maxAttempts: 5) { (result) in
+            DispatchQueue.main.async {
+                switch result {
+                case .failure(let error):
+                    self.showError(error, sticky: true)
```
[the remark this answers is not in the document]
joewalsh: Good catch - I fixed this by explicitly passing the viewController to RMessage. This is the intent everywhere we're using `showError` on `viewController` so it's helpful to make that explicit. There are probably other places that have the same problem with alertManager/RMessage so it's likely worth having a bigger re-think on those the next time we're adjusting error messaging.

Wikipedia/Code/ArticleViewController.swift
staykids: Possibly related to this, the only remaining functional quirk I'm noticing consistently is this scenario: 

1. Make an edit on an article. 
2. After publishing, observe the change and force quit the app.
3. Relaunch the app - state restoration takes you back to the same article.
4. Observe that the change from step 2 isn't visible.
5. Tapping edit again shows the most "up to date" edit.

edit_stale

@tonisevener Do you want to address that with this PR or save it for [T252435](https://phabricator.wikimedia.org/T252435) with 6.6.1 as well?

tonisevener: @staykids oo good catch - I'm pretty certain the same fix would fix both problems. I'll ask @jtminor on what he thinks.

tonisevener: @staykids Per [https://phabricator.wikimedia.org/T252435#6137802](url), this can wait till 6.6.1.

PR #3570, opened 2020-05-08
Allow user to tap article to dismiss references

tonisevener:
https://phabricator.wikimedia.org/T244627#6117483

Attempt to allow the user to both scroll the article underneath and tap to dismiss while the `ReferenceBackLinksViewController` overlay is presenting.

### DIFF
--- a/Wikipedia/Code/ArticleViewController+References.swift
+++ b/Wikipedia/Code/ArticleViewController+References.swift
@@ -206,6 +206,10 @@ extension ArticleViewController: ReferenceBackLinksViewControllerDelegate {
 
 extension ArticleViewController: UIGestureRecognizerDelegate {
     func gestureRecognizer(_ gestureRecognizer: UIGestureRecognizer, shouldRecognizeSimultaneouslyWith otherGestureRecognizer: UIGestureRecognizer) -> Bool {
-        return true
+        if gestureRecognizer === referenceWebViewBackgroundTapGestureRecognizer {
+            return true
+        }
+        
+        return false //default
     }
 }
--- a/Wikipedia/Code/ArticleViewController+References.swift
+++ b/Wikipedia/Code/ArticleViewController+References.swift
@@ -81,6 +81,10 @@ extension ArticleViewController {
         dismiss(animated: true)
     }
     
+    @objc func tappedWebViewBackground() {
+        dismissReferenceBackLinksViewController()
+    }
+    
     func showReferenceBackLinks(_ backLinks: [ReferenceBackLink], referenceId: String, referenceText: String) {
         guard let vc = ReferenceBackLinksViewController(referenceId: referenceId, referenceText: referenceText, backLinks: backLinks, delegate: self, theme: theme) else {
             showGenericError()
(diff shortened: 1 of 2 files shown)


tonisevener:
> Apologies if I'm misunderstanding something about the description, but is part of the intent of this PR/that ticket comment to be able to scroll while _this_ overlay is presented? That is, a tap action on the darkened text area should dismiss the References view, but also any panning actions on that same text area should scroll the underlying article?
> 
> ref

Nope, sorry that was very unclear. It's from this view, when you tap on one of the reference links in the References section at the bottom ([1], [2], etc). Currently it shows this overlay - the article underneath isn't darkened and can be scrolled. This PR adds the tap to dismiss.

IMG_0110

Wikipedia/Code/ArticleViewController.swift
```diff
@@ -57,6 +57,14 @@ class ArticleViewController: ViewController, HintPresenting {
         return rc
     }()
     
+    lazy var referenceWebViewBackgroundTapGestureRecognizer: UITapGestureRecognizer = {
+        let tapGR = UITapGestureRecognizer(target: self, action: #selector(tappedWebViewBackground))
+        tapGR.delegate = self
```
staykids, line 62:
```suggestion
```

We don't appear to be leveraging the delegate methods for the gesture recognizer, so we could remove this.

Wikipedia/Code/ArticleViewController+References.swift
```diff
@@ -197,3 +203,9 @@ extension ArticleViewController: ReferenceBackLinksViewControllerDelegate {
         }
     }
 }
+
+extension ArticleViewController: UIGestureRecognizerDelegate {
+    func gestureRecognizer(_ gestureRecognizer: UIGestureRecognizer, shouldRecognizeSimultaneouslyWith otherGestureRecognizer: UIGestureRecognizer) -> Bool {
+        return true
+    }
+}
```
staykids, line 211:
```suggestion
```

Same note as `referenceWebViewBackgroundTapGestureRecognizer` – removing the unneeded delegate assignment removes the need for this conformance.

Wikipedia/Code/ArticleViewController.swift
tonisevener: Actually without this and the delegate method conformance in `ArticleViewController+References.swift`, `tappedWebViewBackground` never fires so `ReferenceBackLinksViewController` never dismisses upon tap. I did add a gesture recognizer check in 0578c0d before returning true though just to target it a little better.

staykids: Ahhh – I gotcha. Sorry, I was still misunderstanding the different appearance modes of the references controller (with citation modal on in-article tap, without modal when triggered from citation) and wasn't actually even hitting the `tappedWebViewBackground` code path during testing. This led me to incorrectly believe the delegate conformance was unnecessary.

Looks good!

tonisevener: Thanks, no problem! Yeah the different reference view modes are a little confusing.

PR #3574, opened 2020-05-11
Improve iPad share sheet for share-a-fact

mcleinman:
With @carolynlimadeo's approval, we are now now using a width of roughly half the screen size for devices with regular width. 

I also updated when the share sheet comes in. It now comes in after the card completes it's brief animation. Even when updating the source rect during animation, on iPad the share sheet was staying at the bottom of the screen - with an arrow pointing to where the share-a-fact card began its animation. If the share sheet's source frame was updated and `setNeedsLayout` and `layoutIfNeeded` were called throughout animation, it still didn't animate any movement for the share sheet. In that case, at least the source rect was the final position of the animation - so the share sheet would pop and point to nothing, and the share-a-fact card would animate to where the share sheet's arrow was pointing.

**iPad**
Simulator Screen Shot - iPad Pro (9 7-inch) - 2020-05-11 at 13 57 27 Simulator Screen Shot - iPad Pro (9 7-inch) - 2020-05-11 at 13 57 13

**iPhone 8-sized screen**
Simulator Screen Shot - iPhone 8 - 2020-05-11 at 14 00 25  Simulator Screen Shot - iPhone 8 - 2020-05-11 at 14 00 12

**iPhone 11 Pro Max-sized screen**
Simulator Screen Shot - iPhone 11 Pro Max - 2020-05-11 at 13 58 42 Simulator Screen Shot - iPhone 11 Pro Max - 2020-05-11 at 13 58 39


tonisevener:
Looking good! Just one small thing - is it possible to keep the share sheet positioning when you rotate? I notice it gets thrown off a bit.

out

mcleinman:
> Looking good! Just one small thing - is it possible to keep the share sheet positioning when you rotate? I notice it gets thrown off a bit.
> 
> out

Good catch! Now working on rotation.

mcleinman:
Still needs one more review..

Wikipedia/Code/ShareViewController.swift
```diff
@@ -55,6 +55,16 @@ class ShareViewController: UIViewController, Themeable {
             self.showActivityViewController(with: image)
         }
     }
+
+    override func viewWillTransition(to size: CGSize, with coordinator: UIViewControllerTransitionCoordinator) {
+        super.viewWillTransition(to: size, with: coordinator)
+
+        coordinator.animate(alongsideTransition: { (context) in
+            if let presentedActivityVC = self.presentedViewController as? UIActivityViewController {
```
staykids, line 63: One thing I noticed re: rotation on the iPhone 11 Pro Simulator:

1. Start in portrait mode and waiting for share card to animate to top 
2. Rotate to landscape 
3. Rotate back to portrait

The share card returns to the center of the view instead of staying pinned to the top and is covered by the system share sheet. It might be `imageViewVerticallyCenteredConstraint` related, so one possible solution may be something like adding this here in `coordinator.animate(alongsideTransition: ...)`:
```swift
self.imageViewTopConstraint.isActive = true
self.imageViewVerticallyCenteredConstraint.isActive = false
```

iphone11-share

mcleinman: That fix does it. I spent a while digging in trying to figure out another way. It seems like installed/uninstalled constraints are guaranteed to remain [constant upon rotation](https://stackoverflow.com/questions/13739902/ios-change-auto-layout-constraints-when-device-rotates), but I can't find anything from Apple saying this. But it's definitely what's happening here. Further, I think it's just being reset when class sizes change - as on rotation on iPad (regular to regular) the problem doesn't occur. (It does occur when moving from split screen to full screen on iPad, which is more evidence that it's specifically related to class sizes. Working statement: When class sizes change on iOS, installed/uninstalled constraints revert to what exists on the xib.)

TL;DR: Now fixed, ready for another review.

staykids: Awesome – glad it was a small fix for such a strange issue.

PR #3578, opened 2020-05-13
For Explore feed options, use form sheet on iPad

mcleinman:
Fixes T251564.

**To test**
1. Run app on iPad.
1. Tap the three dots next to an explore feed card.
1. Tap Customize Explore Feed
1. Observe: The new screen (with many switches for various explore feed options) should not cover the entire screen on iPad - it should look like a window within iPad.
1. Repeat steps 2-4 on iPhone. It should work as always - cover nearly the entire screen.

**iPad Screenshots**
Simulator Screen Shot - iPad Pro (12 9-inch) (4th generation) - 2020-05-13 at 15 20 33 Simulator Screen Shot - iPad Pro (12 9-inch) (4th generation) - 2020-05-13 at 15 20 46

**iPhone 11 Max Pro Screenshots**
Simulator Screen Shot - iPhone 11 Pro Max - 2020-05-13 at 15 17 32 Simulator Screen Shot - iPhone 11 Pro Max - 2020-05-13 at 15 17 46


staykids:
Thanks for including the test steps! 👍

mcleinman:
> Nice! A much nicer more device-aware presentation of that view controller.
> 
> I anticipate in the future through more user testing we'll need to also add a way to dismiss this view controller by tapping outside of it (in the darkened overlay area) on iPad.

Yup, I thought about the "tap outside" issue - that felt like maybe scope creep for this issue, as (as of last time I checked) you need to roll your own with a gesture recognizer; we don't get that for free. Happy to add that in here if we want - let me know, could definitely include that.

PR #3580, opened 2020-05-15
Fix retain cycle on ArticleWebMessagingController

tonisevener:
https://phabricator.wikimedia.org/T251548

This keeps `ArticleWebMessagingController` instances from building up as you visit and leave articles.


tonisevener:
The original fix was small but I had to do some workarounds to get the tests to run. The lazy `messagingController` property initializer would kick off again in the view controller's `deinit`, causing a crash because the `ArticleViewController` was in the middle of being deallocated. But this only seemed to happen while the tests were run. I think my quick workarounds still preserve the functionality but I'm going to take off the Low Impact in case anyone else wants a deeper look.

tonisevener:
> Nice find. It appears `EditPreviewViewController` still exhibits the same issue re: `ArticleWebMessagingController` not deinitializing, but reading [this last comment](https://phabricator.wikimedia.org/T251548#6141253), it looked you were possibly considering we address that separately when investigating editing. Just wanted to confirm if that was the case or not so we don't miss it.

Thanks, yeah I hadn't looked into the editing flow at all yet but good catch - I went ahead and fixed it since we're touching `EditPreviewViewController` as a part of this anyway.

PR #3584, opened 2020-05-19
Pull article summaries from cache when device is offline

mcleinman:
This is a fix to the "when offline, cached summaries don't show in 3d touch's peek" problem. See **Testing** (below) for important info. I'm only partially confident in this solution, as I'm testing on simulator (no 3D touch devices in the home) and can only somewhat replicate this mode.

This is my first time messing with our caching layer - I spent quite a while digging in, and my understanding is that this change is pulling the summary from the cache, and falling back to making a request if it doesn't exist. (Please let me know if this is understanding is incorrect.) This brings us to a...

**Tradeoff**
When peeking, we're not checking for a new article summary if one is cached. So each time, we show the summary that is from the last time we cached the full article. (I confirmed that the summary gets cached with the full article load.) This does give a slight performance boost for cached article summaries, but at the cost of potentially slightly out-of-date data. Is this an acceptable tradeoff? (And if not, anyone have an idea on a cache policy for "prefer getting new article summary, but fall back to cache if no internet" option? Or do we need to create that ourselves?)

**Test Steps**
1. This PR can only be tested on [a device w/ 3D touch](https://en.wikipedia.org/wiki/Force_Touch#Products), or simulator of such a device. To be extra-clear: This will not be testable on a iPhone 11 simulator, but will be on something like an iPhone 8 simulator.
1. Save an article. Wait a few seconds to ensure it caches.
1. Turn off internet for the device. (Via airplane mode if on a real device. If on simulator, see below.)
1. On the Saved tab, force touch a saved article so that the peek (small view w/ article summary) appears over part of the Saved articles list, but don't press so hard that you pop into the entire article. Ensure you can see the article summary, and don't get an everlasting spinner in the peek mini-view.

**Testing notes**
This would be much better to test on a device. I have not been able to do so, and thus it's possible I've missed some nuances of this problem/fix.

Testing on simulator: To simulate offline/airplane mode on a simulator, turn off wifi on to your laptop. (First, of course, say a temporary goodbye to your streaming music.) I'm getting inconsistent behavior when doing this. There are times that when wifi is off, requests (of articles, article summaries, etc.) just end up seemingly waiting for a long timeout. I've found best results when quitting the simulator, turning off laptop wifi, and then launching the app. At that point, I believe you can toggle the laptop wifi at will.

*Example of article summary peek*

### DIFF
--- a/Wikipedia/Code/ArticlePeekPreviewViewController.swift
+++ b/Wikipedia/Code/ArticlePeekPreviewViewController.swift
@@ -33,7 +33,7 @@ class ArticlePeekPreviewViewController: UIViewController, Peekable {
             completion?()
             return
         }
-        dataStore.articleSummaryController.updateOrCreateArticleSummaryForArticle(withKey: key) { (article, _) in
+        dataStore.articleSummaryController.updateOrCreateArticleSummaryForArticle(withKey: key, cachePolicy: .returnCacheDataElseLoad) { (article, _) in
             defer {
                 completion?()
             }


tonisevener:
@mcleinman thanks for digging into this, it sounds extra painful with simulator so I'm happy to apply any changes and test on my device if you'd like. It looks like this does fix the issue while URLCache is populated - The problem is if the user removes that cache (App Settings > Clear cached data, which basically calls removeAllCachedResponses() on our URLCache instance), we start to see the spinner again so we need to make sure this response is persisted on device along with the saved article data. 

What you need to do is add the article summary url request to the `mustHaveURLRequests` array within `ArticleDBWriter`'s [add method](https://github.com/wikimedia/wikipedia-ios/blob/b2249f2740514c9829c689120973ac80dc0ba0d8/Wikipedia/Code/ArticleCacheDBWriter.swift#L43). Please build this URLRequest the same way the others are built, through `ArticleFetcher` where it leans on the `urlRequest(from url...)` [method](https://github.com/wikimedia/wikipedia-ios/blob/b2249f2740514c9829c689120973ac80dc0ba0d8/Wikipedia/Code/ArticleFetcher.swift#L291) as that adds extra things to the request that the system expects. That will persist the proper `CacheItem` in the Core Data database and the rest of the caching system should take it from there. Then test by unsaving and resaving the article, go to airplane mode, clear the cache via settings, then try the 3D touch.

You can revert the change you did here since it will cause those out-of-date data side-effects that you already predicted. The caching system is set up to work off of `eTag` so automatically it should* verify with the server if there's any change in the response, and if not it'll pull the response from our saved article data, otherwise it'll fetch the data from the server. And it should* also automatically fall back to the saved article data in airplane mode.

*I'm sure I'm missing a thing or two here so if you run into issues I'm happy to answer any questions or pair program on this too.

mcleinman:
> @mcleinman thanks for digging into this, it sounds extra painful with simulator so I'm happy to apply any changes and test on my device if you'd like. It looks like this does fix the issue while URLCache is populated - The problem is if the user removes that cache (App Settings > Clear cached data, which basically calls removeAllCachedResponses() on our URLCache instance), we start to see the spinner again so we need to make sure this response is persisted on device along with the saved article data.
> 
> What you need to do is add the article summary url request to the `mustHaveURLRequests` array within `ArticleDBWriter`'s [add method](https://github.com/wikimedia/wikipedia-ios/blob/b2249f2740514c9829c689120973ac80dc0ba0d8/Wikipedia/Code/ArticleCacheDBWriter.swift#L43). Please build this URLRequest the same way the others are built, through `ArticleFetcher` where it leans on the `urlRequest(from url...)` [method](https://github.com/wikimedia/wikipedia-ios/blob/b2249f2740514c9829c689120973ac80dc0ba0d8/Wikipedia/Code/ArticleFetcher.swift#L291) as that adds extra things to the request that the system expects. That will persist the proper `CacheItem` in the Core Data database and the rest of the caching system should take it from there. Then test by unsaving and resaving the article, go to airplane mode, clear the cache via settings, then try the 3D touch.
> 
> You can revert the change you did here since it will cause those out-of-date data side-effects that you already predicted. The caching system is set up to work off of `eTag` so automatically it should* verify with the server if there's any change in the response, and if not it'll pull the response from our saved article data, otherwise it'll fetch the data from the server. And it should* also automatically fall back to the saved article data in airplane mode.
> 
> *I'm sure I'm missing a thing or two here so if you run into issues I'm happy to answer any questions or pair program on this too.

@tonisevener Thanks for this, and apologies for the delay in getting back to this one.

To make sure I understand before writing this all up (and what it entails), this is the approach you're suggesting, right?

1) In `ArticleCacheCBWriter.add`, create an article summary request in the same way that we create a `mobileHTMLRequest` and `mobileHTMLMediaListRequest`.  Append that request to `mustHaveURLRequests`.
2) To do #1, we'd need to add an `ArticleSummaryFetcher` to this class.
3) To do #1, we'd also need to refactor `ArticleSummaryFetcher` so that we can pull a request out of it. Currently `ArticleSummaryFetcher.fetchSummaryForArticle` takes a bunch of arguments, and passes them on to `Fetcher.performRESTBaseGET` - which in turn calls `session.jsonDecodableTask` to create a `URLSessionTask`. In order to play nice with `mustHaveURLRequests`, we need to refactor this to create a `URLRequest` earlier up the chain, in part so that it can also be used in this `ArticleCacheCBWriter`.

This all makes sense (and I believe is possible), but includes a mini-refactor - which is why I wanted to confirm before going down this path. Let me know - is this what you had in mind? Thanks!

tonisevener:
> > @mcleinman thanks for digging into this, it sounds extra painful with simulator so I'm happy to apply any changes and test on my device if you'd like. It looks like this does fix the issue while URLCache is populated - The problem is if the user removes that cache (App Settings > Clear cached data, which basically calls removeAllCachedResponses() on our URLCache instance), we start to see the spinner again so we need to make sure this response is persisted on device along with the saved article data.
> > What you need to do is add the article summary url request to the `mustHaveURLRequests` array within `ArticleDBWriter`'s [add method](https://github.com/wikimedia/wikipedia-ios/blob/b2249f2740514c9829c689120973ac80dc0ba0d8/Wikipedia/Code/ArticleCacheDBWriter.swift#L43). Please build this URLRequest the same way the others are built, through `ArticleFetcher` where it leans on the `urlRequest(from url...)` [method](https://github.com/wikimedia/wikipedia-ios/blob/b2249f2740514c9829c689120973ac80dc0ba0d8/Wikipedia/Code/ArticleFetcher.swift#L291) as that adds extra things to the request that the system expects. That will persist the proper `CacheItem` in the Core Data database and the rest of the caching system should take it from there. Then test by unsaving and resaving the article, go to airplane mode, clear the cache via settings, then try the 3D touch.
> > You can revert the change you did here since it will cause those out-of-date data side-effects that you already predicted. The caching system is set up to work off of `eTag` so automatically it should* verify with the server if there's any change in the response, and if not it'll pull the response from our saved article data, otherwise it'll fetch the data from the server. And it should* also automatically fall back to the saved article data in airplane mode.
> > *I'm sure I'm missing a thing or two here so if you run into issues I'm happy to answer any questions or pair program on this too.
> 
> @tonisevener Thanks for this, and apologies for the delay in getting back to this one.
> 
> To make sure I understand before writing this all up (and what it entails), this is the approach you're suggesting, right?
> 
> 1. In `ArticleCacheCBWriter.add`, create an article summary request in the same way that we create a `mobileHTMLRequest` and `mobileHTMLMediaListRequest`.  Append that request to `mustHaveURLRequests`.
> 2. To do #1, we'd need to add an `ArticleSummaryFetcher` to this class.
> 3. To do #1, we'd also need to refactor `ArticleSummaryFetcher` so that we can pull a request out of it. Currently `ArticleSummaryFetcher.fetchSummaryForArticle` takes a bunch of arguments, and passes them on to `Fetcher.performRESTBaseGET` - which in turn calls `session.jsonDecodableTask` to create a `URLSessionTask`. In order to play nice with `mustHaveURLRequests`, we need to refactor this to create a `URLRequest` earlier up the chain, in part so that it can also be used in this `ArticleCacheCBWriter`.
> 
> This all makes sense (and I believe is possible), but includes a mini-refactor - which is why I wanted to confirm before going down this path. Let me know - is this what you had in mind? Thanks!

Thanks for pointing that out before going down it. It is more involved than I realized but you are right in that this will be what is needed. Unfortunately this will run into a problem of some of the legacy Session methods (one of which summary fetcher uses) not considering the newer caching pieces that I added for mobile-html. It's something I've been meaning to clean up but it could end up being a pretty high-impact refactor for a bug fix release like this, it may fit better as an early-on refactor in the 6.7 release cycle. Given that I'll mark this as hold for now and ask @jtminor what he thinks.

mcleinman:
Closing as the real fix will be done on a future PR - left link to this on the Phab ticket (https://phabricator.wikimedia.org/T247841) so we can find these notes later on.

PR #3589, opened 2020-05-27
Check survey announcement end time before presenting survey

staykids:
https://phabricator.wikimedia.org/T253156

Checks announcement end time before presenting survey.

Also adds a little `Date` extension with tests to make date comparison a little easier to read and reason in my opinion (`currentDate.isBefore(endDate)` vs `currentDate < endDate`), but @tonisevener I'm totally happy to rip it out if you'd prefer we just use comparison operators.

### DIFF
--- a/Wikipedia/Code/ArticleViewController+SurveyAnnouncements.swift
+++ b/Wikipedia/Code/ArticleViewController+SurveyAnnouncements.swift
@@ -3,8 +3,8 @@ import Foundation
 extension ArticleViewController {
 
     func showSurveyAnnouncementPanel(surveyAnnouncementResult: SurveyAnnouncementsController.SurveyAnnouncementResult) {
-        
-        guard state == .loaded else {
+        let currentDate = Date()
+        guard state == .loaded, let surveyEndTime = surveyAnnouncementResult.announcement.endTime, currentDate.isBefore(surveyEndTime) else {
             return
         }
         
--- a/Wikipedia.xcodeproj/project.pbxproj
+++ b/Wikipedia.xcodeproj/project.pbxproj
@@ -3206,7 +3206,7 @@
 /* Begin PBXFileReference section */
 		00A7946A245CA4E60063BA18 /* ArticleSurveyTimerController.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ArticleSurveyTimerController.swift; sourceTree = "<group>"; usesTabs = 0; };
 		00D280F6247EFFFE006BEE23 /* Date+Extensions.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "Date+Extensions.swift"; sourceTree = "<group>"; usesTabs = 0; };
-		00D280FB247F019C006BEE23 /* Date+ExtensionTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "Date+ExtensionTests.swift"; sourceTree = "<group>"; };
+		00D280FB247F019C006BEE23 /* Date+ExtensionTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "Date+ExtensionTests.swift"; sourceTree = "<group>"; usesTabs = 0; };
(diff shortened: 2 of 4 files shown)


Wikipedia/Code/ArticleViewController+SurveyAnnouncements.swift
```diff
@@ -3,8 +3,8 @@ import Foundation
 extension ArticleViewController {
 
     func showSurveyAnnouncementPanel(surveyAnnouncementResult: SurveyAnnouncementsController.SurveyAnnouncementResult) {
-        
-        guard state == .loaded else {
+        let currentDate = Date()
+        guard state == .loaded, let surveyEndTime = surveyAnnouncementResult.announcement.endTime, currentDate.isBefore(surveyEndTime) else {
```
mcleinman, line 7: Will a survey ever not have an end time? Since that's optional, is there ever a chance it would intentionally resolve to nil? If so, it would always return.

staykids: ### **Testing**

The most complete way to test the surveys would be [Toni's steps here](https://github.com/wikimedia/wikipedia-ios/pull/3518#issue-399963175) - `announcement.js` may need to be locally adjusted to extend the end time. But I've only been testing against [a locally stubbed](https://github.com/wikimedia/wikipedia-ios/pull/3554#issue-411912482) `surveyAnnouncementResult` in order to test the client side logic and appearance.

### **Survey End Time**

Good question! The goal of this PR is to help avoid unintentional presentation of the survey modal – one such case is a user reporting it being presented after the end time had passed. Based on my limited understand of the survey stuff – Short answer: in this situation, it's desirable to `return` if `endTime` is nil. Longer answer: even though `endTime` has the nullable annotation, in use an announcement is considered malformed if its `endTime` is nil. This understanding originates from the [original implementation here](https://github.com/wikimedia/wikipedia-ios/blob/b16f52e8db9c0e9486fe450e5bbcc1a387c178d3/WMF%20Framework/SurveyAnnouncementsController.swift#L59). `WMFAnnouncement` itself appears to have a longer history, so it doesn't seem worth redefining it's interface to support our special survey case.

tonisevener: @mcleinman Another option for testing is to locally map the `en.wikipedia.org/api/rest_v1/feed/announcements` endpoint via Charles. I mapped to a file containing this, which is based on the last time I pulled it from the wikifeeds master branch in this [PR](https://github.com/wikimedia/wikipedia-ios/pull/3577). Just tweak the `start_time` and `end_time` to your liking.

```
{
	"announce": [{
		"type": "survey",
		"start_time": "2020-05-29T00:00:00Z",
		"end_time": "2020-05-29T17:13:00Z",
		"domain": "en.wikipedia.org",
		"id": "IOSSURVEY20IOSV3EN",
		"platforms": ["iOSAppV3"],
		"border": false,
		"placement": "article",
		"text": "Help improve Wikipedia by taking a quick survey about this article",
		"caption_HTML": "View our <a href='https://foundation.m.wikimedia.org/w/index.php?title=Editing_Awareness_and_Trust_Survey_Privacy_Statement'>privacy statement</a>. Survey powered by a third-party. View their <a href='https://policies.google.com/privacy'>privacy policy</a>.",
		"negative_text": "No thanks",
		"action": {
			"title": "Start survey",
			"url": "https://docs.google.com/forms/d/e/1FAIpQLSeoMkmpSHiolH5tmEdbpbtbmphXFCWqyGlqj6GBWrOHV-P1hQ/viewform?entry.366807840={{articleTitle}}"
		},
		"articleTitles": ["Internet", "Chernobyl disaster", "Advertising", "United States", "Wikipedia", "Lists of deaths by year#2019", "Parasite (2019 film)", "Deaths in 2020", "2020 Democratic Party presidential primaries", "2019–20 coronavirus pandemic"],
		"displayDelay": 15
	}]
}
```

mcleinman: I'd love to try to test this end-to-end. I'm having trouble - it doesn't seem to be hitting my local wikifeeds server. I can hit the endpoint just fine in a browser (and see the expected announcements JSON), but when running the local PCS/announcements scheme, it doesn't hit that server. Any idea how my setup needs to be tweaked?

PR #3590, opened 2020-05-27
Add ability to update saved article cache under a custom url header

tonisevener:
Fixes bugs in https://phabricator.wikimedia.org/T252435

This was tweaked from the original reverted 6.6 work in https://github.com/wikimedia/wikipedia-ios/pull/3564/commits/0ad47a79da3e619e43b187c3d6fd9bd8a6111e26. I worry that messing with wider-reaching methods like `itemKeyForURLRequest` might have unintended side-effects, so I modified this to be very light-touch.

**Testing steps:**

**Flow 1:**

1. Edit a saved article's content
2. Save the edit
3. Once article is refreshed, you will see your new edit in the article content. Now enter airplane mode.
4. Tap Back to leave the article, go to the Saved tab and return to the article
5. You will see your new edit in the article content. In 6.6 you would not see your edit, it showed the older article version.

**Flow 2:**

1. Edit a saved article's content
2. Save the edit
3. Once article is refreshed, you will see your new edit in the article content. Terminate the app.
4. Launch the app, navigation state should be restored and you will be back on your recently edited article. In this situation the app leans on the local cache for the mobile-html endpoint so the user sees the content as quickly as possible. You will see your new edit in the article content. In 6.6 you would not see your edit, it showed the older article version.


mcleinman:
Code looks good, from what I know about our caching. Do you have an idea on the best way to test this?

tonisevener:
> Code looks good, from what I know about our caching. Do you have an idea on the best way to test this?

Sorry about that - just added testing steps in the description.

mcleinman:
> > Code looks good, from what I know about our caching. Do you have an idea on the best way to test this?
> 
> Sorry about that - just added testing steps in the description.

Those two tests work great.

Just one small thing, and if it's not easy to update in this PR I'll happily approve this one - When editing something early in the article, the snippet shown on the main Saved page doesn't update. (You have to update something in the first line or so to see this.) If it's easy to refresh the data for that table view, can we do that?

tonisevener:
> > > Code looks good, from what I know about our caching. Do you have an idea on the best way to test this?
> > 
> > 
> > Sorry about that - just added testing steps in the description.
> 
> Those two tests work great.
> 
> Just one small thing, and if it's not easy to update in this PR I'll happily approve this one - When editing something early in the article, the snippet shown on the main Saved page doesn't update. (You have to update something in the first line or so to see this.) If it's easy to refresh the data for that table view, can we do that?

Great catch! I looked into this, unfortunately it seems more involved than that. It does look like the views properly update if the summary endpoint returned the right snippet after the page reloads. This looks like another server-side update delay problem that we tackled in this [PR](https://github.com/wikimedia/wikipedia-ios/pull/3564). I made a separate [task](https://phabricator.wikimedia.org/T254407) for this that we can triage (along with a quick and dirty [fix](https://github.com/wikimedia/wikipedia-ios/commit/23ad228903f65fe759e2df3147f72644322bfd18)).

tonisevener:
This is ready for one more review.

PR #3593, opened 2020-05-29
Add margins for iPad, fix a related iPad layout issue

mcleinman:
Bug: https://phabricator.wikimedia.org/T252918

Other bug found and fixed in this PR: When moving from Table of Contents (TOC) to no-TOC, iPad adds some leading and trailing padding to the image. However, we were never removing the padding if you brought the TOC back. Now we are.

We only use readable width when the TOC isn't shown on iPad. That looks true from screenshots of prior version, and we kept that here. In part because finding the readable width when TOC is on screen proved difficult - you can't use the navcontroller or VC view, as it is wider than the article view. The article's webView returns the wrong readable width, as did the image header view.

The vertical offset placement got moved so it works both with rotation as well as margins changing.

Testing:
- Test turning TOC on and off. (Make sure we have margins, and also make sure the image margins work appropriately.)
- Tap into another article, to make sure margins stay when moving into a new ArticleVC.
- Bring another app on screen for multitasking in various widths - make sure margins appear/disappear as appropriate.
- Rotate the screen, and do all this after rotation.
- Compact sized phones should look identical. But we should get some margins on large phones in landscape, as they are Regular.


tonisevener:
Looks good on iPad - I do feel like my reading place gets lost a bit when I rotate and show/hide the TOC but I think that was already there to some extent. Might be worth trying to improve while you're in there. On iPhone it doesn't seem to scroll to the right section anymore when tapping a TOC title (I'm testing the today's featured article). iPad scrolls to the section fine but there seems to have been a regression on my iPhone 7 (iOS 13):

out

mcleinman:
> Looks good on iPad - I do feel like my reading place gets lost a bit when I rotate and show/hide the TOC but I think that was already there to some extent. Might be worth trying to improve while you're in there. On iPhone it doesn't seem to scroll to the right section anymore when tapping a TOC title (I'm testing the today's featured article). iPad scrolls to the section fine but there seems to have been a regression on my iPhone 7 (iOS 13):
> 
> out

For "keeping the right place", after some testing it looks like this is something many apps struggle with when changing text size and orientation. I played with this a bit in the original PR, and the problem is that we use the scrollView's `verticalOffsetPercentage`. However, verticalOffset doesn't track perfectly when the article changes (orientation or text size), presumably because some things like images and tables have the same height, while paragraphs of text change height.

I found the issue on iPhone - good catch. I know where/why it's happening, will work on the best fix tomorrow morning.

mcleinman:
I added code to fix the iPhone problem.

I then spent quite a while digging more into the issue with the "drift" when adding/removing margins. It's worse than the last version, but also the last version didn't have margins, and this is possibly related. The drift is worse the further down the article you are, which makes sense as we're using the percent down the article that it is.

I dug in, and the problem is the webView's contentSize isn't being recalculated before we're re-applying the scroll percent. So effectively our work to re-set the vertical offset percent isn't doing anything.

We need a way for the webView to tell us when it's done recalculating it's content after we send the two JS messages (`setMargins` and the add'l setup call). 
- I tried placing the code to set the verticalOffsetPercentage in the completion for the JS call for setMargin. Didn't work.
- `func webView(WKWebView, didFinish: WKNavigation!)` does not get called when JS is finished running on an existing page, from my tests. If it would, we could put the code there.
- Of course, I've tried setting `setNeedsLayout` and `layoutIfNeeded` to force a layout pass. 
- I've also tried the ideas in [this post](https://stackoverflow.com/questions/17257629/force-uiwebview-to-redraw), and while those affect the behavior, only about half of the code is run before the verticalOffsetPercentage is run, so it doesn't fix the problem.

Any idea on how to trigger the reset off the percent difference after the webView recalculates its size?

tonisevener:
> I added code to fix the iPhone problem.
> 
> I then spent quite a while digging more into the issue with the "drift" when adding/removing margins. It's worse than the last version, but also the last version didn't have margins, and this is possibly related. The drift is worse the further down the article you are, which makes sense as we're using the percent down the article that it is.
> 
> I dug in, and the problem is the webView's contentSize isn't being recalculated before we're re-applying the scroll percent. So effectively our work to re-set the vertical offset percent isn't doing anything.
> 
> We need a way for the webView to tell us when it's done recalculating it's content after we send the two JS messages (`setMargins` and the add'l setup call).
> 
> * I tried placing the code to set the verticalOffsetPercentage in the completion for the JS call for setMargin. Didn't work.
> * `func webView(WKWebView, didFinish: WKNavigation!)` does not get called when JS is finished running on an existing page, from my tests. If it would, we could put the code there.
> * Of course, I've tried setting `setNeedsLayout` and `layoutIfNeeded` to force a layout pass.
> * I've also tried the ideas in [this post](https://stackoverflow.com/questions/17257629/force-uiwebview-to-redraw), and while those affect the behavior, only about half of the code is run before the verticalOffsetPercentage is run, so it doesn't fix the problem.
> 
> Any idea on how to trigger the reset off the percent difference after the webView recalculates its size?

`ArticleViewController.swift` has a `contentSizeDidChange` callback that should to be triggered when the web view's content size changes. You can try triggering the reset there (or in `debouncedContentSizeDidChange` rather since `contentSizeDidChange` could be called many times in one user action).

staykids:
> We need a way for the webView to tell us when it's done recalculating it's content after we send the two JS messages (`setMargins` and the add'l setup call).
> ...
> * I've also tried the ideas in [this post](https://stackoverflow.com/questions/17257629/force-uiwebview-to-redraw), and while those affect the behavior, only about half of the code is run before the verticalOffsetPercentage is run, so it doesn't fix the problem.

Catching up on this – thanks for posting your research and experimentation on it so far. Just wanted to poke in to say that in my prodding of https://phabricator.wikimedia.org/T253031 your comment about the `WKWebView`'s content characteristics not recalculating when expected feels related to what I'm running into.

By doing something like `webView.evaluateJavaScript("window.scrollBy(0, 1); window.scrollBy(0, -1)")` in `debouncedContentSizeDidChange()` or `handlePCSDidFinishFinalSetup()`, I can persuade the initial TOC taps to scroll where expected, but it's not ideal. Gonna keep digging...

mcleinman:
Should be ready for another look. I reworked a bunch of this - more similar to how it was originally, but now the ArticleVC is responsible for keeping track of its offset percentage.

On iPad, when inline Table of Contents is being added or removed you can see a slight jump in the content. This is also present in 6.5.1, and I couldn't find a good way to get rid of it here.

mcleinman:
> > We need a way for the webView to tell us when it's done recalculating it's content after we send the two JS messages (`setMargins` and the add'l setup call).
> > ...
> > 
> > * I've also tried the ideas in [this post](https://stackoverflow.com/questions/17257629/force-uiwebview-to-redraw), and while those affect the behavior, only about half of the code is run before the verticalOffsetPercentage is run, so it doesn't fix the problem.
> 
> Catching up on this – thanks for posting your research and experimentation on it so far. Just wanted to poke in to say that in my prodding of https://phabricator.wikimedia.org/T253031 your comment about the `WKWebView`'s content characteristics not recalculating when expected feels related to what I'm running into.
> 
> By doing something like `webView.evaluateJavaScript("window.scrollBy(0, 1); window.scrollBy(0, -1)")` in `debouncedContentSizeDidChange()` or `handlePCSDidFinishFinalSetup()`, I can persuade the initial TOC taps to scroll where expected, but it's not ideal. Gonna keep digging...

@staykids I think Toni's comment above about `contentSizeDidChange` is key - that is called after the size changed - so if you set up the work before hand, I think you can guarantee it's being done on the new content size.

mcleinman:
> > On iPad, when inline Table of Contents is being added or removed you can see a slight jump in the content. This is also present in 6.5.1, and I couldn't find a good way to get rid of it here.
> 
> I can't either. I took a video of closing TOC and slowed it way down, it seems like these layout changes happen:
> 
>     1. TOC hides, article content shifts to the left, content width stays the same.
> 
>     2. Article content left margin is added
> 
>     3. Article content expands right out to the right margin
> 
>     4. Content offset is shifted to the previous offset percentage.
> 
> 
> I confirmed that `verticalOffsetPercentage` scrollView setter is only called once and `updateMargins` is only called once with the correct `marginsJSON`. Maybe it's on the PCS side? Or we could hide the content somehow until the content offset call kicks in. If it's in 6.5.1 too, I say we file a task and look into it further in the next release.
> 

Thanks for digging into that. I think it's probably a tough fix, given that we need to coordinate the JS-driven web UI update and the TOC-removing native UI to happen simultaneously. Or possibly just coordinate the JS-driven web UI update (if the jump is from the applying margins and setting the vertical scroll percentage). 

Since neither of us can easily find an answer, let's kick it to a future ticket.

staykids:
Apologies if I'm missing context from the existing comments/ticket and this is expected, but it looks like the updated text content and lead image margins here differ from those present in 6.5.1 (using today's featured article).

margins

tonisevener:
@staykids great catch! I hadn't realized the old code didn't use `readableContentGuide`, which seems more proper to me. @carolynlimadeo how does this change (see Deepak's screenshot above) look to you? The text margins should now match the other list screens for iPad since now we're using the same Apple-provided readable margin. It looks like the top image also changed to full-width in this PR, that should be easy to fix to match the new text margins but I suggest we keep the new text margins themselves the same as in the T252918 screenshot.

mcleinman:
Great catch @staykids! Will wait for @carolynlimadeo to chime in on her preferred solution.

One sub-question for @carolynlimadeo: If we're going with the "image should also have margins" solution, do we also want the additional image padding? The image has 32 pts of padding on each side when on an iPad and TOC isn't currently shown. (This was existing code.) On the image on the left, you can see the image is slightly narrower than the text. Is this how we'd want it?

carolynlimadeo:
Thanks Toni, Deepak and Matt! Text margins are looking great! Is it possible to have the image margins match the text margins? (eg remove the 32 pts of padding but don't go full width)? The full width looks great with this image but I'm worried we're going to get some weird lookin' stuff with the wider ratio on a lot of our images.

mcleinman:
Updated image margins.

Screenshot 8
Screenshot 9
Screenshot 10
Screenshot 11
Screenshot 12
Screenshot 13

mcleinman:
Need 
```if doesArticleUseLargeMargin {
            marginWidth = 32
            marginWidth = articleHorizontalMargin
        }
        }```

because otherwise on compact devices (like most phones) there is a small margin on the image, as even phones have slight margins for text. But in those cases, we want the image to go edge-to-edge (I believe).

PR #3595, opened 2020-06-01
Disable full screen presentation of long-pressed timeline images

staykids:
https://phabricator.wikimedia.org/T252922

## Existing Behavior

I unexpectedly had to do a deep dive to get a better idea of what's up here, but I could still be missing something completely obvious. From my research and experimentation so far, here's my current understanding of what was happening. With "timeline" images in an article (that aren't "links" or included in the page's [media list](https://en.wikipedia.org/api/rest_v1/page/media-list/)), long-pressing them triggers WebKit's default link previewing/context menu instead of our custom one – it doesn't trigger [this delegate method](https://github.com/wikimedia/wikipedia-ios/blob/83961737b9868613e1ffea6366d63510bb082dd2/Wikipedia/Code/ArticleViewController%2BLinkPreviewing.swift#L197). It seems we aren't alone in experiencing this custom context menu issue. Tapping the timeline image preview in this default context menu results in the user being stuck in a non-scrollable or dismissible full screen view.

## Testing

Here are a few articles you can reproduce the issue with: 
https://en.wikipedia.org/wiki/Terrace_House:_Opening_New_Doors
https://en.wikipedia.org/wiki/1918_Atlantic_hurricane_season
https://en.wikipedia.org/wiki/Timeline_of_computing

1. Open any article with a graphical timeline
2. Long-press the timeline image
3. Tap the preview of the image in the context menu
4. Observe it now dismisses the context menu preview instead of pushing into a non-dismissible view controller like before

## Question

It's currently WIP because I'm not sure how to handle the share case. When trying to share a timeline image (long-press a timeline image to show context menu, tap Share), it's trying to share our scheme handled URL (`app://`) instead of either the resolved resource URL or just resource itself. This isn't a 6.6 regression - it occurred with 6.5.1 as well.

Maybe an easy solve is if `mobile-html` just prepended the `https` protocol for timeline URLs? I'm unclear what would be best – any ideas @tonisevener?


staykids:
Should also mention – could maybe [-webkit-touch-callout: none](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-touch-callout) the entire `timeline-wrapper` div to prevent the long-press itself (if that's what it even does – I haven't tested that CSS property yet myself).

tonisevener:
@staykids well my first thought was we can't switch to `https` because then the fallback in offline mode would no longer work - it wouldn't hook into our `SchemeHandler` and pull the saved article timeline. Then I tested it and it's never saving the timeline image anyways when you save an article 🤦‍♀️. This is because it's not represented in the `media-list` endpoint - we use that to get the list of image resources to save. I'm not sure why, it seems like there's a `showInGallery` bool returned there so we could lean on that to filter it out (but another bug is that we currently don't seem to take that boolean into consideration anywhere in the codebase). I have made a bug task for this problem [here](https://phabricator.wikimedia.org/T254290).

Even with all that, I'd still rather keep this scheme as `app://` in the mobile-html side because mixing and matching schemes sounds like a mess and if we can get timeline images returned in media-list and saving on device, `app://` would allow it to work in offline mode.

I played with your `-webkit-touch-callout: none` option on the `timeline-wrapper` div and it does seem to hide the context menu options. I think that's our best choice. Another thing that could work is wrapping the `timeline-wrapper` div in an `a` tag with the same `href` as the `img` tag `src`. It allows for that delegate callback and hides the context menu options, as well as taps to an app-side webview of the timeline & allows the user to zoom in and out, which is a more useful experience IMO. But that would affect Android and may be stepping on editors' toes too much. Either way this will need PCS changes seems like.

out

mcleinman:
Just looked at this for a bit. I don't see anything obvious that you missed, other than maybe wrapping this (which Toni mentioned, and seems a bit heavy handed to me?). This stuff is weird sometimes.

Happy to help w/ PCS changes if needed.

staykids:
> I played with your `-webkit-touch-callout: none` option on the `timeline-wrapper` div and it does seem to hide the context menu options. I think that's our best choice. 

@tonisevener Thanks for poking at this and making that ticket. I agree with you, my scheme mixing proposal is gross. I'm psyched to see that the `-webkit-touch-callout: none` option works. I feel it might be undesirable to `<a>` tag it all – it's definitely advantageous to be able to zoom in/out on the timeline image, but many of them have image maps that would be affected and wouldn't work in the pushed image-only web view. Showing the timeline in the gallery would expose the same image map issue.

I want to find more of a variety of timeline articles and do some more tinkering, but it seems like the `-webkit-touch-callout` route you tested – which would at least disallow the user from sharing non-functional links – would cover our purposes until something more elegant can be devised for the timelines. 

> Happy to help w/ PCS changes if needed.

Thank you for offering help on this, @mcleinman. Because I still haven't set up the local PCS stuff, I'm gonna take you up on that if you don't mind. Will drop a comment in Phab after I get a chance to explore this more.

mcleinman:
@staykids Sounds good. Feel free to assign the Phab ticket to me!

staykids:
> @staykids Sounds good. Feel free to assign the Phab ticket to me!

Done: https://phabricator.wikimedia.org/T254781. Lemme know if you have any questions about it I can answer to help make the solution effortless. Thanks Matt! 

I'm going to remove the WIP label here. Even though the mobile-html fix will prevent a context menu for this particular element, I think checking to confirm a `linkURL` exists might still be helpful in preventing the stuck state for any other elements we're not immediately aware of.

PR #3598, opened 2020-06-04
fix SectionEditorWebViewMessagingController retain cycle

tonisevener:
This wraps up https://phabricator.wikimedia.org/T251548.

**Testing steps:**

1. Open Instruments
2. Filter allocations by `SectionEditorWebViewMessagingController`.
3. Edit an article, complete flow until you're back on the article screen and it has reloaded.
4. 6.6 would still have a `SectionEditorWebViewMessagingController` instance hanging around in memory, this PR fixes that.


Wikipedia/Code/SectionEditorViewController.swift
```diff
@@ -136,6 +147,31 @@ class SectionEditorViewController: ViewController {
...
+        for handler in scriptMessageHandlers {
+            userContentController.removeScriptMessageHandler(forName: handler.name)
+        }
+        
+        scriptMessageHandlers.removeAll()
+    }
+    
+    private func addScriptMessageHandlers(to contentController: WKUserContentController) {
+        
+        guard !scriptMessageHandlers.isEmpty else {
```
mcleinman, line 166: Not a bug, but is this necessary?  If the array is empty, we just won't iterate over anything, no?

Wikipedia/Code/SectionEditorViewController.swift
```diff
@@ -136,6 +147,31 @@ class SectionEditorViewController: ViewController {
         navigationController?.setNavigationBarHidden(false, animated: false)
     }
     
+    private func removeScriptMessageHandlers() {
+        
+        guard let userContentController = webView?.configuration.userContentController,
+            !scriptMessageHandlers.isEmpty else {
```
mcleinman, line 153: same as below, not sure if we need this check - but not a problem to keep it

mcleinman:
Will let you respond before I merge it - happy to merge it as-is.

Wikipedia/Code/SectionEditorViewController.swift
tonisevener: Good point! I removed this.

Wikipedia/Code/SectionEditorViewController.swift
tonisevener: Fixed.

PR #3603, opened 2020-06-10
Event Platform Client for analytics powered by Modern Event Platform

bearloga:
This patch adds the iOS version of the [Event Platform Client](https://www.mediawiki.org/wiki/Wikimedia_Product/Analytics_Infrastructure/Event_Platform_Client) (EPC-iOS) for integration with the modern [Event Platform](https://wikitech.wikimedia.org/wiki/Event_Platform). EPC is an effort to unify algorithms, identifiers, and behaviors across MediaWiki, iOS, and Android.

This initial release supports:
- [remote sampling configurations](https://www.mediawiki.org/wiki/Wikimedia_Product/Analytics_Infrastructure/Stream_configuration#Sampling_settings), which puts sampling rates into the centralized stream configuration
- [stream cc-ing](https://www.mediawiki.org/wiki/Wikimedia_Product/Analytics_Infrastructure/Stream_configuration#Stream_cc-ing), which automatically logs events to multiple streams without requiring multiple calls on the instrumentation side

For additional background information on instrumentation within the Modern Event Platform, please refer to [this page](https://wikitech.wikimedia.org/wiki/Event_Platform/Instrumentation_How_To).

Phabricator: [T228180](https://phabricator.wikimedia.org/T228180)


WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
```diff
@@ -0,0 +1,605 @@
...
+     *
+     * The identifier is a string of 20 zero-padded hexadecimal digits representing a uniformly random
+     * 80-bit integer.
+     */
+    public func session_id() -> String {
+        if SESSION_ID == nil {
+            SESSION_ID = generate_id()
+        }
+        return SESSION_ID!
+    }
```
tonisevener, line 291: This method looks more like a computed property to me. A common pattern in our codebase is to have an underlying stored property that holds the value, and a computed property that controls logic like this. You'll want to rename the `SESSION_ID` property with `_SESSION_ID`, then replace this with:

```
    private var SESSION_ID: String {
        get {
            guard let sessionID = _SESSION_ID else {
                let newID = generate_id()
                _SESSION_ID = newID
                return newID
            }
            
            return sessionID
        }
    }
```

Then you can replace `session_id()` method calls with a cleaner `SESSION_ID` property. Note you will also have to change `SESSION_ID = nil` to `_SESSION_ID = nil` in `begin_new_session()`.

WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
```diff
@@ -0,0 +1,605 @@
...
+
+    /**
+     * Returns the app install ID stored on the device
+     */
+    public func device_id() -> String {
```
tonisevener, line 296: This also feels computed propertyish, though since we're referencing `EventLoggingService` here, I think maybe we should instead add a device_id property to the StorageManager protocol so EPC doesn't have to worry about where it's coming from.

```
public protocol StorageManager {
    func set_persisted(_ key: String, _ value: Any)
    func del_persisted(_ key: String)
    func get_persisted(_ key: String, completion: (Any?) -> Void)
    var device_id: String
}
```

It may ultimately come from EventLoggingService or we may split it off into a separate protocol from StorageManager, but for now I think this would be good enough.

WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
```diff
@@ -0,0 +1,605 @@
...
+    private var network_manager: NetworkManager?
+
+    /**
+     * Store events until the library is finished initializing
+     *
+     * The EPC library makes an HTTP request to a remote stream configuration service for information
+     * about how to evaluate incoming event data. Until this initialization is complete, we store any incoming
+     * events in this buffer.
+     */
+    private var input_buffer = [(stream: String, schema: String, data: [String: Any], domain: String?)]()
```
tonisevener, line 165: I think we will need to take some extra efforts to make sure `input_buffer` and `cache` are thread-safe. There are multiple ways to do this - `ImageControllerCompletionManager.swift` has an example with it's `completions` and `tasks` properties. You make a serial dispatch queue property, then anytime you're reading or writing to `input_buffer` or `cache`, you wrap it up in a `queue.async`. Happy to answer any questions you have on this or pitch in if you run into trouble.

WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
```diff
@@ -0,0 +1,605 @@
...
+     * - Parameter acceptance: the desired proportion of many `token`-s being accepted
+     *
+     * The algorithm works in a "widen the net on frozen fish" fashion -- tokens continue evaluating to
+     * true as the acceptance rate increases. For example, a device determined to be in-sample for a
+     * stream "A" having rate 0.1 will be determined to be in-sample for a stream "B" having rate 0.2,
+     * and its events will show up in tables "A" and "B".
+     */
+    private func determine(_ id: String, _ acceptance: Double) -> Bool {
+        let token: UInt32 = UInt32(id.prefix(8), radix: 16)!
+        return (Double(token) / Double(UInt32.max)) < acceptance
```
tonisevener, line 401: Small suggestion to avoid the `!`:

```suggestion
        guard let token = UInt32(id.prefix(8), radix: 16) else {
            return false
        }
        
        return (Double(token) / Double(UInt32.max)) < acceptance
```

WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
```diff
@@ -0,0 +1,605 @@
...
+                            }
+                            identifier = identifier == "session" ? session_id() : device_id()
+                            let rate = sampling_config["rate"]! as! Double
+                            cache[stream] = determine(identifier, rate)
+                        } // end if there's identifier in sampling config
+                    } // end if there's rate in sampling config
+                } // end if there's a sampling config
+            } // end if there's a config
+        } // end if there's no cached determination
+        return cache[stream]!
```
tonisevener, line 458: Here's a suggestion to avoid the extra nesting & `!`s:

```
        guard let config = config else {
            DDLogDebug("Invalid state, must have config to check for inSample")
            return false
        }
        
        if let cachedValue = cache[stream] {
            return cachedValue
        }

        guard let streamEntry = config[stream],
            let streamConfig = streamEntry["config"] as? [String: Any],
            let samplingConfig = streamConfig["sampling"] as? [String: Any] else {
                DDLogDebug("Failure pulling sampling key from config.")
                return false
        }
        
        /*
        * If cache does not have a determination for stream,
        * generate one.
        */
        guard let rate = samplingConfig["rate"] as? Double else {
            
            /*
            * If stream doesn't have a rate, assume 1.0 (always
            * in-sample). Cache this determination for any future
            * use.
            */
            cache[stream] = true
            return true
        }
        
        /*
        * All platforms use session ID as the default
        * identifier for determining in- vs out-of-sample of
        * events sent to streams. On the web, streams can be
        * set to use pageview token instead. On the apps,
        * streams can be set to use device token instead.
        */
        let identifierType = samplingConfig["identifier"] as? String ?? "session"
        
        guard identifierType == "session" || identifierType == "device" else {
            cache[stream] = false
            return false
        }
        
        let identifier = identifierType == "session" ? sessionID : storageManager.deviceID
        let result = determine(identifier, rate)
        cache[stream] = result
        return result
```

WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
```diff
@@ -0,0 +1,605 @@
...
+        if data.keys.contains("meta") {
+            meta = data["meta"]! as! [String: String]
+        } else {
+            meta = [String: String]()
+        }
+        if domain != nil {
+            meta["domain"] = domain!
+        }
+        var data = data
+        data["meta"] = meta
```
tonisevener, line 513: Suggestion to help avoid the `!`

```
        guard enabled else {
            return
        }
        
        var meta: [String: String] = data["meta"] as? [String: String] ?? [:]

        if let domain = domain {
            meta["domain"] = domain
        }
        
        var data = data
        data["meta"] = meta
```

WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
```diff
@@ -0,0 +1,605 @@
...
+            // CC to other streams, even if this stream does not exist
+            if COPIED.keys.contains(stream) {
+                for cc_stream in COPIED[stream]! {
+                    log(stream: cc_stream, schema: schema, data: data, domain: domain)
+                }
+            }
+            if !(CONFIG!.keys.contains(stream)) {
+                return
+            }
+        }
```
tonisevener, line 549: This snippet is reworked with guard, it's generally preferable to keep things from nesting too much. Also safely unwrapping the copiedStream to avoid the `!`:

```
        guard let config = config else {
            let logItem = LogItem(stream: stream, schema: schema, data: data, domain: domain)
            inputBuffer.append(logItem)
            return
        }

        // CC to other streams, even if this stream does not exist
        if let copiedStream = copied[stream] {
            for ccStream in copiedStream {
                log(stream: ccStream, schema: schema, data: data, domain: domain)
            }
        }
        
        if !(config.keys.contains(stream)) {
            return
        }
```

WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
```diff
@@ -0,0 +1,605 @@
...
+            }
+        })
+    }
+
+    /**
+     * This method is called by the application delegate in `applicationWillResignActive()` and
+     * disables event logging.
+     */
+    public func app_in_background() {
+        ENABLED = false
```
tonisevener, line 235: Do you think we should set `LAST_TS` here as well? It would help match the `session-timed-out()` business logic more I think.

WMF Framework/Event Platform Client/EventPlatformClientProtocols.swift
```diff
@@ -0,0 +1,101 @@
...
+ */
+
+import Foundation
+
+/**
+ * Storage manager provides EPC with the ability to persist data and retrieve persisted data, asynchronously.
+ *
+ * The storage manager needs to be able to handle persistent storage of any type of value. When storing a
+ * session identifier, for example, it needs to store a `String`. When storing queued-up, to-be-sent events
+ * when entering background, it needs to store `[(String, String)]`. When storing raw logged events,
```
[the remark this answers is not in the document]
bearloga: Yep, although the only use right now is saving/loading the `input_buffer` (to be renamed to `inputBuffer`) of events which were logged before the stream configuration is available and need to be processed once the stream config is downloaded.

So the actual concrete scenario is an array of events as tuples: `[(stream: String, schema: String, data: [String: Any], domain: String?)]`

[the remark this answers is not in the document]
bearloga: Do you have recommendations for `data`'s type instead of a Dictionary of `Any`s? Because some event data fields might be strings, some might be numeric, some boolean, some dictionaries. JSON Schema (used by Event Platform for schema specification) supports objects in fields, so

```JSON
{
  "user": {
    "name": "Bearloga",
    "id": 150561
  },
  "session_length_s": 3600,
  "first_app_launch": false,
  "meta": {
     "stream": "test.sessions"
   },
  "$schema": "/analytics/test/session/1.0.0",
  "client_dt": "2020-07-21T09:54:01Z"
}
```

is an example of a POST body that the event ingestion service can handle (provided an appropriate schema)

So we want to be able to persist an event logged with
```swift
EPC.shared.log(
  stream: "test.sessions",
  schema: "/analytics/test/session/1.0.0",
  data: [
     "user": ["name": "Bearloga", "id": 150561],
     "session_length_s": 3600,
     "first_app_launch": false
  ]
)
```

WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
```diff
@@ -0,0 +1,605 @@
...
+ *     PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+ *     LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ *     NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ *     SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import Foundation
+
+/**
+ * **Note**: have not been able to get this into Dictionary+JSON.swift without getting "inaccessible due to
```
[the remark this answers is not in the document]
bearloga: Thank you!

WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
```diff
@@ -0,0 +1,605 @@
...
+     * - Parameter value: value to append to the string array or use as the first value of a new one
+     */
+    mutating func append_if_new(key: String, value: String) {
+        if self.keys.contains(key) {
+            if !self[key]!.contains(value) {
+                self[key]!.append(value)
+            }
+        } else {
+            self[key] = [value]
+        }
```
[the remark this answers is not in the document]
bearloga: Thanks! Will do and wow yeah that looks way better

WMF Framework/Event Platform Client/EventPlatformClientProtocols.swift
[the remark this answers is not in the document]
bearloga: Sorry, JUST remembered your notes from later in the review

WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
bearloga: Makes sense!

WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
bearloga: Oh that's super cool, thanks! I really like that and it reminds me of the `var meta = eventData.meta || {}` syntax from the JavaScript client

WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
```diff
@@ -0,0 +1,605 @@
...
+        storage_manager?.get_persisted("epc_input_buffer", completion: {
+            value in
+            if value != nil {
+                let ib = value as! [(stream: String, schema: String, domain: String, data: [String: Any])]
+                for e in ib {
+                    input_buffer.append(e)
+                }
+                storage_manager!.del_persisted("epc_input_buffer")
+            }
+        })
```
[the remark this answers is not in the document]
bearloga: I was thinking about making events a struct instead of a tuple, so I'm really glad you suggested this.

WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
```diff
@@ -0,0 +1,605 @@
...
+         */
+        if CONFIG == nil {
+            network_manager!.http_download(url: CONFIG_URI, completion: {
+                data in
+                let from_json = try? JSONSerialization.jsonObject(with: data!, options: []) as? [String: [String: Any]]
+                if from_json != nil {
+                    self.set_stream_config(from_json!)
+                } else {
+                    print("Problem processing stream config from response")
+                }
```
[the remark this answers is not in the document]
bearloga: Not for my future self: don't forget about Swift's [argument labels](https://docs.swift.org/swift-book/LanguageGuide/Functions.html#ID167) feature

WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
bearloga: I love this a lot because it's almost exactly like the JS client (https://gerrit.wikimedia.org/r/602739) and cross-platform similarity like that is cool and good.

P.S. Nuria was the one who pushed me away from nested conditionals in that patch.

WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
bearloga: I can't thank you enough for all this great educational feedback.

WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
```diff
@@ -0,0 +1,605 @@
...
+        }
+        print("[EPC] Stream cc-ing:\n\(COPIED.prettyPrintJSON)")
+        if CONFIG != nil && input_buffer.count > 0 {
+            var cached_event: (stream: String, schema: String, data: [String: Any], domain: String?)? = input_buffer.remove(at: 0)
+            while cached_event != nil {
+                log(stream: cached_event!.stream,
+                    schema: cached_event!.schema,
+                    data: cached_event!.data, domain: cached_event!.domain)
+                cached_event = input_buffer.remove(at: 0) // next
+            }
```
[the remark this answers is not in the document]
bearloga: I think I got it:

```swift
        if inputBuffer.count > 0 {
            var cachedEvent: Event
            while !inputBuffer.isEmpty {
                cachedEvent = inputBuffer.remove(at: 0)
                log(stream: cachedEvent.stream,
                    schema: cachedEvent.schema,
                    data: cachedEvent.data,
                    domain: cachedEvent.domain)
            }
        }
```

WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
bearloga: Good call

WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
bearloga: Thank you, this is much better

WMF Framework/Event Platform Client/EventPlatformClientProtocols.swift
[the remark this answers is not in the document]
bearloga: I've renamed everything to use camel case instead of snake case and caps. I'm sorry if it was really annoying to see those naming conventions throughout the PR. When I upload the fixes it should be less annoying to review.

bearloga:
TIL that PRs on GitHub aren't keen on `git commit --amend`

WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
```diff
@@ -0,0 +1,626 @@
...
+     * [mw:Wikimedia Product/Analytics Infrastructure/Stream configuration](https://www.mediawiki.org/wiki/Wikimedia_Product/Analytics_Infrastructure/Stream_configuration)
+     * for more information.
+     */
+    private func inSample(stream: String) -> Bool {
+
+        guard let configs = streamConfigurations else {
+            DDLogDebug("[EPC] Invalid state, must have streamConfigurations to check for inSample")
+            return false
+        }
+        if let cachedValue = samplingCache[stream] {
```
tonisevener, line 407: Actually this piece will need some thread-safety too. For this simple serial queue approach you can try `queue.sync`, that pauses the current thread execution while `queue` goes off and pulls the value and returns it back to the original thread that needs it. A pattern I've noticed in this codebase is to create a bunch of helper methods for getting and setting to these data structures. That buries all the queue.async and queue.sync calls and cleans up the code. It was a bit long to give an example here so I tried it in this [commit](https://github.com/wikimedia/wikipedia-ios/commit/36fc152c3edbd2fff651e8a9f9c556e5a7981ea0).

Also as I'm looking at this I think we might need thread safety when accessing or setting the `streamConfigurations` and `streamCopy` properties as well 😕   so I'll leave that for you. The reason being they are referenced in `log(...)`, which I envision will get called from multiple threads as we pepper them throughout our codebase, so we want to make sure these properties stay intact.

bearloga: Thanks! Hopefully I did it right in the upcoming commit

WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
bearloga: Thanks for all your guidance and help getting this stuff multithread-safe!

bearloga:
@tonisevener Would it be okay if I closed this PR and submitted a new one?

WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
```diff
@@ -352,6 +352,12 @@ public class EPC {
         }
         #endif
 
+        // Make them available to any newly logged events before flushing buffer
+        queue.sync {
+            self.streamConfigurations = config
+            self.streamCopy = copies
```
tonisevener, line 358: You'll be losing any previous values that were in `streamCopy` with this vs. how the old code worked. Seems like that's what you're wanting anyways, but just a heads up in case you expect `setStreamConfig` to be called more than once.

WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
```diff
@@ -329,17 +328,18 @@ public class EPC {
...
                     let child = s[0...i].joined(separator: ".")
                     let parent = s[0..<i].joined(separator: ".")
-                    streamCopy.appendIfNew(key: parent, value: child)
+                    copies.appendIfNew(key: parent, value: child)
                 }
             } else if nPrefixes == 1 {
-                streamCopy.appendIfNew(key: String(s[0]), value: stream)
+                copies.appendIfNew(key: String(s[0]), value: stream)
             }
         }
```
tonisevener, line 345: You probably want `let jsonString = try copies.toPrettyPrintJSONString()` on line 348 below now.

WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
```diff
@@ -362,6 +368,12 @@ public class EPC {
...
     }
 
+    func getStreamConfig() -> [String: [String: Any]]? {
```
tonisevener, line 371: Can you also reference this method instead of using `streamConfigurations` directly in the `configure()` method and `inSample(stream:)` methods?

WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
bearloga: That's a good point! As it is we're good since the stream configs are downloaded & processed only once, but once we figure out how to approach multiple stream configs (if at all) in the BUOD <> Apps sync we'll know if this needs to be modified to do, like, a merge operation.

bearloga:
Thanks for the green light! Going to re-submit as a fresh PR to deal with branch naming and merge conflicts.

PR #3604, opened 2020-06-11
Removed summary cache workaround

tonisevener:
**Fixes Phabricator ticket:** https://phabricator.wikimedia.org/T252845

Unfortunately I wasn't able to test this in the Beta Cluster, it continues to return the old content after a successful POST.

### Test Steps
1. Save article with a Wikidata title description
2. Go to article from Saved tab, note the saved tab cell subtitle.
3. Go to article, change the title description, wait for article to refresh.
4. Confirm you see your new title description in the article content
5. Go back to Saved, confirm you see the new title description in the cell subtitle.


tonisevener:
Hold for digging further into the Wikidata beta cluster test environment.

tonisevener:
> Tests aren't working for me on prod, testing on English Wikipedia. Am I doing it wrong?

I pushed something that hopefully fixes it. I could never reproduce the timeout you were seeing but at the very least a pull to refresh after the timeout should properly update the summary on the Saved tab. I dug in a little bit and the `api/rest_v1/page/summary` endpoint does not return `must-revalidate` in it's `cache-control` response header, and that I think was causing it to act weird. So I'm working around that with a different policy. Now if it's in the reloading state (which is triggered after an edit as well as a pull to refresh), I'm forcing it to revalidate, which means it will lean on local cache only after checking with the server against the cached etag and the server returns a 304. Otherwise it'll fetch the fresh response. Hopefully now you should be reliably seeing the title description subtitle updated in the Saved tab. 🤞

PR #3607, opened 2020-06-15
Use `Configuration.Domain.localhost` for `SchemeHandler`'s local PCS configuration

staykids:
**Phabricator ticket**: N/A

### Notes
Currently, `SchemeHandler` uses a hardcoded "localhost" to determine whether the resource scheme should be `http` or `https` on local PCS builds. This PR replaces that with the existing `Configuration.Domain.localhost` string.

Other than just replacing a magic string, this also has the added benefit of centralizing where a change needs to be made for development deployments of PCS. For example, if one switches from deploying PCS locally to anywhere else, they'd now only have to make a change to the `Configuration.Domain.localhost` for everything to work.

(This PR is based off of 6.6.1, but happy to change the merge base to master as it's not specifically critical for 6.6.1.)

### Test Steps (optional)
These are high commitment test steps for a low impact PR, so I really do not recommend or expect that anyone actually attempt anything other than maybe steps 1 and 2. Just documenting them:
1. Deploy `mobileapps` locally
2. Using the local PCS target, search for any article and confirm it loads as expected
3. Deploy `mobileapps` somewhere that isn't your local development machine
4. Update `Configuration.Domain.localhost`'s value to that new hostname
5. Using the local PCS target, search for any article and confirm it loads as expected


mcleinman:
@staykids, are you comfortable if I merge this into `master` rather than 6.6.1? It looks like it was reset to `master`, but your comment above implies 6.6.1. (FWIW, I'm not sure we need it in 6.6.1, even though it is very low impact.)

tonisevener:
@mcleinman sorry, yeah I moved to to master for that reason without communicating that 😬 Even though it's low impact I think it should go in master. I tested and it looks good to me!

PR #3609, opened 2020-06-16
Pass edit preview parameter in non-local environments

mcleinman:

### Notes
Well, this is why one should keep local and production code identical. 
- To allow testing of edit preview while scheme is set to local PCS, we had to fork the logic.
- We then went back and added code to send a parameter for the edit preview type...
- ...but only did that for the code for local PCS, not the production code.
- (This is yet to rear it's ugly head as a live bug because the server-side code is yet to deploy. It will be held for this, so no time crunch on getting this in.)

While working on this, I tried to write code to simplify testing this locally - some extra code for if you're running local restbase as well as local PCS. This ended up with more forking code running within the forking code, but this new code having to be setup in the `ArticleFetcher`. And assuming that your local restbase port is always 7231. Ultimately, given how edge case it was, adding more code complexity to help testing a corner case seemed like overengineering, (especially because the existing code complexity around this is precisely what led to this bug). I didn't include that code in the PR.

### Test Steps
Therein lies the rub. This is hard to test. You can setup parsoid and restbase locally, add some custom code on restbase (otherwise the second call in the wikitext-to-mobilehtml cocktail won't work), and test it... But that's a lot of work for not actually testing the proper route.

My suggestion, feel free to push back:
1. Test edit preview* on local PCS scheme.
2. Test edit preview* on production scheme.
3. After [this restbase patch](https://gerrit.wikimedia.org/r/c/mediawiki/services/restbase/deploy/+/605650) gets deployed, I'll test both those again.

*Edit an article, see the preview of your edit, ensure edit pencil icons are not shown.


joewalsh:
Should the mode default stay `editPreview` server-side given it's the reason the endpoint exists and it's the only mode in use? The naming of `/transform/wikitext/to/mobile-html` might lead people to believe that it's 'full' mobile-html, but we could update the documentation on RESTBase to be clearer that it's for previewing edits and that's the default mode. If the default mode does change, the Android app would need to be updated as well, it's currently not passing the `output-mode` header either

tonisevener:
It took me a while to get re-acquainted with the history of edit preview, but I think(hope) I’ve got a handle on it now (thanks for explaining in sync @MCleinman ). Although it feels weird to have the default output mode be an option that is not in use anywhere, I thought I remembered some discussion around what the spirit of this endpoint is, and do we heavily suggest that it's only for edit previews. I agree that the naming of the endpoint alone just feels like it would default to `contentAndReferences`, and `editPreview` is more of a optional variant you could configure it with. 

That being said, and correct me if I'm wrong but I think as soon as that default switches server-side, our current 6.0 release will break because this PR isn't in it. If that's true and Android would also break, I think we should hold on the server-side default change (and maybe make a small RestBase change for now to correct the documentation). If we agree the default should be `contentAndReferences`, let's deploy the PCS default change after Android and iOS have had a release that sends .`editPreview` as the header out in the wild for a while.

mcleinman:
@joewalsh I don't recall where - it may have been in a meeting - but we had decided previously to set the default for `html-to-mobilehtml` to `contentAndReferences`.  The idea behind this was:
- Currently this `html-to-mobilehtml` is only ever used for `editPreview`, I believe.
- However, `content` and `references` and `contentAndReferences` were all built into it.
- And the best guess is that someone naive to our codebase would assume that a conversion would come out as basic as possible, that is, `contentAndReferences` and not `editPreview`.

Thus, we decided to do `contentAndReferences`. That said, happy to revisit this. There is a "based on how things work now" argument to keep it as `editPreview`.

mcleinman:
And writing things up at the same time as @tonisevener. 100% agreed to everything she said.

mcleinman:
PCS now keeping the default of `editPreview` - PCS and RestBase both have patches up to update their documentation. 

I believe this PR could still be merged as is, though - unless we want to change `wikitextToMobileHTMLPreviewRequest`'s default argument for `outputMode` to match the server. I'm happy either way with that default argument.

staykids:
Catching up on this now. Unless I'm missing something, it looks like [this](https://github.com/wikimedia/mobileapps/commit/27090db65e2e44989d81b8ff9612efbee6437e45#diff-d9d5db0d62e0b69d9e9aaee33d7ad5a6 ) and [this](https://github.com/wikimedia/mobileapps/commit/b830c69c63f550ba3303883c0aaee300b5f50ac5) were the server-side patches surrounding this PR, right? Just wanted to confirm my understanding of if all the work surrounding this was deployed before I jump into your test steps.

mcleinman:
Main Restbase PR: https://github.com/wikimedia/restbase/pull/1261
Small Restbase documentation PR: https://github.com/wikimedia/restbase/pull/1265
PCS PRs are [here,](https://gerrit.wikimedia.org/r/c/mediawiki/services/mobileapps/+/594294) [here,](https://gerrit.wikimedia.org/r/c/mediawiki/services/mobileapps/+/596284) and [here](https://gerrit.wikimedia.org/r/c/mediawiki/services/mobileapps/+/606529).

PR #3611, opened 2020-06-18
Drop iOS 11

joewalsh:
**Fixes Phabricator ticket:** 
https://phabricator.wikimedia.org/T252163

### Notes
* The only new warnings were around `NSKeyedArchiver/Unarchiver` deprecations, so I fixed those.


Wikipedia.xcodeproj/project.pbxproj
```diff
@@ -12664,7 +12664,7 @@
 				GCC_WARN_UNINITIALIZED_AUTOS = YES;
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
-				IPHONEOS_DEPLOYMENT_TARGET = 11.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 12.4;
```
mcleinman, line 12667: Out of curiosity, why did we go with 12.4 rather than 12.0?

joewalsh: I thought I remembered that targeting the highest minor version was the recommended best practice from Apple but I can't seem to find any documentation for that.

joewalsh: I couldn't find any documentation about this and there's a significant number of sessions on 12.0-12.3, so I changed this to 12.0.

PR #3615, opened 2020-06-29
Fix editing title descriptions when selected from the context menu of selected text

mcleinman:
**Fixes Phabricator ticket:**  https://phabricator.wikimedia.org/T252042

### Notes
* I had to bring in a bit more data from the JS side to figure out how to present the info for title descriptions. This is my first time messing with the app-supplied JS code, so please let me know if I did things in a non-standard way.
* Notably, on the JS side I couldn't find a way to access the PCS-supplied `EditTransform` class from the app-supplied JS code. I've hardcoded the values, which makes it brittle (while showing the source in a comment). Is there a better way fro me to handle this?

### Test Steps
On an article which uses the title description menu, such as `USCGC Dallas (WHEC-716)`:
1. Long press text in the intro section. Select "Edit" from the context menu. Confirm that the edit article VC opens with the first section, and the app doesn't give the "edit intro or first section" menu.
1. Long press text in the second section. Select "Edit" from the context menu. Confirm that the edit article VC opens with the second section, and the app doesn't give the "edit intro or first section" menu.
1. 1. Long press text in the title description. Select "Edit" from the context menu. Confirm that the edit title description VC opens, and the app doesn't give the "edit intro or first section" menu.

On an article which does NOT use the title description menu, such as `Asian Giant Hrnet`:
1. Long press text in the intro section. Select "Edit" from the context menu. Confirm that the edit article VC opens with the first section, and the app doesn't give the "edit intro or first section" menu.
1. Long press text in the second section. Select "Edit" from the context menu. Confirm that the edit article VC opens with the second section, and the app doesn't give the "edit intro or first section" menu.
1. Long press text in the title description. Select "Edit" from the context menu. Confirm that the edit article VC opens with the first section showing and short description selected, and the app doesn't give the "edit intro or first section" menu.


mcleinman:
@joewalsh , thanks for updating the JS. 

I'm currently looking into improving the selection for article description, should have an update soon.

mcleinman:
> Looks good. I found one issue - if you highlight "battlecruiser" in the featured article today, SMS Derfflinger, it scrolls down to a later instance of "battlecruiser" in the editor rather than highlighting the one in the short description template. I tested a few other ones though and they acted fine, so I think this is an anomaly. I'll won't merge it in case you want to look further for a quick fix, but otherwise you're welcome to merge it as-is since the original bug is fixed.
> 
> out

@tonisevener I think I'm willing to let this one go. In this situation, highlighting "German" or "German battlecruiser" in the description works. I believe the problem lies within how the highlighted text is estimated in `editTextSelection.js`:
```
const wikitextRangeForSelectedAndAdjacentText = (selectedAndAdjacentText, wikitext) => {
  const wordsBeforeFromHTML = getNumberOfWordsFromEndOfString(selectedAndAdjacentText.textBeforeSelectedText, adjacentComparisonWordCount)
  const wordsAfterFromHTML = getNumberOfWordsFromBeginningOfString(selectedAndAdjacentText.textAfterSelectedText, adjacentComparisonWordCount)
  const selectedTextRegexPattern = `(${selectedAndAdjacentText.selectedText.replace(/\s+/g, '(?:(?:\\[\\[[^\\]\\|]+\\|)|{{[^}]*}}|<[^>]*>|\\W)+')})`
  const selectedTextRegex = new RegExp(selectedTextRegexPattern, 'gs')

  let bestScoredMatch = null
  while ((match = selectedTextRegex.exec(wikitext)) !== null) {
    const thisMatchScore = scoreForMatch(match, selectedTextRegex.lastIndex, wordsBeforeFromHTML, wordsAfterFromHTML)
    if (bestScoredMatch === null || thisMatchScore > bestScoredMatch.score) {
      bestScoredMatch = {
        match, 
        score: thisMatchScore,
        matchedWikitextBeforeSelection: {start: 0, end: match.index},
        matchedWikitextSelection: {start: match.index, end: selectedTextRegex.lastIndex}
      }    
    }
  }

  if (bestScoredMatch === null) {
    return null
  }

  const matchedWikitextBeforeSelection = bestScoredMatch.match.input.substring(bestScoredMatch.matchedWikitextBeforeSelection.start, bestScoredMatch.matchedWikitextBeforeSelection.end)
  const matchedWikitextSelection = bestScoredMatch.match.input.substring(bestScoredMatch.matchedWikitextSelection.start, bestScoredMatch.matchedWikitextSelection.end)

  return getWikitextRangeToSelect(matchedWikitextBeforeSelection, matchedWikitextSelection)
}
```

We could guarantee accuracy in short description case by adding some forking logic if `editInfo.isSelectedTextInTitleDescription` is `true`. That seems like a lot of work on the JS side for a pretty minor improvement - as this is only a potential problem when _sometimes_ selecting `Edit` _from the context menu_ when the article description comes from short description. Given this, it doesn't seem worthwhile to do the improvement to me. But I'm very, very open to disagreement on this - if anyone feels like it's worth the add'l code, I'm happy to write it.

PR #3616, opened 2020-07-01
Allow external links in EditPreview

mcleinman:
**Fixes Phabricator ticket:** https://phabricator.wikimedia.org/T253892

### Notes
* Don't know when/if the regression was introduced, but an empty string was the problem here - we were assuming a lack of string would be the bat signal and were thus thrown by an existing (yet empty) string.
* I did *not* fix iOS 13's context preview (aka peek & pop). We got some behavior for free on iOS 12, and on the Phab ticket I asked @carolynlimadeo about which behavior we want to create here. Depending on when feedback is provided, I'll either update this PR or create a new one for that work.

### Test Steps
1. Open any article, edit it, make a small change, get to edit preview page. (Ensure the section you're editing has at least one external link. Citations and infoboxes are good sources of external links.)
1. Try tapping a link to another wikipedia article. You should get a small preview with an "OK" button below it. Tapping that "OK" button will drop you back to the edit preview.
1. Try tapping an external link. Ensure you get the external link dialog:
Screenshot

### DIFF
--- a/Wikipedia/Code/EditPreviewViewController.swift
+++ b/Wikipedia/Code/EditPreviewViewController.swift
@@ -183,7 +183,7 @@ extension EditPreviewViewController: ArticleWebMessageHandling {
         case .unknown(let href):
             showExternalLinkInAlert(link: href)
         case .link(let href, _, let title):
-            if let title = title {
+            if let title = title, !title.isEmpty {
                 guard
                     let host = articleURL?.host,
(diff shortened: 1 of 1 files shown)


carolynlimadeo:
Looks good to me, thanks Matt!

mcleinman:
Replaced by this PR, which was based off 6.6.2: https://github.com/wikimedia/wikipedia-ios/pull/3625

PR #3621, opened 2020-07-08
Fix for non-functional "next" button in References

mcleinman:
**Fixes Phabricator ticket:** https://phabricator.wikimedia.org/T255283

### Notes
This is a hacky fix. I believe - though am not confident - that this bad behavior was [caused by Apple in 13.5](https://stackoverflow.com/questions/61787614/toolbar-button-unresponsive-after-ios-13-4-1-update) and is currently on track to be fixed in 14.0 (as of current beta). More details on the Phab ticket, and I'm open to other suggestions.

Tradeoff for the hack is a bit more padding on the trailing side of the toolbar, because of that "fake" button. 

We could:
* not fix it, because it only seems to be a 13.5 issue
* wrap the hacky fix in some code so it's only handled on iOS 13.5

### Test Steps
1. In "references" section, tap a backlink (`[1]`, etc.) for a reference that is cited multiple times in the article.
1. Tap the next button in the bottom right corner, ensure it works.
1. Tap the previous button in the bottom right corner, ensure it works.

### Screenshot of interface inspector
Screenshot 2

### DIFF
--- a/Wikipedia/Code/ReferenceBackLinksViewController.swift
+++ b/Wikipedia/Code/ReferenceBackLinksViewController.swift
@@ -34,8 +34,10 @@ class ReferenceBackLinksViewController: ReferenceViewController {
     lazy var countItem = UIBarButtonItem(customView: countContainer)
 
     func setupToolbar() {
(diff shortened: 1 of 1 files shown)


mcleinman:
Just going to do for iOS versions to 13.5 - 14.0. Will update code shortly.

mcleinman:
Using this one instead: https://github.com/wikimedia/wikipedia-ios/pull/3628

PR #3622, opened 2020-07-08
Fix potential stuck download condition

joewalsh:
**Fixes Phabricator ticket:** 
https://phabricator.wikimedia.org/T257461

### Notes
If there's any discrepancy between the underlying article `key` string and the roundtrip through `NSURL` to `wmf_databaseKey`, it's possible the app would get stuck retrying the same article because it would not appropriately mark it as downloaded.

### Test Steps
None yet, the issue was found via code inspection, need to verify how to repro in the app itself.


tonisevener:
@joewalsh Is there any reason we can't change `didRemoveArticle`, `didFetchArticle`, and `didFailToFetchArticle` to take the `WMFArticle` reference directly, and then those methods no longer need `operateOnArticles`? It probably makes no difference but seems like it would be even fewer hoops for it to jump through to be marked as downloaded.

joewalsh:
@tonisevener  It's possible that by the time we get the callbacks the Article has been deleted elsewhere. IIRC accessing that retaining and accessing that deleted `NSManagedObject` would cause a crash. As a result, the app should re-fetch whenever we leave and return the dispatch queue we originally fetched on. We could pass the `NSManagedObjectID` instead of the `key`, but it seems like we should work out consistency issues with the `key`. The other thing we should consider is updating the key of the article if we figure out the source of the inconsistency and how to best update it.

joewalsh:
Found [the Core Data concurrency documentation](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/CoreData/Concurrency.html). It mentions this as a way to handle issue across queues but if I'm recalling correctly it'a also an issue when leaving and returning to the same queue.

joewalsh:
This approach is quickly becoming unwieldy. Tomorrow morning I'll investigate an alternative solution that would fix the issue with a migration and leave the logic out of the `SavedArticlesFetcher`.

PR #3636, opened 2020-07-15
Fix blurry or pixelated photos caused by excessive zooming (updated)

Muximize:
Zooming a photo past a zoom scale of 0.5 causes a blurry or pixelated look, depending on the device used[1]. Changing the maximum zoom scale to 0.5 fixes this.

This happens most often when double tapping a photo. This is a widespread gesture to zoom a photo to the maximum zoom scale, but the Wikipedia iOS app seems to zoom in too far, which is not what a user would expect based on the behavior of other apps.

The current value of 2.0 was introduced in 55f3390. There is no information about why this value was chosen.

[1]: For example, blurry: iPhone 11 Pro, pixelated: iPhone 8 and iPad Pro 11"


staykids:
Hi @Muximize – thanks again for putting the time into exploring this. I think the points I made above are what's holding us back from merging this currently. For now, I'm going to close this PR. But please do leave a comment if you find time to do further thinking on how you might alter your solution to support the iOS Photos app style of interaction in order to support the use cases I mentioned above.

PR #3637, opened 2020-07-15
Peek and pop for lead image

mcleinman:
**Fixes Phabricator ticket:** https://phabricator.wikimedia.org/T253270

### Notes
* This only works on iOS 13 and higher. I think this broken behavior on iOS 12 is related to [this ticket](https://phabricator.wikimedia.org/T253330), so I focused on iOS 13 in this fix.
* This fix works on iPad - the lead image pops into the gallery controller. Notably, iPad on iOS 13+ has a bug for other (non-lead) images in articles - when coming from the webview, images peek but don't pop. I filed a bug for this: https://phabricator.wikimedia.org/T258109
* If no image has been opened on the article, the mediaList for the article is empty - and we need to fake the media list. (And to do that, we needed to write explicit initializers for all the codable structs.) This also means that there is a slight flash back to the article between peeking and popping. It doesn't seem worthwhile to dig into this further (given that this is supposed to be a quick fix), but am happy to dig in further (or hear ideas on how to improve it) if anyone feels like it is warranted.

### Test Steps
Test on both iPhone and iPad
1. Open an article that is not cached and has a lead image.
1. Long press the lead image. It should open in a peek.
1. Tap the image again. It should pop into gallery view.
1. Long press the lead image again. It should again open in a peek. Tap outside the peekable VC. The peek should close, and you should again be on an article.
1. Tap on the lead image, so you're on the gallery view. (This ensures the media list has been loaded.)
1. Long press the lead image. It should open in a peek.
1. Tap the image again. It should pop into gallery view.

Screenshot 10

Screenshot 10Screenshot 10


Wikipedia/Code/ArticleViewController+LinkPreviewing.swift
```diff
@@ -146,3 +146,43 @@ extension ArticleViewController {
...
+        return UIContextMenuConfiguration(identifier: nil, previewProvider: previewProvider, actionProvider: nil)
+    }
+
+    func contextMenuInteraction(_ interaction: UIContextMenuInteraction, willPerformPreviewActionForMenuWith configuration: UIContextMenuConfiguration, animator: UIContextMenuInteractionCommitAnimating) {
+        animator.addCompletion {
+            self.showLeadImage()
+        }
+    }
+
+    private func oneImageMediaList(from url: URL?) -> MediaList? {
```
joewalsh, line 172: Would it make sense to have this as an initializer on `MediaList`?

Wikipedia/Code/ArticleViewController+LinkPreviewing.swift
```diff
@@ -146,3 +146,43 @@ extension ArticleViewController {
...
+// MARK: Peek/Pop for Lead Image of ArticleVC (iOS 13 and later - no equivalent functionality on iOS 12 and prior)
+@available(iOS 13.0, *)
+extension ArticleViewController: UIContextMenuInteractionDelegate {
+    func contextMenuInteraction(_ interaction: UIContextMenuInteraction, configurationForMenuAtLocation location: CGPoint) -> UIContextMenuConfiguration? {
+        // If gallery has not been opened on that article, self.mediaList is nil - and we need to fake the media list
+        guard let mediaList = self.mediaList ?? oneImageMediaList(from: leadImageView.wmf_imageURLToFetch) else {
+            return nil
+        }
+        let previewProvider: UIContextMenuContentPreviewProvider = {
+            let galleryVC = self.getGalleryViewController(for: nil, in: mediaList)
```
joewalsh, line 159: Should the other gallery previews also use this approach?

Wikipedia/Code/ArticleViewController+LinkPreviewing.swift
mcleinman: Good call, done. Tried to leave a clear comment on the new initializer regarding what it is used for.

Wikipedia/Code/ArticleViewController+LinkPreviewing.swift
mcleinman: By "this", do you mean sending the item as `nil`? If so, just updated this to explicitly use the first item in the medialist. It defaults to the first item when initializing `MediaListGalleryViewController`, so this should be functionally identical, yet (helpfully) now more explicit.

joewalsh: My bad - I put this comment on the wrong line and wasn't specific enough. I meant the approach of using a single item media list rather than asynchronously requesting the actual media list. If there's no downside to this approach, it might make sense to consolidate the gallery previewing logic through one path.

mcleinman: I spun my wheels with that approach for quite a while, got nowhere, and ended up going with this approach. If you (or anyone else) has a suggestion to get the actual media list in this case, I'd love to implement it - I agree it's a much more solid approach. 

What I ran into: `previewProvider` is a `UIContextMenuContentPreviewProvider`, which is a typealias for a function that provides a previewing VC (`() -> UIViewController?`). I don't believe it wants to (or is able to) wait for any sort of async action to get that VC back. On the other hand, our `getMediaList` function asynchronously makes the network request for the media list, and calls the completion argument once the network request returns. I don't believe there is any way to make these two play nice with each other. We could maybe create a new blocking synchronous function to make the network request for the media list, but that seems a bit heavy handed. It's possible there are also other options I'm neglecting to think of.

I think I mentioned this hack in some meeting before settling on it, but you may not have been there. To be clear about the tradeoffs with this approach: When entering gallery view via peek and pop (not via a normal click) AND that is the first image you've tapped on (and thus there is no medialist loaded), you don't get the ability to horizontally scroll between article images. Not the best, but given how corner case it is, it seemed acceptable to everyone present.

joewalsh: Makes sense - the other option that occurs to me would be immediately returning `UIViewController` subclass that acts a container. It could show a spinner until it receives another `UIViewController` that it adds as a `child` covering its whole view. I don't know how finicky the presentation of the `UIContextMenuContentPreview` is, so it might not work well. Given this, and that there was already a discussion of the tradeoffs, I'm good calling this resolved.

mcleinman: Given how rare of a bug this is, I was initially trying to have as light of a fix as possible (both for code complexity purposes, as well as ensuring a small bug didn't eat up a disproportionate amount of time) - not creating new VCs or significant amounts of new logic. That said, let's make it right for the user - I'm happy to spin this up as a new VC. Will keep working on this.

mcleinman: Just added this in. I also had to fake a loading screen (for very slow networks). Using the macOS Network Link Conditioner, this seems to fit the bill - always loads all items for the gallery, and if the network request w/ the medialist doesn't come back quick enough we throw up a quick (user-dismissable) loading screen that looks identical to what is used when loading the actual image. In the background, we're fetching the mediaList and then fetching the first image - but loading both segments now look identical to the user (as if it was one load), as they should.

PR #3638, opened 2020-07-17
Improve voiceover for reference sections

mcleinman:
**Fixes Phabricator ticket:** https://phabricator.wikimedia.org/T250107

We added a new accessibility Voiceover label that is read when the user is over the "references section button" in the top leading corner when viewing a reference. I currently am using "jump to reference section", which has been ok'd by Carolyn.

### Notes
* Voiceover didn't work well on references. When tapping a backlink, when simply tapping a reference in an article (both as a popover or as a reference VC), etc. This improves it dramatically.

### Test Steps
**iPhone, probably needs to be done on device**
1. Open an article. Navigate to the reference section. Turn on VoiceOver. (Pro tip: You can use Siri to turn voiceover on and off.)
1. Tap a reference backlink. Voiceover should be on the point in the article where the reference is.
1. Tap over the buttons and labels on the top and bottom of the screen, ensure they make sense.
1. Close the backlink VC with the button in the trailing top corner. Tap a reference in the article.
1. Ensure the reference VC's buttons and labels makes sense via voice over, and that they're navigatable by swiping backwards and forwards (as when one uses VoiceOver to scan through a view).
1. Tap the button in the leading top corner to jump down to the reference section. Voiceover cursor should be at the actual reference number.
1. Anything else you can think of that involves VoiceOver and references.

**iPad, probably needs to be done on device**
1. Open an article. Turn on VoiceOver. (Pro tip: You can use Siri to turn voiceover on and off.)
1. Tap a reference in the article.
1. Ensure VoiceOver starts reading the pop over, and you can swipe back and forth through the reference popover.
1. Anything else you can think of that involves VoiceOver and references.


Wikipedia/Code/ArticleViewController+References.swift
```diff
@@ -6,7 +6,38 @@ extension ArticleViewController: UIPageViewControllerDelegate {
...
+        if vc is ReferenceBackLinksViewController {
+            dismissReferenceBackLinksViewController()
+        } else {
+            dismissReferencesPopover()
+        }
+    }
+
+    func referenceBackLinksViewControllerUserDidNavigateTo(referenceBackLink: ReferenceBackLink, referenceBackLinksViewController: ReferenceBackLinksViewController) {
+        scroll(to: referenceBackLink.id, centered: true, highlighted: true, animated: true)
+        dispatchOnMainQueueAfterDelayInSeconds(1.0) {
```
joewalsh, line 21: Does this delay exist to wait for the scroll to finish? If so, could this block be sent as the completion to `scroll(to:`? Also, to avoid a retain cycle (or calling this after the VC is gone) should the block weakly capture self?

mcleinman: Updated, thank you for these. I looked at the function below, hadn't realized `scroll to` could take a completion. And `weak self`s are sometimes my achilles heal. Good catch, thank you!

mcleinman: Good call - I clearly replicated the wrong part of the subsequent function. And thanks for catching the missing `weak self` - that's often my achilles heal.

PR #3640, opened 2020-07-22
iOS 14 back button

mcleinman:
**Fixes Phabricator ticket:** https://phabricator.wikimedia.org/T258216

### Notes
* *What branch should this be merged into?* (It was branched off `xcode-12-beta`.) Are we going to make `main` xcode-12 ready? Should this PR be held until that is complete?
* I tested this on iOS 13, on three languages (including an RTL one), and on iPad. I don't believe there are any regressions, but I'd love some good poking at this to make sure there really are no regressions.
* Only tricky part left: For some reason, we're losing the "W" button titleView when jumping back from a long press. I spent quite a while, and couldn't figure out why. I could confirm it happens between `viewWillAppear` and `viewDidAppear`. I'm hackily now re-adding it on `viewDidAppear`, but it means there is a small flash when jumping back via the long press menu - it starts with no button, but then the button appears a split second after the user sees the article. Happy to keep working on the cause, if we want. (Among other things, I did look everywhere we set a `titleView` in the app, to see if something jumps out as being the cause.) 

### Test Steps
1. Load lots of articles. Ensure a long press on iOS 14 allows you to jump back anywhere. Repeat on different tabs (Explore, Search, etc.).
1. Try on iPad.
1. Try on iOS 12/13, to make sure there weren't any regressions. We'll see new back button titles (short article titles show as back button title, long article titles cause back button text to remain as `Back`).
1. Make sure it looks ok in another language.


mcleinman:
Leaving this on hold until the "which branch to merge it into" question is sorted out.

mcleinman:
Took this off hold as we discussed we're going to merge it into `xcode-12-beta`. Also, I removed the request for Josh/Carolyn signoff - with iOS 14 beta 3, we now can maintain the back button in the nav bar saying "back" rather than the article title - and I've implemented that functionality.

Wikipedia/Code/ArticleViewController.swift
```diff
@@ -324,6 +328,9 @@ class ArticleViewController: ViewController, HintPresenting {
     
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
+
+        /// When jumping back to an article via long pressing back button (on iOS 14 or above), W button disappears. Couldn't find cause. It disappears between `viewWillAppear` and `viewDidAppear`, as setting this on the `viewWillAppear`doesn't fix the problem. If we can find source of this bad behavior, we can remove this next line.
+        setupWButton()
```
tonisevener, line 333: Yeah, I think this is related to the standard animation that occurs, I was able to repro with basic back button tap by returning true in `shouldPop` and commenting out the `delegate?.navigationController?.popViewController(animated: true)` line. I think this fix will work though, and it's good that it only happens on the context menu pop.

Wikipedia/Code/ExploreViewController.swift
```diff
@@ -57,6 +57,9 @@ class ExploreViewController: ColumnarCollectionViewController, ExploreCardViewCo
         super.viewWillAppear(animated)
         collectionViewUpdater?.isGranularUpdatingEnabled = true
         restoreScrollPositionIfNeeded()
+
+        /// Terrible hack to make back button text appropriate for iOS 14 - need to set the title on `WMFAppViewController`. For all app tabs, this is set in `viewDidAppear`.
+        parent?.navigationItem.backButtonTitle = title
```
tonisevener, line 62: Yeah, I think this is the lowest-risk fix for this problem. It might also work if you embedded the 5 root view controllers in their own navigation controller, but I wouldn't be comfortable with a fix that big.

WMF Framework/NavigationBar.swift
```diff
@@ -723,4 +706,14 @@ extension NavigationBar: UINavigationBarDelegate {
         delegate?.navigationController?.popViewController(animated: true)
         return false
     }
+
+    public func navigationBar(_ navigationBar: UINavigationBar, didPop item: UINavigationItem) {
```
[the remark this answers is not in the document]
mcleinman: `shouldPop` still isn't called for a long-press pop in the latest beta which came out today (beta 5).

I think the weird animation is due to having to call this in `didPop` rather than `willPop`. (I've added some additional comments around this.) From slowing down the animations, it looks like the system pop is the weird artifact one, and then our manual pop is the second (real) one. I can't think of any way around this w/o being able to interrupt it in `willPop` - I'm very open to ideas though. I left a comment encouraging us to move it to `willPop` when that (presumed) bug gets fixed.

(Notably, `willPush` and `didPush` aren't called when pushing a new article on the stack. This could be because of how we push new articles. Potentially digging further into how we push new articles to the NavController could shed some light on other alternative ways to handle the popping. But that also feels like a potential rabbit hole - If this seems worthwhile to dig into in order to potentially improve this animation, let me know.)

Wikipedia/Code/ExploreViewController.swift
mcleinman: Sounds good - I agree with you on that.

mcleinman:
After playing around for a while, I'm able to reliably replicate getting a reading list to show a generic "Back" item in the list: 
1. Open saved tab, open it to reading lists.
1. Tap into any specific reading list.
1. Tap into any article in the reading list.
1. Tap back to go back to the specific reading list's main controller.
1. Tap into an article in the reading list. (Same one as before, or a different one - doesn't matter.)
1. Get the generic "Back" entry.

I can see roughly where this is happening in the code. Now to fix it.

mcleinman:
Solution for #1: Dug into the disappearing Reading List title. At it's core, it's due to [this line](https://github.com/wikimedia/wikipedia-ios/blob/1bc0cea71e2556b60e8b30701e162318f737ffbb/WMF%20Framework/NavigationBar.swift#L437) setting the `title` back to `nil`. I looked into this lower level code and it seemed unwise to mess with it. I added a permanent back button title by setting an explicit `navigationItem.backButtonTitle` and setting the `navigationItem.backButtonDisplayMode` to `generic`.

mcleinman:
Good news/bad news: This took longer than expected to patch up. But I now understand the navigation of many more corners of the app.

Some notes:
- I believe the only behavior that changes in iOS 13 is that the initial back button from any tab says the tab's name ("Explore"/"Saved"/etc.) rather than just "Back". I can update this to again say "Back" if we want, but it seemed potentially natural.
- I removed the back button boolean when deep linking to an article diff. I believe it's no longer needed - checked on both iOS 13 and 14, both for deep linking in and not. But if I missed something, let me know.
- To get a back button title for article talk and user talk pages, I needed to use something from the URL, as we don't have the page title. I went with the final part of the URL, so they look like `Talk:Robert_E._Howard`. It's not perfect, but I think it's decent. This "use the final part of the URL" is done for any `SinglePageWebViewController`.
- I've only found one spot where we don't get a title when long pressing, and instead get a generic entry of "Back" - a usertalk subpage [like this one](https://en.wikiredia.com/wiki/User_talk:GiantSnowman/2010). I think this is acceptable for now as I think it's s a bit of a corner case, but let me know if I should dig in further on this.
- I did not add long pressing the back button to the editing flow. Editing seems touchy enough that I'd want to make sure I full understand all the possible routes to make sure that jumping back two steps isn't a problem. I'm inclined to think that we don't need to include this feature for the editing flow - especially because it's not behaving badly and showing bad text, it's simply not providing any long-press behavior - but I don't feel strongly, and can happily investigate that if it seems worthwhile.

tonisevener:
@mcleinman This is working great, nice work! I made one small tweak - I combined the new `WMFLocalizedString`s with the old matching keys into the `CommonStrings.swift` file. I think it would have worked fine as-is but seems like a better practice to have them once.

PR #3642, opened 2020-07-22
Fix Core Data value transformer warnings

joewalsh:
**Fixes Phabricator ticket:** 
https://phabricator.wikimedia.org/T231987

### Notes
Creates a custom `WMFSecureUnarchiveFromDataTransformer` that allows `MTLModel` and the other objects we transform. Makes our `MTLModel` subclasses conform to `NSSecureCoding`.  Utilizes `WMFSecureUnarchiveFromDataTransformer` for all transformable Core Data properties.

I did some additional cleanup of the `WMFContentGroup` model to remove a deprecated class. I also fixed an error that surfaced with the `NSKeyedUnarchiver`s that changed when dropping iOS 11.

### Test Steps
1. Use the app normally with a prior version
2. Run the app with this branch
3. Ensure the feed data, user preferences, and event logging continue to work as normal


joewalsh:
> Code looks good. However, on a fresh install on 12.4 (using 8 Plus simulator) I'm getting consisent crashes on launch. Is it possible that crash stems from this PR?
Thanks for catching that! Should be fixed now.

> Also: I don't see One or more models in this application are using transformable properties with transformer names that are either unset, or set to NSKeyedUnarchiveFromDataTransformerName printed on a fresh install from this branch, but still see it if I run master, then run this branch. Is that expected?
Yes, I saw that as well and I believe it's expected. My guess is the error is printed one more time as the Core Data library is migrated from the old to new version.

WMF Framework/Wikipedia.xcdatamodeld/.xccurrentversion
```diff
@@ -3,6 +3,6 @@
 <plist version="1.0">
 <dict>
 	<key>_XCCurrentVersionName</key>
-	<string>Wikipedia 3.xcdatamodel</string>
+	<string>Wikipedia 4.xcdatamodel</string>
```
tonisevener, line 6: @joewalsh Just curious - why the new version for this data model but not the other ones?

joewalsh: Good question, and there's not a good reason. Core Data will generate the mapping automatically even if the version doesn't change, but it's probably better to be explicit about it whenever we change a model. I've added new versions for the other models and reverted the base versions in 2e33e9f.

tonisevener:
Looks good to me! I did see the initial warning that @mcleinman noticed after first running but they went away on the second run. And both times the data showed fine in the app. One thing I wanted to mention just for documentation's sake is that I see `CoreData: Failed to load optimized model at path`on iOS12 simulator, but it happens on both main and this PR so probably unrelated. I tested iOS12 simulator and iOS13 device and all data loaded fine & logging still works.

PR #3646, opened 2020-07-30
Updates for Xcode 12

joewalsh:
### Fixes
https://phabricator.wikimedia.org/T259649

### Notes
* ~~As of Xcode 12 beta 4, there's a blocking iOS 14 simulator-only bug where articles take an excessively long time to load. The app gives the`webView` a `URLRequest` and there's a long pause (10+ seconds) before the `webView` makes the request. The issue persists if the custom scheme handler is removed and the request is made with `https` as the scheme. iOS < 14 simulators and devices work.~~ This is fixed in Xcode 12 beta 5
* Fixed an iPad crash with 1f7d04c
* Cleared view debugger warnings and fixed unresponsive explore feed cells in c00a5b2...97fa10b

### Test Steps
1. Build and run using the latest version of Xcode 12
2. Verify there are zero warnings
3. Verify articles, non-article pages, and editors load and work correctly


staykids:
Just an FYI, I'm rebasing my WidgetKit work off this branch. It's not a big deal though if this branch get mutated wildly or even deleted – for the most part that work is fairly portable.

Wikipedia/Wikipedia-Info.plist
```diff
@@ -65,6 +67,7 @@
 	<key>UIBackgroundModes</key>
 	<array>
 		<string>fetch</string>
+		<string>processing</string>
```
tonisevener, line 70: @joewalsh Just curious as to your reasons for adding this - I see an error on the build server log with ""Missing Info.plist value. The Info.plist key 'BGTaskSchedulerPermittedIdentifiers' must contain a list of identifiers used to submit and handle tasks when 'UIBackgroundModes' has a value of 'processing'. For more information, refer to the Information Property List Key Reference at https://developer.apple.com/library/ios/documentation/General/Reference/InfoPlistKeyReference/Introduction/Introduction.html.""

joewalsh: There was an issue with `WKWebView` without it but that was a few betas ago. I'll remove it and see if the error persists. If it does, there's a `BackgroundTasks` branch that converts our background fetch to a background task and adds the appropriate identifier that we could incorporate.

Wikipedia/Code/WebPageUserScript.swift
```diff
@@ -1,7 +1,9 @@
 /// PageUserScript uses the platform-appropriate WKUserScript initializer to create a user script in WKContentWorld.page
 class PageUserScript: WKUserScript {
     override init(source: String, injectionTime: WKUserScriptInjectionTime, forMainFrameOnly: Bool) {
-        if #available(iOS 14.0, *) {
+        if #available(iOS 14.0, *), UI_USER_INTERFACE_IDIOM() == .phone {
+            // For some reason this crashes on iPadOS.
```
tonisevener, line 5: @joewalsh I'm seeing the opposite actually in Beta 5. With this line iPad reliably crashes when it enters the `else`. Can you confirm this is fixed now & we can remove the `.phone` check?

joewalsh: Confirmed, thanks for fixing this!

PR #3648, opened 2020-07-30
Event logging client for (Modern) Event Platform

bearloga:
**Fixes Phabricator ticket:** [T228180](https://phabricator.wikimedia.org/T228180)

## Notes

This pull request (a re-do of #3603) adds the iOS version of the [Event Platform Client](https://www.mediawiki.org/wiki/Wikimedia_Product/Analytics_Infrastructure/Event_Platform_Client) (EPC-iOS) for integration with the modern [Event Platform](https://wikitech.wikimedia.org/wiki/Event_Platform). EPC is an effort to unify algorithms, identifiers, and behaviors across MediaWiki, iOS, and Android.

This initial release supports:
- [remote sampling configurations](https://www.mediawiki.org/wiki/Wikimedia_Product/Analytics_Infrastructure/Stream_configuration#Sampling_settings), which puts sampling rates into the centralized stream configuration

For additional background information on instrumentation within the Modern Event Platform, please refer to [this page](https://wikitech.wikimedia.org/wiki/Event_Platform/Instrumentation_How_To).

## Next steps

- [x] While the library has already been heavily reviewed and iterated on in the previous PR, but it wouldn't hurt to review again – especially with respect to thread safety
- [x] Implement `StorageManager` and `NetworkManager` to make EPC operational
  - [x] Temporarily remove storage manager facilities (except app install ID & usage data sharing opt-in/out setting retrieval) for initial release until we settle on specification of persistence
- [x] Restore `EPC.shared` singleton (was temporarily changed until storage & network managers are implemented)
- [x] Resolve open question around per-wiki stream configurations (in BUOD <> Apps sync agenda)
  - Will not have per-wiki stream configurations for apps, any language-specific deviations in sampling settings may need to be handled via the[ proposed exemptions idea](https://www.mediawiki.org/wiki/Wikimedia_Product/Analytics_Infrastructure/Stream_configuration#Specifying_exemptions)
  - Apps will query 1 MW API (metawiki's) to download stream configs (side note: EventGate uses metawiki's stream configs as its source of truth)
- [x] Make EPC respect user's usage data sharing preference the way Event Logging does
- [x] Q&A functionality with a simple schema & instrument (test schema+stream for analytics events tracked in [T259714](https://phabricator.wikimedia.org/T259714))
  - [x] Create a schema in [this repository](https://gerrit.wikimedia.org/g/schemas/event/secondary): `/analytics/test/1.0.0`
  - [x] Configure a stream via [this file](https://gerrit.wikimedia.org/r/plugins/gitiles/operations/mediawiki-config/+/master/wmf-config/InitialiseSettings.php): `test.instrumentation` and `test.instrumentation.sampled`
  - [x] Test events verified in database/table (`event.test_instrumentation`)
- [x] Pilot instrumentation (Edit History Compare feature)
  - [x] Schema (`/analytics/mobile_apps/ios_edit_history_compare`)
  - [x] Instrumentation (EditHistoryCompareFunnel.swift)
  - [x] Stream (`ios.edit_history_compare`)
- [x] **BLOCKER** EventStreamConfig API returns empty configs as `[]` instead of `{}` ([T259917](https://phabricator.wikimedia.org/T259917))
  - [x] Merged
  - [x] Deployed (scheduled for 2020-09-08 MediaWiki deployment train)
  - [x] `configURI` changed from fake endpoint at pai-test to production endpoint at meta wiki

Future release may support:
- [stream cc-ing](https://www.mediawiki.org/wiki/Wikimedia_Product/Analytics_Infrastructure/Stream_configuration#Stream_cc-ing), which automatically logs events to multiple streams without requiring multiple calls on the instrumentation side


bearloga:
Just wanted to note that we (Product Infrastructure Data team) decided to put the stream cc-ing feature on the back-burner, as there are still some unresolved questions around how it should really work (implicit mapping as was implemented here or explicit mapping as suggested in [T256169#6254835](https://phabricator.wikimedia.org/T256169#6254835)), and we didn't want this feature to be out on iOS from the get-go and potentially never on MediaWiki. All deliberation will happen in [T256165](https://phabricator.wikimedia.org/T256165).

tonisevener:
@bearloga PR has been updated with `StorageManager` and `NetworkManager` implementations. I also added a couple of (identical) test log calls - one when you pull to refresh on the Explore tab and another when you tap the Settings icon on the Explore tab. I did a little bit of renaming and some more thread safety stuff in the library, hope that's cool. Let me know how this looks on your end, if it seems good we can bring it back to our side for converting a particular feature's logging & final code review.

WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
```diff
@@ -0,0 +1,765 @@
...
+        do {
+            let jsonString = try data.toPrettyPrintJSONString()
+            DDLogDebug("[EPC] Event logged to stream '\(stream)' (schema: \(schema):\n\(jsonString)")
+        } catch let error {
+            DDLogError("[EPC] \(error.localizedDescription)")
+        }
+        #endif
+
+        guard let streamConfigs = streamConfigurations else {
+            let event = EPCBufferEvent(stream: stream, schema: schema, data: data, domain: domain)
```
tonisevener, line 612: @bearloga do you think we should persist a timestamp within EPCBufferEvent? Otherwise when it eventually does get posted because streamConfigurations is populated, I think the `client_dt` field will represent the post date but not the date the event was triggered.

WMF Framework/Event Platform Client/EPCNetworkManager.swift
```diff
@@ -0,0 +1,140 @@
...
+        operationQueue = OperationQueue()
+        operationQueue.maxConcurrentOperationCount = 1
+    }
+    
+    func httpPost(url: URL, body: NSDictionary) {
+        storageManager.createAndSavePost(with: url, body: body)
+    }
+    
+    func httpDownload(url: URL, completion: @escaping (Data?) -> Void) {
+        httpDownload(url: url, attempt: 0, maxAttempts: 5, attemptDelay: 2, completion: completion)
```
tonisevener, line 23: @bearloga This is arbitrary logic I chose on config download retry attempts. Not sure who needs to OK this, maybe @jtminor ? This tries for 5 times, with 2 seconds in between attempts. Then it fails and the library will persist in the buffer until the next launch.

bearloga: That works! 5 attempts with 2 seconds between attempts sounds reasonable and generous :) We can run it by in the next BUOD<>Apps sync, but I doubt anyone would suggest a different configuration.

@linehan @dbrant: If we're assuming that the Android client will need to have same/similar behavior, do these settings for re-trying stream configuration retrieval make sense for y'all?

WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
bearloga: We should be good as is! When a buffered event gets re-`log()`-ed once stream configurations are available, that event's `data` will already have have a `client_dt` (and a `session_id`) populated from when the event was originally logged, before it was put into the input buffer.

WMF Framework/Event Platform Client/EPCNetworkManager.swift
linehan: This seems fine to me. We might re-architect this a bit down the road. I see this logic as being delegated to the implementation because it's part of fetching the information required to initialize the library (i.e. it takes place before the library 'comes into existence' in an operational sense). How the implementation gets to that starting line may be arbitrary. But, it's probably not a bad idea for them to match.

WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
linehan: Without looking too closely, I think @tonisevener may be alluding to the issue where the timestamp is not generated until the event is being processed (i.e., pulled out of the input buffer). The time at which the event actually *happens* in software will thus be offset by the time it spends in the buffer. IMO the best option here is to just leave it simple, as it's a identified source of timestamp bias that we're going to correct.

linehan: Oh, I see, yes here it's already been set in `data`. All good then.

tonisevener: @bearloga @linehan 👍  doh, I didn't catch that part. I see it now, thanks for explaining!

WMF Framework/Event Platform Client/EPCNetworkManager.swift
```diff
@@ -0,0 +1,140 @@
...
+            return
+        }
+        
+        let task = session.dataTask(with: url) { (data, response, error) in
+            
+            let failureBlock = {
+                dispatchOnMainQueueAfterDelayInSeconds(attemptDelay) {
+                    self.httpDownload(url: url, attempt: attempt + 1, maxAttempts: maxAttempts, attemptDelay: attemptDelay, completion: completion)
+                }
+            }
```
bearloga, line 39: @tonisevener This is clever!! I like this a lot! I think I improved it slightly in an upcoming commit by getting rid of `attempt`, switching to `maxAttempts: maxAttempts - 1`, and checking for `maxAttempts > 0`

bearloga:
**Note to self**: depending on the outcome of the discussion taking place in the identifiers fragment patch (https://gerrit.wikimedia.org/r/c/schemas/event/secondary/+/615559/3/jsonschema/fragment/analytics/identifiers/current.yaml#22), `session_id` and `device_id` may need to be renamed to `app_session_id` and `app_install_id`

bearloga:
@joewalsh Thanks for fixing the tests and making those updates!

**Status update**: We're getting closer to being able to create the pilot schema. There was a lot of discussion and iteration on https://gerrit.wikimedia.org/r/c/schemas/event/secondary/+/615559 and we've arrived at a version we're all satisfied with, so that'll be merged tomorrow (Friday) and I'll probably have pilot schema + pilot instrument up for review by EOD/EOW. Exciting!!!

WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
```diff
@@ -0,0 +1,776 @@
...
+            self.samplingCache.removeAll()
+        }
+    }
+}
+
+//MARK: PeriodicWorker
+
+extension EPC: PeriodicWorker {
+    public func doPeriodicWork(_ completion: @escaping () -> Void) {
+        networkManager.httpTryPost()
```
bearloga, line 764: @tonisevener: Since EPC is extended to have `httpTryPost` which then calls network manager's `httpTryPost`, should this be `self.httpTryPost()` instead? Or should we remove

``` swift
/**
 * Passthrough method to tell `networkManager` to attempt to post its
 * scheduled events
 */
@objc public func httpTryPost() {
  networkManager.httpTryPost()
}
```

above?

WMF Framework/Event Platform Client/EPCNetworkManager.swift
```diff
@@ -0,0 +1,135 @@
...
+        var queuedEvent: (url: URL, body: NSDictionary)?
+        while !outputBufferIsEmpty() {
+            queuedEvent = removeOutputBufferAtIndex(0)
+            if let queuedEvent = queuedEvent {
+                self.submit(url: queuedEvent.url, payload: queuedEvent.body)
+            }
+        }
+    }
+
+    private func submit(url: URL, payload: Any? = nil) {
```
linehan, line 89: Consider calling this `send`. See comment above about `schedulePost` being `schedule` and `httpTryPost` being `sendAllScheduled`

I would also change `payload` to `body`.

WMF Framework/Event Platform Client/EPCNetworkManager.swift
```diff
@@ -0,0 +1,135 @@
...
+                return
+            }
+            
+            completion(data)
+        }
+        task?.resume()
+    }
+    
+    func httpTryPost() {
+        var queuedEvent: (url: URL, body: NSDictionary)?
```
linehan, line 80: Consider renaming `queuedEvent` to `event`, or `item`, since we don't refer to the `outputBuffer` as a queue here.

WMF Framework/Event Platform Client/EPCNetworkManager.swift
```diff
@@ -0,0 +1,135 @@
...
+     * - Parameters:
+     *   - url: where to request data from
+     *   - retriesRemaining: maximum number of retries allowed for this download
+     *     operation
+     *   - attemptDelay: time between each retry
+     *   - completion: what to do with the downloaded data
+     *
+     * The operation will be performed up to `retriesRemaining + 1` times.
+     */
+    private func httpDownload(url: URL, retriesRemaining: Int, attemptDelay: TimeInterval, completion: @escaping (Data?) -> Void) {
```
linehan, line 41: Because this is only ever used to fetch stream configurations, and only ever receives one completion handler, I would suggest we name this `fetchStreamConfiguration` and investigate inlining it in the main library class.
```swift
private func fetchStreamConfiguration(retries: Int, retryDelay: Date) {
    networkManager.httpGet(url: streamConfigurationServiceURI, completion: { 
        (data, response, error) in 
            
        if let res = response, res.statusCode == 200 {
            if let dat = data {
                loadStreamConfiguration(dat)
            }
        } else {
            if retries > 0 {
                dispatchOnMainQueueAfterDelayInSeconds(retryDelay) {
                    fetchStreamConfiguration(retries: retries-1, retryDelay: retryDelay)
                }
            } else {
                    /* do the thing that you want to do when you can't load stream configuration */
            }
        }
    })
}
```
Here, `loadStreamConfiguration` would be the completion handler currently in `configure()`. `configure` then basically becomes the code above. The other thing you need is the method `httpGet`, which could look like:
```swift
private func httpGet(url: String, completion: @escaping (Data?, HTTPURLResponse?, Error?) -> Void) {
    let request = URLRequest( url: URL(string: url) )
    request.httpMethod = "GET"

    let task = session.dataTask(with: request, completionHandler: completion) 
    task?.resume()
}
```

WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
tonisevener: @bearloga Oh yeah, good catch. I can't remember why I thought we needed the passthrough method (maybe I didn't realize all calls would be able to flow through the `PeriodicWorker`), but it doesn't seem needed now. You can remove it.

WMF Framework/Event Platform Client/EPCNetworkManager.swift
```diff
@@ -0,0 +1,135 @@
...
+    
+    init(session: Session = Session.shared) {
+        self.session = session
+    }
+
+    func schedulePost(url: URL, body: NSDictionary) {
+        appendPostToOutputBuffer((url: url, body: body))
+    }
+    
+    func httpDownload(url: URL, completion: @escaping (Data?) -> Void) {
```
linehan, line 26: This isn't necessary; see comment below.

WMF Framework/Event Platform Client/EPCNetworkManager.swift
```diff
@@ -0,0 +1,135 @@
...
+    /**
+     * Holds events that have been scheduled for POSTing
+     */
+    private var outputBuffer: [(url: URL, body: NSDictionary)] = []
+    
+    init(session: Session = Session.shared) {
+        self.session = session
+    }
+
+    func schedulePost(url: URL, body: NSDictionary) {
```
linehan, line 22: Consider re-naming this to simply `schedule`.

WMF Framework/Event Platform Client/EPCNetworkManager.swift
```diff
@@ -0,0 +1,135 @@
...
+                failureBlock()
+                return
+            }
+            
+            completion(data)
+        }
+        task?.resume()
+    }
+    
+    func httpTryPost() {
```
linehan, line 79: I would re-name this `sendAllScheduled`.

WMF Framework/Event Platform Client/EPCNetworkManager.swift
```diff
@@ -0,0 +1,135 @@
+
```
linehan, line 1: If the below changes are able to be made to EPC/EPCNetworkManager, and the extraneous accessors are wrapped up in 3-4 functions for thread-safe arrays, then there will be a lot of room to put the remaining `schedule`/`send`/`sendAllScheduled` and the `fetchStreamConfiguration` in the main class. The truly implementation-dependent functions are `httpPost` and `httpGet` (see review below for what these are), but those are small. 

The reason I'm after this is because I think that's what this code "wants" -- by which I mean we are duplicating state without needing to (e.g. the DispatchQueue), encapsulating functions that are only ever called once (e.g. httpDownload), and instantiating without much reason (Session is already instantiated and Session.shared already exists). We are also not cleanly separating the truly implementation-dependent parts (e.g. the issuing of the HTTP requests) from the other parts of the biz logic. 

Unification would allow us to remove the protocol file as well, bringing us down to just the one file with a section for implementation-dependent threadsafe array access and implementation-dependent http get/post requests, and nothing else. This too makes sense to me since I don't think we plan to re-implement these components wth different logic but the same protocol/contract.

I'm happy to prepare a patchset if it would help

WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
```diff
@@ -0,0 +1,776 @@
...
+            }
+        }
+    }
+    private var _streamConfigurations: [String: [String: Any]]? = nil
+    
+    /**
+     * Updated with every `log` call and when app enters background, used for determining if the
+     * session has expired.
+     */
+    private var lastTimestamp: Date {
```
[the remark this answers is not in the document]
tonisevener: Good catch and agreed we should remove - if it's only called from the app delegate then I only see it running on the main thread which is serial. Could be that `lastTimestamp` was called from other areas in previous iterations or I just became a little too thread-safety happy :)

WMF Framework/Event Platform Client/EPCNetworkManager.swift
```diff
@@ -0,0 +1,135 @@
...
+        queue.sync {
+            return self.outputBuffer.isEmpty
+        }
+    }
+    /**
+     * Thread-safe synchronous removal of scheduled event at index
+     * - Parameter index: The index from which to remove the scheduled event
+     * - Returns: a previously scheduled event
+     */
+    func removeOutputBufferAtIndex(_ index: Int) -> (url: URL, body: NSDictionary)? {
```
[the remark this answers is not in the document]
tonisevener: You can also use `return self.outputBuffer.popFirst()` to clean this up. It returns an optional for in case the array is empty, so you can clean out the count checks.

WMF Framework/Event Platform Client/EPCNetworkManager.swift
tonisevener: This sounds good to me. My only ask is to lean on putting the network methods in a private extension on EPC, similar to how we put all of it's thread safety methods in a private extension.

WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
```diff
@@ -0,0 +1,776 @@
...
+     */
+    @objc public func submit(stream: String, data: [String: NSCoding], domain: String? = nil) -> Void {
+        guard self.sharingUsageData else {
+            return
+        }
+
+        /*
+         * EventGate needs to know which version of the schema to validate against
+         */
+        let schemaKey = "$schema"
```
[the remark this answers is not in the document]
tonisevener: I put it in `schemaKey` originally, at the time I think the same literal was used in multiple places so I wanted to avoid a situation where it accidentally is only changed in one place in the future. Agree it can be moved back to a literal now though.

WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
```diff
@@ -0,0 +1,776 @@
...
+        guard data.keys.contains(schemaKey) else {
+            DDLogError("EPC: Event data is missing the required '$schema' field")
+            return
+        }
+
+        let metaKey = "meta"
+        var meta: [String: NSCoding] = data[metaKey] as? [String: NSCoding] ?? [:]
+
+        if let domain = domain {
+            meta["domain"] = domain as NSCoding
```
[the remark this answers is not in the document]
tonisevener: Yeah, at the time I thought I was limited to `NSCoding` to persist items in Core Data, though looking again now I can't remember how I arrived at that. The swiftier way for serialization in general is to use `Codable`, and we may be able to use generics (shoutout to @joewalsh for inspiration) for the event's data, which could potentially clean a lot of this dictionary construction and the log call site up. But I don't have any more concrete suggestions beyond that without tinkering with it myself first and seeing what works. @bearloga feel free to run with that or wait on this part till I have time to revisit (probably tomorrow afternoon) and try to arrive at more specific feedback.

WMF Framework/Event Platform Client/EPCNetworkManager.swift
linehan: Yeah totally agree about putting them in a private extension. Whatever minimally opinionated structure we can grab that keeps us from accidentally coupling to the main class. Private extensions don't exactly prevent us from two-way coupling but I it works for me, all we need is a reminder for now that these are distinct. The interface will formalize over the next couple revs as things grow/evolve and it'll tell us what it wants to look like.

WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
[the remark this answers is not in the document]
bearloga: Hah! Our operating theory was that it was for performance, micro-optimization reasons. Alright, I'll change this and the other `{field}Key`s too.

WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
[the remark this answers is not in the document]
bearloga: I'm like 99% sure the `NSCoding` is for Core Data storage reasons, which we can probably change to `Any` for v1 since we got rid of the persistence facilities. @tonisevener do you concur?

WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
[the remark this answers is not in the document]
bearloga: Aye, it used to be updated in the logging call but we changed it to only be updated on entering background, so we're good to simplify this. Thanks for clarification!

WMF Framework/Event Platform Client/EPCNetworkManager.swift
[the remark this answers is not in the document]
linehan: Yeah that sounds goood.

WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
```diff
@@ -0,0 +1,776 @@
...
+            DDLogDebug("EPC: Event logged to '\(stream)' but only the following streams are configured: \(streamConfigs.keys.joined(separator: ", "))")
+            return
+        }
+
+        guard let installID = self.installID else {
+            DDLogError("EPC: Missing install ID. Fallbacking to not in sample.")
+            return
+        }
+        
+        if !inSample(stream: stream, deviceID: installID) {
```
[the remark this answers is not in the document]
bearloga: I actually got rid of the second argument entirely since we have access to `installID` directly now and that can be retrieved inside `inSample`

WMF Framework/Event Platform Client/EPCNetworkManager.swift
bearloga: I like `postAllScheduled`

tonisevener:
@bearloga I tried my hand at adding the ability for you to use a `Session` instance that is configured with `waitsForConnectivity = true`, located [here](https://github.com/wikimedia/wikipedia-ios/commit/f6c3f75f8ac8de99fa28ab0a17a95287a31ec12f). Feel free to pull it in and test with it. Though `Session` instantiation was a bit of a mess (and still is), so I think this particular commit merits a separate write up and PR for another iOS engineer to review. I'll make that PR by EOW.

https://github.com/wikimedia/wikipedia-ios/commit/f6c3f75f8ac8de99fa28ab0a17a95287a31ec12f

WMF Framework/Event Platform Client/EPCNetworkManager.swift
```diff
@@ -0,0 +1,135 @@
...
+            queuedEvent = removeOutputBufferAtIndex(0)
+            if let queuedEvent = queuedEvent {
+                self.submit(url: queuedEvent.url, payload: queuedEvent.body)
+            }
+        }
+    }
+
+    private func submit(url: URL, payload: Any? = nil) {
+        let request = session.request(with: url, method: .post, bodyParameters: payload, bodyEncoding: .json)
+        let task = session.dataTask(with: request, completionHandler: { (_, response, error) in
```
tonisevener, line 91: Just curious, is there an easy way to batch post multiple events with one call? I thought I remember some mention of that being possible with the endpoint. It would be an quick win if so and you could avoid the `while` in `httpTryPost`.

tonisevener, line 91: One last thing before I forget - it would be good to ask the system to extend the background execution time before calling this. This [documentation](https://developer.apple.com/documentation/uikit/app_and_environment/scenes/preparing_your_ui_to_run_in_the_background/extending_your_app_s_background_execution_time) has a good write-up on it, though note the network call is synchronous in their sample code. I think you should be able to end the background task in the completionHandler here since this is asynchronous.

bearloga: Yep! We could definitely batch post multiple events with one call. From [T239996#5723920](https://phabricator.wikimedia.org/T239996#5723920):

> So with current settings, each message must be smaller than 1MB, and the entire POST body must be smaller than 4MB.

A conservative batch size would probably be 32 or 64 events. @tonisevener @linehan @mdholloway what do you think? Should we try batching in v1 or table it for next iteration and leave it one-at-a-time for now?

linehan: There are arguments both ways. I personally feel like batching is fine, but you can make the argument that if the network (or server) is unreliable, batching risks losing more data if the unreliability is uniform (which is probably isn't, so in practice there is maybe less of a difference between these two in terms of expected data loss). 

Setting the moral arguments to one side, and though I want to get to it, maybe it makes sense to wait until next rev. We don't actually have anything using the batching functionality yet that I know of, so there may be unknown unknowns there. Maybe we can revisit if everything else is locked down and we have time for a little extra work to prove it out before merge, though.

Insert usual warning sticker that `while` loops are the leading cause of non-halting programs, etc. etc.

tonisevener: 👍  sounds good!

linehan: Wow, cool, yeah I think we should definitely do this.

WMF Framework/Event Platform Client/EPCNetworkManager.swift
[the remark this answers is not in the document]
bearloga:
```
Referencing instance method 'popFirst()' on 'Collection' requires the types '[(url: URL, body: NSDictionary)]' and 'Array<(url: URL, body: NSDictionary)>.SubSequence' (aka 'ArraySlice<(url: URL, body: NSDictionary)>') be equivalent
```

So I just changed it to:

``` swift
func outputBufferPopFirst() -> (url: URL, body: NSDictionary)? {
  queue.sync {
    if self.outputBuffer.isEmpty {
      return nil
    }
    return self.outputBuffer.remove(at: 0)
  }
}
```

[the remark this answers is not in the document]
linehan: Lol... indeed how silly we all were to presume that a subset of an array be an array. How silly, when we know that Bread, once sliced, is Bread no longer, but is BreadSlice.

WMF Framework/Event Platform Client/EventPlatformClientLibrary.swift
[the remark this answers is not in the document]
bearloga: Just changed `NSCoding` to `Codable` in `submit` method and we have our answer:

```
Method cannot be marked @objc because the type of the parameter 2 cannot be represented in Objective-C
```

So it looks like if we want to be able to use this library in any of the existing instrumentation written in Obj-C we have to use `NSCoding`

WMF Framework/Event Platform Client/EPCNetworkManager.swift
bearloga:
```
'shared' is unavailable in application extensions for iOS: Use view controller based solutions where appropriate instead.
```

Let's table this for the revision?

WMF Framework/Event Platform Client/EPCNetworkManager.swift
bearloga: Just pushed a commit where this was done.

WMF Framework/Event Platform Client/EPCNetworkManager.swift
mdholloway: Did you try using [performExpiringActivity](https://developer.apple.com/documentation/foundation/processinfo/1617030-performexpiringactivity) instead as suggested at the bottom of the docs that Toni linked?

bearloga: @mdholloway I totally didn't see that! Good eyes. Alright, maybe I can work Toni's suggestion in after all.

tonisevener: Another thought I had is, move these EPC protocol conformances to `PeriodicWorker` and `BackgroundFetcher` into their own file, then move that file out of WMF and into the Wikipedia project (uncheck the WMF target and check the (Wikipedia + Staging + Experimental + User Testing targets, you also might want to move the file location out of the WMF area in the project directory). So the code to do periodic work lives in the app, but the rest of the library still lives in WMF Framework. Then you'd have to rework postAllScheduled() to take a completion block, and add the `beginBackgroundTask` code around it. @mdholloway 's solution sounds potentially faster and easier though, I'm just not familiar with it (we don't have any places in the codebase where we use `performExpiringActivity`).

```
//MARK: PeriodicWorker
extension EPC: PeriodicWorker {
    public func doPeriodicWork(_ completion: @escaping () -> Void) {
        self.postAllScheduled()
    }
}

//MARK: BackgroundFetcher
extension EPC: BackgroundFetcher {
    public func performBackgroundFetch(_ completion: @escaping (UIBackgroundFetchResult) -> Void) {
        doPeriodicWork {
            completion(.noData)
        }
    }
}
```

bearloga: Captured in wishlist for v2 ([T261987](https://phabricator.wikimedia.org/T261987))

bearloga: Captured in wishlist for v2 ([T261987](https://phabricator.wikimedia.org/T261987))

joewalsh (commented):
Looking great overall! I did my review in the form of proposed changes via a PR on this PR. The review comments/questions are in that PR: #3682

tonisevener:
I have been reviewing this as a part of the refactor PR and from my perspective these are ready to merge into `main`. I had one possible cleanup comment [here](https://github.com/wikimedia/wikipedia-ios/pull/3682#discussion_r488004790), but `client_dt` has tripped me up in the past so I could just be missing something. We're happy to discuss the refactor further in our meeting today.

bearloga:
From the meeting: We decided to move forward with this PR once #3682 is merged into it

tonisevener (approved):
🎉  🎊  Thanks everyone for their hard work on this! Very excited to get this in.

PR #3651, opened 2020-08-06
Switch to GitHub Actions for CI

joewalsh:
**Fixes Phabricator ticket:** 
https://phabricator.wikimedia.org/T259805

### Notes
* Increased test timeouts to work on slower environments
* Removed Beta Cluster Tests that are invalid after the switch to the Page Content Service
* Added GitHub Actions config, removed Jenkins Config
* Removed Jenkins-specific output settings from Fastlane
* Our build server will still be used for deploying


joewalsh:
> 1. Was the SecurityTests file just dead code?

Previously, the app had a specific entitlement to set the default file protection to `NSFileProtectionCompleteUntilFirstUserAuthentication`. However, that is the default setting, and having the entitlement set to that caused issues with automatic signing, so [it was removed](https://github.com/wikimedia/wikipedia-ios/pull/3496). `SecurityTests` was verifying that was actually set. Given that we're now staying with the default setting, `SecurityTests` can be removed.

> 2. What are the best ways to test the localization and testing flows?

For localization, I tested by setting the branch to`twn-test`, creating that branch off of this branch, and then pushing localization changes. For testing testing, it's been running on this PR (and any PR created off of this branch).

PR #3659, opened 2020-08-07
Add Picture of the day widget

staykids:
**Phabricator ticket:** https://phabricator.wikimedia.org/T258522

### Notes
* Still several little things to do (including this description), but wanted to go ahead and get this up if it's helpful.


Widgets/Widgets/PictureOfTheDayWidget.swift
```diff
@@ -0,0 +1,248 @@
...
+// MARK: - Widget
+
+struct PictureOfTheDayWidget: Widget {
+	private let kind: String = WidgetController.SupportedWidget.pictureOfTheDay.identifier
+
+	public var body: some WidgetConfiguration {
+		StaticConfiguration(kind: kind, provider: PictureOfTheDayProvider(), content: { entry in
+			PictureOfTheDayView(entry: entry)
+		})
+		.configurationDisplayName("Picture of the day")
```
mcleinman, line 15: These should be translated, no?

Widgets/Widgets/PictureOfTheDayWidget.swift
```diff
@@ -0,0 +1,248 @@
...
+	// MARK: TimelineProvider
+
+	func placeholder(in: Context) -> PictureOfTheDayEntry {
+		return dataStore.placeholderEntry
+	}
+    
+	func getTimeline(in context: Context, completion: @escaping (Timeline<PictureOfTheDayEntry>) -> Void) {
+		dataStore.fetchLatestAvailablePictureEntry { entry in
+			let currentDate = Date()
+			let timeline = Timeline(entries: [entry], policy: .after(currentDate.dateAtMidnight() ?? currentDate))
```
mcleinman, line 125: I'm not dealing with it yet, but been thinking about this for the widget I'm working on, and need to dig into this to investigate more.

It's not always just midnight, right? I'm not sure when these refresh on the server. But even if it's midnight in some timezone, no guarantees the phone is in that timezone. This is definitely more pressing for "On This Day", but I'm going to need to figure out when we can get refreshed data so we're not surfacing day-old "On This Day" data. (I'm just providing comments as I think of them at this point - not a full review yet.)

staykids: Thanks for peeking at all this. 

The way I understand the [Picture of the day on Commons](https://commons.wikimedia.org/wiki/Commons:Picture_of_the_day) is that they are often determined much in advance. So today, I could potentially request [the picture for a week from now](https://en.wikipedia.org/api/rest_v1/#/Feed/aggregatedFeed). We _could_ construct a timeline for all those forthcoming days, but the approach taken here instead is more DRY-style reuse of the existing data store/data the app already has rather than constructing a network stack in the widget. 

I wouldn't read too much into the `dateAtMidnight` stuff here for this widget, I'll probably end up ripping it out. Ideally, I think the refresh scenario for this widget will be something like this:
1. On main app's background refresh, reload this widget (work to do that already in here)
2. On any manual refresh of the explore feed in the main app, reload this widget
3. Every few hours (using the timeline refresh after policy mechanism above), refresh the widget.

I imagine a strategy like this might also be useful for the other widgets (this was why I made [WidgetController](https://github.com/wikimedia/wikipedia-ios/pull/3659/files#diff-40c9f930826459cea23eaa885cfa4f16) and defined the widget identifiers there – we can choose to reload only specific widgets here if we want to if some contextually relevant action occurs in the app).

Honestly the data modeling I'm using here is something I want to revisit again after I work on the Top read widget's data layer. I think there are probably some simplifications and changes I could make for the better after I dig into the Core Data store a little more. Feel free to share though if you figure out something that feels better suited for these widgets

Widgets/Widgets/PictureOfTheDayWidget.swift
staykids: Indeed, localization is on my list of what to figure out how to do "the SwiftUI" way. If you figure out best practices here, lemme know, otherwise I'll dig into it again when I next pass this widget. As I understood it, the pictures of the day are English-only, so this wasn't an immediate priority I had when first working on this widget.

mcleinman: Sounds good. In the widget I'm working on, I'm using translated strings just like anywhere else in the app. And since we have similar work for the explore feed, all the strings I needed already existed.

Wikipedia/Code/ExploreViewController.swift
```diff
@@ -640,6 +640,8 @@ class ExploreViewController: ColumnarCollectionViewController, ExploreCardViewCo
     }
     
     func collectionViewUpdater<T: NSFetchRequestResult>(_ updater: CollectionViewUpdater<T>, didUpdate collectionView: UICollectionView) {
+		WidgetController.shared.reloadAllWidgets()
```
tonisevener, line 643: I wonder if we ought to put these `reloadAllWidgets` calls somewhere closer to the Explore feed data refresh layer, maybe in `WMFExploreFeedContentController`. This one in particular I'm wondering how it will act if the Explore feed is not on screen (which you can see via Settings > Explore feed > turn off Explore tab toggle at the bottom). It may be no difference though, I would also expect `WMFExploreFeedContentController` to not be around or active if the user has disabled Explore on screen.

Widgets/Widgets/PictureOfTheDayWidget.swift
```diff
@@ -0,0 +1,255 @@
...
+	// MARK: Properties
+
+	static let shared = PictureOfTheDayData()
+
+	private var imageInfoFetcher = MWKImageInfoFetcher()
+	private var dataStore: MWKDataStore {
+		MWKDataStore.shared()
+	}
+
+	let sampleEntry = PictureOfTheDayEntry(date: Date(), image: #imageLiteral(resourceName: "PictureOfTheYear_2019"), imageDescription: "Two bulls running while the jockey holds on to them in pacu jawi (from Minangkabau, \"bull race\"), a traditional bull race in Tanah Datar, West Sumatra, Indonesia. 2015, Final-45.")
```
tonisevener, line 46: We should localize imageDescription.

Widgets/Widgets/PictureOfTheDayWidget.swift
```diff
@@ -0,0 +1,255 @@
...
+	let placeholderEntry = PictureOfTheDayEntry(date: Date(), contentDate: nil, contentURL: nil, imageURL: nil, image: nil, imageDescription: nil)
+
+	// MARK: Public
+
+	func fetchLatestAvailablePictureEntry(usingImageCache: Bool = false, _ completion: @escaping (PictureOfTheDayEntry) -> Void) {
+		// We could Result type the completion, but it's not probably worth the complexity for the widget's use cases
+		guard let contentGroup = dataStore.viewContext.newestGroup(of: .pictureOfTheDay), let imageContent = contentGroup.contentPreview as? WMFFeedImage else {
```
tonisevener, line 53: One thing I've wondered about when seeing `viewContext.newestGroup(of:)` is whether it'll scoop up the wrong (non-primary) language POTD content groups, simply because it's considered "newer" than the latest primary language content group. Doesn't seem like these managed object utilities prioritize their primary language but I could be missing it somewhere.

tonisevener (commented):
Just a couple of comments that are more open questions that I'm not sure of on the answer, and another little localization thing.

Widgets/Widgets/PictureOfTheDayWidget.swift
tonisevener: FYI I did this in a PR feedback commit.

Widgets/Widgets/PictureOfTheDayWidget.swift
tonisevener: I updated to `newestVisibleGroup`, per discussion that users intentionally may have hidden a particular card on the Explore feed and wouldn't want to see it again. That discussion also confirmed that this method already takes primary language into account.

Wikipedia/Code/ExploreViewController.swift
tonisevener: I moved the placement of these reload calls into `WMFExploreFeedContentController.m`, but I did confirm those calls aren't hit (nor is this one) if the Explore tab is hidden via app settings. So a network call fallback might be in order at some point to handle that edge case.

tonisevener: Update: `WMFExploreFeedContentController.m` is hit if I lower the refresh threshold. So I think we're good there, though the widget still had trouble refreshing when the Explore feed was hidden so not sure where it's getting dropped. We can look into it in a followup PR.

tonisevener (approved):

PR #3664, opened 2020-08-14
Fix scroll to section on redirect to section of another article

joewalsh:
**Fixes Phabricator ticket:** 
https://phabricator.wikimedia.org/T260429

**Hold pending triage of the ticket**

### Notes
Consolidates article scroll state restoration logic. There are a few different areas we were attempting to scroll to a given area of the page and potentially needed to wait for that scroll offset to become available as the page is loading. This handles those states through a single enum and consolidated logic for retrying after failing to find the offset.

### Test Steps
See ticket for the repro steps of the bug. Also test state restoration (open an article, scroll down the page, force quit the app, re-open) and scroll offset restoration after rotating the device.


tonisevener:
Sorry for the delay on reviewing this! Here's where I'm at with it:

I found that Artie's web view wouldn't unhide after restoring the navigation state from launch, so I pushed a commit fixing that. Before we called `setWebViewHidden(false)` whether we restored via an offset or an anchor so I tried to match that in this PR. I'm also finding a few other bugs:

1. Scrolling to the Artie fragment from the Sopranos fails about 50% of the time for me. After digging a little bit more it seems that the JS bridge `scroll_to_anchor` is not called in these failure situations.
2. When restoring a deep navigation stack, after pressing Back from the screen I land on after launch, the previous web view is shown, then jumps to the previous offset. In `main` it seems like the web view is hidden initially until the article is scrolled to the offset, then shown, resulting in a smoother experience.
3. I made a few fragments on TEST Wikipedia to see if we can fix additional cases while we're here. Neither of these work on `main` or in this PR: A) Tap the "grew up" link under "Early life and education" in the "Erna Solberg" article. and B) Tap the "anchor on another page" link on the "Triple3" article.

No pressure to fix these @joewalsh (hope you're doing well!), I can pick away at them. Just wanted to get my notes down so I know where I left off.

tonisevener:
Update - I think I have fixed the first issue, albeit in a bit of a gross way. Will look at the remaining two next week.

tonisevener (approved):

PR #3668, opened 2020-08-23
Use built-in Publishers.MergeMany instead of CombineExt

joewalsh:
### Notes
There's a built-in publisher in `Combine`, `Publishers.MergeMany`, that can be used for merging many publishers instead of `CombineExt`. Even though it's removed in this PR, there are other aspects of `CombineExt` that could be useful as we implement more functionality with `Combine`, so we should keep it in mind.


joewalsh:
Self merging due to the extremely low impact and to remove the `CombineExt` fork while doing a clean up pass on the [list of GitHub hosted repos](https://www.mediawiki.org/wiki/Gerrit/GitHub). Feel free to post-review.

PR #3671, opened 2020-08-26
Migrate MWKLanguageLinkController to Core Data

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T257250 

### Notes

When `UserDefaults` were [migrated off the named suite](https://github.com/wikimedia/wikipedia-ios/pull/3439) into using the standard suite, the today widgets stopped accurately polling the currently selected app language. This was because the today widgets were referencing the standard `UserDefaults` suite, which isn't shared between the main app and the extensions. Joe migrated the app language component here from `UserDefaults` into Core Data, which should mitigate the issue.

I'll review this one myself because it relates to some widget stuff I'm working on. (It won't let me self-assign for review, so I'll self-assign as an assignee).

### Test Steps
1. Open app, change primary language
2. Open today widgets (e.g. Top read), confirm the widget's language matches the primary language selected in app

### DIFF
--- a/TopReadWidget/WMFTodayTopReadWidgetViewController.swift
+++ b/TopReadWidget/WMFTodayTopReadWidgetViewController.swift
@@ -63,12 +63,13 @@ class WMFTodayTopReadWidgetViewController: ExtensionViewController, NCWidgetProv
     override func viewDidLoad() {
         super.viewDidLoad()
         
-        guard let appLanguage = MWKLanguageLinkController.sharedInstance().appLanguage else {
+        userStore = MWKDataStore.shared()
+
+        guard let appLanguage = userStore.languageLinkController.appLanguage else {
             return
         }
     
         siteURL = appLanguage.siteURL()
-        userStore = MWKDataStore.shared()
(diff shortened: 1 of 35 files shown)


staykids:
Nice - thanks again for unblocking me with this one @joewalsh. I believe one potential code smell is `migratePreferredLanguagesToManagedObjectContext` – since it still references `NSUserDefaults`, if it's called from an extension target that hasn't yet migrated the `WMFKeyValue` would differ from the app preference. But if I'm tracing the call stack properly, it appears the only path into that migration occurs in the main app target itself, so in practice it's probably not problematic. I pushed a comment on that method just for posterity.

joewalsh:
@staykids correct, the migration is only called by `WMFAppViewController` in the main app. Thanks for adding the comment to clarify! If we want to go one step further, we could add a preprocessor macro to the app targets, something like `MAIN_APP`, and ensure the migration code is only shipped with the main app. I couldn't find a way to target non-extension codes with `if @available` but that approach would probably be preferable if it exists.

PR #3673, opened 2020-08-28
Add Top Read widget

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T259841

Details forthcoming.


Widgets/Widgets/TopReadWidget.swift
```diff
@@ -0,0 +1,371 @@
...
+	// MARK: Public
+
+	private var dataStore: MWKDataStore {
+		MWKDataStore.shared()
+	}
+
+	func fetchLatestAvailableTopRead(_ completion: @escaping (TopReadEntry) -> Void) {
+		guard
+			let appLanguage = dataStore.languageLinkController.appLanguage,
+			let topRead = dataStore.viewContext.group(of: .topRead, for: Date(), siteURL: appLanguage.siteURL()),
```
tonisevener, line 43: This looks like it does what I was thinking of with POTD - fetching the latest content group for today and the primary language. I was also going to suggest we trigger a widget reload after the user changes their primary language in the app. Though on further thought hooking into Explore Feed's content reloading might cover it since that updates with a primary language change as well.

tonisevener: Just a documentation comment, confirmed that updating languages hits `WMFAppViewControllers.m's `preferredLanguagesDidChange` method, which eventually calls the `updateFeedSourcesWithDate` method where the `reloadAllWidgets` call lives. So no need for separate reload call on language change.

tonisevener (approved):

PR #3677, opened 2020-09-03
Deep link into On This Day

mcleinman:
**Fixes Phabricator ticket:**  https://phabricator.wikimedia.org/T172759

### Notes
* This is needed for opening the On This Day Widget (https://phabricator.wikimedia.org/T259840).
* @carolynlimadeo - Please confirm these views are what we want. (She and I went over the details previously, but I want to confirm this is as expected.) Can make tweaks in this PR (or later on).
* We're now opening this in a navigation stack, and so the header changes (and we no longer have an ending "back to Explore Feed" button).
* This only grabs the deep link for English Wikipedia's On This Day (https://en.wikipedia.org/wiki/Wikipedia:On_this_day/Today). It doesn't grab the pages that feed into On This Day (https://en.wikipedia.org/wiki/Wikipedia:Selected_anniversaries/August_31), as those are editable and perhaps the user wants to edit them. Additionally, it doesn't deep link for On This Day in non-English languages. This works for now - as we need to finish this up in order for the widget to work - but I've made a separate Phab ticket regarding these simplifications (and assigned to Josh).

### Test Steps
1. Test on iPad and iPhone.
1. Run app. Open On This Day card from Explore feed. Confirm it looks normal (especially with the header and button on bottom). Make sure you can open an article.
1. In a terminal window, run `xcrun simctl openurl booted 'https://en.wikipedia.org/wiki/Wikipedia:On_this_day/Today'`. This should open an On This Day VC. Confirm the header looks like the mock ups below, and there is no "back to explore feed" at the bottom. Make sure you can open an article.
1. In a terminal window, run `xcrun simctl openurl booted 'https://en.wikipedia.org/wiki/Wikipedia:On_this_day/Today?5'`. Confirm that this opens the VC to the sixth event.
1. In a terminal window, run `xcrun simctl openurl booted 'https://en.wikipedia.org/wiki/Wikipedia:On_this_day/Today?afw'`. Confirm this opens and doesn't scroll. Garbage argument shouldn't crash the app or anything, it should just silently fail.

### Screenshots
Screenshot 29


Wikipedia/Code/OnThisDayViewController.swift
```diff
@@ -88,6 +115,34 @@ class OnThisDayViewController: ColumnarCollectionViewController, DetailPresentin
...
+        let locale = NSLocale.wmf_locale(for: language)
+        let semanticContentAttribute = MWLanguageInfo.semanticContentAttribute(forWMFLanguage: language)
+
+        headerView.topSmallLine.semanticContentAttribute = semanticContentAttribute
+        headerView.middleLargeLine.semanticContentAttribute = semanticContentAttribute
+        headerView.bottomSmallLine.semanticContentAttribute = semanticContentAttribute
+
+        headerView.topSmallLine.text = CommonStrings.onThisDayTitle.uppercased()
+
+        let eventCountText = String(format: WMFLocalizedString("on-this-day-detail-header-title", language: language, value:"{{PLURAL:%1$d|%1$d historical event|%1$d historical events}}", comment:"Title for 'On this day' detail view - %1$d is replaced with the number of historical events which occurred on the given day"), locale: locale, events.count).uppercased(with: locale)
```
tonisevener, line 133: I suggest moving this WMFLocalizedString (and the one in OnThisDayViewControllerHeader) to CommonStrings.swift.

Wikipedia/Code/OnThisDayViewController.swift
```diff
@@ -88,6 +115,34 @@ class OnThisDayViewController: ColumnarCollectionViewController, DetailPresentin
...
+
+        headerView.middleLargeLine.text = DateFormatter.wmf_monthNameDayNumberGMTFormatter(for: language).string(from: midnightUTCDate)
+
+        if let firstEventEraString = events.first?.yearString, let lastEventEraString = events.last?.yearString {
+            let dateRangeText = String(format: WMFLocalizedString("on-this-day-detail-header-date-range", language: language, value:"from %1$@ - %2$@", comment:"Text for 'On this day' detail view events 'year range' label - %1$@ is replaced with string version of the oldest event year - i.e. '300 BC', %2$@ is replaced with string version of the most recent event year - i.e. '2006', "), locale: locale, lastEventEraString, firstEventEraString)
```
tonisevener, line 138: Same for this string

tonisevener (changes requested):
Just some small things from looking at the code, going to run and test it now.

Wikipedia/Code/ViewControllerRouter.swift
```diff
@@ -76,6 +76,17 @@ class ViewControllerRouter: NSObject {
                 return false
             }
             return presentOrPush(talkPageVC, with: completion)
+        case .onThisDay(let indexOfSelectedEvent):
+            let dataStore = MWKDataStore.shared()
+            guard let contentGroup = dataStore.viewContext.newestGroup(of: .onThisDay), let onThisDayVC = contentGroup.detailViewControllerWithDataStore(dataStore, theme: theme) as? OnThisDayViewController else {
```
tonisevener, line 81: @mcleinman just thought of one other thing - could this pull a non-EN Wikipedia on this day group too, since you're only checking for the latest? Maybe we can add a language check here too. I can look into this later today.

tonisevener:
I see a couple of layout issues (screenshots attached). I think they can be done later in a followup PR though, doesn't have to be fixed for the beta IMO. I can also look into this later this afternoon if I have time. I tested on iPhone X iOS12 and iPad iOS14

iPad & iPhone X - looks like you need to align to a safe area / layoutMargin, it pushes over to the left too far. I also notice when rotating a large content inset shows up. Often this just requires a call to our `updateScrollViewInsets` method at the right time. 

Screen Shot 2020-09-03 at 1 03 31 PM

Screen Shot 2020-09-03 at 12 22 48 PM

Screen Shot 2020-09-03 at 12 22 16 PM

mcleinman:
Also - until it scrolls, it shows the title. We don't want a title showing on launch if the header is visible.
Screenshot 34

tonisevener (approved):

Wikipedia/Code/ViewControllerRouter.swift
mcleinman: From changing the language of the app, this seems to work as expected.

PR #3679, opened 2020-09-04
Handle underbar header when programatically scrolling

mcleinman:
**Fixes Phabricator ticket:** none, bug came from PR https://github.com/wikimedia/wikipedia-ios/pull/3677

### Notes
* A programatic scroll doesn't call the relevant delegate functions that allow `NavigationBarHider` to work. This adds `isProgramaticallyScrolling` to allow it to work in these case.

### Test Steps
1. Run app.
1. In a terminal window, run `xcrun simctl openurl booted 'https://en.wikipedia.org/wiki/Wikipedia:On_this_day/Today'`. Ensure `On This Day` VC opens appropriately, with header looking good. Scroll up and down, nothing should jump.
1. In a terminal window, run `xcrun simctl openurl booted 'https://en.wikipedia.org/wiki/Wikipedia:On_this_day/Today?0'`. Ensure `On This Day` VC opens appropriately, with header looking good. Scroll up and down, nothing should jump.
1. In a terminal window, run `xcrun simctl openurl booted 'https://en.wikipedia.org/wiki/Wikipedia:On_this_day/Today?1'`. Ensure `On This Day` VC opens appropriately, with header looking good. Scroll up and down, nothing should jump.
1. In a terminal window, run `xcrun simctl openurl booted 'https://en.wikipedia.org/wiki/Wikipedia:On_this_day/Today?6'`. Ensure `On This Day` VC opens appropriately, with header looking good. Scroll up and down, nothing should jump.

### DIFF
--- a/Wikipedia/Code/NavigationBarHider.swift
+++ b/Wikipedia/Code/NavigationBarHider.swift
@@ -45,6 +45,10 @@ public class NavigationBarHider: NSObject {
     }
 
     public func setIsProgramaticallyScrolling(_ isProgramaticallyScrolling: Bool, yOffset: CGFloat, topContentInset: CGFloat) {
+        guard navigationBar?.isInteractiveHidingEnabled == true else {
+            return
+        }
+        
         self.isProgramaticallyScrolling = isProgramaticallyScrolling
 
         if isProgramaticallyScrolling {
@@ -65,9 +69,11 @@ public class NavigationBarHider: NSObject {
             return
         }
 
-        /// When programatically scrolling on a just-created scroll view, we need a layout pass to get accurate heights.
-        scrollView.layoutIfNeeded()
-        navigationBar.layoutIfNeeded()
+        if isProgramaticallyScrolling {
+            /// When programatically scrolling on a just-created scroll view, we need a layout pass to get accurate heights.
+            scrollView.layoutIfNeeded()
+            navigationBar.layoutIfNeeded()
+        }
 
         guard scrollView.contentSize.height > 0 else {
             if navigationBar.isAdjustingHidingFromContentInsetChangesEnabled  {
--- a/Wikipedia/Code/NavigationBarHider.swift
+++ b/Wikipedia/Code/NavigationBarHider.swift
@@ -32,23 +32,43 @@ public class NavigationBarHider: NSObject {
     
     fileprivate var isUserScrolling: Bool = false
     fileprivate var isScrollingToTop: Bool = false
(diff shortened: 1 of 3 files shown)


Wikipedia/Code/NavigationBarHider.swift
```diff
@@ -32,23 +32,43 @@ public class NavigationBarHider: NSObject {
...
+    private func setDataForScroll(yOffset: CGFloat, topContentInset: CGFloat) {
+        guard let navigationBar = navigationBar else {
+            return
+        }
+        initialScrollY = yOffset + topContentInset
+        initialNavigationBarPercentHidden = navigationBar.navigationBarPercentHidden
+    }
+
+    public func setIsProgramaticallyScrolling(_ isProgramaticallyScrolling: Bool, yOffset: CGFloat, topContentInset: CGFloat) {
+        self.isProgramaticallyScrolling = isProgramaticallyScrolling
```
tonisevener, line 48: Might want a `guard let isInteractiveHidingEnabled == true` here, just to confirm that the user has set the right navbar settings before you update these internal state properties.

Wikipedia/Code/NavigationBarHider.swift
```diff
@@ -32,23 +32,43 @@ public class NavigationBarHider: NSObject {
...
     }
 
     @objc public func scrollViewDidScroll(_ scrollView: UIScrollView) {
         guard let navigationBar = navigationBar else {
             return
         }
-        
+
+        /// When programatically scrolling on a just-created scroll view, we need a layout pass to get accurate heights.
+        scrollView.layoutIfNeeded()
```
tonisevener, line 69: If it only seems needed when programmatically scrolling, might want to wrap these in `if isProgramaticallyScrolling {` to limit unneeded layout passes in other states.

tonisevener (changes requested):
Just a couple of small suggestions. Having trouble installing the widgets though to test (Widgets extension is now missing, but nothing changed in the project file in this PR diff, so I'm very confused!)

Wikipedia/Code/NavigationBarHider.swift
mcleinman: Good call, updated!

Wikipedia/Code/NavigationBarHider.swift
mcleinman: Good call, updated!

tonisevener (approved):

PR #3680, opened 2020-09-08
Use BackgroundTasks.framework on iOS 13 and above

joewalsh:
### Notes
Potential fix for issues with infrequent background updates 

### Test Steps
0. Ensure background updates are enabled for the app in system Settings
1. Use the app during the day
2. Don't use the app overnight
3. Check the app the next day to verify content has loaded


staykids (approved):
👍 nice. Left a device running this branch on overnight, and it _seemed_ to work. Read over some docs and the API and seems like our bases are covered here. Also, followed the steps [here](https://developer.apple.com/documentation/backgroundtasks/starting_and_terminating_tasks_during_development) to test triggering the background task code path on device and it seemed to be hitting it as expected.

Crossing my fingers that this magically resolves all our widget reload issues 🤞

PR #3682, opened 2020-09-09
Update Event Platform Client library to utilize Codable and concrete types where possible, fix minor issues

joewalsh:
One of the challenges we had in the past with event logging was that utilizing unstructured dictionaries made it easy to mistype a key or have the structure of the dictionary wrong. This proposal utilizes Swift's built in `Codable` protocol to allow events to be defined as structs so the types and structure are enforced. It also creates enums for `Stream`s and `Schema`s so they're defined from a concrete set and can be re-used without the potential for mistyping.

This change removes the ability to log from Objective-C. Any place we currently log from Objective-C should be converted to Swift when updating to the MEP. 

There were a few minor fixes made here as well:
- Fixed a missing completion call in `doPeriodicWork` that was causing a background task to timeout
- Fixed a potential race condition between `isEmpty` and `popFirst` - both checks could be combined into one


bearloga:
This is a major enough patch that we shouldn't rush through it, there are also fundamentally different ideas being introduced here that need to be deliberated on. I've added this to discussion for v2 ([T261987](https://phabricator.wikimedia.org/T261987)).

WMF Framework/EventStreams.swift
```diff
@@ -0,0 +1,26 @@
...
+     * Schema specifies which schema (and specifically which version of that schema)
+     * a given event conforms to. Analytics schemas can be found in the jsonschema directory of
+     * [secondary repo](https://gerrit.wikimedia.org/g/schemas/event/secondary/).
+     * As an example, if instrumenting client-side error logging, a possible
+     * `$schema` would be `/mediawiki/client/error/1.0.0`. For the most part, the
+     * `$schema` will start with `/analytics`, since there's where
+     * analytics-related schemas are collected.
+     */
+    public enum Schema: String, Codable {
+        case editHistoryCompare = "/analytics/mobile_apps/ios_edit_history_compare/1.0.0"
```
bearloga, line 23: I'm a bit apprehensive about this.

The schema is a contract between the event producer and the event ingester. Without this, each piece of instrument declares explicitly which schema and which version of that schema it conforms to by including `$schema` in the event data.

When a schema is updated to have more fields or more allowed values in enum fields and the instrumentation is updated to match, I worry that having the schema+version centralized here, separate from the instrumentation, will make it too easy to forget to update. Of course, if the schema is declared explicitly in the same file as the instrumentation code it's also entirely possible to miss in that case too, but not as easy as if it's different locations.

Also, I really really hope that this isn't taken as me underestimating anybody!

joewalsh: That's a good point! It would likely help with maintainability to have the schema and the event structure in the same place. I will push a patch that moves the `schema` definition inside the structure of the event so we can see how it feels.

WMF Framework/EventPlatformClientLibrary.swift
```diff
@@ -134,10 +126,21 @@ public class EPC: NSObject {
...
+    /**
+     * An individual stream's configuration.
+     */
+    private struct StreamConfiguration: Codable {
+        let sampling: Sampling?
+        struct Sampling: Codable {
+            let rate: Double?
+            let identifier: String?
+        }
+    }
```
bearloga, line 138: What would happen when deserializing to this if a stream's config included additional fields besides `sampling`?

Let's say the returned stream configurations suddenly look like this:

```json
{
  "streams": {
    "ios.edit_history_compare": {},
    "ios.daily_checkin": {
      "tags": { "anonymous": true }
    },
    "ios.some_old_stream": {
      "tags": { "active": false }
    }
  }
}
```

joewalsh: In your example, the decoding would succeed, `tags` would be ignored and `sampling` would be `nil`.

bearloga: Awesome! :D

WMF Framework/EventStreams.swift
joewalsh: I pushed this change and I agree it helps. Here's what an event looks like after the update:
```swift
    private struct Event: EventInterface {
        static let schema = "/analytics/mobile_apps/ios_edit_history_compare/1.0.0"
        let action: Action
        let primary_language: String
        let is_anon: Bool
    }
```

tonisevener (commented):
I'm seeing the POSTs fail on this branch with 400 in Charles, though the completion block doesn't return an error so nothing is printed to the log. I haven't figured out what is causing the failure yet.

joewalsh:
> I'm seeing the POSTs fail on this branch with 400 in Charles, though the completion block doesn't return an error so nothing is printed to the log. I haven't figured out what is causing the failure yet.
@tonisevener thanks for finding this! Should be fixed now.

WMF Framework/EventStreams.swift
tonisevener: @bearloga would we ever need to share schema identifiers across multiple types of event logs? That's the only potential problem I see with this approach, multiple string literals of the same identifier peppered throughout as we move forward. But I could definitely be misremembering how the backend works and streams vs. schemas.

joewalsh: Another potential option would be to keep the `static` definition in the struct but use `EPC.Schema` for the type instead of `String`:
```swift
private struct Event: EventInterface {
        static let schema: EPC.Schema = .editHistoryCompareV1
        let action: Action
        let primary_language: String
        let is_anon: Bool
    }
```

bearloga: Quick recap of MEP to clarify some of this stuff: In the legacy system stream=table=schema. In the modern system schemas and streams have a 1:many relationship. Multiple streams can declare the same schema, and that schema specifies what the incoming the data should look like and how the table in the database looks like. Each stream gets its own table.

When you produce an event to the stream there's 3 conditions which are checked: (1) that the stream exists in the stream config, (2) that the title of the schema in the stream config matches the schema declared by the event, and (3) that the event data conforms to the schema it claims to be conforming to.

I can imagine having a more generic schema that is re-used by multiple instruments that send data to different streams. For example, we can generalize `ios_edit_history_compare` schema by making the `action` field a free-form string instead of an enum like it currently is.

joewalsh: Makes sense, thanks for the recap! Given that we might re-use schemas, I pushed the change that restores the central definition of schemas, but keeps the per-event static schema. This way we could re-use them but would also have them declared on the event struct so it's easier to remember it should be updated when changing the event's structure.

WMF Framework/EventPlatformClientLibrary.swift
```diff
@@ -389,63 +386,71 @@ public class EPC: NSObject {
...
+            group.enter()
+            httpPost(url: item.url, body: item.body) { result in
+                switch result {
+                case .success:
+                    break
+                case .failure(let error):
+                    switch error {
+                    case .networkingLibraryError:
+                        /// Retry on networking library failure
+                        self.appendPostToOutputBuffer(item)
```
tonisevener, line 441: Seems like this could get stuck in a retry loop, maybe we should retry up to a max number of times it fails with this error?

WMF Framework/EventPlatformClientLibrary.swift
```diff
@@ -389,63 +386,71 @@ public class EPC: NSObject {
...
+
+            // Process event buffer after making stream configs available
+            // NOTE: If any event is re-submitted while streamConfigurations
+            // is still being set (asynchronously), they will just go back to
+            // input buffer.
+            while let event = inputBufferPopFirst() {
+                guard inSample(stream: event.stream) else {
+                    continue
+                }
+                appendPostToOutputBuffer((url: streamIntakeServiceURI, body: data))
```
tonisevener, line 417: `body` should be `event.data`, not `data` which is the stream configuration data.

WMF Framework/EventPlatformClientLibrary.swift
```diff
@@ -597,98 +660,59 @@ public class EPC: NSObject {
...
+    }
 
-        if !(streamConfigs.keys.contains(stream)) {
-            DDLogDebug("EPC: Event submitted to '\(stream)' but only the following streams are configured: \(streamConfigs.keys.joined(separator: ", "))")
-            return
-        }
-        
-        if !inSample(stream: stream) {
+    /// Private, synchronous version of `submit`.
+    private func _submit<E: EventInterface>(stream: Stream, event: E, domain: String? = nil, date: Date? = nil){
```
tonisevener, line 670: This might be more of a comment with the original PR and less with the refactor, but is there any need anymore for the `date` parameter? I don't see how it could be used, even if events are queued in the inputBuffer first and posted after the config fetches (which I think is how it was originally going to be used).

WMF Framework/EventPlatformClientLibrary.swift
joewalsh: Good catch - fixed. I added a struct for persisting event requests. This would have prevented the `event.data` bug below as well.

WMF Framework/EventPlatformClientLibrary.swift
joewalsh: My understanding of the original PR was that it's for providing a custom `client_dt` but if we don't plan to utilize that, it could be removed.

tonisevener (approved):

WMF Framework/EventPlatformClientLibrary.swift
tonisevener: Thanks! The request struct feels much better.

tonisevener:
From my perspective this with the original library can be merged into `main`. But first we can confirm how other folks feel about this later today.

WMF Framework/EventPlatformClientLibrary.swift
bearloga: `client_dt` needs to be the timestamp of when the event was originally submitted. The original PR checks for its presence in the event data and if it's not already there (meaning the event is being submitted for the first time and not re-submitted after stream config is available) it sets it. I'm not entirely sure where this is set in this PR.

bearloga:
From the meeting: We decided to move forward with this PR and merge the original PR (once this is merged into that)

PR #3685, opened 2020-09-11
Top read network fallback & in-memory images for snapshot

tonisevener:
### Test Steps
1. Verify that you see top read content on a clean install without opening the app.
2. Smoke test image loading in top read widgets.


staykids:
Seems to be working well so far, thanks for pulling in those changes. In regard to the test steps, it seems the widget extensions don't register with the widget gallery until after the app is first launched (unless I'm missing something). Same experience for you?

Widgets/Widgets/TopReadWidget.swift
```diff
@@ -89,6 +89,14 @@ final class TopReadData {
                 group.leave()
                 continue
             }
+            
+            if usingImageCache {
+                if let cachedImage = fetcher.memoryCachedImage(withURL: thumbnailURL) {
+                    rankedElements[index].image = cachedImage.staticImage
+                    group.leave()
+                    continue
+                }
```
joewalsh, line 98: It looks like `fetcher.fetchImage()` checks the memory cache first so the behavior would be the same if this goes on to fetch the image from the network. Should the continue be outside of the `if let cachedImage...` block so it doesn't go on to fetch the image from the network?

tonisevener: @joewalsh well I was thinking if there was no memory cached image, then fallback to the network just as a backup to try to get the image, but I guess that defeats the purpose of a quick method for `getSnapshot`. I didn't catch that `fetchImage` leans on memory anyways. I could move the continue, but keep this commit as a whole, so we are sure a fallback/network hit never occurs? Or if we are fine hitting the network as a backup, I could revert this commit & remove `usingImageCache` entirely.

joewalsh:
>Seems to be working well so far, thanks for pulling in those changes. In regard to the test steps, it seems the widget extensions don't register with the widget gallery until after the app is first launched (unless I'm missing something). Same experience for you?

I'm seeing that as well. However, I was able to open the app, close it without completing onboarding (so the feed doesn't load), add the top read widget, and verify it loaded content. That test for me was less about those repro steps and more about verifying the feed fetch worked from the widget extension

Widgets/Widgets/TopReadWidget.swift
```diff
@@ -323,8 +353,7 @@ struct TopReadOverlayView: View {
 	var content: some View {
 		VStack(alignment: .leading) {
 			if isExpandedStyle {
-				// TODO: Localize
-				Text("\(currentViewCountOrEmpty) readers")
+				Text(currentViewCountOrEmpty)
```
staykids, line 356: As per the designs, I believe this expanded style overlay (meaning, no image in the small widget family) still needs to include the "readers" text segment: `Text("\(currentViewCountOrEmpty) readers")`

widget_small_overlay

Widgets/Widgets/TopReadWidget.swift
joewalsh: Looking at it again, it seems the intent of the snapshot is to be as quick as possible, so it makes sense to not hit the network for images. I pushed an update in f8210e6...40d5bae that only hits the memory and file cache for snapshots, for the content itself as well as the images.

Widgets/Widgets/TopReadWidget.swift
tonisevener: @staykids This is still working for me, are you seeing issues when you run? `currentViewCountOrEmpty` included "readers" via the localized string. I made a commit updating the variable names to be more clear, hopefully that helps.

staykids: My bad, you're right - I missed that. Thanks for updating my vague variable name.

staykids (approved):
I'm going to bring this style of network fallback into POTD.

PR #3689, opened 2020-09-14
Improve iOS 14 deep-press back button behavior

mcleinman:
**Fixes Phabricator ticket:** https://phabricator.wikimedia.org/T262305

### Notes
Three things in this PR:
1. Like in normal navigation, remove the SearchController from the navigation stack once you've navigated more than one VC beyond the SearchController.
2. On the SearchController, improve the code a bit. No behavior changes.
3. Comment updates, for accuracy.

### Test Steps
1. On an iOS 14 device/simulator.
1. Open a few controllers. Deep press the back button. Ensure you see a list of all VCs you've tapped through, and can jump back to any of them.
1. Check for regressions: Open a Search Controller. Search, tap into an article*, and tap into another article. Tap back button twice, ensure you never see the Search Controller.
1. Confirm new code is good: Open a Search Controller. Search, tap into an article*, and deep press the back button. Ensure the first item is `Search`. Tap it, ensure you jump back to SearchController.
1. Confirm new code is good (con't): Search, tap into an article*, tap into another article, and deep press the back button. Ensure the `Search` item isn't visible in the list. Tap into a prior article, make sure you jump back to it as expected.

* On this screen, instead of `< Back` you should see `< Search`

### Screenshots/Videos
Screenshot 79


Wikipedia/Code/SearchViewController.swift
```diff
@@ -26,9 +31,6 @@ class SearchViewController: ArticleCollectionViewController, UISearchBarDelegate
             navigationBar.isAdjustingHidingFromContentInsetChangesEnabled = false
             searchBar.becomeFirstResponder()
         }
-
-        /// Terrible hack to make back button text appropriate for iOS 14 - need to set the title on `WMFAppViewController`. For all app tabs, this is set in `viewDidAppear`.
-        parent?.navigationItem.backButtonTitle = title
```
tonisevener, line 31: I think you need at least this part - I notice from the last Search tab, the root view controller is no longer updating to say "Search". (Launch (lands on Explore) > Search tab > Search for something and drill in > Long press back button > Root context menu item still says "Explore")

tonisevener (commented):
I noticed 1 regression, see below. Also I do see Places occasionally not updating it's root title either, but it's intermittent & is also in the 1756 beta build, so not caused by this PR. It tends to happen right after launch, so my guess is the location permissions modal messes up the setting of the parent navigationItem title.

Wikipedia/Code/SearchViewController.swift
mcleinman: 🤦 I should have really trusted that Matt-of-a-few-weeks-ago knew what he was doing. Thanks, reverted.

mcleinman:
Just updated the refactor to make it more swift-y. When looking at it again saw a better way to write it. I tested it and it seems to work just as well, but should get some eyes.

tonisevener (approved):
Working well for me!

PR #3690, opened 2020-09-14
Fix On This Day widget link

mcleinman:

### Notes
When tapping an On This Day widget, the On This Day View Controller is supposed to open with the highlighted event int the middle of the screen. This morning I noticed it was broken. Looks like it was a regression that snuck in.

### Test Steps
Add an "On This Day" widget. Note the event being shown. Tap the widget. Ensure the View Controller is scrolled to the event shown on the widget.


mcleinman:
On hold to figure out best branch to merge this into. Will likely be merged into whatever https://github.com/wikimedia/wikipedia-ios/pull/3678 is ultimately merged into, but after that merge happens.

Widgets/Widgets/OnThisDayWidget.swift
```diff
@@ -211,7 +211,8 @@ extension OnThisDayEntry {
             let latestEventYear = allEvents.first?.yearString,
             let article = previewEvents.first?.articlePreviews?.first,
             let year = previewEvent.year?.intValue,
-            let eventsCount = contentGroup.countOfFullContent?.intValue
+            let eventsCount = contentGroup.countOfFullContent?.intValue,
+            let previewEventIndex = allEvents.firstIndex(of: previewEvent)
```
tonisevener, line 215: It's probably near-impossible for this to fail but, just in case I think you should keep it out of the guard and if it's nil, default to the index-less contentURL below, otherwise use the one with an index. It would just be a shame if the deep link url was the only reason a TimelineEntry wasn't instantiated.

tonisevener (commented):
Just one little thing, but it's working well for me.

Widgets/Widgets/OnThisDayWidget.swift
mcleinman: Good point - on it!

mcleinman: Done. And one of the few times I'm comfortable with force unwrapping is for a URL from a static string. Though even then it makes me a tad uneasy. Let me know if you'd prefer it gone.

tonisevener (approved):

PR #3691, opened 2020-09-15
On This Day deep link improvements

mcleinman:
**Fixes Phabricator ticket:** https://phabricator.wikimedia.org/T261979

### Notes
* 3 of the items in the Phab ticket were already fixed by another PR, it seems.
* Safe area for header: Underbar issue was a pre-existing bug - It also shows up when viewing a reading list on a rotated X-sized device.
* On This Day restoration: Now saving what type of navigation header the VC should have.

### Test Steps
On this day restoration:
1. Open On This Day from Explore feed. Confirm: It should have an "X" to close the view.
1. Tap to an article.
1. In terminal, open On This Day via a deep link: `xcrun simctl openurl booted "https://en.wikipedia.org/wiki/Wikipedia:On_this_day/Today"` (alternatively: tap an On This Day widget) Confirm: it should have a normal nav bar
1. Tap an article.
1. Background the app, so the state is saved.
1. Re run the app. Tap the back button through the stack, and ensure the two On This Day VCs have their expecteed nav styles.

Header safe area:
1. Run app on a phone/simulator with a notch.
1. Open On This Day VC (from Explore feed or deep link).
1. Rotate phone so notch is on leading edge. Ensure the On This Day header respects the safe area.

### Screenshots/Videos
Screenshot

### DIFF
--- a/WMF Framework/NavigationBar.swift
+++ b/WMF Framework/NavigationBar.swift
@@ -268,8 +268,8 @@ public class NavigationBar: SetupView, FakeProgressReceiving, FakeProgressDelega
         underBarViewTopTitleBarBottomConstraint = titleBar.bottomAnchor.constraint(equalTo: underBarView.topAnchor)
         underBarViewTopBottomConstraint = topAnchor.constraint(equalTo: underBarView.topAnchor)
         
-        let underBarViewLeadingConstraint = safeAreaLayoutGuide.leadingAnchor.constraint(equalTo: underBarView.leadingAnchor)
-        let underBarViewTrailingConstraint = safeAreaLayoutGuide.trailingAnchor.constraint(equalTo: underBarView.trailingAnchor)
+        let underBarViewLeadingConstraint = leadingAnchor.constraint(equalTo: underBarView.leadingAnchor)
+        let underBarViewTrailingConstraint = trailingAnchor.constraint(equalTo: underBarView.trailingAnchor)
         
         extendedViewHeightConstraint = extendedView.heightAnchor.constraint(equalToConstant: 0)
         extendedView.addConstraint(extendedViewHeightConstraint)
--- a/WMF Framework/NavigationBar.swift
+++ b/WMF Framework/NavigationBar.swift
@@ -268,8 +268,8 @@ public class NavigationBar: SetupView, FakeProgressReceiving, FakeProgressDelega
         underBarViewTopTitleBarBottomConstraint = titleBar.bottomAnchor.constraint(equalTo: underBarView.topAnchor)
         underBarViewTopBottomConstraint = topAnchor.constraint(equalTo: underBarView.topAnchor)
         
-        let underBarViewLeadingConstraint = leadingAnchor.constraint(equalTo: underBarView.leadingAnchor)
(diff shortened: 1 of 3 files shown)


WMF Framework/NavigationBar.swift
```diff
@@ -268,8 +268,8 @@ public class NavigationBar: SetupView, FakeProgressReceiving, FakeProgressDelega
         underBarViewTopTitleBarBottomConstraint = titleBar.bottomAnchor.constraint(equalTo: underBarView.topAnchor)
         underBarViewTopBottomConstraint = topAnchor.constraint(equalTo: underBarView.topAnchor)
         
-        let underBarViewLeadingConstraint = leadingAnchor.constraint(equalTo: underBarView.leadingAnchor)
-        let underBarViewTrailingConstraint = trailingAnchor.constraint(equalTo: underBarView.trailingAnchor)
+        let underBarViewLeadingConstraint = safeAreaLayoutGuide.leadingAnchor.constraint(equalTo: underBarView.leadingAnchor)
+        let underBarViewTrailingConstraint = safeAreaLayoutGuide.trailingAnchor.constraint(equalTo: underBarView.trailingAnchor)
```
tonisevener, line 272: @mcleinman This area makes me a bit nervous since so many screens hit it. Is it possible for you to consider the safe area only in ThreeLineHeaderView? I realize we'll lose the the reading list detail fix but I'd rather lose that (or have neither layout fixes) than have a fix here, considering we won't have much time for regression testing. The state restoration code change is looking and testing fine. I'll switch to reviewing the other and will get back to this one afterwards.

tonisevener (changes requested):

WMF Framework/NavigationBar.swift
mcleinman: I can do that if we want. I first started fixing it on ThreeLineHeaderView, but then realized the problem should probably be fixed on a higher level. I definitely understand discomfort with touching `NavigationBar.swift` when we're going to release today/tomorrow. 

By searching for `addundernavigationbarview(`, it looks like it's only used in 12 places. The UI bug also exists in reading lists. From a quick glance at the other spots that use it, the bug may only exist for reading lists and On This Day.
Screenshot

It seems like to me that is where it should be fixed, though - in the under bar holder view, rather than needing developers to remember the margins for every view is passed to the under bar holder. If we don't want to make such a chunky fix right now, what if we didn't fix it for this release but put it in a PR to fix it on the next one (when we have more time to test)? Or if we do want to fix it at the lower level ThreeLineHeaderView - and if so, do we also want to fix the Reading List header at a lower level?

tonisevener: @mcleinman At this point I'm good with the former - don't fix for this release but put in a PR for the next one. I agree this seems like the right place to fix it. The one place that jumped out at me that might not make this the best placement is in `InsertMediaViewController`. The underbar view is the gray area with the "Select a file from Wikimedia Commons". It looks like it extends past the safe area by design, so we might need to tinker with that screen's setup if we want to keep the fix here. But we can worry about that for 6.7.1.

Screen Shot 2020-09-15 at 3 26 44 PM

mcleinman: Sounds good. Will update momentarily.

tonisevener (approved):

PR #3692, opened 2020-09-15
Use feed refresh as fallback for Picture of the day widget

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T262358

### Notes

Adds use of `updateFeedSourcesUserInitiated` as network fallback if user's data store doesn't contain current day's picture. Also updates the previously erroneous use of `memoryCachedImage` to `cachedImage` to expand pool of available cached data.

### Test Steps
Similar to https://github.com/wikimedia/wikipedia-ios/pull/3685. In general, if network connectivity is available, the placeholder 2019 entry should not be visible after the POTD widget is added to the home screen or today view.


Widgets/Widgets/PictureOfTheDayWidget.swift
```diff
@@ -37,41 +37,69 @@ final class PictureOfTheDayData {
...
+        }
+    }
+
+	// MARK: Private
+
+    private func fetchLatestAvailablePictureEntryFromNetwork(completion: @escaping (PictureOfTheDayEntry) -> Void) {
+        dataStore.feedContentController.updateFeedSourcesUserInitiated(false) {
+            let moc = self.dataStore.viewContext
+            moc.perform {
+                guard let latest = moc.newestVisibleGroup(of: .topRead) else {
```
tonisevener, line 61: `.topRead` here

staykids: 😅 Code Review: It Really Works!™️

Widgets/Widgets/PictureOfTheDayWidget.swift
```diff
@@ -159,7 +187,7 @@ struct PictureOfTheDayProvider: TimelineProvider {
 	}
 
 	func getSnapshot(in context: Context, completion: @escaping (PictureOfTheDayEntry) -> Void) {
-		dataStore.fetchLatestAvailablePictureEntry(usingImageCache: context.isPreview) { entry in
+		dataStore.fetchLatestAvailablePictureEntry(usingCache: context.isPreview) { entry in
```
tonisevener, line 190: I have noticed Top Read was changed to `usingCache: true` here instead of considering `context.isPreview`. I'm not sure which way is best but I thought worth pointing out the discrepancy.

Widgets/Widgets/PictureOfTheDayWidget.swift
```diff
@@ -37,41 +37,69 @@ final class PictureOfTheDayData {
...
             self.fetchImageLicense(canonicalPageTitle: canonicalPageTitle) { license in
                 let entry = PictureOfTheDayEntry(date: Date(), contentDate: contentDate, contentURL: contentURL, imageURL: imageThumbnailURL, image: fetchedImage.image.staticImage, imageDescription: imageDescription, license: license)
                 completion(entry)
             }
-		})
-	}
-
-	// MARK: Private
+        })
+    }
```
tonisevener, line 102: (it won't let me comment below so I'm doing it here) - I've been looking and I don't think the `imageInfoFetcher.fetchGalleryInfo` will run the completion on the main queue. We might need to wrap up that completion calls in a `DispatchQueue.main`

tonisevener: Update - I went ahead and made this change. Seems to be testing okay.

tonisevener (approved):

Widgets/Widgets/PictureOfTheDayWidget.swift
staykids: Nice catch – I got used to `fetchImage` on `ImageCacheController` wrapping the completion for me and forgot to do it myself here. Thanks for making the change.

Widgets/Widgets/PictureOfTheDayWidget.swift
staykids: I think we should probably change it in Top read to respecting `context.isPreview` – I'll make the change and spin up a PR and we can test it both ways. Of course the actual behavior is completely unknown to me, but I can imagine a scenario where the system maybe tries to opportunistically create a snapshot giving our widget a bunch of time to do whatever network operations we need, and if we aren't respecting that flag we may end up delivering less relevant information than we're actually able to. Apple's conceptual documentation on widgets does something similar for `getSnapshot` so probably worth sticking with the pattern until we discover it's wrong.

PR #3694, opened 2020-09-15
underbar should respect safe area by default, with option to override

mcleinman:
**Fixes Phabricator ticket:** part of already-completed https://phabricator.wikimedia.org/T261979

### Notes
* This is being pushed to 6.7.1 to have some more testing time.

### Test Steps
1. Run app on a phone/simulator with a notch.
1. Open On This Day VC (from Explore feed or deep link).
1. Rotate phone so notch is on leading edge. Ensure the On This Day header respects the safe area. 
1. Open a reading list (on saved tab). Rotate phone so notch is on leading edge. Ensure the title header (`x articles`/`[title of list]`) respects the safe area.
1. Edit a section of any article, tap the insert image icon.
1. Rotate phone so notch is on leading edge. Ensure the header goes edge-to-edge (does NOT respect the safe area).

### DIFF
--- a/WMF Framework/NavigationBar.swift
+++ b/WMF Framework/NavigationBar.swift
@@ -191,6 +191,16 @@ public class NavigationBar: SetupView, FakeProgressReceiving, FakeProgressDelega
 
     private var shadowHeightConstraint: NSLayoutConstraint!
     private var extendedViewHeightConstraint: NSLayoutConstraint!
+
+    private lazy var safeAreaUnderBarConstraints = [
+        safeAreaLayoutGuide.leadingAnchor.constraint(equalTo: underBarView.leadingAnchor),
+        safeAreaLayoutGuide.trailingAnchor.constraint(equalTo: underBarView.trailingAnchor)
+    ]
+
+    private lazy var fullWidthUnderBarConstraints = [
+        leadingAnchor.constraint(equalTo: underBarView.leadingAnchor),
+        trailingAnchor.constraint(equalTo: underBarView.trailingAnchor)
+    ]
     
     private var titleBarTopConstraint: NSLayoutConstraint!
     private var barTopConstraint: NSLayoutConstraint!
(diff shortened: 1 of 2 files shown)


mcleinman:
On hold, as this is destined for 6.7.1

mcleinman:
Off hold

WMF Framework/NavigationBar.swift
```diff
@@ -361,6 +361,20 @@ public class NavigationBar: SetupView, FakeProgressReceiving, FakeProgressDelega
...
+        didSet {
+            guard shouldUnderBarIgnoreSafeArea != oldValue else {
+                return
+            }
+
+            safeAreaLayoutGuide.leadingAnchor.constraint(equalTo: underBarView.leadingAnchor).isActive = !shouldUnderBarIgnoreSafeArea
+            safeAreaLayoutGuide.trailingAnchor.constraint(equalTo: underBarView.trailingAnchor).isActive = !shouldUnderBarIgnoreSafeArea
+
+            leadingAnchor.constraint(equalTo: underBarView.leadingAnchor).isActive = shouldUnderBarIgnoreSafeArea
+            trailingAnchor.constraint(equalTo: underBarView.trailingAnchor).isActive = shouldUnderBarIgnoreSafeArea
```
joewalsh, line 375: `constraint(equalTo:` creates a new constraint and doesn't return an existing one, so new constraints will be created each time `shouldUnderBarIgnoreSafeArea` is set. Could these be created and stored in properties in `setup()`?
```
underBarRespectSafeAreaConstraints = [underBarViewLeadingConstraint, underBarViewTrailingConstraint]
underBarIgnoreSafeAreaConstraints = [...]
```
Then you could toggle `isActive` on those arrays here

joewalsh (changes requested):

WMF Framework/NavigationBar.swift
mcleinman: My bad - I had it in my head that identically created constraints were treated as if they were the same. Not sure where I got that from. Will update this shortly.

joewalsh (approved):

PR #3703, opened 2020-09-18
Fixes 0xdead10cc crashes reported by the widgets

joewalsh:
**Fixes Phabricator ticket:** 
https://phabricator.wikimedia.org/T263247

### Major changes
* Untangles the app's model to allow the widgets to spin up and down the entire stack
* Removes singletons from the following classes:
  * `Session`
  * `WMFAuthenticationManager`
  * `CacheController`
  * `ArticleCacheController`
  * `ImageCacheController`
* Makes the appropriate adjustments to inject dependencies and clarify the relationships between the objects that used to be singletons. This should continue into defining appropriate protocols for the relationships, but was not done in most cases here to reduce the impact.
* Keeps the `MWKDataStore.shared` for the main app to mitigate impact of this PR. This is now the root of the app's entire model stack. The widgets now create a `MWKDataStore`, fetch the necessary info, and release it when complete. This frees up any shared resources and locks on shared files, which fixes the `0xdead10cc` crashes.
* Cleans up and clarifies the cross process Core Data sync code. Adds a cross process sync for the `PersistentCacheController`'s context. This is only necessary when both processes are running simultaneously. If changes occur while either process is dormant, the other process will get the changes from the shared persistent store.

### Minor fixes
* 44f633e, 9d5e3e8, 99ba23c - Fixes the [language change issue](https://phabricator.wikimedia.org/T263158) by ensuring the managed object context is saved after language changes. Also adds a missing widget timeline update in `saveNewExploreFeedPreferences`
* 6d236b3 - Fixes a crash that occurred when the system was trying to encode the picture of the day timeline entry. Seems to fix the issue with flashing widgets in the widget gallery after a phone reboot. Re-tested on a clean install with previous repro steps and the widgets weren't flashing.


Wikipedia/Code/ImageController.swift
```diff
@@ -32,27 +32,6 @@ let WMFExtendedFileAttributeNameMIMEType = "org.wikimedia.MIMEType"
 open class ImageController : NSObject {
```
tonisevener, line 32: I think this class is only used in a small unit test class, so maybe we can delete it entirely.

Wikipedia/Code/MWKDataStore.m
```diff
@@ -68,84 +73,95 @@ + (MWKDataStore *)shared {
...
-        self.remoteNotificationsController = [[RemoteNotificationsController alloc] initWithSession:[WMFSession shared] configuration:[WMFConfiguration current] preferredLanguageCodesProvider:self.languageLinkController];
-        WMFArticleSummaryFetcher *fetcher = [[WMFArticleSummaryFetcher alloc] initWithSession:[WMFSession shared] configuration:[WMFConfiguration current]];
+        self.wikidataDescriptionEditingController = [[WikidataDescriptionEditingController alloc] initWithSession:session configuration:configuration];
+        self.remoteNotificationsController = [[RemoteNotificationsController alloc] initWithSession:session configuration:configuration preferredLanguageCodesProvider:self.languageLinkController];
+        WMFArticleSummaryFetcher *fetcher = [[WMFArticleSummaryFetcher alloc] initWithSession:session configuration:configuration];
         self.articleSummaryController = [[WMFArticleSummaryController alloc] initWithFetcher:fetcher dataStore:self];
         self.mobileviewConverter = [[MobileviewToMobileHTMLConverter alloc] init];
```
tonisevener, line 111: I could be missing this somewhere but I don't see the new `notificationsController` property getting instantiated anywhere in this class.

Wikipedia/Code/SavedPageSpotlightManager.swift
```diff
@@ -35,9 +35,10 @@ public extension WMFArticle {
 
         searchableItem.subject = wikidataDescription
         searchableItem.contentDescription = snippet
-        if let imageURL = imageURL(forWidth: WMFImageWidth.medium.rawValue) {
-            searchableItem.thumbnailData = ImageCacheController.shared?.data(withURL: imageURL)?.data
-        }
+        /// SINGLETONTODO
+//        if let imageURL = imageURL(forWidth: WMFImageWidth.medium.rawValue) {
```
tonisevener, line 39: I didn't dig into what this is actually doing but seems like at the very least we should pull from `MWKDataSource.shared` to assign this data right?

WMF Framework/PermanentCacheController.swift
```diff
@@ -0,0 +1,62 @@
...
+        imageCache = ImageCacheController(moc: moc, session: session, configuration: configuration)
+        articleCache = ArticleCacheController(moc: moc, imageCacheController: imageCache, session: session, configuration: configuration, preferredLanguageDelegate: preferredLanguageDelegate)
+        urlCache = PermanentlyPersistableURLCache(moc: moc)
+        managedObjectContext = moc
+        super.init()
+        session.permanentCache = self
+    }
+    
+    /// Performs any necessary migrations on the CacheController's internal storage
+    /// Exists on this @objc ImageCacheControllerWrapper so it can be accessed from WMFDataStore
```
tonisevener, line 24: We should update this comment to not reference `ImageCacheControllerWrapper`.

WMF Framework/PermanentCacheController.swift
```diff
@@ -0,0 +1,62 @@
...
+    
+    @objc public func memoryCachedImage(withURL url: URL) -> Image? {
+        return imageCache.memoryCachedImage(withURL: url)
+    }
+    
+    @objc public func fetchImage(withURL url: URL?, priority: Float, failure: @escaping (Error) -> Void, success: @escaping (ImageDownload) -> Void) -> String? {
+        return imageCache.fetchImage(withURL: url, priority: priority, failure: failure, success: success)
+    }
+    
+    @objc public func cancelFetch(withURL url: URL?, token: String?) {
```
tonisevener, line 41: Since this passthrough method only operates on the `imageCache` within, I think we should name it `cancelImageFetch` just to be more clear to outside callers.

WMF Framework/PermanentCacheController.swift
```diff
@@ -0,0 +1,62 @@
...
+    
+    @objc public func cachedImage(withURL url: URL?) -> Image? {
+        return imageCache.cachedImage(withURL: url)
+    }
+    
+    @objc public func fetchData(withURL url: URL, failure: @escaping (Error) -> Void, success: @escaping (Data, URLResponse) -> Void) {
+        imageCache.fetchData(withURL: url, failure: failure, success: success)
+    }
+    
+    @objc public func data(withURL url: URL) -> TypedImageData? {
```
tonisevener, line 53: Same here, `imageData(withURL url:URL)`

tonisevener (commented):
This is great, very much looking forward to merging this and only having 1 singleton to worry about. Here's some initial comments on the singleton-wrangling portion of this, now I'm going to review the Core Data synchronizer, Widget and test changes.

Wikipedia/Code/MWKDataStore.m
joewalsh: Great catch - it's definitely missing. I'll add it.

WMF Framework/WMFCrossProcessCoreDataSynchronizer.m
```diff
@@ -0,0 +1,161 @@
...
+    return bundleHash;
+}
+
+@implementation WMFCrossProcessCoreDataSynchronizer
+
+- (instancetype)initWithIdentifier:(NSString *)identifier storageDirectory:(NSURL *)directoryURL {
+    self = [super init];
+    if (self) {
+        _semaphore = dispatch_semaphore_create(1);
+        self.containerURL = directoryURL;
```
tonisevener, line 31: Need a `self.identifier = identifier` here I think.

tonisevener (commented):
Reviewed synchronizer portion, just 1 thing.

WMF Framework/WMFCrossProcessCoreDataSynchronizer.m
joewalsh: Good catch - fixed!

Widgets/Widgets/TopReadWidget.swift
```diff
@@ -31,56 +31,54 @@ final class TopReadData {
...
-        let moc = dataStore.viewContext
-        moc.perform {
-            guard let latest = moc.newestGroup(of: .topRead, forSiteURL: self.primaryAppLanguageSiteURL), latest.isForToday else {
-                guard !usingCache else {
-                    completion(self.placeholder)
+    func fetchLatestAvailableTopRead(usingCache: Bool = false, completion userCompletion: @escaping (TopReadEntry) -> Void) {
+        WidgetController.shared.startWidgetUpdateTask(userCompletion) { (dataStore, completion) in
+            let moc = dataStore.viewContext
+            moc.perform {
+                guard let latest = moc.newestGroup(of: .topRead), latest.isForToday else {
```
tonisevener, line 39: Do we need siteURL check here?

WikipediaUnitTests/Code/SectionEditorFindAndReplaceTests.swift
```diff
@@ -274,7 +275,8 @@ class SectionEditorFindAndReplaceTests: XCTestCase {
         webView.evaluateJavaScript("""
                 var line = editor.state.search.posFrom.line
                 var ch = editor.state.search.posFrom.ch
-            window.webkit.messageHandlers.\(messageHandlerKeyNextSearchLocation).postMessage({'\(cursorLineKey)': line, '\(cursorCharacterKey)': ch})
+                window.webkit.messageHandlers.\(messageHandlerKeyNextSearchLocation).postMessage({'\(cursorLineKey)': line, '\(cursorCharacterKey)': ch})
+                true
```
tonisevener, line 279: Are these `true`s necessary or were they just tinkering code?

tonisevener (commented):
Last of the code review comments. I still haven't run on device and tested, I'll do that later this afternoon.

WikipediaUnitTests/Code/SectionEditorFindAndReplaceTests.swift
joewalsh: Oddly enough they seem to be necessary on iOS 14 - I fount this issue while running the tests locally against iOS 14.

Widgets/Widgets/TopReadWidget.swift
joewalsh: Yup! Done

tonisevener (approved):

PR #3704, opened 2020-09-18
Widgets: If error, update in 2 hours

mcleinman:
**Fixes Phabricator ticket:** follow up from widget work

### Notes
* If a timeline update results in an error, retry in 2 hours.
* Updated files to use spaces to indent.
* @staykids - Please confirm that the `isError` variable actually is `true` in the error condition.

### Test Steps
1. This is somewhat hard to test.


Widgets/Widgets/OnThisDayWidget.swift
```diff
@@ -38,7 +38,14 @@ struct OnThisDayProvider: TimelineProvider {
     func getTimeline(in context: Context, completion: @escaping (Timeline<OnThisDayEntry>) -> Void) {
         dataStore.fetchLatestAvailableOnThisDayEntry { entry in
             let currentDate = Date()
-            let timeline = Timeline(entries: [entry], policy: .after(currentDate.dateAtMidnight() ?? currentDate))
+            let nextUpdate: Date
+            if entry.error == nil {
```
staykids, line 42: I think we can just skip the error checking and move to a 2 or 3 hour update interval for all the widgets and see how far that gets us. I can see the potential value of doing the error checking first, but I wonder if iOS decides to perform timeline updates based on our policy _and_ its own mix of policy decisions like network reachability.

I think our eventual ideal model of constructing the timeline may be something like:
- First check network reachability. If unavailable, and we have a sufficient budget of widget updates, schedule another update for after a few minutes. If we don't have a sufficient budget of widget updates, just deliver cached/placeholder data with another update scheduled for after a couple hours.
- If network is reachable, perform a fetch. 
- If fetched content matches our criteria (for the current day, not placeholder entry, no errors), present it and schedule a timeline update at midnight. Otherwise, present what's available, and schedule another update after a few minutes as the widget update budget allows. 
- Have a local cache recording repeated "after a few minutes" attempts to be sure we're not burning through our available widget update budget

But I think maybe that's a bit too much before we get a better understanding of widgets in use to see where the problems are. So for now maybe just: 
```swift
let components = DateComponents(hour: 2)
let nextUpdate = Calendar.current.date(byAdding: components, to: currentDate) ?? currentDate
let timeline = Timeline(entries: [entry], policy: .after(nextUpdate))
```
Just moving everything to this 2 or 3 hour update interval will help us cover our stale data cases as well.

What do you think?

mcleinman: I really, really like your eventual model plan.

I'm open to the idea for ignoring error checking, but I'm curious what the value is in refreshing if we're fairly confident there was no error. It seems like we don't actually know our budget of widget updates, and there is no way to query for it. (Though I could have missed that somewhere in documentation, it seemed like it was a bit of a black box, like on WatchOS.) While there is potentially no downside to spamming a refresh every 2-3 hours (if we run out of our budget, we presumably already have our latest version), I'm not sure we're guaranteed that budget resets at midnight. If they use something other than a hard midnight cut (which we can't know), we may not get a refresh at midnight if we spammed too much during the day.

staykids: The main things I think a blanket 2-3 hour refresh would afford us:

1. simpler conceptual model (no branching based on error), so as we move forward into understanding more about these widgets in production use we know they are always (as much as we can hope) getting updates on a 2-3 hour interval
2. covering more stale or insufficient data cases - for example, the absence of image data on a first pass of top read would be ok and not considered an error, but if network conditions improve/the system is allowed to retry with more frequency, the fidelity of the presented data could be improved to include the images.

If you feel strongly about your current approach, let's go for it, but if you're on the fence about it, I think the blanket 2-3 hour approach could be a good alternative to start. Up to you!

Widgets/Widgets/PictureOfTheDayWidget.swift
```diff
@@ -170,126 +170,134 @@ struct PictureOfTheDayEntry: TimelineEntry {
...
-	func getSnapshot(in context: Context, completion: @escaping (PictureOfTheDayEntry) -> Void) {
-		dataStore.fetchLatestAvailablePictureEntry(usingCache: context.isPreview) { entry in
-			completion(entry)
-		}
-	}
+    func getTimeline(in context: Context, completion: @escaping (Timeline<PictureOfTheDayEntry>) -> Void) {
+        dataStore.fetchLatestAvailablePictureEntry { entry in
+            let currentDate = Date()
+            let nextUpdate: Date
+            let isError = (entry.imageDescription == dataStore.sampleEntry.imageDescription)
```
staykids, line 191: Because the sample entry's `imageDescription` is localized (which I think might be the wrong approach for this particular widget), I think we need to consider the error state using something else. Maybe `imageURL` instead? It's kind of strange because it's like, what do we consider an error state for this particular widget? Using `image` directly might be the most "correct" approach to cover multiple "error" states.

Widgets/Widgets/OnThisDayWidget.swift
mcleinman: I think I went from neutral to favoring the "only update if problems" approach while writing my last comment. (Apologies - I think my writing starts with neutrality, but doesn't quite end there.)

The most important thing in my opinion is to ensure we don't burn through our budget and not get a requested midnight update (after the day turns). My concern about every 2-3 hours is that we don't know our budget, and can't be confident when the budget resets. So if for some reason we burn through and don't get a midnight refresh, the widget will look out of date to the users.

If there is documentation around the budget that renders my concerns moot, I'm happy to go with the every 2-3 hours approach. But if we can't be sure, I think I'd rather err on the safe side.

Widgets/Widgets/PictureOfTheDayWidget.swift
mcleinman: I'll use whatever you recommend for error states for these - Though I think we need a check on an optional or a check on a non-class (as the objects won't be identical for the images). Image is a struct, so I think this would compare just fine, right?

staykids: Sorry if I was unclear - I just meant the `UIImage` `image` directly, comparing if their data is equivalent. Conceptually here, I think error is either the failure to have an image for the entry or the entry image being the same as the sample entry's image.

```swift
entry.image == nil || entry.image == dataStore.sampleEntry.image
```

Widgets/Widgets/OnThisDayWidget.swift
staykids: Ok sounds good. While I'm pretty certain I've seen/heard some official source suggest something to the effect of having a more liberal budget than what we have here, I'm not having good luck finding that source again. So safe like you suggest sounds good. Thanks for exploring my alternative a bit.

Widgets/Widgets/PictureOfTheDayWidget.swift
mcleinman: Updated the error check. Thanks!

Widgets/Widgets/TopReadWidget.swift
```diff
@@ -125,49 +125,58 @@ final class TopReadData {
...
-			let timeline = Timeline(entries: [entry], policy: .atEnd)
-			completion(timeline)
-		}
-	}
+    func getTimeline(in context: Context, completion: @escaping (Timeline<TopReadEntry>) -> Void) {
+        dataStore.fetchLatestAvailableTopRead { entry in
+            let timeline: Timeline<TopReadEntry>
+            let isError = (entry.groupURL == nil)
+            if !isError {
+                timeline = Timeline(entries: [entry], policy: .atEnd)
```
staykids, line 168: We _could_ change this to match the after midnight policy, but let's leave it. `.atEnd` is the default policy, in use it currently feels like it has the behavior characteristics we desire, and I'm curious if in production it exposes any kind of difference in behavior from the other widgets.

staykids (approved):
💪🏽 nice - may our widgets always be loaded.

I think if we ever find more concrete information about the widget reload budget supporting at least a once an hour refresh policy, we should adjust these to the just load every 2-3 hours model to reduce complexity.

PR #3705, opened 2020-09-18
On This Day widget tweaks

mcleinman:
**Fixes Phabricator ticket:** https://phabricator.wikimedia.org/T262876

### Notes
UI changes:
- Placeholder widget is now localized
- The largest dynamic type respected by widgets (the "even larger" sizes provided in an accessibility setting do not make it larger than `extraExtraLarge`) was cutting off some text. It no longer is. (@carolynlimadeo: If instead of tightening it up, we wanted to drop the article box for these sizes - allowing a more-useful amount of event snippet text - let me know.)

Code improvements:
- Force unwraps removed (and made the articleURL optional in the process)
- Removed some placeholder text
- Use a cached version of the article image when fetching it for a preview.

### Test Steps
1. Try the widget with the largest dynamic type size.
1. Easiest way to test placeholder entry text: Replace `completion(entry)` in `OnThisDayWidget.swift` with `completion(self.placeholderEntry)`. Look at the widget in, say, English and Arabic, make sure both look appropriate. (Some text is yet to be translated in non-English languages, so you'll likely still see some English.)

### Screenshots/Videos
**Before**
Second largest size (looks fine)
Screenshot
 
Largest size (broken)
Screenshot 1

**After**
Largest size (fixed w/ slightly tighter vertical spacing in a few spots)
Screenshot 2


Widgets/Widgets/OnThisDayView.swift
```diff
@@ -29,21 +29,26 @@ struct OnThisDayColors {
 
 struct OnThisDayView: View {
     @Environment(\.widgetFamily) private var widgetSize
+    @Environment(\.sizeCategory) var textSize
     var entry: OnThisDayProvider.Entry
 
+    var doesNeedVerticalCompression: Bool {
```
staykids, line 35: What do you think about just `needsVerticalCompression` for these?

mcleinman: I'm in the habit of having booleans get an `is`/`was`/'does`/etc. prefix, but `needs` provides the same clarity I think. Good call. Updating now.

Widgets/Widgets/OnThisDayView.swift
```diff
@@ -29,21 +29,26 @@ struct OnThisDayColors {
...
+    @Environment(\.sizeCategory) var textSize
     var entry: OnThisDayProvider.Entry
 
+    var needsVerticalCompression: Bool {
+        return textSize == .extraExtraExtraLarge
+    }
+
     @ViewBuilder
     var body: some View {
         GeometryReader { proxy in
```
staykids, line 41: Not worth messing with the delicate magical chemistry of SwiftUI for this now, but I think this `GeometryReader` is superfluous. I think you could get away with writing this kinda like:

```swift
@ViewBuilder
var body: some View {
	widgetContent
		.overlay(errorBox)
		.environment(\.layoutDirection, entry.isRTLLanguage ? .rightToLeft : .leftToRight)
		.widgetURL(entry.contentURL)
}

@ViewBuilder
var widgetContent: some View {
	switch widgetSize {
		...
	}
}

```

staykids (approved):
👍 works `<h1><h2><h3>great</h1></h2></h3>`!

Widgets/Widgets/OnThisDayView.swift
mcleinman: 100% in agreement. Good catch on this. I think there were a million readers at one point when I was trying to get the dot spacing working, and apparently I missed one in the removals.

PR #3707, opened 2020-09-21
Maybe fix AppViewController crash

tonisevener:

### Notes
* I could be talked out of this one - it's been more of a warmup task for me this morning. This is the highest non-widget crash in 6.7.0. It fails in `WMFAppViewController`'s `viewDidLoad`, opening in project points to the line `self.navigationItem.backButtonDisplayMode = UINavigationItemBackButtonDisplayModeGeneric;` with a `NSObject doesNotRecognizeSelector` failure in the stack trace. I can't reproduce it, but I noticed all of the crash samples show an iOS14 build number of either beta 1 or beta2, so my guess is that this was an Apple-side bug that has since been fixed, and these crashes will eventually go away on their own.

### Test Steps
1. Add a breakpoint at the `backButtonDisplayMode` assignment in `AppViewController` and in one of the Swift usages (I did `ArticleViewController`), run on iOS12 or iOS13. Confirm breakpoints are not hit.
2. Do the same for iOS14. Confirm breakpoints are still hit.


mcleinman:
If this is an Apple-side beta bug, do we want to add code to check for this? Unless it's going to be rushed out in a hotfix, by the time it's in users hands I'd expect there to be very very few iOS 14 beta 1/2 users out there.

tonisevener:
> That said, if this is an Apple-side beta bug, do we want to add code to check for this?

Yeah, I'm leaning towards no, but I'll leave this up a little while longer before closing this PR. I am surprised that the crash count is as high as it is given that it's from beta 1 or 2, but maybe most of those are the same few users repeatedly hitting it, since I imagine it acts like a crash on launch.

tonisevener:
Closing this one out, it seems to be steadily decreasing in 6.7.0.

PR #3713, opened 2020-09-24
Fix CocoaLumberjack debug errors

tonisevener:

### Notes
This hopefully fixes the debugging errors we've been seeing.
- Carthage was no longer updating, changed the `carthage_update` script to include [this suggested solution](https://github.com/Carthage/Carthage/issues/3019#issuecomment-665136323).
- Removed `FBSnapshotTestCase` & associated tests. The tests were already disabled because of Xcode 11 beta issues last year, now there's Xcode 12 issues.
- Removed Cocoalumberjack from Carthage and added with Swift Package Manager. Worth noting that I had to reference it's `master` branch rather than an official release number. I was getting build warnings otherwise. ([Here's](https://github.com/CocoaLumberjack/CocoaLumberjack/issues/1160#issuecomment-674067839) some documentation suggesting to use master)
- WIP for some further testing

### Test Steps
1. Run on device and simulator, confirm you still see logging in the console.
2. Confirm log looks right when submitting a bug.
3. Confirm you can debug and print various things out to the console (test Swift & Obj-C, both app-side and in a WMF file).


tonisevener:
This is ready for review.

mcleinman (approved):
This works wonderfully. Thank you so much!

PR #3714, opened 2020-09-24
Widget 0xdead10cc fixes part III: it was the old widgets all along

joewalsh:
**Fixes Phabricator ticket:** 
https://phabricator.wikimedia.org/T263247

### Notes
* Updates old widgets to use new widget fetching method
* Fixes retain cycles in image fetching exposed by image fetching via the `UIImageView` extension in the old widgets

### Test Steps
1. Simulator > Erase all content and settings
2. Restart Xcode
3. Set a breakpoint on [MWKDataStore shared]
4. Run the old widget targets
5. Ensure the breakpoint isn't hit
6. Inspect the memory graph
7. Ensure there are no data stores or `NSManagedObjectContext`s remaining

PR #3716, opened 2020-09-24
Request picture of the day image closer to display size

joewalsh:
**Fixes Phabricator ticket:** 
https://phabricator.wikimedia.org/T263278

### Notes
* Update branch after #3714 is merged

### Test Steps
1. Open POTD on a day there's a panoramic image
2. Verify the image requested is the closest thumbnail size (one of the WMFImageWidth buckets) we can get to a 1:1 image pixel:display pixel

### Screenshots
Before:
IMG_1865

After:
IMG_1866

### DIFF
--- a/Wikipedia/Code/WMFFeedImage.m
+++ b/Wikipedia/Code/WMFFeedImage.m
@@ -3,13 +3,14 @@
 #import <WMF/WMFImageURLParsing.h>
 #import <WMF/NSURL+WMFExtras.h>
 #import <WMF/MWLanguageInfo.h>
+#import <WMF/UIScreen+WMFImageWidth.h>
 
 NS_ASSUME_NONNULL_BEGIN
 
 @implementation WMFFeedImage
 
 + (NSUInteger)modelVersion {
-    return 2;
+    return 3;
 }
 
 + (NSDictionary *)JSONKeyPathsByPropertyKey {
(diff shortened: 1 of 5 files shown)


Wikipedia/Code/WMFFeedImage.m
```diff
@@ -26,8 +29,8 @@ + (NSValueTransformer *)imageThumbURLJSONTransformer {
                                               BOOL *success,
                                               NSError *__autoreleasing *error) {
             NSInteger sizePrefix = WMFParseSizePrefixFromSourceURL(urlString);
-            if (sizePrefix < 640) {
-                urlString = WMFChangeImageSourceURLSizePrefix(urlString, 640);
+            if (sizePrefix < WMFImageWidthExtraLarge) {
+                urlString = WMFChangeImageSourceURLSizePrefix(urlString, WMFImageWidthExtraLarge);
```
staykids, line 33: Is the intent here just to have a more sensible default given the prevalence of 2x scale and above devices?

staykids (approved):
I'm slightly worried about the potential risk this introduces in regards to POTD widgets crashing (as we know there's a WidgetKit memory cap, but we're not precisely sure what it is or what the requested image's ultimate file size will be), but it's also nice to not see blurry images in the widget size variants.

Wikipedia/Code/WMFFeedImage.m
joewalsh: Yup! And to use a defined constant so if those thumbnail widths ever change it'd be updated here

PR #3717, opened 2020-09-25
Follow-on widget fixes

joewalsh:
* Added a dimension limit to images requested for the POTD widget
* Updated the widgets to use the background fetch for updating the feed. The background fetch updates fewer content sources and utilizes the main context instead of spinning up a separate import context. Consolidate logic that fetches the latest `WMFContentGroup` from the cache and updates the feed if it isn't found.
* Move the responsibility of tearing down the `MWKDataStore` and waiting til tasks finish to the data store itself


Widgets/Widgets/PictureOfTheDayWidget.swift
```diff
@@ -33,37 +33,20 @@ final class PictureOfTheDayData {
     // MARK: Public
 
     func fetchLatestAvailablePictureEntry(for imageSize: CGSize, usingCache: Bool = false, completion userCompletion: @escaping (PictureOfTheDayEntry) -> Void) {
-        WidgetController.shared.startWidgetUpdateTask(userCompletion) { (dataStore, completion) in
-            let moc = dataStore.viewContext
-            moc.perform {
-                guard let latest = moc.newestGroup(of: .pictureOfTheDay), latest.isForToday else {
```
tonisevener, line 39: @joewalsh not sure if this will affect anything user-facing but the old POTD code did not consider `siteURL` when fetching `newestGroup`, and the new shared code does.

TopReadWidget/WMFTodayTopReadWidgetViewController.swift
```diff
@@ -348,18 +348,7 @@ class WMFTodayTopReadWidgetViewController: ExtensionViewController, NCWidgetProv
     }
 
     func fetch(dataStore: MWKDataStore, completionHandler: @escaping ((NCUpdateResult) -> Void)) {
-        updateUIWithTopRead(with: dataStore, moc: dataStore.viewContext) { result in
-            guard result == .failed else {
-                completionHandler(result)
-                return
-            }
-            dataStore.feedContentController.updateFeedSourcesUserInitiated(false) {
```
tonisevener, line 356: I haven't tested to confirm but seems like there's a chance this could miss a daily update by stripping this network update. Did you purposefully not lean on your new `fetchNewestWidgetContentGroup(.topRead, in: dataStore, isNetworkFetchAllowed: true)` method here?

Widgets/Widgets/PictureOfTheDayWidget.swift
joewalsh: Good catch! It looks like the POTD content group has siteURL set but it's unclear if it'll always match the user's primary language - I've updated it to no longer constraint to siteURL for POTD

TopReadWidget/WMFTodayTopReadWidgetViewController.swift
joewalsh: Yeah, thanks for pointing this out, I should have mentioned this in the notes. The other legacy widgets don't perform a network fetch but instead rely on background updates for the app. I thought it might make more sense to be consistent with the other widgets, but it's probably better to stay consistent with the previous behavior of this widget. I restored the fetch using `fetchNewestWidgetContentGroup`.

tonisevener (approved):

PR #3721, opened 2020-09-30
Article as a Living Document - Article Content Insert piece

tonisevener:
**Fixes Phabricator ticket:** 

Epic with original mocks: https://phabricator.wikimedia.org/T241253
Subticket with additional business logic in comment discussions: https://phabricator.wikimedia.org/T257595

### Remaining Tasks

Nothing major, shouldn't block overall code review at this point.

- [x] Uncomment unit tests and fix
- [x] Add bot icon for bots
- [x] Fix small bug where some articles are slightly horizontally scrollable
- [x] Fix bug with new changes badge logic being overwritten

### Test Steps
1. You should be able to see the article content insert on LTR devices in any EN Wikipedia article for iOS13+


WMF Framework/Significant Events Endpoint/SignificantEventsModels.swift
```diff
@@ -0,0 +1,616 @@
...
+            case .large:
+                if let event = Event.Large(untypedEvent: untypedEvent) {
+                    typedEvents.append(.large(event))
+                }
+            case .vandalismRevert:
+                if let event = Event.VandalismRevert(untypedEvent: untypedEvent) {
+                    typedEvents.append(.vandalismRevert(event))
+                }
+            case .newTalkPageTopic:
+                if let event = Event.NewTalkPageTopic(untypedEvent: untypedEvent) {
```
staykids, line 58: The test article I've typically been primarily using is COVID-19 Pandemic (https://mobileapps-ios-experiments.wmflabs.org/en.wikipedia.org/v1/page/significant-events/COVID-19_pandemic). Not sure if it's related to any endpoint changes made on Friday, but it looks like there's a parsing issue here that prevents the significant changes for that page from displaying.

1. The `if let event = Event.NewTalkPageTopic(untypedEvent: untypedEvent)` here produces no value because `NewTalkPageTopic`'s initializer https://github.com/wikimedia/wikipedia-ios/blob/917c7335e957200109033f91a35be349b0024ca8/WMF%20Framework/Significant%20Events%20Endpoint/SignificantEventsModels.swift#L232 expects `snippet` to be non-nil. Take a look at the data below to see what was causing the failure for me.

2. This causes the `guard typedEvents.count == untypedEvents.count` statement to fail with false and throws the `unableToParseIntoTypedEvents` error.


```
▿ WMF.SignificantEvents.UntypedEvent
  - outputType: WMF.SignificantEvents.EventOutputType.newTalkPageTopic
  ▿ revId: Optional(981925194)
	- some: 981925194
  ▿ timestampString: Optional("2020-10-05T05:57:47Z")
	- some: "2020-10-05T05:57:47Z"
  ▿ user: Optional("Investigatory")
	- some: "Investigatory"
  ▿ userId: Optional(40140363)
	- some: 40140363
  ▿ userGroups: Optional(["*", "user", "autoconfirmed"])
	▿ some: 3 elements
	  - "*"
	  - "user"
	  - "autoconfirmed"
  ▿ userEditCount: Optional(297)
	- some: 297
  - count: nil
  - sections: nil
  ▿ section: Optional("== RFC to update and confirm current standing of transmission in lead ==")
	- some: "== RFC to update and confirm current standing of transmission in lead =="
  - snippet: nil
  - untypedChanges: nil
```

tonisevener: @staykids thanks for the heads up! yeah it seems to me like talk page snippets *shouldn't* return null, but I'll dig into the endpoint to see if it's justified in this case. If so we may have to flag the snippet as optional and hide it in the cell.

tonisevener: @staykids this should be fixed in the endpoint now.

staykids: 💪 working great now

staykids: Hmm seems like this popped up for me again on a different article: https://mobileapps-ios-experiments.wmflabs.org/en.wikipedia.org/v1/page/significant-events/Proud_Boys

In addition to the `snippet` being nil, this one's `section` is also nil - both with non-nil requirements in the `NewTalkPageTopic` initializer.

```
▿ WMF.SignificantEvents.UntypedEvent
  - outputType: WMF.SignificantEvents.EventOutputType.newTalkPageTopic
  ▿ revId: Optional(982152939)
    - some: 982152939
  ▿ timestampString: Optional("2020-10-06T13:18:38Z")
    - some: "2020-10-06T13:18:38Z"
  ▿ user: Optional("ItsPugle")
    - some: "ItsPugle"
  ▿ userId: Optional(30023517)
    - some: 30023517
  ▿ userGroups: Optional(["extendedconfirmed", "*", "user", "autoconfirmed"])
    ▿ some: 4 elements
      - "extendedconfirmed"
      - "*"
      - "user"
      - "autoconfirmed"
  ▿ userEditCount: Optional(3201)
    - some: 3201
  - count: nil
  - sections: nil
  - section: nil
  - snippet: nil
  - untypedChanges: nil
```

tonisevener: @staykids thanks for the heads up, I'll look at this next

tonisevener: @staykids this should be fixed up better in the endpoint now, it was a side-effect of my previous fix. Let me know if you see any other issues.

staykids: Nice – looks like an `UntypedEvent` with that revision ID is now being classified as a small change and no longer going through the `NewTalkPageTopic` path. Also just FYI, I'll stick to using test articles from Shay's spreadsheet from here on in – just so happened this article was already on there.

WMF Framework/Significant Events Endpoint/SignificantEventsModels.swift
```diff
@@ -0,0 +1,616 @@
...
+                var typedTemplates: [Template] = []
+                self.outputType = untypedChange.outputType
+                self.sections = untypedChange.sections
+                self.untypedTemplates = untypedTemplates
+                
+                for untypedTemplate in untypedTemplates {
+                    guard let name = untypedTemplate["name"] else {
+                        continue
+                    }
+                    let lowercaseName = name.lowercased()
```
staykids, line 313: Is there a functional reason we're lowercasing `name` that I'm missing? If not, might be worth using one of these to skip the indirection:

```swift
if name.localizedCaseInsensitiveContains("cite")  { ... }
if name.caseInsensitiveCompare("cite") == .orderedSame { ... }
```

WMF Framework/Significant Events Endpoint/SignificantEventsModels.swift
```diff
@@ -0,0 +1,616 @@
...
+                    if lowercaseName.contains("cite") {
+                        if lowercaseName.contains("book") {
+                            if let bookCitation = Citation.Book(dict: untypedTemplate) {
```
staykids, line 316: Since these don't have else cases, could simplify and reduce the nesting by pulling the second `if let`s up to the parent in each of these: 

```swift
if lowercaseName.contains("book"), let bookCitation = Citation.Book(dict: untypedTemplate) {
    typedTemplates.append(.bookCitation(bookCitation))
}
```

WMF Framework/Significant Events Endpoint/SignificantEventsModels.swift
```diff
@@ -0,0 +1,616 @@
...
+                self.publisher = dict.nonEmptyValueForKey(key: "publisher") ??
+                                dict.nonEmptyValueForKey(key: "distributor") ??
+                                dict.nonEmptyValueForKey(key: "institution") ??
+                                dict.nonEmptyValueForKey(key: "newsgroup")
+                
+                self.pagesCited = dict.nonEmptyValueForKey(key: "pages") ??
+                    dict.nonEmptyValueForKey(key: "pp")
+                
+                self.isbn = dict.nonEmptyValueForKey(key: "isbn") ??
+                            dict.nonEmptyValueForKey(key: "ISBN13") ??
```
staykids, line 413: This is another opportunity where case insensitive `String` operations might reduce code. You could probably even adapt your `Dictionary` helper to support that pretty easily - maybe something like:

```swift

dict.nonEmptyValueForKey(key: "isbn13", caseInsensitive: true)

...

private extension Dictionary where Key == String, Value == String {
    func nonEmptyValueForKey(key: String, caseInsensitive: Bool = false) -> String? {
        guard let key = caseInsensitive 
                ? keys.first(where: {$0.caseInsensitiveCompare(key) == .orderedSame})
                : key else { 
                    return nil		
         }

        if let value = self[key], !value.isEmpty {
            return value
        }

        return nil
    }
}
```

WMF Framework/Significant Events Endpoint/SignificantEventsModels.swift
```diff
@@ -0,0 +1,616 @@
...
+                dict.nonEmptyValueForKey(key: "author") ??
+                dict.nonEmptyValueForKey(key: "author1") ??
+                dict.nonEmptyValueForKey(key: "authors") ??
+                dict.nonEmptyValueForKey(key: "last1")
+                
+                self.firstName = dict.nonEmptyValueForKey(key: "first") ??
+                dict.nonEmptyValueForKey(key: "first1")
+                
+                self.sourceDateString = dict.nonEmptyValueForKey(key: "date")
+                self.urlString = dict.nonEmptyValueForKey(key: "url")
```
staykids, line 450: I think the doc states this can also be an uppercased `URL`: 

```swift
self.urlString = dict.nonEmptyValueForKey(key: "url") ?? dict.nonEmptyValueForKey(key: "URL") // or something like the `caseInsensitive` approach from above 😉
```

WMF Framework/Significant Events Endpoint/SignificantEventsModels.swift
```diff
@@ -0,0 +1,616 @@
...
+                
+                self.sourceDateString = dict.nonEmptyValueForKey(key: "date")
+                self.publication = dict.nonEmptyValueForKey(key: "work") ??
+                                    dict.nonEmptyValueForKey(key: "journal") ??
+                                    dict.nonEmptyValueForKey(key: "magazine") ??
+                                    dict.nonEmptyValueForKey(key: "periodical") ??
+                                    dict.nonEmptyValueForKey(key: "newspaper") ??
+                                    dict.nonEmptyValueForKey(key: "website")
+                
+                self.urlString = dict.nonEmptyValueForKey(key: "url")
```
staykids, line 490: Same comment re: uppercased `URL` being a possibility.

WMF Framework/Significant Events Endpoint/SignificantEventsModels.swift
```diff
@@ -0,0 +1,616 @@
...
+            public let urlString: String
+            public let title: String
+            public let publisher: String?
+            public let accessDateString: String?
+            public let archiveDateString: String?
+            public let archiveDotOrgUrlString: String?
+            
+            init?(dict: [String: String]) {
+                guard let title = dict.nonEmptyValueForKey(key: "title"),
+                      let urlString = dict.nonEmptyValueForKey(key: "url") else {
```
staykids, line 507: Same comment re: uppercased URL being a possibility.

staykids (commented):
Still working through all this, and like I mentioned haven't surfaced anything major yet. But wanted to go ahead and mention these small minor notes in case you think they are worth doing at all.

WMF Framework/CommonStrings.swift
```diff
@@ -231,6 +231,91 @@ public class CommonStrings: NSObject {
...
+    
+    static let multipleChangesMadeDescription = WMFLocalizedString("aaald-multiple-changes-description", value: "Multiple changes made", comment: "Title displayed in article as a living document content insert explaining that multiple changes were made in a revision.")
+    
+    static let twoDescriptionsFormat = WMFLocalizedString("aaald-two-descriptions-format", value: "%1$@ and %2$@", comment: "Format for two change types to insert into a revision's event description in an article as a living document timeline cell. %1$@ is replaced by the first change type and %2$@ is replaced by the second change type, e.g. '612 characters added and 323 characters removed'")
+    
+    static let finalDelimiter = WMFLocalizedString("aaald-multiple-descriptions-last-delimiter", value: " and ", comment: "Text to show as the last delimiter in a list of multiple event changes. These changes are shown in the description area of an article as a living document timeline cell. e.g. '3 references added, 612 characters added and 100 characters removed'")
```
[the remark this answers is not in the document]
tonisevener: Great catch! I didn't realize this class existed. This has been updated.

WMF Framework/DateFormatter+WikipediaLanguage.swift
```diff
@@ -2,6 +2,16 @@ import Foundation
 
 extension DateFormatter {
     // Returns year string - i.e. '1000' or '200 BC'. (Negative years are 'BC')
+    
+    //tonitodo: consider combining with NSDateFormatter WMFExtensions implementation (either via bridging or convert all intances to this swift usage)
+    public static var wmf_iso8601Formatter: DateFormatter = {
+        let dateFormatter = DateFormatter()
```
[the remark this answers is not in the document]
tonisevener: Yeah, I agree that makes sense, I based this off of the similarly-named formatter in `NSDateFormatter+WMFExtensions.m` so I'm not sure why it wasn't set up that way to begin with. After further digging it looks like this formatter isn't used anymore in the feature (I checked both PRs), so I just removed it altogether.

WMF Framework/Significant Events Endpoint/SignificantEventsModels.swift
```diff
@@ -0,0 +1,616 @@
...
+                    typedEvents.append(.vandalismRevert(event))
+                }
+            case .newTalkPageTopic:
+                if let event = Event.NewTalkPageTopic(untypedEvent: untypedEvent) {
+                    typedEvents.append(.newTalkPageTopic(event))
+                }
+            }
+        }
+
+        guard typedEvents.count == untypedEvents.count else {
```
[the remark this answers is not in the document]
tonisevener: Yeah I agree, it's too rigid to fail if one event isn't typed. I changed this to error out against a case where there are no typed events at this point whatsoever, which is probably pretty rare but I wanted to avoid any empty events state that might happen at the UI level.

WMF Framework/Significant Events Endpoint/SignificantEventsModels.swift
tonisevener: Nope no reason, this is fixed now with `localizedCaseInsensitiveContains`. The names can look something like

```
"name": "cite web ",
``` 

in the endpoint, so I still needed to keep the contains logic.

WMF Framework/Significant Events Endpoint/SignificantEventsModels.swift
tonisevener: 👍  Done

WMF Framework/Significant Events Endpoint/SignificantEventsModels.swift
tonisevener: 👍  Done.

WMF Framework/Significant Events Endpoint/SignificantEventsModels.swift
tonisevener: 👍  Done.

WMF Framework/Significant Events Endpoint/SignificantEventsModels.swift
tonisevener: 👍  Done.

WMF Framework/Significant Events Endpoint/SignificantEventsModels.swift
tonisevener: Good catch, this is done.

WMF Framework/CommonStrings.swift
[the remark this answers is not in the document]
staykids: Nice! I'm glad `ListFormatter` worked out. I could be missing something obvious, but I think you can actually use it to simplify this entire `switch` statement here: https://github.com/wikimedia/wikipedia-ios/blob/42a05c48de14357165374f8286d9e7bcbb942933/WMF%20Framework/Significant%20Events%20Endpoint/ArticleAsLivingDocViewModels.swift#L568-L586

to just something like:
```swift
let mergedDescription = ListFormatter.localizedString(byJoining: sortedDescriptions.map { $0.text })
return mergedDescription.isEmpty ? nil : mergedDescription
```

WMF Framework/Significant Events Endpoint/SignificantEventsModels.swift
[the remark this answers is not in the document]
staykids: Cool, that sounds like a good failure case here.

WMF Framework/CommonStrings.swift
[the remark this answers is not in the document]
tonisevener: Doh! 🤦‍♀️  You're right, this cleanup is done and it tested well against the other PR. Thanks for the careful eye.

staykids (approved):
Wow nice job Toni – this and the endpoint itself were clearly quite an undertaking!

Some super minor comments included, but I'm encouraging you to ignore them if they are uninteresting because they don't affect functionality and I'm providing feedback a day later than you were hoping to receive it. Feel free to merge whenever you're ready.

Wikipedia/Code/Article as a Living Document/ArticleAsLivingDocController.swift
```diff
@@ -0,0 +1,470 @@
...
+}
+
+@available(iOS 13.0, *)
+class ArticleAsLivingDocController: NSObject {
+    
+    enum Errors: Error {
+        case viewModelInstantiationFailure
+    }
+    
+    private weak var delegate: (ArticleAsLivingDocControllerDelegate & UIViewController & HintPresenting)? //will need ArticleAsLivingDocViewControllerDelegate
```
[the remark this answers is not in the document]
tonisevener: Good idea, this is done.

Wikipedia/Code/Article as a Living Document/ArticleAsLivingDocController.swift
```diff
@@ -0,0 +1,470 @@
...
+    
+    var _articleAsLivingDocViewModel: ArticleAsLivingDocViewModel?
```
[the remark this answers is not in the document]
tonisevener: Yeah it is a bit messy here, agreed. I'm going to leave it as-is for now since it feels a little more impactful, but I'll make a note to revisit with the modal code in place (so it's easier to test against).

Wikipedia/Code/Article as a Living Document/ArticleAsLivingDocController.swift
```diff
@@ -0,0 +1,470 @@
...
+    var shouldAttemptToShowArticleAsLivingDoc: Bool {
+        
+        guard let delegate = delegate,
+              let view = delegate.view else {
+            return false
+        }
+        
+        //todo: need A/B test logic (falls in test and visiting article in allowed list)
+        let isDeviceRTL = view.effectiveUserInterfaceLayoutDirection == .rightToLeft
+        let isENWikipediaArticle: Bool
```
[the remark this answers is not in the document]
tonisevener: 👍 This is done.

Wikipedia/Code/Article as a Living Document/ArticleAsLivingDocController.swift
```diff
@@ -0,0 +1,470 @@
...
+        if let host = delegate.articleURL.host,
+           host == Configuration.Domain.englishWikipedia {
+            isENWikipediaArticle = true
+        } else {
+            isENWikipediaArticle = false
+        }
+        
+        let shouldAttemptToShowArticleAsLivingDoc: Bool
```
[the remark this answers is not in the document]
tonisevener: 👍   This is done.

Wikipedia/Code/Article as a Living Document/ArticleAsLivingDocController.swift
```diff
@@ -0,0 +1,470 @@
...
+        self.delegate = delegate
+    }
+    
+    func articleTitleAndSiteURL() -> (title: String, siteURL: URL)? {
+        
+        guard let delegate = delegate else {
+            return nil
+        }
+        
+        if let title = delegate.articleURL.wmf_title?.denormalizedPageTitle,
```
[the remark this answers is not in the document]
tonisevener: 👍   This is done.

Wikipedia/Code/Session.swift
```diff
@@ -264,6 +264,18 @@ public class Session: NSObject {
     
     public func dataTask(with request: URLRequest, callback: Callback) -> URLSessionTask? {
         
+        //odd workaround to show an article as living doc bot icon in the article content web view.
```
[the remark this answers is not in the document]
tonisevener: Yeah, I wasn't sure how to pull an icon from the asset catalog and place into a web view's content, but maybe I'm overcomplicating it. Our SchemeHandler intercepts any article web view request (including images) and asks Session for the data, so this my backwards way of getting it to work.

WikipediaUnitTests/Code/Significant Events Tests/ArticleAsLivingDocViewModelTests.swift
```diff
@@ -0,0 +1,138 @@
...
+                    switch firstEvent {
+                    case .small(let smallEvent):
+                        XCTAssertNil(smallEvent.eventDescription)
+
+                        let traitCollection = UITraitCollection(preferredContentSizeCategory: UIContentSizeCategory.large)
+                        let attributedText = smallEvent.eventDescriptionForTraitCollection(traitCollection, theme: Theme.light)
+                        var attributes = attributedText.attributes(at: 0, effectiveRange: nil)
+                        var font = attributes[NSAttributedString.Key.font] as! UIFont
+                        var color = attributes[NSAttributedString.Key.foregroundColor] as! UIColor
+                        XCTAssertEqual(font.pointSize, 15.0)
```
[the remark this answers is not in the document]
tonisevener: Yeah agreed, directly comparing point sizes is probably going a bit overboard in unit tests. Maybe we could be a little looser on it, like just confirm that the point size is larger than it was in the last check, and confirming the font color is different than in the last check. I mainly just wanted to test that regenerating an event description with a different trait collection and theme changes the attributed string as expected.

tonisevener:
Going to go ahead and merge with the simpler PR feedback fixes. Thanks for reviewing this @staykids !

PR #3722, opened 2020-10-01
[To Discuss] Remove PR creation code from GitHub action

Abijeet:
The PR creation will now be taken care by the translatewiki
bot automatically.

Merge after https://gerrit.wikimedia.org/r/c/translatewiki/+/631413

The `GITHUB_TOKEN` from secrets can be removed if its not being
used elsewhere.

Tag: @translatewiki 

Bug: [T264003](https://phabricator.wikimedia.org/T264003)

PR #3728, opened 2020-10-05
Github action for PR being created by Translatewiki

mcleinman:

### Notes
* Along with [this patch](https://gerrit.wikimedia.org/r/c/translatewiki/+/631413), this will be the new way we handle translations. Basically it's the same as the existing github action, but now we're not creating the PR - we're responding to a PR's creation
* Once this runs and we confirm it's working, I'll put up another PR to remove the pre-existing GitHub action. For now, we'll see two PRs every time translations come in - and hopefully they'll be identical.

### Test Steps
1. I'm not aware of any good test environment for Github Actions beyond ones that take a lot of local work. (Please let me know if there is a good test flow.) Since I believe there is little risk to this script, I'd propose merging as is and when translation strings are updated, we'll compare the two created PRs to ensure this flow is working as expected.


mcleinman:
Between https://github.com/wikimedia/wikipedia-ios/pull/3728 and https://github.com/wikimedia/wikipedia-ios/pull/3722, only need one of the two to be merged. The first one allows both to run for a bit to confirm the new way works. Second one is a knife cut, though this is a very low risk change to how we process incoming translations.

.github/workflows/localization-update.yml
```diff
@@ -0,0 +1,29 @@
...
+#
+# As of late 2020, TranslateWiki creates PRs automatically, in line with
+# how they work with other repos. However, the strings still need to be
+# translated to work well with the iOS app. This script runs when a new PR
+# is created by TranslateWiki, to add everything the app needs to the PR.
+#
+name: Update localizations from TranslateWiki
+
+on:
+  pull_request:
```
joewalsh, line 11: I'd suggest running this on `push` instead of `pull_request` to handle cases where the localization script adds new localizations after the PR is already opened

joewalsh (approved):
I think this should be merged to get the documentation changes in, but I'd say take the configuration change #3722 instead of making a separate config. If they're both running at the same time, it's undetermined if the old or new will run first.

PR #3732, opened 2020-10-07
Update midnight widget refresh extension to use random time after midnight

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T264881

### Notes
To possibly help mitigate spikes of traffic on the hour, randomize the requested refresh time of Picture of the day and On this day from midnight local time to some random minute and second within an hour after midnight local time.

### DIFF
--- a/Widgets/Utilities/Date+Extensions.swift
+++ b/Widgets/Utilities/Date+Extensions.swift
@@ -2,11 +2,12 @@ import Foundation
 
 extension Date {
 
-	// TODO: Document nil Date
-    func dateAtMidnight(calendar: Calendar = .current) -> Date? {
-		let components = DateComponents(day: 1)
-		let startOfDay = calendar.startOfDay(for: Date())
+    /// Return a randomized `Date` in the next midnight hour. Will return nil if a `Date` can't be constructed given the calendar
+    /// and component constraints, though in practice will likely always return some value.
+    func randomDateShortlyAfterMidnight(calendar: Calendar = .current) -> Date? {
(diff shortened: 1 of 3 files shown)


mcleinman:
Do we also need to update Picture of the Day?

staykids:
I think Picture of the day is covered here – did you mean Top read? If so, I didn't touch the logic there because it doesn't explicitly use the midnight path and I'm not sure it's worth changing the "working" logic there before we understand if this approach is sufficient to mitigate the spikes in production or not.

tonisevener:
> I think Picture of the day is covered here – did you mean Top read? If so, I didn't touch the logic there because it doesn't explicitly use the midnight path and I'm not sure it's worth changing the "working" logic there before we understand if this approach is sufficient to mitigate the spikes in production or not.

@staykids Interesting - I hadn't realized it didn't use the midnight path. Do you know how it's currently working now? It seems to update around midnight anyways for me.

mcleinman:
> I think Picture of the day is covered here – did you mean Top read? If so, I didn't touch the logic there because it doesn't explicitly use the midnight path and I'm not sure it's worth changing the "working" logic there before we understand if this approach is sufficient to mitigate the spikes in production or not.

Guilty as charged. I saw two of three, and didn't pause long enough when writing that. Top Read looks up at the end of each day, right? (I'm a tad confused, but not important for now - it uses an `.atEnd` policy, but seems to set a date of now.)

mcleinman (approved):
Looks great, thanks!

PR #3738, opened 2020-10-17
Fix cross process notifier regression, fix missing explore feed cards

joewalsh:
### Fixes Phabricator Ticket
https://phabricator.wikimedia.org/T265793

### Notes
When the explore feed content is updated by the widgets, the app can get in a state where the explore feed is missing content and won't be updated until force-quitting and reopening the app. I've seen this with 6.7.2 on an iPhone X, iPhone 6S, and iPad Pro all running 14.0.1 with all three widgets installed. I caused this regression when updating the Core Data cross process sync to work with the cache database in #3703. The unique channel name that needs to be the same in the app and the extension were no longer the same due to a bug in the way the keys were generated and persisted. The main context's channel name would be re-generated on every launch because only the cache context channel name was persisted.

### Test Steps
Here's how I was able to repro on the simulator:

1. Set the computer to a date in the past
2. Erase all content and settings on the simulator you want to use
3. Build & run the WidgetsExtension target
4. View the widget
5. Open the app
6. Advance the computer's date by a day
7. View the widget
8. Open the app

[Before](https://vimeo.com/467690912)
[After](https://vimeo.com/467690944)


tonisevener (approved):
Working well for me, thanks for fixing this!

PR #3742, opened 2020-10-21
Test all AAALD navigations as pushes alternative

tonisevener:
Fixes Phabricator tickets: 
[https://phabricator.wikimedia.org/T265674](https://phabricator.wikimedia.org/T265674)
[https://phabricator.wikimedia.org/T266833](https://phabricator.wikimedia.org/T266833)

**Test Steps**
- View changes, View discussion, Username, and Small event taps in the modal should now push on from the right rather than dismissing modal > push on from right on top of article.
- View changes and small event taps should go directly to the diff view, pushing on from the right. If small events are > 1, we show the diff as the compare (orange) view across all of those revisions, not the single revision view.
- For any snippet links that previously pushed on an article view, instead show the same article preview window as we do in the Edit Preview screen.


WMF Framework/Significant Events Endpoint/ArticleAsLivingDocViewModels.swift
```diff
@@ -427,25 +428,21 @@ public extension ArticleAsLivingDocViewModel {
...
+                    parentId = vandalismRevert.parentId
                 default:
                     assertionFailure("Shouldn't happen")
+                    revId = 0
+                    parentId = 0
                 }
                 
             }
             
-            public convenience init?(forPrototypeText prototypeText: String) {
```
tonisevener, line 440: Cleanup - this method wasn't used anywhere.

WMF Framework/Significant Events Endpoint/SignificantEventsModels.swift
```diff
@@ -558,22 +572,6 @@ public extension SignificantEvents {
                 case snippet
                 case untypedChanges = "significantChanges"
             }
-        
-        init(forPrototypeText prototypeText: String) {
```
tonisevener, line 562: This also wasn't used anywhere.

mcleinman:
This code looks good to me, and tests well. However, I just put up a final commit to remove duplicated code (per an earlier comment I left), so @staykids should do the final approval/merge of this one.

staykids:
> This code looks good to me, and tests well. However, I just put up a final commit to remove duplicated code (per an earlier comment I left), so @staykids should do the final approval/merge of this one.

👍 looking at it now

Wikipedia/Code/InternalLinkPreviewing.swift
```diff
@@ -0,0 +1,32 @@
...
+            exists = !query.contains("redlink=1")
+        } else {
+            exists = true
+        }
+        if !exists {
+            showRedLinkInAlert()
+            return
+        }
+        let dataStore = MWKDataStore.shared()
+        let internalLinkViewController = EditPreviewInternalLinkViewController(articleURL: url, dataStore: dataStore)
```
staykids, line 18: If the article link doesn't exist (like this screenshot, to "https://en.wikipedia.org/wiki/List_of_Republicans_who_have_denounced_Donald_Trump" on "2020 United States presidental election"), the preview hangs indefinitely.

loading

staykids (approved):
Nice – the new navigation pattern definitely helps. Two questions below that I don't immediately have solutions for. We can just merge if they are beyond the scope of this PR.

Wikipedia/Code/InternalLinkPreviewing.swift
mcleinman: This one probably should be filed separately. There are two failures here: 
1) We're looking for `redlink=1` in the url's query to decide whether to show this as a "no actual article" alert or this mini-article view. That is not present in [the data returned from the server](https://mobileapps-ios-experiments.wmflabs.org/en.wikipedia.org/v1/page/significant-events/2020_United_States_presidential_election?rvstartid=987405390). Maybe that's related to why [we're not handling redlinks more generally](https://phabricator.wikimedia.org/T267438), I'm not sure.
2) This bit of code, which formerly was just used for EditPreview, didn't actually do anything on a server error. I've done a quick and dirty stopping of the spinner, so you just get a white screen. But this should probably be improved to show some sort of error in this case.

Wikipedia/Code/Article as a Living Document/ArticleAsLivingDocViewController.swift
```diff
@@ -386,46 +385,49 @@ class ArticleAsLivingDocViewController: ColumnarCollectionViewController {
...
-                self.dismiss(animated: true) {
-                    self.delegate?.handleLink(with: url.absoluteString)
-                }
-            }
+        guard let fullURL = delegate?.articleURL.resolvingRelativeWikiHref(url.absoluteString) else {
+            return
+        }
+        switch Configuration.current.router.destination(for: fullURL) {
+            case .article(let articleURL): showInternalLink(url: articleURL)
+            default: navigate(to: fullURL)
```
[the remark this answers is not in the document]
mcleinman: This one is getting fixed in 3756

staykids (approved):

PR #3744, opened 2020-10-23
AaaLD - article content interstitial design polish

tonisevener:
**Fixes Phabricator ticket:** 
https://phabricator.wikimedia.org/T265666
https://phabricator.wikimedia.org/T265669

### Notes
* Open up article you've never seen on EN Wikipedia in LTR mode on device.
* Look at timestamp next to the New Changes badge.
* Now pull to refresh, confirm new changes badge switches to Last Updated and timestamp looks the same (it was different in some cases before)
* Confirm you don't see any bugs on the article interstitial (colors and spacing were tweaked). Test all themes.
* Button background colors on View Changes / View Discussion / Thank buttons in modal slightly changed, confirm that looks alright (note Carolyn still wants to rethink the Thank selected state).


Wikipedia/Code/ArticleWebMessagingController.swift
```diff
@@ -548,7 +548,7 @@ extension ArticleWebMessagingController {
             return
         }
         
-        var articleAsLivingDocBoxInnerHTML = "\(articleAsLivingDocBoxInnerContainerHTMLStart)<h4 id='significant-changes-header'>RECENT CHANGES</h4><ul id='significant-changes-list'>"
+        var articleAsLivingDocBoxInnerHTML = "\(articleAsLivingDocBoxInnerContainerHTMLStart)<h4 id='significant-changes-header'>SIGNIFICANT EVENTS</h4><ul id='significant-changes-list'>"
```
mcleinman, line 551: Should we translate this? While AaaLD is only in English for now, we've used translated strings for everything else.

Wikipedia/Code/ArticleWebMessagingController.swift
```diff
@@ -637,13 +637,6 @@ extension ArticleWebMessagingController {
                         seContainer.style.paddingRight = "\(layoutMargins.right)px";
                  }
 
-                 if (skeletonContainer) {
```
mcleinman, line 640: Can we also remove the `var skeletonContainer = document.getElementById('\(articleAsLivingDocBoxSkeletonContainerID)');` line above?

Wikipedia/Code/ArticleWebMessagingController.swift
mcleinman: Also: This now says "Significant Events", but the title of the modal is "Recent Changes". Should the modal also be updated?

www/significant-events-styles-sepia.css
```diff
@@ -1,10 +1,10 @@
 
 div#significant-changes-container{
-    background-color:#F8F1E3;
+    background-color:#E1DAD1;
```
mcleinman, line 3: Just wanted to confirm the background colors on these buttons - they seemed a bit more divergent than what was is on other colors.
Screenshot

mcleinman (changes requested):
Great tweaks overall. A few questions/comments. 

When playing with text sizes, I found another bug. Feel free to fix it here, or spin it out into its own Phab ticket (and assign to me if you want - it's likely a bug I didn't find before, almost definitely not something you introduced).
- The AaaLD modal uses the device's text size. At the largest (non-accessibility) text size and the smallest text size, open the AaaLD modal. (I didn't check on large accessibility sizes.)
- Note that the scrollview is scrolled all the way to the bottom - instead of the top, as expected. Also, the navbar underbar view is showing. 
- I'm unclear if this is a bug in the navbar, the auto-scrolling (which shouldn't be happening in this situation), or something else. But I'd be happy to dig in if you'd like.

Wikipedia/Code/ArticleWebMessagingController.swift
tonisevener: Oh good catch! The article content insert doesn't show small events, so we changed it here since it might be hiding more recent (small) changes. The modal still shows small events, so maybe this is fine. @carolynlimadeo what do you think, should the modal nav bar title and modal header view title label change from "Recent Changes" to "Significant Events" for consistency?

www/significant-events-styles-sepia.css
tonisevener: Pretty sure this came up in our polish and @carolynlimadeo was good with it. She also might be rethinking them as a part of [T266354](https://phabricator.wikimedia.org/T266354). I'll ask her to confirm for this PR though.

tonisevener: @mcleinman is that white snippet text a separate bug you noticed?

mcleinman: I couldn't replicate. I assume it's related to going between themes. But I tried a bunch, and couldn't replicate it. (Meant to use another screen shot that didn't show it, but put the wrong one up there.)

carolynlimadeo: Hi all! Yeah, I need to redefine the button colors, but this color will work for now for user testing. Sorry about the confusion!

Wikipedia/Code/ArticleWebMessagingController.swift
tonisevener: @mcleinman on your first point, I just pushed a commit that added translations for this method. We can ask about the headers in planning today.

Wikipedia/Code/ArticleWebMessagingController.swift
tonisevener: 👍  yep, this has been removed now.

Wikipedia/Code/ArticleWebMessagingController.swift
tonisevener: @mcleinman confirmed with Carolyn, she wants to leave the modal headers as "Recent Changes" since it includes small edits, and the article content insert does not. Thanks for checking!

mcleinman (approved):
Working great, thanks!

mcleinman:
> Great tweaks overall. A few questions/comments.
> 
> When playing with text sizes, I found another bug. Feel free to fix it here, or spin it out into its own Phab ticket (and assign to me if you want - it's likely a bug I didn't find before, almost definitely not something you introduced).
> 
>     * The AaaLD modal uses the device's text size. At the largest (non-accessibility) text size and the smallest text size, open the AaaLD modal. (I didn't check on large accessibility sizes.)
> 
>     * Note that the scrollview is scrolled all the way to the bottom - instead of the top, as expected. Also, the navbar underbar view is showing.
> 
>     * I'm unclear if this is a bug in the navbar, the auto-scrolling (which shouldn't be happening in this situation), or something else. But I'd be happy to dig in if you'd like.

Phab ticket for this bug: https://phabricator.wikimedia.org/T266621

PR #3745, opened 2020-10-23
AALD: highlight history cells, allow for multiple cells to be highlighted

mcleinman:
**Fixes Phabricator ticket:** https://phabricator.wikimedia.org/T266370

### Test Steps
1. In AALD modal, tap various "View changes" buttons. Ensure the appropriate cell (or cells, if multiple small edits were made) are temporarily highlighted.


Wikipedia/Code/PageHistoryViewController.swift
```diff
@@ -483,10 +483,21 @@ class PageHistoryViewController: ColumnarCollectionViewController {
                     cell.isSelected = false
                 }
                 cell.selectionOrder = SelectionOrder(cachedCellContent.selectionOrderRawValue)
+            } else if isTargetOfScroll {
```
tonisevener, line 486: Might be good to refactor this into a separate method or closure, so you aren't repeating code below.

tonisevener (commented):
Looks great! Just one small suggestion. I'm going to keep this on Hold for now and kick off a couple of Experimental builds for Carolyn to look at, we can merge after User Testing if we keep this transition to History.

Wikipedia/Code/PageHistoryViewController.swift
mcleinman: I refactored this, but then realized it's not actually animating the de-selection. Working on that now.

mcleinman: Improved the animation code. It was animating, but the border color and part of the button doesn't animate. It's possible to hack them in as animated, but that seemed heavy-handed for this experiment. Instead, we're now just using the background to highlight (which looks decent, to my untrained eyes), and more properly made the highlighting a function on the view, and the VC now just tells it to animate.

mcleinman:
Should we change the base of this to T265666?

Wikipedia/Code/PageHistoryViewController.swift
tonisevener: @mcleinman I'm not seeing your commit, can you try pushing again?

mcleinman: Now pushed up, thanks.

tonisevener:
Closing since we decided to push on directly to diff screen from user testing.

PR #3746, opened 2020-10-26

translatewiki:
Translation updates


mcleinman (approved):
👍

mcleinman:
Those tests are some localization check failures - will look into this in a bit.

mcleinman:
Updated the two `id` translation issues w/ proper variable. Asked team best way to handle `he` issue - ideally we'd just remove the translation and let someone else take a crack at it, but it seems like translation strings can only be updated, not removed.

mcleinman:
Emailed Yaron, a Hebrew translator, to ask for another look at the Hebrew string throwing errors. After that, tests should hopefully pass and we can merge this in.

mcleinman (approved):
👍

PR #3749, opened 2020-10-29
Merge sequential small event only sections into ranged section

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T265670

### Notes
Merges and collapses sequential sections with small changes matching this criteria:
- The section contains only small events
- Those small events are already compressed into one small event (so the `Section`'s `TypedEvent` count is 1)
- The sections being merged are sequential neighbors – even though their dates might not be strictly sequential, there are no intervening sections between them. For example: M1, M2, M3 where M is a matching section will be collapsed and merged into a new section M0, but M1, L1, M2 will not be merged into a new section.

This merging and collapsing behavior happens at two points - first when a `ArticleAsLivingDocViewModel` is constructed from `SignificantEvents` then again when a new page is fetched and those new events are appended to the existing visible sections. When merging sequential small event sections, a date range is also constructed that represents the collapsed range. This will be used in a future PR to help craft the timestamps in the modal.

I find my work here to be terribly inefficient and gross, _but_ it seems to work on a variety of articles in my testing so far without any noticeable performance implications (because the page size for new data is so small) so maybe the inefficiency isn't that bad here... 🤷🏽‍♂️

### Test Steps
1. Observe AAaLD modal without PR
	- Sequential small events within a page of data are not collapsed and merged.
	- Sequential small events at paging join points are not collapsed and merged.
2. Observe AAaLD modal with PR
	- Sequential small events within a page of data are collapsed and merged.
	- Sequential small events at paging join points are collapsed and merged.

Test articles I used often while implementing this were these two, as they had a lot of recent small events:

https://en.wikipedia.org/wiki/A_History_of_the_Birds_of_Europe
https://en.wikipedia.org/wiki/Ornithology

### Screenshots

small_changes


Wikipedia.xcodeproj/project.pbxproj
```diff
@@ -3399,7 +3399,7 @@
 		676070A1227CE60800A81F09 /* TalkPageReplyFooterView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = TalkPageReplyFooterView.swift; sourceTree = "<group>"; };
 		676070A32280987C00A81F09 /* TalkPageTopicNewViewController.xib */ = {isa = PBXFileReference; lastKnownFileType = file.xib; path = TalkPageTopicNewViewController.xib; sourceTree = "<group>"; };
 		676A8A8223A4013D0084B967 /* ArticleFetcher.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; name = ArticleFetcher.swift; path = ../Wikipedia/Code/ArticleFetcher.swift; sourceTree = "<group>"; };
-		6771298E24FF76AC00E89CA5 /* ArticleAsLivingDocViewController.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ArticleAsLivingDocViewController.swift; sourceTree = "<group>"; };
+		6771298E24FF76AC00E89CA5 /* ArticleAsLivingDocViewController.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ArticleAsLivingDocViewController.swift; sourceTree = "<group>"; usesTabs = 0; };
```
mcleinman, line 3402: Not a bug: I don't know what this flag does, and a quick google doesn't pop anything up. But many other files have the same flag - possibly something from the new Xcode version and how it handles window/tab management? Happy to approve the PR.

Wikipedia/Code/Article as a Living Document/ArticleAsLivingDocViewController.swift
```diff
@@ -162,19 +162,67 @@ class ArticleAsLivingDocViewController: ColumnarCollectionViewController {
...
+                        existingSections.append(proposedSection)
+                    }
                 }
             }
         }
 
+        // Collapse first proposed section into last current section if both only contain small events
+        if
+            let lastCurrentSection = currentSnapshot.sectionIdentifiers.last, lastCurrentSection.containsOnlySmallEvents,
+            let firstProposedSection = sections.first, firstProposedSection.containsOnlySmallEvents {
```
mcleinman, line 203: Should this first proposed section be the first section that doesn't contain the existing sections?

mcleinman (commented):
This looks great! Just a few questions.

Wikipedia.xcodeproj/project.pbxproj
staykids: This one's actually about using tabs vs. spaces. In the File inspector in Xcode when any file is selected, the bottom Text Settings section has the option to explicitly select which indent type the file should use (regardless of Xcode's settings). Since the files in the project mostly all use spaces for indentation, I just try to explicitly set it to "Spaces" when I'm working on files I haven't touched before.

WMF Framework/Significant Events Endpoint/ArticleAsLivingDocViewModels.swift
```diff
@@ -241,7 +333,22 @@ public extension ArticleAsLivingDocViewModel {
         
         public func hash(into hasher: inout Hasher) {
```
tonisevener, line 334: I'm a bit afraid to touch this since it's working - so feel free to ignore - but using `subtitleTimestampDisplay` for `SectionHeader`'s equality check and hash still feels odd, like it should be based on a more concrete identifier rather than a generated string meant to be displayed in UI. This was existing and my own doing, but if you want to I would welcome an attempt at this cleanup in your timestamp PR (especially if you find you need to rework `subtitleTimestampDisplay` anyways).

Wikipedia.xcodeproj/project.pbxproj
mcleinman: Awesome, thanks!

WMF Framework/Significant Events Endpoint/ArticleAsLivingDocViewModels.swift
```diff
@@ -241,7 +333,22 @@ public extension ArticleAsLivingDocViewModel {
...
             hasher.combine(subtitleTimestampDisplay)
+            hasher.combine(typedEvents)
         }
+
+        // MARK: - Helpers
+
+        public var containsOnlySmallEvents: Bool {
+            typedEvents.count == 1 && typedEvents.allSatisfy { event in
```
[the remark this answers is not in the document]
staykids: Good catch – wrote this bit before I created the other method. Updated.

WMF Framework/Significant Events Endpoint/ArticleAsLivingDocViewModels.swift
```diff
@@ -187,6 +188,94 @@ public struct ArticleAsLivingDocViewModel {
...
+
+            typedEvents = []
+            collapsedSectionHashes = []
+        }
+
+        var newlyCollapsedSectionHashes: [Int] = []
+
+        // Reconstruct sections with newly eligible small event sections collapsed in proper order
+        for section in sections {
+            if collapsedSections.contains(where: { $0.1.compactMap{ $0 }.contains { $0 == section.hashValue} }) && !newlyCollapsedSectionHashes.contains(section.hashValue) {
```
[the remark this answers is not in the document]
staykids: Amen – I mentioned in the description I find my work in this PR gross, and this is no exception. A particular offender. Was definitely in "just get it done" mode. Updated it - refactored out a lil' bit of the grossness and I _think_ it's still logically equivalent.

WMF Framework/Significant Events Endpoint/ArticleAsLivingDocViewModels.swift
```diff
@@ -187,6 +188,94 @@ public struct ArticleAsLivingDocViewModel {
...
+                if let collapsedSection = collapsedSections.first(where: { $0.1.compactMap { $0 }.contains(section.hashValue) }) {
```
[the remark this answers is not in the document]
staykids: Dropped with refactor.

Wikipedia/Code/Article as a Living Document/ArticleAsLivingDocViewController.swift
staykids: I certainly could be missing something, but I _believe_ it's sufficient without checking that in practice. That is, I think `!existingSections.contains(firstProposedSection)` will always evaluate to `true` in that if statement.

WMF Framework/Significant Events Endpoint/ArticleAsLivingDocViewModels.swift
```diff
@@ -187,6 +188,94 @@ public struct ArticleAsLivingDocViewModel {
         self.newChangesTimestamp = newChangesTimestamp
         self.lastUpdatedTimestamp = lastUpdatedTimestamp
     }
+
+    /// Collapses sequential sections that contain only small events into one section, including a date range that represents the collected events
+    static func collapseSmallEvents(from sections: [SectionHeader]) -> [SectionHeader] {
+        guard let dayMonthNumberYearDateFormatter = DateFormatter.wmf_monthNameDayOfMonthNumberYear(), let isoDateFormatter = DateFormatter.wmf_iso8601() else {
```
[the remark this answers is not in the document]
staykids: If it's cool with you, I'd prefer to address this if necessary in the timestamp PR rather than muddying this PR - was already messing around with it and `Section` hashability again there.

WMF Framework/Significant Events Endpoint/ArticleAsLivingDocViewModels.swift
```diff
@@ -187,6 +188,94 @@ public struct ArticleAsLivingDocViewModel {
...
+                typedEvents.append(contentsOf: sectionElement.typedEvents)
+            }
+
+            collapsedSectionHashes.append(contentsOf: sections[range].compactMap { $0.hashValue })
+
+            if let startIndex = range.first {
+                var smallChanges: [SignificantEvents.Event.Small] = []
+                for event in typedEvents {
+                    smallChanges.append(contentsOf: event.smallChanges)
+                }
```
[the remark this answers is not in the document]
staykids: Updated. A perfect way to hide my gross nested loops, thanks 😉

WMF Framework/Significant Events Endpoint/ArticleAsLivingDocViewModels.swift
staykids: I was definitely in "don't break, just make it work mode" on this PR, too. Same note above about giving this a try in the timestamp PR.

staykids:
Alright y'all – updated based on your feedback and ready for another review/merge if ready.

WMF Framework/Significant Events Endpoint/ArticleAsLivingDocViewModels.swift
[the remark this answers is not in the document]
mcleinman: Looks great, thank you!

WMF Framework/Significant Events Endpoint/ArticleAsLivingDocViewModels.swift
```diff
@@ -249,13 +252,19 @@ public struct ArticleAsLivingDocViewModel {
 
         var newlyCollapsedSectionHashes: [Int] = []
 
+        // Returns the small event collapsed section that represents the `sectionHash`, if one exists
+        func firstCollapsedSectionContaining(sectionHash: Int) -> CollapsedSection? {
+            return collapsedSections
+                .first { collapsedSection in collapsedSection.sectionHashes.compactMap { $0 }
```
mcleinman, line 258: Do we need the compactMap here? It just pulls out any nils, but sinc sectionHash can never be nil we don't need to worry about that? (This already was in the code, I just didn't catch it before.)

WMF Framework/Significant Events Endpoint/ArticleAsLivingDocViewModels.swift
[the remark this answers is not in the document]
mcleinman: Thanks. This is so much more readable now, much appreciated!

Wikipedia/Code/Article as a Living Document/ArticleAsLivingDocViewController.swift
mcleinman: Sounds reasonable to me - I think when I wrote that I neglected to consider that we're at most combining two sections into one in this block.

WMF Framework/Significant Events Endpoint/ArticleAsLivingDocViewModels.swift
staykids: Good catch - updated to a plain ole `map`.

mcleinman: Why do we need to use `map`... can it just be 

`return collapsedSections.first { collapsedSection in collapsedSection.sectionHashes.contains(sectionHash) }`

or even (depending on personal preference)

`return collapsedSections.first { $0.sectionHashes.contains(sectionHash) }`

staykids: Haha oops – this officially means no more programming for me today 🥴. Updated – now permanently shutting down my computer forever.

mcleinman: We all have those moments sometimes - I certainly have more than my share. :)

mcleinman (approved):
Looks great to me, thanks.

tonisevener (approved):
Looks great to me too! Nice work on this one - it looked like a doozy.

PR #3750, opened 2020-10-30
add loading indicator to bottom of AaaLD

mcleinman:
**Fixes Phabricator ticket:** https://phabricator.wikimedia.org/T265677

### Notes
* LAYOUT BUG: The footer is sized to fit around the loading indicator (w/ edge insets). Thus, it currently isn't centered, as the footer isn't full width. @tonisevener, is there a gotcha w/ how autolayout works w/ our collection views? Or am I just forgetting some gotcha on footers and collection views? Do I need to manually set the width to the collection view size?

* We talked about keeping it on screen for a minimum of 1 second. It gets replaced by the just-loaded items, and so it disappears in context. The only time it would just stop being shown is if we don't load more items (because of a network error, or otherwise). This seems rare enough that it didn't seem worth building the "1 second minimum" logic into an experiment. @carolynlimadeo, can you please let me know if you'd like it in there anyway?

### Test Steps
1. Open AaaLD modal. Scroll to bottom. See loading indicator and text saying "LOADING" beneath it. 
1. Repeat for other themes.

### DIFF
--- a/Wikipedia/Code/Article as a Living Document/ActivityIndicatorCollectionViewFooter.swift
+++ b/Wikipedia/Code/Article as a Living Document/ActivityIndicatorCollectionViewFooter.swift
@@ -15,19 +15,20 @@ class ActivityIndicatorCollectionViewFooter: UICollectionReusableView {
         loadingLabel.font = UIFont.wmf_font(.footnote, compatibleWithTraitCollection: traitCollection)
         loadingLabel.text = WMFLocalizedString("loading-indicator-text", value: "Loading", comment: "Text shown underneath loading indicator.").uppercased()
 
-        let stackView = UIStackView(arrangedSubviews: [loadingIndicator, loadingLabel])
-        stackView.axis = .vertical
-        stackView.spacing = 4.0
-        stackView.alignment = .center
+        addSubview(loadingLabel)
+        addSubview(loadingIndicator)
+        loadingLabel.translatesAutoresizingMaskIntoConstraints = false
+        loadingIndicator.translatesAutoresizingMaskIntoConstraints = false
 
-        addSubview(stackView)
-        stackView.translatesAutoresizingMaskIntoConstraints = false
+        // Need to give this a non-1000 priority to avoid constraint issues in the logs.
+        let separationConstraint = loadingLabel.topAnchor.constraint(equalTo: loadingIndicator.bottomAnchor, constant: 4)
+        separationConstraint.priority = UILayoutPriority(999)
         NSLayoutConstraint.activate([
-            stackView.topAnchor.constraint(equalTo: safeAreaLayoutGuide.topAnchor, constant: 30),
-            stackView.bottomAnchor.constraint(equalTo: safeAreaLayoutGuide.bottomAnchor, constant: -30),
-            stackView.leadingAnchor.constraint(greaterThanOrEqualTo: leadingAnchor, constant: 30),
-            stackView.trailingAnchor.constraint(lessThanOrEqualTo: trailingAnchor, constant: -30),
-            stackView.centerXAnchor.constraint(equalTo: centerXAnchor)
+            loadingIndicator.topAnchor.constraint(equalTo: safeAreaLayoutGuide.topAnchor, constant: 30),
+            separationConstraint,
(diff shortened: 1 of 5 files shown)


Wikipedia/Code/Article as a Living Document/ActivityIndicatorCollectionViewFooter.swift
```diff
@@ -0,0 +1,33 @@
...
+
+    private func setupView() {
+        loadingLabel.font = UIFont.wmf_font(.footnote, compatibleWithTraitCollection: traitCollection)
+        loadingLabel.text = WMFLocalizedString("loading-indicator-text", value: "Loading", comment: "Text shown underneath loading indicator.").uppercased()
+
+        let stackView = UIStackView(arrangedSubviews: [loadingIndicator, loadingLabel])
+        stackView.axis = .vertical
+        stackView.spacing = 4.0
+
+        wmf_addSubview(stackView, withConstraintsToEdgesWithInsets: UIEdgeInsets(top: 30, left: 30, bottom: 30, right: 30))
```
tonisevener, line 22: @mcleinman I actually didn't think we could mix and match auto layout cells with our manually laid out ones, so you made it farther than I thought. I would keep going with this, I think all you need to do is replace this line with something like:

```
        stackView.alignment = .center
        addSubview(stackView)
        stackView.translatesAutoresizingMaskIntoConstraints = false
        NSLayoutConstraint.activate([ stackView.centerXAnchor.constraint(equalTo: self.centerXAnchor),
                                      stackView.topAnchor.constraint(equalTo: self.topAnchor, constant: 30),
                                      stackView.bottomAnchor.constraint(equalTo: self.safeAreaLayoutGuide.bottomAnchor, constant: 30)])
```

Because `wmf_addSubview` doesn't add any centering constraints. You probably still need some more tweaks, like maybe add >= 30 constraints for leading and trailing in case the screen gets really small and the type is large. Also my `safeAreaLayoutGuide` in the snippet isn't working quite right:

Screen Shot 2020-10-30 at 10 22 30 AM
 
If you were to hook into our manual system, you would subclass `SizeThatFitsReusableView` and override sizeThatFits to position the indicator and label (see `ArticleAsLivingDocSectionHeaderView`. But I don't think you need to do that here.

tonisevener:
> The only time it would just stop being shown is if we don't load more items (because of a network error, or otherwise). This seems rare enough that it didn't seem worth building the "1 second minimum" logic into an experiment.

Yeah I agree, and more often than not it feels like the endpoint takes a second for paging anyway. I say we add it in later if we notice it feeling too quick.

Wikipedia/Code/Article as a Living Document/ActivityIndicatorCollectionViewFooter.swift
mcleinman: I mistakenly assumed the stack view would take full width, had some end-of-day brain spasm. Thanks, check it out now - should be working well.

Wikipedia/Code/Article as a Living Document/ActivityIndicatorCollectionViewFooter.swift
```diff
@@ -0,0 +1,42 @@
...
+        loadingLabel.text = WMFLocalizedString("loading-indicator-text", value: "Loading", comment: "Text shown underneath loading indicator.").uppercased()
+
+        let stackView = UIStackView(arrangedSubviews: [loadingIndicator, loadingLabel])
+        stackView.axis = .vertical
+        stackView.spacing = 4.0
+        stackView.alignment = .center
+
+        addSubview(stackView)
+        stackView.translatesAutoresizingMaskIntoConstraints = false
+        NSLayoutConstraint.activate([
```
tonisevener, line 25: I see some layout constraint errors on iOS13 simulators (14 seems fine). The UI lays out fine though, and they only seem to show when cells are appended to the collection view, not when the footer itself is displayed. So maybe it's just a matter of changing the priority on one of these to 999 for some state that it gets in when new cells are appended. Everything else looks great!

mcleinman: Good catch. This is now fixed!

tonisevener (approved):
Looks good to me! Feel free to merge whenever it's updated with `main`.

PR #3753, opened 2020-11-03
Fix scrolling bug for large/small text sizes

mcleinman:
**Fixes Phabricator ticket:** https://phabricator.wikimedia.org/T266621

### Notes
* Ultimately a slight rounding issue causes this bug.
* Feel free to push back if you think this is too hacky of a fix.
* See investigation [on Phab task](https://phabricator.wikimedia.org/T266621#6601603).
* While the bug only related to the `isAtTop` calculation, I added padding to the `isAtBottom` calculation under the assumption there could sometimes be rounding issues there, as well.

### Test Steps
1. Use smallest text size.
1. Run app.
1. Open an AaaLD modal. Ensure it is scrolled to the top, as expected.


WMF Framework/UIScrollView+Limits.swift
```diff
@@ -18,11 +18,13 @@ extension UIScrollView {
     }
 
     public var isAtTop: Bool {
-        return contentOffset.y <= topOffsetY
+        /// 0.5 is to account for  rounding errors. Sometimes when we expect them to be equal, these are less than .2 different (due to rounding) - and with multiple layout passes, it caused a large scrolling bug on a VC's launch.
+        return contentOffset.y <= topOffsetY + 0.5
```
staykids, line 22: What a strange bug! Is there special significance to using `0.5`? If not, what do you think about using `topOffsetY.rounded(.up)` instead, so we're at the next integral number without fear of more rounding errors?

mcleinman: No significance to `0.5` - just big enough to seem safe. You're absolutely correct, rounding is much cleaner. Updated this.

staykids (approved):

PR #3755, opened 2020-11-05
Add A/B Testing system for Article as a Living Doc

tonisevener:
**Fixes Phabricator ticket:** 
https://phabricator.wikimedia.org/T262941

### Notes
This PR adds A/B testing to the Article as a Living Doc experiment. Basic functions are:
1. Fetch the survey announcement on launch (but not the very first launch), which has the article list and testing percentage
2. Continue saving the article list in memory as we did before via `SurveyAnnouncementsController`
3. Evaluate the percentage in the announcement, and based on that assign a bucket.
4. Send bucket assignment through the legacy event logging system via the `MobileWikiAppiOSUserHistory` schema. Set its `test_group` field to either `LivingDoc_Test` or `LivingDoc_Control`.
4. Consider article list and bucket they landed in when deciding whether or not to show the experiment in the article view.
5. Do not reassign the bucket unnecessarily so that the user sees a consistent experiment.

### Test Steps
1. Pull and run the [wikifeeds patch](https://gerrit.wikimedia.org/r/c/mediawiki/services/wikifeeds/+/639409/) changes locally.
2. Go to project > Build Settings > search other swift flags > delete `-DWMF_LOCAL_PAGE_CONTENT_SERVICE` out of `LocalDebug` build configuration so you can test local wikifeeds, but hit production PCS. (I'm not sure why we'd ever need both to be local, but I didn't want to muddy up this PR with project build configuration improvements.)
3. Fresh launch project on the "Local Page Content Service & Announcements" scheme. Enable analytics in onboarding.
4. You should not see any indication that the test is running for the first launch (announcements aren't fetched, event logging `MobileWikiAppiOSUserHistory` schema does not send `test_group`, no experiment or survey in UI.
5. Terminate the app from simulator. Relaunch.
6. The announcements endpoint should be fetched on launch.`MobileWikiAppiOSUserHistory` schema should send a `test_group` assignment in it's event logging call.
7. If you are testing before 11/09 (the startDate in the announcements endpoint), you shouldn't see any experiment. Move your device date to 11/10 and relaunch (or change the announcements endpoint startDate and re-run locally).
8. Test an article that is not in the list. You should not see the experiment.
9. Test an article that is in the list. You should see the experiment IF you fell into the experiment bucket.
10. On your local wikifeeds, tweak the `percent_receiving_experiment` value, relaunch and repeat steps 8 - 10 and confirm experiment shows or hides as expected. (0 - no experiments show, 100 - always shows experiment, 50 - half show experiment).

Some notes as you're testing this:

- Bucket assignment is only generated on each new percentage the app detects. If it was at 50 percent the last time it generated a bucket, it does not randomly pick a new one, so the user sees a consistent experience.
- Related to this - the `MobileWikiAppiOSUserHistory` schema only posts if the user has agreed to share usage data via app settings or onboarding, and if its values have changed since last time it posted. So if you are assigned a new `test_group` bucket, it will post, otherwise you won't see it log.
- You can use a proxy to watch when the announcements endpoint is fetched and to watch when event logging is posted (to `meta.wikimedia.org/beacon/event...`, event is encoded in the url) with the `MobileWikiAppiOSUserHistory` and it's `test_group` assignment. I also find it's easier to throw this somewhere into `UserHistoryFunnel.swift`:

```
    override func log(_ eventData: [AnyHashable : Any]) {
        print("📘logging user history funnel")
        if let testGroup = eventData["test_group"] as? String {
            print("📗llogging test group: \(testGroup)")
        }
        super.log(eventData)
    }
```


WMF Framework/SurveyAnnouncementsController.swift
```diff
@@ -67,8 +92,10 @@ public final class SurveyAnnouncementsController: NSObject {
                 let googleFormattedArticleTitle = normalizedArticleTitle.googleFormPercentEncodedPageTitle else {
                     continue
             }
-            
-            guard let actionURL = announcement.actionURLReplacingPlaceholder("{{articleTitle}}", withValue: googleFormattedArticleTitle) else {
+    
+            //TODO: need to create this elsewhere to capture didSeeModal and isInExperiment
```
tonisevener, line 96: This TODO can be done as a part of https://phabricator.wikimedia.org/T264057.

WMF Framework/WMFContentGroup+Extensions.m
```diff
@@ -543,11 +543,6 @@ - (void)updateVisibilityForUserIsLoggedIn:(BOOL)isLoggedIn {
         return;
     }
     
-#if WMF_ANNOUNCEMENT_DATE_IGNORE
```
tonisevener, line 546: I did do one small project cleanup - I didn't see this used in any configurations and I would rather test as close to the Prod code as possible, so I decided to remove this macro.

mcleinman (approved):
Looks great and tests great.

PR #3756, opened 2020-11-05
Fix nav bar in place on modal

mcleinman:
**Fixes Phabricator ticket:** https://phabricator.wikimedia.org/T266898

### Notes
* Merging this into the new navigation branch

### Test Steps
1. Run on iPad.
1. Open an AaaLD modal.
1. Tap a "View discussion" button. (Article on United States has one fairly close to the top.)
1. Ensure the nav bar doesn't disappear when you scroll down.


Wikipedia/Code/ViewControllerRouter.swift
```diff
@@ -26,6 +26,10 @@ class ViewControllerRouter: NSObject {
             }
         }
 
+        if let vc = viewController as? SinglePageWebViewController, navigationController.modalPresentationStyle == .pageSheet {
+            vc.navigationBar.isBarHidingEnabled = false
+        }
+
```
staykids, line 32: I get slightly worried when high level specific logic seeps into these lower level controllers, especially in the case of something possibly temporary (like this experiment). What do you think of something like this below to be a bit more defensive regarding the cases the custom logic is applied to?

```suggestion
if #available(iOS 13.0, *), navigationController.children.contains(where: { $0 is ArticleAsLivingDocViewController }) {
	// All article as living document specific behavior
	if UIDevice.current.userInterfaceIdiom == .pad, let vc = viewController as? SinglePageWebViewController, navigationController.modalPresentationStyle == .pageSheet {
		vc.navigationBar.isBarHidingEnabled = false
	}
}
```

staykids (approved):
Nice – the navigation bar on iPad stays in place as the user scrolls. Small item below, feel free to ignore and merge if you don't think it's necessary.

Wikipedia/Code/ViewControllerRouter.swift
mcleinman: This is great, thank you. I really cringed at writing this, but the way the router is written I couldn't do it earlier. Adding more checks to this is good, agreed. Thanks.

staykids (approved):
👍 I fixed one minor thing with the simple navigation bar check – it was wrapped in the iPad idiom check, so I just moved that iPad specific check so the simple navigation bar is supported on both iPad and iPhone now.

PR #3762, opened 2020-11-09
Fix Xcode 12 warnings

staykids:
**Phabricator:** N/A

### Notes
* Used Xcode's automatic "upgrade settings" for Xcode 12 (it seems like the Sticker Pack was still set with a 11.0 deployment target, which was updated to 12.0 during the upgrade)
* The newly added build setting `CLANG_WARN_QUOTED_INCLUDE_IN_FRAMEWORK_HEADER` defaults to `YES`, but I changed it to `NO` because some of our dependencies are using double quoted includes
* Removes unused `googleFormattedArticleTitle` variable in `SurveyAnnouncementsController`


tonisevener (approved):
Looks good to me! Thanks for doing this.

PR #3763, opened 2020-11-09
Ignore announcement endpoint and always show AaaLD experiment

staykids:
**Phabricator:** N/A

### Notes
Just in case we need it if the announcement patch doesn't get merged.


tonisevener:
Looks like the [survey announcement](https://en.wikipedia.org/api/rest_v1/feed/announcements) is being returned from the endpoint now. Closing this.

PR #3765, opened 2020-11-10
Use number formatter for long character and edit counts in AaaLD experiment

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T265679

### Notes
* Use our `localizedThousandsStringFromNumber` number formatter logic for characters (added or deleted) and editor edit counts over 1000. 
* The counts in the header of the AaaLD modal are left unchanged.

### Test Steps
Until the survey announcement patch is merged, [this manual patch](https://github.com/wikimedia/wikipedia-ios/pull/3763/commits/d5b599ed32c212c1c1dd5b7f0cecf045e4d66330) is required to see the AaaLD experiment.

1. Observe the in article experiment - confirm that the interstitial character and author edit counts greater than 1000 are formatted as per the `localizedThousandsStringFromNumber`'s K, M, B pattern.
2. Observe the modal - confirm that other than in the page header, the same pattern as step 1 holds.

### Screenshot
number formatter


tonisevener (approved):
Looks good - tests were failing at `testIncomingTranslationStringForBracketSubstitutionsNotPresentInEN` because `aaald-added-text-description` and `aaald-deleted-text-description` had existing translations with the `PLURAL` synax, but EN did not. I renamed the key (appended a `-2` at the end) so it is seen as a new string needing a new translation. I think in this case this approach valid since the string value changed so much. My hope is the original translation will be cleaned out on `translatewiki`s side, I assume there's already some automated way of cleaning out no longer used translations just from any old code cleanup we might do.

staykids:
> Looks good - tests were failing at `testIncomingTranslationStringForBracketSubstitutionsNotPresentInEN` because `aaald-added-text-description` and `aaald-deleted-text-description` had existing translations with the `PLURAL` synax, but EN did not. I renamed the key (appended a `-2` at the end) so it is seen as a new string needing a new translation.

@tonisevener Ahhh thank you – I was wondering what was up with that.

PR #3766, opened 2020-11-10
Rework survey timer to handle Article as a Living Doc experiment

tonisevener:
**Fixes Phabricator ticket:** 
https://phabricator.wikimedia.org/T264057

### Notes

- Reworked survey timer logic to display survey prompt at the right moment around Article as a Living Doc experiment
- Confirmed the control experience survey prompt and timer has no regressions.
- Add extra logic for survey announcement to take priority over fundraising, should the user qualify for both.

#### Tips

It is handy to take what is returning from the [announcements endpoint](https://en.wikipedia.org/api/rest_v1/feed/announcements) currently, tweak it and map locally via a proxy. I did this for the `displayDelay` value so I don't have to wait as long for the survey to show, and for the `percent_receiving_experiment` value to easily turn on or off the experiment to test each flow.

#### Experiment Flow

1. Fresh install, launch, go to an article **not** in list, experiment and survey shouldn't show. Go to article in list, experiment and survey doesn’t show (nothing shows on first launch).
2. Terminate from app and relaunch. Go to article **not** in list, experiment and survey shouldn't show. Go to article in list (on a small device so content insert is off screen), experiment shows. Wait `displayDelay` seconds. Survey announcement should not show. Scroll until content insert shows. Wait `displayDelay` seconds. Survey should show. Tap through to survey, confirm article title, `true` for `inExperiment` and `false` for `sawModal` are passed through. Go back to app, confirm survey prompt has dismissed. Leave article and come back, confirm survey doesn't show again but experiment shows. Terminate app and come back to article, confirm survey doesn't show again but experiment shows.
3. Do step 2, except this time on a larger device so that the content insert is seen on load without any scrolling. Confirm survey shows after waiting `displayDelay` seconds.
4. Reinstall the app, launch and terminate to get to 2nd launch. Go to article in list, scroll to content insert. As timer is going after scrolling, tap "Read more updates". Timer should extend by 5 seconds. After timer is up survey prompt should display on top of AaaLD modal. Tap through to survey, confirm article title, `true` for `inExperiment` and `true` for `sawModal` are passed through.
5. Reinstall app, launch and terminate to get to 2nd launch. Go to article in list, scroll to content insert. Wait until survey prompt shows. Dismiss survey prompt. Pull to refresh on article and scroll back to content insert. Confirm survey prompt doesn't show. Leave article and come back, confirm survey doesn't show again but experiment shows. Terminate app and come back to article, confirm survey doesn't show again but experiment shows.
6. Reinstall app, launch and terminate to get to 2nd launch. Go to article in list, scroll to content insert. As timer is going, tap a link in the article (one NOT in the article list). Confirm timer stops and survey is not shown on the other article. Tap back, confirm timer starts back up and you see the survey after it fires.
7. Repeat step 6, but this time tap link before content insert is seen on screen. you should not see survey on new article. tap back. You should not see survey on old article (because you never saw the content insert).
8. Repeat step 6, but this time tap Read More Updates as timer is going. Then tap View Changes so that a new screen is pushed on from the right. Confirm survey doesn't show when timer fires. Tap back. Confirm survey shows after timer time has elapsed.
9. Reinstall, go to 2nd launch, visit article in list. Before scrolling to content insert, background and foreground. Confirm no survey ever shows up. Then scroll to content insert. Timer should start. As timer is going, background for a while then foreground. Confirm timer restarts where it left off then survey shows after it fires.

#### Control Flow (test for regressions)

1. Fresh install, launch, go to an article **not** in list, experiment and survey shouldn't show. Go to article in list, experiment and survey doesn’t show (nothing shows on first launch).
2. Terminate from app and relaunch. Go to article **not** in list, experiment and survey shouldn't show. Go to article in list, experiment shouldn't show and survey shows after waiting `displayDelay` seconds after article view appearance. Tap through to survey, confirm article title, `false` for `inExperiment` and `false` for `sawModal` are passed through. Go back to app, confirm survey prompt has dismissed. Leave article and come back, confirm survey doesn't show again. Terminate app and come back to article, confirm survey doesn't show.
3. Reinstall app, launch past onboarding and terminate to get to 2nd launch. Go to article in list. Background while timer is running on article. Confirm timer is paused. Foreground, confirm timer resumes where left off and survey shows after remaining seconds have elapsed.
4. Reinstall app, launch past onboarding and terminate to get to 2nd launch. Go to article in list. While timer is running, tap a link in the article to go to a different article NOT in the list. Confirm survey doesn't show. Background and foreground, confirm survey still doesn't show. Tap back to get back to article in list. Confirm survey shows after `displayDelay` seconds.
5. Reinstall app, launch past onboarding and terminate to get to 2nd launch. Go to article in list. While timer is running, tap a link to a different article ALSO in the list (i.e. Joe Biden > Kamala Harris). Survey should show after waiting `displayDelay` seconds. Tap through to survey, confirm you see Kamala Harris’s article title in survey, not Joe Biden. Go back to Joe Biden, confirm survey does not display.

#### Fundraising Announcement

One business rule for this is that the survey announcement should take priority over the fundraising announcement. Once the survey announcement is acted upon, the fundraising announcement should show the next time the article is visited. We will test this by mapping locally.

1. Map local the announcements endpoint. Find the fundraising announcement with `id` = `FUNDRAISING20IOSV2EN` and `countries` = `US`. Change the `start_time` to today's date.
2. Fresh install and launch. Nothing should show on any articles.
3. Terminate from app and relaunch. Go to an article in the list. If you are in the A/B test (depending on the percentage), you might see the experiment. You will see the survey prompt at the correct time according to the experiment/control flow above. Dismiss the survey prompt. Leave the article and come back. You should see the fundraising announcement on appearance of the article view (it is okay that both prompts show in the same session). Dismiss the fundraising announcement. Terminate the app and relaunch. You shouldn't see the survey prompt or the fundraising announcement (but might still see the experiment depending on the A/B test assignment).


tonisevener:
This one is ready for review.

Wikipedia.xcodeproj/project.pbxproj
```diff
@@ -13920,7 +13920,7 @@
 				);
 				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
 				INFOPLIST_FILE = "Wikipedia Stickers/Info.plist";
-				IPHONEOS_DEPLOYMENT_TARGET = 12.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 11.0;
```
tonisevener, line 13923: After sending off to TestFlight we received an email with this warning:

"Architecture incompatible with MinimumOSVersion. The app bundle at 'Wikipedia.app/PlugIns/Wikipedia Stickers.appex' specifies a MinimumOSVersion of '12.0' but contains a 32-bit architecture that is unsupported on iOS 12 and later."

This fixes that. We had changed it to 12.0 originally because the project settings automatic update recommended it after upgrading Xcode.

staykids:
This is probably beyond the scope of this PR itself since it's only marginally related to our implementation, but wanted to mention: would it be more ideal if the Google Forms answer fields for `sawModal`,`inExperiment`, and `Article Source` were read-only and non-editable? Apologies if I missed discussion about this in the Phab ticket, but it seems like we're potentially inviting bad data by leaving them editable. 

I have very limited working knowledge of Google Forms though – lazy Googling seems to suggest there is a way to make answer fields read-only, but I can't be certain.

staykids:
Again, thanks for the detailed test steps - helping keep my testing very focused. So far it's going swimmingly, but one unexpected item I keep running into is during step 3 in "Fundraising Announcement".

Here's what's happening - it seems like the last statement in that step ("You should still not see the experiment or survey") is the opposite of what happens. For example:

1. Follow steps to show fundraising announcement on COVID-19 pandemic article. Dismiss it.
2. Return to explore feed, then search for and open COVID-19 pandemic article again.
3. AAaLD experiment is displayed.

I want to confirm – is this unexpected? And if so, are these the proper proxy mapping conditions to test that flow:

• I'm not manipulating `percent_receiving_experiment` - leaving it as the API provides
• I'm only updating the `start_time` for `FUNDRAISING20IOSV2EN` (as the `platforms` value appears to have been merged)

tonisevener:
> Again, thanks for the detailed test steps - helping keep my testing very focused. So far it's going swimmingly, but one unexpected item I keep running into is during step 3 in "Fundraising Announcement".
> 
> Here's what's happening - it seems like the last statement in that step ("You should still not see the experiment or survey") is the opposite of what happens. For example:
> 
> 1. Follow steps to show fundraising announcement on COVID-19 pandemic article. Dismiss it.
> 2. Return to explore feed, then search for and open COVID-19 pandemic article again.
> 3. AAaLD experiment is displayed.
> 
> I want to confirm – is this unexpected? And if so, are these the proper proxy mapping conditions to test that flow:
> 
> • I'm not manipulating `percent_receiving_experiment` - leaving it as the API provides
> • I'm only updating the `start_time` for `FUNDRAISING20IOSV2EN` (as the `platforms` value appears to have been merged)

@staykids thanks for finding that - I do think that's a unexpected and a poor experience. Seeing two different popups in the same session sounds annoying. I have a couple more things related to this to confirm with Josh later today so I'll ask him about that.

tonisevener:
> This is probably beyond the scope of this PR itself since it's only marginally related to our implementation, but wanted to mention: would it be more ideal if the Google Forms answer fields for `sawModal`,`inExperiment`, and `Article Source` were read-only and non-editable? Apologies if I missed discussion about this in the Phab ticket, but it seems like we're potentially inviting bad data by leaving them editable.
> 
> I have very limited working knowledge of Google Forms though – lazy Googling seems to suggest there is a way to make answer fields read-only, but I can't be certain.

Yeah definitely agreed on that - Shay said she looked into this and this was the closest we could get. I was having a hard time finding a clear-cut answer on it but let us know if you found a way to make it non-editable (we can discuss in planning with Shay). Here's a couple of links that seem to suggest it won't work:

https://stackoverflow.com/a/58800891
https://support.google.com/docs/thread/22663508?hl=en

tonisevener:
> Again, thanks for the detailed test steps - helping keep my testing very focused. So far it's going swimmingly, but one unexpected item I keep running into is during step 3 in "Fundraising Announcement".
> 
> Here's what's happening - it seems like the last statement in that step ("You should still not see the experiment or survey") is the opposite of what happens. For example:
> 
> 1. Follow steps to show fundraising announcement on COVID-19 pandemic article. Dismiss it.
> 2. Return to explore feed, then search for and open COVID-19 pandemic article again.
> 3. AAaLD experiment is displayed.
> 
> I want to confirm – is this unexpected? And if so, are these the proper proxy mapping conditions to test that flow:
> 
> • I'm not manipulating `percent_receiving_experiment` - leaving it as the API provides
> • I'm only updating the `start_time` for `FUNDRAISING20IOSV2EN` (as the `platforms` value appears to have been merged)

Per Josh in planning, it is okay if the user sees two popups in the same session. I pushed a commit swapping the priorities around, now the survey prompt should show first before the fundraising announcement. I updated the test steps for that section as well.

staykids (commented):
Nice! Two minor ignorable comments below. Will wait to take another code pass and run through the test steps when the dust settles. Until then, gonna move over to review https://github.com/wikimedia/wikipedia-ios/pull/3769.

Wikipedia/Code/Article as a Living Document/ArticleAsLivingDocController.swift
```diff
@@ -85,8 +93,8 @@ class ArticleAsLivingDocController: NSObject {
         let isDeviceRTL = view.effectiveUserInterfaceLayoutDirection == .rightToLeft
         let isENWikipediaArticle = delegate.articleURL.host == Configuration.Domain.englishWikipedia
```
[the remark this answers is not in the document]
tonisevener: Good idea - I fixed this and moved EN Wikipedia and LTR checking to the guard. `willShowFundraisingAnnouncement` ended up getting scrapped but I do check A/B test buckets which hit persistence so it was still valid feedback.

Wikipedia/Code/ArticleViewController+Announcements.swift
```diff
@@ -2,6 +2,24 @@ import WMF
 import CocoaLumberjackSwift
 
 extension ArticleViewController {
+    
+    var willShowFundraisingAnnouncement: Bool {
+        guard UserDefaults.standard.shouldCheckForArticleAnnouncements else {
+            return false
+        }
+        let predicate = NSPredicate(format: "placement == %@", "article")
```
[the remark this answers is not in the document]
tonisevener: Sounds good, this is fixed. I had to add single quotes around it to prevent a crash.

tonisevener:
@staykids Logic has been tweaked and the test steps are working for me. This is ready for your full review again, thanks!

staykids:
> @staykids Logic has been tweaked and the test steps are working for me. This is ready for your full review again, thanks!

Just finished another code pass of the new changes. About to run through all the test steps again now.

Wikipedia/Code/Article as a Living Document/ArticleAsLivingDocController.swift
```diff
@@ -69,24 +77,21 @@ class ArticleAsLivingDocController: NSObject {
     var shouldAttemptToShowArticleAsLivingDoc: Bool {
         
         guard let delegate = delegate,
-              let view = delegate.view else {
+              let view = delegate.view,
+              view.effectiveUserInterfaceLayoutDirection == .leftToRight,
+              delegate.articleURL.host == Configuration.Domain.englishWikipedia else {
```
staykids, line 82: No change needed, just wanted to provide more detail about my previous rationale for having `delegate.articleURL.host == Configuration.Domain.englishWikipedia` higher up in the `guard` if it's helpful at all to read (or if it seems I'm thinking about it incorrectly, lemme know 🥴):

Typically, I always treat the order of these guard statements as most restrictive (general) to least restrictive (specific) and least expensive performance wise to most. So in the case of these `.leftToRight` and `englishWikipedia`, to me they are both probably effectively equivalent performance wise. But restriction wise, the `englishWikipedia` restricts the audience we're interested in here much more widely than the `.leftToRight` check.

It's probably some sense of silly premature optimization and organization on my part, but I think sometimes that type of precedence organization also has benefits beyond performance gains, like readability.

staykids (approved):
Quite a feat covering all these announcement pathways, Toni! Good stuff. There's one minor test case condition I had unexpected results with I'll detail here, but otherwise went ahead and approved if it's just me misunderstanding the expectation or we're comfortable with the result.

In the fundraising announcement flow, the expectation you stated is that the survey should take priority over the fundraising announcement. This indeed seems to be the case when the user is in the A/B test (I just set `percent_receiving_experiment` to 100 to include them) – the survey prompt comes up on a qualifying article first, then the next visit shows the fundraising alert. But if the user isn't in the A/B test (`percent_receiving_experiment` to 0), then the flow is reversed – they see the fundraising alert first, then the survey prompt.

I honestly could be doing something wrong, but wanted to mention.

Wikipedia/Code/Article as a Living Document/ArticleAsLivingDocController.swift
tonisevener: Makes sense, I can see how that helps with readability. I pushed a quick commit shifting this and hopefully fixing your reverse priorities of the prompts if the user is not in the A/B test. I'm confirming there were not regressions related to that part now though, if I find any I'll revert. Good catch on that!

tonisevener: I've gone through the test cases again with my last change - they still look good and the survey shows up before fundraising now if they don't land in the experiment. I do notice the fundraising announcement occasionally doesn't show up until an additional launch or explore feed refresh, which is a bit unexpected but actually more in line with the ideal functionality, that is that they don't see back to back modals. Since Josh sounded okay with back to back modals anyways, I figure this slight inconsistency is okay. I'll mention in sync tomorrow.

PR #3767, opened 2020-11-16

translatewiki:
Translation updates


mcleinman:
Failing test caused by Ukranian translation for [on this day footer](https://translatewiki.net/w/i.php?title=Wikimedia:Wikipedia-ios-on-this-day-footer-with-event-count/uk&oldid=9080476). I've reverted that translation change, so it should now work.

That said, the string should support plural. I'll put a PR up to update that.

mcleinman (approved):
👍

PR #3768, opened 2020-11-17
On this day translation -> plural

mcleinman:
A volunteer translated this w/ a plural string, causing our tests to fail. But y'know what, it should be plural. They were right.

### Notes
* In English Wikipedia, it is highly unlikely that we'll ever need the singular form - but I'm not sure how this is used in other language's Wikipedias.
* I changed the key to ensure we get new translations for all languages.

### Test Steps
1. Run the app, confirm that the On This Day card on the explore feed says something like "37 more historical events on this day" beneath it.
1. Change the line `return CommonStrings.onThisDayFooterWith(with: otherEventsCount)` to `return CommonStrings.onThisDayFooterWith(with: 1)`, and run the app again.
1. Confirm that the On This Day card on the explore feed now says "1 more historical event on this day" beneath it, using the singular version of "event".

### Screenshots/Videos
Screenshot 6


tonisevener (approved):
Looks good to me!

PR #3769, opened 2020-11-17
AaaLD - Navigate to particular talk page topic if possible

tonisevener:
**Fixes Phabricator ticket:** https://phabricator.wikimedia.org/T267849

### Notes
* This is using the talk page topic section title in the significant events models to navigate the user to a specific topic after tapping "View discussion" by passing it in as a fragment. If there is an issue with generating the fragment, default to the article talk page like it was before. 

### Test Steps
1. Go to AaaLD experiment for various articles and tap "View discussion".
2. Confirm it goes directly to the article talk page topic rather than the homepage.
3. If it doesn't, check to see if that topic still exists on desktop - it may have been archived in which case there's not much we can do. If it does exist, please mention as a bug comment on this PR. It's likely something with the encoding (which has been finicky).


WMF Framework/Significant Events Endpoint/ArticleAsLivingDocViewModels.swift
```diff
@@ -905,17 +910,23 @@ public extension ArticleAsLivingDocViewModel.Event.Large {
...
-            }
-            
-            return section
-        }
+        let finalSet = set.map { Self.sectionTitleWithEqualSignsStripped(originalTitle: $0) }
         
         return Set(finalSet)
     }
     
+    private static func sectionTitleWithEqualSignsStripped(originalTitle: String) -> String {
```
staykids, line 918: `==` is the title demarcator for these original titles, but I don't know enough about talk pages to know if `=` is a protected character _within_ the topic name itself. If so, we could potentially return a malformed title here.

Is there a particular advantage to this regex approach over using something higher level on `String` or just explicitly removing the known prefixes and suffixes? I'm probably missing something, but in my exploring so far the title format always seems to be `== Title ==`. If `==` is not protected within the body, removing the known prefix and suffix would prevent the malformed case I mentioned. If `==` is protected within the entire body, I think something like this would be equivalent:
```swift
originalTitle.replacingOccurrences(of: "==", with: "").trimmingCharacters(in: .whitespacesAndNewlines)
```

To be clear, I'm totally fine with how it is now. Just wanna learn if there's something I'm missing.

staykids (approved):
Love it – so nice we're able to take the user to what they're actually after instead of making them hunt for it.

Been working great so far. So far, I've noticed the cases of either straight to the discussion itself or straight to the homepage, and in the case of straight to the homepage confirmed the page is inaccessible on desktop web. I have not run into the potential bug case of getting pushed to the homepage, then finding it's indeed accessible on desktop web.

One note below, feel free to ignore the change.

WMF Framework/Significant Events Endpoint/ArticleAsLivingDocViewModels.swift
tonisevener: @staykids yeah, I was basing this complication on the way [section titles](https://en.wikibooks.org/wiki/MediaWiki_User_Guide/Sections_and_Headings) in general can be from 2 to 6 equal signs, and also have 0 or more spaces between the equal signs and the title content, so I was trying to write something loose enough to strip all of that. I haven't seen a situation where a higher level heading than 2 is used for the talk page topic that gets extracted from the endpoint, but I don't think there's any guarantee that it'll only be 2 equal signs. My regex skills are sorely lacking though so happy to hear of any better way to write this. I just tested and you're right that two or more equal signs within the title itself will also get stripped. I'll try for a bit to fix that, but I do think it's pretty edge-casey and they'll just land on the home talk page instead of a specific topic if it does happen.

tonisevener: @staykids I updated the regex - now it's limited to only checking the beginning and ending of a title so that middle equal signs aren't stripped. Also decreased the number of equal signs to look for down to 1. That's still a wikitext possibility though I don't see that happening since that would be the title of a page.

staykids: Awesome - thanks for the details and the update (and the opportunity for me to brush up on my weak regex.) Test cases seem to be working identically to yesterday for me.

tonisevener: 👍  thanks for confirming - this PR is a bit stuck until the base branch PR is merged (I did that to make it easier to kick off a beta build with this fix for the demo yesterday). Once https://github.com/wikimedia/wikipedia-ios/pull/3766 is merged we can merge this one.

PR #3771, opened 2020-11-18
Improve anonymous user display in AaaLD

mcleinman:
**Fixes Phabricator ticket:** https://phabricator.wikimedia.org/T265680

### Notes
* Updated IP users to just say "anonymous user"
* Icon for bots was already showing on both interstitial and modal.
* Added icon for anonymous users to modal and interstitial. (Doing it at start of "Edit by anonymous user" string. Otherwise, we need to get the translation in two parts in order to do "Edit by [image] anonymous user", and we don't have a grammatical piece in our translation style for "insert image here".) I think it ultimately looks okay.
* We're not updating the color of the interstitial icons, as they are injected into the HTML. (We could do that by changing the images as the theme changes on the article, or some extensive CSS - either way, probably a larger change for another time.) For the bot, the blue is close enough that it might be okay. For anonymous users the text is various shades of gray, and the single gray icon seems good enough. @carolynlimadeo seemed okay with this, for now.

### Test Steps
1. Open an article with a recent anonymous editor. (It may help if you adjust the computed `var shouldAttemptToShowArticleAsLivingDoc` to always return true, so you see the experiment on all articles.)
1. Ensure the interstitial doesn't show the user's IP address, and includes an anonymous user icon.
1. Open AaaLD modal. 
1. Ensure user IPs don't show for anonymous users, and that they use an anonymous user icon.

### Screenshots/Videos 
Screenshot 7 Screenshot 8


tonisevener (approved):
Looks good! I noticed our build script made some changes to the .strings files - manually comparing one of them showed that the translation for `aaald-revision-userInfo-anonymous` was deleted, so I committed them. I didn't see an addition for `aaald-revision-by-anonymous` but I think that's because translations for it do not exist yet.

mcleinman:
> Looks good! I noticed our build script made some changes to the .strings files - manually comparing one of them showed that the translation for `aaald-revision-userInfo-anonymous` was deleted, so I committed them. I didn't see an addition for `aaald-revision-by-anonymous` but I think that's because translations for it do not exist yet.

Thanks for taking care of that.

PR #3774, opened 2020-11-19
Remove extraneous HTML from AAaLD article insert event descriptions.

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T268201

### Notes
To guard against a JavaScript injection error, use our existing `removingHTML` helper before injecting interstitial article snippet. We may not necessarily need this depending on where your endpoint changes land, but honestly it might be a nice extra line of defense in case the endpoint returns something unexpected that causes the experiment to fail. So far, I haven't knowingly witnessed any regressions using this with other articles in the experiment, but I've only been testing for a couple hours so far so I could very well be missing something.

The part I keep going back and forth on is the element I commented out - whether to apply it to both the article insert description and the modal content descriptions, or just the insert itself. I left it commented out in `localizedStringFromSections`. I guess I slightly lean towards it being ok to use this in the modal as well (see screenshot below), since at worst case so far it removes some formatting but at best it removes a bunch of useless cruft. But again, I could be missing something.

What do you think?

### Test Steps
1. Open `Joe Biden` article
2. Confirm AAaLD experiment loads and that the article insert segment renders
3. Open known to be currently working on 6.7.3 TestFlight build articles in experiment
4. Confirm same elements as Step 2

### Screenshots

Here's a screenshot of before and after removing the HTML in the modal - we lose the italic formatting, but in the article insert itself that's more desirable than the experiment failing to load at all.

section

### DIFF
--- a/WMF Framework/Significant Events Endpoint/ArticleAsLivingDocViewModels.swift
+++ b/WMF Framework/Significant Events Endpoint/ArticleAsLivingDocViewModels.swift
@@ -601,7 +601,7 @@ public extension ArticleAsLivingDocViewModel {
                     userGroups = newTalkPageTopic.userGroups
                     
                     if let talkPageSection = newTalkPageTopic.section {
-                        self.buttonsToDisplay = .viewDiscussion(sectionName: Self.sectionTitleWithWikitextStripped(originalTitle: talkPageSection))
+                        self.buttonsToDisplay = .viewDiscussion(sectionName: Self.sectionTitleWithWikitextAndHtmlStripped(originalTitle: talkPageSection))
                     } else {
                         self.buttonsToDisplay = .viewDiscussion(sectionName: nil)
                     }
@@ -910,14 +910,15 @@ public extension ArticleAsLivingDocViewModel.Event.Large {
         }
         
         //strip == signs from all section titles
-        let finalSet = set.map { Self.sectionTitleWithWikitextStripped(originalTitle: $0) }
+        let finalSet = set.map { Self.sectionTitleWithWikitextAndHtmlStripped(originalTitle: $0) }
 
         return Set(finalSet)
     }
     
     //remove one or more equal signs and zero or more spaces on either side of the title text
-    private static func sectionTitleWithWikitextStripped(originalTitle: String) -> String {
-        var loopTitle = originalTitle
+    //also removing html for display and potential javascript injection issues - https://phabricator.wikimedia.org/T268201
+    private static func sectionTitleWithWikitextAndHtmlStripped(originalTitle: String) -> String {
+        var loopTitle = originalTitle.removingHTML
         
         let regex = "^=+\\s*|\\s*=+$"
         var maybeMatch = loopTitle.range(of: regex, options: .regularExpression)
@@ -945,7 +946,7 @@ public extension ArticleAsLivingDocViewModel.Event.Large {
             localizedString = String.localizedStringWithFormat(CommonStrings.manySectionsDescription, sections.count)
         }
 
-        return " " + localizedString.removingHTML
+        return " " + localizedString
     }
     
     private func localizedSectionHtmlSnippet(sectionsSet: Set<String>) -> String? {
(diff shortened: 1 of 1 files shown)


WMF Framework/Significant Events Endpoint/ArticleAsLivingDocViewModels.swift
```diff
@@ -944,7 +944,9 @@ public extension ArticleAsLivingDocViewModel.Event.Large {
         default:
             localizedString = String.localizedStringWithFormat(CommonStrings.manySectionsDescription, sections.count)
         }
-        
+
+        // Uncomment to remove extraneous HTML in the modal
+        // return " " + localizedString.removingHTML
```
tonisevener, line 949: @staykids Actually I think it would work best if this was the only line you had. Generating the content insert html snippet above eventually calls this, so we would be stripping html **before** wrapping them in italic tags for the insert, thus getting to keep that styling (win-win). And I agree with you that it would be nice to remove that cruft from the modal descriptions. It would be good to test the View Discussion > talk page transition with this though. We should confirm that still works since we're relying on section titles to make that fragment.

tonisevener (commented):
Thanks for digging in and investigating this! I have one comment - I think we can find a way to gain this without losing styling for sections in the content insert.

WMF Framework/Significant Events Endpoint/ArticleAsLivingDocViewModels.swift
staykids: Cool - makes sense if we're dropping it both places. Updated. 

I don't know if there's a particular talk page case you've noticed, but so far I haven't noticed any behavior out of the ordinary when using the same test methodology I used in https://github.com/wikimedia/wikipedia-ios/pull/3769, using the Joe Biden and COVID-19 pandemic articles as test articles. Lemme know if you noticed a particular case or anything else specific to the view discussion > talk page transition you want me to test with this update.

Since our bad data is gone from the endpoint, I've been stubbing in our known bad data (from the [Phab ticket](https://phabricator.wikimedia.org/T268201#6634654)) as appropriate to make sure the result is intended. Still poking at this one, but wanted to go ahead and put up this update if you felt it was sufficient to approve/merge (as I don't wanna hold up any of your new RC plans.)

tonisevener:
@staykids I noticed italics were still getting stripped in both locations (guess I was reading the code wrong), so I moved it to a place that strips it earlier in the process. This seems to work and the talk page topic transition still seems to work as it did before. I'll approve soon after some final testing but can you also do a last little bit of testing on your end before I merge?

Before [fff01d3](https://github.com/wikimedia/wikipedia-ios/pull/3774/commits/fff01d33f1842b77a96f6b027b73996bb2ca605d):

Screen Shot 2020-11-20 at 4 03 34 PM
Screen Shot 2020-11-20 at 4 03 41 PM
Screen Shot 2020-11-20 at 4 03 44 PM

After:
Screen Shot 2020-11-20 at 3 57 01 PM
Screen Shot 2020-11-20 at 3 57 11 PM

staykids:
Thanks for looking at it – that's so weird, I could have sworn the italics were fine when I was stubbing the bad data, but I guess I might not have stubbed in the best places. Will do another pass with your latest changes.

tonisevener (approved):

staykids:
I expanded my testing beyond the two I was using to more articles on the allow list.  So far so good:

- Italics seem to be displayed appropriately (section names)
- Thank and view buttons aren't impacted (because they aren't reliant on the section name)
- View discussion doesn't seem to be affected (probably because the titles are intended by the authors to be human readable anyway, so that provides us with some additional security beyond our HTML stripping)
- Remaining content appears to be the same

I'm happy to continue free form or more focused testing in specific areas on this if you'd like me to, or feel free to merge if you're feeling good about it too.

tonisevener:
> I expanded my testing beyond the two I was using to more articles on the allow list. So far so good:
> 
> * Italics seem to be displayed appropriately (section names)
> * Thank and view buttons aren't impacted (because they aren't reliant on the section name)
> * View discussion doesn't seem to be affected (probably because the titles are intended by the authors to be human readable anyway, so that provides us with some additional security beyond our HTML stripping)
> * Remaining content appears to be the same
> 
> I'm happy to continue free form or more focused testing in specific areas on this if you'd like me to, or feel free to merge if you're feeling good about it too.

Great, thanks @staykids! I think this is good to go.

PR #3775, opened 2020-11-20
Fix bugs found with AaaLD event logging

tonisevener:
**Fixes Phabricator ticket:** https://phabricator.wikimedia.org/T262616

### Notes
* Position was failing because we were sending it as a string
* Also starting position counting at 0 to match how we log tapping on search results.
* Also added single quotes around the `type` value to help Shay parse on the back end.

WIP to fix issues with the `type` field. We need more info from Analytics side on the proper way to send this value.

#### (copied from the original PR)
1. Fresh install app. In onboarding toggle on the sending of analytics data on the last page.
2. Perform various actions in the schema deck. You should see something like EventLoggingService: 0x60000337dd20 <x-coredata:///WMFEventRecord/t2B6FC470-29D4-46F2-B143-AE3D8F2D0B4A7> recorded! in the console with each event you trigger.
3. Watch the network via a proxy. You will see dumps every so often of events that start with https://meta.wikimedia.org/beacon/event....
4. Take that event url and decode it through a url decoder. You should see events with the schema value of MobileWikiAppiOSLivingDoc.
#### (new things to look for)
5. Thank someone, tap a View Changes button, View Discussion button, or Small Changes link, and inspect the post that goes out.
6. Confirm `position` is now sent as a 0-based integer (before it started at 1 and was wrapped in quotes).
7. Confirm `type` value has single quotes around each type, e.g. `"type":"'char_added','char_deleted'"`

### DIFF
--- a/Wikipedia/Code/Article as a Living Document/ArticleAsLivingDocViewController.swift
+++ b/Wikipedia/Code/Article as a Living Document/ArticleAsLivingDocViewController.swift
@@ -156,7 +156,7 @@ class ArticleAsLivingDocViewController: ColumnarCollectionViewController {
     }
     
     private func updateLoggingPositionsForItemsInSections(_ sections: [ArticleAsLivingDocViewModel.SectionHeader]) {
-        var currentPosition: Int = 1
+        var currentPosition: Int = 0
         for section in sections {
             for item in section.typedEvents {
                 switch item {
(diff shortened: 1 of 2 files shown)


mcleinman:
This makes sense and code looks good. Will hold off on testing until it's off WIP.

tonisevener:
@mcleinman this is ready for review/testing - Shay was able to confirm that thanks are now coming in.

mcleinman (approved):
Looks good, tests good.

PR #3777, opened 2020-11-20
Adjust alignment of Thanks icon in toolbar

dempseyatgithub:
**Fixes Phabricator ticket:** 
https://phabricator.wikimedia.org/T254832

### Notes
Since the inner custom view needed to be adjusted using a UIButton-specific method, this seemed to be the least intrusive approach.

This approach adds an init method that takes `iconInsets`. This is an Obj-C method, so we can't use a default argument value in the method signature and since UIEdgeInsets is a struct in Obj-C, cannot make the argument an optional.

So, made the existing init call through to the init with the new parameter. This leaves all call sites the same except the one we need to change.

I hard-coded the offer, but if there is a place to put those metrics. I can definitely make the change.

Note the adjustment amount lines the body of the Thanks icon up with the body of the Share icon.

Some other approaches considered:
`(item.customView as? UIButton)?.imageEdgeInsets = UIEdgeInsets(...)` at the call site.

Override `imageInsets` and pass through to get/set `(customView as? UIButton)?.imageEdgeInsets`. Still need to fall back to super in case custom view some how ends up not being a UIButton. Seemed to over-complicate `imageInsets`.

Expose a `var customButton: UIButton?` to hide the casting nastiness at the call site and then use `item.customButton?. imageEdgeInsets = UIImageEdgeInsets(...) at the call site. Seemed to expose a property just for a single use case.

### Test Steps
1. View an article
2. Scroll to View edit history in About this Article
3. Tap a revision history entry
4. Adjusted icon is visible in bottom toolbar 

Thanks-Icon-Before
Thanks-Icon-After


mcleinman (approved):
Code looks great, and this tests great. Thank you!

PR #3781, opened 2020-11-23
Event Platform Client: Isolate sampling controller logic

mdholloway:
A first step toward decomposing the Event Platform Client from a large single class into individual sub-components.

### DIFF
--- a/WMF Framework/Event Platform/EventPlatformClient.swift
+++ b/WMF Framework/Event Platform/EventPlatformClient.swift
@@ -10,7 +10,7 @@
  *     stream intake service.
  *
  * LICENSE NOTICE
- *     Copyright 2019 Wikimedia Foundation
+ *     Copyright 2020 Wikimedia Foundation
  *
  *     Redistribution and use in source and binary forms, with or without
  *     modification, are permitted provided that the following conditions are
@@ -53,7 +53,7 @@ import CocoaLumberjackSwift
  *   originally submitted
  */
 @objc (WMFEventPlatformClient)
-public class EventPlatformClient: NSObject {
+public class EventPlatformClient: NSObject, SamplingControllerDelegate {
     // MARK: - Properties
 
     @objc(sharedInstance) public static let shared: EventPlatformClient = {
@@ -64,6 +64,8 @@ public class EventPlatformClient: NSObject {
     /// Session for requesting data
     let session = MWKDataStore.shared().session
     
+    let samplingController: SamplingController
+
     /**
      * Streams are the event stream identifiers that can be utilized with the EventPlatformClientLibrary. They should
      *  correspond to the `$id` of a schema in
@@ -128,7 +130,7 @@ public class EventPlatformClient: NSObject {
     /**
      * An individual stream's configuration.
      */
-    private struct StreamConfiguration: Codable {
+    struct StreamConfiguration: Codable {
         let sampling: Sampling?
         struct Sampling: Codable {
             let rate: Double?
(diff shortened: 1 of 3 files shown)


WMF Framework/EventPlatformClient/SamplingController.swift
```diff
@@ -0,0 +1,133 @@
...
+     * times `in_sample` simply returns the cached determination.
+     *
+     * Only cache determinations asynchronously via `queue.async`
+     */
+    private var samplingCache: [EPC.Stream: Bool] = [:]
+
+    /**
+     * Compute a boolean function on a random identifier
+     * - Parameter stream: name of the stream
+     * - Returns: `true` if in sample or `false` otherwise
```
tonisevener, line 29: Would be good to add the new config parameter documentation here.

WMF Framework/EventPlatformClient/EventPlatformClient.swift
```diff
@@ -263,9 +255,12 @@ public class EPC: NSObject {
 
         self.streamIntakeServiceURI = streamIntakeServiceURI
         self.streamConfigServiceURI = streamConfigServiceURI
-        
+
         super.init()
 
+        self.samplingController.appInstallId = installID
```
tonisevener, line 261: I'm a bit wary of `SamplingController` having its own copies of these values, since they could easily get out of sync (such as when `resetSession` is called from this class). I think better would be for SamplingController to have EventPlatformClientLibrary as a delegate, then whenever SamplingController needs those identifiers it asks the delegate (I worked something up right quick [here](https://github.com/wikimedia/wikipedia-ios/commit/530075c4b964d1a0d2e1b8405ea4158d50a557ac) if you want to reference it).

tonisevener (changes requested):
Looking good, just one thing about identifiers potentially getting out of sync.

WMF Framework/EventPlatformClient/EventPlatformClient.swift
```diff
@@ -721,7 +630,10 @@ public class EPC: NSObject {
                 return
             }
             
-            guard inSample(stream: stream) else {
+            guard let config = streamConfigs[stream] else {
```
tonisevener, line 633: This check is equivalent to the `guard streamConfigs.keys.contains(stream) else {` check a few lines earlier, but I like your way of writing it better. Can you delete lines 628 - 631 and add the `DDLogDebug("EPC: Event submitted to...` line inside this guard before the `return`?

WMF Framework/EventPlatformClient/EventPlatformClient.swift
mdholloway: Thanks for this.   Done.

WMF Framework/EventPlatformClient/EventPlatformClient.swift
mdholloway: Done

WMF Framework/EventPlatformClient/SamplingController.swift
mdholloway: Done

mdholloway:
OK, I think I've addressed all comments on this one.

mdholloway:
Oh, I also renamed the "EventPlatformClient" group to "Event Platform" for stylistic consistency with the other group names.

tonisevener (approved):
Looks good to me. I'll look tomorrow into the CircleCI failures (looks like something on their end), then I'll merge.

PR #3782, opened 2020-11-23
Ensure alerts go out on main thread

mcleinman:
**Fixes Phabricator ticket:** https://phabricator.wikimedia.org/T268522

### Notes
* The startup code is mostly in obj-c, and it crashes the app when we try to send out an alert.
* This PR ensures all alerts go out on the main thread.
* The "two Saved lists" error (the one used in the bug to trigger this carsh) isn't the situation Sharvani's account was in, but her account was using the same "alert on launch" code, and so I believe this should fix the crash.

### Test Steps
1. Sign into iOS app.
1. Sign into Android app with same account. (If you need Matt to setup an account and do the Android side of this, let me know.)
1. Android: Open any article, tap the bookmark icon to save it. In the in-article dialog, tap the "add to reading list" button. "Create a new reading list", and name it "Saved". (Note: Must be done in-article, as you aren't allowed to create a list called "Saved" in the Android app when creating it from the Reading List screen.)
1. iOS: Hard quit app, re-open it. Ensure app does not crash when trying to sync reading lists, and that it shows an alert about the updated reading list name.


tonisevener (approved):
Looks good to me!

PR #3783, opened 2020-11-24
Disable swipe actions to fix visual issues

dempseyatgithub:
**Fixes Phabricator ticket:** 
https://phabricator.wikimedia.org/T260579

### Notes
- Add a isSwipeEnabled property to SwipeableCell protocol
- Implement with default of true in ArticleCollectionViewCell
- Turn off swiping when save button present in ArticleFullWidthImage… cell
- Turn off swiping in ArticleLocationAuthorizationCollectionViewCell
- Check for isSwipeEnabled in CollectionViewEditController

The most straightforward approach seemed to be adding a property to the `SwipeableCell` protocol to indicate whether swiping was enabled. I had considered using an empty action array as a proxy as to whether a cell should be have swiping enabled, but since it is a relatively complex class hierarchy of cells, it seemed that might introduce unexpected issues and require more testing to validate now and in the future.

Only `ArticleCollectionViewCell` implements `SwipeableCell` and the default value of `isSwipeEnabled` is true. This maintains the current swiping behavior for all subclasses unless `isSwipeEnabled` is overridden. The property is overriden in  `ArticleLocationAuthorizationCollectionViewCell` to always return `false`, and in `ArticleFullWidthImageCollectionViewCell` to return `false` when the save button is visible.

In `CollectionViewEditController`, where `isSwipeEnabled` is checked, I added the check to every entry point where a `SwipeableCell` is required. Specifically, in `configureSwipeableCell()`, in `panGestureRecognizerShouldBegin()` and `handlePanGesture()`. There's a little bit of belt and suspenders going on since the check is in both 'should begin' and the 'handle gesture' methods, but my thinking is that a false `isSwipeEnabled` value essentially turns a SwipeableCell into an 'unswipeable cell' so it makes sense to check. I didn't add the check in `openActionPane()` and `closeActionPane()` since those are called from `handlePanGesture()`.  

### Test Steps
1. Launch app with Location Services permission turned off (or a fresh app install)
2. Swipe trailing-to-leading on a location-based article with 'Enable Location' button
3. No action buttons should slide in.

4. Swipe trailing-to-leading on a Random Article and an Article of the Day card.
5. No action buttons should slide in and the summary text should *not* disappear.


mcleinman (approved):
Looks great, tests great. Thank you!

PR #3784, opened 2020-11-24
Event Platform Client: Consolidate input and output buffers

mdholloway:
Currently, the Event Platform Client persists events in two different
buffers: an "input buffer" used before stream configurations are loaded,
and an "output buffer" to which events are transferred from the input
buffer when stream configurations are loaded and which is used
thereafter.

This patch removes the input buffer as unnecessary and uses a single
event buffer throughout the app lifecycle. We would expect stream
configurations to be loaded very quickly on app startup in most cases.
In the event the periodic worker triggers event submission prior to
stream configurations being loaded, the submission is skipped and the
events remain in buffer.


WMF Framework/EventPlatformClientLibrary.swift
```diff
@@ -698,24 +669,22 @@ public class EPC: NSObject {
...
-            
+
             #if DEBUG
             let jsonString = String(data: jsonData, encoding: .utf8)!
             DDLogDebug("EPC: Scheduling event to be sent to \(streamIntakeServiceURI) with POST body:\n\(jsonString)")
             #endif
-            
-            let request = EventRequest(url: streamIntakeServiceURI, data: jsonData, stream: stream)
+
             guard let streamConfigs = streamConfigurations else {
```
tonisevener, line 684: I think lines 684 - 695 should go too - any event logged before the stream configuration is fetched will get snatched up by these guards and dropped before it can be appended to the buffer below.

You'll need to re-add this config keys check and `inSample` check later before posting though since you're removing it here, since it serves as a filter to prevent unwanted or unrecognized events from posting. Probably in `postAllScheduled` within the `while`.

WMF Framework/EventPlatformClientLibrary.swift
```diff
@@ -190,28 +190,10 @@ public class EPC: NSObject {
...
-     *
-     * Only modify (append events to, remove events from) *synchronously* via
-     * `queue.sync`
-     */
-    private var inputBuffer: [EventRequest] = []
-
-    /**
-     * Maximum number of events allowed in the input buffer
-     */
-    private let inbutBufferLimit = 128
```
tonisevener, line 209: Just wanted to point out that we'll now have no in-memory event limit since the input buffer was the only one that seemed to have a max. Probably moot though since your persistence is coming, we'll just need to be sure there's logic in that to prevent events from piling up indefinitely.

tonisevener (changes requested):
Looking good, feels better having only 1 queue to worry about. I haven't tested to be sure but there's one potential problem I see with events getting dropped before the config fetch comes through.

WMF Framework/EventPlatformClientLibrary.swift
tonisevener: Update - upon re-reading your description, this code dropping the event might have been left on purpose. I still think we ought to try to keep these events though, in case a user launches while completely offline.

WMF Framework/EventPlatformClientLibrary.swift
mdholloway: Hmm, yeah, probably best to leave this limit in place in the meantime.  I'll put it back in.

WMF Framework/EventPlatformClientLibrary.swift
mdholloway: Ah, right.  Yeah, this isn't correct.  Actually, most of this stuff is getting overhauled in the WIP persistence patch I started writing after pushing this anyway, so maybe it's not worth putting more time into consolidating the buffers.  I'll leave this PR open for now, but there's a decent chance I'll end up abandoning it once the persistence PR is up.

mdholloway: Okay, now the two buffers are starting to make more sense to me.  The input buffer contains essentially arbitrary content (since we're not able to evaluate whether to send them yet, without the stream config data) and is therefore capped at 128 events, whereas events in the output buffer have been validated as targeting a configured stream and being in sample, and it is therefore allowed to grow arbitrarily.

I'm starting to think it might be worth keeping the input buffer in place for this reason even with the persistence layer restored.  What do you think, @tonisevener?

tonisevener: @mdholloway yeah, if it's getting overhauled anyways maybe we can evaluate it after your persistence patch. We could do something like have the persisted objects have a `validated` flag. After the config comes through, fetch unvalidated events from Core Data, loop through them and mark the flag or delete the event. Then in `postAllScheduled` only fetch events from Core Data that are validated and post them. In a case like that both buffers / array properties would go away and we'd just be fetching directly from the db. To keep the cap, before saving any unvalidated items to the database, you could delete any older unvalidated ones so that you're always maintaining a max of 128. This may be straying far from what you have already done so far with persistence, just a direction we could go in if we decide to.

tonisevener:
Gonna go ahead and close this one out, since we discussed keeping the separate buffers in the meeting today.

PR #3785, opened 2020-11-24
Update unit testing documentation to include current `en-US` requirem…

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T259859 (does not directly address the ticket's issue, just documents the testing requirements – that ticket should probably remain in place until resolved)


mcleinman (approved):
🙌

PR #3787, opened 2020-11-27
Tap on random article card header now navigates to previewed article

dempseyatgithub:
**Fixes Phabricator ticket:** 
https://phabricator.wikimedia.org/T207708

### Notes
So far all explore cards navigate to the the same destination if the header and footer of the card is tapped.

In this case, tapping above or below the article preview require different destinations.

To detect the tap:
In ExploreCardCollectionViewCell, the cellBackgroundView is now a small subclass of UIView that does hit testing and sets a value when the tap is in the title area of the cell. This just detects the location of the tap and then calls super.hitTest to allow the normal event handling to proceed.

The cell itself is a delegate of the background view. It takes a little more code, but seemed more resilient than `superview?.superview?.titleAreaTapped` to reference the cell.

Putting the hit test override on the cell itself causes the override to be called no matter where in the cell the tap happens. Putting it in the background, it only is called in spots where the background is exposed.

Because of the way the cellBackgroundView is already set up, its bounds are set so the y=0 is at the start of the preview contents of the cell. So checking for y < 0 returns the correct result. I added a comment where the cellBackgroundView metrics are set up to double check the hit test still works if changes are made to the frame/bounds.

Since the cell is available in collectionView(_:didSelectItemAt:) titleAreaTapped is a property on the cell.

In collectionView(_:didSelectItemAt:) the flag `useRandomArticlePreviewItem` is true if titleAreaTapped and if the moreType of the item's group is random article. If it is true, we skip the first method to get the detail view controller, but instead go through the method that asks for the view controller for the preview item.

This seemed to be the least intrusive way of adding this functionality. If different action needs to be taken based on tapping in the title area rather than the footer of a cell, that can be used separately than deciding which view controller to use next.

- Add hit testing to card background to detect tap in title area
- Add titleAreaTapped property to ExploreCardCollectionViewCell
- Add check for this case in ExploreViewController selection method
- Use detail view controller for preview item in this case

### Test Steps
1. Scroll to a random article in the explore feed
2. Tap in the header ("Random article from English Wikipedia")
3. The currently previewed article should display in the detail view
4. Go back to explore feed
5. Tap in the footer ("Another random article")
6. A new random article should display in the detail view


WMF Framework/ExploreCardCollectionViewCell.swift
```diff
@@ -5,23 +5,35 @@ public protocol CardContent {
     func contentHeight(forWidth: CGFloat) -> CGFloat
 }
 
+fileprivate protocol CardBackgroundViewDelegate: AnyObject {
+    var titleAreaTapped: Bool { get set }
+}
+private class CardBackgroundView: UIView {
+    fileprivate weak var delegate: CardBackgroundViewDelegate?
+    override public func hitTest(_ point: CGPoint, with event: UIEvent?) -> UIView? {
+        delegate?.titleAreaTapped = point.y < 0
```
mcleinman, line 14: Could the `0` be changed to something like `frame.maxY` so that it covers the areas above the bottom of the card. 

This is a slight change - In testing, I realized that tapping the margin between the card and the edge of the screen* results in going to a novel random article.

*In real life, a finger is unlikely to hit this. But you never know, and if people are using other types of devices... can't hurt?

WMF Framework/ExploreCardCollectionViewCell.swift
```diff
@@ -5,23 +5,35 @@ public protocol CardContent {
     func contentHeight(forWidth: CGFloat) -> CGFloat
 }
 
+fileprivate protocol CardBackgroundViewDelegate: AnyObject {
```
mcleinman, line 8: My personal style, no need to update if you don't agree: When I'm writing code with nonobvious reasons to exist (what this does is obvious, but why it was written is only really discoverable if you dig into the commit/Phab history), I like to leave a 1 or 2 line comment explaining why it was added. Something to the effect of "Random Explore feed cards navigate to a specific random article when the card is tapped, and a truly random article when the card background (with `Another random article >` text) is tapped. This protocol allows the top of the background to navigate to the specific random article, as a user would expect."

mcleinman (changes requested):
Looks great! One request for a small tweak. Thanks!

WMF Framework/ExploreCardCollectionViewCell.swift
dempseyatgithub: That's a great catch! I hadn't realized the margins weren't part of the content itself.

I have it working, but of course, it's a little more complex since the card needs to report the right offset to the background view when hit testing happens.

I'll look at doing the hit testing override in the cell again. The method will get called more often (even when the content is tapped) but that should not add much overhead, but it cuts down on the complexity communicating to the background view.

WMF Framework/ExploreCardCollectionViewCell.swift
dempseyatgithub: Yes, I prefer comments, but it seemed like the files I was working in didn't have any, so I thought that might be a preference. Will gladly add comments.

mcleinman (approved):
Looks great, thank you!

PR #3788, opened 2020-12-02
Only display gallery images matching `showInGallery` flag.

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T269288

### Notes
Currently, the article gallery controller attempts to display all photos from the media list endpoint for an article, regardless of the value of the `showInGallery` flag from the API response. This leads to unexpected situations, like trying and failing to display media that shouldn't be displayed in the gallery.

### Test Steps

1. Open https://en.wikipedia.org/wiki/Basic_reproduction_number article
2. Tap the image in "Contact rate and infectious period" section
3. With the gallery open, swipe left or right to the next or previous images. Confirm you're not able to because there are no other images in the article.


mcleinman (approved):
Great catch! Looking forward to getting all this image gallery bugs stomped out.

PR #3789, opened 2020-12-03
Show proper navigation bar for references

mcleinman:
**Fixes Phabricator ticket:** https://phabricator.wikimedia.org/T268586

### Notes
* After lots of research, the fix ended up being simple: `ReferenceViewController` is never displayed as a push, and its nav bar should know it is displayed modally. (While not actually a modal presentation, it's the best choice of those options - and naturally causes the proper nav bar behavior.)
* Bug would have snuck in [this PR](https://github.com/wikimedia/wikipedia-ios/pull/3640/files), as we changed how we're assembling the `navigationItem` stack.

### Test Steps
1. Open app to an article, scroll to Reference section. Tap a reference number. App should scroll up to that reference, and have a new reference toolbar and nav bar. 
1. When you tap anywhere on the article (or the x in the top right corner), the nav bar and toolbar should go back to normal.
1. Confirm it works on iPad, iPhone 8, and iPhone X-sized devices.

### Screenshots/Videos 
Screenshot 8Screenshot 7


staykids (commented):
Return of the reference nav bar!

I'm unsure if it's related/a known issue and wanted to check: it seems like the reference toolbar, which allows quickly navigating between references, is no longer functioning.

ref_toolbar

mcleinman:
> Return of the reference nav bar!
> 
> I'm unsure if it's related/a known issue and wanted to check: it seems like the reference toolbar, which allows quickly navigating between references, is no longer functioning.
> 
> ref_toolbar

@staykids That bar should remain unselectable unless there are multiple times a specific reference is used in an article. I just tested it out with reference 8 in `United States`, and it properly moved between the four instances of reference 8 in the article. If you're seeing it broken, could you give an example?

staykids:
@mcleinman Ahh, my bad. I was misinterpreting the utility of those toolbar buttons. 👍🏽

mcleinman:
> @mcleinman Ahh, my bad. I was misinterpreting the utility of those toolbar buttons. 👍🏽

No worries! I figured that might be it, but didn't want to assume. :)

PR #3791, opened 2020-12-04
Remove dependabot warnings

tonisevener:
**Fixes Phabricator ticket:** 
Initial work for https://phabricator.wikimedia.org/T266837

### Notes
We have been a bit stuck on dealing with these Github warnings because the version of [grunt-browserify](https://www.npmjs.com/package/grunt-browserify) hasn't updated it's dependencies in 3 years. From what I gather this package gives you the ability to break up your javascript into clean separate files, reference each other using the Node-esq `require` methods, then when you run `grunt` it combines them into 1 file into the assets folder.

What I did here is remove `grunt-browserify` from our package.json and Gruntfile and manually combined that code as they would have been combined ultimately in the assets folder. Now running the grunt command will simply copy those files from the `www` folder into the `assets` folder.

What we lose with this is the ease of working in separated files for some codemirror text detection logic and the client-side javascript we still run for the article screen (areas are listed in test steps below). We also lose some easy namespacing, so a lot of the methods and variables in these combined files are global now. I feel like generally fewer dependencies are better but giant asterisk that this stuff is definitely not in my wheelhouse. Happy to cancel this PR if anyone feels like this isn't the right path.

Potential followup is to remove this whole grunt setup entirely. We would lose the CSS compression and linting of js files.

TLDNR:

1. Updated `package.json` in the `www` folder to latest available packages.
2. Combined the 9 `codemirror-range-*.js` files in the `www/codemirror` folder into `www/codemirror-range-determination-bundle.js`. 
3. Combined the 4 files (`utilities.js`, `findInPage.js`, `elementLocation.js`, `editTextSelection.js`) in the `www/js` folder into `www/index.js`
4. Added command to include these new combined js files in the copy command section of Gruntfile (along with `about.js`, which only seemed to do a copy before anyways through browserify).
5. Removed browserify stuff from Gruntfile and `package.json`.

### Test Steps
Perform regression on article, editing and the about page. Confirm javascript commands still work which from what I could tell were:
         1. Tapping a TOC item, it scrolls properly to section.
         2. Scrolling in article across multiple sections, TOC item properly highlights
         3. Find and highlight still works
         4. Select text in the article. Tap "Edit" from the context menu. Confirm for the most part text is properly highlighted in the editing flow. (If it doesn't, confirm it does in `main` before marking it as a bug, sometimes it just fails to find the text).
         5. After highlighting text on the editing screen, tap the remove formatting (no sign) button from the keyboard menu. Confirm formatting removes as well as it did before.
         6. Confirm the About app page themes properly.


staykids:
@tonisevener Thanks for jumping on this one! It's def. giving me an opportunity to learn more about Grunt and the Browserify module.

Our custom `/www/js` stuff is limited enough that a single file for it doesn't seem that bad (and great idea retaining the original filenames in the comments to help our future selves debug stuff). I initially felt iffier about the custom CodeMirror concatenating, but since we're using our own fork of the package, and all around it's not something that we update much, realized my initial reaction was irrational. 

I'm still going through it all, but is this kind of your general mentality as well about why this approach feels like it'd reasonably serve our goal of removing that outdated dependency?

tonisevener:
> @tonisevener Thanks for jumping on this one! It's def. giving me an opportunity to learn more about Grunt and the Browserify module.
> 
> Our custom `/www/js` stuff is limited enough that a single file for it doesn't seem that bad (and great idea retaining the original filenames in the comments to help our future selves debug stuff). I initially felt iffier about the custom CodeMirror concatenating, but since we're using our own fork of the package, and all around it's not something that we update much, realized my initial reaction was irrational.
> 
> I'm still going through it all, but is this kind of your general mentality as well about why this approach feels like it'd reasonably serve our goal of removing that outdated dependency?

@staykids Yeah, I'm trying to glean what I can from the setup, and it looks like Carthage pulls our codemirror fork into Carthage/Checkouts/wikipedia-ios-codemirror, and our Gruntfile in www has a command to copy the resources folder from Carthage/Checkouts/wikipedia-ios-codemirror into Wikipedia/assets. That fork seems to only contain the files we really need (the official [codemirror release](https://github.com/codemirror/CodeMirror/releases/tag/5.35.0) contains a lot of extra files, such as syntax highlighting for all sorts of other languages). Any custom changes that we needed to make within any of the existing, out-of-the-box functionality of CodeMirror (like find & replace), we made in the fork repo and updated via Carthage & Grunt. But anything that was entirely additional that CodeMirror never supported, such as seeking out and highlighting text selected from the previous screen's context menu (i.e. all the range determination bundle stuff), was added directly as JS files in this repo. I'm not sure if this helps at all but it helped in my mind to see the separation. That's not to say that a change in the fork won't break the range determination bundle code, just that all changes in this PR reside in here at this layer and the fork is untouched.

Looking again at this, I do think I could separate out the codemirror range determination files again, and link with multiple script tags within `codemirror-index.html` instead of the one. That might help with the feeling that we're moving backwards by creating one large file to develop in. As for `www/index.js`, since it's smaller and it's embedded in a different way via `UtilitiesScript`, I suggest we keep it to the one file.

www/package.json
```diff
@@ -4,9 +4,8 @@
   "repository": "https://gerrit.wikimedia.org/r/apps/ios/wikipedia",
```
mcleinman, line 4: Hygiene, don't need to do this if you're not going to make any other changes: Should we update this to our GH repo?

```suggestion
  "repository": "https://github.com/wikimedia/wikipedia-ios/",
```

mcleinman:
Thanks for taking care of this one!

staykids:
@tonisevener Awesome – thanks for confirming. So far, I've tested these steps a few times on both iPhone and iPad with several different articles. Haven't noticed any regressions that don't also already occur on `main` (one such example that occurs on both branches on iPad: TOC open, find in page, using the arrows to navigate to a found text highlight not in current section, but TOC section doesn't update until user manually scrolls article). 

Going to do one last sweep then approve.

staykids (approved):

PR #3792, opened 2020-12-04
Display multiple languages subtitle in order of user's preferred languages

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T263162

### Notes
Updates the "Customize Explore feed" subtitle for each content type to display the enabled languages in the same order as the user's preferred languages.

### Test Steps
1. Open Settings > My languages.
2. Add a few languages, and observe the order they are in. (e.g. English, Italian, Spanish)
3. Tap "Customize Explore feed".
4. Observe order of languages in the subtitle for each content type cell. Confirm they match the order of languages in step 2. (e.g. English, Italian, Spanish)


mcleinman (approved):
Looks great, tests great. Thanks!

PR #3794, opened 2020-12-07
Remove extension method on NSLocale, replace with more recent API.

dempseyatgithub:

### Notes
When the code was originally written, this API had just been introduced and so couldn't be used immediately.
This PR removes an extension method and simplifies the call sites where the extension was used.

### Test Steps
1. There should be no changes to the behavior to the app.


mcleinman (approved):
Thank you!

PR #3795, opened 2020-12-09
Refactor NSLocale/Locale extensions

dempseyatgithub:

### Notes
- Make internally used constants fileprivate
- Drop wmf_ prefix from fileprivate constants
- Migrate two Obj-C NSLocale tests to LocaleTests.swift
- Remove WMFLocaleTests.m
- Move extensions used by Swift-only code from NSLocale to Locale
- Update call sites to use migrated methods
  - Session.swift
  - TableOfContentsHeader.swift

A number of extension methods were on NSLocale and exposed to Obj-C only to be used in two Obj-C test methods. Moving the tests to Swift made the Obj-C test file unnecessary, allowed extension methods to move to Locale and allowed one method to be removed, since it wrapped an existing extension method that used a default argument value, which could not be exposed in Obj-C directly.

Cleaned up naming and access level.

### Test Steps
1. Run LocaleTests.swift test methods.

There should be no change in behavior in the app with these changes.


mcleinman (approved):
Thanks!

PR #3797, opened 2020-12-11
Some MWKLanguageLinkController cleanup

dempseyatgithub:

### Notes
- Remove unused imports and @class declarations of MWKLanguageLinkController
- Remove two unused method from MWKLanguageLinkController
- Move testing-only method declaration to MWKLanguageLinkController private header

### Test Steps
1. Builds after a clean, none of the removed imports or methods are required to build.
2. MWKLanguageLinkControllerTests pass after the changes.


mcleinman (approved):
Looks great, tests great*. Thanks for handling this cleanup! 

*An article as a living doc test fails, but I see why - and will update that on another PR.

PR #3800, opened 2020-12-11
EPC: Update to edit_history_compare schema v2.0.0

mdholloway:
Basically an administrative change renaming 'client_dt' to 'dt'.

### DIFF
--- a/WMF Framework/EventPlatformClientLibrary.swift
+++ b/WMF Framework/EventPlatformClientLibrary.swift
@@ -47,7 +47,7 @@ import CocoaLumberjackSwift
  *
  * iOS schemas will always include the following fields which are managed by EPC
  * and which will be assigned automatically by the library:
- * - `client_dt`: client-side timestamp of when event was originally submitted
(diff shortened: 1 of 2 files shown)


Wikipedia/Code/EditFunnel.swift
```diff
@@ -96,7 +96,7 @@
             return eventData
         }
 
-        let requiredData: [String: Any] = ["session_token": sessionID, "anon": isAnon, "app_install_id": appInstallID, "client_dt": timestamp]
```
tonisevener, line 99: @mdholloway let's keep this `client_dt` for now - `EditFunnel` still flows only through the old legacy system, only `EditHistoryCompareFunnel` will have the call sites to the new system.

tonisevener (changes requested):
Just one small change needed, thanks!

Wikipedia/Code/EditFunnel.swift
mdholloway: Whoops! Thanks for catching that. Updated.

tonisevener (approved):

PR #3802, opened 2020-12-14
fix crash on picture of the day

mcleinman:
**Fixes Phabricator ticket:** N/A

### Notes
* There were two issues here: One, this wasn't triggering an error. Two, an error in this flow wasn't actually doing anything.

### Test Steps
1. Run app. Tap picture of the day. Ensure you get a nice error message, rather than a crash.

PR #3803, opened 2020-12-15
 fix crash on picture of the day

mcleinman:
**Fixes Phabricator ticket:** https://phabricator.wikimedia.org/T270126

### Notes
* There were two issues here: One, this wasn't triggering an error. Two, an error in this flow wasn't actually doing anything.

### Test Steps
1. Run app. Tap picture of the day. Ensure you get a nice error message, rather than a crash.


staykids (approved):
Nice! Works with today's problem entry and some control entries in the past and future. Also works when invoked via widget.

Great that we have a pre-existing localized error message, and still allowing the POTD to be displayed in the feed (since we have no current issue retrieving the image itself, just some meta data).

tonisevener:
Closing in favor back to #3802.

PR #3805, opened 2020-12-19
Allow iPads to pop into images/articles

mcleinman:
**Fixes Phabricator ticket:** https://phabricator.wikimedia.org/T258109

### Notes
* The animator can have 2 commit styles - `dismiss` or `pop`. The safety check being removed in this PR was added in the [first commit](https://github.com/wikimedia/wikipedia-ios/commit/921c96ed950c15c811c6adedb0d96a0ff007d9ad#diff-b3a4e8778a7c51d6747010167b742d27ffe6d9bc1613fb09070e7a0340f2c9e0) while working on peek-and-pop and PCS, and I can't find any reason for it to exist beyond defensive coding. I believe in all cases we want to allow peeked article elements to pop, so it is safe to remove. (And we don't currently have any other method that accesses the animator. IMO: It's not worth adding another delegate method, but if we already had one that accessed the animator we could have updated the commit style there.)
* I'm guessing the problem doesn't exist on iPhone because the default commit style is likely `pop` on phones.
* The issue is about images, but article previews also don't currently pop. This fixes the problem.

### Test Steps
1. Run on an iPad. Within an article, ensure images and articles can be peeked, then popped.


Wikipedia/Code/ArticleViewController+LinkPreviewing.swift
```diff
@@ -93,7 +93,6 @@ extension ArticleViewController: ArticleContextMenuPresenting, WKUIDelegate {
     func webView(_ webView: WKWebView, contextMenuForElement elementInfo: WKContextMenuElementInfo, willCommitWithAnimator animator: UIContextMenuInteractionCommitAnimating) {
         guard
```
tonisevener, line 94: I noticed the popping is a bit slow on iPad, like it waits a while for the preview to disappear, then it triggers, whereas iPhone seems to pop in place as soon as you tap the preview. But when I add this line:

```
animator.preferredCommitStyle = .pop
```

It seems to pop quickly, like iPhone. I'm not sure why, in fact the [documentation](https://developer.apple.com/documentation/uikit/uicontextmenuinteractioncommitstyle?language=objc) seems to suggest the opposite (dismiss = no animation, pop = animation). Anyways just thought I'd mention it as a consideration, but I'm not sure if this is the right thing to do in this case. At any rate it looks good for approval.

tonisevener (approved):

tonisevener (approved):

PR #3806, opened 2020-12-22
EPC: Add StorageManager

mdholloway:


mdholloway:
This isn't quite fully baked yet, but I wanted to get it pushed up for discussion.  Events are making it into the database but never making it out again. 😅   The faulty conditional in the SQL query seems to be `ZFAILED <> 1` part, which is apparently, sort of counterintuitively, evaluating to false where `ZFAILED` is null.  But the query predicates are literally copy-pasted from EventLoggingService, and seem to be working fine there, so not sure why it's not working here.

tonisevener:
@mdholloway I haven't tested to confirm but in `EventPlatformEvents.xcdatamodeld`, `failed` has no Default Value whereas `EventLogging.xcdatamodeld` has a Default Value of NO. That will probably prevent null values for ZFAILED down in the db table.

Screen Shot 2020-12-22 at 9 47 07 AM

tonisevener:
@mdholloway It also looks like "Optional" is not checked, that might make a difference too.

mdholloway:
Yep, I made some adjustments to the attribute settings this morning and I'm hopeful that will get things fixed up.  Now dealing with a separate build issue, though (`Duplicate interface definition for class 'WMFEPEventRecord'`); looks like it occurred in CI as well, so it's not just a local issue.

tonisevener:
@mdholloway yeah that's more xcdatamodel weirdness, try changing "Codegen" value to "Manual/None"

Screen Shot 2020-12-22 at 10 05 37 AM

mdholloway:
That did the trick, thanks!  🎉

WMF Framework/Event Platform/EventPlatformClient.swift
```diff
@@ -383,26 +353,56 @@ public class EventPlatformClient: NSObject, SamplingControllerDelegate {
...
+            DDLogDebug("EPC: Nothing to send.")
+            return
+        }
+
         let group = DispatchGroup()
-        while let item = outputBufferPopFirst() {
+        for record in records {
             group.enter()
-            httpPost(url: item.url, body: item.data) { result in
+
```
tonisevener, line 370: We'll need to wrap up any Core Data object property accessors (calls to `record.stream` and `record.data` in this method) in a `performAndWait` call from `StorageManager`, to prevent threading corruption. My thinking is make a public method on StorageManager like this:

```
    func streamAndData(record: EPEventRecord) -> (stream: String, data: Data)? {
        var result: (stream: String, data: Data)? = nil
        performAndWait { moc in
            result = (record.stream, data: record.data)
        }
        return result
    }
```

Then here pull these values like so:

```
            guard let streamAndData = storageManager.streamAndData(record: record) else {
                DDLogError("EPC: Unable to extract stream and data from EPEventRecord.")
                storageManager.markFailed(record: record)
                continue
            }
            
            let streamRaw = streamAndData.stream
            let data = streamAndData.data
```

And then use `streamRaw` and `data` in place of `record.stream` and `record.data` below.

I'm pretty sure this will be enough to plug the holes, but we can try adding the [concurrency debug launch argument](https://oleb.net/blog/2014/06/core-data-concurrency-debugging/) while we test to help find any other issues.

WMF Framework/Event Platform/EventPlatformClient.swift
```diff
@@ -383,26 +353,56 @@ public class EventPlatformClient: NSObject, SamplingControllerDelegate {
...
+                /// Continue without marking as failed so we can try again later.
+                DDLogWarn("EPC: Event submitted for stream '\(stream.rawValue)' but only the following streams are configured: \(self.streamConfigurations.keys.map(\.rawValue).joined(separator: ", "))")
+                continue
+            }
+
+            guard samplingController.inSample(stream: stream, config: streamConfig) else {
+                /// Stream is out of sample. Mark as failed so it will be dropped from storage on the next cleanup.
+                DDLogDebug("EPC: Stream '\(stream.rawValue)' is not in sample")
+                storageManager.markFailed(record: record)
+                return
```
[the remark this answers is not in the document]
mdholloway: Yep, fixed.

WMF Framework/Event Platform/EventPlatformClient.swift
```diff
@@ -383,26 +353,56 @@ public class EventPlatformClient: NSObject, SamplingControllerDelegate {
...
+
+            httpPost(url: EventPlatformClient.eventIntakeURI, body: record.data) { result in
                 switch result {
                 case .success:
+                    storageManager.markPosted(record: record, date: Date())
                     break
                 case .failure(let error):
                     switch error {
                     case .networkingLibraryError:
-                        guard item.attempt < 10 else {
```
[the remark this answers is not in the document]
mdholloway: Yeah, I finally got around to looking yesterday and couldn't find anywhere that the old library was actually tracking retry attempts (despite the existence of the DB column), so I just zapped it.

WMF Framework/Event Platform/EventPlatformClient.swift
```diff
@@ -183,39 +184,22 @@ public class EventPlatformClient: NSObject, SamplingControllerDelegate {
...
         super.init()
 
         self.samplingController.delegate = self
 
-        self.fetchStreamConfiguration(retries: 10, retryDelay: 30)
+        if (self.storageManager != nil) {
+            self.fetchStreamConfiguration(retries: 10, retryDelay: 30)
+        } else {
+            DDLogError("EPC: Error initializing the storage manager. Event intake and submission will be disabled.")
+        }
```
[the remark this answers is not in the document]
mdholloway: Done.  Actually, while we're talking about this section: @linehan and I had our weekly tech huddle a couple of hours ago and were thinking about whether the client should be disabled completely in the (unlikely?) event that the storage setup goes sideways, or whether it should continue to attempt to send events through without permanent storage like it's doing now.  Our tentative conclusion was that it would be best to simply fail, as this PR is currently written, in the first instance, and perhaps do some beta testing to see how prevalent storage setup failures are in the wild and whether it turns out that they are prevalent enough to justify adding the complexity of a fallback code path.  What do you think?

WMF Framework/Event Platform/StorageManager.swift
```diff
@@ -0,0 +1,185 @@
+import Foundation
+import CocoaLumberjackSwift
+
+class StorageManager: NSObject {
+
+    private var managedObjectContext: NSManagedObjectContext
```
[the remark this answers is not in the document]
mdholloway: Done

WMF Framework/Event Platform/StorageManager.swift
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    func markFailed(record: EPEventRecord) {
+        perform { moc in
+            record.failed = true
+            self.save(moc)
+        }
+    }
+
+    private init?(storageURL: URL) {
```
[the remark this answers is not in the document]
mdholloway: Wasn't sure whether the convention here would be to keep the initializer up top, or to follow decreasing order of visibility.  Done!

WMF Framework/Event Platform/StorageManager.swift
```diff
@@ -0,0 +1,185 @@
...
+        let modelURL = Bundle.wmf.url(forResource: "EventPlatformEvents", withExtension: "momd")!
+        let psc = NSPersistentStoreCoordinator(managedObjectModel: NSManagedObjectModel(contentsOf: modelURL)!)
```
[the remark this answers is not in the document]
mdholloway: Done

WMF Framework/Event Platform/EventPlatformClient.swift
mdholloway: I was thinking (and Jason and I were talking) about how I wasn't super happy about `EPEventRecord` objects getting returned to the client. This comment makes me even less happy about it!  

Would it work to update StorageManager.popAll() to return (say) `PersistableEvent` objects containing the relevant data values, rather than the raw `EPEventRecord` objects?  Something like:

```
    func popAll() -> [EventPlatformClient.PersistableEvent] {
        var result: [EventPlatformClient.PersistableEvent] = []
        performAndWait { moc in
            let fetch: NSFetchRequest<EPEventRecord> = EPEventRecord.fetchRequest()
            fetch.sortDescriptors = [NSSortDescriptor(keyPath: \EPEventRecord.recorded, ascending: true)]
            fetch.predicate = NSPredicate(format: "(posted == nil) AND (failed != TRUE)")

            do {
                var count = 0
                let records = try moc.fetch(fetch)
                for record in records {
                    guard let stream = EventPlatformClient.Stream(rawValue: record.stream) else {
                        continue
                    }
                    result.append(EventPlatformClient.PersistableEvent(data: record.data, stream: stream))
                    count += 1
                }
                DDLogDebug("EPC: Found \(count) events awaiting submission")
            } catch let error {
                DDLogError(error.localizedDescription)
            }
        }
        return result
    }
```

Then all of our references to the `EPEventRecord` objects stay within the initial `performAndWait` block for the fetch, which (I think/hope) would resolve the threading concerns.  `PersistableEvent` would have to be updated with fields for the rest of the DB columns (which I was thinking of doing anyway) and could also perhaps take an optional `moid` field for when we want to refer back to a DB record for update based on the result of a submission attempt.

WMF Framework/Event Platform/EventPlatformClient.swift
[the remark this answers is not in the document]
tonisevener: @mdholloway sure that sounds good to me, that is, don't add the complexity unless it's prevalent enough. Do you have thoughts on how to measure storage setup failures? Is the plan to hit a logging endpoint if `self.storageManager = nil` to mark an instance of a storage failure, or are we going to assume `edit_history_compare` count differences from the legacy system mean the storage setup failed?

WMF Framework/Event Platform/StorageManager.swift
```diff
@@ -0,0 +1,185 @@
...
+    static let shared: StorageManager? = {
+        let fileManager = FileManager.default
+        var storageDirectory = fileManager.wmf_containerURL().appendingPathComponent("Event Platform", isDirectory: true)
+
+        do {
+            try fileManager.createDirectory(at: storageDirectory, withIntermediateDirectories: true, attributes: nil)
+        } catch {
+            do {
+                let storageDirectoryIsReachable = try storageDirectory.checkResourceIsReachable()
+                if !storageDirectoryIsReachable {
```
linehan, line 18: Hmm, I feel like there are more branches here than there is information or behavior. I assume this is the "always-try-to-create-and-fail-if-exists" pattern, maybe we can make it the "create-if-doesn't-exist" pattern instead? It would be nice if we didn't need nested do blocks.

```swift
do {
  if ! try storageDirectory.checkResourceIsReachable() {
    try fileManager.createDirectory(at: storageDirectory, withIntermediateDirectories: true, attributes: nil)
  }
} catch {
  DDLogError("EPC: Event Platform event storage directory is unreachable and cannot be created.")
  return nil
}
```

WMF Framework/Event Platform/StorageManager.swift
```diff
@@ -0,0 +1,185 @@
+import Foundation
+import CocoaLumberjackSwift
+
+class StorageManager: NSObject {
+
+    private var managedObjectContext: NSManagedObjectContext
+    private let pruningAge: TimeInterval = 60*60*24*30 // 30 days
```
linehan, line 7: Did we pick a magic number here? Should we do some napkin calculations and make sure this makes sense?

WMF Framework/Event Platform/StorageManager.swift
```diff
@@ -0,0 +1,185 @@
...
+                return nil
+            }
+        }
+
+        do {
+            var values = URLResourceValues()
+            values.isExcludedFromBackup = true
+            try storageDirectory.setResourceValues(values)
+        } catch let error {
+            DDLogError("EPC: Error excluding from backup: \(error)")
```
linehan, line 33: 1. What kind of thing would cause setResourceValues to throw?
2. Should we be treating this as a StorageManager instantiation failure?

WMF Framework/Event Platform/EventPlatformClient.swift
tonisevener: @mdholloway yep that sounds good with using `PersistableEvent` like that - I like that it would keep full Core Data records out of `EventLoggingService`. For referring back to the DB record, your `moid` field will work, and for referencing to mark as posted or failed, I suggest using the `existingObject(with: NSManagedObjectID)` method. (https://developer.apple.com/documentation/coredata/nsmanagedobjectcontext/1506686-existingobject). Note it will throw an error if that record has since been deleted which seems possible, vs `object(with: NSManagedObjectID)` which throws an exception.

WMF Framework/Event Platform/EventPlatformClient.swift
```diff
@@ -383,26 +353,56 @@ public class EventPlatformClient: NSObject, SamplingControllerDelegate {
...
-        while let item = outputBufferPopFirst() {
+        for record in records {
             group.enter()
-            httpPost(url: item.url, body: item.data) { result in
+
+            guard let stream = Stream(rawValue: record.stream) else {
+                /// This is an unrecoverable logic error. Mark as failed.
+                DDLogError("EPC: Invalid event stream name '\(record.stream)'")
+                storageManager.markFailed(record: record)
+                continue
```
tonisevener, line 375: You need to add a `group.leave()` before each of these continues - every group.enter() needs to be balanced with one, and these early continues won't reach the one in the `httpPost` completion block.

WMF Framework/Event Platform/StorageManager.swift
```diff
@@ -0,0 +1,185 @@
...
+    private func perform(_ block: @escaping (_ moc: NSManagedObjectContext) -> Void) {
+        let moc = self.managedObjectContext
+        moc.perform {
+            block(moc)
+        }
+    }
+}
+
+extension StorageManager: PeriodicWorker {
+    public func doPeriodicWork(_ completion: @escaping () -> Void) {
```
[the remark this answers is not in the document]
mdholloway: Yeah, I saw earlier that this wasn't firing and added it to the periodicWorkerController in WMFAppViewController.m; just hadn't pushed that update yet.  

The motivation here is that the main EPC class shouldn't be concerned with the storage manager's bookkeeping; all it should need to know how to do are (1) push events onto the queue, (2) pop events from the queue for submission, and (3) update enqueued events with `posted` or `failed` state.  But I suppose I can live with the main client directly calling StorageManager.pruneStaleEvents for now and cleaning up this (and some other implementation detail that's leaking in with the latest changes) at some possible future date.

mdholloway:
By the way, why is the User-Agent string stored with all events in the old client?  I added it here, too, just to match the old client, but can it be removed?

WMF Framework/Event Platform/EventPlatformClient.swift
mdholloway: OK, this is done in the latest version (using `existingObject`). The implementation is still a bit rough and not quite DRY but I think the idea is there at least.

WMF Framework/Event Platform/StorageManager.swift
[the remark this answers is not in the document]
tonisevener: @mdholloway aah I see now. I'm also good with you doing it in `WMFAppViewController.m` instead in that case. Also to quote my earlier comment:

`Otherwise you could be marking a record as failed or posted after it has since been deleted, and I'm not sure that the underlying system would handle that well.`

This chaining might not be necessary once you work on the changes mentioned [here](https://github.com/wikimedia/wikipedia-ios/pull/3806#discussion_r547498524). Even if `markPosted` and `markFailed` runs on a record that has since been pruned, fetching the object via `existingObject` should let you ignore that error rather than crash.

tonisevener:
> By the way, why is the User-Agent string stored with all events in the old client? I added it here, too, just to match the old client, but can it be removed?

Hm good catch - it looks like the old system sets the request header value for "User-Agent". The only reason I can think is we save the user agent as the request is generated and persisted. If they happen to update the app or OS before the event is posted, this way the user agent sent is the old one when the event occured, and not the new one when the event is posted. In `httpPost`, when you're creating the request from `session`, you can pass in the header from the record as an overridden default header:

```
let request = session.request(with: url, method: .post, bodyData: body, bodyEncoding: .json, headers: ["User-Agent": userAgent])
```

WMF Framework/Event Platform/EventPlatformClient.swift
mdholloway: Done.

WMF Framework/Event Platform/StorageManager.swift
mdholloway: I checked the commit adding this (https://github.com/wikimedia/wikipedia-ios/commit/9fe6d9389a3cfe059f9747bfc826e25ff4d00384) and its enclosing PR (https://github.com/wikimedia/wikipedia-ios/pull/1484) and I can't find any specific justification for a 30-day window.  The relevant Phab task indicates that technical planning was being done in Google Docs, but the one I was able to find in my email has been deleted. :(

There is a wiki page here with useful discussion about what they were trying to accomplish, but nothing directly on point: https://www.mediawiki.org/wiki/Wikimedia_Apps/App_Analytics

WMF Framework/Event Platform/EventPlatformClient.swift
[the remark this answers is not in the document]
mdholloway: I didn't have anything in mind beyond the latter.  Would it be possible to determine the effect (if any) based only on the numbers of events coming back from beta testers, or is that pool too small to tell us anything conclusive?

WMF Framework/Event Platform/StorageManager.swift
mdholloway: One thing that occurs to me is that if we're trying to gauge the effect on event numbers of putting DB persistence back in, it would be best to match the old client on implementation details like this one as a control.  

In the long run, it's probably worth creating a Phab task to hash it out with PM input and to document the decision for posterity.

WMF Framework/Event Platform/StorageManager.swift
mdholloway: Yeah, this code is a bit of a mess that I haven't looked back at since I got it working... 

Looking at @tonisevener's earlier implementation here (https://github.com/wikimedia/wikipedia-ios/pull/3648/commits/ab60d394ec4f866921af84f588158271e85e60ce#diff-ce0064ced62c208c2fdcccf81df261f4c1fa7aa8021e9e3d3a405a18a7495121R23-R27) it looks like we may be able to dispense even with the `checkResourceIsReachable` call and just call `createDirectory`; it'll return `true` if the directory already exists and `createIntermediates` is `true`.

https://developer.apple.com/documentation/foundation/filemanager/1407884-createdirectory
https://developer.apple.com/documentation/foundation/url/1779667-checkresourceisreachable

WMF Framework/Event Platform/StorageManager.swift
mdholloway: The docs don't specify exactly what would cause `setResourceValues` to throw (though they specify things that don't 🙃 ).  At any rate, it does seems cleaner to me to keep all of this in a single block devoted to StorageManager instantiation.  I'll update accordingly.

mdholloway: https://developer.apple.com/documentation/foundation/nsurl/1413819-setresourcevalue

WMF Framework/Event Platform/StorageManager.swift
linehan: Thanks for doing all of that digging! Now that I'm pulling the sweater thread a little bit, I still think we aren't quite all on the same page about how our various requirements are interacting, though. 

The output buffer fulfills two basic purposes:
1. It allows events to pool so we can apply bursting and batching strategies.
2. It allows events to survive transient network connectivity loss

The key word here is 'transient', since events will only find their way to the output buffer for reason #2 if they occur
- DURING a loss of network connectivity
- AFTER stream configuration load
- BEFORE app termination

This is because stream config load requires a network connection. That means that the actual TTL is likely to be the time until the user terminates the app for the first time while offline. At that point, no further events will ever make their way to the output buffer until a new stream configuration is fetched.

This behavior is intentional -- both the stream configuration and the input buffer should not be persisted to disk, and therefore should not persist beyond app termination. That means that after app termination, no further events will ever make their way to the output buffer until a new stream configuration is fetched.

My basic point here is that the true TTL is `min(30 days, time to first app termination after network loss)`. I think my suggestion would be to set a size limit rather than a TTL, since there is already an implicit TTL (the time to app termination). This would also protect users against any bugs which could cause an instrument to spray a lot of events and start filling up storage. Certainly after 30 days without phoning home, it would be possible to get into a bad situation.

WMF Framework/Event Platform/EventPlatformClient.swift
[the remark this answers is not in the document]
linehan: @tonisevener There are a couple of paths that could lead the event system to disable itself. At least user opt-out, stream configuration fetch failure, and storage failure, though there may be more. I could imagine all of these going to the same method to disable the system, perhaps this method takes a `reason` argument, but at least stream configuration fetch and storage failure could generate an event and POST it directly to one of our backends (similar to how client error logging is done in MediaWiki). If a POST couldn't happen immediately (we'd expect this if stream configuration fetch failure occurred), perhaps this diagnostic event could be persisted until the next time the device is online.

Something like this was my thought. I don't think we have sufficient information to diagnose a delta between new and legacy as being caused by storage failure otherwise.

WMF Framework/Event Platform/StorageManager.swift
linehan: Sounds good to me. No comment on the fact that a fn called `createDirectory` returns `true` if it fails to create a directory.

WMF Framework/Event Platform/StorageManager.swift
tonisevener: > This behavior is intentional -- both the stream configuration and the input buffer should not be persisted to disk, and therefore should not persist beyond app termination.

@linehan It seems like this will still drop many events if we are only persisting the output buffer. In my mind input buffer events that are made while offline without a stream configuration are valid and should be given a chance to post between app terminations. But I admit I am clouded by the old client library design.

I'm fine with having a pruning be a size limit of the db events table rather than a particular date - just don't understand the reasons for the input / output buffer distinction now that we have persistence. This PR does still have the config analysis / inSample mapping of course, it just occurs later right before posting rather than coordinating between the two separate buffers.

linehan: > > This behavior is intentional -- both the stream configuration and the input buffer should not be persisted to disk, and therefore should not persist beyond app termination.
>
> In my mind input buffer events that are made while offline without a stream configuration are valid and should be given a chance to post between app terminations.

@tonisevener I've drifted a bit here, myself, but the reason we initially included an input buffer was to cover a "blind spot" between app launch and stream config fetch. This is a pattern from MediaWiki that has proven useful. It wasn't intended as a general way to "delay judgement" on events in the absence of stream configuration, except in that small window between load and fetch. We should talk more about this, but here are a couple of properties that result from minimizing the input buffer's role:
- The system will "fail safe" in the absence of stream config. If there is a bug or problem with any instrument or stream, it can be either shut off in stream config; if a device cannot fetch steam config, it will be disabled by default (along with all streams and instruments)
- We can be sure that the budget for events/storage/performance on a given device is an exact reflection of the stream configuration. If input buffer is optimistic about the possibility of one day getting a stream config, our budgets will need to instead reflect the possibility that, no matter what our stream config is, all events from all instruments and all streams will fire and be stored to disk for the TTL.

The summary is that it makes the stream configuration a better reflection of the conditions on a device, which helps us reason about and control various factors better.

WMF Framework/Event Platform/StorageManager.swift
tonisevener: Yeah, that earlier implementation was lifted straight from the old library (complete with the unused `var didGetDirectoryExistsError = false` var 😅). It seems to me like `createDirectory` could serve as both checks in one (I assume it would throw an error if it was unreachable, the documentation is a little hazy). At any rate I'm not aware of any problems we've had in the wild with the old library instantiation.

WMF Framework/Event Platform/StorageManager.swift
```diff
@@ -0,0 +1,220 @@
...
+        perform { moc in
+            defer {
+                completion()
+            }
+
+            let pruneFetch = NSFetchRequest<NSFetchRequestResult>(entityName: "WMFEPEventRecord")
+            pruneFetch.returnsObjectsAsFaults = false
+
+            let pruneDate = Date().addingTimeInterval(-(self.pruningAge)) as NSDate
+            pruneFetch.predicate = NSPredicate(format: "(recorded < %@) OR (posted != nil) OR (failed == TRUE)", pruneDate)
```
linehan, line 155: @mdholloway @tonisevener Are we actually using the distinction between `posted` and `failed` anywhere? Functionally speaking, both of them mean "not eligible for POSTing and need to be pruned". It seems like we could just have a single property called `pruneable` or something. We can replace `mark{Posted,Failed}` with `markPruneable`.

mdholloway: +1, again I was mimicking the older behavior, but I'm all for consolidating this unless there's some reason not to.  @tonisevener thoughts?

tonisevener: @mdholloway @linehan yep agreed, I'm not aware of any need for the distinction and a single flag sounds way better. Nice cleanup catch!

WMF Framework/Event Platform/StorageManager.swift
tonisevener: @linehan Makes sense. I can see the argument both ways. If today's stream config is supposed to dictate all past and present events (i.e. the current config sampling rate of zero should shut down events logged 2 weeks ago), we should scrap the input buffer. Or if you need events posted to represent what the config settings were at the time the event was generated, we should keep the input buffer.

My main concern was that by keeping the input buffer, persistence only catches those whose stream config fetch succeeds but posting fails and the app is terminated (like you said, transient network connection issues, not full offline mode). I would guess that number is far lower than situations where both stream config and posting fails.

mdholloway: I guess I was assuming that we wanted unsent event persistence across app termination and relaunch simply by virtue of the fact that we're discussing DB persistence.  Otherwise, why take on the complexity?  Is there any point?

That said, I think that dropping unsent events at app termination is the right thing to do, and 30 days seems quite excessive as a max TTL.

mdholloway: OK, I can see a justification if we expect the amount of pending event data between bursts and/or transient losses of network connectivity to become larger than we want to hold in memory at a given time. Although if that's the case, we're arguably creating an excessive number of events...

linehan: > My main concern was that by keeping the input buffer, persistence only catches those whose stream config fetch succeeds but posting fails and the app is terminated (like you said, transient network connection issues, not full offline mode). I would guess that number is far lower than situations where both stream config and posting fails.

Yes, so we agree persistence in this situation is more confusing than useful, *if* the other behavior is what I described, I hear what you're saying though, that offline mode is a thing. I don't know how much of a thing it is, but let's say it's enough of a thing.

I'll lay out the full three approaches (at least as far as I can formulate them) rather than just talking about one or the other.

Approach 1 - "The future is the present"
- Don't persist the stream configuration
- Persist the input buffer
- Input buffer events are processed according to the stream configuration that is eventually fetched, regardless of when they occurred.
  - This leads to events in seeming to have been generated prior to their stream configuration allowing it, or with properties out of step with their stream configuration
  - It's possible to solve this, but it's not solved out-of-the-box today

Approach 2 - "The past is the present"
- Persist the most recent stream configuration
- Don't persist the input buffer
- If stream configuration fetch fails, use the persisted one instead; events are processed through to the output buffer.
   - This leads to the same situation in which events may seem to have been generated subsequent to their stream configuration disallowing it, or with properties out of step with their stream configuration.
      - It's possible to solve this, in a way that is more likely to get implemented soon (the stream configurations have a version or modtime, and events get rejected at the server if their event doesn't give the current version/modtime).
   - The other problem here is that this will not "fail safe" if the stream configuration change disables a buggy instrument or stream

Approach 3 - "The present is the present"
- Don't persist the stream configuration
- Don't persist the input buffer
- If stream configuration fetch fails, no events get sent.
  - Events will not be recorded for users that have not fetched a stream configuration. Offline mode, etc. gets hard to instrument.
    - Possible to solve this by having different behavior for 'offline mode' events, or by caching some stream configuration, etc. 
    - Seems easier to evolve from this to a more sophisticated solution (at least to me)

WMF Framework/Event Platform/StorageManager.swift
mdholloway: OK, I'm still going to hold off on re-pushing while I work on updating the input buffer stuff, but for the record, this is done.

mdholloway:
> > By the way, why is the User-Agent string stored with all events in the old client? I added it here, too, just to match the old client, but can it be removed?
> 
> Hm good catch - it looks like the old system sets the request header value for "User-Agent". The only reason I can think is we save the user agent as the request is generated and persisted. If they happen to update the app or OS before the event is posted, this way the user agent sent is the old one when the event occured, and not the new one when the event is posted. In `httpPost`, when you're creating the request from `session`, you can pass in the header from the record to override the default header:
> 
> ```
> let request = session.request(with: url, method: .post, bodyData: body, bodyEncoding: .json, headers: ["User-Agent": userAgent])
> ```

If I recall correctly, the old EventLogging system supplements the event data that ultimately gets stored in the DB with the producer's user-agent.  I don't think that currently happens (yet) for the Event Platform, but I'd have to double-check on that (though maybe @linehan knows off the top of his head).  In any event, is this just being used as a proxy for the app version?  Can we update the relevant schema(s) to explicitly require an app version string instead, rather than continuing to spoof older user-agent headers?

tonisevener:
> If I recall correctly, the old EventLogging system supplements the event data that ultimately gets stored in the DB with the producer's user-agent.  I don't think that currently happens (yet) for the Event Platform, but I'd have to double-check on that.

The `let request = session.request(with: url, method: .post, bodyData: body, bodyEncoding: .json)` line in the new service [populates](https://github.com/wikimedia/wikipedia-ios/blob/b73da75a73490d877e42f5a87ef73d4396ac9583/Wikipedia/Code/Session.swift#L217) the request header with the same value that the legacy service uses to [populate](https://github.com/wikimedia/wikipedia-ios/blob/b73da75a73490d877e42f5a87ef73d4396ac9583/WMF%20Framework/EventLoggingService.swift#L143) the userAgent value on it's records. So I think for the most part the functionality is the same, we just aren't persisting it, and there's a chance the value of user agent has changed in the days since the event was first generated and persisted.

mdholloway:
> > If I recall correctly, the old EventLogging system supplements the event data that ultimately gets stored in the DB with the producer's user-agent.  I don't think that currently happens (yet) for the Event Platform, but I'd have to double-check on that.
> 
> The `let request = session.request(with: url, method: .post, bodyData: body, bodyEncoding: .json)` line in the new service [populates](https://github.com/wikimedia/wikipedia-ios/blob/b73da75a73490d877e42f5a87ef73d4396ac9583/Wikipedia/Code/Session.swift#L217) the request header with the same value that the legacy service uses to [populate](https://github.com/wikimedia/wikipedia-ios/blob/b73da75a73490d877e42f5a87ef73d4396ac9583/WMF%20Framework/EventLoggingService.swift#L143) the userAgent value on it's records. So I think for the most part the functionality is the same, we just aren't persisting it, and there's a chance the value of user agent has changed in the time since the event was first generated and persisted.

Ah, sorry, I was talking about the server-side behavior of the legacy eventlogging service upon receiving an event via the beacon endpoint (see https://github.com/wikimedia/eventlogging#eventcapsule-mediawiki-extension), not the EventLoggingService class in the app.

linehan:
> > > By the way, why is the User-Agent string stored with all events in the old client? I added it here, too, just to match the old client, but can it be removed?
> > 
> > 
> > Hm good catch - it looks like the old system sets the request header value for "User-Agent". The only reason I can think is we save the user agent as the request is generated and persisted. If they happen to update the app or OS before the event is posted, this way the user agent sent is the old one when the event occured, and not the new one when the event is posted. In `httpPost`, when you're creating the request from `session`, you can pass in the header from the record to override the default header:
> > ```
> > let request = session.request(with: url, method: .post, bodyData: body, bodyEncoding: .json, headers: ["User-Agent": userAgent])
> > ```
> 
> If I recall correctly, the old EventLogging system supplements the event data that ultimately gets stored in the DB with the producer's user-agent. I don't think that currently happens (yet) for the Event Platform, but I'd have to double-check on that 

The fields in `http`, specifically `http.request_headers.user_agent`, should be filled out by EventGate based on the HTTP request it receives. So whatever we put in the heade as User-Agent will make it into that field.

> In any event, is this just being used as a proxy for the app version? Can we update the relevant schema(s) to explicitly require an app version string instead, rather than continuing to spoof older user-agent headers?

I was thinking along these lines, except I was thinking to let the client set a `user_agent` string directly rather than needing to fish 
it out of the header. This would also simplify our awkward userAgent state we are needing to tag along with the event the whole way, since it would just be part of the event's data to begin with.

I think the intention of `user_agent` as "the software and version of the originator of this HTTP request" makes sense and has a cross-platform interpretation (it's tied to HTTP as a protocol, rather than to any specific notion of how that user agent is specified), so it's probably not necessary for us to distinguish app version versus browser version, as long as user agent strings are distinct and canonical, it makes sense to me.

mdholloway:
> I was thinking along these lines, except I was thinking to let the client set a `user_agent` string directly rather than needing to fish it out of the header. This would also simplify our awkward userAgent state we are needing to tag along with the event the whole way, since it would just be part of the event's data to begin with.
> 
> I think the intention of a user agent as, roughly, "the software and version of the originator of this HTTP request" makes sense and has a cross-platform interpretation (it's tied to HTTP as a protocol, rather than to any specific notion of how that user agent string is structured), so it's probably not necessary for us to distinguish app version versus browser version versus anything else as long as user agent strings are distinct and canonical, it makes sense to me.

I guess I would want to know how it's being used by PMs and data analysts (or whomever is querying this data).  I would think in general that breaking out the specific data of interest would make queries easier to write and more performant to execute than doing something like `WHERE user_agent LIKE '%v1.3.37%'` or what have you.  But I guess I wouldn't object in principle to having a `user_agent` field in the event data object.  I don't mean to be pedantic but it just seems wrong for the app to represent itself to the intake server as something it isn't anymore in order to set this data with the intended value(s).

linehan:
> 
> I guess I would want to know how it's being used by PMs and data analysts (or whomever is querying this data). I would think in general that breaking out the specific data of interest would make queries easier to write and more performant to execute than doing something like `WHERE user_agent LIKE '%v1.3.37%'` or what have you. 

Aah yeah, that is a good point. I think the practice is for a back-end processor to eat these strings and spit out something categorical like "firefox", "wikipedia-ios", "mobile-safari", etc. This is most often used as a dimension to break down events by browser or (possibly) version.

> I don't mean to be pedantic but it just seems wrong for the app to represent itself to the intake server as something it isn't anymore in order to set this data with the intended value(s).

Being pedantic is totally cool. Is user agent a bad place/way for the app to represent itself, though? I always thought about it as the piece of software (including version) which is the user's agent in the HTTP request. So, `curl`, or `quake`, could be a user agent. That's how I think about it at least, which I think is aligned with [RFC 7231](https://tools.ietf.org/html/rfc7231#section-5.5.3). I don't think it's wrong to separate the user agent from a user agent version, or something, but probably both pieces of data are important for analytics.

IIRC Chrome and some other browser vendors want to move away from having UA strings in their HTTP requests. I don't know if this means that `navigator.userAgent` would also be deprecated.

WMF Framework/Event Platform/StorageManager.swift
mdholloway: OK, I put the input buffer back in to implement Approach 3.

mdholloway:
OK, I zapped the User-Agent from the DB but we can revert that if need be, of course.

I also noticed that the earlier Core Data model for EPC storage was still floating around in the repo, so I added a commit deleting that, too.

I'll remove the Draft tag since I think this is beyond that point now.

WMF Framework/Event Platform/EventPlatformClient.swift
```diff
@@ -383,26 +388,53 @@ public class EventPlatformClient: NSObject, SamplingControllerDelegate {
...
+            }
+
+            guard samplingController.inSample(stream: event.stream, config: streamConfig) else {
+                /// Stream is out of sample. Mark as failed so it will be dropped from storage on the next cleanup.
+                DDLogDebug("EPC: Stream '\(event.stream.rawValue)' is not in sample")
+                storageManager.markPurgeable(event: event)
+                group.leave()
+                continue
+            }
```
tonisevener, line 422: Are these checks needed here? They already go through this filter in `_submit` and `loadStreamConfiguration` before entering the output buffer. Seems like this check would be pruning against the future config rather than keeping the present config pruning that has already happened. That's more of a backend question for y'all - I could see the argument for enforcing a hard drop against the most up-to-date config.

WMF Framework/Event Platform/EventPlatformClient.swift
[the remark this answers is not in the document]
tonisevener: @linehan your error logging all sounds great, should we maybe think about putting that in for a v3?

tonisevener (commented):
One last code review comment from my end. I did some lightweight testing and it's working well!

WMF Framework/Event Platform/EventPlatformClient.swift
mdholloway: Nice catch.  I think you're right, these last-second rechecks are neither necessary nor desirable.  I'll remove them.

tonisevener (approved):
I've been testing this morning and everything looks great. I believe I have marked all the issues I brought up as resolved, leaving the others for @linehan to confirm. Thanks @mdholloway!

WMF Framework/Event Platform/EventPlatformClient.swift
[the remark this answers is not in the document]
linehan: @tonisevener Yeah, let's revisit for v3.

linehan:
Looks good to me, many thanks @mdholloway and @tonisevener, excellent work.

PR #3808, opened 2020-12-22
Refactor and clean up MWKLanguageLink

dempseyatgithub:

### Notes

- Remove unused namespace private property
- Remove unused import and class declarations
- Move convenience methods that depend on Explore Feed content controller
  - Move to fileprivate extensions in the one place they are used: BaseExploreFeedSettingsViewController
- Move isEqual: macro closer to isEqualToLanguageLink: method

The biggest change is removing the dependency on MWKDataStore and the explore feed content controller from the main class definition of MWKLanguageLink. Since the language link is a simple data object, it did not seem appropriate for it to have that dependency.

Since those convenience methods are only used in BaseExploreFeedSettingsViewController, a fileprivate extension in that file keep the convenience without exposing the dependency widely.

I moved the isEqual: macro because in reading the file multiple times, I completely missed its presence.

### Test Steps
No changes in behavior are expected.
Tested explore feed settings.
All relevant unit tests passed.


tonisevener (approved):
Looks good to me! Thanks for the cleanup.

PR #3811, opened 2020-12-23
Dependency updates

mcleinman:

### Notes
* Removes Tweaks, as I don't think we use it anymore. (Notably, we still have exists because the our NYTimesPhotoViewer version uses it.)
* Updates LibrariesUsed info shown in app.
* Next up - updating our forks of a few libraries.

### Test Steps
1. Run app. Ensure it works. 
2. Go to Settings, ensure it works.


Wikipedia/Resources/LibrariesUsed.plist
```diff
@@ -40,25 +40,7 @@ Redistribution and use in source and binary forms, with or without modification,
  * Neither the name of the Bitswift, Inc. nor the names of its contributors may be used to endorse or promote products derived from this software without specific prior written permission.
 
 THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
-SMIT�[LicenseText[LicenseNameZlibextobjc_3**Copyright (c) 2012 - 2014 Justin Spahr-Summers**
```
tonisevener, line 43: I think we are still using some of libextobjc under the WMF Framework > Third Party project group. We should clean that out if we can, if not restore this license in LibrariesUsed.

Wikipedia/Resources/LibrariesUsed.plist
```diff
@@ -152,34 +134,225 @@ DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
...
-LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
-ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
-(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
-SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
-  
  $ + 1 = I N � � � � � �
$07������-4@L^�����
%1=F�����$$$$ $3)r)y)�)�)�             :              /�
\ No newline at end of file
+�12345[LicenseText[LicenseNameYswift-log_,]
```
tonisevener, line 137: Is swift-log used in the project? I couldn't find reference to it.

Wikipedia/assets/codemirror/config/codemirror-config-skr.json
```diff
@@ -0,0 +1,6 @@
+{"extCodeMirrorConfig":{"pluginModules":[],"tagModes":{"ref":"text/mediawiki","pre":"mw-tag-pre","nowiki":"mw-tag-nowiki"},"tags":{"pre":true,"nowiki":true,"gallery":true,"indicator":true,"timeline":true,"hiero":true,"charinsert":true,"ref":true,"references":true,"inputbox":true,"imagemap":true,"source":true,"syntaxhighlight":true,"poem":true,"categorytree":true,"section":true,"score":true,"templatestyles":true,"templatedata":true,"math":true,"ce":true,"chem":true,"graph":true,"maplink":true,"mapframe":true},"doubleUnderscore":[{"__notoc__":"notoc","__nogallery__":"nogallery","__forcetoc__":"forcetoc","__toc__":"toc","__noeditsection__":"noeditsection","__notitleconvert__":"notitleconvert","__notc__":"notitleconvert","__nocontentconvert__":"nocontentconvert","__nocc__":"nocontentconvert"},{"__NEWSECTIONLINK__":"newsectionlink","__NONEWSECTIONLINK__":"nonewsectionlink","__HIDDENCAT__":
```
tonisevener, line 1: I'm curious as to how these were added. It probably only helps - perhaps these are auto-generated and they've been missing since the last time we ran a carthage update?

tonisevener (commented):
Thanks for going over this! I found that my carthage updating would fail (tried by both running the `scripts/carthage_update` script and by running `carthage update --platform iOS` directly in terminal) with the same Xcode 12 carthage issue that I ran into a while back. Might be something different with my machine than yours, or our Xcode versions, but I fixed it for myself in the dependency-updates-suggestions branch ([cda896](https://github.com/wikimedia/wikipedia-ios/commit/cda8969b7c01fe9c5cbd78fcb934b15c925ba6f0) & [9bdc1f](https://github.com/wikimedia/wikipedia-ios/commit/9bdc1f6cf92a49864ac4b0a5aa4d420dfd648e34), if you want to consider pulling those in). It is definitely not a long-term fix though. It looks like Carthage is about to put out a new release that should fix this and avoid the need for these architecture exclusion workarounds ([original issue](https://github.com/Carthage/Carthage/issues/3019) and [issue for releasing](https://github.com/Carthage/Carthage/issues/3102))

Wikipedia.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -0,0 +1,25 @@
+{
```
tonisevener, line 1: I think you can remove this now that we've gotten rid of the outer workspace file.

Wikipedia/Resources/LibrariesUsed.plist
mcleinman: It's a dependency that CocoaLumberjack uses. Wasn't sure if we list those embedded dependencies?

Wikipedia.xcworkspace/xcshareddata/swiftpm/Package.resolved
mcleinman: I tried to update this branch to `main`, but think I made things worse. Going to close this out, and open a new PR. Anyhow, this comment should be addressed by recreating this off `main`.

Wikipedia/Resources/LibrariesUsed.plist
mcleinman: Good catch!

Wikipedia/assets/codemirror/config/codemirror-config-skr.json
mcleinman: You're correct - this happened when updating carthage.

Wikipedia/Resources/LibrariesUsed.plist
tonisevener: Ah I see - yeah I think that's fine, though now that we're here I wonder if we should also list our web dependencies. If we go down that path the list of embedded dependencies there is extensive. Might be worth syncing with @jtminor on it.

mcleinman:
Replaced by https://github.com/wikimedia/wikipedia-ios/pull/3827

PR #3813, opened 2020-12-23
MWKLanguageLinkController Refactoring

dempseyatgithub:

#### • Add `primarySiteURL` convenience property to MWKDataStore and update call sites

The path `dataStore.lanaguageLinkController.appLanguage?.siteURL` is used throughout the app. This change makes each call site `dataStore.primarySiteURL` which is much more concise and better indicates what is happening. It also begins to clarify the many API with 'siteURL' in the name. Note that MWKDataStore already has a `loginSiteURL` property that returns this same information, so it seemed reasonable to add the property and make the call sites more clear.
 
#### • Make MWKLanguageLink siteURL and articleURL true Obj-C properties and update call sites

Just a small cleanup where the call sites look very weird in Swift since they come across as functions.

#### • Move NSLocale processing method to NSLocale extension

The goal is to consolidate methods that only depend on NSLocale in the NSLocale extension  

#### • Rename readPreferredLanguageCodesWithoutOSPreferredLanguages
Moving the locale manipulation method to an NSLocale extension made the with/without naming distinction less useful. The shorter name `-readSavedPreferredLanguageCodes` still conveys that this reads only the saved value.

### Test Steps
1. No changes in app behavior.
2. All test pass.

### DIFF
--- a/FeaturedArticleWidget/FeaturedArticleWidget.swift
+++ b/FeaturedArticleWidget/FeaturedArticleWidget.swift
@@ -87,7 +87,7 @@ class FeaturedArticleWidget: ExtensionViewController, NCWidgetProviding {
     func widgetPerformUpdate(completionHandler: @escaping (NCUpdateResult) -> Void) {
         WidgetController.shared.startWidgetUpdateTask(completionHandler) { (dataStore, completion) in
             let moc = dataStore.viewContext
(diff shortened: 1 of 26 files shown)


mcleinman (approved):
Code looks great, and it tests well. Thanks for doing this cleanup!

PR #3815, opened 2020-12-29

translatewiki:
Translation updates


mcleinman:
This is failing due to the new N'Ko string:
`"aaald-characters-text-description" = "{{PLURAL:|ߛߓߍߘߋ߲߫=߀$1 ߟߎ߬|ߛߓߍߘߋ߲|ߛߓߍߘߋ߲ ߠߎ߬}}";`

While imperceptible, the 0 in $10 is actually an [N'Ko zero](https://unicode-table.com/en/07C0/). I'm updating this upstream, and the next time twn submits a patch it should work fine.

PR #3817, opened 2021-01-04
Add and use user info to WMFPreferredLanguagesDidChangeNotification

dempseyatgithub:

### Notes

Currently two MWKLanguageLinkController properties expose the prior state of preferred languages to allow WMFAppViewController to calculate if a reorder, add, or remove has happened and to which language.

All of this info is available without calculation when the notification is created. It also seems like a break in encapsulation to keep prior internal state hanging around so an external client can perform calculations on it.

This change removes those methods and adds the appropriate info to the user info and updates the client method to use the user info.

- Add two user info keys for WMFPreferredLanguagesDidChangeNotification
- Add enum of change types for WMFPreferredLanguagesDidChangeNotification
- Remove mostRecentlyModifiedPreferredLanguage property
- Remove previousPreferredLanguages
- Update WMFAppController method to use the user info data
- Remove unused testing methods / declarations

### Test Steps
1. Unit tests passed
2. Stepped through code for each case (add/remove/reorder) to ensure same behavior.
3. After adding and removing languages, the appropriate language should appear/disappear from Explore Feed settings.


mcleinman (approved):
Much cleaner. Thank you!

PR #3819, opened 2021-01-05
Cleanup logging

mcleinman:
**Fixes Phabricator ticket:** https://phabricator.wikimedia.org/T260724

### Notes
* This cuts down on noise, moves everything (except testing logging) to CocoaLumberjack, and improves logging format.

### Test Steps
1. Run app. Make sure log lines look decent. Feel free to make comments if you want improvements on how logging looks.

### Screenshots
Screenshot 4


Wikipedia/Code/ArticleViewController.swift
```diff
@@ -130,7 +130,9 @@ class ArticleViewController: ViewController, HintPresenting {
     }()
     
     lazy var webView: WKWebView = {
-        return WMFWebView(frame: view.bounds, configuration: webViewConfiguration)
+        let webView = WMFWebView(frame: view.bounds, configuration: webViewConfiguration)
+        view.addSubview(webView)
```
mcleinman, line 134: This one stops some log spamming - the view wasn't being added to the superview, causing some warnings.

Wikipedia/Code/BasicLogger.swift
```diff
@@ -0,0 +1,7 @@
+@objc class BasicLogger: NSObject, DDLogger {
+    func log(message logMessage: DDLogMessage) {
+        print(logFormatter?.format(message: logMessage) ?? logMessage.message)
```
mcleinman, line 3: Couldn't find a way to knock off the gross start of lines without doing it this way. Sample project also included that NSLog-ish start of line, until I did something like this.

mcleinman:
Made changes from our discussion in engineering sync.

staykids:
Nice! So great starting to get this stuff under control and provide actual value to us rather than just being noise. Thanks again for undertaking this. Some thoughts below:

1. I think we should just use the emoji and strip out the log level names. I understand your point about wanting new contributors to understand what they mean, but I think for us it definitely just becomes extra noise, and for new contributors who become regular contributors it also becomes noise. But to support both us and new contributors, what do you think about taking out the log level names and: 1. documenting those log levels on the [README](main/README.md) or some other Markdown file in the repo or 2. just documenting it inline as comments in that `formatLogMessage` function or 3. both?

2. What do you think about these possible formats? Along with removing the log level name, it also removes the "from: " from the file details. Some sense of time in the logs I think is of extreme value by default, but if I'm in the minority seeing that time value, totally ok with me compact being our default and I can just use the more verbose Release format myself.

```
Release 
{Level Emoji} {Medium date style, Short time style}: {Message} [{Filename}#L{Line Number}]
ℹ️ Jan 1, 2021 at 10:00 am: Did change authorization status 0. [LocationManager#L213]

Default 
{Level Emoji} {HH:mm a}: {Message} [{Filename}#L{Line Number}]
ℹ️ 10:00 am: Did change authorization status 0. [LocationManager#L213]

Compact 
{Level Emoji} {Message} [{Filename}#L{Line Number}]
ℹ️ Did change authorization status 0. [LocationManager#L213]
```

3. I wonder if we have a more apt emoji for the Debug level. I personally agree that 🪲 reads as bug more than 🐛, but looking at some of the existing stuff logged using that level, nothing really connotes "this is a bug" to me. I don't have an alternate, just surfacing that 🪲/🐛 makes me think the log element is notating details about a bug, when in actuality it's just general info or debugging info.

4. Related: maybe filed away for future us, but is there a compelling reason to have both debug and info levels? It doesn't really seem like we're conceptually separating their uses (see `LocationManager` as an example of this). Maybe it'll make more sense in a future where we _do_ conceptually separate what is an info level log and what is a debug level log but right now the lack of distinction is odd.

5. Slightly extreme: what do you think about going all the way removing logging cruft that isn't created by us and setting the `OS_ACTIVITY_MODE` environment variable to `disable` to mute OS log noise (which unfortunately is more messy). It's extreme in that it probably dumps some useful OS level logging. Adding it in steps like here at least allows _some_ flexibility:

OS_Logging

Would be ideal if we could somehow control this is code. I dunno, actually maybe we shouldn't do it, and if we don't do it, retaining something like `File:` (in `ℹ️: Did change authorization status 0. [File: LocationManager#L213]`) still lets us filter out the OS garbage by filtering the Xcode Console by `File:`.

mcleinman:
>     1. I think we should just use the emoji and strip out the log level names. I understand your point about wanting new contributors to understand what they mean, but I think for us it definitely just becomes extra noise, and for new contributors who become regular contributors it also becomes noise. But to support both us and new contributors, what do you think about taking out the log level names and: 1. documenting those log levels on the [README](main/README.md) or some other Markdown file in the repo or 2. just documenting it inline as comments in that `formatLogMessage` function or 3. both?

It seems like the tradeoff is improved ease-of-use vs. 6 extra characters?  It seems like ease-of-use is helpful, though I'm happy to be outvoted on this. I'd think even moderately experienced folks w/ our codebase may forget the difference between info/debug (updated below)/verbose, and so it's not just brand new developers. I'd also see ease-of-use not just as "can you go find the info in a readme" but "how often do I get confused when getting set up in this new codebase, and need to go find an answer"... and so even if the information is available somewhere else, having to go find it makes it a less friendly code base. All that said - I clearly thing it's a worth the tradeoff to have it. Shall we have Toni be a tiebreaker for this one? 

> 
>     2. What do you think about these possible formats? Along with removing the log level name, it also removes the "from: " from the file details. Some sense of time in the logs I think is of extreme value by default, but if I'm in the minority seeing that time value, totally ok with me compact being our default and I can just use the more verbose Release format myself.

I'm open to adding as many options as we want. It seems that the easiest way to switch this is with the DEBUG flag, no? So there will be a default for debug builds, a default for release builds, and then you'd have to manually switch to another one? That feels kinda clunky. what do you think?

As for time stamp specifics. Dd you pick those styles intentionally? I'm a pretty big fan of full timestamps (down to milliseconds) in logs. I usually find that if timing matters when looking at some logs, it really matters. At least for the release one, it feels like we might as well include the entire thing?

I've updated the end of the lines.
> 
>     3. I wonder if we have a more apt emoji for the Debug level. I personally agree that 🪲 reads as bug more than 🐛, but looking at some of the existing stuff logged using that level, nothing really connotes "this is a bug" to me. I don't have an alternate, just surfacing that 🪲/🐛 makes me think the log element is notating details about a bug, when in actuality it's just general info or debugging info.

You're 100% correct - the DEBUG log level is more detailed than INFO, but less than VERBOSE. It's not supposed to connotate a bug - that's what ERROR (or WARN) is for. I updated this level to `💬`. Let me know if you think of something better.
 
>     4. Related: maybe filed away for future us, but is there a compelling reason to have both debug and info levels? It doesn't really seem like we're conceptually separating their uses (see `LocationManager` as an example of this). Maybe it'll make more sense in a future where we _do_ conceptually separate what is an info level log and what is a debug level log but right now the lack of distinction is odd.

And we're hardly even using Verbose! We don't define these levels, they come by default. So I'm not sure it makes sense for us to combine the two levels in our output. We could prepend them identically if we wanted to, but that seems to be a tough one - because changing `ddLogLevel` should be all that changes what gets logged. If for some reason we put both Info and Debug into our Info path, but then set `ddLogLevel` as `info`, we'd lose some of what we had been logging as `info` (as it was actually Debug under the hood.)

>     5. Slightly extreme: what do you think about going all the way removing logging cruft that isn't created by us and setting the `OS_ACTIVITY_MODE` environment variable to `disable` to mute OS log noise (which unfortunately is more messy). It's extreme in that it probably dumps some useful OS level logging. Adding it in steps like here at least allows _some_ flexibility:

Very good call. I played around with this earlier in my exploration, but didn't come back to after I rewrote how it's printing the log statements. I pulled it for now. The framework-level logging is probably rarely relevant to bugs we're finding. We can always add it back in, but for now I think this keeps it all cleaner w/o needing to filter on any of our default text. Thanks. 💯

tonisevener:
This is looking great! I pushed a commit reducing some of the chattier mobile-html article saving debug logs that I made a while back. I removed some success logs and changed debug logs in error cases to, appropriately, DDLogError.

> Related: maybe filed away for future us, but is there a compelling reason to have both debug and info levels?

Yeah I think both is a bit odd too and could use some rules establishing when to use what. My initial thoughts are DebugLog should only be logs we add to help chase down a very _particular_ bug. InfoLog are longer-lived logs not related to fixing a specific bug. Then we could go in and clean out DebugLogs every now and then, they are not meant to stick around forever. Just initial thoughts though, I haven't done any digging to confirm that makes sense in practice.

> We can always add it back in, but for now I think this keeps it all cleaner w/o needing to filter on any of our default text.

One bummer I've noticed since `OS_ACTIVITY_MODE` was disabled is it seems like NSLayoutConstraints console errors are suppressed now. I would also like to know if Javascript exception errors are suppressed from the web views too (didn't test but that might be the case). Personally I would rather fully delete some Info & Debug logs if it helps make room for these two things.

mcleinman:
My todos:
- Cut down to emoji
- Date formatting - full for release, short time style w/ milliseconds for a debug
- turn OS_ACTIVITY_MODE back on
- Update readme docs w/ log levels and emojis

mcleinman:
This should be ready for another look.

There is a lot of log spam associated with the webviews and networking - enough that it's pretty tough to try navigating articles unless you filter on `#L`. I fixed some of the easier ones in this PR, but I believe we're entering a somewhat unknown territory in trying to fix these, from what internet searches could tell me. Ultimately, with that log spam there, this PR doesn't go as far as I'd have hoped. But we got a bunch of spam out, and can lock on `#L` it ignore (the sometimes useful) system-level logging, as needed.

Screenshot 1
Screenshot 2

staykids (approved):
Looks great. I hope you don't mind, just pushed a small README change, rewriting "warn" as "warning".

PR #3822, opened 2021-01-06
Language link support for variants

dempseyatgithub:

There are three commits in the PR. It may be easiest to review each commit in turn.

### Notes
**Commit 1: Extend MWKLanguageLink for language variants** 
Essentially adding a property for the variant and a computed property to return the variant if present, the language code if no variant. Including all the required isEqual:, hash updates, and adding a test.

Note I did not add a convenience init method that assumes a null variant. It seemed better to have anyone creating a MWKLanguageLink see and consider the languageVariantCode value.

**Commit 2: Add language variants to allLanguages** 
The language variants are read in using a similar mechanism as the main languages. Because we need to replace the main language with its corresponding variants, the json file is keyed by the main language.

Note that the two sets of variants in this commit are Chinese and Serbian, both variants are currently supported by the current app language variant mechanism.

Replacing the main language with its corresponding variants is factored into its own method because it will also be used to allow the user to choose the language of an article.

There is a flag in WikipediaLookup to turn variants on or off. It is set to off in this diff, so app behavior should be unchanged.

**Commit 3: Support variants in preferred languages** 
This updates MWKLanguageLinkController to use contentLanguageCode instead of languageCode, so that language variants can be uniquely identified in saving preferred languages.

### Test Steps
1. App behavior should be unchanged since the variants are not read in by default.
2. All test should continue to pass.
3. Compiling with flag turned on, variants should appear in language choices and changes to preferred language settings should persist, even if one or more language variants are in the list.


Wikipedia/Code/WikipediaLookup.swift
```diff
@@ -26,8 +26,30 @@ import CocoaLumberjackSwift
...
+    // Flag to be removed once language variants feature can be permanently turned on
+    private static let languageVariantsEnabled = false
+    @objc static let allLanguageVariantsByWikipediaLanguageCode: [String:[MWKLanguageLink]] = {
+        guard languageVariantsEnabled else { return [:] }
+        guard let languagesFileURL = Bundle.wmf.url(forResource: "wikipedia-language-variants", withExtension: "json") else {
+            return [:]
+        }
+        do {
+            let data = try Data(contentsOf: languagesFileURL)
+            let entries =  try JSONDecoder().decode([String : [WikipediaLanguageVariant]].self, from: data)
```
mcleinman, line 42: Super, super small - but there is an extra space after the `=`, I believe.

Wikipedia/Code/wikipedia-language-variants.json
```diff
@@ -0,0 +1,66 @@
+{
```
mcleinman, line 1: I think it's appropriate that this is manual (given how rarely these are updated), but we should probably update [this document](https://github.com/wikimedia/wikipedia-ios/blob/main/docs/languages.md) to mention that language variants are manually updated here.

mcleinman (changes requested):
Looks great, tests great. For sake of keeping up with documentation, would it be possible to update the file mentioned in a comment below? Thanks!

It's really, really awesome to see this work coming together. I got a giant smile when I selected a language variant! :)

Wikipedia/Code/WikipediaLookup.swift
dempseyatgithub: Done in the new commit.

Wikipedia/Code/wikipedia-language-variants.json
dempseyatgithub: Updated the document with details.

mcleinman (approved):
So wonderful! Thank you!

PR #3823, opened 2021-01-06
Explore feed content controller refactor

dempseyatgithub:

### Notes
The PR has two commits. The first is adding an assume non-null macro to the header file and removing many individual annotations.

The second commit is the meat of the refactoring.

WMFExploreFeedContentController was using MWKLanguageLinkController's `preferredSiteURLs` (itself a transformation of `preferredLanguages`) in two different ways.

First, a siteURLs property was being set with this value externally at controller creation time and to update content sources when preferred languages changed in the app.

The `setSiteURL:` setter method updated content sources as a side effect of setting the value. The value itself was used as part of the update, but not otherwise used in the class.

Second, an internally-defined `preferredSiteURLs` property called MWKLanguageLinkController's `preferredSiteURLs` and returned those values as an NSSet. These values were just iterated over, there is no practical use in having them be a set rather than an array. (MWKLanguageLinkController already enforces uniqueness of preferred languages.)

The internally-defined `preferredSiteURLs` property (and generally the use of this class) requires the dataSource of the controller be non-nil.

So, some changes:
- Pass in required MWKDataStore in init method
- Make dataStore property private
- Replace setSiteURLs side effects with updateContentSources method
- Change return type of preferredSiteURLs property from NSSet to NSArray
- Remove siteURLs property and call preferredSiteURLs instead
- Update call sites to use updated interfaces

This ensures the WMFExploreFeedContentController always has its required data store. It makes external client's calls to update content sources explicit instead of a hidden side effect of `setSiteURLs`. It also encapsulates WMFExploreFeedContentController's dependency on MWKLanguageLinkController, so the public interface of the feed content controller is out of the 'language business'.

### Test Steps
1. App should behave exactly as before
2. All tests should pass


mcleinman (approved):
Looks great, tests great! These PRs are really, really well-sized. Thank you!

PR #3827, opened 2021-01-14
Dependency updates 

mcleinman:

### Notes
* Removes Tweaks, as I don't think we use it anymore. (Notably, we still have exists because the our NYTimesPhotoViewer version uses it.) Removing it also removes some localization strings.
* Updates LibrariesUsed info shown in app.
* Updates carthage dependencies

### Test Steps
1. Run app. Ensure it works.
2. Go to Settings, ensure it works.


mcleinman:
This replaces https://github.com/wikimedia/wikipedia-ios/pull/3811.

I believe the main outstanding question: Is "librariesused" supposed to be "all libraries we include and the libraries they include" or just "the libraries we include"? https://github.com/wikimedia/wikipedia-ios/pull/3811#discussion_r554139158

tonisevener (approved):
Working well for me - as for the dependencies question, I'll bring it up in sync and we can make an additional PR if necessary.

PR #3828, opened 2021-01-14
Fix announcement modal bugs

tonisevener:
**Fixes Phabricator ticket:** 
Fundraising bug: https://phabricator.wikimedia.org/T242347
Investigation spike (see comments for notes): https://phabricator.wikimedia.org/T247976

### Notes
This PR fixes a few things that stood out as odd to me and could be the reason our announcements (both survey and fundraising) are still displaying on the article view.
1. As discussed in engineering sync, I simplified the logic in `ArticleViewController+Announcements.swift`. Now instead of relying solely on the content group's `isVisible` flag and UserDefault's `shouldCheckForArticleAnnouncements` flag being properly set, we are pulling all visible announcement content groups and checking the end time immediately before displaying. I suspect there was some sort of race condition where this code is being hit **before** the `isVisible` and `shouldCheckForArticleAnnouncements` flags are set from `WMFAnnouncementContentSource.m`. I could repro a fundraising banner displaying after the end time upon deep linking to an article, which supports that theory.
2. I noticed in offline mode, the fetcher would still return a cached announcements response from the last time it fetched announcements. This despite the fact that we are asking session to [ignoreCache](https://github.com/wikimedia/wikipedia-ios/blob/fe864305e128ceadb67da94fc37f2122d91eac91/WMF%20Framework/WMFAnnouncementsFetcher.m#L21. Digging in it looks like some of our caching code added from the mobile-html release in Session.swift takes over, so I have added a `prefersPersistentCacheOverError = false` to fix this. This could be one possible explanation for why the survey announcement is still being displayed despite us purposefully not saving to the database and checking against the end date before presentation.
3. I also noticed, after a `WMFContentGroup` announcement's end date is reached and it's visiblity flag is flipped, it still hangs around in the database. It's probably harmless since all of our presentation logic now checks for isVisible and the proper end date, but I still see no reason to keep these old announcements in the database past the end date. I added a housekeeper task to clean these out, which is triggered when the user backgrounds the app.

#### Test fundraising modal
1. Map locally via proxy some announcements response with fundraising that ends sometime in the near future.
2. Terminate from simulator/device and launch a 2nd time to confirm announcements are fetched.
3. Terminate, stop mapping locally, change simulator/device date to a date after the announcement end date. 
4. Relaunch the app via an article deeplink (one not in the survey campaign, like https://en.wikipedia.org/wiki/Cat). Fundraising modal should not show on article user deep linked to. In `main` it does.

#### Test survey modal
1. Map locally via proxy some announcements with a survey that ends on sometime in the near future. 
2. Upon first launch, save an article in the campaign list. 
3. Terminate from simulator/device and launch a 2nd time to confirm announcements are fetched.
4. Terminate, stop mapping locally, and enter offline mode on device. Launch again and go to your saved article.
5. Survey modal should not show after `displayDelay` seconds from announcement. In `main` it does.

#### Do some light regression testing to confirm dismissal flags are remembered
1. Have locally mapped announcements that end in the near future, both fundraising and survey.
2. Go to 2nd launch so that announcements are fetched.
3. Go to article in campaign list, confirm you see survey modal after `displayDelay` seconds.
4. Dismiss modal, leave the article and terminate. Relaunch and return to the article. You should now see the fundraising modal. Dismiss modal, leave article and terminate. Relaunch and return to the article, you should now see no modals.

#### Inspect SQLLite db to confirm housekeeping works
1. Map locally via proxy some announcements response with fundraising that ends sometime in the near future.
2. After 2nd launch and announcements are fetched, inspect db. Note you see an announcement type `ZWMFCCONTENTGROUP` (you'll see "announcement" as a part of the ZKEY field) with ZISVISIBLE = 1.
3. Background, check db again. Confirm it's unchanged (announcement content group still there, zisvisible still there.
4. Change device/simulator date to after the announcement end date and relaunch. Check the db again. Confirm ZISVISIBLE now equals 0. Background the app. Check db. Confirm announcement content group row is now gone (cleaned out by housekeeper).

### DIFF
--- a/Wikipedia/Code/ArticleViewController+Announcements.swift
+++ b/Wikipedia/Code/ArticleViewController+Announcements.swift
@@ -9,20 +9,19 @@ extension ArticleViewController {
         }
         let predicate = NSPredicate(format: "placement == 'article' && isVisible == YES")
         let contentGroups = dataStore.viewContext.orderedGroups(of: .announcement, with: predicate)
+        let currentDate = Date()
         
         //get the first content group with a valid date
         let contentGroup = contentGroups?.first(where: { (group) -> Bool in
-            let currentDate = Date()
             guard group.contentType == .announcement,
                   let announcement = group.contentPreview as? WMFAnnouncement,
                   let startDate = announcement.startTime,
-                  let endDate = announcement.endTime,
-                  (startDate...endDate).contains(currentDate)
+                  let endDate = announcement.endTime
                   else {
                 return false
             }
             
-            return true
+            return (startDate...endDate).contains(currentDate)
         })
         
         guard
--- a/Wikipedia/Code/ArticleViewController+Announcements.swift
+++ b/Wikipedia/Code/ArticleViewController+Announcements.swift
@@ -4,22 +4,35 @@ import CocoaLumberjackSwift
 extension ArticleViewController {
     
     func showAnnouncementIfNeeded() {
-        guard UserDefaults.standard.shouldCheckForArticleAnnouncements,
-              ((isInValidSurveyCampaignAndArticleList && userHasSeenSurveyPrompt) || !isInValidSurveyCampaignAndArticleList) else {
+        guard ((isInValidSurveyCampaignAndArticleList && userHasSeenSurveyPrompt) || !isInValidSurveyCampaignAndArticleList) else {
             return
         }
-        let predicate = NSPredicate(format: "placement == 'article'")
-        let contentGroup = dataStore.viewContext.newestVisibleGroup(of: .announcement, with: predicate)
+        let predicate = NSPredicate(format: "placement == 'article' && isVisible == YES")
+        let contentGroups = dataStore.viewContext.orderedGroups(of: .announcement, with: predicate)
+        
+        //get the first content group with a valid date
+        let contentGroup = contentGroups?.first(where: { (group) -> Bool in
+            let currentDate = Date()
+            guard group.contentType == .announcement,
+                  let announcement = group.contentPreview as? WMFAnnouncement,
+                  let startDate = announcement.startTime,
+                  let endDate = announcement.endTime,
+                  (startDate...endDate).contains(currentDate)
+                  else {
+                return false
+            }
+            
+            return true
+        })
+        
         guard
             let contentGroupURL = contentGroup?.url,
             let announcement = contentGroup?.contentPreview as? WMFAnnouncement,
             let actionURL = announcement.actionURL
         else {
-            UserDefaults.standard.shouldCheckForArticleAnnouncements = false
             return
         }
         let dismiss = {
-            UserDefaults.standard.shouldCheckForArticleAnnouncements = false
             // re-fetch since time has elapsed
             let contentGroup = self.dataStore.viewContext.contentGroup(for: contentGroupURL)
             contentGroup?.markDismissed()
--- a/Wikipedia/Code/WMFDatabaseHouseKeeper.swift
+++ b/Wikipedia/Code/WMFDatabaseHouseKeeper.swift
@@ -14,11 +14,11 @@ import Foundation
     }
     
     private func deleteStaleAnnouncements(_ moc: NSManagedObjectContext) throws {
-        guard let hiddenAnnouncementContentGroups = moc.orderedGroups(of: .announcement, with: nil) else {
+        guard let announcementContentGroups = moc.orderedGroups(of: .announcement, with: nil) else {
             return
         }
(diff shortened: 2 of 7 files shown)


Wikipedia/Code/ArticleViewController+Announcements.swift
```diff
@@ -4,22 +4,35 @@ import CocoaLumberjackSwift
...
             return
         }
-        let predicate = NSPredicate(format: "placement == 'article'")
-        let contentGroup = dataStore.viewContext.newestVisibleGroup(of: .announcement, with: predicate)
+        let predicate = NSPredicate(format: "placement == 'article' && isVisible == YES")
+        let contentGroups = dataStore.viewContext.orderedGroups(of: .announcement, with: predicate)
+        
+        //get the first content group with a valid date
+        let contentGroup = contentGroups?.first(where: { (group) -> Bool in
+            let currentDate = Date()
```
mcleinman, line 15: Move this a couple lines up for performance?

Wikipedia/Code/ArticleViewController+Announcements.swift
```diff
@@ -4,22 +4,35 @@ import CocoaLumberjackSwift
...
+            guard group.contentType == .announcement,
+                  let announcement = group.contentPreview as? WMFAnnouncement,
+                  let startDate = announcement.startTime,
+                  let endDate = announcement.endTime,
+                  (startDate...endDate).contains(currentDate)
+                  else {
+                return false
+            }
+            
+            return true
```
mcleinman, line 25: return `(startDate...endDate).contains(currentDate)` and remove it from the guard statement?

mcleinman:
Per conversation elsewhere, I've reassigned this to Deepak as I'm not around tomorrow. Feel free to ignore any of my comments you don't agree with. Thanks for this!

Wikipedia/Code/ArticleViewController+Announcements.swift
tonisevener: Good call, this is done!

Wikipedia/Code/ArticleViewController+Announcements.swift
tonisevener: Sounds good, this is done too.

staykids:
@tonisevener Do you happen to have an `/announcements` response you'd recommend using for the test steps here?

tonisevener:
@staykids sure, try this. I moved the fundraising and survey campaign dates to 1/20/2021, feel free to change them as-needed.

[announcements.txt](https://github.com/wikimedia/wikipedia-ios/files/5822548/announcements.txt)

Wikipedia/Code/WMFDatabaseHouseKeeper.swift
```diff
@@ -8,9 +8,30 @@ import Foundation
         let urls = try deleteStaleUnreferencedArticles(moc, navigationStateController: navigationStateController)
 
         try deleteStaleTalkPages(moc)
+        try deleteStaleAnnouncements(moc)
 
         return urls
     }
+    
+    private func deleteStaleAnnouncements(_ moc: NSManagedObjectContext) throws {
+        let predicate = NSPredicate(format: "isVisible == NO")
```
staykids, line 17: Feel free to ignore if I'm misunderstanding, but is this particular check for `isVisible == NO` necessary? If the goal is to clean up "stale" no longer relevant announcements, doesn't just using all the announcements in the for loop guard below, removing any announcement that has an end date before the current date, cover what we're trying to accomplish (whether or not it's visible)?

I only bring this up because if there's some unexpected `isVisible` toggling logic elsewhere, it would potentially prevent the stale announcements in the upcoming for loop from being purged.

staykids (approved):
🕵🏽 love your investigation here, and I think the solutions you've listed are all great to help harden against us running into this again.

Just one question below, and still testing, but went ahead and approved if it's at all holding you up on spinning up the beta - so feel free to merge as needed.

Wikipedia/Code/WMFDatabaseHouseKeeper.swift
tonisevener: @staykids hm yeah that is a fair point - I think my reasoning was to be a little more light touch with this cleanup since it was more of a last-minute change. But you're right that if we can't rely on `isVisible` to be entirely accurate, maybe we should remove that consideration from this cleanup. I'll push an update with that fix and will do some testing on my end with it.

tonisevener:
@staykids My testing still looked good after making your suggestion. Going to go ahead and merge this.

PR #3830, opened 2021-01-15
Variant-aware feed settings

dempseyatgithub:

There are four commits in the PR. It may be easiest to review each commit in turn.

### Notes
The overall purpose of this PR is for the very flexible and intricate preferences for the explore feed to support language variants.

Currently, these preferences are stored keyed by site URL string. (i.e. https://zh.wikipedia.org). This PR changes this to be stored by language variant code (i.e. en, zh-hans). The language variant code is now being used to store preferred languages and uniquely identifies a language / language variant.

**Commit 1: Add wmf_contentLanguageCode computed property to NSURL/URL**
Mirrors the behavior of MWKLanguageLink to provide the content language code.
Should be useful in general, but in this case it reduces the amount of change necessary in the WMFExploreFeedContentController by allowing it to continue to operate on site URLs.
Added tests as well.

**Commit 2: Have MWKLanguageLink URL properties include languageVariantCode**
When siteURL or articleURL is called on a MWKLanguageLink instance, make sure the languageVariantCode is set on the returned URL.
Added test as well.

**Commit 3: Add variant property to Wikipedia model entities**
Add a new Wikipedia.xcdatamodel version that adds a _variant_ property to WMFContentGroup and WMFArticle. This includes adding the appropriate header and implementation file code for the new properties. The use of the name _variant_ as opposed to _contentLanguageCode_ is for consistency for the existing _variant_ property on the CacheItem entity in the Cache model.

Update URL computed properties to propagate language variant code. Note that methods that create the WMFArticle do not yet set the language variant code. This was to keep this PR focused on the changes needed for making the feed preferences location-aware. The added property in WMFArticle was to try to keep model migrations/versions to a minimum.

Addition of these properties is handled by Core Data lightweight migration and the persistent store coordinator is already setup to perform lightweight migrations if needed.

**Commit 4: Make explore feed preferences language-variant aware**
The bulk of this commit is using the wmf_contentLanguageCode as the language key in the preferences instead of the wmf_databaseKey (which is basically the siteURL as a string).

There is code added to update the preferences from the old keys to the new keys. _Note that this code is a bootstrapped placeholder until the eventual 'move to variants' initial migration is in place._

Finally, much of this commit are updating method and variable names to accurately reflect the use of contentLanguageCode.

**Commit 5 & 6: Test additions**
Add more to an existing test and add a comment to clarify a test.

### Test Steps
1. With language variants turned off (the current default) there should be no visible change in the behavior of explore feed settings or the app at all.

2. With language variants turned on in the WikipediaLookup.swift file, and language variants chosen as preferred languages, all explore settings should display and work correctly, even with multiple variants from the same language.

To test, I used Serbian, Cyrillic and Serbian, Latin separated by another language. Putting a different language between the variants makes it easier to tell that a new set of explore cards for a different variant is present.

NOTE: The *explore feed settings* will work. The explore feed itself should show cards based on the settings, but the variant fetched and displayed will not be correct yet.


WMF Framework/WMFExploreFeedContentController.m
```diff
@@ -448,7 +448,11 @@ - (NSDictionary *)exploreFeedPreferencesInManagedObjectContext:(NSManagedObjectC
     WMFKeyValue *keyValue = [moc wmf_keyValueForKey:WMFExploreFeedPreferencesKey];
     NSDictionary *exploreFeedPreferences = (NSDictionary *)keyValue.value;
     if (exploreFeedPreferences && [exploreFeedPreferences objectForKey:WMFExploreFeedPreferencesGlobalCardsKey]) {
-        return exploreFeedPreferences;
+        if ([self exploreFeedPreferencesNeedMigration:exploreFeedPreferences]) {
```
tonisevener, line 451: I know you mention this is placeholder so no need to address it now, but for a manual-ish migration like this (i.e. nothing handled automatically by Core Data), we like to kick it off from `MWKDataStore.m`'s `performUpdatesFromLibraryVersion` method. We already have set up some migrations for our legacy event logging, so you could tack it on after the `migrateShareUsageAndInstallIDToUserDefaults` [line](https://github.com/wikimedia/wikipedia-ios/blob/b733d672810fda8a55e366919ca66011d433e57e/Wikipedia/Code/MWKDataStore.m#L442).

tonisevener (approved):
Looking good! I noted one thing about migration placement that isn't a must for this PR. I also notice if I test with both Chinese Traditional and Chinese Simplified turned on, I would have expected some doubled up cards (say, two Chinese featured article cards for today, that look identical because as you point out the variant fetched and displayed will not be correct yet). But I could be misreading your test steps and notes. Just wanted to point it out in case you need to discuss or follow up on it. Still going to approve and merge to unblock you. Excited to see this stuff come together!

WMF Framework/WMFExploreFeedContentController.m
dempseyatgithub: Thanks Toni! I've copied this and added to my running notes for the migration piece.

dempseyatgithub:
Thanks Toni. I will see if I can figure out why you aren't seeing two sets of Chinese variant cards.

PR #3832, opened 2021-01-19
Variant-aware Mantle classes

dempseyatgithub:

There are six commits in the PR. It may be easiest to review each commit in turn.

### Notes

The Mantle framework is used to convert JSON responses into model objects in the app. Each may have other model objects or arrays of other model objects as properties. Each model object may also have properties of type NSURL.

When the JSON is fetched, the correct language variant should be used in the URL request. However, the language variant also needs to propagate to the URL properties of the graph of the deserialized Mantle objects.

This PR implements the infrastructure for doing so in the WMK-framework specific subclass WMFMTLModel.

**Commit 1: Add infrastructure to propagate variant code through Mantle objects**

The class `WMFMTLModel` adds a method `-propagateLanguageVariantCode:` as well as two class properties. One returns an array of property keys that are URL values. The other returns an array of property keys that should be drilled down into as subelements, or arrays of subelements.

The `-propagateLanguageVariantCode:` and its supporting private methods set the languageVariantCode on all URLs specified in the class's array of URL property keys and recursively drills down to all subelements defined in the class's array of subelement property keys.

This approach mirrors the key-based approach that Mantle already uses.

An `-initWithLanguageVariantCode:` method allows creation and propagation of new instances.

The changes include category methods on `MTLJSONAdapter` to wrap its factory methods and pass in the language variant code.

Since the team desires to move away from Mantle eventually, it didn't seem to make sense to create new category .h/.m files for `MTLJSONAdapter`. Also, it seemed to make sense to put all of the related Mantle language-variant code in one place.

This commit is just the core of the feature. 

**Commit 2: Add variant support to WMFLegacySerializer**
A number of the Mantle classes are created using `WMFLegacySerializer`.

While updating, discovered only one `WMFLegacySerializer` method is used publicly, one method is only used privately, and one is no longer used at all. So adjusted accordingly. (I realize I could have merged the now-private method with the existing public method, but was trying to keep logic changes to minimum.)

Updated the `WMFLegacySerializer` factory methods to take a language variant code and called through to the new `MTLJSONAdapter` methods.

Updated the call sites of the `WMFLegacySerializer` factory method to pass in the language variant code. Considered leaving the non-variant versions, but thought since we are able to require the call site to at least consider passing in a variant, that was a better choice.

**Commit 3: Call variant-aware MTLJSONAdapter factor methods**
First I needed to accept a number of clang linter changes that are unrelated in order to make this commit.

The other changes are updating call sites to use the new factory methods.

**Commit 4: Add URL and subelement property keys to WMFMTLModel classes**
This adds the URL and subelement property keys to the model classes. Note that not all WMFMTLModel subclasses were done, only those created by the updated factory methods.

Note also that all URL properties are included, even those that may not absolutely need the language variant, like an image URL. It seemed a good rule of thumb that all URLs in all subelements have the variant propagated.

Finally, I added a comment in cases where the superclass's empty array of keys was being used so it was clear the lack of an overridden value was not an oversight.

If keys do need to be changed or additional models need this support, it can be easily adjusted or added in a subsequent PR.

**Commits 5 & 6: Add tests**
The existing `WMFSearchResultsSerializationTests.m` file had two small tests already using `MTLJSONAdapter`. Since Mantle is being moved away from at some point, adding another file didn't seem to make sense, so renamed the file and test class `WMFMTLModelSerializationTests`.

Added two test fixtures, JSON for a WMFFeedDayResponse and a WMFFeedOnThisDayEvent.

Implemented three tests to propagate the variant. Also test propagating nil.

### Test Steps
These changes should cause no changes in the behavior of the app.
The added tests should all pass.


mcleinman (approved):
This code looks wonderful and tests great! Thank you!

PR #3837, opened 2021-01-25
URL creation method language variant propagation

dempseyatgithub:

### Notes
Conceptually, this PR does a few things:
- Converts methods that accepted a host name and returned URLComponents to methods that accept a URL and return a URL. The language variant code of the incoming URL is propagated to the returned URL.
- Updates call sites for these updated methods
- Adds a convenience method to NSURLComponents/URLComponents to return a URL from the components with the provided language variant code.
- Propagates the language variant code in all NSURL instance category methods that return an NSURL

I've broken it down into separate commits, so reviewing each commit should be pretty straightforward.

### Test Steps
1. All app tests should pass
2. No change in behavior of app


mcleinman (approved):
Looks good, tests good. Thank you!

PR #3838, opened 2021-01-26
Make WMFContentGroup and WMFArticle language variant-aware

dempseyatgithub:

Each commit is self-contained it is probably easiest to review them individually.

### Notes
**Commit 1**
In WMFContentGroup URL methods, propagates the language variant to the created URL.

In cases where an argument needed to be added, only the language variant is passed in, not a source URL. This is to indicate that those methods do not require a source URL, only the language variant.

Call sites are updated where necessary.

**Commit 2**
Update appropriate WMFContentGroup fetch methods to include the language variant as part of the predicate.

**Commit 3**
Clean up how the variant value is assigned in WMFContentGroup. Ensure URL, siteURL, and articleURL properties propagate the variant. Assert that any set URL value has the same variant value as the content group.

**Commit 4**
Add a class to serve as a compound key in the NSCache article cache in the MWKDataStore. Add a test to ensure it works as expected as a key.

**Commit 5**
Update WMFArticle and MWKDataStore to include the variant in fetches and use the new compound key for NSCache operations.

Note a few non-language variant-aware methods are left as temporary shims. Once those subsystems are variant-aware, the shim methods will be removed.

**Commit 6**
Refactor of opportunity. The deleted MWKDataStore methods are unused.

**Commit 7**
The 'contentPreview' property of WMFContentGroup and 'object' property of WMFContent are archived NSURLs, WMFMTLModel subclasses, or collections of those types. Since archived NSURLs have no notion of a language variant, the variant must be propagated to the unarchived objects. This is done in -awakeFromFetch so it only occurs for fetched objects once after fetch.

**Commit 8**
This commit makes sure Random articles do not appear as blank in the explore feed if you turn on language variants. It is a temporary shim until ArticleSummary is made variant-aware. This only has an effect if language variants are turned on.

### Test Steps
1. With variants turned off (current default) there should be no changes in app behavior.
2. With variants turned on, if you set multiple variants for the same language, you should see a set of explore feed cards for each variant. Content of cards should be visible on first launch and after quit and restart.
NOTE: The contents of the cards will not yet reflect the proper variant.

### DIFF
--- a/WMF Framework/WMFArticle+Extensions.h
+++ b/WMF Framework/WMFArticle+Extensions.h
@@ -63,13 +63,15 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (nullable WMFArticle *)fetchArticleWithURL:(nullable NSURL *)articleURL;
 
-- (nullable WMFArticle *)fetchArticleWithKey:(nullable NSString *)key;
+- (nullable WMFArticle *)fetchArticleWithKey:(nullable NSString *)key variant:(nullable NSString *)variant;
 
-- (nullable NSArray<WMFArticle *> *)fetchArticlesWithKey:(nullable NSString *)key error:(NSError **)error;
+- (nullable NSArray<WMFArticle *> *)fetchArticlesWithKey:(nullable NSString *)key variant:(nullable NSString *)variant error:(NSError **)error;
+- (nullable NSArray<WMFArticle *> *)fetchArticlesWithKey:(nullable NSString *)key error:(NSError **)error; // Temporary shim for ArticleSummary that is not yet variant-aware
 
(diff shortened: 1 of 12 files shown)


tonisevener (approved):
Looks great! Thanks for breaking these commits down and documenting them - makes it much easier to review.

I noticed one odd thing with the Explore feed UI when I tap a Chinese card's "..." link and select "Hide this card". Sometimes scrolling a bit after the card collapse causes the UI to muck up. Notably I couldn't repro on the other languages, but this isn't the first time we've seen weirdness on this screen with UI ([this](https://phabricator.wikimedia.org/T261024) is a similar bug that's in this release). Anyways just an FYI, doesn't have to be looked into now and is maybe just that T261024 bug manifesting.

Screen Shot 2021-01-26 at 6 32 11 PM

PR #3841, opened 2021-01-27
Enable Accept-Language header for language variants

dempseyatgithub:

Each commit is self-contained it is probably easiest to review them individually.

### Notes
**Commit 1**
One liner in Session to use the language variant of the request URL in the Accept-Languages header. Otherwise, uses the standard  Accept-Languages header.

**Commit 2**
Most of the complexity of this PR is in this commit. Two things are happening:
1. In Configuration, updating a method that creates the Accept-Languages header for page content service requests. It now takes a URL instead of a language string and uses the language variant code if present as the value of the returned Accept-Languages header. The two call sites in ArticleFetcher are updated accordingly.

2. The urlRequest() method in ArticleFetcher has a language argument that exists only to pass through the value to the updated configuration method. This commit removes the language argument and update call sites. In ArticleCacheDBWriter, updating the call sites required setting the language variant on the passed-in URL. 

**Commit 3**
A one liner in ArticleFetcher to propagate the language variant code to a created URL. Considered moving this to its own URL extension method, but since it is only used once, that seemed unnecessary.

### Test Steps
1. No change in behavior if language variants are turned off. One exception is that saved articles in Chinese and Serbian may display the wrong variant. (I have not tested this extensively since it is a temporary issue). This is temporary and will be resolved once ArticleCacheDBWriter / ArticleCache is fully variant-aware.

2. If language variants are turned on, explore feed card should display the correct variant and tapping through to the article should display the article in the correct variant.

Known issue: When language variants are turned on, returning to the explore feed after navigating to the article will result in all cards for the same language displaying one variant. (Looking at that next)


WMF Framework/Configuration.swift
```diff
@@ -182,17 +182,17 @@ public class Configuration: NSObject {
     }
     
     /// Returns the default request headers for Page Content Service API requests
-    public func pageContentServiceHeaders(for wikipediaLanguage: String? = nil) -> [String: String] {
-        guard let apiLanguage = wikipediaLanguage else {
-            return [:]
-        }
+    public func pageContentServiceHeaders(for url: URL) -> [String: String] {
```
tonisevener, line 185: I don't think we need this method anymore now that `Session`'s default header attempts to assign a single variant at first. This was added at the time because the backend needed a single variant rather than `Session`'s exhaustive list, otherwise it would send a response with mixed characters.

tonisevener (approved):
So cool seeing this work end-to-end! Just one minor comment on a method you could possibly cleanout. I'll approve anyways and wait a little bit before merging in case you want to do that change.

WMF Framework/Configuration.swift
dempseyatgithub: I should have mentioned that in the PR review. I have that on a list of things to look at after I get the bigger features in the app working. For now, I had wanted to leave it in, as one less behavior change among all the other changes. But I also think by the end of the feature this method will be gone, I just want to take the time to test and double check before removing it.

PR #3842, opened 2021-01-28
Remove Tweaks from additional targets

tonisevener:
There were a couple of other references to Tweaks in the build settings that caused the Experimental build to fail. This fixes that.

Test steps:

Choose the Experimental target, select Product > Archive. Confirm it archives without failure.


mcleinman (approved):
Thanks for catching this, apologies!

PR #3844, opened 2021-01-29
Improve POTD size in explore feed

tonisevener:
**Fixes Phabricator ticket:** https://phabricator.wikimedia.org/T164642

### Notes
* We added extra logic in https://github.com/wikimedia/wikipedia-ios/pull/3716/files to choose a clearer POTD url for displaying in the iOS14 widget. This ticket uses that same helper method to choose a better image URL for the explore feed card.

### Test Steps
1. Map locally the https://en.wikipedia.org/api/rest_v1/feed/featured/2021/xx/xx endpoint (depending on which day you're testing) with the response at https://en.wikipedia.org/api/rest_v1/feed/featured/2017/05/04. That will allow you to directly test the image that the original bug was filed against.
2. General POTD explore feed regression testing without mapping locally.

Before:
IMG_0219

After:
IMG_0220

### DIFF
--- a/WMF Framework/WMFContentGroup+Display.swift
+++ b/WMF Framework/WMFContentGroup+Display.swift
@@ -10,14 +10,27 @@ extension WMFFeedDisplayType {
 }
 
 extension WMFContentGroup {
-    public func imageURLsCompatibleWithTraitCollection(_ traitCollection: UITraitCollection, dataStore: MWKDataStore) -> Set<URL>? {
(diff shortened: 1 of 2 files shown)


staykids (approved):
Nice! Offers a _pinch_ higher quality. The only thing I wonder is if this increase in image bucket size will make any kind of noticeable difference for folks on slower connections. But in my basic iPhone/iPad testing so far works fine.

PR #3846, opened 2021-02-01

translatewiki:
Translation updates


mcleinman:
I fixed the problem string upstream on translatewiki. I'm just going to wait for the next automatic push from the translation repo, as we don't have a release that will be built before we get another push.

PR #3849, opened 2021-02-02
Language Variant Support To View Article In Another Language

dempseyatgithub:

### Notes
This PR enables variants in in the Change Language feature when viewing an article.

NOTE: After posting the PR, added two small commits that ensure tapping links in an article navigate to new article in same language variant.

**Commit 1 & 2**
The first two commits make small changes in support of enabling this feature.

**Commit 3**
The third commit is the biggest change in the PR:
For a given article URL, an array of language link instances is returned for each language that has a corresponding article.

To support language variants this array needs to be processed to do two things:
If a language that supports variants is in the returned array, the article language link for that language needs to be replaced by one article language link per variant. This allows the user to choose which variant to view the article in.

If the original article URL being viewed is a language with a variant, an article language link for the remaining variants in that language needs to be added to the list. This allows the user to read the same article in a different variant.
 
This is added as a single-method category on MWKLanguageLinkController. I made it a separate category to emphasize that it is a very specialized method. The implementation uses two private methods to do the processing.

**Commit 4**
Accepting some formatting changes from the clang formatter.

**Commit 5**
Three changes to put the feature into action:
In MWKTitleLanguageController.m, use the language link results processing method and use contentLanguageCode for matching.
I wasn't convinced about the thread safety of the array processing method, so I do that on the main thread. It should not affect responsiveness.

In WMFAppViewController.m, switch to the in-memory URL key to check if a new URL needs to be loaded. Without this, the code will not navigate from one variant to another in the same language.

**Commit 6**
In URL extension instance methods that vend URLs, propagate the language variant code to the newly created URL. (This was already done for NSURL category methods.)

**Commit 7**
Use the in-memory key when checking if handling the same link.

### Test Steps
1. With variants turned off, no change in behavior
2. With variants turned on, viewing an article like Zebra and tapping the Change Language button in the toolbar:
 - Instead of 'Chinese' and 'Serbian' all of the language variants for those languages should be available.
 - Selecting a language variant will load the article in that variant
 - Tapping the Change Language button again should:
 - Include all of the *other* variants of that language in the list of choices
 - Tapping another variant of the same language should load the article in the selected variant
3. With variants turned on, tapping a link in an article displayed in a language variant will load the tapped article using the same language variant.


tonisevener (commented):
Working great for me - just one thing that I had a question about, but it doesn't seem to affect the UI for me so it might be fine.

tonisevener (approved):

PR #3852, opened 2021-02-05
Remove MWLanguageInfo

dempseyatgithub:

Each commit is probably best reviewed in isolation as each is meant to be an atomic change.

### Notes
This PR removes the deprecated MWLanguageInfo class and moves its functionality to class methods on MWKLanguageLinkController.

The PR also recasts the API in terms of content language code, which includes language variants.

The language variants of Kazakh, as stated in https://phabricator.wikimedia.org/T268275, are Latin, Cyrillic, and Arabic. Since there is at least one case where layout direction may be different among variants, it seemed that being positioned to support that was necessary.
(Note that Kazakh language variants have not been added to the app yet.)

**Commit 1**
Move class methods from MWLanguageInfo to MWKLanguageLinkController. Largely updating call sites and changing #include statements. There are a lot of files changed in this commit, but the changes are mechanical.

**Commit 2**
The WMFPreviewSectionLanguageInfoDelegate protocol relies on MWLanguageInfo, but is not used at all. Remove it.

**Commit 3**
Add class methods to MWKLanguageLinkController that cover two of the three main use cases for MWLanguageInfo objects. (The third is already present in the method that returns the semantic content attribute)

**Commit 4**
Remove the class method that returns an MWLanguageInfo. At two call sites, replace with calls to new 'layoutDirection' method. At one call site, add an assert to check that old and seemingly unused behavior is not relied upon.

The private -codeForCode: method took a language code. If the string passed in was "test" or "simple" instead of a language code, the language code "en" would be returned. Both call sites that moved to the 'layoutDirection' method get their language parsed from a URL, so those special cases are never valid. The call site in the NSLocale extension is called from many different place. I added the assert so it would surface any places where this is still being used (although unlikely).

**Commit 5**
The public use of the +rtlLanguages method was always to see if a language code was a RTL language. Move to new method that answers that question explicitly and make the method private.

**Commit 6**
Update the 'semanticContentAttribute' name to match the other two methods. Update call sites.
At a couple of call sites, this method was being called with the result checked to see if a language was RTL. Moved these call sites to the method that returns this directly.

**Commit 7**
Remove MWLanguageInfo files and remaining references from the project. Update the category name and comments in the MWKLanguageLinkController. Kept these methods in a category because they do have a specialized purpose.

**Commit 8**
Added a comment indicating that if language variants are added to the set of RTL languages, the call sites of the category methods need to be updated to ensure that content language code and not just language code is passed in.

This does not need to happen to support Chinese variants or Serbian variants, so this does not need to happen immediately.

### Test Steps
1. App behavior should remain unchanged


Wikipedia/Code/NSLocale+WMFExtras.swift
```diff
@@ -120,21 +120,25 @@ extension NSLocale {
...
         }
         
-        let languageInfo = MWLanguageInfo(forCode: language)
-        let code = languageInfo.code
+        // Previous code took the wikipediaLanguage value and called a method
+        // that would return the same language code that was passed in
+        // unless 'simple' or 'test' was passed in instead of a language code
+        // in which case "en" would be returned. This asserion double checks that
+        // that old substituion is no longer being relied upon.
+        assert(language != "simple" && language != "test")
```
tonisevener, line 128: I see this assertion getting hit a couple of times when I follow these steps:

1. Add "Simple English" wiki to languages via app settings.
2. Assertion failure is hit whenever Explore feed tries to load. Although if you don't reach this you probably will reach it when digging into the article's edit history on an article in Simple.
3. Add "Test" wiki to languages via app settings.
4. Search for any article in Test. Go to article. Go down to footer and tap "View edit history".

tonisevener (commented):
All this cleanup looks good to me, thanks! I was able to hit that new assertion failure a couple of times though, I added test steps below. Test Wikipedia is just a test environment, set up in production as if it were a language that we use to test editing. Seems fine if we wanted that to rely on `Locale.autoupdatingCurrent` but Simple Wikipedia is a real wiki that should probably have the "en" override.

Wikipedia/Code/NSLocale+WMFExtras.swift
dempseyatgithub: I did not know there was a Simple English version of Wikipedia.

I've updated and removed the assert and added a new conditional that uses the "en" locale if the incoming language is "simple" or "test".

This way, once the locales are cached once, we don't need to keep checking if the language string is "simple" or "test". (Which we were doing before these changes.)

The other use cases in where this transformation happened before the refactor were both checking if the incoming language string was in a set of RTL language codes. This will work with 'simple' and 'test' as-is without any special transformation to 'en'.

I've also added a comment describing why the special casing is necessary.

tonisevener (approved):
Thanks!

PR #3853, opened 2021-02-06
Build server: use App Store Connect API key

mcleinman:
**Fixes Phabricator ticket:** https://phabricator.wikimedia.org/T271350

### Notes
* The key is on the server - and is not committed to the repo.

### Test Steps
1. Using jenkins, build a build - using this branch. Ensure it's uploaded to app store connect.


fastlane/Fastfile
```diff
@@ -272,6 +272,12 @@ platform :ios do
     ipa_path = "#{build_dir}/#{product_name}.ipa"
     build_number =  options[:build] || get_latest_build_number(prefix: "betas") + 1
     version = get_version_number_from_options(options)
+    apple_key = `cat ~/AuthKey_8L2XY35BLF.p8`
+    api_key = app_store_connect_api_key(
```
tonisevener, line 276: Possibly overkill but I think we ought to try and mask the key and issuer ID as well, either via Jenkins environment variables or use a similar pattern as how you're pulling the auth key here.

mcleinman:
Will fix this next week.

fastlane/Fastfile
mcleinman: At long last, removed. (Ran into some issues - the files can't have any newline whitespace at the end of the file, but it's working now.) 

I'm going to cycle the keys on the server, but I think this is ready to review again.

tonisevener (approved):
Works like a dream!

PR #3855, opened 2021-02-11

translatewiki:
Translation updates


mcleinman (approved):
I don't fully understand why this diff is so large and includes so many deletions, but it seems safe:
- I spot checked a bunch of the commonly deleted strings. None of them appeared in our repo (iOS code or otherwise), though some existed if you go back a year or more.
- Most of the languages with many deletions also had some additions.
- Very commonly translated languages (French, Hebrew, etc.) didn't have the deletions.
- Maybe nothing gets deleted for a language unless we also have something added?

PR #3856, opened 2021-02-12
Make onboarding variant-aware

dempseyatgithub:

Unlike previous PRs, this one is best reviewed all at once instead of commit by commit. 

### Notes

For onboarding to be variant-aware, the language step needs to suggest language variants based on the user's OS settings for languages and region.

- The feature flag boolean needed to be marked `@objc` so it could be used in MWKLanguageLinkController.
- MWKLanguageLinkController needed to return a different list of languages that included Wikipedia language variant codes when the feature flag is turned on.

There is an existing property `wmf_preferredLanguageCodes` - and extension on NSLocale that relies on a method called `uniqueWikipediaLanguages()`.

This method currently takes an array of OS locale identifiers. For each locale, it adds a language variant string, if one is available. Otherwise, it adds the locale identifier.

For onboarding, we need this exact transformation, except using the locale's language code instead of locale identifier.

So, extended `uniqueWikipediaLanguages()` to take a boolean that controls which is used. Made the default the existing behavior so existing call sites do not need to be updated.

- Added a new property `wmf_preferredWikipediaLanguageCodes` that returns the array that contains language variant codes and language codes.

- Refactored some complex lookup code out into a separate method.

Note that the Locale/NSLocale extension is likely at 'peak complexity'. It contains extensions that support both having the language variant feature turned on and off. I believe once the feature it turned on, there can simplification and possible renaming to make things clearer, once it is clear exactly what needs to stay.

### Test Steps
1. With the language variants feature flag set to false, no change in app behavior.
2. With the feature flag set to true, language variants set in the OS languages should correctly appear in the onboarding language step and if the user does not edit, should be used for the app's preferred languages setting.


dempseyatgithub:
Sorry, I jumped the gun posting this. No need to review just yet.

dempseyatgithub:
It is ready for review now.

tonisevener (approved):

PR #3858, opened 2021-02-13
Add article inspector mapping (part 1 of 4)

tonisevener:
**Fixes Phabricator ticket:** https://phabricator.wikimedia.org/T270426

This doesn't complete the Phabricator task - it's just the beginning stages to get all the data we need in a controller. Further processing to break it down into sentences and combine the data from each response will be in followup PRs.

### Notes
* On article load, we fetch the article content html over the JS bridge. At the same time we fetch the WikiWho response from the labs instance and decode the data.

### Test Steps
2. Set isEnabled = true in `ArticleInspectorController`
3. Add a breakpoint to the `processContent(articleHtml: String, wikiWhoResponse: WikiWhoResponse)` method.
3. Run and go to the article that is cached on the endpoint. You might need to visit https://wikiwho-ios-experiments.wmflabs.org/whocolor/{article_title} to kick off the caching. See that the `articleHtml` and `wikiWhoResponse` are properly populated at the breakpoint.


tonisevener:
Marking as Hold since we don't want this reviewed for 6.8.

tonisevener:
Closing this for now to clean up GitHub, will reopen once we get closer to heavy development on it.

PR #3862, opened 2021-02-18
Variant-aware save buttons and list

dempseyatgithub:

It is probably best reviewed commit by commit.

### Notes
This PR has a few commits that are prep-work for future PRs and then updates to SaveButtonsController and MWKSavedPageList to handle variants.

*Note that there are more changes needed to get saved pages / reading lists working when language variants are turned on. These changes should keep behavior the same when language variants are turned off.*

**Commit 1**
Add methods to MWKLanguageLinkController and NSLocale to make a best guess of language variant when an article comes in from a synched reading list.

The language link controller method searches for a match in the user's app preferred languages, if none found returns the value of the NSLocale method.

The NSLocale method searches the user's OS languages settings for a preferred variant. If no preferred variant is found, uses a default variant value.

Added a test, but only runs if language variants are turned on.

**Commit 2**
In an upcoming PR, need to fetch articles by inMemoryKey. We already do this in one place in the code.
Added a fetch method and made the method that returns the appropriate predicate private.
This is straight refactoring. No logic change.

**Commit 3**
Update SaveButtonsController to use in-memory keys. Does not change behavior. Note, it still calls into non-variant aware APIs. This is expected for this commit.

**Commit 4**
Add a way to track all visible save buttons for the same article across all variants. This will enable keeping the save buttons in sync across different variants of the same article in the explore feed. No change in behavior when language variants are turned off.

**Commit 5**
Use isAnyVariantSaved property for state of saved buttons.
Update all cases where save button state is set. Note many call sites were using `saveDate == nil` which indicates *unsaved* so the results needed to flip in those cases.

**Commit 6**
Make MWKSavedPageList variant-aware. This is a change to three methods.

**Commit 7**
Revert to using the string articleKey when coordinating between remote and local reading list entries.
Create a typealias `RemoteReadingListArticleKey` for `String`. Using that type as the key for various tracking dictionaries where appropriate makes it less error prone than both variant and non-variant-aware keys using `MWKInMemoryURLKey`.

### Test Steps
1. With language variants turned off, behavior of saved buttons, saved list and reading lists should remain unchanged.

### DIFF
--- a/WMF Framework/ReadingListsSyncOperation.swift
+++ b/WMF Framework/ReadingListsSyncOperation.swift
@@ -742,7 +742,7 @@ internal class ReadingListsSyncOperation: ReadingListsOperation {
         let group = WMFTaskGroup()
         let semaphore = DispatchSemaphore(value: 1)
         var remoteEntriesToCreateLocallyByArticleKey: [WMFInMemoryURLKey: APIReadingListEntry] = [:]
-        var requestedArticleKeys: Set<WMFInMemoryURLKey> = []
+        var requestedArticleKeys: Set<RemoteReadingListArticleKey> = []
         var articleSummariesByArticleKey: [WMFInMemoryURLKey: ArticleSummary] = [:]
         var entryCount = 0
         var articlesByKey: [WMFInMemoryURLKey: WMFArticle] = [:]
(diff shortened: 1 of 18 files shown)


WMF Framework/ReadingListsSyncOperation.swift
```diff
@@ -752,14 +752,14 @@ internal class ReadingListsSyncOperation: ReadingListsOperation {
...
                 remoteEntriesToCreateLocallyByArticleKey[articleKey] = remoteEntry
-                guard !requestedArticleKeys.contains(articleKey) else {
+                guard !requestedArticleKeys.contains(remoteArticleKey) else {
                     return
                 }
-                requestedArticleKeys.insert(articleKey)
+                requestedArticleKeys.insert(remoteArticleKey)
                 if let article = dataStore.fetchArticle(with: articleURL, in: moc) {
                     articlesByKey[articleKey] = article
                 } else {
```
tonisevener, line 765: I'm seeing a crash at [line 822](https://github.com/wikimedia/wikipedia-ios/blob/VariantAwareSaveButtonsAndList/WMF%20Framework/ReadingListsSyncOperation.swift#L822) when I log in with an account with reading lists and it begins syncing. It's expecting `articleKey` to be of type `NSString`. That `moc` call seems pretty untyped (it accepts `Any` for value), maybe we can replace it with one of our article-specific `fetchOrCreate` calls like `moc.fetchOrCreateArticle(withKey: articleKey.databaseKey, variant: articleKey.languageVariantCode)`?

tonisevener (commented):
Code looks good to me, though I found a couple of things when testing without the variants flag on. One crash (detailed below) and also sometimes when I tap the Featured Article's "Save" button in the Explore feed, it's state doesn't change until I scroll off and scroll it back on. It's intermittent though.

out

WMF Framework/ReadingListsSyncOperation.swift
dempseyatgithub: I'll take a look at both and thanks for finding those.

dempseyatgithub: The crash is a line I missed in putting that change into a commit - and the fix is exactly as you suggested.

Will see what is up with the save button selection and hopefully have both fixes in a small commit.

tonisevener (approved):

PR #3863, opened 2021-02-19
Add additional language variant strings for early translation

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T273666 (just the adding strings component of this ticket)

### Notes
Apologies - this one's a lot of reading and finicky detail stuff to review. You should only need to review the `CommonStrings.swift` changes, as the rest is autogenerated.

Adds language variant strings for early translation by the community (like https://github.com/wikimedia/wikipedia-ios/pull/3839). I agree with [Toni's comment here](https://phabricator.wikimedia.org/T272416#6838809) that it may be advantageous to have generic strings to use, because as you can see here the original approach for just Chinese doesn't necessarily scale nicely as the languages list grows. But, at least getting this work merged now means we're more likely to have fully translated specialized strings before the feature launches.

### Test Steps
1. Confirm each language from https://phabricator.wikimedia.org/T273666 is represented and the string values match
2. Ensure each localized string's a) variable name b) key c) value and d) description are using the corresponding and appropriate language
3. Ensure there are no errant semicolons (semicolons from the Phab ticket body text should have been replaced with newlines `\n`)


WMF Framework/CommonStrings.swift
```diff
@@ -337,4 +339,94 @@ public extension CommonStrings {
...
 
+	// Crimean Tatar
+
+	static let crimeanTatarVariantsAlertTitle =  WMFLocalizedString("crimean-tatar-variants-alert-title", value: "Updates to Crimean Tatar variant support", comment: "Title of alert used to inform users about Crimean Tatar variant support.")
+
+	static let crimeanTatarVariantsAlertBody = WMFLocalizedString("crimean-tatar-variants-alert-body", value: "The Wikipedia app now supports the following Crimean Tatar variants as primary or secondary languages within the app, making it easier to read, search and edit in your preferred variants:\n\nQırımtatarca, Latin Crimean Tatar Latin (chr-latn)\nкъырымтатарджа, Кирил Crimean Tatar Cyrillic (crh-cyrl)", comment: "Body text of alert used to inform users about Crimean Tatar variant support. Please do not translate the newlines (\n) or Crimean Tatar characters (къырымтатарджа, etc.).")
+
+	static let crimeanTatarVariantsAlertPreferencesButton = WMFLocalizedString("crimean-tatar-variants-alert-preferences-button", value: "Review your preferences", comment: "Action button on alert used to inform users about Crimean Tatar variant support.")
```
tonisevener, line 348: @staykids any need to have separate strings for preferences and dismiss? Maybe we could combine them behind two more generic keys, `variants-alert-preferences-button` & `variants-alert-dismiss-button`.

staykids: Nope - no reason I'm aware of other than mimicking the same pattern I used for Chinese variants. I'll generalize them now. 👍🏽

staykids: Updated!

tonisevener (approved):

PR #3864, opened 2021-02-19
Add article inspector mapping (part 2 of 4)

tonisevener:
**Fixes Phabricator ticket:** https://phabricator.wikimedia.org/T270426

Do not review until https://github.com/wikimedia/wikipedia-ios/pull/3858 is merged and this PR is updated with `main`.

This doesn't complete the Phabricator task - this is the second step that processes the html from the article content and WikiWho response into sections, paragraphs, and sentences. Further processing to map together the WikiWho sentences and article content sentences, and create NSAttributedStrings for use in the modal will be in followup PRs. 

### Notes
* This PR involves processing the html from the article content and the WikiWho response, breaking them down into section, paragraph, and sentence models (the `IndividualSentence` flavor). Mapping together these sentence models from both responses will lean on the `CombinedSentence` model. I have it in this PR just to complete out the generic structure of the `Section` and `Paragraph` models, but it isn't in use yet.

I decided to go with parsing the html via the third-party `SwiftSoup` library. We could ask the web view to do it instead (i.e. have a javascript bridge call that returns the sections and paragraphs), but since the WikiWho response isn't represented in the web view we would need to send it there and create another document object with it, which felt gross. Plus with this library more processing can be done with Swift. I also looked a little bit into trying to process both html responses outside of a web view with JavascriptCore, but that wasn't very straightforward - there was no document object for me to access, so I abandoned it before digging further.

### Test Steps
1. Set isEnabled = true in `ArticleInspectorController`
3. Add a breakpoint to the `processContent(articleHtml: String, wikiWhoResponse: WikiWhoResponse)` method, after the `articleHtmlSections` and `wikiWhoSections` are created. 
3. Run and go to the article that is cached on the endpoint. You might need to visit https://wikiwho-ios-experiments.wmflabs.org/whocolor/{article_title} to kick off the caching. See that the `articleHtmlSections` and `wikiWhoSections` are properly populated at the breakpoint (note I also added some light unit tests that we can flesh out more fully as we run into issues).


benkaday3218 (approved):

tonisevener:
Closing this for now to clean up GitHub, will reopen once we get closer to heavy development on it.

PR #3867, opened 2021-02-22
Variant aware reading lists

dempseyatgithub:

It is probably best reviewed commit by commit.

### Notes
Each commit message should explain that commit pretty well.
Most changes involve moving to the in-memory key.

The `savedVariant` property in the second commit is used later to ensure the correct article is sent in the 'article saved / unsaved' notification.

The biggest commit includes `ReadingListsSyncOperation` and is largely determining the variant for an incoming remote reading list entry, and then using the variant-aware URL/in-memory key.

### Test Steps
1. There should be no change in behavior with language variants turned off.
2. With language variants turned on, this will sync remote entries using the best guess heuristic. Locally, one variant can be saved at a time, but the variant the user was viewing when saved will be used locally. 

### Known issue
I found an edge case:
- Have a saved article
- View article in article viewer
- Tap to view in another language
- Choose another variant and view article
- Tap the save button
- The article unsaved, but the button does not update

I thought it would probably be okay to land this change and track down that issue separately.


tonisevener (approved):
Working great for me!

PR #3868, opened 2021-02-23
Add variants for additional languages

dempseyatgithub:
This is work towards adding additional language variants beyond Chinese https://phabricator.wikimedia.org/T268275.

It is not the entire feature/ticket, but enables testing with all language variants to be added.

### Notes
The first commit adds the additional mappings between OS locale identifiers and language variant codes. This will be used during onboarding and migration, as well as helping to determine the variant for articles without any variant info (like sync'd reading list entries).

The second commit just adjusts indentation.

The third commit adds the language variant JSON objects that are read at runtime.

### Test Steps
1. With language variants turned on, you can choose the variants for these seven languages, along with Chinese and Serbian.

2. Names / variant codes, etc. should match those provided in the Phabricator for this issue.

###Known Issue
Viewing Kazakh and Kurdish in the Arabic variant does not display right-to-left yet.


docs/languages.md
```diff
@@ -36,6 +36,33 @@ _Troubleshooting: Variants missing from this file cause onboarding to fail to id
...
+
+In Chinese, the region affects the Wikipedia langauge variant. So, the mapping file specifies regions in the most deeply nested dictionary.
+
+In all other languages with variants so far, the script identifier is enough to identify the variant, and does not change regardless of region.
+For these languages, only the "default" key is present in the region dictionary.
+
+Also note that not all scripts/variants supported by a particular Wikipedia language are available identifiers in the OS.
+In these cases, finding the language in the user's OS preferences implies a particular language variant.
+
+Note that "gan" and "crh" are not available in the OS, although their scripts are. A default language variant value is provided in case a future OS includes these locale identifiers.
```
tonisevener, line 56: I may be getting things mixed up but I also couldn't find any Kurdish variants as an OS language to choose (ku-latn and ku-arab) either. All I see is Kurdish, Sorani which mapped to the `ckb` language code. Just noting in case it's worth mentioning here.

dempseyatgithub: I found the same thing. I had researched by filtering the list of `Locale.availableIdentifiers` which does include Kurdish identifiers. I updated the note to add that Kurdish is currently not selectable in the OS language settings.

That is different from 'gan' and 'crh' which do not appear in the Locale available identifiers list.

Wikipedia/Code/wikipedia-language-variants.json
```diff
@@ -1,66 +1,170 @@
...
+            "languageVariantCode": "zh-tw",
+            "languageName": "臺灣正體",
+            "localName": "Taiwanese Traditional",
+            "languageCode": "zh"
+        }
+    ],
+    "sr": [
+        {
+            "languageVariantCode": "sr-ec",
+            "languageName": "Србија",
```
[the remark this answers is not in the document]
dempseyatgithub: The language names of the newly added languages in the variants file are copy-pasted from the Phab ticket, so they should be as specified (lower or upper case).

I had left the Serbian names as-is, but have now changed them to match the Phab ticket.

tonisevener (approved):

PR #3875, opened 2021-02-26
Fix article theming bug

tonisevener:
**Fixes Phabricator ticket:** https://phabricator.wikimedia.org/T275239

### Notes
* Intermittently (but fairly often), the chrome of an article view will theme but not the web view content. This happens when a new theme trigger to ArticleViewController occurs after the web view initial setup call occurs with the old theme but before the web view is fully loaded to accept the new theme. In this case, I am checking the themes after the .loaded state is triggered (meaning the web view is ready to be re-themed) and see if they are out of sync. If so, trigger a retheming message to the web view.

### Test Steps
1. Launch the app on a device in Light Mode, be sure the theme in app settings is set to "Default" so it follows the system settings.
2. Background the app, go to OS settings and change the theme to Dark.
3. Now go to an article in Safari and enter the app via an article deep link.
4. Confirm that the article theming looks good once everything is loaded. On `main`, pretty often this does not happen. The web view still shows the old theme, causing hidden words against the `ArticleViewController` background color.


mcleinman (approved):
Looks great, tests great!

PR #3877, opened 2021-02-26
Validate project settings with no target changes performed.

staykids:
**Phabricator:** N/A

### Notes
Removes the "Validate Project Settings" Xcode project warning. Performing the changes Xcode requested previously resulted in App Store validation warning emails regarding the Stickers target, so that change was reverted. In this PR, I went the route Toni recommended of unselecting the recommended changes and completing the validation instead of alternative approaches to ignore the warnings. This means that the Stickers target (hopefully) shouldn't be affected again in a way that triggers App Store warning emails.

### Test Steps
1. Confirm project builds and no "Validate Project Settings" warning is displayed in Xcode.


mcleinman (approved):
Looks great, thank you for this!

PR #3880, opened 2021-03-05
Allow native article description to update EN Wikipedia wikitext

tonisevener:
**Fixes Phabricator ticket:** https://phabricator.wikimedia.org/T257867

### Notes
This PR should fix our broken English article description update flow. It allows us to use our native article description screen to update English Wikipedia, for both new descriptions and existing descriptions. 

The general approach is to move any Wikidata-specific code out of `DescriptionEditViewController.swift`, and instead have it reference a protocol that performs the update specifics depending on which implementation we have passed in. Some other bug fixes and cleanup was added along the way. 

This is best reviewed commit-by-commit. Details are below:

**Commit 1**
Perform some general class renaming and cleanup first. `WikidataDescriptionEditingController` is a Fetcher, but you wouldn't know it from the name. Changed the name to `WikidataFetcher`. Changed to previous `WikidataFetcher` to an extension under `WikidataFetcher+Places.swift`, since it was only used in that one location. I am also removing the reference to `WikidataDescriptionEditingController` out of `MWKDataStore`. It does not use any persistent store within it's calls, so I see no need for it to hang off of that data store object.

**Commit 2**
Commit making tweaks to the JS bridge calls. We need to fetch the current description for pre-populating the textfield on the native screen. Non-EN articles already have this as a part of the wikidata description fetched from the summary endpoint, but EN articles do not. We already have this loaded in the web view so I'm choosing to get it from there rather than making any additional endpoint calls. This also adds some slight cleanup to the ArticleDescriptionSource enum. There's already a .none option, so made sense to lean on that rather than use an optional. This allows me to not deal with the optional later on when presenting the `DescriptionEditViewController`.

**Commit 3**
Tweaks needed to our fetchers and processing. The Short Description API (which we'd be using if it wasn't for oAuth) takes `baserevid` into [account](https://github.com/wikimedia/mobileapps/blob/3a3ac209ddd3ce4e7b676f331c28b4b1afb79520/lib/mwapi.js#L486), so this is pulling the revisionID when fetching the section, to add in later as a `baserevid` when updating the section.

**Commit 4**
This is the bulk of the new logic. Establishes a protocol for updating article descriptions. Copies Wikidata-specific logic out of `DescriptionEditViewController` and into `WikidataDescriptionController`. `ShortDescriptionController` contains the new logic for updating an EN Wikipedia article description. It detects if it already has a Short Description template, and if so, update the description within it. Otherwise prepend a new Short Description template with the new description to the first section. It also uses our JS bridge call in Commit 2 to fetch the article's current description for pre-populating the textfield in `DescriptionEditViewController`.

**Commit 5**
Here we are updating `DescriptionEditViewController` and `ArticleViewController+Editing` to use our new protocol & implementing controllers.

**Commit 6**
Adding some basic tests for `ShortDescriptionController`. Also I noticed when going through the "Learn more" link for EN Wikipedia, the setting to suppress the extra navigation buttons are lost when you navigate to a link within the web view. Here I'm passing this setting along in our `presentOrPush` method, I think it's fairly safe.

**Commit 7**
This fixes a bug I found when updating a non-EN Wikipedia article. Although the article content would reflect the new Wikidata description, the textfield would not pre-populate the latest description when tapping to edit the article description again. This is because the summary endpoint is not attempting to fetch again after performing an update. It seems our [cachePolicy](https://github.com/wikimedia/wikipedia-ios/blob/3fe61d389368499920a175ba8271b68d50536f4b/Wikipedia/Code/ArticleViewController.swift#L434) setting wasn't taking effect because the view state wasn't right (it already switches back to .loaded by this point) and deep within the summary fetching, it totally drops the cachePolicy we have set anyways. This commit fixes these 2 things.

### Test Steps
1. In EN Wikipedia, test adding a new article description and updating an existing article description. Update should go through and you should see it reflected in the article content. Check the wikitext in section 0 (either on Desktop or by choosing to "Edit introduction" in the alert view) - confirm that a {{Short description}} template has been added. Also confirm that the "Learn more" link on the Add/Edit Description modal goes to a web view about Short Descriptions.
2. In any other language Wikipedia, test adding a new article description and updating an existing article description. Update should go through and you should see it reflected in the article content. Check the wikitext in section 0 (either on Desktop or by choosing to "Edit introduction" in the alert view) - confirm that there is NO {{Short description}} template. Go to Wikidata page for that article on desktop, confirm you do see your latest update. Also confirm that the "Learn more" link on the Add/Edit Description modal goes to the FAQ page about Wikidata as it did before.

**Final FYIs**
I have not fixed any abuse filter or captcha handling here. I was able to intermittently reproduce an abuse warning on Test Wiki, but it seemed messed up in our standard editing flow as well, and this PR is already large enough.

Pretty often you will see an error after updating about the article content not being able to refresh in time. I believe this is existing and expected functionality [from this PR](https://github.com/wikimedia/wikipedia-ios/pull/3564).


staykids:
> Pretty often you will see an error after updating about the article content not being able to refresh in time. I believe this is existing and expected functionality from this PR.

@tonisevener Maybe it's just something goofy about my internet connection, but it feels like I'm running into this _almost_ every time, and just wanted to check if your experience with it was the same or not. The related [upstream task](https://phabricator.wikimedia.org/T252142) seemed to be resolved (unless I'm misinterpreting.)

Specifically, this failure (with an error banner presented on the article) is getting hit pretty consistently for me when making EN title description edits (not any other edits). 
https://github.com/wikimedia/wikipedia-ios/blob/66ee9e99437e1b3500536eed05af97da99a5f542/Wikipedia/Code/ArticleFetcher.swift#L332

Strangely, increasing the `maxAttempts` here to something higher (like 10), seems to provide enough time for the revision to change, load, and present without the error banner:
https://github.com/wikimedia/wikipedia-ios/blob/66ee9e99437e1b3500536eed05af97da99a5f542/Wikipedia/Code/ArticleViewController.swift#L690

So my two Qs: 
1. Are you experiencing the same behavior as me, or does it seem to work without this error for you more often than not?
2. If you're also encountering it, do you think it's worth increasing the `maxAttempts`, or prefer some other solution?

tonisevener:
> > Pretty often you will see an error after updating about the article content not being able to refresh in time. I believe this is existing and expected functionality from this PR.
> 
> @tonisevener Maybe it's just something goofy about my internet connection, but it feels like I'm running into this _almost_ every time, and just wanted to check if your experience with it was the same or not. The related [upstream task](https://phabricator.wikimedia.org/T252142) seemed to be resolved (unless I'm misinterpreting.)
> 
> Specifically, this failure (with an error banner presented on the article) is getting hit pretty consistently for me when making EN title description edits (not any other edits).
> https://github.com/wikimedia/wikipedia-ios/blob/66ee9e99437e1b3500536eed05af97da99a5f542/Wikipedia/Code/ArticleFetcher.swift#L332
> 
> Strangely, increasing the `maxAttempts` here to something higher (like 10), seems to provide enough time for the revision to change, load, and present without the error banner:
> https://github.com/wikimedia/wikipedia-ios/blob/66ee9e99437e1b3500536eed05af97da99a5f542/Wikipedia/Code/ArticleViewController.swift#L690
> 
> So my two Qs:
> 
> 1. Are you experiencing the same behavior as me, or does it seem to work without this error for you more often than not?
> 2. If you're also encountering it, do you think it's worth increasing the `maxAttempts`, or prefer some other solution?

Yeah I was seeing it a lot, and you're right, enough to dig in more and try to fix. Actually looking at the comment thread [here](https://phabricator.wikimedia.org/T252146), it seems like for updating EN Wikipedia I should be able to grab the latest revision ID and update in that manner, which would avoid the polling. I think the polling is needed for updating a Wikidata description for non-EN, but maybe we can try bumping up the `maxAttempts` or the exponential delay a bit for those cases. I'll work on this next.

tonisevener:
@staykids I pushed an update that requests the specific `revisionID` if the update is from EN Wikipedia, that should fix the polling errors on that one. The non-EN Wikipedia updates respond more reliably for me, the past few times I've tried seem to update before the polling error triggers. Let me know how those feel, if it's still bad I'll look into tweaking the attempts / retry delay.

staykids:
> @staykids I pushed an update that requests the specific `revisionID` if the update is from EN Wikipedia, that should fix the polling errors on that one.

👍🏽 Nice. The only thing I keep running into is the  result edited page isn't consistently loading/displaying the new revision after being edited.

1. Open any EN article
2. Observe article description
3. Make edit to article description, then publish
4. Blue edit successfully published toast at bottom, no red error at top
5. Observe article description - it isn't the new one, still previous one

It appears the base revision is correct, and then new revision ID from `replaceDescriptionInWikitextAndUpload` is eventually passed to the `ArticleViewController`'s `waitForNewContentAndRefresh`.

Interestingly, just delaying this refresh by a half second seems to resolve the issue, but I haven't dug deep enough into it to know if it's a threading issue or if that just provides enough time for backend things to catch up:

https://github.com/wikimedia/wikipedia-ios/blob/65973c4fc17d5d6ff1f53b0145c0ce25c72e7eee/Wikipedia/Code/ArticleViewController.swift#L736



I know this general edit revision display problem is something we've encountered shades of before. Would you rather us merge the current PR and try and address this problem separately, or address it in this PR?

tonisevener:
> Would you rather us merge the current PR and try and address this problem separately, or address it in this PR?

Hmm I hadn't noticed that, but I'll take another look, I'd rather address it here. It's possible that I caused some sort of regression by messing with the cache flags in this PR.

tonisevener:
@staykids I pushed a commit that updates the article content locally if the description post succeeds, as discussed yesterday. This should (hopefully) paper over most of the refresh problems you've been seeing. If any of the JS code fails, it will attempt the web view refresh as it was before this commit. I also made a [Phab task](https://phabricator.wikimedia.org/T277522) to track the server-side loading issue.

staykids:
> @staykids I pushed a commit that updates the article content locally if the description post succeeds, as discussed yesterday.

👍🏽 Awesome, will jump back into this soon.

staykids (approved):
Thanks so much for documenting the commit by commit - made it easier to follow along! I can't shake the feeling that I'm missing something, but so far this seems to be working pretty well in my testing on EN wiki and ES wiki.

Few notes:
- Minor completion block Q in review
- I'm not clear if this image attached is the intended behavior or not (because I don't remember what the Learn more link did when the EN article description stuff was working 🥴), but tapping "Learn more" from an ES article description edit and an EN one displays different content:

es_en_learn_more


- Related to Commit 7: I know we talked a bit about it, just want to confirm if we're ok or not for now with how it's possible to edit an article description, then possibly have the wrong information be prefilled if you try to immediately edit it after. Right now, following these steps for an article on ES wiki results in the wrong article description populating in the textfield for me:
1. Find ES article with existing description, tap to edit and notice it's correctly prefilled
2. Edit the article description and publish, notice your change is nicely reflected in the article page
3. Immediately tap to edit again, and notice the original description is prefilled instead of your current one

I went ahead and approved (because this PR functionally accomplishes the Phab tasks's purpose, which is otherwise broken in the shipping version) but I defer to you on if we should merge what works now and address the other points later or resolve them before merging.

tonisevener:
> Minor completion block Q in review

Added PR feedback commit fixing this.

> Tapping "Learn more" from an ES article description edit and an EN one displays different content

Yeah sorry I didn't explain that further - that was a later decision that was made in the [Phab task](https://phabricator.wikimedia.org/T257867#6883904). The non-EN native modal mentions Wikidata, which is no longer true for EN, so we changed out the Learn more link for EN.

As for the description being out of date when you edit again, I think it's worth fixing, good catch. I'll do this in a follow-up PR to get a 2nd pair of eyes on it without holding this one up.

PR #3885, opened 2021-03-10
Move from KVO to an explicit delegate for language changes

dempseyatgithub:
This is prep work for https://phabricator.wikimedia.org/T273666 (Language Variant Migration)

### Notes
During language variant migration, it is necessary to *not* send notifications that the preferred languages have changed.

MWKLanguageLinkController ends up doing a KVO notification on 'allLanguages' whenever 'preferredLanguages' changes. This makes for ugly logic to suppress the notification, needing to skip `willChange` and `didChange` in different places in the method.

I've moved to an explicit delegate which is now used for both data sources for MWKLanguageFilter. This makes the dependencies much more clear and also removes an Obj-C-centric mechanism that makes the move to Swift more difficult.

- Current implementation was misusing KVO on allLanguages for changes to preferredLanguages
- Add MWKLanguageFilterDataSourceDelegate protocol
- Add languageFilterDelegate property to MWKLanguageFilterDataSource protocol and implementors
- Remove manual KVO triggering from MWKLanguageLinkController
- Add calls to delegate in MWKLanguageLinkController and MWKTitleLanguageController
- Replace KVO registration with delegate setting in MWKLanguageFilter
- Replace KVO observation method with delegate notification method in MWKLanguageFilter
- This change makes the dependency explicit
- It also eases transition to Swift w/o reliance on Obj-C properties

### Test Steps
1. Ran all tests
2. In Settings, choose Languages, add a language. Search and choose one. Choose to Add a language again, search, the language you just chose should no longer be available.
3. When viewing an article, tap the 'Change Language' button. Languages should appear after the list of that article in other languages has been fetched.

### DIFF
--- a/Wikipedia/Code/MWKLanguageFilter.h
+++ b/Wikipedia/Code/MWKLanguageFilter.h
@@ -4,20 +4,18 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-@protocol MWKLanguageFilterDataSourceDelegate
-- (void)noteLanguagesDidChange;
-@end
+// Notification sent by an MWKLanguageFilterDataSource when language array values change
+extern NSString *const MWKLanguageFilterDataSourceLanguagesDidChangeNotification;
 
 @protocol MWKLanguageFilterDataSource <NSObject>
 
-@property (nonatomic, weak, nullable) id <MWKLanguageFilterDataSourceDelegate> languageFilterDelegate;
 @property (readonly, copy, nonatomic) NSArray<MWKLanguageLink *> *allLanguages;
 @property (readonly, copy, nonatomic) NSArray<MWKLanguageLink *> *preferredLanguages;
 @property (readonly, copy, nonatomic) NSArray<MWKLanguageLink *> *otherLanguages;
 
 @end
 
-@interface MWKLanguageFilter : NSObject <MWKLanguageFilterDataSourceDelegate>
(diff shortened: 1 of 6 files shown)


tonisevener (commented):
Looking good, just see one bug. It's not apparent to me yet how it's happening so not something obvious via code, but when I add a language in the app settings, the "Your languages" list doesn't update until I close the modal and come back.

out

dempseyatgithub:
> Looking good, just see one bug. It's not apparent to me yet how it's happening so not something obvious via code, but when I add a language in the app settings, the "Your languages" list doesn't update until I close the modal and come back.

Thank you - I wonder if there is one more KVO notification that I've missed. I'll take a look.

dempseyatgithub:
I see what's going on. Although the preferred language settings view controller does not display a language filter, it *has* one, so it also needs to find out filtered languages have changed.

This has turned into a bit of a rathole.

dempseyatgithub:
I switched to a notification instead of a delegate. It's probably best to review the changes as a single set of changes, instead of commit by commit since the latest commit essentially goes in a completely different direction than the previous one.

I tested and this addresses the issue you were seeing and Change Language for articles also continues to work. I added a comment in MWKLanguageFilter explaining that multiple can be in use simultaneously.

tonisevener:
Cool, looks good! Thanks for fixing.

tonisevener (approved):

PR #3886, opened 2021-03-12
add altISOCode to MWKLanguageLink to work around Norwegian bugs

tonisevener:
**Fixes Phabricator tickets:** https://phabricator.wikimedia.org/T276645, https://phabricator.wikimedia.org/T272193

### Notes
This ain't pretty but I believe this fixes the Norwegian bugs without causing regression. I tried to avoid any explicit callout to "no" or "nb" so that this could be a more generic fix that possibly fixed any other similar issues on other wikis, but I found that those other wikis acted inconsistently in the langlinks endpoint, so a completely generic fix wasn't possible. Since we only have complaints with Norwegian it's probably a good idea to limit the fix to that language for now anyways.

I also decided against changing the `languageCode` for Norwegian in the app from "no" to "nb". There are lots of assumptions in the app that the subdomain is the language code (for example, our `wmf_language` extension on NSURL), so I felt it safer to leave it as-is. I am instead adding an optional `altISOCode` to `MWKLanguageLink` to serve as basically just another code to check, but only as band-aids in the area of these two bugs.

### Test Steps
1. Change device language to **ONLY** Norwegian (Bokmål), no secondary languages.
2. Fresh install the app and launch. Confirm you see a Norwegian language chosen in Onboarding (it's blank on `main`). Confirm the Explore feed is fetched and populates (it's blank on `main`). Search for an article and tap into it. Confirm article loads. This concludes testing for https://phabricator.wikimedia.org/T276645.
3. On Norwegian, search for any article that's common enough to be listed in other languages (I chose USA). Go to that article, choose the language picker, and open article in another language. Choose language picker from that article, and confirm you see Norwegian Bokmål listed at the top as a language option. On `main` Norwegian does not show in this case because the language codes do not line up. This concludes the test for https://phabricator.wikimedia.org/T272193.

### DIFF
--- a/Command Line Tools/Update Languages/WikipediaLanguageCommandLineUtilityAPI.swift
+++ b/Command Line Tools/Update Languages/WikipediaLanguageCommandLineUtilityAPI.swift
@@ -20,7 +20,7 @@ class WikipediaLanguageCommandLineUtilityAPI {
     //     ...
     // }
     func getSites() -> AnyPublisher<[Wikipedia], Error> {
-        let sitematrixURL = URL(string: "https://meta.wikimedia.org/w/api.php?action=sitematrix&format=json&formatversion=2&origin=*")!
+        let sitematrixURL = URL(string: "https://meta.wikimedia.org/w/api.php?action=sitematrix&smsiteprop=url%7Cdbname%7Ccode%7Csitename%7Clang&format=json&formatversion=2&origin=*")!
         return URLSession.shared
             .dataTaskPublisher(for: sitematrixURL)
             .tryMap { result -> [String: Any] in
@@ -44,10 +44,19 @@ class WikipediaLanguageCommandLineUtilityAPI {
                     else {
                         return nil
                 }
-                return Wikipedia(languageCode: code, languageName: name, localName: localname)
+                
+                guard code == "no" else {
+                    return Wikipedia(languageCode: code, languageName: name, localName: localname, altISOCode: nil)
+                }
+                
+                //Norwegian (Bokmål) has a different ISO code than it's subdomain, which is useful to reference in some instances (prepopulating preferredLanguages from iOS device languages, and choosing the correct alternative article language from the langlinks endpoint).
+                //https://phabricator.wikimedia.org/T276645
+                //https://phabricator.wikimedia.org/T272193
+                return Wikipedia(languageCode: code, languageName: name, localName: localname, altISOCode: "nb")
+                
             }
             // Add testwiki, it's not returned by the site matrix
-            wikipedias.append(Wikipedia(languageCode: "test", languageName: "Test", localName: "Test"))
+            wikipedias.append(Wikipedia(languageCode: "test", languageName: "Test", localName: "Test", altISOCode: nil))
             return wikipedias
         }.eraseToAnyPublisher()
     }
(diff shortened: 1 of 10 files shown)


Wikipedia/Code/WikipediaLookup.swift
```diff
@@ -45,14 +45,14 @@ import CocoaLumberjackSwift
...
-                    
+ 
                     var localizedName = wikipediaLanguageVariant.localName
                     if !Locale.current.isEnglish,
                         let iOSLocalizedName = Locale.current.localizedString(forIdentifier: wikipediaLanguageVariant.languageVariantCode) {
                         localizedName = iOSLocalizedName
                     }
                     
-                    return MWKLanguageLink(languageCode: wikipediaLanguageVariant.languageCode, pageTitleText: "", name: wikipediaLanguageVariant.languageName, localizedName: localizedName, languageVariantCode: wikipediaLanguageVariant.languageVariantCode)
+                    return MWKLanguageLink(languageCode: wikipediaLanguageVariant.languageCode, pageTitleText: "", name: wikipediaLanguageVariant.languageName, localizedName: localizedName, languageVariantCode: wikipediaLanguageVariant.languageVariantCode, altISOCode: nil)
```
mcleinman, line 55: I realize we don't need an altcode for any language variants currently, but is there any harm in providing a non-nil property here (in case we get any in the future)?

WikipediaUnitTests/Code/MWKLanguageLinkControllerTests.m
```diff
@@ -47,15 +47,15 @@ - (void)testSaveSelectedLanguageUpdatesTheControllersFilteredPreferredLanguages
     //    NSAssert(![[self preferredLanguageCodes] containsObject:@"test"],
```
mcleinman, line 47: Is it worth adding a test for these altcodes? Don't need one strictly speaking, but could be useful...

Command Line Tools/Update Languages/WikipediaLanguageCommandLineUtilityAPI.swift
```diff
@@ -44,10 +44,19 @@ class WikipediaLanguageCommandLineUtilityAPI {
                     else {
                         return nil
                 }
-                return Wikipedia(languageCode: code, languageName: name, localName: localname)
+                
+                guard code == "no" else {
```
mcleinman, line 48: Minor quibble, not a bug, feel free to ignore: I'd sorta expect to see the `guard` check if it's not `no` - and if it is `no` (and thus non-standard), let the guard statement handle that case.

mcleinman (commented):
Couple small comments/questions

Command Line Tools/Update Languages/WikipediaLanguageCommandLineUtilityAPI.swift
tonisevener: Good catch! This is fixed.

WikipediaUnitTests/Code/MWKLanguageLinkControllerTests.m
tonisevener: Yeah, agreed, it would be nice. I'm having a hard time getting `MWKLanguageLinkController` to think the OS languages have "nb" in there, which would be the best test. I'm running low on time today but I can look into it next week if necessary.

Wikipedia/Code/WikipediaLookup.swift
tonisevener: This is done, let me know if isn't what you had in mind.

mcleinman: Thanks!

mcleinman (approved):
Tests great, code looks great. Thanks!

PR #3891, opened 2021-03-24
Article description followup - hide wikidata description update delay

tonisevener:
**Fixes Phabricator ticket:** Feedback from https://github.com/wikimedia/wikipedia-ios/pull/3880#pullrequestreview-616851526

### Notes

I am forcing an update from the summary endpoint if the article description successfully posts, which should update WMFArticle's wikidataDescription property to match the server. Unfortunately it does seem like there's a server delay so the response we get still does not reflect the user's most recent update.

To get around this, I am holding onto the last successful description update for Wikidata descriptions and preferring that for prepopulating the textfield over the WMFArticle's wikidataDescription property. This does patch over the particular bug in the PR feedback comment. But you will notice elsewhere in the app the description is still the old one (say, the Continued Reading card for example). I don't think fixing it across the app is worth looking into at this point, but I'm open to a change of approach here if there's a better way.

I am still keeping the summary update even though it's a little useless, since it seems like good practice to fetch the latest summary from the server after an update and if that delay does become reduced server-side it will fix the descriptions across the app. 

### Test Steps
1. On a non-EN Wikipedia, update an article description.
2. Confirm the article content updates to the new description.
3. Tap the edit article description button again
4. Confirm the textfield on the native screen is pre-populated with the latest description that the user just posted. On `main` it is not.


tonisevener:
Closing this PR while we watch the fix on https://phabricator.wikimedia.org/T277522.

PR #3892, opened 2021-03-26
Support language variants in user talk pages

tonisevener:
**Fixes Phabricator ticket:** https://phabricator.wikimedia.org/T277603

### Notes
These are a few changes needed to support variants in talk pages.

**Commit 1** 
Here I am considering the `wmf_contentLanguageCode` as logic for whether to reload the TalkPageContainerVC with the new language variant or not. Before it only considered `siteURL`, causing the container to assume it was the same and not reload.

I also added a `variant` property to the `TalkPage` managed object and am saving and referencing that for pulling persisted objects. I didn't bump up the database version, but if we go out to public beta before this PR is merged that will be needed.

**Commit 2**
Here I am adding talk page variant migration, in the same way we are migrating the other Wikipedia entities. I found that after migration, the talk page would not try to fetch again from the network to get any updated characters if the latest revisionID was the same, so I added a one-time `forceRefresh` boolean to each migrated talk page to force it to refresh from the network again.

**Commit 3**
I came across an existing bug on a Serbian user talk page. There was an issue where posting a reply to a topic resulted in doubled-up replies in the UI once the screen refreshed. Digging in it looks like it was simply a copy-paste issue, changing the cast here to `TalkPageReply` allows this method to delete any old local replies that are no longer returned from the endpoint.

### Test Steps
**Reloading**
1. Fresh install app. In Search on Chinese Simplified, type "User talk:", and choose a user that loads a list of topics. Tap into one of the topics, then tap back to confirm the unread indicator is gone.
2. Go to the language picker and choose Chinese Traditional. Confirm the talk page reloads and reflects the new character set. Confirm the unread indicator is displayed on the topic you had selected if the characters in the topic are different.

**Updating**
1. Search and find a user talk page you own (I've been searching "User talk: TSevener (WMF)" and updating that talk page). Switch to Serbian Latin or Cyrillic. Post a new topic and/or a new reply to a topic with characters in that variant or the alternative variant. Confirm that when the page loads, it reflects the new topic/reply in the expected variant character set, regardless of which characters you used when posting. Switch language variants to the other using the language picker. Confirm you see your new topic/reply in the other character set.

**Migration**
1. On simulator/device, set up device language preferences to contain a Chinese Traditional language variant. Check out `releases/6.7.4`and fresh install app.
2. Search and visit a Chinese user talk page. Tap into a topic to remove the unread indicator.
3. Check out this branch and revisit that talk page under Chinese Traditional. Confirm talk page looks the same and unread indicator remains removed.

**Migration and Forcing Refresh**
4. Repeat steps 1 and 2 above under Migration.
5. Before checking out this branch, change device language preferences to Chinese Simplified.
6. Check out this branch and revisit that talk page under Chinese Simplified. Confirm that the page loads the new simplified characters and unread indicator is back to showing, because the character set has changed. Note you may briefly see the old traditional characters, because this screen is set up to show whatever was last in the managed objects while it's fetching from the network.


WMF Framework/Wikipedia.xcdatamodeld/Wikipedia 6.xcdatamodel/contents
```diff
@@ -73,6 +73,7 @@
     <entity name="TalkPage" representedClassName="TalkPage" syncable="YES" codeGenerationType="class">
         <attribute name="dateAccessed" attributeType="Date" defaultDateTimeInterval="582233760" usesScalarValueType="NO"/>
         <attribute name="displayTitle" attributeType="String"/>
+        <attribute name="forceRefresh" attributeType="Boolean" defaultValueString="NO" usesScalarValueType="YES"/>
```
dempseyatgithub, line 76: It seems a shame to have to add a permanent property to the model to handle a one-time migration.

Do you foresee other uses for this property?

If not I was wondering if it might make sense to do something a little sneaky like set the minutes and seconds of the dateAccessed to "33" minutes and "33" seconds (or '42' in honor of Douglas Adams), and set and check for those date components to refresh instead of the boolean.

It looks like you use the dateAccessed for cleanup purposes, but that shouldn't significantly affect sorting and maybe occasionally the rare talk page gets re-fetched unnecessarily.

Or I might be overthinking just to save a Bool.

Alternately, could there be a way to refetch as part of the migration?

Or - a user defaults key that just holds the appropriate record IDs, and they get checked and refreshed instead of the permanent bool on every record? (I am liking this second idea better.)

dempseyatgithub: (Or a list of revision IDs)

tonisevener: @dempseyatgithub good ideas! Totally agreed on it feeling unnecessary - I don't foresee needing to keep track of a separate bool per talk page record in the future. I like the user defaults, I'll try that (though I must admit `dateAccessed` tinkering has the allure of not having an additional user defaults flag in the codebase). I'd rather not fetch on migration just because of the small chance that they don't have a connection in that case and/or the unnecessary traffic with fetching pages they may not intend to visit again.

tonisevener: @dempseyatgithub I have posted a commit for using the user defaults method. I looked into `dateAccessed` first but found that it was being reset before the `forceRefresh` check ( within `userDidAccess`) which undid the seconds/nanoseconds specification I had made, and I didn't feel comfortable changing around the area of that reset at this point.

dempseyatgithub (approved):
The user defaults changes look good! I completely understand about not wanting to change with the date logic just to do something that might have been a bit on the 'clever' side to begin with.

This is clean, understandable and could still be removed at some point in the future.

PR #3893, opened 2021-03-26
Fix language variant alert issues

dempseyatgithub:
**Additional work to finish Phabricator ticket:** 
https://phabricator.wikimedia.org/T270210

### Notes
* Added a user dismissal completion block to the language view controller.
The block is called in all three cases where the user could dismiss the language view controller, including handling dismissal via accessibility gesture.

* Add a method to present the preferred language settings and not use a new NSUserActivty type
* Remove the WMFUserActivityTypeLanguageSettings

* Commit 4 is the largest commit. It turns the existing alert presentation into a function that takes a completion block. Instead of passing it in as the `resumeApp:` completion block, it is called as part of the `resultApp:` completion block before `finishResumingApp` is called. It essentially adds one more phase in resuming just before the finish step.

This commit also plumbs the completion block through methods called by `presentLanguageVariantAlerts()` so that every path does eventually call the completion block.

* The last commit adds a check to ensure the user is not in the middle of a deep link, and the alerts would appear on top of existing presented controllers or nested view controllers within a navigation controller.

### Test Steps
The issues described in https://github.com/wikimedia/wikipedia-ios/pull/3888 should be addressed.

I did not see any additional issues, but will continue to try different things.


Wikipedia/Code/WMFLanguagesViewController.m
```diff
@@ -480,7 +480,7 @@ - (void)addLanguageButtonTapped {
 - (void)languagesController:(WMFLanguagesViewController *)controller didSelectLanguage:(MWKLanguageLink *)language {
     [MWKDataStore.shared.languageLinkController appendPreferredLanguage:language];
     [self reloadDataSections];
-    [controller dismissViewControllerAnimated:YES completion:NULL];
+    [controller dismissViewControllerAnimated:YES completion:self.userDismissalCompletionBlock];
```
tonisevener, line 483: I found that this caused the completion to be called after the user tapped a language from the "Add another language" screen. It just felt like a step too soon, because the `WMFPreferredLanguagesViewController` was still on screen when `finishResumingApp` kicked off, potentially causing whatever followup modals from that sequence to present on top of that screen. I feel like we should keep this as NULL.

Wikipedia/Code/WMFLanguagesViewController.m
```diff
@@ -387,7 +387,7 @@ - (void)searchBarSearchButtonClicked:(UISearchBar *)searchBar {
 #pragma mark - UIAccessibilityAction
 
 - (BOOL)accessibilityPerformEscape {
-    [self dismissViewControllerAnimated:YES completion:nil];
+    [self dismissViewControllerAnimated:YES completion:self.userDismissalCompletionBlock];
```
tonisevener, line 390: Good catch on this one!

tonisevener (commented):
Looking good! Just one small thing I found, I think we'll be good to merge after that.

Wikipedia/Code/WMFLanguagesViewController.m
dempseyatgithub: I put it back to NULL in this case and updated the method comment in the header.

tonisevener (approved):
🎉

PR #3894, opened 2021-03-26
Apply user's preferred language variant to incoming links when applicable

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T278581

### Notes
* Applies user's preferred language variant code to incoming URLs if they match the site URL of the incoming link.

### Test Steps
When tapping a top read widget row element, it should open in the app in your primary (preferred) language/variant.

1. Add a language with variants (Chinese, Serbian, etc.) as your primary app language
2. Add Top read widget to home screen
3. Tap a top read element on widget, confirm the app opens to that article in your current primary language variant
4. Update your variant to a different variant in the same language
5. Repeat step 3


Wikipedia/Code/WMFAppViewController.m
```diff
@@ -1211,15 +1211,19 @@ - (BOOL)processUserActivity:(NSUserActivity *)activity animated:(BOOL)animated c
...
-                    preferredLanguagesViewController];
+            WMFPreferredLanguagesViewController *languagesVC = [WMFPreferredLanguagesViewController preferredLanguagesViewController];
             languagesVC.showExploreFeedCustomizationSettings = YES;
             languagesVC.delegate = self.settingsViewController;
             [languagesVC applyTheme:self.theme];
             [self showSettingsWithSubViewController:languagesVC animated:animated];
         } break;
         default: {
             NSURL *linkURL = [activity wmf_linkURL];
+            MWKLanguageLink *preferredLanguage = self.dataStore.languageLinkController.appLanguage;
```
dempseyatgithub, line 1222: It would be better to use -preferredLanguageVariantCodeForLanguageCode: here.

You can call:
`linkURL.wmf_languageVariantCode = [linkController.preferredLanguageVariantCodeForLanguageCode: linkURL.wmf_language]` and it will do the correct thing if `wmf_language` is nil, if the language code doesn't have a variant, etc.

dempseyatgithub: Also, it will handle any language that comes in, even though we're pretty sure it will be in the user's primary language.

dempseyatgithub (changes requested):
I added a note inline of a preferred method to use to get the language variant.

Wikipedia/Code/WMFAppViewController.m
staykids: What good is having new API if we don't even use it? 😉 Thanks for flagging this – updated.

dempseyatgithub (approved):
The updated change looks good!

Wikipedia/Code/WMFAppViewController.m
```diff
@@ -1213,6 +1213,8 @@ - (BOOL)processUserActivity:(NSUserActivity *)activity animated:(BOOL)animated c
         } break;
         default: {
             NSURL *linkURL = [activity wmf_linkURL];
+            // Ensure incoming link is fetched in user's preferred variant if applicable
+            linkURL.wmf_languageVariantCode = [self.dataStore.languageLinkController preferredLanguageVariantCodeForLanguageCode:linkURL.wmf_language];
```
tonisevener, line 1217: @staykids sorry to resurface this but, I notice this area is hit when tapping a link in any article, as well as from search results, so we might get `wmf_languageVariantCode` set here in an improper way.

Example:

On an app with only Simplified Chinese in app language settings, go to an article in Chinese Simplified. Then select the language switcher in the article and choose Traditional Chinese (Hong Kong). During that transition this line changes the URL's `wmf_languageVariantCode` from `zh-hk` to 'zh-hans'. Is there a way to wrap this in some check where we know for sure they are coming from the widgets?

staykids: @tonisevener Great catch, and thanks for surfacing this!

Hmm, so the widgets just pass a URL (with no additional info) to the host app for it deal with, which then goes through the whole process of parsing it into an appropriate `WMFUserActivityType`, and ultimately ends up in this function. We probably want to retain that flow.

Some ideas that immediately come to mind:

1. We could create a new `WMFUserActivityType` for the widget links. We'd have to inject a parameter into the widget's URL that indicates it's from the widget, parse it in the host app to the new type and populate the variant there, and handle it here in a different switch case so it doesn't affect our default link parsing. The unknown to me is if there are any conformance gotchas in creating a new `WMFUserActivityType` or if it's pretty straightforward.

2. We could skip the new `WMFUserActivityType`, just inject the URL parameter widget-side, and parse and only inject the variant here in this case statement if it contains that "from the widget" URL parameter. This path is more lightweight, but makes this switch case slightly less direct (in that it'd include the logic to determine if it's from a widget or not before injecting the variant).

Do you want me to try any of these (or have ideas for something else I'm missing)?

tonisevener: @staykids Hm well I just tested and it looks like `wmf_languageVariantCode` is missing via standard deep linking too, like if you search a ZH article in Safari, so it would be nice to catch those while we're here. Maybe in AppDelegate's `openURL` method, you can flag in the url via a query item to indicate that it will need the language variant code assigned based on the user's app language preferences down the line. Then check for the query item flag here in this `processUserActivity` `WMFAppViewController` class before assigning.

dempseyatgithub: @staykids @tonisevener Would checking to see if the URL already has a `wmf_langaugeVariantCode` and leaving it as-is if it does be enough to solve the issue?

The existing `preferredLanguageVariantCodeForLanguageCode:` will not add a variant code for languages that don't support variants, and URLs coming from within the app should have the variant code properly set if needed.

dempseyatgithub: So doing:
```
if (!linkURL.wmf_languageVariantCode) {
    linkURL.wmf_languageVariantCode = 
            [self.dataStore.languageLinkController preferredLanguageVariantCodeForLanguageCode:linkURL.wmf_language];
}
```

tonisevener: @dempseyatgithub great point! That sounds good to me, I'm all for keeping it simple.

tonisevener: FYI I did some initial testing on that simple check, it looked good so I went ahead and made a PR for this and deployed the latest beta from that branch: https://github.com/wikimedia/wikipedia-ios/pull/3905

PR #3895, opened 2021-03-27
Update top read widget to be language variant aware

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T277697 (this PR is a stepping stone towards addressing this task)

### Notes
Previously, the top read widget would get into states where the title and description would be a. displayed in contrasting variants from one another and b. inconsistent with what's displayed in the Explore feed in the app. For example, if you were using `Serbian (Latin)` as your primary language, you might see a `Serbian (Latin)` description but a `Serbian (Cyrillic)` title.

While it's certainly possible you may still see mixed character sets (for backend content reasons), this PR seeks to reach some sense of parity between the top read widget's displayed titles and descriptions and the in-app Explore feed's top read titles and descriptions.

- Updates the top read widget to use variant aware URLs
- Populates the widget elements' title and description from the fetched `WMFArticle` instead of the `WMFFeedTopReadArticlePreview`

### Test Steps
I don't have great test steps here because getting the widgets to reload as desired has always been problematic, but generally:

1. Make your preferred language one with many variants
2. Observe the top read widget title and description, ensure they match the in-app Explore feed's top read elements (specifically, that the title and description are using the same variant as the Explore feed's top read group)
3. Change to a different variant, repeat step 2.

Sometimes you'll encounter cases of the widgets not reloading, in which case sometimes deleting and re-adding the widget helps. The intent of this PR is not so much to address the reloading problem, but to help address the consistency issues of the titles and descriptions as described in step 2.

top_read


Widgets/Widgets/TopReadWidget.swift
```diff
@@ -54,14 +54,19 @@ final class TopReadData {
...
         let groupURL = topRead.url
         let isCurrent = topRead.isForToday // even though the top read data is from yesterday, the content group is for today
         var rankedElements: [TopReadEntry.RankedElement] = []
+
         for article in results {
+            article.articleURL.wmf_languageVariantCode = languageVariant
             if let articlePreview = dataStore.fetchArticle(with: article.articleURL) {
                 if let viewCounts = articlePreview.pageViewsSortedByDate {
-                    rankedElements.append(.init(title: article.displayTitle, description: article.wikidataDescription ?? article.snippet ?? "", articleURL: article.articleURL, thumbnailURL: article.thumbnailURL, viewCounts: viewCounts))
+                    let title = articlePreview.displayTitle ?? article.displayTitle
```
dempseyatgithub, line 67: I was initially confused because the PR notes say that the title and description of the article would be preferred over those values in the article preview, but the code seems to indicate the opposite. I see that the logic is correct, but the variable names seem to indicate the opposite.

Maybe consider swapping the names 'article' and 'articlePreview' so that 'articleResult' refers to the `WMFFeedTopReadArticlePreview` and 'article' refers to the 'WMFArticle'.

Widgets/Widgets/TopReadWidget.swift
```diff
@@ -54,14 +54,19 @@ final class TopReadData {
...
+                    let description = articlePreview.wikidataDescription ?? article.wikidataDescription ?? articlePreview.snippet ?? article.snippet ?? ""
+                    rankedElements.append(.init(title: title, description: description, articleURL: article.articleURL, thumbnailURL: article.thumbnailURL, viewCounts: viewCounts))
```
dempseyatgithub, line 69: I'm not personally a big fan of .init because often it's difficult to know what type of thing is being initialized just by reading the call site. That's not to say change it - more of a stylistic preference.

dempseyatgithub (changes requested):
I had a number of comments.

It also makes me think that I need to create a document with some details about the variant feature, because everything you've done is reasonable, but should be able to be done more simply.

It makes me realize I haven't given the team anything to help you all reason about doing variant-specific things.

WMF Framework/WidgetController.swift
```diff
@@ -54,18 +54,21 @@ public final class WidgetController: NSObject {
...
         }
     }
     
-    private func fetchCachedWidgetContentGroup(with kind: WMFContentGroupKind, isAnyLanguageAllowed: Bool, in dataStore: MWKDataStore, completion:  @escaping (WMFContentGroup?) -> Void) {
+    private func fetchCachedWidgetContentGroup(with kind: WMFContentGroupKind, isAnyLanguageAllowed: Bool, shouldApplyPreferredLanguageVariant: Bool = false, in dataStore: MWKDataStore, completion:  @escaping (WMFContentGroup?) -> Void) {
         assert(Thread.isMainThread, "Cached widget content group must be fetched from the main queue")
         let moc = dataStore.viewContext
-        let siteURL = isAnyLanguageAllowed ? dataStore.primarySiteURL : nil
+        var siteURL = isAnyLanguageAllowed ? dataStore.primarySiteURL : nil
+        if shouldApplyPreferredLanguageVariant {
```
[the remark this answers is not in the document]
staykids: I pulled out the variant setting stuff here altogether, because it looks like it may be unneeded. Also, the Top read and On this day widget don't appear to trigger the path where `siteURL` is populated here.

Widgets/Widgets/TopReadWidget.swift
staykids: The naming here is indeed a mess. Might as well use this as an opportunity to clean it up. Updated.

staykids:
Updated and should be ready for another look.

dempseyatgithub (approved):
The changes look good!

PR #3898, opened 2021-03-30
Fix bug with non-norwegian altISOCodes being populated

tonisevener:

### Notes
I found a case where `altISOCode` was populated on a Chinese `MWKLanguageLink` object, which was unexpected since it was only meant for Norwegian. This fixes that hasty PR feedback and adds some missing checks in some of the comparison methods.

See for reference - https://github.com/wikimedia/wikipedia-ios/pull/3886.

### Test Steps
Set a breakpoint and confirm `altISOCode` in `MWKLanguageLink` is only ever populated to "nb" for the "no" language code.


mcleinman (approved):
Works great, tests great. Thanks!

PR #3899, opened 2021-03-30
Remove zh-hant and zh-cn as variant options, tweak names

tonisevener:
**Fixes Phabricator ticket:** 
https://phabricator.wikimedia.org/T276280#6902051
https://phabricator.wikimedia.org/T276280#6947417

### Notes
* Here I am making tweaks to the json files to remove `zh-hant` and `zh-cn` variants, as requested by @carolynlimadeo. I am also adjusting the wording slightly to so the English subtitles on the language list screen match her comment [here](https://phabricator.wikimedia.org/T276280#6902051).

We already have some translations in for the `chinese-variants-alert-body` key that translated all 8 variants, so I am proactively going into the `.strings` files and removing them (thanks to the "\n"s and variant codes it's pretty easy to tease these out). As soon as this is merged I plan to go into translatewiki.net and update the translations there so they don't get rewritten with 8 variants the next time our auto-magical setup creates a PR.

### Test Steps
Do some light regression testing on: 
1. Variant choices presented to the user when fresh installing, confirm it's 6 and not 8.
2. Upgrade from 6.7.4 on a device that has Chinese Traditional in it's device language settings - confirm nothing seems broken and you see the variant modal listing the 6 options.


WMF Framework/MediaWikiAcceptLanguageMapping.json
```diff
@@ -11,16 +10,6 @@
...
         "hans": {
             "default": "zh-hans",
-            "cn": "zh-cn",
-            "hk": "zh-hk",
-            "mo": "zh-mo",
-            "my": "zh-my",
-            "sg": "zh-sg",
-            "tw": "zh-tw"
-        },
-        "hant": {
```
dempseyatgithub, line 21: This mapping is going from OS locale identifiers to Wikipedia variants.

So, hant is still a valid OS locale identifier which we need to handle. It's where we map to the correct variants of Chinese, Traditional, which is still needed.

I've added to the ticket, since we also need to choose which variant of Traditional Chinese we will use if the user has 'Chinese, Traditional' selected. (It seems wrong to choose simplified if the user has clearly preferred traditional.)

(I wish you could put comments in json!)

dempseyatgithub (changes requested):
Overall, this looks good.

We need to restore the 'hant' dictionary. My comment goes into more detail.

And we need to find out which Traditional Chinese variant we should use if the user has "Chinese, Traditional" selected in the OS.

tonisevener:
@dempseyatgithub this is ready for another look. I am defaulting to `zh-hk` for a Chinese Traditional device language per https://phabricator.wikimedia.org/T276280#6961508.

WMF Framework/MediaWikiAcceptLanguageMapping.json
dempseyatgithub: The changes look good.

dempseyatgithub (approved):
These changes look good and resolve to Hong Kong Traditional as described in the Phab ticket.

PR #3903, opened 2021-04-02
Fix remote notification API errors for variants

tonisevener:
**Fixes Phabricator ticket:** https://phabricator.wikimedia.org/T277608

### Notes
This fixes the API errors around remote notifications and variants. Let's lean on QA to regression test any user-facing side of this.

### Test Steps
1. Go to `RemoteNotificationsOperationsController.swift`. Comment out this chunk, this will allow you to see the notifications endpoint call more often:
```
//        guard deadlineController?.isBeforeDeadline ?? false else {
//            completeEarly()
//            return
//        }
```
2. Run app, add app languages with variants in Settings. Inspect the notifications call (...`api.php?action=query&meta=notifications`...), confirm the query item for `notwikis` uses the Wiki language only for it's value, not the entire variant code: `notwikis=srwiki|enwiki|zhwiki|wikidatawiki`. In `main` it sends `notwikis=enwiki|zh_hanswiki|sr_ecwiki|wikidatawiki` and in the response we get the warning:

```
{
  "batchcomplete": true,
  "warnings": {
    "notifications": {
      "warnings": "Unrecognized values for parameter \"notwikis\": zh_hanswiki, sr_ecwiki"
    }
  },
```

3. On `main`, I would occasionally see a token error when it tried to mark a notification as read. The call looks like this:

POST https://www.mediawiki.org/w/api.php?action=echomarkread&format=json&list={{notificationID}}&wikis=enwiki
Body: token=%3CWMFToken:%200x600002e65500%3E

Response:

```
{
  "error": {
    "code": "badtoken",
    "info": "Invalid CSRF token.",
    "*": "See https://www.mediawiki.org/w/api.php for API usage. Subscribe to the mediawiki-api-announce mailing list at &lt;https://lists.wikimedia.org/mailman/listinfo/mediawiki-api-announce&gt; for notice of API deprecations and breaking changes."
  },
  "servedby": "mw1285"
}
```

Fixing that one line in `RemoteNotificationsAPIController.swift` to `token.value` fixed this API error. It's a little difficult to test this. I would:

1. Make a change to an article from the app on EN Wikipedia. 
2. On desktop, log into a different account, go to that article, go to history and click "undo" next to the revision you just made from the device.
3. Wait a while on device watching the calls, eventually after a minute or two you will see a call that tries to mark that edit revert notification as read.


Wikipedia/Code/MWKLanguageLinkController.m
```diff
@@ -232,6 +232,16 @@ - (void)getPreferredLanguageCodes:(void (^)(NSArray<NSString *> *))completion {
     }];
 }
 
+- (void)getPreferredWikiCodes:(void (^)(NSArray<NSString *> *))completion {
```
tonisevener, line 235: I'm open to an alternative name if "WikiCodes" just adds to the confusion.

dempseyatgithub: I think the naming issue is a symptom of my not trying to update existing method names to the new `langugageCode` / `languageVariantCode` / `contentLanguageCode` naming.

(I was going to suggest that for either the dot release or the clean up period that Josh mentioned that I make those changes).

So, the existing method should be `getPreferredContentLanguageCodes` and the new method should be `getPreferredLanguageCodes`.

Maybe the name `WMFPreferredLanguageInfoProviding` would work better. I think having 'language' in the name is better. We got rid of the MWKLanguageInfo type during refactoring and 'preferred language info' can contain any info about preferred languages.

Would `WMFPreferredLanguageInfoProvider` work. I always find 'Provider' to be better than 'Providing' especially since the Swift naming guidelines seem to prefer using a noun like 'Provider' if the item *is* a thing.

Wikipedia/Code/MWKLanguageLinkController.m
```diff
@@ -232,6 +232,16 @@ - (void)getPreferredLanguageCodes:(void (^)(NSArray<NSString *> *))completion {
...
+    NSArray<NSURL *> *preferredSiteURLs = [self preferredSiteURLs];
```
dempseyatgithub, line 236: I'm curious, is there are reason to go through preferredSiteURLs instead of going through preferredLanguages collecting the languageCode of each?

dempseyatgithub (commented):
I mainly had a lot of naming comments, but also one implementation question.

Wikipedia/Code/MWKLanguageLinkController.m
tonisevener: For whatever reason I had thought going through siteURLs saved me a step, but now I don't see how I thought that 😅 I'll change, I think it makes more sense to go your way.

dempseyatgithub (approved):
The changes look good.

PR #3909, opened 2021-04-07
Ensure ISO code 'nb' becomes wiki language code 'no' within the app

dempseyatgithub:
**Fixes Phabricator ticket:** 
This is a refinement of the fix for https://phabricator.wikimedia.org/T276645

### Notes
This change ensures two things:
1. During onboarding, if the user has Norwegian set as a language, the Locale identifier 'nb' will be mapped to the wikimedia language code of 'no'

2. When migrating from a previous version of the app, various language-related settings are being updated to support language variants. Essentially the language code, such as 'zh' is being replaced with a variant language code such as 'zh-hans'.

This change adds the mapping "nb"->"no" to ensure any setting that use the ISO code are updated to use the wikipedia language code.

- Add mapping from locale 'nb' to wikimedia language code 'no'
- Add mapping from 'nb' to 'no' as part of language variant settings migration
- Fix typo in description string of MWKLangaugeLink.m

### DIFF
--- a/WMF Framework/MWKDataStore+LanguageVariantMigration.swift
+++ b/WMF Framework/MWKDataStore+LanguageVariantMigration.swift
@@ -52,9 +52,13 @@ extension MWKDataStore {
             result[languageCode] = languageVariantCode
         }
         
-        languageLinkController.migratePreferredLanguages(toLanguageVariants: migrationMapping, in: moc)
-        feedContentController.migrateExploreFeedSettings(toLanguageVariants: migrationMapping, in: moc)
-        migrateSearchLanguageSetting(toLanguageVariants: migrationMapping)
+        // Ensure any settings that currently use 'nb' are updated to use 'no'
+        var languageCodeMigrationMapping = migrationMapping
+        languageCodeMigrationMapping["nb"] = "no"
+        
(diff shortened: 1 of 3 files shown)


tonisevener:
Thanks for catching this! The only unexpected difference I've noticed (on a simulator with only Norwegian Bokmal as a device language) is that now `Accept-Language` headers send as `no` whereas on `main` and 6.7.4 (after manually adding an app language after onboarding) it sent as `nb_us` for me. Since this isn't a Wiki language that supports variants I don't think it matters to the server on the backend, but just a change I thought worth noting.

tonisevener (approved):
This looks good to me in my testing.

dempseyatgithub:
> Thanks for catching this! The only unexpected difference I've noticed (on a simulator with only Norwegian Bokmal as a device language) is that now `Accept-Language` headers send as `no` whereas on `main` and 6.7.4 (after manually adding an app language after onboarding) it sent as `nb_us` for me. Since this isn't a Wiki language that supports variants I don't think it matters to the server on the backend, but just a change I thought worth noting.

For non-variant languages the app has sent in the 'Accept-Language' header that lists all languages in decreasing priority, but which the servers don't seem to use. So, if the app had English as a primary language and Norwegian as the second preferred language, "en" would be the first language included.

In my testing, it seems that for sites without language variants the correct language is returned regardless of what value "Accept-Language" contains.

PR #3912, opened 2021-04-09
Update search languages bar button design

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T277058

### Notes
Updates based on feedback from Carolyn:
- Removes the horizontal scrolling indicator which was previously clashing with the selected element's underline
- Updates the selected and deselected state color schemes
- Truncates displayed language codes to 4 characters max. The approach I took also included removing any trailing punctuation (something like `map-` looked odd). At least based on our current list of `languageCode`s, this approach seems ok.

### Test Steps
1. Confirm design of `SearchLanguageButton`s matches Phabricator task.
2. Confirm no language code "box" has more than 4 characters in it, and that they do not end with punctuation.
3. Confirm no horizontal scrolling indicator is visible when scrolling the languages bar.

### Screenshots
languages

### DIFF
--- a/Wikipedia/Code/SearchLanguagesBarViewController.swift
+++ b/Wikipedia/Code/SearchLanguagesBarViewController.swift
@@ -90,9 +90,10 @@ class SearchLanguageButton: UnderlineButton {
     // MARK: - Configuration
 
     func apply(theme: Theme) {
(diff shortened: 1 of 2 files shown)


tonisevener (approved):
Looks great!

PR #3914, opened 2021-04-13
allow scrolling to final screen in a scrollview

mcleinman:
**Fixes Phabricator ticket:** https://phabricator.wikimedia.org/T280063

### Notes
* [This PR](https://github.com/wikimedia/wikipedia-ios/pull/3664) introduced a bug. We checked if the scroll offset was reasonable, and didn't scroll if it wasn't. However, we were discarding any scroll offsets in the final screen of a scrollview - such as `Read More`. 

### Test Steps
1. Run app. Scroll to "Read More" section via Table of Contents. Scroll back to an earlier section via Table of Contents. Ensure scrolls work.

### DIFF
--- a/Wikipedia/Code/ArticleScrolling.swift
+++ b/Wikipedia/Code/ArticleScrolling.swift
@@ -70,14 +70,16 @@ extension ArticleScrolling where Self: ViewController {
         } else {
             adjustmentY = overlayTop
         }
-        let y = offset.y + adjustmentY
         let minY = 0 - scrollView.contentInset.top
-        guard y > minY else {
+        let maxY = scrollView.contentSize.height - scrollView.bounds.height + scrollView.contentInset.bottom
+
+        /// If y lies within the last screen, scroll should be to the final full screen.
+        let y = min(offset.y + adjustmentY, maxY)
+        guard y >= minY else {
             completion?(false)
             return
         }
-        let maxY = scrollView.contentSize.height - scrollView.bounds.height + scrollView.contentInset.bottom
-        guard y < maxY else {
+        guard y <= maxY else {
             completion?(false)
             return
         }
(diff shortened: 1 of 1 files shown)


Wikipedia/Code/ArticleScrolling.swift
```diff
@@ -70,14 +70,16 @@ extension ArticleScrolling where Self: ViewController {
         } else {
             adjustmentY = overlayTop
         }
-        let y = offset.y + adjustmentY
         let minY = 0 - scrollView.contentInset.top
-        guard y > minY else {
+        let maxY = scrollView.contentSize.height - scrollView.bounds.height + scrollView.contentInset.bottom
+
+        /// If y lies within the last screen, scroll should be to the final full screen.
+        let y = min(offset.y + adjustmentY, maxY)
```
tonisevener, line 77: @mcleinman nice catch, I didn't realize that PR had caused it! The only regression I notice is the the navigation state restoration doesn't scroll as well as it does in `main`. For example, go to any article, tap TOC and scroll to a section in the middle. Background, then terminate. Relaunch, and upon state restoration it scrolls to an area much higher. Digging in it seems like for whatever reason `maxY` is much smaller at that point in the restoration sequence.

ArticleVC's `articleLoadWaitGroup` seems to be populated at this point only for state restoration, so changing this line to something like `let y = articleLoadWaitGroup != nil ? offset.y + adjustmentY : min(offset.y + adjustmentY, maxY)` fixes it.

tonisevener (commented):
Thanks! Looking good, found one regression and a possible light-touch fix, though I don't think it is necessarily a blocker on merging. Let me know what you think.

Wikipedia/Code/ArticleScrolling.swift
mcleinman: Good catch, thanks! I also realized that the `guard y <= maxY` is useless when `y = min(blah, maxY)` and so updated this code to be a bit more specific on when we override to maxY, and also be more readable (hopefully). Seems to work for state restoration as well as "scroll to bottom section". Let me know if you see any problems, or think the code readability could be improved.

tonisevener (approved):
Looks good, thanks!

PR #3915, opened 2021-04-14
get tyap working, update codemirror

mcleinman:

### Notes
* Gets Tyap working in the app.
* I misspoke in a meeting - the app permits translations even if there is not a wiki. (I was looking at another part of how we handle languages.) In this case, we had to follow step 3 of the [languages documentation](https://github.com/wikimedia/wikipedia-ios/blob/main/docs/languages.md), which I've updated in this PR to make even clearer.
* Most of this is codemirror stuff. The key pieces are the project file and `Wikipedia/iOS Native Localizations/kcg.lproj/InfoPlist.strings`.

### Test Steps
1. Add `Tyap` to a device (or sim's) preferred languages, and place it above English.
2. Run app, ensure you see Tyap.

### Screenshots
Screenshot 2


mcleinman:
> Apologies if I'm missing it, but the CodeMirror config update stuff here isn't by some manual process, but by using the script generated by `WikipediaLanguageCommandLineUtility.swift` right?

Yes, it's from an automated part of script.

staykids (approved):
I think there's some kind of localization caching strangeness with iOS that was throwing me for a loop for a while, where 1. switching to `add-tyap` and installing then deleting the app and then 2. switching back to `main` and installing showed Tyap working on the `main` branch build. It _seems_ like erasing all content and settings on the Simulator cleared up that weirdness.

This appears to work as expected - that is, app chrome displays translated strings localized in Katab/Tyap as in the original screenshot. 👍🏽

PR #3917, opened 2021-04-15
Rename wmf_language

dempseyatgithub:
**Fixes Phabricator ticket:** 
https://phabricator.wikimedia.org/T280266

### Notes
This PR is renaming properties and argument names to more clearly differentiate _language_ which is an instance of MWKLanguageLink and _languageCode_ which is a language code string.

The first two commits change the property name and call sites, first on NSURL, then URL.

The second two commits change argument names and call sites, first on NSURL then on NSURLComponents.

Note. I originally thought this would be more than one PR, but I've continued to add new commits to it.

Each commit message should be self-explanatory, but essentially, it is changing the code base to adhere to the 'languageCode' naming convention.

### Test Steps
1. No change in app behavior
2. All existing tests should pass


WMF Framework/MWKDataStore+LanguageVariantMigration.swift
```diff
@@ -65,7 +65,7 @@ extension MWKDataStore {
     private func migrateSearchLanguageSetting(toLanguageVariants languageMapping: [String:String]) {
         let defaults = UserDefaults.standard
         if let url = defaults.url(forKey: WMFSearchURLKey),
-           let languageCode = url.wmf_language {
+           let languageCode = url.wmf_languageCode {
             let searchLanguageCode = languageMapping[languageCode] ?? languageCode
             defaults.wmf_setCurrentSearchContentLanguageCode(searchLanguageCode)
```
tonisevener, line 70: I have noticed a few instances of method arguments named with "contentLanguageCode", yet we are passing in a wmf_languageCode and not wmf_contentLanguageCode. Could that cause any bugs?

You can do a find on this PR for "contentLanguageCode" to see other placements, I decided against commenting on all of them.

WMF Framework/Localization/WMFLocalization.h
```diff
@@ -3,7 +3,7 @@
 NS_ASSUME_NONNULL_BEGIN
 
 // Don't rename this function. It'll break the script that generates the strings. https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/LoadingResources/Strings/Strings.html
```
tonisevener, line 5: I haven't dug into what this means but just wanted to highlight this comment warning against renaming this.

tonisevener (commented):
Code review done - found a couple of things that could be fine with some investigation, but just wanted to point out. I'll do some light regression testing later this afternoon.

WMF Framework/MWKDataStore+LanguageVariantMigration.swift
dempseyatgithub: The case I am aware of is the refactored RTL methods use `contentLanguageCode` in the name, but we pass in language code.

Fixing that is essentially the core of the Phab ticket for supporting the Arabic variant of Kazakh.

In this case, the existing search language setting will only have language codes, and if there is no migration to a variant the original code is the correct value.

When I work on the RTL issue, I can also do a double check that we aren't passing in the wrong thing elsewhere when contentLanugageCode is an argument. But, generally, since these are all naming changes, they should all be correct as-is, except for the RTL methods.

WMF Framework/Localization/WMFLocalization.h
dempseyatgithub: I believe this is because `genstrings` is run on the code to generate the strings files. (For a while, I owned `genstrings` when I worked on AppKit!)

Only the name of the function is used - all custom localized string functions use the same signature. 

Changing the name of the argument doesn't change the signature of the function, so this should not break anything.

tonisevener (approved):
Looks good!

PR #3919, opened 2021-04-19
Scrolling: Ensure range is valid to fix crash

mcleinman:
**Fixes Phabricator ticket:** N/A (crash found in beta)

### Notes
* All ranges in this function should now be valid.
* I'd love to refactor this into some functions for readability, but given that this is a last fix before a release, that seems too risky.

### Test Steps
1. Run app. Scroll down midway through an article.
2. Background app. Quit app.
3. Reopen app. Ensure article is at same scroll position.


tonisevener (approved):
👍

PR #3920, opened 2021-04-20
Analyzer fixes

tonisevener:

### Notes
* I ran some analyzers for the 6.8 pre-release testing. This PR addresses some of those findings.

**Commit 1**
Addressing a couple of issues that popped up after running Product > Build > Analyze. I ignored the issues in 3rd party libraries.

**Commit 2**
I set `-com.apple.CoreData.ConcurrencyDebug 1` as a launch argument and fixed a couple of Core Data threading complaints from that.

### Regression Test Steps
Just some light regression testing in these areas I think will be enough:
1. Searching an item and the initial Explore feed population should hit the `WMFMTLModel.m` change.
2. Visit and article and save it. This should trigger the `SavedPageSpotlightManager.swift` change.
3. Relaunch app. This should trigger the continued reading card change in `WMFContinueReadingContentSource.m` from your step 2.


staykids (approved):
`__Multithreading_Violation_AllThatIsLeftToUsIsHonor__`. 👍🏽 Changes make sense I believe, during testing hits the breakpoints you specified, and haven't run into any unexpected behavior yet.

I put a few tiny suggestions [in this branch](https://github.com/wikimedia/wikipedia-ios/commits/checker-fixes-suggestions) if any were interesting to you to cherry pick, but we can just as easily save them for later. So feel free to go ahead and merge this as-is now if you prefer.

1. https://github.com/wikimedia/wikipedia-ios/commit/e5e72c31d5ddabe5f6cb1ba303d7121315126d91 - fixed typo with method name
2. https://github.com/wikimedia/wikipedia-ios/commit/6eebb8304f0efa9de187426e175804fd160982f7 - removed unused `removeFromIndex` method (though it's probably sensible to actually use) 
3. https://github.com/wikimedia/wikipedia-ios/commit/286e559e29e2531cbc6c707583234c3eba254c2a - this is perhaps the one needing more scrutiny, but maybe we can get away with removing this internal queue. The doc header of `CSSearchableItemAttributeSet` said it should only be mutated from a single thread at a time, so I can understand the initial serial queue intent. I have no meaningful experience with Core Spotlight though, so I'm unclear of best practices.

tonisevener:
@staykids thanks for the suggestions! I looked into it and I think removing the article from CoreSpotlight upon unsave was just a missed piece during our rewrite to mobile-html 😬  so I added it back in. Good catch on that, I was able to confirm the search result goes away from the spotlight results after unsaving an article.

I tracked the internal queue changes down to [this PR](https://github.com/wikimedia/wikipedia-ios/pull/2062). I think I'm going to leave it as-is - if there was some UI hanging at some point during a large sync, I'd hate to bring that back.

PR #3923, opened 2021-04-22
Resolve visual issue when transitioning in and out of search view controller

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T277786 (specifically, [this comment](https://phabricator.wikimedia.org/T238863#6949757))

### Notes
Resolves issue where entering and exiting the search view caused the entire screen to "flash". This was because the background color wasn't explicitly set. 

I still haven't been able to figure out a great solution to discarding the overlay entirely (which hides some of the scroll offset issues) that works well with empty and filled tables. But I think this PR is a definite improvement in that the overall view's background color stays consistent and only the collection view content itself fades in and out.

### Test Steps

Lightly check the same steps as https://github.com/wikimedia/wikipedia-ios/pull/3882 to ensure there aren't any regressions. Confirm the before fix GIFs below's "background flashing" does not occur in these scenarios for each of our themes:

1. From "Explore" tab, tap into search, then use left edge swipe gesture to interactively dismiss the controller.
2. From "Search" tab, tap Search then the Cancel button.

### Before fix GIFs

- Explore Feed: left edge interactive dismiss
explore_feed_search

- Search tab: Unexpected background color on appearance and dismissal
search_tab_search


tonisevener (approved):
Definitely better!

PR #3924, opened 2021-04-24
Automate release tagging in Jenkins

mcleinman:

### Notes
* This takes the last beta commit, and tags it as a release and marks it in GitHub as a release.
* I had to create a new GitHub token for Jenkins to get this to run, from the wmf-apps-ci account.
* Ran `bundle update`.

### Test Steps
0. Remove existing tags: On the `wmf2249` machine, run `git tag -d releases/6.8.1 && git push --delete origin "releases/6.8.1"`
1. On the build server, run `Tag last beta as release`. When asked for a branch, put in this one: `release_tag`.
2. Confirm that there is a releases tag for 6.8.0: https://github.com/wikimedia/wikipedia-ios/tags
3. Confirm that tag is listed as the latest release: https://github.com/wikimedia/wikipedia-ios/releases

*After this gets merged, I need to remove the option in Jenkins to choose a branch - as it will always be run on `main`.*


fastlane/Fastfile
```diff
@@ -189,31 +189,40 @@ platform :ios do
     sh "git push origin --tags"
   end
 
-  desc "Add a tag for the current version number push to repo."
+  desc "Add a release tag for the latest beta and push to repo."
   lane :tag_release do |options|
```
tonisevener, line 193: I wonder if it'd be nice to have an optional `sha` to pass in here that takes precedence over `get_latest_build_number()` that we could send in through the Jenkins job. With your work on [https://phabricator.wikimedia.org/T275167](https://phabricator.wikimedia.org/T275167) I could see a situation where the latest beta has been bumped up (due to us working ahead to the next release and merging things) before Apple approves the current release, so the wrong commit gets tagged.

But I'm also fine with waiting until we run into that situation first before adding complexity.

tonisevener (commented):
Great stuff! I haven't tested yet, but wanted to propose one thing before moving to that step.

fastlane/Fastfile
mcleinman: Done. And jenkins has a spot to put it in. If you leave it blank, it uses the last beta. Otherwise, you can put any valid tag or commit hash in. And invalid tags/commit hashes will cause the process to fail.

I also ran `bundle update` to update the gems.

tonisevener (approved):
Works like a charm!

mcleinman:
Thanks! Per the notes above, now that this is merged I've removed the option to select a branch from the Jenkins project, for simplicity.

PR #3927, opened 2021-04-26
Remove altISOCode check

dempseyatgithub:
**Fixes Phabricator ticket:** 
https://phabricator.wikimedia.org/T280269

### Notes
There are two main logical pieces to this PR:

1. Remove the `altISOCode` logic from the MWKLanguageLinkController `-preferredLanguages` method.

Since we took steps to ensure that the desired value 'no' is present in preferred languages settings and 'nb' never makes it into preferred language settings, this check is no longer needed.

I added an NSAssert to catch if somehow an 'nb' made it through.

2. Switch to the correct code as soon a possible for links to articles in other languages.

In the 6.8 implementation, we create a MWKLanguageLink with the incorrect language code in the MWKLangaugeLinkFetcher and then use an `altISOCode` check in MWKTitleLanguageController to replace that instance with one with the correct code.

This PR makes the initial MWKLanguageLink that is created in MWKLangaugeLinkFetcher  correct. It also encapsulates the `altISOCode` mapping to `languageCode` in MWKLanguageLinkController.

This replaces the existing check in MWKTitleLanguageController.

### Test Steps
1. Set Norwegian as a language in the OS
2. Do a fresh install of the app
3. Accept the default languages suggested in onboarding
4. Norwegian should work as expected and assertion should not be hit

1. In a language that *isn't* Norwegian search for the article 'Dog'.
2. From the article 'Dog' tap the 'Change Language' button
3. Norwegian should appear as a language choice
4. Tapping Norwegian should display the article correctly

### DIFF
--- a/Wikipedia/Code/MWKLanguageLinkController.m
+++ b/Wikipedia/Code/MWKLanguageLinkController.m
@@ -96,6 +96,27 @@ static NSString *const WMFPreviousLanguagesKey = @"WMFPreviousSelectedLanguagesK
     return [NSLocale wmf_bestLanguageVariantCodeForLanguageCode:languageCode];
 }
 
++ (nullable NSString *)languageCodeForISOLanguageCode:(nullable NSString *)isoLanguageCode {
+    // Map altISOCodes to languageCodes once for fast lookups
+    static dispatch_once_t onceToken;
+    static NSDictionary<NSString *, NSString *> *languageCodesByAltISOCode;
+    dispatch_once(&onceToken, ^{
+        NSMutableDictionary *tempDict = [[NSMutableDictionary alloc] init];
+        for (MWKLanguageLink *language in self.allLanguages) {
+            if (language.altISOCode) {
+                tempDict[language.altISOCode] = language.languageCode;
+            }
+        }
+        languageCodesByAltISOCode = [tempDict copy];
+    });
+    if (!isoLanguageCode) {
+        return nil;
+    }
+    
+    NSString *languageCode = languageCodesByAltISOCode[isoLanguageCode];
+    return languageCode ? : isoLanguageCode; // If no alternative ISO code, use the original value
+}
+
 - (nullable MWKLanguageLink *)appLanguage {
     return [self.preferredLanguages firstObject];
 }
--- a/Wikipedia/Code/MWKLanguageLinkController.m
+++ b/Wikipedia/Code/MWKLanguageLinkController.m
@@ -16,6 +16,7 @@ static NSString *const WMFPreviousLanguagesKey = @"WMFPreviousSelectedLanguagesK
 @interface MWKLanguageLinkController ()
 
 @property (weak, nonatomic) NSManagedObjectContext *moc;
+@property (nullable, copy, nonatomic) NSArray<MWKLanguageLink *> *cachedPreferredLanguages;
 
 @end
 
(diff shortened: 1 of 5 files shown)


Wikipedia/Code/MWKLanguageLinkController.m
```diff
@@ -104,12 +125,10 @@ - (nullable MWKLanguageLink *)appLanguage {
...
         return [self.allLanguages wmf_match:^BOOL(MWKLanguageLink *langLink) {
-            //Note, sometimes the device iOS language codes will return a code that doesn't line up with the Wikipedia language codes we have set,
-            //so we are also checking for a match against the altISOCode (currently only set for "no.wikipedia.org")
-            //Fixes https://phabricator.wikimedia.org/T276645
-            return [langLink.contentLanguageCode isEqualToString:langString] ||
-                   (langLink.altISOCode &&
-                    [langLink.altISOCode isEqualToString:langString]);
+            // The Norwegian language code 'nb' used by the OS should be automatically converted to 'no' used by Wikipedia during onboarding.
+            // Issue described in  https://phabricator.wikimedia.org/T276645
+            NSAssert(![langString isEqualToString:@"nb"], @"Inaccurate language code found in preferred language codes. Language code 'nb' should be 'no'.");
```
tonisevener, line 130: I came across a situation where this assertion is hit - 

1. On a device with Norwegian Bokmal as a language, check out releases/6.7.4 and run.
2. Go through onboarding and add any language on the choose languages screen. (It will be in that weird empty state due to "nb" not being found on the old version).
3. Check out this branch and run.
4. Assertion is hit

It seems on `MWKDataStore`'s init, this code is run before migration happens.

tonisevener (commented):
Looks good - thanks for cleaning this up! Only thing I found is that I was able to trigger the assertion from upgrading to 6.7.4. I didn't dig further than that but I think it's worth looking into what it's doing when it's triggered to make sure it's not setting something improperly when initializing the data store.

Wikipedia/Code/MWKLanguageLinkController.m
dempseyatgithub: Looking at it, the preferred site URLs, which are based on the preferred languages are set on the explore feed *before* migration happens.

I need to investigate a bit more but there are a few things I want to check:

- In 6.7.4 Norwegian in the preferred languages list is just ignored, so won't appear in the explore feed.
- In 6.8 it is not ignored because of the altISOCode check
- In 6.8.1 upgrading from 6.7.4, it would be ignored when preferredSiteURLs is initially set. What happens to the explore feed settings with Norwegian going from 6.7.4 to 6.8 / 6.8.1?

2. It seems (on 6.7.4 at least) that if you go through onboarding without making a change to preferred languages that preferred languages are never set.

It seems like that means that every time the app is launched, we end up getting the same list we would get from onboarding. That doesn't seem like what we'd want to do though.

dempseyatgithub: I've updated my approach for `-preferredLanguages` in the most recent commit.

The new approach does the following:
- Caches the preferred languages in a `cachedPreferredLanguages` array that is invalidated when preferred languages change. This prevents repeated fetching and searching to generate the same array repeatedly.
- Uses the new `+languageCodeForISOLanguageCode:` method added in an earlier commit for article language links. This will do the altISOCode -> languageCode substitution even before the language variant migration updates the settings.

tonisevener (approved):
Thanks! Working great - I found one more bug with the Norwegian language, but it's existing on 6.8.0 (😬) so nothing from what you did here. I'll go ahead and approve and make a separate Phab task for this since it might be better to approach it separately.

1. Check out `releases/6.7.4` on a device with only Norwegian Bokmal as a language.
2. Launch app, go through onboarding, again it will be in that weird empty state. Add Norsk as a language.
3. Check out this branch
4. Now Norsk shows up twice in language settings. It should only be once.

Maybe this would be as simple as deduping when generating the cached preferred languages? I inspected and it looks like the same object in there twice.

PR #3928, opened 2021-04-27
Perform Xcode 12.5 and Swift 5.4 updates

staykids:
**Phabricator:** N/A

### Notes
A few bits to update the project for Xcode 12.5 and remove Swift 5.4 warnings.

* Performs "recommended project settings" update without making any actual target changes (see screenshot below)
* Ignores a string formatting compiler warning on Mantle.
* Replaces `class` with `AnyObject` for protocol inheritance. Seems like this was [in the works](https://github.com/apple/swift-evolution/blob/main/proposals/0156-subclass-existentials.md#class-and-anyobject) for a bit.

### Test Steps
1. Make sure the app still works 😉

### Screenshots
settings


tonisevener (approved):
Looks good!

Hope you don't mind but I snuck in an .xcversion file change. That should be the last thing needed for the build server.

tonisevener:
> Hope you don't mind but I snuck in an .xcversion file change. That should be the last thing needed for the build server.

Nevermind, seems CircleCI isn't ready for that yet and it breaks our tests. I'll make a followup PR when they have the 12.5 image.

PR #3929, opened 2021-04-28
Force animated GIF support in `NYTPhotoViewer`

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T249056

### Notes
Adds animated GIF support back to `NYTPhotoViewer`. It seems like this regression was likely introduced during my removal of Carthage and integration of those dependencies directly in the project. Carthage injected a preprocessor macro for `ANIMATED_GIF_SUPPORT` which was used to determine whether `NYTPhotoViewer` used a regular `UIImageView` or a `FLAnimatedImageView`.

This solution just directly removes those branching macro checks in the `NYTPhotoViewer` source to force favoring a `FLAnimatedImageView` to display animated images. I intentionally went this route over adding a  `ANIMATED_GIF_SUPPORT = 1` macro in the project – it's our intention to eventually remove these third party dependencies that we currently directly include in the main repo, so better not to leak these little details into the project file.

### Test Steps
1. Navigate to article with static image, tap image and ensure it renders in the fullscreen "gallery" photo view
2. Navigate to article with animated image, tap image to display fullscreen and ensure it animates (Phab ticket mentions these article examples with animated images:  `Continental Drift`, `Animation`)


benkaday3218 (approved):

dempseyatgithub (commented):
This looks fine but it looks like there are a couple of mentions of `ANIMATED_GIF_SUPPORT` remaining.

It probably makes sense to remove all mention of that macro since it is no longer present, even if that code path is not currently used by the app.

NYTScalingImageView.m, Line 126. Remove `#ifdef ANIMATED_GIF_SUPPORT`.

NYTPhoto.h, Line 28. Update comment so that it doesn't reference `ANIMATED_GIF_SUPPORT`.

staykids:
Good catch - updated!

dempseyatgithub (approved):
Looks good! Works as expected.

PR #3930, opened 2021-04-28
Remove unneeded search lookup

dempseyatgithub:
**Fixes Phabricator ticket:** 
https://phabricator.wikimedia.org/T281295

### Notes
The current search language is stored as a content language code.

In 6.8 and earlier, the code was used to look up its corresponding MWKLanguageLink. However the language link object was mainly used to get the content language code back out. Its one other use was to retrieve the site URL for that language.

In general once a subsystem is using language codes it shouldn't go back to using language link objects (that's a sign the subsystem should be using language links and not codes.)

In this case, the content language code has enough information to return the site URL.

This pull request has three commits:

1. Adds a category method to NSURL to create a site URL from a content language code. The resulting URL has the appropriate value set for `wmf_langaugeVariantCode` if appropriate. Also adds a test to confirm the behavior.

2. Replace uses of the MWKLanguageLink instance with the content language code. Also adds a public `selectedSiteURL` property while making an unused property private for better encapsulation.

3. Remove the now unused method `-langugageForContentLanguageCode:`

I updated this so that `selectedSiteURL` filters the search language MWKLanguageLink instances that are already present to find the matching content language code and return the site URL.

I removed the newly introduced NSURL category method since it is no longer needed.

### Test Steps
There should be no change in app behavior with this PR.

Test variants:
1. Add two language variants to preferred languages — Serbian, Latin and Serbian, Cyrillic.
2. Search for something with one variant selected.
3. Tap on a result and make sure correct variant is displayed in article view
4. Return to Search
5. Select the other variant and search
6. Tap on a result and make sure newly selected variant is displayed in article view

Test selection:
1. Ensure multiple preferred languages are set in the app
2. Go to Search
3. Change search language
4. Selection should update correctly
5. Leave search and navigate elsewhere in the app
6. Return to search
7. Selected language should still be selected
8. Quit app
9. Relaunch app
10. Go to Search
11. Previously selected language should be selected


Wikipedia/Code/NSURL+WMFLinkParsing.m
```diff
@@ -23,6 +23,23 @@ + (NSURL *)wmf_URLWithDefaultSiteAndLanguageCode:(nullable NSString *)languageCo
     return [self wmf_URLWithDomain:WMFConfiguration.current.defaultSiteDomain languageCode:languageCode];
 }
 
++ (NSURL *)wmf_URLWithDefaultSiteAndContentLanguageCode:(nullable NSString *)contentLangaugeCode {
+    if (!contentLangaugeCode) return nil;
+    NSString *languageCode = nil;
+    NSString *languageVariantCode = nil;
+    if ([contentLangaugeCode containsString:@"-"]) {
```
tonisevener, line 30: I don't think we should rely on the hyphen right away, there are [some language codes](https://github.com/wikimedia/wikipedia-ios/blob/main/Wikipedia/Code/wikipedia-languages.json#L1614) with a hyphen where this will break (i.e. I send in `zh-classical` language code, and expect to get a `zh-classical.wikipedia.org` URL.

Wikipedia/Code/NSURL+WMFLinkParsing.m
```diff
@@ -23,6 +23,23 @@ + (NSURL *)wmf_URLWithDefaultSiteAndLanguageCode:(nullable NSString *)languageCo
     return [self wmf_URLWithDomain:WMFConfiguration.current.defaultSiteDomain languageCode:languageCode];
 }
 
++ (NSURL *)wmf_URLWithDefaultSiteAndContentLanguageCode:(nullable NSString *)contentLangaugeCode {
```
tonisevener, line 26: Tiny spelling thing - contentLangaugeCode > contentLanguageCode.

Wikipedia/Code/NSURL+WMFLinkParsing.m
dempseyatgithub: Thanks for catching that!

Wikipedia/Code/NSURL+WMFLinkParsing.m
dempseyatgithub: This has been removed.

Wikipedia/Code/NSURL+WMFLinkParsing.m
dempseyatgithub: I've changed the implementation so we don't need this method at all.

benkaday3218 (approved):

tonisevener (approved):
Works great!

PR #3933, opened 2021-05-03
Add empty state for no other article languages

tonisevener:
**Fixes Phabricator ticket:** https://phabricator.wikimedia.org/T277066

### Notes
* This is hooking into our existing empty state helper methods to add an empty view when there are no other article languages to choose from.

### Test Steps
1. Go to article for "Myriam Gurba" on EN Wikipedia.
2. Tap the languages button
3. Confirm you see empty state as described in the Phab ticket (note I didn't make any changes to the navigation bar or search to match mocks exactly, see [Phab comment](https://phabricator.wikimedia.org/T277066#7039973) for approval). Confirm it works across device and type sizes, orientations, and themes.
4. Regression - Go to an article with languages (any common one like Cat). Confirm languages show and work as before.
5. Regression - Go to languages list page from Settings and Search, confirm they work as before.

Screen Shot 2021-05-03 at 5 51 48 PM
Screen Shot 2021-05-03 at 5 52 00 PM


Wikipedia/Code/WMFLanguagesViewController.m
```diff
@@ -678,6 +683,13 @@ - (void)downloadArticlelanguages {
     [self.titleLanguageController
         fetchLanguagesWithSuccess:^{
             @strongify(self)
+                if (self.titleLanguageController.allLanguages.count == 0) {
+                    [self wmf_showEmptyViewOfType:WMFEmptyViewTypeNoOtherArticleLanguages theme:self.theme frame:self.view.bounds];
+                    self.hideLanguageFilter = YES;
```
staykids, line 688: I could be missing a path it's mutated elsewhere, but since this is already set in the lifecycle method, we could just remove this line.

staykids (commented):
Working great so far! A minor element commented and a suggestion below.

A thing that’s slightly distracting is that the true empty state of the view, before it gets populated with a `WMFEmptyView`, is… well… _empty_, and doesn’t communicate that some activity (awaiting an API response) is occurring. I think we can do a little to help smooth the experience a bit.

I put these suggestions [in this branch](https://github.com/wikimedia/wikipedia-ios/compare/T277066...T277066-suggestions) (forgive my quick hacky Objective-C):
- Adds an activity indicator that’s visible until the data required for updating the display is fetched. I think this helps communicate to the user that _some_ action is happening while they wait for the translated languages to be fetched.
- Adds a short and subtle fade in of the `WMFEmptyView` when the fetch completes with no results. I think this helps the transition and overall experience feel less jarring.

(As always, if you’d rather we roll this into a separate PR totally fine by me.)

tonisevener:
@staykids thanks for the suggestions! Agreed on the loading state - briefly I tried to add one with the blue progress bar from `FakeProgressController.swift` that we have under navigation areas, but it was less plug-n-play for `WMFLanguageViewController` due to being in Objective-C and not a subclass of `ViewController.swift`, so I bailed since it wasn't exactly an ask in the ticket. But yours looks good and is definitely an improvement. I just made one slight adjustment in making `activityIndicator` nonatomic since it should only ever be accessed from the main thread anyways.

staykids (approved):

PR #3934, opened 2021-05-04
Create FUNDING.yml

mcleinman:

### Notes
* Add a support WMF link. From conversation w/ advancement.
* This will cause a link to show up on the right side of our main repo webpage.

### Test Steps
1. N/A


staykids (changes requested):
Nice - looking at other `FUNDING.yml`s, it looks like we can strip everything except the `custom` one we actually want to use.

mcleinman:
Very good point, thanks. 

That said, I went org-wide (so we didn't need to replicate this in Android): https://github.com/wikimedia/.github/tree/master.

Closing this out.

PR #3935, opened 2021-05-04
Allow tap focusing on User Talk page UI elements when VoiceOver is enabled

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T280838

### Notes
Currently, a user with VoiceOver enabled can use the left or right swipe gestures to navigate through the next/previous User Talk page UI elements, but they can’t arbitrary tap or tap and drag onto header view elements. It appears this was because our custom hit testing superseded VoiceOver’s hit testing. This PR ignores the custom hit testing if VoiceOver is enabled (as is done in other places in the app) – this allows richer accessibility support when VoiceOver is enabled and preserves the previous behavior when it’s disabled.

### Test Steps
1. Enable VoiceOver in iOS’s Settings
2. Access either your own User Talk page (Settings → Account → Your talk page) or another user’s User Talk page (one way is select an article → View edit history at bottom → Tap a user name → Tap user name again → tap Talk in page)
3. Confirm the left or right swipe gestures (next/previous) navigate you through all the presented UI
4. Confirm arbitrarily tapping header view elements (User Talk header, user’s name, subheading) works and gets announced via VoiceOver
5. Confirm tapping on one element and dragging to a different UI element in this view works and gets announced via VoiceOver


dempseyatgithub (approved):
This works as described with VoiceOver and the changes match the pattern set by other similar cases in the code. Looks good!

PR #3936, opened 2021-05-05
Update right-to-left methods to handle language variants

dempseyatgithub:
**Fixes Phabricator ticket:** 
https://phabricator.wikimedia.org/T280720

### Notes
This PR does two main things:

1. Updates the call sites of the various methods that take a content language code and return an indication of RTL. Previously the language code was passed in. Call sites are updated to pass in language content code.

Note image descriptions do not support language variants and still require the use of language code with these methods. Those cases are commented for clarity.

Also, a `contentLanguageCode` property is added to WMFFeedOnThisDayEvent to follow the pattern of the existing `languageCode` property

2. The RTL language variants are added to the set of RTL content language codes. The private property `rtlLanguages` is renamed `rtlContentLanguageCodes`.

Note that I also audited all methods that take a contentLanguageCode argument to ensure that a content language code is being passed in, or annotated with a comment if there is some reason using the language code is appropriate in that case.

### Test Steps
1. Add Arabic variants of Kurdish and Kazakh.
2. In Explore feed, Top Read card should appear RTL
3. Tap into an article
4. Due to a server-side issue, the contents *will not* appear RTL
5. Tap on the TOC button
6. The TOC should appear RTL

(Note, only the title is returned in Arabic in Kazakh, this appears to be a server-side issue.)


tonisevener (approved):
Looks great!

PR #3937, opened 2021-05-05
[WIP] Add logging of MobileWikiAppiOSNavigationEvents via legacy system

tonisevener:
**Fixes Phabricator ticket:** https://phabricator.wikimedia.org/T280739

### Notes
* [WIP, will add after testing]

### Test Steps
* [WIP, will add after testing]


staykids (approved):
Nice! I’m glad to see the solution wasn’t too invasive. Thanks for the test steps - here’s what I tried:
- Confirmed the enum matches the specified schema revision of  `MobileWikiAppiOSNavigationEvents` on Meta
- Triggered each of the events (pretty sure I hit them all 🤪)
- Proxied traffic to confirm the encoded URLs mapped to the expected event action
- Switched primary language to confirm that key updated as expected

PR #3938, opened 2021-05-06
Retain edit summary when moving back and forth through edit preview

mcleinman:
**Fixes Phabricator ticket:** https://phabricator.wikimedia.org/T255490

### Notes
* From looking at the git history of this file, it was converted from ObjC to Swift a couple years ago, and hasn't seen much work since. At the time, Monte noted that it could use some love as it's a little sloppy. I did a small refactor to make it make a bit more sense. Also removed code from `viewWillAppear`, which fixes the bug in the next bullet point.
* * There is a bug currenty in there - if you enter some summary text on the final edit screen, then tap "Learn more", then  go back - the text disappears.
* The Edit flow could probably use some larger love. Delegates on delegates, and `SectionEditorViewController` (the first one a user sees) controls some of the remainder of this flow. (Also a bonus `EditSummaryVC` which is the textfield within the `EditSaveViewController`.)  I don't love this architecture, but I took advantage of it and added one more function to the delegate.
* I don't love how I'm persisting the state here. Open to other ideas.
* There is an unused "watch this page" toggle that has never been used. This code should allow us to persist that toggle as well, if it ever does get used.

### Test Steps
1. Edit some text. On the last page, add in an edit summary, maybe toggle the toggle - then go back a screen or two, then go back forward - your text/toggle should persist.

### DIFF
--- a/Wikipedia/Code/EditSaveViewController.swift
+++ b/Wikipedia/Code/EditSaveViewController.swift
@@ -8,6 +8,7 @@ struct SectionEditorChanges {
 
 protocol EditSaveViewControllerDelegate: NSObjectProtocol {
     func editSaveViewControllerDidSave(_ editSaveViewController: EditSaveViewController, result: Result<SectionEditorChanges, Error>)
+    func editSaveViewControllerWillCancel(_ saveData: EditSaveViewController.SaveData)
 }
 
 private enum NavigationMode : Int {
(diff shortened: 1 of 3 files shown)


dempseyatgithub:
I tried the steps to test and the summary information did not persist after going back and forth.

I also have never done an edit in the app before, so here's a screen capture of what I did to make sure I was trying/testing the correct thing.

https://user-images.githubusercontent.com/470139/117373027-61cf7e80-ae7f-11eb-8956-cc84fb63d023.mov

PR #3939, opened 2021-05-06
Change the no-other-article-languages

amire80:
Articles are not necessarily translated. They are often written
from scratch in different languages.

**Fixes Phabricator ticket:** 

### Notes
* 

### Test Steps
1.


mcleinman (approved):
Great catch, thank you! 

(iOS team - @carolynlimadeo OK'd this change.)

PR #3944, opened 2021-05-07
Use user's default mail client when reporting a bug

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T269042

### Notes
* Removes unused `MFMailComposeViewController` references in `WMFAlertManager`
* Uses `mailto` approach to "Report a bug" in order to respect user's default mail app preference. This moves from an in-app modal to report a bug to pushing the user to that preferred app.

### Test Steps
Generally, the steps from the ticket apply.

1. Open app → Settings → Help and feedback → tap "Report a bug"
2. Confirm, with the exception of including the `Log.txt` attachment, the constructed email's address, subject, and body content is identical to the previous in-app mail modal but opens in your email client app
3. Install an email client alternative to Mail.app that has iOS 14's default mail client entitlements (Gmail is an example)
4. Open iOS Settings → scroll to Gmail (or whatever email client you install) → tap Default Mail App → select Gmail
5. Repeat step 1 and confirm step 2 opens in your newly selected Default Mail App

So far I've noticed: a) Gmail doesn't respect the included newlines in the email body and 2) when I initially started with only Mail then installed a new client, it took a toggle or two of "Default Mail App" to have `mailto` actually use the newly selected client – I haven't been able to repeat this delayed behavior again, and I don't believe this is due to any error on our part.


dempseyatgithub (approved):
It looks like there are conflicts that need to be resolved.

This behaved as described. I tested with Mail and Gmail. The code looks good.

I tried a few possible workarounds for the Gmail newline issue, but it doesn't seem to respect newlines at all.

staykids:
I don’t want to unintentionally/unexpectedly crush any string changes that I may have missed in the interim, so I just reverted the string cleanup commit (which removes an unused code block). I’ll remake it separately after this is merged.

This should be ready to merge.

PR #3951, opened 2021-05-13

translatewiki:
Translation updates


tonisevener (approved):
Looks good now. I updated the string that caused the tests to fail in translatewiki and locally in the project and re-ran the import script.

PR #3952, opened 2021-05-14
Remove grunt from project

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T266837

### Notes
This is a proposal that we remove `grunt`, which in my opinion did not buy us much in its current state and is not worth the complexity. Grunt would:
1. Runs `es-lint` on the `js` files in `www`.
2. Combine some `.less` css styles in `www` into a single `styleoverrides.css` file, and save that into `assets`.
3. Copy various other files from the `www` folder into the `assets` folder.

This PR removes the `www` entirely so that now we update files within the `assets` folder directly. From what I gather, we lose js linting on these files and some minimizing of `styleoverrides.css`, but everything else was pretty much a direct copy. The 3 `.less` files that were combined into `styleoverrides.css` are very small themselves, so working in `styleoverrides.css` directly would not be too long and unwieldy.

I also cleaned out unused files from `assets` and updated the documentation in `web_dev`. There are some subdirectories in `assets` that we shouldn't update directly, and I documented the process for those. These processes mostly already existed and were just undocumented. There is one slight change in process when updating from the codemirror submodule:

Before:
Download submodule, run `grunt` from `www` which copies the submodule `resources` folder into `assets/codemirror/resources`

After:
Download submodule, run  `copy_code_mirror_resources` script which copies the submodule `resources` folder into `assets/codemirror/resources`

### Test Steps
1. Test Settings > About, Article, Editing and Edit Preview views and confirm there are no regressions.
2. Instantiate submodule and run `copy_code_mirror_resources` script from the `scripts` folder. Confirm resources folder from submodule gets copied into `assets/codemirror/resources`.


staykids (approved):
Awesome – love getting rid of these smaller dependencies. I made a minor commit fixing some typos in `web_dev.md`. Here’s generally how I tested based on your steps:

1. Deleted CodeMirror’s `resources`, confirmed editing didn’t work
2. Used new script, confirmed it pulled resources back into proper location, confirmed editing worked
3. Lightly regression tested (against the shipping build) the About page, article view, CodeMirror editor view, and edit preview, in both light and dark themes, to confirm they still appeared to function as normal

I’m happy to test any other specifics if they come to mind, but otherwise looks great.

PR #3956, opened 2021-05-27
Project settings and configuration cleanup

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T283880

### Notes
This is a cleanup effort to simplify our environments and project build settings. Notable commits are explained below:

**Commit 1** - removing old unused macros and custom flags, though note the MTL flags were added back in in commit 7 due to some Metal exceptions I was seeing when testing.

**Commit 2** - This commit tries to solve two problems:
1. We have several macros and a few build configurations that only exist to target one particular non-production environment for one group of endpoints. Rather than continuing to track these separately I have reduced them down to two macros: `WMF_LOCAL` and `WMF_STAGING` with the ability to fine-tune which endpoints apply within the `Configuration.current` property. I think this grouping will be easier to deal with moving forward.
2. Before the `Local Page Content Service & Announcements` scheme would have both local pcs and local announcements toggled on, and to only test one (such as announcements, which is fairly often with our experiments) you would have to dig into the project build settings and find the right macro and build config to tweak. Now they are still both on by default for `WMF_LOCAL`, but you can easily switch one or the other off by removing them from the local options in `Configuration.current`.

**Commit 3**  - removing Event Logging Dev Debug and Beta Cluster schemes,  Event Logging Dev Debug functionality can now be tested with the Staging scheme.

**Commit 4** - here I became a bit refactor-happy but I find this easier to understand (ymmv). The Configuration init factories were basically closures to run whenever a url needed to be created in the app. These closures really only preserved what host and base path components will need to be sent to the `APIURLComponentsBuilder`, which varied depending on what environment you are pointing at (staging might need to start with `https://mobileapps.wmflabs.org/en.wikipedia.org/v1/` vs production `https://en.wikipedia.org/api/rest_v1`, for example). So instead of keeping track of this as a closure, I am instantiating `Configuration` with `BuilderType` enums that live within `APIURLComponentsBuilder`. These enums have the ability to know what their host and base components will need to be whenever a url needs to be generated.

**Commit 5 - 8** - smaller additional cleanup commits.

**Commit 9** - After testing on `main`, we were pretty inconsistent with how we were pointing to the Beta Cluster, and the tests that ran against it were [scrapped after the mobile-html refactor](https://github.com/wikimedia/wikipedia-ios/pull/3651). I checked with Android and they aren't using it as a staging environment at all, so here I decided to remove it.

**Commit 10** - Updated documentation with these changes, plus old changes in Event Logging and changes to our PR process document (I don't think we need as many PR labels, and I can remove most of them once this is agreed upon and merged). I also removed mentions of UI tests since we're not actively maintaining them.

### Test Steps
1. The main thing to do is regression test production. Run the Wikipedia scheme and proxy, and confirm the calls we make all look like they're going to production. Turn on event logging and confirm those calls look right. Also test Edit Preview.
2. I think any Staging and Local scheme issues introduced here can be fixed as follow-ons as we find them, but here's the main > this PR branch mapping if you want to test these:

- `main` - `Local Page Content Service & Announcements` scheme calls should match this PR - `Local Page Content Service & Announcements` scheme
- `main` - `Staging` scheme calls should match this PR - `Staging` scheme
- `main` - `Event Logging Dev Debug` scheme calls should match this PR - `Staging` scheme - `Configuration.staging(options: [.deploymentLabsForEventLogging])` setup in the [current property](https://github.com/wikimedia/wikipedia-ios/blob/b6fd4d90a18224c0a1fd96b8eb881366913d679b/WMF%20Framework/Configuration.swift#L44).


dempseyatgithub (approved):
As far as I can tell, this looks good to me and seems to be hitting production.

PR #3959, opened 2021-05-28
Fastlane: PR creation for version bumps

mcleinman:
**Phabricator:** N/A

### Notes
* This automates version changes, but requiring human approval.
* After this is approved, I need to change the branch that Jenkins points to. (https://wmf2249:8043/job/Increase%20version/configure). Additionally, I need to merge two Jenkins jobs - this one and `Tag last beta as release`, so they happen as one.

### Test Steps
1. Log into Jenkins.
2. Run the `Increase version` job. 
3. Ensure it opens a PR, and it looks appropriate. Close the PR, and delete the branch (via GH).


fastlane/Fastfile
```diff
@@ -180,6 +180,37 @@ platform :ios do
     )
   end
 
+  desc "Change version number and create PR with changes"
+  lane :change_version do |options|
+
+    size = options[:size]
+    case size
```
tonisevener, line 187: I haven't figured out where but it seems like this size is getting dropped somewhere. All the PRs that open are patch bumps, regardless of what I select in the Jenkins dropdown.

mcleinman: This should now be working, after some adjustments.

tonisevener (approved):
The version bump commits look good now, thanks! I'm guessing the git commit history in the PR will clean up once we have Jenkins configured to `main` for this job.

PR #3969, opened 2021-06-04
Fix gallery crash

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T284357

### Notes
We have a couple of threading crashes related to the `WMFPOTDImageGalleryViewController` on 6.8.1. It looks like sometimes upon failure of the image info endpoint, the failure block is not run on the main thread. Dispatching to main fixes it. This also includes whitespace changes from ClangFormat.

### Test Steps
1. Map locally the image info endpoint that is called after tapping the POTD cell in the Explore feed. Map with a response structured like this:

```
{
    "error": {
        "error": {
		"code": "123"
	}
    }
}
```
2. Ensure it doesn't crash, `main` crashes here.


dempseyatgithub (approved):
This looks good to me.

PR #3991, opened 2021-06-30
Add MetricsClientBridge for Objective-C compatibility

mdholloway:
Splits the EventPlatformClient's Objective-C compatilibity layer into a bridging class which will remain in the app repository after the client library is consolidated with the other platform libraries as part of the work falling under [T280182](https://phabricator.wikimedia.org/T280182).  This is needed because the standalone Swift client library will need to run on Linux, where the `@objc` annotation providing interoperability is not supported.

The bridging class will likely also be used to expose main app functionality to the client library once the library itself is extracted.

Note: Metrics Client is the new name for the Event Platform Client project.


WMF Framework/Event Platform/EventPlatformClient.swift
```diff
@@ -751,3 +726,12 @@ public protocol EventInterface: Codable {
      */
     static var schema: EventPlatformClient.Schema { get }
 }
+
+// MARK: Objective-C
+
+public protocol MetricsClientObjectiveCCompat: BackgroundFetcher, PeriodicWorker {
```
tonisevener, line 732: It doesn't look like this protocol is used anywhere, can you remove it? Your extensions in `MetricsClientBridge.swift` should be enough to conform to `BackgroundFetcher` and `PeriodicWorker`.

tonisevener (commented):
@mdholloway This looks good! Just one small ask about `MetricsClientObjectiveCCompat`.

WMF Framework/Event Platform/EventPlatformClient.swift
mdholloway: Done!

tonisevener (approved):

PR #3992, opened 2021-06-30
Convert AppDelegate and WMFSettingsViewController to Swift

mdholloway:
I tried my hand at converting a couple of classes to Swift. The motivation for this was to eliminate the need for Objective-C interoperability in the EventPlatformClient class, but wholesale conversion of all callers to Swift proved to be too much to take on right now, so I ended up taking a different approach to resolving our immediate need (#3991).  

Disclaimer: I kept the logic pretty close to how it was in the Obj-C classes, the automated tests pass, and the app seems to be working well, but I'm not an iOS or Swift expert by any means, so please review with care!


Wikipedia.xcodeproj/project.pbxproj
```diff
@@ -651,6 +650,10 @@
 		67F73E752267B9070079DEEF /* TalkPageReplyListViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 67F73E742267B9070079DEEF /* TalkPageReplyListViewController.swift */; };
 		67F73E792267B9510079DEEF /* TalkPageTopicNewViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 67F73E782267B9500079DEEF /* TalkPageTopicNewViewController.swift */; };
 		67F9AE4923AD2F38003D4F5E /* Array+SafeIndex.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7A3EE1532267DC3800709CF6 /* Array+SafeIndex.swift */; };
+		7016040E2684DE32009C84EC /* AppDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7016040D2684DE32009C84EC /* AppDelegate.swift */; };
```
tonisevener, line 653: Can you also add the new files (`main.swift`, `AppDelegate.swift`, `WMFSettingsViewController.swift`) to our other 3 targets? These are just some alternative builds, they won't compile without these checked:

Screen Shot 2021-09-17 at 12 06 23 PM

tonisevener (changes requested):
👏 

@mdholloway Thanks so much for taking the time to do this, it looks like a lot of work! Most of my feedback is very small style stuff. Some comments may not be as explanatory as they could be, so please let me know if you have any questions.

tonisevener:
Going to go ahead and close this one due to other team priorities.

PR #3998, opened 2021-07-12
Add additional error handling in localization updates.

mcleinman:
**Phabricator:** N/A

### Notes
* Last few localization failures have been due to bad formatting of plurals. (Notably, they've been in RTL languages, which might be more confusing because of how the LTR `PLURAL:$1|` is shown within these strings.) This gives a better message in this situation, rather than just a crash w/ no explanation.
* Tested this against the currently failing TWN branch, and it worked as expected.
* While it would be nice to have the string name and language in the message, that info isn't passed into the failing function. It doesn't seem worth making this function throw and handling the error upstream, so we'll just deal with what we have here - it's easy to figure out the failing translation when we run the script locally. Additionally, while the script crashes on the first failure, this specific failure type seems to come 2+ at a time (if a person has been adding some translations, they've usually been doing a few) - and so it's nice to force the engineer to actually run this locally a few times, to ensure we've caught all instances of misformatted translation strings.

### Test Steps
1. Ensure automated tests pass


tonisevener (approved):
Looks good!

PR #3999, opened 2021-07-14
Allow interactive pop gesture in Settings

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T217937

### Notes
Currently, the interactive pop gesture – which allows you to pan the left edge of the screen to pop a view controller in a navigation controller – doesn’t work in the Settings navigation stack. This PR adds that functionality. It seems likely this is currently nonfunctional as a side effect of some of our custom navigation bar/gesture intercepting stuff higher up. Instead of applying a solution to some wider touch base class, I felt better about just applying this solution constrained to the Settings navigation controller stack. I’m not entirely sure what, if any, potential side effects this fix may have in the Settings stack, but so far it seems to function as expected.

### Test Steps
1. Launch app
2. Tap the Settings button in the Explore tab
3. Navigate to a subpage that is pushed on the navigation stack (like Search, Explore feed, Notifications, etc.)
4. Confirm the left edge interactive pop pan gesture works to interactively pop the current view controller.
5. Force quit the app while in Settings, and reopen the app. Confirm the interactive pop pan gesture still works in this state restored Settings stack.
6. Disable the Explore feed. Confirm the interactive pop pan gesture still works.

Note that it's expected that the subpages that present themselves modally (like Log in and My languages) don’t support this interactive dismiss behavior.


tonisevener (approved):
Working great! Thanks for keeping it light touch.

PR #4004, opened 2021-07-27
Manually check for Farsi when searching for languages

mcleinman:
**Phabricator:** https://phabricator.wikimedia.org/T107530

### Notes
* Quick hack to check for Farsi in search box, per discussion.
* Also refactored out some duplicated code.

### Test Steps
1. Run app, go to adding languages.
2. Search for "Farsi". After every letter typed, "Persian" should be an option as well as the expected results.
3. "Persian" should also be available when searching for "Persian", of course.


tonisevener (approved):
Working well for me! For whatever reason GitHub is struggling to update with `main` for me. I'll try merging tomorrow.

PR #4005, opened 2021-07-28
Add snapshot preview content for Featured Article widget

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T259842#7230812

### Notes

Adds the EN "Moon" featured article as the Widget Gallery preview for the Featured article widget when other data is unavailable. We're not really worried about full localization for this content because the states in which this would appear are quite transient.

With this addition, the expected behavior of what you should see in the Widget Gallery (where you can add a widget to your home screen) is something like this:
1. If you open the widget gallery and scroll to Featured article in Wikipedia, and no cached widget data is available, you should see the Moon article.
2. If you're at that same place in the gallery, but the app has already cached a featured article, you should see that cached featured article.

Unfortunately, the Wikifeeds content featured endpoint doesn't support content before 2016 (and this article was featured in 2007), so I had to construct the sample data (`Featured Article Widget Preview Content.json`) manually. Please forgive that it's practically unreadable – you can prettify the JSON to see it's the Moon article with encoded image data for article's image.

### Test Steps
Unfortunately, I don't have great test steps here because I can't determine Apple's policies for removing cached widget gallery data.

The sure way to do it, which I _do not_ recommend is:
1. Install fresh build to device that hasn't previously had a build of Wikipedia deployed to it, while in Airplane Mode
2. Confirm you see Moon article in Widget Gallery
3. Confirm after adding this widget to the home screen that an error state is rendered that states "no featured article content is available for this date".

But I think a more realistic and sufficient test path would be:
1. Add this code block [to this line](https://github.com/wikimedia/wikipedia-ios/blob/4675187412b990421e894ca266ae4ee7bbbf0399/WMF%20Framework/Widget/WidgetController.swift#L236) to display the Moon article content on widget on the home screen, and confirm it renders correctly in all widget sizes on home screen and links through to the Moon article:
```swift
performCompletion(result: .success(WidgetFeaturedContent.previewContent()!))
return
```
2. Replace [this line](https://github.com/wikimedia/wikipedia-ios/blob/4675187412b990421e894ca266ae4ee7bbbf0399/WMF%20Framework/Widget/WidgetController.swift#L238
) with the one below, and confirm all widget sizes render correctly in the Wikipedia Widget Gallery itself:
```swift
let previewSnapshot = WidgetFeaturedContent.previewContent()!
```

### Screenshots
featured_article_preview


tonisevener (approved):
Working well for me with your test steps - do you want me to merge it now or wait until the base branch is merged into `main`?

staykids:
👍🏽 thanks for looking at it. If it's cool with you, I'd rather we wait to merge it into `main` after its parent branch gets merged.

tonisevener:
@staykids can you resolve the conflicts on this one?

staykids:
> @staykids can you resolve the conflicts on this one?

👍🏽 done!

tonisevener:
@staykids thanks, I'm actually unable to trigger this this moon state at all now - it seems like something with the merge from `main` changed things. I haven't looked into why though. When I go back to the branch as it was in [467518](https://github.com/wikimedia/wikipedia-ios/pull/4005/commits/4675187412b990421e894ca266ae4ee7bbbf0399) I can see it pretty easily.

staykids:
> @staykids thanks, I'm actually unable to trigger this this moon state at all now

@tonisevener I think I know what might be up – lemme reach a stopping point with writing up this notifications UI stuff and I'll swing back to update this ticket this afternoon.

staykids:
Updated and added a unit test. This occured because of a JSON decoding problem with the sample content (going from `normalizedtitle` to `displaytitle`).

I remember when initially creating this sample JSON I thought “It’d be wise to create a unit test for this in case we change the data model“ but I didn’t because I figured we wouldn’t change the data model 😆. Glad this bit us sooner rather than later. Should be ready for another look!

tonisevener (approved):
Thanks! Looks good 👍

PR #4011, opened 2021-08-07
Add searching via search tab

yodagamaheshan:
**Phabricator: https://phabricator.wikimedia.org/T257708** 

### Notes
* Since TabBarItem is not inherited from UIView we cannot add gesture directly to it just calling by `addGestureRecognizer()` so that I have used `tabBarItem.value(forKey: "view")` to filter and find the view.

### Test Steps
1. Double-tapping on search tab directly when user switching from another tab gives the user the ability to search directly 
2. Enable searching , after tap on search tab bar item when user already in the search tab.

### Screenshots/Videos

iPhone

https://user-images.githubusercontent.com/50470084/128607873-68d16706-1dd3-4dbd-aebc-b2183b4adb21.mp4

iPad

https://user-images.githubusercontent.com/50470084/128607878-e141c6b0-3d99-4afd-b9fd-de1272f2b8d6.mp4

### DIFF
--- a/Wikipedia/Code/SearchViewController.swift
+++ b/Wikipedia/Code/SearchViewController.swift
@@ -17,6 +17,7 @@ class SearchViewController: ArticleCollectionViewController, UISearchBarDelegate
         view.bringSubviewToFront(resultsViewController.view)
         resultsViewController.view.isHidden = true
         useNavigationBarVisibleHeightForScrollViewInsets = true
+        handleBecomingFirstResponderAfterTappingOnSearchTab()
     }
 
     override func viewWillAppear(_ animated: Bool) {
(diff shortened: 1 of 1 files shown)


tonisevener (changes requested):
@yodagamaheshan Thanks for this PR! This code looks good to me - there's just one experience discrepancy I noticed when testing this. When the app settings are configured to land on the Search screen after launching (App settings > Search > Open app on Search tab > toggle ON, terminate and retest), it seems like this second tab tap no longer focuses on the search bar. Can you fix this when you get a chance?

yodagamaheshan:
> @yodagamaheshan Thanks for this PR! This code looks good to me - there's just one experience discrepancy I noticed when testing this. When the app settings are configured to land on the Search screen after launching (App settings > Search > Open app on Search tab > toggle ON, terminate and retest), it seems like this second tab tap no longer focuses on the search bar. Can you fix this when you get a chance?

sure I will take a look and fix it.

tonisevener:
> SearchViewController becomes the first responder automatically ( without taping the search tab - even in the current release ) when the app settings are configured to land on the Search screen after launching.

@yodagamaheshan sorry for the delay on this! Yes, there is some code to become the first responder in `SearchViewController`'s appearance lifecycle methods, but those are conditional to `shouldBecomeFirstResponder` being true. I did a little bit of digging and that value doesn't seem to be true for this context as tab bar controller's child, which lines up with how it's working when I run it. It sounds like it's acting differently for you - can you let me know what device or simulator you're testing on?

I do think this code looks safe and is a great addition for the standard settings setup. I'll confirm with others on the team if this difference is fine, and if so I'll merge later this afternoon.

tonisevener (changes requested):
Hi @yodagamaheshan! After digging around a bit, I discovered another way to do this that I think feels a bit more future proof, and it fixes that edge case failure when the user lands on Search after launch. 

For the Explore feed, we are doing something similar where we scroll to the top if the user taps it's tab bar item a second time. The placement of that logic is [here](https://github.com/wikimedia/wikipedia-ios/blob/b30caaed7cf75139e033702bc7d0981bd15e3855/Wikipedia/Code/WMFAppViewController.m#L1568), in `UITabBarControllerDelegate`'s `shouldSelectViewController` method. Can you lean on this method to trigger the search bar focus in the SearchViewController? You could add something like this here:

```
case WMFAppTabTypeSearch: {
                SearchViewController *searchVC = (SearchViewController *)[self searchViewController];
                [searchVC handleTapOnTabBarItem];
            } break;
```

Then you can remove `handleBecomingFirstResponderAfterTappingOnSearchTab`, and won't need the additional tap gesture recognizer or the private-ish `valueForKey` API. Let me know if you have any concerns, thanks!

PR #4013, opened 2021-08-10
Push notifications subscription flow

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T287305

### Notes
This PR is meant to lay the groundwork for getting push notifications to appear in the app. It involves subscribing the device token from the notifications framework to our echo backend, as well as asking users for notifications permissions.

Also note the changes to `NotificationSettingsViewController` are meant to be rather temporary and are just a happy path to get to an enabled permissions and subscription state. Fine-tuning to it will be made when https://phabricator.wikimedia.org/T288688 is tackled.

It's probably easiest to review this commit-by-commit. The commit messages are hopefully clear enough.

After review, please wait for #4010 to be reviewed and merged. Then update this with feature/notifications and merge.

### Test Steps
1. First update [this line](https://github.com/wikimedia/wikipedia-ios/blob/T288297/WMF%20Framework/Configuration.swift#L47) with `return Configuration.staging(options: [.betaCluster])`. We're going to temporarily force the Experimental scheme to point to the beta cluster (the sandbox flow so far only pushes to the Experimental bundle ID).
2. Create two accounts on the [beta cluster](https://en.wikipedia.beta.wmflabs.org/wiki/Main_Page). Log in on desktop with the first one. Go into preferences and make sure all web and apps column checkmarks are checked. Make an edit to any article.
3. Fresh install this app on a device running the Experimental scheme.
4. Go to Settings > Notifications.
5. ~~Proxy here to watch calls, you will reference device token posted here in later step.~~ Log in to your first account if needed here, then toggle ON and allow OS push permissions. ~~Make a note of `providertoken` value and `topic` value in call made to `action=echopushsubscriptions&command=create`.~~
6. Background the app. Now on desktop, log in to your 2nd account you made in step 2. Go to the article that your first account edited, tap into history, and thank the first account for that edit.
7. ~~Temporary step because sandbox flow is [broken again](https://phabricator.wikimedia.org/T287897#7302176): Go to push notification utility and send a visible using the device token and topic noted in step 5~~. Actually this step shouldn't be necessary anymore. The sandbox flow should be back in place now without the need to trigger via the push utility.
8. Your device should display a "checkEchoV1" push notification.

### DIFF
--- a/Wikipedia/Code/WMFNotificationsController.m
+++ b/Wikipedia/Code/WMFNotificationsController.m
@@ -27,15 +27,19 @@ NSString *const WMFNotificationInfoFeedNewsStoryKey = @"feedNewsStory";
 @property (weak, nonatomic) MWKDataStore *dataStore;
 @property (nonatomic, readwrite, copy, nullable) NSData *remoteRegistrationDeviceToken;
 @property (nonatomic, readwrite, strong, nullable) NSError *remoteRegistrationError;
+@property (nonatomic, strong) WMFEchoSubscriptionFetcher *echoSubscriptionFetcher;
+@property (nonatomic, strong) MWKLanguageLinkController *languageLinkController;
 
 @end
 
 @implementation WMFNotificationsController
 
-- (instancetype)initWithDataStore:(MWKDataStore *)dataStore {
+- (instancetype)initWithDataStore:(MWKDataStore *)dataStore languageLinkController:(MWKLanguageLinkController *)languageLinkController {
     self = [super init];
     if (self) {
         self.dataStore = dataStore;
+        self.languageLinkController = languageLinkController;
+        self.echoSubscriptionFetcher = [[WMFEchoSubscriptionFetcher alloc] initWithSession:dataStore.session configuration:dataStore.configuration];
     }
     return self;
 }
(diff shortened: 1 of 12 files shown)


staykids:
@tonisevener Haven't dug into this yet and just wanted to check: I believe this came in before we sync'd about architecture ideas – was there anything you wanted to changed before I reviewed it or is it good to go?

tonisevener:
@staykids I think the only potential thing that came up from architecture review was if AppDelegate is doing too much here. I immediately vend the device token or error to `WMFAppViewController` (and further into `WMFNotificationsController` after that), so hopefully that's fine.

I added some complexity in [e7f4db](https://github.com/wikimedia/wikipedia-ios/pull/4013/commits/e7f4db3bf1e437e097c1c5f4480a344c246a2d1d) that I think I will back out and simplify. I don't think it makes sense to review logic like this until we hash out https://phabricator.wikimedia.org/T271855 and/or https://phabricator.wikimedia.org/T288688. I'll put on hold until I fix this. Thanks for checking!

Wikipedia/Code/WMFNotificationsController.m
```diff
@@ -27,51 +24,74 @@
...
+    UNUserNotificationCenter *center = [UNUserNotificationCenter currentNotificationCenter];
+    [center getNotificationSettingsWithCompletionHandler:^(UNNotificationSettings *_Nonnull settings) {
+        completionHandler(settings.authorizationStatus);
+    }];
+}
+
+- (void)requestPermissionsIfNecessaryWithCompletionHandler:(void (^)(BOOL isAllowed, NSError *__nullable error))completionHandler {
     if (![UNUserNotificationCenter class]) {
         completionHandler(NO, nil);
         return;
```
staykids, line 57: I know this wasn't added in this PR, but I wonder if we can just remove this historical code (also in `sendNotificationWithTitle:`). Maybe I'm missing something, but perhaps it was just some pre-`UNUserNotificationCenter` safety code?

staykids (approved):
Beautiful! So great seeing this stuff starting to flow (even if it's just through the beta cluster).

Going commit by commit to review it was great. And it looks like you edited the description before I could give you the good news about `T287897` 🙂 – I didn't have to use our push utility to manually trigger anything.

- I ignored digging too deep into the Settings changes because as you mentioned this is stuff that'll be revisited in a future task
- I tried both some different thanks and edit reverts and all popped up as `checkEchoV1` pushes on device after about a minute each.

I pushed some minor extraneous space cleanups, and had one optional note we can surely ignore about some historical code.

Wikipedia/Code/WMFNotificationsController.m
tonisevener: I think that's a good suggestion - I pushed a commit removing these checks.

tonisevener:
> And it looks like you edited the description before I could give you the good news about T287897 🙂 – I didn't have to use our push utility to manually trigger anything.

@staykids great! yeah I tried update test steps right quick before you reached the testing phase since I was pretty certain the sandbox flow worked again. Glad it wasn't too confusing even if the test steps were a little off.

PR #4016, opened 2021-08-16
Some edit reverted notification cleanup

tonisevener:

### Notes
This is part one of many for T287298. This task is just some early code cleanup I decided to do before digging into the RemoteNotifications database work. From this point forward I think it would be easier if we thought of edit reverted holistically with the other notification types than tiptoeing around this code. It also removes local notification scheduling for edit revert.

### Test Steps
Nothing major to test for now, just confirm it builds and runs.


tonisevener:
Meant to base this against the feature branch, will reopen a new one.

PR #4017, opened 2021-08-16
Some edit reverted notification cleanup

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T287298

### Notes
This is part one of many for T287298. This task is just some early code cleanup I decided to do before digging into the RemoteNotifications database work. From this point forward I think it would be easier if we thought of edit reverted holistically with the other notification types than tiptoeing around this code. It also removes local notification scheduling for edit revert.

After review, please wait for #4013 to be reviewed and merged. Then update this with feature/notifications and merge.

### Test Steps
Nothing major to test for now, just confirm it builds and runs.

PR #4018, opened 2021-08-16
Add new RemoteNotifications database with extension, remove old database on migration

tonisevener:

### Notes
I believe we had discussed before that there is no need to maintain old remote notification database records, so this PR deletes the old files upon library migration in `MWKDataStore`, and creates a new one by adding the `.sqlite` extension. I found I needed the `.sqlite` extension anyways to open and inspect it, so adding `.sqlite` as an extension while setting up the stack allows it to both be inspectable and to be seen as a new database that needs to be created. There is a brief time period before `performUpdatesFromLibraryVersion` is called where we have both the old database files and the new ones in the app container, but then the old ones are deleted after migration completes.

I am also changing the managed object contexts slightly. `managedObjectContext` (which was already a background context) becomes a more descriptive `backgroundContext`, which will be used as a "write" context (importing, refreshing, and marking as read will happen through here). The new `viewContext` will be used as a "read-only" context, that the views will reference. The `automaticallyMergesChangesFromParent` value and merge policy should allow it to automatically take in changes from the underlying store.

After review, please wait for https://github.com/wikimedia/wikipedia-ios/pull/4017 to be reviewed and merged. Then update this with `feature/notifications` and merge.

To confirm data makes it into the new database, there are a few changes I had to make to allow for more frequent polling and looser API parameters:

1. Return `true` in `isBeforeDeadline` in `RemoteNotificationsOperationsController` [here](https://github.com/wikimedia/wikipedia-ios/blob/T287298-2/WMF%20Framework/Remote%20Notifications/RemoteNotificationsOperationsController.swift#L167).
2. Replace `components.host` with `en.wikipedia.org` (or whatever subdomain you have generated the most notifications on) in `RemoteNotificationsAPIController` [here](https://github.com/wikimedia/wikipedia-ios/blob/T287298-2/WMF%20Framework/Remote%20Notifications/RemoteNotificationsAPIController.swift#L10).
3. Remove `notfilter` field and value from the dictionary in `RemoteNotificationsAPIController` [here](https://github.com/wikimedia/wikipedia-ios/blob/T287298-2/WMF%20Framework/Remote%20Notifications/RemoteNotificationsAPIController.swift#L282).
4. Change `notwikis` value from `wikis.joined(separator: "|")` to `*` in `RemoteNotificationsAPIController` [here](https://github.com/wikimedia/wikipedia-ios/blob/T287298-2/WMF%20Framework/Remote%20Notifications/RemoteNotificationsAPIController.swift#L285).

### Test Steps
1. Run app in simulator on `main`.
2. Switch to this PR branch and run. Open app container and confirm you only see one set of "RemoteNotifications" database files with the `sqlite` extension.
3. Wait for polling to occur (you can proxy if necessary for "notifications" in the url).
4. When you see a polling call, open the `RemoteNotifications.sqlite` database. Confirm you see data. We will capture and persist more fields in a followup PR.

### Screenshots/Videos

Before migration, notice two sets of `RemoteNotifications` database.
Screen Shot 2021-08-16 at 3 55 56 PM

After migration, old database files are gone.
Screen Shot 2021-08-16 at 3 56 09 PM

### DIFF
--- a/WMF Framework/Remote Notifications/RemoteNotificationsOperationsController.swift
+++ b/WMF Framework/Remote Notifications/RemoteNotificationsOperationsController.swift
@@ -19,7 +19,7 @@ class RemoteNotificationsOperationsController: NSObject {
         apiController = RemoteNotificationsAPIController(session: session, configuration: configuration)
         var modelControllerInitializationError: Error?
         modelController = RemoteNotificationsModelController(&modelControllerInitializationError)
-        deadlineController = RemoteNotificationsOperationsDeadlineController(with: modelController?.managedObjectContext)
+        deadlineController = RemoteNotificationsOperationsDeadlineController(with: modelController?.backgroundContext)
         if let modelControllerInitializationError = modelControllerInitializationError {
             DDLogError("Failed to initialize RemoteNotificationsModelController and RemoteNotificationsOperationsDeadlineController: \(modelControllerInitializationError)")
             isLocked = true
@@ -38,6 +38,18 @@ class RemoteNotificationsOperationsController: NSObject {
 
     deinit {
         NotificationCenter.default.removeObserver(self)
+    }
+    
+    func deleteOldDatabaseFiles() throws {
+        let modelName = RemoteNotificationsModelController.modelName
+        let sharedAppContainerURL = FileManager.default.wmf_containerURL()
+        let legacyRemoteNotificationsStorageUrl = sharedAppContainerURL.appendingPathComponent(modelName)
+        let legecyJournalShmUrl = sharedAppContainerURL.appendingPathComponent("\(modelName)-shm")
+        let legecyJournalWalUrl = sharedAppContainerURL.appendingPathComponent("\(modelName)-wal")
+        
+        try FileManager.default.removeItem(at: legacyRemoteNotificationsStorageUrl)
+        try FileManager.default.removeItem(at: legecyJournalShmUrl)
+        try FileManager.default.removeItem(at: legecyJournalWalUrl)
     }
 
     public func stop() {
--- a/WMF Framework/Remote Notifications/RemoteNotificationsOperationsController.swift
+++ b/WMF Framework/Remote Notifications/RemoteNotificationsOperationsController.swift
@@ -40,16 +40,8 @@ class RemoteNotificationsOperationsController: NSObject {
         NotificationCenter.default.removeObserver(self)
     }
     
-    func deleteOldDatabaseFiles() throws {
-        let modelName = RemoteNotificationsModelController.modelName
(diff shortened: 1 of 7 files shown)


WMF Framework/Remote Notifications/RemoteNotificationsOperationsController.swift
```diff
@@ -39,6 +39,18 @@ class RemoteNotificationsOperationsController: NSObject {
...
     }
+    
+    func deleteOldDatabaseFiles() throws {
+        let modelName = RemoteNotificationsModelController.modelName
+        let sharedAppContainerURL = FileManager.default.wmf_containerURL()
+        let legacyRemoteNotificationsStorageUrl = sharedAppContainerURL.appendingPathComponent(modelName)
+        let legecyJournalShmUrl = sharedAppContainerURL.appendingPathComponent("\(modelName)-shm")
+        let legecyJournalWalUrl = sharedAppContainerURL.appendingPathComponent("\(modelName)-wal")
+        
+        try FileManager.default.removeItem(at: legacyRemoteNotificationsStorageUrl)
```
staykids, line 50: I think the preferred way to delete a persistent store is via [destroyPersistentStore](https://developer.apple.com/documentation/coredata/nspersistentstorecoordinator/1468888-destroypersistentstore), though it looks like some folks are reporting this isn't also deleting the meta optimization files for them.

Frankly, I can't personally recall deleting an in-use Core Data store and in place recreating it myself, so I don't have good guidance to offer here. But maybe, because of folks reporting some of their issues with the meta files not also being deleted, and the fact we're intending to rewrite a store in place with a very similar name, your approach seems like the more reasonable approach.

staykids (approved):
Very nice – great seeing content populating. I followed your code changes to trigger the polling, which populated the `RemoteNotifications` store.

One note below about the migration.

WMF Framework/Remote Notifications/RemoteNotificationsOperationsController.swift
tonisevener: @staykids thanks for catching this! I would certainly rather do it the official way, I didn't realize this method existed. I'll explore this a bit today.

tonisevener: @staykids  I pushed a commit calling `destroyPersistentStore` in the migration method, before deleting the old files. It's maybe unnecessary, but I found that `destroyPersistentStore` alone didn't delete the files, and to gain that `sqlite` extension I would rather delete the old files and redo the stack than rename. `destroyPersistentStore` is here really in case it does any extra cleanup internally that I wouldn't know to do. I'll merge EOD or pause that if you have any concerns here. Thanks!

PR #4019, opened 2021-08-16
Save additional notification fields in RemoteNotifications database

tonisevener:

### Notes
Building off of work in https://github.com/wikimedia/wikipedia-ios/pull/4018, this PR captures as much data as I think we will need from the notifications objects returned from the API. This also updates fields that do not need to be optional. A couple of callouts worth noting:

1. Previously we ignored the server-side `read` flag and maintained a local read / seen / excluded state. Now we are simplifying into a single `isRead` flag that is powered off of the API value. I think it's good to keep this simple for now, and add back in a local read / seen / excluded flag later if we find we need it.
2. This also stores the links dictionary from the notifications object as a transformable field in the database. To me this seemed simpler than maintaining a complicated separate set of links managed objects with primary and secondary relationships. It decodes into a `RemoteNotificationsLinks` object.

After review, please wait for #4018 to be reviewed and merged. Then update this with `feature/notifications` and merge.

It would probably help to add the code changes mentioned in https://github.com/wikimedia/wikipedia-ios/pull/4018 to allow for more frequent polling and looser API parameters.

### Test Steps
1. Launch on freshly installed app. Wait for notifications polling to occur.
2. Inspect database, confirm the records added looks right.
3. Terminate app. Add `let messageLinks = readNotifications.first?.messageLinks` right before the `markAsRead` call to the `apiController` in `RemoteNotificationsMarkAsReadOperation` [here](https://github.com/wikimedia/wikipedia-ios/blob/T287298-2/WMF%20Framework/Remote%20Notifications/Operations/RemoteNotificationsMarkAsReadOperation.swift#L9). Also add a breakpoint here. Launch app. When breakpoint is hit, confirm `messageLinks` value looks right.

### DIFF
--- a/WMF Framework/Remote Notifications/Model/RemoteNotificationLinks.swift
+++ b/WMF Framework/Remote Notifications/Model/RemoteNotificationLinks.swift
@@ -44,7 +44,7 @@ public class RemoteNotificationLink: NSObject, NSSecureCoding, Codable {
         let type = decoder.decodeObject(of: NSString.self, forKey: "type")
         let urlString = decoder.decodeObject(of: NSString.self, forKey: "url")
         let label = decoder.decodeObject(of: NSString.self, forKey: "label")
-        let url = urlString == nil ? nil : URL(string: urlString! as String)
+        let url = URL(string: urlString as String? ?? "")
         self.init(type: type, url: url, label: label)
     }
 
--- a/WMF Framework/Remote Notifications/Model/RemoteNotificationLinks.swift
+++ b/WMF Framework/Remote Notifications/Model/RemoteNotificationLinks.swift
@@ -1,0 +1,57 @@
+import Foundation
+
+@objc(RemoteNotificationLinks)
+public class RemoteNotificationLinks: NSObject, NSSecureCoding, Codable {
+    public static var supportsSecureCoding: Bool = true
+    
+    let primary: RemoteNotificationLink?
+    let secondary: [RemoteNotificationLink]?
+
+    init(primary: RemoteNotificationLink?, secondary: [RemoteNotificationLink]?) {
+        self.primary = primary
+        self.secondary = secondary
+    }
+
+    public required convenience init(coder decoder: NSCoder) {
+        
+        let primary = decoder.decodeObject(of: RemoteNotificationLink.self, forKey: "primary")
+        let classes = [NSArray.classForCoder(), RemoteNotificationLink.classForCoder()]
+        let secondary = decoder.decodeObject(of: classes, forKey: "secondary") as? [RemoteNotificationLink]
+        self.init(primary: primary, secondary: secondary)
+    }
+
+    public func encode(with coder: NSCoder) {
+        coder.encode(primary, forKey: "primary")
+        coder.encode(secondary, forKey: "secondary")
+    }
+}
+
+@objc(RemoteNotificationLink)
+public class RemoteNotificationLink: NSObject, NSSecureCoding, Codable {
+    public static var supportsSecureCoding: Bool = true
+    
+    let type: String?
+    let url: URL?
+    let label: String?
+
+    init(type: NSString?, url: URL?, label: NSString?) {
+        self.type = type as String?
+        self.url = url
+        self.label = label as String?
+    }
+
+    public required convenience init(coder decoder: NSCoder) {
+        let type = decoder.decodeObject(of: NSString.self, forKey: "type")
+        let urlString = decoder.decodeObject(of: NSString.self, forKey: "url")
+        let label = decoder.decodeObject(of: NSString.self, forKey: "label")
+        let url = urlString == nil ? nil : URL(string: urlString! as String)
+        self.init(type: type, url: url, label: label)
+    }
+
+    public func encode(with coder: NSCoder) {
+        coder.encode(type, forKey: "type")
+        let urlString = url?.absoluteString
+        coder.encode(urlString, forKey: "url")
+        coder.encode(label, forKey: "label")
+    }
+}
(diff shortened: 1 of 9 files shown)


staykids:
>  Add let messageLinks = readNotifications.first?.messageLinks right before the markAsRead call to the apiController in RemoteNotificationsMarkAsReadOperation here. Also add a breakpoint here. Launch app. When breakpoint is hit, confirm messageLinks value looks right.

@tonisevener For clarity, you're meaning to confirm that the `primary` and `secondary` URLs here match what's persisted with Core Data for a particular `RemoteNotifications` entry, right?

tonisevener:
> @tonisevener For clarity, you're meaning to confirm that the primary and secondary URLs here match what's persisted with Core Data for a particular RemoteNotifications entry, right?

@staykids Sorry for being vague there - yeah I mainly just want to make sure that this `messageLinks` property is saving and transforming properly to and from Core Data and that we're not losing any `url` or `label` data. You can proxy and compare the results against the the json response we receive from the notifications API `https://en.wikipedia.org/w/api.php?action=query&format=json&formatversion=2&meta=notifications&notformat=model&notlimit=max&notwikis=*` call. I ended up doing that because I don't think I was able to inspect `messageLinks` in the database files myself. Also sorry the link no longer works - here's a screenshot of me inspecting the stuff. You can inspect the notifications ID too to help you find that object in the API json response. The urls don't seem to show in the variables view, just when I right click and print it to the console.

Screen Shot 2021-09-15 at 1 51 36 PM

staykids:
> @staykids Sorry for being vague there - yeah I mainly just want to make sure that this `messageLinks` property is saving and transforming properly to and from Core Data and that we're not losing any `url` or `label` data.

Cool - no worries, thanks for confirming!

WMF Framework/Remote Notifications/Model/RemoteNotificationLinks.swift
```diff
@@ -0,0 +1,57 @@
...
+        self.type = type as String?
+        self.url = url
+        self.label = label as String?
+    }
+
+    public required convenience init(coder decoder: NSCoder) {
+        let type = decoder.decodeObject(of: NSString.self, forKey: "type")
+        let urlString = decoder.decodeObject(of: NSString.self, forKey: "url")
+        let label = decoder.decodeObject(of: NSString.self, forKey: "label")
+        let url = urlString == nil ? nil : URL(string: urlString! as String)
```
staykids, line 47: To make the logic a bit more direct and readable, what do you think about something like:
```suggestion
        let url = URL(string: urlString as String? ?? "")
```

WMF Framework/Remote Notifications/Model/RemoteNotificationsModelController.swift
```diff
@@ -160,25 +150,32 @@ final class RemoteNotificationsModelController: NSObject {
...
             assertionFailure("Notification should have a date")
             return
         }
-        guard let id = notification.id else {
-            assertionFailure("Notification must have an id")
-            return
-        }
 
-        let message = notification.message?.header?.wmf_stringByRemovingHTML()
+        let isRead = notification.readString == nil ? NSNumber(booleanLiteral: false) : NSNumber(booleanLiteral: true)
```
staykids, line 158: Little question to clarify my understanding about `readString`/the JSON response's `read` property value: is it expected that the API returns a date string if a notification has been read (presumably the date the notification was marked as read), but nil for that property otherwise?

staykids (approved):
👍🏽 – one minor code suggestion and one question below.

WMF Framework/Remote Notifications/Model/RemoteNotificationsModelController.swift
tonisevener: @staykids this is pretty investigative, but yeah that's my understanding. The [notifications API](https://www.mediawiki.org/wiki/Notifications/API) doesn't have documentation on the response, but I've been pointed to [this](https://github.com/wikimedia/mediawiki-extensions-Echo/blob/66bb02fd8b8e74cd0277b20a26b000abcb35dcc2/includes/DataOutputFormatter.php#L134) area of code for determining what optional values to expect from here. If you follow that through ([here](https://github.com/wikimedia/mediawiki-extensions-Echo/blob/66bb02fd8b8e74cd0277b20a26b000abcb35dcc2/includes/model/Notification.php#L207) > [here](https://github.com/wikimedia/mediawiki-extensions-Echo/blob/4dd2a651e154fed4689ecdb7da6f3536a5c77f96/includes/model/Notification.php#L156) > [here](https://github.com/wikimedia/mediawiki-extensions-Echo/blob/37c602c5326ecd171b8eec3eedadade987cbdb26/includes/gateway/UserNotificationGateway.php#L75), it seems like the database table's `notification_read_timestamp` is set with the current date here within this `markRead` method, and that ultimately bubbles up to the API response value. I also found this comment [here](https://github.com/wikimedia/mediawiki-extensions-Echo/blob/839b833033a5cc4286ca0618b12188b670b87562/echo.sql#L43) that the `notification_read_timestamp` is the "Timestamp when the user read the notification, or null if unread".

WMF Framework/Remote Notifications/Model/RemoteNotificationLinks.swift
tonisevener: Done - thanks for the suggestion!

WMF Framework/Remote Notifications/Model/RemoteNotificationsModelController.swift
staykids: Wow great detective work 🕵️

PR #4020, opened 2021-08-18
Remove legacy polling and add new notification fetch operation

tonisevener:

### Notes
This adds the ability to trigger the RemoteNotificationsController to fetch the first page (read and unread) of all app preferred language notifications (+ wikidata and commons), to be triggered from the UI layer. It also has a helper method to generate an NSFetchedResultsController. 

Future PRs will fetch or bulk import additional pages, add mark as read logic, add fetched results controller tab predicates and whatever else we may need. This is just to make the Core Data + UI connection.

After review, please wait for #4019 to be reviewed and merged. Then update this with feature/notifications and merge.

**Note:** As of now, these notification calls only work in the production scheme. I will make fixes up the stack in https://github.com/wikimedia/wikipedia-ios/pull/4010 to get this working in the beta cluster, so we can test end-to-end from a push notification.

### Test Steps
@staykids I purposefully left out any UI code to test this, but attached are my prototype screens that I was comparing against if interested. They automatically fetch and populate a basic table view.

[NotificationCenterViewController.txt](https://github.com/wikimedia/wikipedia-ios/files/7010410/NotificationCenterViewController.txt)
[NotificationCenterTableViewCell.txt](https://github.com/wikimedia/wikipedia-ios/files/7010411/NotificationCenterTableViewCell.txt)

### Screenshots/Videos
[link to gif](https://user-images.githubusercontent.com/3620196/130106306-7387c45e-498a-4487-84f6-e596075d4b15.gif)


WMF Framework/Remote Notifications/Operations/RemoteNotificationsFetchFirstPageOperation.swift
```diff
@@ -0,0 +1,41 @@
+class RemoteNotificationsFetchFirstPageOperation: RemoteNotificationsOperation {
+    let languageCode: String
+    init(with apiController: RemoteNotificationsAPIController, modelController: RemoteNotificationsModelController, languageCode: String) {
+        self.languageCode = languageCode
+        super.init(with: apiController, modelController: modelController)
+    }
+    override func execute() {
+        self.backgroundContext.perform {
```
staykids, line 8: The intermixing of Core Data and NSOperation is something I have less experience with, so please push back if something I’m saying seems off, but would it make sense for these operations to use a fresh background context (e.g. `container.newBackgroundContext()`) instead? I have some thread worry with these NSOperation’s all using the same shared `RemoteNotificationsModelController` `backgroundContext` .

Sorry I shoulda thought about this more in #4018 but I failed to fully consider in what contexts this context (unintentional pun) might be used.

tonisevener: @staykids oh no problem on not getting to it earlier, you didn't know where this was heading. 

This was mainly inspired by reading around that it's helpful to have a read-only view context and a write-only background context, but I took that to mean a single background context. Looking around at similar examples in our own codebase (`feedImportContext`) and a [sample](https://developer.apple.com/documentation/coredata/loading_and_displaying_a_large_data_feed) of one Apple's, it does look like these background contexts are spun up in the moment right before importing.

I **think** this particular single background context would be fine, as long as I call `perform` (which I now think is not necessary on this line, because it's already within `createNewNotifications`) then it should be thread-safe if the background context is a serial queue. But I also have less experience with this and it's hard to argue with those two examples doing it the way you describe. I'll rework to spin up a `newBackgroundContext` in the moment.

staykids: >  I'll rework to spin up a `newBackgroundContext` in the moment.

👍🏽

> This was mainly inspired by reading around that it's helpful to have a read-only view context and a write-only background context, but I took that to mean a single background context. 

I think this pattern is certainly valid when we can ensure the context _itself_ is always being accessed on the same thread it was created.

> I **think** this particular single background context would be fine, as long as I call `perform` (which I now think is not necessary on this line, because it's already within `createNewNotifications`) then it should be thread-safe if the background context is a serial queue.

I could certainly be wrong, but I think this is where the context's usage within an `NSOperation` throws a wrench in the the works. The background context itself is first created on the main thread, and the notification fetch Operations are created on the main thread with a reference to the main thread background context, but all the Operations themselves are performing work on any number of threads that _aren't_ the main thread. While the context's internal queue may be serial/private, it being created on a main thread then accessed on a non-main thread may not be its internal expectation and cause unexpected strangeness beyond my understanding. I just popped open https://developer.apple.com/documentation/coredata/nsmanagedobjectcontext and the first paragraph of the `Concurrency` section at the bottom actually talks about this a bit.

A side note (just want to share, totally not suggesting we do this here – it seems like just creating a new background context in the operation will be fine for our needs): `NSOperation` certainly doesn't make it easy, but I typically favor architecting this type of async network work where the operation itself just does its unit of work and provides back some data, and then the caller itself then decides what to do with that data – so in this case, essentially moving the Core Data stuff out of the Operation itself a layer up into the call site. That way 1) the operation itself has less dependencies and its purpose is more direct (in this case, the fetch operation just being a simple network operation with no side effects) and 2) we can constrain the thread pool (in this case, we can use the same single background context on the main thread).

tonisevener: @staykids Aah, I see, my main rule in my head has been was that managed object saving/manipulation must happen on the context's queue, I didn't notice that `NSManagedObjectContext` itself must only ever be referenced from the queue it was created on, though that makes sense now. My appreciation for you catching this grows more!

> A side note 

Thanks for this, I'll explore that while I work on my feedback today. I do feel a bit like if an operation only does a simple fetch, then the operation abstraction starts to feel like overkill. But it does make sense to move references to background contexts out of here, so we aren't tempted to reintroduce this issue in the future.

tonisevener: @staykids this fix has been made in https://github.com/wikimedia/wikipedia-ios/pull/4020/commits/787834767c418c5a50d2e21c96eaa91e012a196c. I am instantiating a new background context in the operation's `init` method, and I am asserting that this and the reference during execution occurs on the same thread. Some legacy dead code still uses the model controller's `backgroundContext` property (now renamed to `legacyBackgroundContext`), I didn't want to turn this into too big of a cleanup PR for that, so I listed it as a TODO. It's related to mark as read so we can clean it out when we tackle [T287302](https://phabricator.wikimedia.org/T287302).

staykids: Thanks for rolling with the changes on this one – hopefully the reward of our work here is that Core Data doesn’t throw us cryptic errors at some random point in the future 🥴. I have a bit more feedback about getting this to the finish line and figured it best to throw it in a branch: https://github.com/wikimedia/wikipedia-ios/compare/56fc59d7798e...0855e5932f38

Take a look and lemme know what you think, or if any of my thinking seems off. Notes on the changes in the branch:

- I think the occurrences of `DispatchQueue.main.async`ing and `assert(Thread.isMainThread)`ing might be unnecessary. While they do a good job of constraining the `backgroundContext` to the main thread, that may not actually be desirable and makes the approach a bit indirect. We want our `NSOperations` to perform their work on whatever thread the OS decides they need to with minimal side effects. In this case, I think the `DispatchQueue.main.async`ing out in the operations is not needed. Instead, we can rely on the `RemoteNotificationsFetchFirstPageOperation`  itself doing its work on whatever thread the OS tells it to, using a background context in that operation for data operations to save the data we’re interested in, then the  `operationQueue` itself’s completion being a main/calling thread signal as it is now.

- Removing the `DispatchQueue.main.async`ing, and letting the `NSOperation`’s do their work on whatever thread the OS decides, exposes a minor threading issue. The thread on which a `RemoteNotificationsFetchFirstPageOperation` is initialized and the thread `execute` is performed within may actually be different threads. For this reason, using the managed object context initialized in the initializer in the `execute(...)` method means we’re back to our original problem of passing a context initialized in one thread to a different thread. I’d propose we just create a new `backgroundContext` when we need it in `execute` – this should resolve our cross-thread context issue by creating the context just in time when we need it on the thread it’s going to be used in.

tonisevener: @staykids definitely agreed - excellent and well-reasoned suggestions as always. Thanks for helping me clean this up! I merged your branch in.

staykids (approved):
👍🏽

PR #4024, opened 2021-08-25
Add empty Notifications Center

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T288654

### Notes
Adds empty Notifications Center view to app, displayed when tapping the Notifications Bell in either the Explore feed or Settings when Explore is off.

- Adds empty `UICollectionView`
- Adds empty `NotificationsCenterViewModel` to transit data/`RemoteNotificationsController` to view controller
- Modifies `ViewControllerTransitionsController` to support standard push/pop animation style
   - [This approach](https://github.com/wikimedia/wikipedia-ios/blob/30dc4fcc59603d8cd8686b654753d632b06a0b7d/Wikipedia/Code/ViewControllerTransitionsController.swift#L110
) may be slightly controversial. One might argue it'd be better to just have a property on the `ViewController` base class to flag this desired behavior, but I think that would create an unclear expectation of what that property does. Truly, we're just wanting to avoid the transition style for `ViewController`s under the influence of `ViewControllerTransitionsController` – not all `ViewController`s period. So while this may look a little odd, I think it focuses behavioral expectations better than adding a `ViewController` bool.
- Adds `NotificationsCenterPresentationDelegate` to allow `WMFAppViewController` to handle conditional presentation logic of the Notifications Center based on the origin `ViewController`. I'd imagine keeping this presentation logic in `WMFAppViewController` makes sense, as this is where we'll likely want to handle and conditionally present the Notifications Center on incoming deep link taps of iOS push notifications as well.

### Test Steps
1. Confirm Notifications Center displays in Explore feed and Settings when Explore is off when tapping the Notifications Bell button
2. Confirm it displays in either device screen orientation
3. Ensure its appearance and disappearance animations are the standard navigation push/pop style
4. Ensure other Explore feed animation transitions (like tapping a featured article) perform as they did before

### Screenshots
Notifications Center


Wikipedia/Code/WMFSettingsViewController.m
```diff
@@ -103,7 +103,7 @@ - (void)viewWillAppear:(BOOL)animated {
 
 - (void)configureNotificationsButton {
     if ([[NSUserDefaults standardUserDefaults] defaultTabType] == WMFAppDefaultTabTypeSettings && [[[self dataStore] authenticationManager] isLoggedIn]) {
-        UIBarButtonItem *notificationsBarButton = [[UIBarButtonItem alloc] initWithImage:[UIImage imageNamed:@"notifications-bell"] style:UIBarButtonItemStylePlain target:self action:nil];
+        UIBarButtonItem *notificationsBarButton = [[UIBarButtonItem alloc] initWithImage:[UIImage imageNamed:@"notifications-bell"] style:UIBarButtonItemStylePlain target:self action:@selector(userDidTapNotificationsCenter)];
```
tonisevener, line 106: I haven't looked into it, but for some reason this button doesn't show on Settings when I launch the app after turning the Explore feed off.

Wikipedia/Code/WMFAppViewController.m
```diff
@@ -1938,6 +1939,7 @@ - (nonnull WMFSettingsViewController *)settingsViewController {
         WMFSettingsViewController *settingsVC =
             [WMFSettingsViewController settingsViewControllerWithDataStore:self.dataStore];
         [settingsVC applyTheme:self.theme];
+        _settingsViewController.notificationsCenterPresentationDelegate = self;
```
tonisevener, line 1942: Also haven't looked into this, but when I turn off the Explore feed and tap the notifications button on the Settings screen, it doesn't navigate away. So it seems like this delegate might be nil in that case.

Wikipedia/Code/NotificationsCenterViewController.swift
```diff
@@ -0,0 +1,52 @@
+import UIKit
```
tonisevener, line 1: I think our tab vs. spaces settings might differ - I notice these new class indentions are different. Here are my settings:

Screen Shot 2021-09-03 at 12 47 06 PM

tonisevener (changes requested):
Found a few things, but it's ok if you want to tackle any of them (besides maybe indentions) in a followup PR. Thanks!

Wikipedia/Code/NotificationsCenterViewController.swift
```diff
@@ -0,0 +1,52 @@
...
+		self.viewModel = viewModel
+		super.init(theme: theme)
+	}
+
+	required init?(coder aDecoder: NSCoder) {
+		fatalError("init(coder:) has not been implemented")
+	}
+
+	override func loadView() {
+		super.loadView()
```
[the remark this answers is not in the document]
staykids: Great catch! I think their recommendation is slightly misleading – since the default implementation of `loadView` seems to populate `view` with a valid default value, there may be cases where it's not unreasonable to call it (like maybe if your custom view is quite simple and you just want to build it in place).

I think it works as implemented now (though isn't optimized) because it's effectively just assigning a value to `view` twice (in fact, we do it in a [couple](https://github.com/wikimedia/wikipedia-ios/blob/32ca59d18352c20419605c6b469cc530f304948d/Wikipedia/Code/ExploreCardViewController.swift#L57) [other](https://github.com/wikimedia/wikipedia-ios/blob/32ca59d18352c20419605c6b469cc530f304948d/Wikipedia/Code/ReferenceBackLinksViewController.swift#L62) places). But, who knows what side effects lurk in the `UIKit` source – I pushed a commit removing it!

Wikipedia/Code/WMFSettingsViewController.m
staykids: Let's cover this one in https://phabricator.wikimedia.org/T288669 – that seems like the best opportunity to be sure all our bell display logic is solid.

Wikipedia/Code/NotificationsCenterViewController.swift
staykids: Yeah, I'm set to `Prefer Ident Using` tabs here. But to make the files ident using spaces, regardless of a contributor's Xcode settings, I tend to use the File Inspector to set the `Ident Using` property explicitly to `Spaces` on a per file basis. I missed setting it on this file, which I just updated and pushed.

Wikipedia/Code/WMFAppViewController.m
staykids: This was just a silly ordering mistake by me – updated.

staykids:
Thanks for looking at this – I think I've addressed your feedback.

tonisevener (approved):
Looks good, thanks! I also added that setting you mentioned to the project file, maybe that'll help.

PR #4029, opened 2021-08-31
Increase deployment target & fix warnings

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T289535

### Notes
This is part 1 of 2. This fixes all the low-hanging fruit warnings after increasing the deployment target to iOS13, as well as some code cleanup. Warning fixes are grouped by commit into type of warning, explained by the commit message. For the last commit I decided to revert previous fixes in third party libraries and suppress the warnings, to match patterns we've taken in the past.

After this PR, the remaining warnings are related to using deprecated peek and pop APIs, which are a different beast and will be handled in a follow up PR.

### Test Steps
I think we can lean on QA for regression testing this (possible regression areas are in Phab task). For code review I would just confirm that the project builds and runs, and that nothing looks obviously wrong while exploring the app a bit.


Wikipedia/Code/UIApplication+WindowWorkarounds.swift
```diff
@@ -0,0 +1,12 @@
+
+import Foundation
+
+extension UIApplication {
+    @objc var workaroundKeyWindow: UIWindow? {
```
mcleinman, line 5: We can remove `@objc` since we're not using it for the objc class anymore, right?

Wikipedia/Code/UIApplication+WindowWorkarounds.swift
```diff
@@ -0,0 +1,12 @@
...
+        return windows.first { $0.isKeyWindow }
+    }
+    
+    @objc var workaroundStatusBarFrame: CGRect {
```
mcleinman, line 9: Same as above: We can remove `@objc` since we're not using it for the objc class anymore, right?

mcleinman (commented):
Looking great and testing great. Just a handful of small comments. I'm also happy to make the updates if you want.

Thanks for doing this - lots of small research bits to keep us modern. Much appreciated!

tonisevener:
@mcleinman great catches 😅 thanks! I think I have addressed them all, but let me know if I missed any.

mcleinman (approved):
This looks wonderful! Thank you!

mcleinman:
I believe we're ready to merge this one, but wanted to confirm. Please let me know - or feel free to merge it.

PR #4034, opened 2021-09-14
Enable Search Swipe to Save or Share

colealanroberts:
https://phabricator.wikimedia.org/T208892

### Notes
* Enables Save / Share actions when swiping a search result

### Test Steps
1. Perform a search
2. Swipe left on the cell to reveal the two new Save and Share actions.

### Screenshots/Videos

https://user-images.githubusercontent.com/53830238/133188945-d953afcb-87b8-4a1e-a148-e0e6f6bb6012.mp4

https://user-images.githubusercontent.com/53830238/133188956-98d14f39-5af7-484e-b95e-41cfd7c8108d.mp4

### DIFF
--- a/Wikipedia/Code/SearchResultsViewController.swift
+++ b/Wikipedia/Code/SearchResultsViewController.swift
@@ -102,11 +102,11 @@ class SearchResultsViewController: ArticleCollectionViewController {
         }
         cell.configureForCompactList(at: indexPath.item)
         cell.setTitleHTML(result.displayTitleHTML, boldedString: resultsInfo?.searchTerm)
-       
         cell.articleSemanticContentAttribute = MWKLanguageLinkController.semanticContentAttribute(forContentLanguageCode: contentLanguageCode)
         cell.titleLabel.accessibilityLanguage = languageCode
         cell.descriptionLabel.text = descriptionForSearchResult(result)
         cell.descriptionLabel.accessibilityLanguage = languageCode
+        cell.actions = availableActions(at: indexPath)
         if layoutOnly {
             cell.isImageViewHidden = result.thumbnailURL != nil
         } else {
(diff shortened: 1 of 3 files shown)


colealanroberts:
This is my first PR for Wikipedia iOS!

I believe I've moved the ticket to the correct release ([6.9.1](https://phabricator.wikimedia.org/tag/ios-app-v6.9.1/)) which I see was created today. However, I'm not familiar with the release cycle and perhaps the next release (6.8.X) is correct?

Any insight would be greatly appreciated. Thanks!

Wikipedia/Code/SearchResultsViewController.swift
```diff
@@ -102,11 +102,11 @@ class SearchResultsViewController: ArticleCollectionViewController {
         }
         cell.configureForCompactList(at: indexPath.item)
         cell.setTitleHTML(result.displayTitleHTML, boldedString: resultsInfo?.searchTerm)
-       
         cell.articleSemanticContentAttribute = MWKLanguageLinkController.semanticContentAttribute(forContentLanguageCode: contentLanguageCode)
         cell.titleLabel.accessibilityLanguage = languageCode
         cell.descriptionLabel.text = descriptionForSearchResult(result)
         cell.descriptionLabel.accessibilityLanguage = languageCode
+        cell.actions = availableActions(at: indexPath)
```
mcleinman, line 109: Could we change this to a pattern used elsewhere in the app?
```suggestion
        editController.configureSwipeableCell(cell, forItemAt: indexPath, layoutOnly: layoutOnly)
```

mcleinman (commented):
Thanks for this! A great improvement! It works great - just one suggestion for you.

mcleinman:
> This is my first PR for Wikipedia iOS!
> 
> I believe I've moved the ticket to the correct release ([6.9.1](https://phabricator.wikimedia.org/tag/ios-app-v6.9.1/)) which I see was created today. However, I'm not familiar with the release cycle and perhaps the next release (6.8.X) is correct?
> 
> Any insight would be greatly appreciated. Thanks!

It can be hard to know. In this case, I'm going to move this into 6.9, which is going to be our next release.

Wikipedia/Code/SearchResultsViewController.swift
colealanroberts: @mcleinman Excellent, I've made that change!

mcleinman (approved):
This is wonderful, thank you!

PR #4036, opened 2021-09-17
fix crash in SearchViewController

mcleinman:
**Phabricator:** N/A, but follow up from https://phabricator.wikimedia.org/T208892

### Notes
* I believe we're using the wrong check here. We want to make sure there are some search results (otherwise we should be allowing deletes, as the cell is a historical search term), but we need to make sure the index is smaller than the search results, not the number of saved historical search terms.
* This was crashing on device but not on sim b/c my sims had fewer historical saved searches.

### Test Steps
1. Search screen - ensure deleting old search terms work. 
2. Make several searches.
2. Ensure swiping right on search results works
3. Cancel the search, ensure deleting old search terms still works.


Wikipedia/Code/SearchViewController.swift
```diff
@@ -532,12 +532,9 @@ class SearchViewController: ArticleCollectionViewController, UISearchBarDelegate
     override func articleURL(at indexPath: IndexPath) -> URL? {
         // This guard checks prevents an out of bounds exception, this is due to
```
tonisevener, line 533: I could totally be reading this wrong, but I think we should revert this method and the method below back to what it originally was, `return nil`. It seems to me that `SearchViewController`'s `collectionView` only displays a list of the persisted recent search terms, whereas `SearchResultsViewController` only displays a list of search results from the network. We shouldn't be mixing and matching these data sources.

These two methods are here because it's superclass asserts false if they aren't implemented. Really cells on `SearchViewController` have little to do with articles, it's just persisted terms they have searched before. In hindsight we probably should never have subclassed `SearchViewController` from `ArticleCollectionViewController`, because we inherit and are expected to implement unnecessary methods like these.

mcleinman: These have been reverted.

mcleinman:
I’m was running into a corner case - if you have a search results screen active and save an article from a swipe action - and then navigate to the Saved tab, delete that article - and then navigate back to our search results…. in that case the swipe action is out of sync (unless you scroll enough to trigger cellForRowAt). Hitting an assertion failure. This now listens for updates to articles from other parts of the app, and updates the swipe actions appropriately.

colealanroberts:
I believe I introduced this bug. Admittedly, tested using the simulator and not on device. @mcleinman want me to take a look at this again and reimplement the desired behavior?

Wikipedia/Code/SearchResultsViewController.swift
```diff
@@ -123,5 +124,12 @@ class SearchResultsViewController: ArticleCollectionViewController {
         collectionView.backgroundColor = theme.colors.midBackground
     }
 
+    @objc func updateVisibleCells() {
```
tonisevener, line 127: @mcleinman Works great! Just a small suggestion - I noticed other listeners of this notification inspect the article from the notification and exit early if it's not something the screen is interested in. Currently all cells on this screen will get reloaded for any article that changes throughout the app, whether it's on this screen or not. Can you change it to something like this so we're only configuring the cell of the updated article?

```
@objc func updateArticleCell(_ notification: NSNotification) {
        guard
            let updatedArticle = notification.object as? WMFArticle,
            let updatedArticleKey = updatedArticle.inMemoryKey
        else {
            return
        }
        
        guard updatedArticle.hasChangedValuesForCurrentEventThatAffectSavedState else {
            return
        }
        
        for indexPath in collectionView.indexPathsForVisibleItems {
            guard articleURL(at: indexPath)?.wmf_inMemoryKey == updatedArticleKey else {
                continue
            }
            
            guard let cell = collectionView.cellForItem(at: indexPath) as? ArticleRightAlignedImageCollectionViewCell else {
                continue
            }
            
            configure(cell: cell, forItemAt: indexPath, layoutOnly: false)
        }
    }
```

tonisevener:
> I believe I introduced this bug. Admittedly, tested using the simulator and not on device. @mcleinman want me to take a look at this again and reimplement the desired behavior?

@colealanroberts it's no problem at all - this fix is really close to being merged and I think we'll be good after that. Thanks for offering though!

Wikipedia/Code/SearchResultsViewController.swift
mcleinman: Good catch. Just added this (and combined some of the `guard`s. Let me know if you think there are any more tweaks I should make!

tonisevener (approved):
Looks good, thanks!

PR #4041, opened 2021-09-23
Implement all notification cell types

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T288658

### Notes
Adds Notifications Center cell and designs for our notification types.

- Implement the design of a notifications cell
- Ensure each of our notification types can be represented in a notifications cell
- Ensure the cell has two display presentations (read, unread) in three display states (default, edit selected, edit unselected) that match designs
- Ensure cell's leading image container can transform into a check-style selection control
- Add View Model (that will [later be populated with data](https://phabricator.wikimedia.org/T288662) from the Remote Notifications Controller) that represents all the notifications content for display in a cell

As I mentioned before, this PR is rather large. The commit history may help a bit, but if it feels easier to review as a sequence of smaller PRs, let me know and I'm certainly happy to break it up as so!

### Test Steps

1. Log in with an account that has notifications
2. Confirm notification cells appear in Notifications Center (ignore the actual content/data/images of the notifications, that's covered in a future Phab task)
3. Confirm notifications support the cell states below (using Figma as our visual guide)

#### Cell States Configuration

Unfortunately, the current steps to test the different presentation states of the cells is quite manual at the moment. It would benefit us to eventually have a test database of `RemoteNotification`s that include a notification of every type with every reasonable configuration. 

- **Notification Type**: `RemoteNotificationType` (thanks Toni) represents our available notification type. Update the type returned by `NotificationsCenterCellViewModel`'s `var notificationType: RemoteNotificationType?` to see each type, the image for that type, and the colors and visually match with Figma. This is not exhaustive.
- **Read/Unread**: cells can be presented as either read or unread. The color and font weights alter depending on this state. Update the Display State used within `NotificationsCenterCell`'s `func updateCellStyle(forDisplayState displayState: NotificationsCenterCellDisplayState)` to see the range of available presentations.
- **Selected/Unselected**: as well as being presented as read or unread, a cell can also be in edit mode as either selected or unselected. In edit mode, the leading image transforms from a notification type image to a selected or unselected checkmark. Update the same Display State above to see these presentations.


tonisevener (commented):
This is looking great! I have some minor code comments and also a couple of things I noticed when running:

1. When the cell display style is `.defaultUnread`, it seems like the date, subheader, metaActionButton label text sizes don't increase immediately when the dynamic text sizes changes (i.e. background on this screen, change size, foreground). Other labels do, and these labels also do when the display style is `defaultRead`, so I think it must be a bug somewhere.
2. While on this screen, changing the device to dark/light mode then foregrounding doesn't change the cell's theme, if app reading preferences are set to default.

Wikipedia/Code/NotificationsCenterCellStyle.swift
```diff
@@ -0,0 +1,176 @@
...
+            return UIImage(systemName: "link", withConfiguration: UIImage.SymbolConfiguration(weight: .semibold))
+        case .thanks:
+            return UIImage(named: "notifications-type-thanks")
+        case .welcome, .translationMilestone(_), .editMilestone:
+            return UIImage(systemName: "heart.fill")
+        case .loginFailKnownDevice, .loginFailUnknownDevice, .loginSuccessUnknownDevice,
+             .unknownSystemAlert, .unknownAlert:
+            return UIImage(named: "notifications-type-login-notify")
+        case .emailFromOtherUser:
+            return UIImage(systemName: "mail", withConfiguration: UIImage.SymbolConfiguration(weight: .semibold))
```
tonisevener, line 168: FYI, discovered this after initial review - it doesn't look like this system icon shows up in iOS13.

Wikipedia/Code/NotificationsCenterCell.swift
```diff
@@ -0,0 +1,360 @@
...
+
+		label.insets = UIEdgeInsets(top: 4, left: 4, bottom: -4, right: -4)
+
+		return label
+	}()
+
+	lazy var projectSourceImage: UIImageView = {
+		let imageView = UIImageView()
+		imageView.translatesAutoresizingMaskIntoConstraints = false
+		imageView.image = UIImage(named: "notifications-project-commons")
```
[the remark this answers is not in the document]
staykids: Are you referring to these?
notifications-projects

If so, I've already included them in the `Images` Asset Catalog in this PR (sorry, I know it's a big PR).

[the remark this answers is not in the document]
tonisevener: Oh great! Yeah that's what I meant, I just missed it.

Wikipedia/Code/NotificationsCenterCellStyle.swift
staykids: Great catch on this – looks like the `mail` SF Symbol is iOS 14+. I replaced it with `envelope` for iOS 13. I could have just done some nil coalescing (`??`) [here](https://github.com/wikimedia/wikipedia-ios/pull/4041/commits/c28b97ea70e4d76ecd3c073d5fc526f809c3b83f#diff-b450f3bce014695328d8d0c58c2a112de015165cfe0912a26fd36369741c97f2) to be more direct, but I think calling out the `if #available` is valuable here to communicate in code the reason for this fallback. I also audited the rest of these notification type images and I believe that was the only case of iOS 14+ specific SF Symbols here.

We do use `doc.plaintext.fill` which is iOS 14+ in the `NotificationCenterCell`’s `metaActionButton`, but that is temporary – I added a comment.

staykids:
Awesome - got your code review notes addressed. Lemme know if I missed anything.

> 1. When the cell display style is `.defaultUnread`, it seems like the date, subheader, metaActionButton label text sizes don't increase immediately when the dynamic text sizes changes

I'm gonna need more time to investigate this one.

> 2. While on this screen, changing the device to dark/light mode then foregrounding doesn't change the cell's theme, if app reading preferences are set to default.

👍🏽 I followed the pattern of elsewhere in the app to also [reloadData](https://github.com/wikimedia/wikipedia-ios/pull/4041/commits/c1db61b2b6602f67f937665a0a53832826c7d5e3), which is kind of heavy but I suppose not unreasonable.

Wikipedia/Code/NotificationsCenterCell.swift
```diff
@@ -0,0 +1,342 @@
+import UIKit
+
+protocol NotificationsCenterCellDelegate: AnyObject {
+	func userDidTapSecondaryActionForCellIdentifier(id: String)
+}
+
+final class NotificationsCenterCell: UICollectionViewCell {
```
tonisevener, line 7: @staykids One more thing - the RTL scheme has things looking slightly off.  The date label needs to be right aligned and I think the truncation is on the wrong side:


Screen Shot 2021-09-29 at 9 47 23 AM

staykids: > @staykids One more thing - the RTL scheme has things looking slightly off. The date label needs to be right aligned 

Good catch – updated the time label to change its text alignment depending on the user interface layout direction. 

> and I think the truncation is on the wrong side:

Regarding the truncation, my thinking here could be wrong given the _interesting_ ways we support LTR/RTL, but copy and pasting some RTL language text into the labels, it looks like the truncation is already properly happening on the left side.

RTL

tonisevener: @staykids aah nice, I didn't actually try pasting in RTL text. Thanks for confirming!

staykids:
> > 1. When the cell display style is `.defaultUnread`, it seems like the date, subheader, metaActionButton label text sizes don't increase immediately when the dynamic text sizes changes

This was a strange one. I think it ultimately came down to this line:

https://github.com/wikimedia/wikipedia-ios/blob/46fde6ebc917d844953e9abf6307821909d46043/WMF%20Framework/UIFont%2BWMFDynamicType.swift#L126

Non-regular weights (like the time label) were relegated to using a fixed font size rather than a dynamic preferred one, and thus did not scale until the collection view requested a cell to reconfigure itself. 

I don't feel it's wise to try to make a change directly to this type system without significantly more testing time because of how high touch it is, so I chose to instead follow a pattern seen elsewhere in the app. The reason it seems other views don't appear to encounter this scaling bug is because they listen for changes to the trait collection's `preferredContentSizeCategory` and reload the collection view's data if one has occurred.

I followed that approach here. Like the theme change, it's again a heavy `reloadData` operation, but I suppose it's also defensible given how infrequently it'd be called.

staykids:
@tonisevener I believe I've addressed everything - this one's ready for another look!

tonisevener (approved):
Looks great!

PR #4042, opened 2021-09-24
[DRAFT] Add empty state view to Notifications Center

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T288668


staykids:
Looking back at this, there are a few corner cases I want to address before opening it for review.

staykids:
@tonisevener This one should be ready for a look!

tonisevener (commented):
Looking good! Just one comment on the bouncing. I tried to set this fix up in a suggestions branch but it's getting away from me, so it's either harder than it sounds or I'm extremely rusty. Still I think it's worth a (maybe timeboxed) attempt. We could also spin off a polish Phab task around this to try tackling later on.

staykids:
@tonisevener Should be ready for a final look!

PR #4047, opened 2021-10-01
Implement notifications import operation

tonisevener:

### Notes
[Description TBD]


tonisevener:
> The final test is failing - when I relaunch the app and go to notificaiton center, it starts pulling my notifications from the start. Are you seeing the same behavior, @tonisevener ?
> 
> All other tests are working well for me.

@mcleinman Hmm, nope I'm not seeing that behavior on my machine. Let's sync up and see if we can figure out what's going on.

PR #4062, opened 2021-10-27
Fix link in article has wrong prefix "app://"

kojiGit55:
**Phabricator:** 
https://phabricator.wikimedia.org/T291298

### Notes
* Currently, link in article has a custom url scheme `app://`, so when it's copied & pasted into other App, the link doesn't work. So I changed `app://` to `https://` to fix this

### Test Steps
1. Go to any article
2. Copy a sentence with a link in it
3. Open up Apple's Notes app
4. Paste copied text

### DIFF
--- a/Wikipedia/Code/ArticleViewController.swift
+++ b/Wikipedia/Code/ArticleViewController.swift
@@ -35,7 +35,6 @@ class ArticleViewController: ViewController, HintPresenting {
     /// Called when initial JS setup is complete
     @objc public var initialSetupCompletion: (() -> Void)?
     
-    internal let schemeHandler: SchemeHandler
     internal let dataStore: MWKDataStore
     
     private let cacheController: ArticleCacheController
@@ -100,7 +99,6 @@ class ArticleViewController: ViewController, HintPresenting {
         self.article = article
         
         self.dataStore = dataStore
-        self.schemeHandler = schemeHandler ?? SchemeHandler(scheme: "app", session: dataStore.session)
         self.cacheController = cacheController
         
         super.init(theme: theme)
@@ -123,7 +121,6 @@ class ArticleViewController: ViewController, HintPresenting {
     lazy var webViewConfiguration: WKWebViewConfiguration = {
         let configuration = WKWebViewConfiguration()
         configuration.processPool = ArticleViewController.webProcessPool
-        configuration.setURLSchemeHandler(schemeHandler, forURLScheme: schemeHandler.scheme)
         return configuration
     }()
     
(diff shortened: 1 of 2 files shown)


Wikipedia/Code/ArticleViewController.swift
```diff
@@ -123,7 +121,6 @@ class ArticleViewController: ViewController, HintPresenting {
     lazy var webViewConfiguration: WKWebViewConfiguration = {
         let configuration = WKWebViewConfiguration()
         configuration.processPool = ArticleViewController.webProcessPool
-        configuration.setURLSchemeHandler(schemeHandler, forURLScheme: schemeHandler.scheme)
```
tonisevener, line 126: Our article view is heavily dependent on this scheme handler for supporting offline mode in saved articles. Without this, visiting a saved article without a network connection no longer displays. Rather than touching this area, I suggest trying to modify the links at the moment they are copied. Here are some approaches that might work:

1. Suppress the contextual menu's default "Copy" button (in [WMFWebView.m](https://github.com/wikimedia/wikipedia-ios/blob/160c00fa4e4e104854e4057697c57480a504244a/Wikipedia/Code/WMFWebView.m#L6), we're already doing this with `_share`) and provide our own custom one with implementation (in `ArticleViewController+ContextMenu.swift`, [updateMenuItems](https://github.com/wikimedia/wikipedia-ios/blob/160c00fa4e4e104854e4057697c57480a504244a/Wikipedia/Code/ArticleViewController%2BContextMenu.swift#L2) method. Again see `shareMenuItem` here as an example.). In this implementation we'll probably need to go across the javascript bridge to copy the selected text in the web view, then replace any instances of "app://" with "https://", then save it to the [clipboard](https://developer.apple.com/documentation/uikit/uipasteboard).
2. Override `WMFWebView`'s or `ArticleViewController`'s `copy:(id)sender` [method](https://developer.apple.com/documentation/uikit/uiresponderstandardeditactions/2354191-copy) and do some custom handling to tweak the links. This seems to get called right when the copy button is pressed.
3. Listen for `UIPasteboard`'s `changedNotification` (doc [here](https://developer.apple.com/documentation/uikit/uipasteboard/1622104-changednotification)) in `ArticleViewController` and modify the [clipboard](https://developer.apple.com/documentation/uikit/uipasteboard) in your handler method.

I tried each of these approaches and got closest with the 3rd, so that might be the one to aim for.

tonisevener (changes requested):
Hi @kojiGit55! Thanks so much for the PR. Unfortunately we have an unusual setup with our article view that we'll need to preserve for offline support, so we'll need to try a different approach for this fix. I documented a few options you can try in my comments.

That being said, since this ticket is more complicated than we originally thought, feel free to close this PR if you don't have the time to dig into this - we will pick it back up when we're able. Thanks for helping us think this one through.

tonisevener:
Hi @kojiGit55! Since this PR hasn't seen movement in a while, I am going to go ahead and close it. Feel free to open up another when the comments in https://github.com/wikimedia/wikipedia-ios/pull/4062#pullrequestreview-796935664 are addressed. Thanks again!

PR #4046, opened 2021-10-01
Determine new notifications in extension and display

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T288773
https://phabricator.wikimedia.org/T285353
& Some of https://phabricator.wikimedia.org/T287033

### Notes
This extends the work done in #4026 to pull the latest unread push notifications from the notifications API, filters out previously seen notifications via an app container cache file and displays the results. If the result is a single new unread notification, it modifies the push content to display it's `header` value. There is also some extra handling for talk page notification types (adds subtitle, bundles if necessary) based on the [Figma mocks](https://www.figma.com/file/cedgOU5CyOR0UVqtjDOvzE/iOS-Notifications?node-id=549%3A1752). For all other situations, it defaults to "New activity on Wikipedia" text.

After review, please wait for #4044 to be reviewed and merged. Then update this with feature/notifications and merge into that feature branch.

These steps should be tested through the [Beta Cluster](https://en.wikipedia.beta.wmflabs.org/wiki/Main_Page) via the Staging scheme. 

### Test Steps
1. Fresh install from this branch and log in from an account that has all checkboxes checked on desktop Wikipedia notifications preferences. Go to Settings > Notifications and enable notifications.
2. Do something to trigger a notification to yourself from another account on desktop Wikipedia. I tend to test by thanking an edit, undoing an edit, or writing on my talk page with another account.
3. Wait for notifications to appear on your phone. On Staging scheme / beta cluster desktop it takes about 10 seconds for me lately.
4. The first notification will likely say "New activity on Wikipedia", since all notifications appear new to the device, and it sees it as condensing multiple notification types.
5. Trigger another notification (thanks or undo edit).
6. Now you should see a push with content specific push notification content.
7. Go to your user talk page and write on it.
8. Now you should see a push with content specific push notification content, plus a "New message" subtitle.
9. Get a couple of browser tabs open with editing your talk page. Try to post two updates back to back to trigger a batched talk page notification
10. Now you should see a push with batched content, "N new messages on {Title}", as well a "New messages" subtitle.
11. Go to desktop notifications preferences, uncheck "Thanks" from the app column.
12. Trigger a thanks again to yourself. Confirm no notification appears on your device.

PR #4064, opened 2021-10-27
Notifications - add primary and secondary routing from cells

tonisevener:
**Phabricator:**
https://phabricator.wikimedia.org/T288670

### Notes
This PR adds the primary and secondary tap targets and routing in our Notifications Center cells.  It expands on https://github.com/wikimedia/wikipedia-ios/pull/4044 in filling out the `NotificationsCenterCellViewModel` with more data. I am also adding the ability for talk pages to deep link directly to a reply.

### Test Steps
1.. Run on the main/production scheme with an account that has lots of notifications. It's helpful if you're able to find an account with notifications on Commons, Wikidata, and multiple languages. Be sure to add those other languages as an app language in Settings before going to Notification Center. I can share my testing credentials if needed.
2. Browse through notification center via the Explore bell icon. Tap various cells (cell as a whole, and the header link) and see that they are pushed to another screen. In particular find a user talk page cell that has the section header populated. Tap that, and notice that it navigates to a particular reply. There are a lot of types, so we should lean on QA for a more thorough test of that logic.

Google sheet of type > tap destinations: https://docs.google.com/spreadsheets/d/10Gkjd6GWFNDXzPnZhDJ2NvPqNvqoTTCFHMnSUtf3ex8/edit#gid=0


Wikipedia/Code/NotificationsCenterViewModel+LinkExtensions.swift
```diff
@@ -0,0 +1,263 @@
...
+        return components.url
+    }
+    
+    var connectionWithWikidataItemURL: URL? {
+
+        //Note: Sample notification json indicates that the wikidata item link is the second secondary link.
+        //Return this link if we're fairly certain it's what we think it is
+        
+        guard let secondaryLinks = notification.secondaryLinks,
+              secondaryLinks.count > 1,
```
staykids, line 198: This is fine as is, but just wanted to mention that generally for safe indexing purposes I'd favor a technique that directly checks the availability of that index (`secondaryLinks.indicies.contains(1)`) rather than relying on a secondary property (`count`).

Wikipedia/Code/TalkPageContainerViewController.swift
```diff
@@ -92,9 +92,22 @@ class TalkPageContainerViewController: ViewController, HintPresenting {
...
+    //Overlay and activity indicator that displays while routing to reply thread
+    private lazy var replyRoutingOverlay: UIView = {
+        let view = UIView()
+        view.translatesAutoresizingMaskIntoConstraints = false
+        view.backgroundColor = theme.colors.paperBackground
+        return view
+    }()
+    private lazy var replyRoutingActivityIndicator: UIActivityIndicatorView = {
+        let view = UIActivityIndicatorView(style: .medium)
+        view.color = theme.isDark ? .white : .gray
```
staykids, line 107: This same color convention is already applied in other places so please push back if it doesn't make sense, but wouldn't it be preferable to just use `theme.colors.primaryText`, so the indicator always renders with the same readability as text?

Wikipedia/Code/NotificationsCenterCell.swift
```diff
@@ -66,10 +68,17 @@ final class NotificationsCenterCell: UICollectionViewCell {
         label.textAlignment = .left
         label.numberOfLines = 1
         label.setContentCompressionResistancePriority(.defaultLow, for: .horizontal)
-        label.setContentHuggingPriority(.required, for: .horizontal)
+        label.setContentHuggingPriority(.defaultLow, for: .horizontal)
```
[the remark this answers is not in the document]
tonisevener: @staykids My only thinking with this particular change was to let the `headerLabel` expand so that it's easier for the user to tap, rather than having `relativeTimeAgoLabel` be the one that takes up most of the container view. I can revert this though if you think it's not necessary or causes a layout regression.

The other constraint tweaks in this file were necessary for the tap gesture on `headerLabel` to fire. After inspecting the view hierarchy,  I noticed the `headerTextContainer`'s width was evaluating to zero (even though it's subviews were displaying fine!), which I think caused the tap gesture recognizer on the headerLabel to be ignored. Hopefully these other tweaks didn't cause any layout regressions. Let me know if you see any!

Wikipedia/Code/NotificationsCenterCellStyle.swift
```diff
@@ -21,14 +21,35 @@ struct NotificationsCenterCellStyle {
...
             return theme.colors.error
         default:
             return theme.colors.link
         }
     }
+    
+    func isHeaderLabelTapGestureEnabled(_ displayState: NotificationsCenterCellDisplayState) -> Bool {
+
+        guard displayState.isUnread else {
+            return false
```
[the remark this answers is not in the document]
tonisevener: @staykids you're right, that was a wrong assumption on my part. Also confirmed with @carolynlimadeo.

Wikipedia/Code/NotificationsCenterCellStyle.swift
```diff
@@ -21,14 +21,35 @@ struct NotificationsCenterCellStyle {
...
         case .loginFailKnownDevice, .loginFailUnknownDevice, .loginSuccessUnknownDevice:
             return theme.colors.error
         default:
             return theme.colors.link
         }
     }
+    
+    func isHeaderLabelTapGestureEnabled(_ displayState: NotificationsCenterCellDisplayState) -> Bool {
+
+        guard displayState.isUnread else {
```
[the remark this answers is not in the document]
tonisevener: @staykids good catch, this is done. Can't really test it yet until the editing mode PR is merged, but the logic should be there now.

Wikipedia/Code/NotificationsCenterViewModel+LinkExtensions.swift
tonisevener: @staykids good technique, the intention feels more clear with that way. This is updated.

Wikipedia/Code/TalkPageContainerViewController.swift
tonisevener: @staykids that does make sense, I'm not sure why it's been done in this pattern in the past. Maybe holdover from the style having incorporated only white and gray options? I made this change.

Wikipedia/Code/NotificationsCenterCell.swift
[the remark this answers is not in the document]
staykids: Gotcha – thanks for the details! I haven't yet noticed a layout regression with the change.

tonisevener:
@staykids slipped one link change commit just now after a sync this morning with @carolynlimadeo - now login notifications have a secondary link. The spreadsheet has already been updated.

Wikipedia/Code/NotificationsCenterCell.swift
```diff
@@ -291,6 +301,35 @@ final class NotificationsCenterCell: UICollectionViewCell {
...
+        
+        headerLabelTapGestureRecognizer.isEnabled = isHeaderLabelTapGestureEnabled
+    }
+}
+
+//MARK: - Private
+
+private extension NotificationsCenterCell {
+    
+    var isHeaderLabelTapGestureEnabled: Bool {
```
staykids, line 313: Nice, it looks you introduced some logical separation within `NotificationsCenterCell` here to help readability. It looks like you moved `isHeaderLabelTapGestureEnabled` here as well. I'm not suggesting you change anything here, this approach certainly works and I'm happy to merge. I just want to outline my rationale behind originally suggesting `NotificationsCenterViewModel` as its destination instead.

A property like this computed in the view model more strongly adheres to view models as something that transform more complex objects/logic into simpler "primitives" for consumption by a view, and views as something with less or only needed responsibility/knowledge of the outside world. Same vibe as how `headerText` etc. are computed in the view model now for naive consumption by the view. My approach with `NotificationsCenterCellStyle` isn't perfect - ideally, it'd be computed and originate from the view model as well, but its reliance on the active trait collection makes it a bit more direct to just treat as a view utility.

If, for example, the view model computed a boolean like `shouldAllowSecondaryTapAction` (with the same logic as `isHeaderLabelTapGestureEnabled`), the cell itself's call site could be `headerLabelTapGestureRecognizer.isEnabled = viewModel.shouldAllowSecondaryTapAction`. This moves the bulk of this logic responsibility to the view model (which has enough knowledge about the involved objects to distill the logic into a simple boolean for use by the view), reduces the responsibility of the view itself (it more naively just takes the minimum data needed to do its job of configuring itself), and as a naming bonus I think that the call site reads more naturally.

Again, I'm happy to merge this as-is, just wanted to provide my thought process if this thinking seems useful for our future decision making!

Wikipedia/Code/NotificationsCenterCell.swift
```diff
@@ -315,7 +354,7 @@ final class NotificationsCenterCell: UICollectionViewCell {
         projectSourceLabel.layer.borderColor = cellStyle.projectSourceColor.cgColor
         projectSourceImage.tintColor = cellStyle.projectSourceColor
 
-        selectedBackgroundView?.backgroundColor = cellStyle.selectedCellBackgroundColor
+        //selectedBackgroundView?.backgroundColor = cellStyle.selectedCellBackgroundColor
```
staykids, line 354: See optional note about uncommenting.

staykids (approved):
Wow, so beautiful using the routing in action! Great work. I only have so many test notification types, so relying on QA for thorough testing of routing correctness like you suggest makes sense. 

Just an architectural comment and one optional change on the background view left below. Lemme know – I'm happy to merge as-is.

Wikipedia/Code/NotificationsCenterCell.swift
tonisevener: @staykids agh, I'm sorry, I misread your original comment as asking it to be in the cell. Your explanation totally makes sense, and I like that you take any knowledge of the how out of the computed property name. I added your suggestion. Thanks as always for the careful eye 😬.

staykids: No problem at all, thanks for the change – glad the explanation was helpful.

Wikipedia/Code/NotificationsCenterCell.swift
```diff
@@ -197,7 +206,7 @@ final class NotificationsCenterCell: UICollectionViewCell {
         let topMargin: CGFloat = 13
         let edgeMargin: CGFloat = 11
 
-        selectedBackgroundView = UIView()
+        //selectedBackgroundView = UIView()
```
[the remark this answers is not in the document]
tonisevener: @staykids Thanks for checking and confirming! I think that's a remnant comment from early on in this PR when I didn't want to consider deselecting the cell. I reverted these. I think #4046 has the most workable selection state for these cells so far, and it has these lines uncommented as well. Of course you're free to change this selection logic however you need for your cell swiping.

PR #4066, opened 2021-10-28
Notifications - add swipe actions from cells

tonisevener:
**Phabricator:**
https://phabricator.wikimedia.org/T288687 (data and routing portion, mark as read to come in a followup PR)

### Notes
This PR establishes the swipe action label and routing urls for each notification. It builds on the swipe gesture recognizer from https://github.com/wikimedia/wikipedia-ios/pull/4063 and the url generation helper methods in https://github.com/wikimedia/wikipedia-ios/pull/4064, so do not merge until those are in place. For now we are using a standard action sheet / popover to display these options - it will be replaced with a custom sheet in the future.

### Test Steps
1. Run on the main/production scheme with an account that has lots of notifications. It's helpful if you're able to find an account with notifications on Commons, Wikidata, and multiple languages. Be sure to add those other languages as an app language in Settings before going to Notification Center. I can share my testing credentials if needed.
2. Browse through notification center via the Explore bell icon. Swipe right on each cell and confirm the action sheet labels look right. Tap one of the custom options in the action sheet (i.e. not mark as read or notification settings). Confirm the user routes to the correct place.

**Optional**
I have a custom payload of notifications I've gathered to mimic some harder-to-reach notifications. If interested you can proxy and map the https://en.wikipedia.org/w/api.php?action=query&format=json&formatversion=2&meta=notifications&notfilter=read%7C!read&notformat=model&notlimit=max&notwikis=enwiki call to this before navigating to Notifications Center. In general though, we shouldn't get too in depth on this testing and instead lean on QA. Feel free to reach out to me for this file.


Wikipedia/Code/NotificationsCenterViewController.swift
```diff
@@ -175,32 +175,49 @@ extension NotificationsCenterViewController: NotificationCenterViewModelDelegate
...
+                })
+            case .custom(let data):
+                alertAction = UIAlertAction(title: data.text, style: .default, handler: { alertAction in
+                    let url = data.url
+                    self.navigate(to: url)
+                })
+            }
+            
+            alertController.addAction(alertAction)
+        }
```
staykids, line 213: Currently, after presenting the action sheet, the user can’t dismiss it without selecting an action. We should add a cancel `UIAlertAction` to this action sheet. That’ll allow both tapping `Cancel` to close the sheet and tapping outside the action sheet to dismiss it.

Wikipedia/Code/NotificationsCenterCellViewModel+SwipeActionExtensions.swift
```diff
@@ -0,0 +1,445 @@
+
+import Foundation
+
+extension NotificationsCenterCellViewModel {
+    
+    enum SwipeAction {
```
staykids, line 6: I'd like to propose renaming `SwipeAction` (and these related types/occurrences/file name) to something that doesn't reference `Swipe`, like `NotificationSheetAction`, `SheetAction`, or `ActionSheetAction` or something. My reason being, `Swipe` is slightly confusing in this context, as it implies a cell's swipe actions (like [existing API](https://developer.apple.com/documentation/uikit/uiswipeactionsconfiguration)) instead of what they are in this case which is sheet actions.

Since it's a nested type of `NotificationCenterCellViewModel`, maybe the short `SheetAction` is most direct. What do you think?

Wikipedia/Code/NotificationsCenterViewController.swift
tonisevener: @staykids good idea, this is done!

Wikipedia/Code/NotificationsCenterCellViewModel+SwipeActionExtensions.swift
tonisevener: @staykids that sounds good, I like `SheetAction` since sheet is a generic enough term to cover this action sheet and our future half sheet modal.

I did a mindless find & replace, but I can trim it down more if you think it helps. Seems like some of those local `var sheetActions` could be simplified to just `var actions`.

Wikipedia/Code/NotificationsCenterViewController.swift
staykids: @tonisevener 👍🏽 

I think the preferable approach here would actually be using the `.cancel` style and dropping the handler parameter: `let cancelAction = UIAlertAction(title: CommonStrings.cancelActionTitle, style: .cancel)`

This allows the tapping outside of the sheet to dismiss it action, and also logically separates the Cancel button from the primary sheet actions to allow a persistent escape hatch for users as the list of actions grows.

staykids: The `.cancel` style approach also prevents iPad presentations of the sheet from displaying the extraneous cancel action, as a tap outside the sheet to dismiss it is the popover presentation default behavior.

Wikipedia/Code/NotificationsCenterCellViewModel+SheetActionExtensions.swift
```diff
@@ -0,0 +1,445 @@
...
+        let markAsReadOrUnreadActionData = SheetActionData(text: markAsReadOrUnreadText, url: nil)
+        sheetActions.append(.markAsReadOrUnread(markAsReadOrUnreadActionData))
+        
+        switch notification.type {
+        case .userTalkPageMessage:
+            sheetActions.append(contentsOf: userTalkPageActions(for: configuration))
+        case .mentionInTalkPage,
+             .editReverted:
+            sheetActions.append(contentsOf: mentionInTalkAndEditRevertedPageActions(for: configuration))
+        case .mentionInEditSummary: //done
```
staykids, line 32: Minor: leftover `//done` comment, unless I'm conceptually missing something about its purpose.

staykids (approved):
Nice – great to play with this stuff with real notifications (again, using the notification types available to me)!

Optional minor comment about a comment, and a minor update to the cancel `UIAlertAction` presentation style.

Wikipedia/Code/NotificationsCenterViewController.swift
tonisevener: @staykids doh! This is done. Thanks again for cleaning up my sloppiness.

Wikipedia/Code/NotificationsCenterCellViewModel+SheetActionExtensions.swift
tonisevener: @staykids Good catch, those were testing comments to keep track of notification types I was testing. This is removed.

PR #4069, opened 2021-11-01
Notifications - add mark as read functionality

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T287302
https://phabricator.wikimedia.org/T288687

### Notes
This PR sets up some basic functionality of marking notifications as read or unread, or marking all notification as read via the MediaWiki notifications API. It builds on PR #4066 to trigger marking a single notification as read or unread via the swipe action sheet option. I decided to do a bit of a fire-and-forget approach on these without completion blocks, but I'm open to altering it here or later when we work on the other mark as read [task](https://phabricator.wikimedia.org/T288665).

Note that even though I am triggering one of these methods through a swipe action, the state won't update itself because the data flow [PR](https://github.com/wikimedia/wikipedia-ios/pull/4049) is not merged yet. For a preview of all PRs working together you can test the `notifications-plus-prs-2` branch.

### Test Steps
1. Fresh launch app. Login and go to Notifications Center. Also open the Notifications Center on desktop Wikipedia (https://en.wikipedia.org/wiki/Special:Notifications), and go to the "Unread" section in a browser.
2. Swipe on one of your notifications on device and see the the action sheet display. Tap the "Mark as read/unread" option.
3. When you swipe and mark something as unread, the notification should appear on this Desktop unread section after a browser refresh. When you swipe again mark something as read, the notification should disappear from this Desktop unread section after a browser refresh. 
4. Repeat step 2 to mark a few notifications as unread, ideally notifications from multiple languages or projects. Refresh in step 3 to confirm they show as unread on Desktop.
5. Add temporary hook to mark all as read when tapping the edit button [here](https://github.com/wikimedia/wikipedia-ios/blob/410766f45385a272d63a5b47e1294f0a4c368b94/Wikipedia/Code/NotificationsCenterViewController.swift#L81): `viewModel.markAllAsRead()` 
6. Rerun project and tap Edit button. Refresh desktop Wikipedia. Confirm all project notifications were marked as read.


WMF Framework/Remote Notifications/Model/RemoteNotificationsModelController.swift
```diff
@@ -222,32 +200,39 @@ final class RemoteNotificationsModelController: NSObject {
...
-
-    public func markAsRead(notificationWithID notificationID: String) {
-        processNotificationWithID(notificationID) { (notification) in
-            notification.isRead = true
+        
+        //it is assumed notifications are from view context. We want to only write on background context.
+        var keys: [String] = []
+        viewContext.performAndWait {
+            keys = notifications.compactMap { $0.key }
         }
```
staykids, line 214: Since we already have the set of `RemoteNotification`s here and we're only interested in the stable `key` property on a `RemoteNotification`, is it necessary to access their managed object context here?

Would it be sufficient to replace this with just:
```swift
let keys = notifications.compactMap { $0.key }
```

staykids (commented):
Nice – been working well confirming changes against `Special:Notifications` as you suggested!

One note below.

WMF Framework/Remote Notifications/Model/RemoteNotificationsModelController.swift
tonisevener: @staykids I might be overly cautious here but I was worried about accessing the notifications' `key` property from a non-main thread. Even though the notifications passed in originated from the view context, this method here is called from the `RemoteNotificationsMarkReadOrUnreadOperation`, which is likely not running on the main thread. Let me know if I'm thinking of this wrong though. I think another pattern I've seen referenced is to pass around `NSManagedObjectID`s instead, and use those to call `object(with objectID: NSManagedObjectID)` on a background context to pull background notifications and flip their `isRead` flag. That might be a more clear and common way to do this.

staykids: @tonisevener If the objective is Core Data thread safety in the operation, maybe the best solution is just avoiding dealing with passing around the `NSManagedObject` `RemoteNotification`s as much as possible in it. What if instead of passing a `Set<RemoteNotification>` to the controller at the beginning of the process in the view model, we just pass a `Set<String>` of `RemoteNotification.key`s? This functionally provides what the operation needs (a bunch of notification keys to operate with) without having to do any `NSManagedObjectIDs` or cross-context object access dancing.

Then the sequence looks like: main thread provides a bunch of `String`s to the remote notifications controller (from Notifications Center view model, something like `remoteNotificationsController.markAsReadOrUnread(notificationKeys: Set(notificationKeys), shouldMarkRead: shouldMarkRead)`), controller spins up an operation that operates on non-main thread, operation creates a new managed object context to operate with on that non-main thread, and just uses those non-`NSManagedObject`, non-context bound `key` Strings to fetch with.

I'm certainly not an expert in this area though. If you'd prefer to leave it as-is lemme know and I can merge.

tonisevener: @staykids that is a good idea and much simpler, I like it. I'm working towards that the api controller also expects to be fed notifications in the operation, where it extracts the notification `id` and `wiki` properties. So there was already a potential threading bug there that this surfaced. Just giving you a heads up that this will be slightly more complicated than just switching over to `Set<String>` in those spots. I'll keep working at it but might not get this in by EOD.

staykids: > where it extracts the notification `id` and `wiki` properties.  ... Just giving you a heads up that this will be slightly more complicated than just switching over to `Set<String>` in those spots.

@tonisevener If it's helpful at all: if you're looking for something to simplify all call sites and pass around the pertinent data, maybe a new type could help. This would probably provide better context to the method parameters and also keep the call sites simple.

```swift
extension RemoteNotification {
    public typealias Identifiers = (key: String?, id: String?, wiki: String?)

    public var identifiers: Identifiers {
        return (key, id, wiki)
    }
}
```
	
Unfortunately, I think you run into hashable limitations for these tuples to be Sets, so maybe a plain ole struct would do.

```swift
extension RemoteNotification {
	public struct Identifiers: Hashable {
		let key: String?
		let id: String?
		let wiki: String?
	}

	public var identifiers: Identifiers {
		return Identifiers(key: key, id: id, wiki: wiki)
	}
}
```

Maybe there's a more apt name for the struct, best I could think of quickly.

> I'll keep working at it but might not get this in by EOD.

No worries at all on the timeline.

tonisevener: @staykids thanks for the tip! I basically used what you suggested, but used the name `IdentifierGroup`. I don't love it, but it helps that it's a singular form since I'm passing arrays and sets of this type around, and `Identifiers` having a property of `id` felt weird as well.

tonisevener:
@staykids I pushed one more commit fixing bugs that were discovered while building the user testing branch. The API threw errors when marking notifications as read across multiple wikis in a single call, so I am breaking them down by wiki and calling to each separate project, similar to how we do importing and refreshing. I didn't catch it before because so far the test steps only had us mark a single notification, which worked fine. Test steps are mostly the same, but I also updated them with a way to test marking all notifications as read.

staykids:
> @staykids I pushed one more commit fixing bugs that were discovered while building the user testing branch. The API threw errors when marking notifications as read across multiple wikis in a single call, so I am breaking them down by wiki and calling to each separate project, similar to how we do importing and refreshing.

Nice – so far with the latest commit, adding in the temporary call to `markAllAsRead()` as you noted has been working fine to mark a group of separate project notifications (English, Spanish, Arabic, Commons, and Wikidata in one of my test account's case) as read without any console API errors as far as I can tell. Going to do a final pass of the code.

staykids (approved):
Nice! I think your choice of `IdentifierGroup` is much better than my original suggestion. A minor optional change below if you also feel it reads better.

WMF Framework/Remote Notifications/Model/RemoteNotificationsModelController.swift
```diff
@@ -214,33 +157,95 @@ final class RemoteNotificationsModelController: NSObject {
...
+            guard let dictionaries = (try? moc.fetch(fetchRequest)) as? [[String: String]] else {
+                completion([])
+                return
+            }
+
+            let results = dictionaries.reduce([String]()) { results, item in
+                var array = results
+                array.append(contentsOf: item.values)
+                return array
+            }
```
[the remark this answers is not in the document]
tonisevener: @staykids Nice, thanks! I updated this.

PR #4073, opened 2021-11-08
Use shared cookie storage for fundraising announcement logic

tonisevener:
**Phabricator:**
https://phabricator.wikimedia.org/T294996

### Notes
We made some changes to cookie storage earlier in the notifications feature branch (see #4026), which caused a regression with fundraising announcements. Earlier releases (tested 6.8.1 and 6.8.0) work fine, but this needs to be in place before we release the notifications feature.

### Test Steps
**Before**
1. Switch to `feature/notifications` branch.
2. Run on device on main scheme. Proxy device to watch traffic. When explore feed shows, background app, then foreground and pull to refresh the Explore feed to cause announcements to be fetched.
3. Proxy will show https://en.wikipedia.org/api/rest_v1/feed/announcements call, which for now is empty. Map this call locally to this expected 2021 fundraising [json](https://github.com/wikimedia/wikipedia-ios/files/7499033/2021fundraising.txt). 
4. Terminate app and change the date to sometime in December. Relaunch app. After launch proxy should show announcements call again, this time with your locally mapped json as a response.
5. Visit an article on EN Wikipedia. Fundraising announcement should show, but it does not.

**After**
1. Reset device date, uninstall app, and switch to this PR branch and re-run from main scheme with proxy still running and locally mapped.
2. Background app, foreground, pull to refresh on Explore feed to be sure your locally mapped announcements are fetched.
3. Terminate app, change the date to sometime in December. Relaunch app.
4. Visit an article on EN Wikipedia. Fundraising announcement now shows.


staykids (approved):
Nice find – working as expected with your test steps.

PR #4075, opened 2021-11-12
Update to Xcode 13.1

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T291382

### Notes
This attempts to fix bugs that building with Xcode 13.1 surfaces, as well as update various settings files to the latest (Circle CI tests, build server version target, etc.)

### Test Steps
These are the areas that changed, so here's some optional regression test steps you can go through in iOS 13, 14, & 15. I will also add these to the Phabricator task for QA to do heavier testing with.
1. Navigation bars. Particularly Settings navigation bars and the Editing view navigation bar, where I was seeing UI bugs. Sometimes they showed as clear for me, with the UI underneath poking through. Test theming for these as well.
2. The tab bar. Similar to navigation bars, it's background color sometimes showed as clear. Test changing the theme through Settings, Article view, and the Editing flow, and be sure the tab bar updates correctly.
3. Settings section headers and footers theming and spacing. These weren't theming well on iOS15. I updated these to a custom header / footer view (which we were already halfway using on these screens) which already themed properly.
4. Collection view cell & header/footer sizes - made a change to fix a console error here. Look at Explore feed, Saved article tab, History tab, and Search results cell sizing.
5. Announcements deserialization - launch app, background, foreground, then pull to refresh on Explore feed. Announcements will be fetched. Before https://github.com/wikimedia/wikipedia-ios/pull/4075/commits/c7e762f89395bbf9e7d3abb33d9c074918a63c06, a console error would show at this point with something like:

```
 -[NSKeyedUnarchiver validateAllowedClass:forKey:] allowed unarchiving safe plist type 'NSString' for key 'NS.objects', even though it was not explicitly included in the client allowed classes set: '{(
    "'WMFMTLModel',
    "'NSArray'
)}'. This will be disallowed in the future.
```

Confirm this console error no longer shows. 

To confirm fundraising announcement still works:
1. Go to an article, confirm no fundraising announcement shows. 
2. Enter airplane mode and terminate the app.
3. Change device date time to sometime in December.
4. Relaunch app via Xcode. Confirm you still see no NSKeyedUnarchiver console errors.
5. Go to an article. Confirm you see fundraising announcement. Dismiss popup.
6. Revisit article, confirm you no longer see fundraising announcement. Terminate app.
7. Relaunch app, visit article, confirm you still no longer see fundraising announcement.


Wikipedia/Code/WMFThemeableNavigationController.m
```diff
@@ -34,30 +34,19 @@ - (UIStatusBarStyle)preferredStatusBarStyle {
...
-            backgroundImage = theme.sheetNavigationBarBackgroundImage;
-            break;
-        case WMFThemeableNavigationControllerStyleGallery:
-            self.modalTransitionStyle = UIModalTransitionStyleCrossDissolve;
-            [self setNavigationBarHidden:YES animated:NO];
-            break;
-        default:
-            backgroundImage = theme.navigationBarBackgroundImage;
-            break;
+    UINavigationBarAppearance *appearance = [UINavigationBarAppearance appearanceForTheme:theme style:self.style];
```
tonisevener, line 39: Note this is going off of advice found here: https://developer.apple.com/forums/thread/682420

Wikipedia/Code/WMFSettingsViewController.m
```diff
@@ -489,13 +493,28 @@ - (NSInteger)tableView:(UITableView *)tableView
...
+    if (header) {
+
+        if (@available(iOS 15.0, *)) {
+
+        } else {
+            //Hides odd content inset bug in iOS 13 & 14
+            if (section == 0) {
+                return 0;
+            }
+        }
```
mcleinman, line 507:
```suggestion
        if (@available(iOS 15.0, *) == false && (section == 0)) {
                //Hides odd content inset bug in iOS 13 & 14
                return 0;
        }
```

Wikipedia/Code/WMFThemeableNavigationController.m
```diff
@@ -34,30 +34,19 @@ - (UIStatusBarStyle)preferredStatusBarStyle {
...
+    self.navigationBar.standardAppearance = appearance;
+    self.navigationBar.scrollEdgeAppearance = appearance;
+    self.navigationBar.compactAppearance = appearance;
+
+    if (@available(iOS 15.0, *)) {
+        //do nothing
+    } else {
```
mcleinman, line 46:
```suggestion
    if (@available(iOS 15.0, *) == false) {
```

Wikipedia/Code/UINavigationBarAppearance+Extensions.swift
```diff
@@ -0,0 +1,27 @@
...
+        appearance.backgroundColor = theme.colors.chromeBackground;
+        appearance.titleTextAttributes = theme.navigationBarTitleTextAttributes
+        let backgroundImage: UIImage?
+        switch style {
+            
+        case .editor:
+            backgroundImage = theme.editorNavigationBarBackgroundImage
+        case .sheet:
+            backgroundImage = theme.sheetNavigationBarBackgroundImage
+        case .gallery:
```
mcleinman, line 17: Does navigation bar need to be hidden in this case as well?

mcleinman: From looking at the old obj c code.

Wikipedia/Code/UINavigationBarAppearance+Extensions.swift
```diff
@@ -0,0 +1,27 @@
...
+            backgroundImage = nil
+        default:
+            backgroundImage = theme.navigationBarBackgroundImage
+        }
+        
+        appearance.backgroundImage = backgroundImage
```
mcleinman, line 23:
```suggestion
        switch style {
        case .editor:
            appearance.backgroundImage = theme.editorNavigationBarBackgroundImage
        case .sheet:
            appearance.backgroundImage = theme.sheetNavigationBarBackgroundImage
        case .gallery:
            appearance.backgroundImage = nil
        default:
            appearance.backgroundImage = theme.navigationBarBackgroundImage
        }
```

mcleinman (commented):
This looks great! Just a few questions/suggestions.

Wikipedia/Code/WMFSettingsViewController.m
tonisevener: Unfortunately with this code I see a warning that says `@available does not guard availability here; use if (@available) instead`. It does seem to work functionally the same, but I'm wary of putting this in with the warning.

Wikipedia/Code/WMFThemeableNavigationController.m
tonisevener: I see the same warning message here - `@available does not guard availability here; use if (@available) instead`.

Wikipedia/Code/WMFTableHeaderFooterLabelView.m
```diff
@@ -22,9 +22,11 @@ - (void)awakeFromNib {
 - (void)setType:(WMFTableHeaderFooterLabelViewType)type {
     switch (type) {
         case WMFTableHeaderFooterLabelViewType_Header:
-            self.topConstraint.constant = 15;
+            self.topConstraint.constant = [[UIFontMetrics defaultMetrics] scaledValueForValue:30];
```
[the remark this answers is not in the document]
tonisevener: Yeah, once I switched away from the standard `titleForHeaderInSection` methods to get around the theming issue, I noticed the spacing was thrown off in the headers. So I tried to get it closer to how it looked before by doubling this up. I think I've tackled all the spacing issues, but let me know if you see any problems. While I was tweaking this I also noticed `WMFTableHeaderFooterLabelView` didn't scale up with dynamic type at all, so I fixed that. The scaled spacing here is just a part of adding that ability.

Wikipedia/Code/UINavigationBarAppearance+Extensions.swift
tonisevener: Awesome catch, that was an oversight. I added that logic back in where it was before in `WMFThemeableNavigationController`'s `applyTheme` method. I don't really love that lived here before - I think logically it probably belongs more somewhere that isn't theming - but I didn't want to go down a rabbit hole of changing too much at this point.

Wikipedia/Code/SearchSettingsViewController.swift
```diff
@@ -82,8 +85,18 @@ extension SearchSettingsViewController {
 }
 
 extension SearchSettingsViewController {
-    @objc public func tableView(_ tableView: UITableView, titleForFooterInSection section: Int) -> String? {
-        return getSection(at: section).footerTitle
+    @objc func tableView(_ tableView: UITableView, heightForFooterInSection section: Int) -> CGFloat {
+        guard let _ = self.tableView(tableView, viewForFooterInSection: section) as? WMFTableHeaderFooterLabelView else {
```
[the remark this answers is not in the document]
tonisevener: Thanks, I think you make an excellent point, I have updated this. Ultimately the height for footers were based on the existence of footer text in the various section models (and in one screen's case, if it's the last section), so there really isn't any need to try to instantiate the view and/or check if it's nil. The section models aren't all uniform across these screens, so I decided to keep this new logic in the view controllers.

Wikipedia/Code/WMFSettingsViewController.m
mcleinman: Thank you, good catch! (And sorry for not fully examining my suggestion.) We'll keep it as is.

Wikipedia/Code/WMFThemeableNavigationController.m
mcleinman: Thank you, good catch! (And sorry for not fully examining my suggestion.) We'll keep it as is. Great work on this.

mcleinman (approved):
This is looking great to me. And I just spent quite a while playing with this on a touch ID simulator, a Face ID simulator, and an iPad simulator. I couldn't find any issues, but may have missed something. I definitely think we should all continue to hammer on this after this is merged and a nightly build comes out.

That said, we need to resolve the conflict in this branch before merging.

PR #4076, opened 2021-11-18

translatewiki:
Translation updates


mcleinman:
Fixed problems in 3 strings. Next twn transfer should allow tests to pass.

PR #4080, opened 2021-12-06
Fix iPad layout bug fix on search results

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T208892#7550319

### Notes
This fixes a regression from work done in https://github.com/wikimedia/wikipedia-ios/pull/4036. The `cell.configureForCompactList(at: indexPath.item)` call caused the layout margins to reset to the compact measurements when the configuration came through our new `updateArticleCell(_ notification: NSNotification)` method.

I'm not sure what benefit `configureForCompactList` gives us at all, but to keep the risk low I adjusted the methods so that only when configuring the cell through the `updateArticleCell` method, it no longer tries to reset the layout margins.

### Test Steps
1. Run on iPad
2. Search for an article
3. On the search results list, swipe right on a cell and select the save icon.
4. When cell drawer closes, confirm cell layout margins are retained (before margins would shrink down to compact sizes).
5. Perform same steps on iPhone. Confirm cell layout margins are retained.

### DIFF
--- a/Wikipedia/Code/SearchResultsViewController.swift
+++ b/Wikipedia/Code/SearchResultsViewController.swift
@@ -93,6 +93,10 @@ class SearchResultsViewController: ArticleCollectionViewController {
     }
     
     override func configure(cell: ArticleRightAlignedImageCollectionViewCell, forItemAt indexPath: IndexPath, layoutOnly: Bool) {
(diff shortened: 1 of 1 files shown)


staykids (approved):
What a strange one! I spent a bit of time digging in the layout margins, but as I started looking into the superclasses I took a step back. It seems sensible to me addressing this in a localized way like you have here.

Tests as expected so far on iPad and iPhone.

PR #4082, opened 2021-12-08
Use `leadImage` flag to determine lead image for gallery presentation

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T252920

### Notes
This PR addresses the problem where the user taps a lead image in an article, but is taken to a different/inconsistent image in gallery. It checks the [newly added](https://phabricator.wikimedia.org/T269312) `leadImage` property offered by the `media-list` endpoint when presenting a lead image `MediaItem` in the gallery view. While there still may be instances of the incorrect lead image opening in the gallery when the user taps it, those are likely API problems and not client side problems. With this client side fix, I believe we’re doing the best we can on the iOS side to respect what the API offers. Additional API/PCS changes may likely be required to completely resolve the inconsistent gallery/lead image issues.

I remembered this yesterday after reviewing PRs as something that has user facing benefit that's ready, but it’s ok that this will likely not make it into 6.8.2. 

### Test Steps
1. Open https://en.wikipedia.org/wiki/Pago_Pago on `main` branch build
2. Tap the lead image, and confirm the wrong image opens in the gallery
3. Open the same article on this branch
4. Tap the lead image, confirm it opens in the gallery


mazevedofs (approved):
Working perfectly!

PR #4084, opened 2021-12-10
Update `themeAnalyticsName` values to match MobileWikiAppiOSUserHistory schema

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T289604

### Notes
* This PR updates the `themeAnalyticsName` method to send values that match our analytics schema, and to make sure it sends `dark` as a value when the Default theme is selected, but the device is set to dark mode. To ensure that, a validation is made with both the applied theme and the system appearance mode, to decide the correct value to be sent.

### Test Steps
1. Launch the app, select the Default theme
2. Set the device appearance to Light Mode, the value sent to analytics should be default.
3. Set the device appearance to Dark Mode, the value sent to analytics should be dark.
4. Set the theme to Light, the value sent should be Default


Wikipedia/Code/NSUserDefaults+WMFExtensions.swift
```diff
@@ -116,10 +116,17 @@ let WMFSendUsageReports = "WMFSendUsageReports"
     
     @objc var themeAnalyticsName: String {
         let name = string(forKey: WMFAppThemeName)
+        let darkMode =  bool(forKey: "SystemDarkMode")
```
tonisevener, line 119: Can you make a property with this string, just so we're not hardcoding it twice? Maybe a private computed one just for this extension to see.

Wikipedia/Code/NSUserDefaults+WMFExtensions.swift
```diff
@@ -116,10 +116,17 @@ let WMFSendUsageReports = "WMFSendUsageReports"
...
+            let theme = Theme.defaultAnalyticsThemeName
+            return darkMode ? Theme.dark.name : theme
+        }
+        
+        if Theme.withName(name)?.name == Theme.light.name {
             return Theme.defaultAnalyticsThemeName
         }
-        return Theme.withName(name)?.name ?? Theme.light.name
+        
+        return Theme.withName(name)?.name ?? Theme.defaultAnalyticsThemeName
```
tonisevener, line 129: This is working great! I did find one odd instance of the name "dark-dimmed" and "black-dimmed" sending if you have the "dim images" toggle checked in the Reading preferences settings and you change the theme from an article view or the editor theming popovers. This is likely already an existing bug but we should fix it while we're here.

While I was digging around I also noticed an old `analyticsName` property on Theme.swift that appears to be unused. Because of this "-dimmed" bug though, I think it has value in us using it again. If you [update](https://github.com/wikimedia/wikipedia-ios/blob/7fd42e1d1e178d9f732574101dcf6598f937058d/WMF%20Framework/Theme.swift#L839) the light theme's `analyticsName` to be Theme.defaultAnalyticsThemeName, I think this method could then become something like:

```
guard name != nil, name != Theme.defaultThemeName else {
    return systemDarkMode ? Theme.black.analyticsName : Theme.light.analyticsName
}

return Theme.withName(name)?.analyticsName ?? Theme.light.analyticsName
```

This way this way the analytics value can hopefully be a bit more stable than the theme name, which apparently changes with that dimming setting.

Wikipedia/Code/NSUserDefaults+WMFExtensions.swift
```diff
@@ -116,10 +116,17 @@ let WMFSendUsageReports = "WMFSendUsageReports"
     
     @objc var themeAnalyticsName: String {
         let name = string(forKey: WMFAppThemeName)
+        let darkMode =  bool(forKey: "SystemDarkMode")
         guard name != nil, name != Theme.defaultThemeName else {
+            let theme = Theme.defaultAnalyticsThemeName
+            return darkMode ? Theme.dark.name : theme
```
tonisevener, line 122: Actually for the system dark mode, we automatically choose our app's "black" theme, not the app's "dark" theme (confusing, I know). Can you send `Theme.black` here instead of `Theme.dark`? It also might help to rename the local `darkMode` var to `systemDarkMode` to maybe help a little with code confusion.

tonisevener (changes requested):
@mazevedofs Awesome, great job working around this tricky theming logic! I have some small cleanup suggestions and one larger fix for a different (likely already existing) bug that I found.

Wikipedia/Code/NSUserDefaults+WMFExtensions.swift
```diff
@@ -116,17 +116,16 @@ let WMFSendUsageReports = "WMFSendUsageReports"
...
-            return darkMode ? Theme.dark.name : theme
+            return systemDarkMode ? Theme.black.analyticsName : Theme.light.analyticsName
         }
         
         if Theme.withName(name)?.name == Theme.light.name {
             return Theme.defaultAnalyticsThemeName
         }
         
-        return Theme.withName(name)?.name ?? Theme.defaultAnalyticsThemeName
+        return Theme.withName(name)?.name ?? Theme.light.analyticsName
```
tonisevener, line 128: @mazevedofs one tiny thing - the "-dimmed" name is still trickling through, I think this needs to be `Theme.withName(name)?.analyticsName` instead of `Theme.withName(name)?.name`.

tonisevener (changes requested):

Wikipedia/Code/NSUserDefaults+WMFExtensions.swift
mazevedofs: Thanks for catching this one as well! :)

tonisevener (approved):
🎉  Thanks! Looks good!

PR #4086, opened 2021-12-10
Present `More` swipe action popover centered over container view

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T288687 (in part)

### Notes
This updates the presentation of the `More` action popover to centered over the containing action view rather than centered over the containing cell view.

### Test Steps
1. Swipe open a notification cell on iPad
2. Tap `More` and confirm the popover is presented centered from the `More` swipe action view


tonisevener (approved):
Thanks, this is much better!

PR #4090, opened 2021-12-17
Add toolbar buttons to Notifications Center

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T288874 (partial)

### Notes

This PR includes some baseline architecture for the Notifications Center toolbar. It includes:
- Adding toolbar buttons that change when the primary mode of the view controller changes (type and project filter buttons with status text when not editing, Mark and Mark all as read buttons when in editing mode)
- Toni's initial model for `RemoteNotificationsFilterState` to represent filters on the presented notifications
- A `RemoteNotificationsController` notification that fires whenever the filter state is changed
- Some localized strings

There's more that will be continued in a future PR, including:
- Adding the actual logic for determining status bar text (the primary purpose of the Phab task)
- Addressing theming in the bar buttons
- More String localization

An engineering note about the _system_ notification added in `RemoteNotificationsController`: if the relationship between the `RemoteNotificationsController` and the `NotificationsCenterViewModel` was one of direct ownership by `NotificationsCenterViewModel`, it'd make sense to use a single delegate instead of firing a notification. But because `RemoteNotificationsController` is owned by `MWKDataStore`, a single delegate is not ideal and could be confusing for anything else that uses `RemoteNotificationsController`. Instead of implementing a more appropriate delegation pattern here, I chose to just fire a system notification. I dislike that this is less direct in terms of the class's influence, but it serves our purpose in this case.

### Test Steps

For this PR, mostly just code review and confirm toggling the Edit button changes the toolbar items appropriately. In the continuation of this PR, I'll include some test code to simplify seeing the different states of the status bar text.


Wikipedia/Code/NotificationsCenterViewModel.swift
```diff
@@ -62,8 +77,10 @@ final class NotificationsCenterViewModel: NSObject {
     // MARK: - Public
     
     func refreshNotifications() {
+        isLoading = true
```
tonisevener, line 80: I see what you mean about this loading state getting messy because importing and refreshing can step on each other a bit deeper within the remote notifications controllers. I'll give some thought to cleaning that area up.

tonisevener (approved):
Looks good! Thanks for packaging up a partial implementation. I'm going to go ahead and merge, and that should allow us to wrap up https://phabricator.wikimedia.org/T288665.

PR #4092, opened 2021-12-22
Settings swipe

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T217937

### Notes
* This PR adds the swipe to pop gesture on Settings' navigation stack, adding support for the swipe gesture form left to right on some view controllers pushed from this navigation controller.

### Test Steps
1. Launch the app, tap the Settings icon 
2. Tap on some option that is pushed from the controller (Search, Account/Login, Explore Feed and Notifications are eligible - on screens with the X close button on the top right this gesture shouldn't work).
3. Use the swipe from left to right gesture to return to the main Settings page.
4. Try to swipe back on the Settings page, nothing should happen here.
5. With the Settings page open, close the app and quit it on the background. Launch the App again (it should open on the Settings page, and repeat steps 2 - 4.
6. Disable the explore feed (it should be replaced by the Settings page), repeat steps 2 - 5.


tonisevener (changes requested):
This is working well! Just a few comments from me. Also I think the general approach is fine so you can take this out of Draft mode. Thanks!

Wikipedia/Code/NavigationStateController.swift
```diff
@@ -8,6 +9,7 @@ protocol DetailPresentingFromContentGroup {
 final class NavigationStateController: NSObject {
     private let dataStore: MWKDataStore
     private var theme = Theme.standard
+    private var settingsNavController: UINavigationController?
```
tonisevener, line 12: @mazevedofs One last minor thing - add `weak` here -

`private weak var settingsNavController: UINavigationController?`

Unless you disagree of course. Thanks!

mazevedofs: I totally agree, completely missed it from your previous comment

tonisevener (approved):
Looks great, thank you!

PR #4095, opened 2021-12-23

translatewiki:
Translation updates


mcleinman:
Fixed 2 Catalan translations, next import should work fine. (Catalan has a plural-sensitive word for "new", which we unfortunately do not currently support.)

PR #4099, opened 2022-01-05
Update `MobileWikiAppiOSUserHistory` to add `inbox_count` and `device_level_enabled` values

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T297752

### Notes
* This PR updates the `MobileWikiAppiOSUserHistory` schema to be able to collect data from our notifications, We're adding two more properties - `inbox_count` that contains the amount of notifications a user has as an Integer, and `device_level_enabled` which sends the  current `UNAuthorizationStatus` as a String.

### Test Steps
1. Launch the app
2. In Settings, go to Notifications and enable them
3. If possible, do some actions that generate notifications (so we can see the count go up and update accordingly)
4. Check if `MobileWikiAppiOSUserHistory` has the two new properties added, and if they reflect currently the user's inbox count and notifications authorization status.


WMF Framework/Remote Notifications/RemoteNotificationsController.swift
```diff
@@ -61,6 +61,24 @@ import CocoaLumberjackSwift
...
+        
+        guard let viewContext = self.viewContext else {
+            DDLogError("Failure fetching notifications from persistence: missing viewContext")
+            return 0
+        }
+        
+        let fetchRequest: NSFetchRequest<RemoteNotification> = RemoteNotification.fetchRequest()
+        
+        do {
+            return try viewContext.fetch(fetchRequest).count
```
tonisevener, line 76: Try `return try viewContext.count(for: fetchRequest)` instead - I think it's more efficient under the hood than first fetching an array of notifications and getting the count.

Wikipedia/Code/UserHistoryFunnel.swift
```diff
@@ -53,9 +53,35 @@ private typealias ContentGroupKindAndLoggingCode = (kind: WMFContentGroupKind, l
             event["test_group"] = articleAsLivingDocBucket.rawValue
         }
         
+        var autorizationStatus: UNAuthorizationStatus = .notDetermined
+        dataStore.notificationsController.notificationPermissionsStatus { status in
```
tonisevener, line 57: I noticed the event always seems to post `notDetermined`, even if I have enabled notifications permissions (which you can enable via app Settings > Notifications > flip the toggle).

`dataStore.notificationsControllernotificationPermissionStatus` is asynchronous so it's probably not returning in time to update the `authorizationStatus` var before it gets added to the event dictionary. 

Instead I'd first try determining the status outside of the `event` method, then passing it in as a parameter once we definitely know we have it. Here's an example of what the `logStartingSnapshot()` event call would look like:

```
dataStore.notificationsController.notificationPermissionsStatus {[weak self] authorizationStatus in
            
    guard let self = self else {
        return
    }
            
    DispatchQueue.main.async {
        self.log(self.event(authorizationStatus: authorizationStatus))
    }
}
```

tonisevener (changes requested):
Nice! Just a couple of comments, one little performance suggestion and the other on a bug I was seeing.

tonisevener (approved):

tonisevener:
Looks great! Thanks!

PR #4105, opened 2022-01-14
Notifications prototype work

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T288665
https://phabricator.wikimedia.org/T288669
https://phabricator.wikimedia.org/T288870
https://phabricator.wikimedia.org/T288668
https://phabricator.wikimedia.org/T288873
https://phabricator.wikimedia.org/T288694
https://phabricator.wikimedia.org/T297746
https://phabricator.wikimedia.org/T297747

### Notes
This PR syncs up our `main` branch with our prototype work. Feel free to commit any lighter-touch cleanup & bug fixes. For any larger missing pieces for the Phab tasks above or larger refactoring efforts, please make a note of it and we will track it with another followup PR.

### Test Steps
Test Notifications Center:

- Notifications importing, refreshing and displaying
- Notifications marking as read / unread
- Notifications read status & type filtering
- Notifications inbox filtering


WMF Framework/Remote Notifications/Model/RemoteNotificationsModelController.swift
```diff
@@ -102,22 +167,27 @@ final class RemoteNotificationsModelController: NSObject {
...
+                notification.readString != nil
+            })
+                
+            if newPageContainsUnread {
+                NotificationCenter.default.post(name: Notification.Name.NotificationsCenterBadgeNeedsUpdate, object: nil)
+            }
+            
             self.save(moc: moc)
             completion()
         }
```
staykids, line 193: @tonisevener I was noticing some strange behavior around the in-app explore feed bell not properly updating to its indicated state (bell with dot indicator) when a user initially logs in, has the initial import happen, and has unread notifications. The explore feed properly updates to show the bell, but it's in the wrong (un-indicated) state. 

I think this suggestion can address that issue, but I wanted to check with you before I commit – do you see any drawbacks with it I'm missing?

```suggestion
    public func createNewNotifications(moc: NSManagedObjectContext, notificationsFetchedFromTheServer: Set<RemoteNotificationsAPIController.NotificationsResult.Notification>, completion: @escaping () -> Void) throws {
        moc.perform {
            for notification in notificationsFetchedFromTheServer {
                self.createNewNotification(moc: moc, notification: notification)
            }
            
            self.save(moc: moc)
            NotificationCenter.default.post(name: Notification.Name.NotificationsCenterBadgeNeedsUpdate, object: nil)
            completion()
        }
    }
```

tonisevener: @staykids ah, good catch! Your suggestion looks good. I think I was just trying to optimize and limit the `NSNotification` from firing needlessly (and thus an unnecessary count fetch from the database and view update), but now that I think about it, we do want it to update whether or not it contains an unread notification. I could see a situation where it's in a badged state, but a refresh call updates all notifications in the database to be read (if say they marked them from Desktop). We'd need this to fire and un-badge the bell icon.

PR #4111, opened 2022-01-27
Notifications data layer cleanup

tonisevener:
**Phabricator:** 
N/A

### Notes
This consists of general cleanup and refactoring on the data layer for notifications. The main files I aimed to clean up are `RemoteNotificationsController`, `RemoteNotificationsOperationsController`, `RemoteNotificationsModelController`, and `RemoteNotificationsAPIController`. It may be easiest to pull down and review in Xcode because the diffs can be large (some due to refactoring, but also due to basic reordering of methods).

Note each commit had a particular file in mind for cleanup (see git commit description for details), but changes trickled out to other files to get it compiling.

### Test Steps
General regression testing across the Notifications Center features.


Wikipedia/Code/NotificationsCenterViewModel.swift
```diff
@@ -89,47 +98,80 @@ final class NotificationsCenterViewModel: NSObject {
             delegate?.update(types: updateTypes)
         }
     }
+    
+    @objc private func remoteNotificationsControllerDidUpdateFilterState() {
+        //Not doing anything here yet
+        //Because filter screen disapperances call self.resetAndRefreshData from the view controller, which fetches the first page again and relays state changes back to the view controller, there's no need to react to this notification.
```
staykids, line 104: RE: the interim load state flashing you mentioned – it seems like sometimes cases of opening the center > opening the filter settings > switching the read status from all to read and > dismissing the filters modal results in a lil' empty state loading flash.

I think if we forced the main center to reload here, while the user is likely still in the filters modal, we could prevent that. But I don't think I'm suggesting we actually do that, as it's more just masking that problem rather than directly addressing it. Probably better we just address that separately outside of this PR – what do you think?

WMF Framework/Remote Notifications/RemoteNotificationsOperationsController.swift
```diff
@@ -1,349 +1,250 @@
...
-                self?.isImporting = false
-                self?.importingCompletionBlocks.forEach { completionBlock in
-                    completionBlock(error)
+                self?.isLoadingNotifications = false
+                self?.loadingNotificationsCompletionBlocks.forEach { completionBlock in
+                    completionBlock(result)
                 }
 
-                self?.importingCompletionBlocks.removeAll()
+                self?.loadingNotificationsCompletionBlocks.removeAll()
```
staykids, line 64: In practice the main thread assertions and serial queue operations may result in it all being fine, but this array mutation gives me a bit of thread safety anxiety.

WMF Framework/Remote Notifications/Model/RemoteNotificationsModelController.swift
```diff
@@ -153,177 +141,162 @@ final class RemoteNotificationsModelController: NSObject {
...
+        assert(Thread.isMainThread)
         let fetchRequest: NSFetchRequest<RemoteNotification> = RemoteNotification.fetchRequest()
         fetchRequest.predicate = unreadNotificationsPredicate
-        return try? viewContext.count(for: fetchRequest)
+        return try viewContext.count(for: fetchRequest)
+    }
+    
+    func numberOfAllNotifications() throws -> Int {
+        assert(Thread.isMainThread)
+        let fetchRequest: NSFetchRequest<RemoteNotification> = RemoteNotification.fetchRequest()
```
staykids, line 164: I think here (and elsewhere within `RemoteNotificationsModelController` and `RemoteNotificationsRefreshOperation`) we can remove the `: NSFetchRequest<RemoteNotification>` and let type inference take care of us.

staykids (commented):
Nice! Thanks for taking the time to document some of the computed variables/functions. I still want to do another pass and more on device testing, but below are some things I noticed on my first pass.

WMF Framework/Remote Notifications/Model/RemoteNotificationsModelController.swift
```diff
@@ -356,4 +329,83 @@ final class RemoteNotificationsModelController: NSObject {
...
+    }
+    
+    private func createNewNotification(moc: NSManagedObjectContext, notification: RemoteNotificationsAPIController.NotificationsResult.Notification) throws {
+        guard let date = notification.date else {
+            assertionFailure("Notification should have a date")
+            throw ReadWriteError.missingDateInNotification
+        }
+
+        let isRead = notification.readString == nil ? NSNumber(booleanLiteral: false) : NSNumber(booleanLiteral: true)
+        let _ = moc.wmf_create(entityNamed: "RemoteNotification",
```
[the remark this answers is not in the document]
tonisevener: Done!

Wikipedia/Code/NotificationsCenterViewModel.swift
tonisevener: Yeah, I had a similar set up like that at some point, that is, updating the collection view whenever the filter state changed at all. The problem was iPad - because the modal is presented as `.pageSheet`, you could see the collection view underneath showing empty states and animating inserts/deletes with every little toggle the user did on the filter/inbox views. I found it distracting, so I updated it so that the collection view updates only when those modals are about to disappear. 

But that is a good bug you found and exactly what I was talking about, thanks for finding an instance of it. I'm not sure how to tackle it at this point, but I'll add it to my bug list and make a Phab task on it so that we can address it separately.

WMF Framework/Remote Notifications/RemoteNotificationsOperationsController.swift
tonisevener: Yeah, that's fair. I do think it's safe but it would be nice to avoid keeping track of these completion blocks at all. The only reason I'm tracking them this way is if importing/refreshing takes a long time - I want to be able to react when it completes if it's already in the middle of it.

What do you think about removing the completion block param from this method entirely, and we have callers react to it via the `NotificationsCenterLoadingDidEnd` notification that posts? Originally I thought of `NotificationsCenterLoadingDidEnd` as only being used for a basic UI loading state, but I can explore using it for whatever the completion blocks are doing.

staykids: Interesting! So essentially the idea is `RemoteNotificationsController` becomes an observer for the `LoadingDidEnd` notifications, it calls the `RemoteNotificationsOperationsController`'s `loadNotifications`, then upon receiving that notification inspects something (maybe `userInfo` on the notification) to know if it was a success or failure to then perform whatever actions it wants?

If your intent for those notifications was more to serve the UI though, maybe it's best not to overload them and stick with the blocks. This would help prevent its specific intent leaking into and influencing other parts of the app. I personally don't have a problem with the completion block approach – I think it actually nicely draws a pretty clear path of intent for the reader, which is helpful when tracing and debugging.

tonisevener: Yep! That's what I was thinking, but if the intent seems clear to you as it is then maybe I'll hold off until we're sure it'll be a problem.

WMF Framework/Remote Notifications/RemoteNotificationsController.swift
```diff
@@ -1,180 +1,334 @@
...
-        }
+    @objc private func applicationDidBecomeActive() {
+        loadNotifications(force: false)
     }
     
-    @objc func deleteLegacyDatabaseFiles() {
+    @objc private func authManagerDidLogOut() {
         do {
-            try operationsController.deleteLegacyDatabaseFiles()
+            try modelController?.resetDatabaseAndSharedCache()
```
staykids, line 100: This is venturing into totally fine if you'd rather we address it in a separate PR territory, but I think to be good data citizens we should also reset the `filterState` and `allInboxProjects` in `RemoteNotificationsController` when the user logs out. If we don't, the previous `filterState` is retained, and if they decide to log back in their first Notifications Center experience is in an unexpected state. It might be a little tricky because of `filterState`'s `didSet` property observer – it may be a little wasteful and silly and not proper, but maybe even this little bit in the listening for log out method is sufficient:

```swift 
@objc private func authManagerDidLogOut() {
    do {
	filterState = RemoteNotificationsFilterState(readStatus: .all, offTypes: [], offProjects: [])
        allInboxProjects = []
        try modelController?.resetDatabaseAndSharedCache()
    }
	...
}
```

tonisevener: Great catch! I agree this is needed. I think your approach works well - it crossed my mind that maybe we should allow `filterState` to be nullable but then we'll have to handle the optional everywhere. We may be persisting some unnecessary data but I'm okay with that tradeoff at this point. Plus I'm pretty sure we already are persisting this default elsewhere, if they reset everything on the filters & inbox views.

staykids (approved):
👍🏽 nice – I pushed a tiny commit to resolve a build error (after merging in `main`) as a result of the switch to throwable.

PR #4114, opened 2022-02-01
Fix "Find in Page" bug introduced by iOS 15 accessibility

mcleinman:
**Phabricator:** https://phabricator.wikimedia.org/T292471

### Notes
* Confirmed that this bug is an Apple-side one. For some reason, if `Full Keyboard Access` in enabled within Accessibility settings, the IBAction `Editing Changed` for a UITextField is not called. We're using a simple workaround of manually setting an `editingChanged` target in code, and this is not affected by the bug. (The bug is not present in iOS 14.)
* Sample code that I'll be submitting to Apple: [SampleInputView.zip](https://github.com/wikimedia/wikipedia-ios/files/7975464/SampleInputView.zip)
* Further testing seems to show that this bug only exists for `UITextField`s within `inputAccessoryView`, so this is the only place this bug reveals itself in our app. Other `inputAccessoryView`s and other `UITextField`s do not have this issue.

### Test Steps
1. Turn on `Full Keyboard Access` accessibility setting on a device. (It isn't available on Simulator.)
2. Run app on a device. 
3. Ensure Find in Page works.
4. Ensure find/replace in editing works.

### DIFF
--- a/Wikipedia/Code/FindAndReplaceKeyboardBar.swift
+++ b/Wikipedia/Code/FindAndReplaceKeyboardBar.swift
@@ -116,6 +116,10 @@ final class FindAndReplaceKeyboardBar: UIInputView {
         updateShowingReplaceState()
         updateReplaceLabelState()
         updateReplaceButtonsState()
+
+        // See comment on findTextFieldTextChanged for why we use these.
+        findTextField.addTarget(self, action: #selector(findTextFieldTextChanged), for: .editingChanged)
+        replaceTextField.addTarget(self, action: #selector(replaceTextFieldTextChanged), for: .editingChanged)
     }
     
     override var intrinsicContentSize: CGSize {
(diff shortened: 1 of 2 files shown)


tonisevener (commented):
@mcleinman Awesome, thanks for digging into this one. This is definitely better now than not working at all, but I wanted to point out some strangeness I'm still seeing. It seems like the other input accessory view buttons do not react as long as the device keyboard is open. I wonder if it could be a similar issue that requires us to delete the `IBAction`s for those.

1. Go to an article view
2. Tap the Find button and type in a search term.
3. Notice the arrow up and arrow down buttons enable whenever results highlight, however they don't do anything when you tap them. The close button also has the same issue on, we can not exit Find mode.
4. Now tap "search" on the device keyboard OR scroll a little bit in the article to minimize the device keyboard.
5. Now the arrow up and down buttons work, as well as the X button.

There are similar issues in the Editing flow.

mcleinman:
> @mcleinman Awesome, thanks for digging into this one. This is definitely better now than not working at all, but I wanted to point out some strangeness I'm still seeing. It seems like the other input accessory view buttons do not react as long as the device keyboard is open. I wonder if it could be a similar issue that requires us to delete the `IBAction`s for those.
> 
>     1. Go to an article view
> 
>     2. Tap the Find button and type in a search term.
> 
>     3. Notice the arrow up and arrow down buttons enable whenever results highlight, however they don't do anything when you tap them. The close button also has the same issue on, we can not exit Find mode.
> 
>     4. Now tap "search" on the device keyboard OR scroll a little bit in the article to minimize the device keyboard.
> 
>     5. Now the arrow up and down buttons work, as well as the X button.
> 
> 
> There are similar issues in the Editing flow.

I did an audit to see if this was affecting buttons other input accessory views, and it was not. So I dug into it some more, and... This behavior was caused because the xib connects to the first responder, not the file's owner. Switching the connection to file's owner fixes the problem. (Still, since this is new behavior in iOS 15 I think this could be described as a regression.) 

I'm also backing out the change to the text fields, and properly connecting that piece as well.

mcleinman:
Closing this in favor of https://github.com/wikimedia/wikipedia-ios/pull/4120

PR #4118, opened 2022-02-07
Update NotificationServiceExtension deployment target to 13.0

tonisevener:
**Phabricator:** None

### Notes
I noticed the Notification Service Extension target had a deployment target set to 14.5, which I found odd. Updating to 13.0 to match everything else.

I don't have an iOS13 test device so I can't be sure if there was a bug with push notifications that this fixes.


staykids (approved):
Great catch! I also do not have a physical device still on iOS 13. I'll pull https://phabricator.wikimedia.org/T287310 back into Needs QA and specifically call this out.

PR #4122, opened 2022-02-10
T287970 mark as seen

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T287970

### Notes
* This PR adds the capability to mark notifications as seen from the app, replicating them to the web. As the user enters the notifications view, a call will be made to the notifications API and mark all notifications as seen unless a notifications inbox filter is applied.

### Test Steps
1. Be sure to have unseen notifications on a project. You can check it by going to the web and making sure the badge on the bell icon is red
2. On the iOS App, click on the notifications icon on the upper left corner of the Explore feed. Do not click on the notifications.
3. Go to the desktop version, and check if the badge is now grey.

### Screenshots/Videos
- Badge with unseen notifications
Screen Shot 2022-02-10 at 18 16 13

- Badge with seen and unread notifications
Screen Shot 2022-02-10 at 18 16 45


WMF Framework/Remote Notifications/RemoteNotificationsAPIController.swift
```diff
@@ -385,6 +443,15 @@ public class RemoteNotificationsAPIController: Fetcher {
                               "format": "json"]
             return dictionary
         }
+        
+        static func markAllAsSeen(project: RemoteNotificationsProject) -> Parameters? {
+            let dictionary = ["action": "echomarkseen",
+                              "all": "true",
```
tonisevener, line 449: I think you can remove this `"all": "true"` line.

WMF Framework/Remote Notifications/RemoteNotificationsController.swift
```diff
@@ -191,6 +196,26 @@ public enum RemoteNotificationsControllerError: Error {
         operationsController.markAllAsRead(languageLinkController: languageLinkController, completion: completion)
     }
     
+    /// Asks server to mark all notifications as seen for wikipedias that contain local unread notifications.
+    public func markAllAsSeen(completion: @escaping ((Result<Void, Error>) -> Void)) {
+        guard !areFiltersEnabled else {
+            return
```
tonisevener, line 202: I think it might be good practice to call the completion block here too, just for the sake of consistency. You could add a new case for it to `RemoteNotificationsControllerError` and pass that through like the guards below.

I'm a little torn because it's not _really_ an error, and not really a success either, but I think I lean more towards passing back an error in this case.

Wikipedia/Code/NotificationsCenterViewModel.swift
```diff
@@ -162,6 +162,18 @@ final class NotificationsCenterViewModel: NSObject {
         }
     }
     
+    func markAllAsSeen() {
+        remoteNotificationsController.markAllAsSeen { result in
+            switch result {
+            case let .failure(error):
+                DDLogError("Error marking all notifications as seen: \(error)")
```
tonisevener, line 169: You might want to skip this logging if the error is the new filters-enabled error you'll make, since it'll be a common path that's not actually a problem to need to know about.

tonisevener (changes requested):
@mazevedofs Great work! Just a couple of minor things from code review. I'll let you know if I find anything else after pulling and testing.

tonisevener (changes requested):
@mazevedofs thanks! The code looks good. I pulled it down and tested it this morning. I'd like us to do a few more tweaks here to make the mark as seen call a little more permissive. I think if the user is seeing any data, we go ahead and call mark as seen, even if they have filters on. But if they are seeing an empty state (which can happen due to filters or during an initial slow import), we prevent the call.

I took a crack at this in a `T287970-mark-as-seen-suggestions` branch [here](https://github.com/wikimedia/wikipedia-ios/commit/bc20ba6cf1affd285abdab2d4c8610edd89b7476). I ended up having to move more of the validation logic up a layer into `NotificationsCenterViewModel` since that's where we can check to see what's displaying screen via the model controller. Let me know what you think!

mazevedofs:
> @mazevedofs thanks! The code looks good. I pulled it down and tested it this morning. I'd like us to do a few more tweaks here to make the mark as seen call a little more permissive. I think if the user is seeing any data, we go ahead and call mark as seen, even if they have filters on. But if they are seeing an empty state (which can happen due to filters or during an initial slow import), we prevent the call.
> 
> I took a crack at this in a `T287970-mark-as-seen-suggestions` branch [here](https://github.com/wikimedia/wikipedia-ios/commit/bc20ba6cf1affd285abdab2d4c8610edd89b7476). I ended up having to move more of the validation logic up a layer into `NotificationsCenterViewModel` since that's where we can check to see what's displaying screen via the model controller. Let me know what you think!

@tonisevener I think it works best this way! Great catch with the verification for the empty states. I cherry-picked your commit onto this branch so we keep the git history

tonisevener (approved):

PR #4124, opened 2022-02-15
Notifications performance fix - cell heights

tonisevener:
**Phabricator:**
https://phabricator.wikimedia.org/T301837 (Part 1 of 3)

### Notes
When performance testing, I noticed the scroll view would start to lag when I reached around 3,000 notifications. I was testing on an iPhone 7, iOS 15.2.1. Leaning on any previously calculated heights in-memory seemed to fix it.

**Note:** The `notifications-performance-testing` branch is based off of `main` as it is now, plus one small loading tweak that's coming in a future PR. It is also set up to import around 10,000 mock notifications if you add 7+ app languages. Try that and scroll until 3,000 notifications have loaded in Notifications Center. It might be helpful to print out the index path in `willDisplayCell` on `NotifiationsCenterViewController` so you can see how many have loaded. You can also try tweaking the default `fetchLimit` parameter in `RemoteNotificationsController` to immediately fetch a lot of notifications without having to page.

### Test Steps
0. Pull the `notifications-performance-testing` branch down and cherry-pick this commit into it (do not push up this change though).
1. Open Notifications Center and scroll until 3,000 notifications have loaded. Confirm it still scrolls smoothly.
2. Background the app and change the user's dynamic type setting. Go back to Notifications Center. Confirm the cells resize properly to the new font size and it still scrolls smoothly.

### DIFF
--- a/Wikipedia/Code/NotificationsCenterView.swift
+++ b/Wikipedia/Code/NotificationsCenterView.swift
@@ -13,22 +13,11 @@ final class NotificationsCenterView: SetupView {
     // MARK: - Properties
 
 	lazy var collectionView: UICollectionView = {
-		let collectionView = UICollectionView(frame: .zero, collectionViewLayout: tableStyleLayout)
+		let collectionView = UICollectionView(frame: .zero, collectionViewLayout: tableStyleLayout())
 		collectionView.register(NotificationsCenterCell.self, forCellWithReuseIdentifier: NotificationsCenterCell.reuseIdentifier)
 		collectionView.alwaysBounceVertical = true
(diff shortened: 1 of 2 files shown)


staykids (approved):
👍🏽 I've scrolled 3-5k notifications a few times on my test iPhone 11 before and after applying this fix, but personally haven't encountered devastating scroll performance before or after. That said, it does have a perceivably more "buttery" scroll feel with this pre-calculated absolute height dimension approach. Another benefit is that we no longer run into that one little visual glitch where marking as read/unread would change the cell height slightly (to accommodate the updated header text font).

PR #4125, opened 2022-02-15
Notifications performance fix - improve edit mode lag

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T301837 (Part 2 of 3)

### Notes
When performance testing I noticed there was a big lag in updating cell states after switching the edit mode. I saw this when I scrolled to about 2,000 notifications.

### Test Steps
1. Pull the notifications-performance-testing branch down and cherry-pick this commit into it (do not push up this change though).
2. Open Notifications Center and scroll until 2,000 notifications have loaded. Tap the Edit button in the upper right corner. Confirm the state of the cells quickly switch to the edit state (with empty circles instead of type icons).
3. Tap the Edit button in the upper right corner again. Confirm the state of the cells quickly switch out of the edit state.


mazevedofs (approved):
Working great

PR #4127, opened 2022-02-15
Notifications filter bug fixes

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T301838

### Notes
I found a few bugs related to notifications displaying at the wrong time while performance testing. 

### Test Steps
0. Pull the notifications-performance-testing branch down and cherry-pick these commits into it (do not push up this change though).

[eac59a7](https://github.com/wikimedia/wikipedia-ios/commit/eac59a789a8b2d6fd1e6b7fe1690d72ac698e1cb)
This commit fixed a small issue with the NSPredicates for translation types - some translation types were slipping through even when it was toggled off on the filters screen.

1. Go to notifications center. Wait until all 10,000 notifications are imported and you see data on screen. Go to filters screen. Toggle off the translation type. Confirm translation types are no longer showing in the list.

[377ba3](https://github.com/wikimedia/wikipedia-ios/commit/377ba311998392768bc4ef66b06b54b356fa8ccb) (Part 1)
Part of this commit fixes a bug where notifications would insert on screen upon refresh even if the filter types had that type toggled off. 

2. Go to notifications center. Wait until all 10,000 notifications are imported and you see data on screen. Tap filters button. Toggle off a notable type (maybe all of the orange ones). Dismiss filters screen. Notifications should refresh and insert when Notifications Center appears again. If not, try backgrounding and foregrounding until you see "Checking for notifications" in the bottom toolbar. Confirm none of the orange ones bypass the types filter and insert during this refresh.

[377ba3](https://github.com/wikimedia/wikipedia-ios/commit/377ba311998392768bc4ef66b06b54b356fa8ccb) (Part 2)
The other part of this commit is another insert bug when refreshing - sometimes upon refresh the API may return a very old unread notification (say the user marked it as unread on Desktop). If the user has only the first page loaded, this old notification would get inserted to the very bottom of the first page. Then when the user scrolled, the next page would get fetched from Core Data and inserted in the sorted order, _above_ this very old notification. This fix does not insert any notifications older than the last one displaying upon refresh. It should get fetched and displayed eventually during our normal paging.

3. Go to notifications center. Wait until all 10,000 notifications are imported and you see the first page of data on screen. Wait a while, then background and foreground until you see Notifications Center switch to a refreshing state ("Checking for notifications" in the bottom toolbar). Then slowly scroll to the bottom of the first page. Confirm the last notification on the first page isn't a super old notification, and the next page of notifications insert at the very bottom.


mazevedofs (approved):
Working perfectly!

PR #4128, opened 2022-02-15
Fix crash from diffable data source update queue

tonisevener:
**Phabricator:**
Maybe https://phabricator.wikimedia.org/T301132?

### Notes
I came across this crash once or twice when doing performance tests. Here's a screenshot of what I saw:

Screen Shot 2022-02-14 at 9 49 34 AM

Here's what I _think_ is happening: we're trying to access a managed object's property (in this case, the `RemoteNotification` `key` property) from the wrong thread. The notifications hanging off of `NotificationsCenterCellViewModel` were created from the `viewContext`, but diffable data source is considering them from the `snapshotUpdateQueue`, not the main thread. My approach here is to just have a constant in-memory `key` property hanging off of `NotificationsCenterCellViewModel` for the `snapshotUpdateQueue` to reference when using `Equatable`/`Hashable`.

This crash happens rarely for me, so it's hard to test beyond basic regression. Even when I've confirmed that the `==` is running on the `shapshotUpdateQueue`, it usually runs fine.

### Test Steps
Regression test importing and refreshing notifications. You can consider cherry-picking this fix into the `notifications-performance-testing` branch to easily increase the number of mock notifications that are imported or refreshed.


mazevedofs (approved):
Looks great!

PR #4129, opened 2022-02-16
Notifications loading state bug fixes

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T301125

### Notes
These are some bug fixes around the notifications loading states. I also removed some unnecessary `print` calls that slipped through.

### Test Steps
1. Fresh install the app. Add a lot of languages, and quickly go into the Notifications Center when the bell icon appears on the Explore feed.
2. Confirm you see an importing state (empty view with "Checking for notifications..." text, bottom toolbar with "Checking for notifications..." text). Now confirm the new logic - that the bottom filter and inbox buttons are disabled during this importing state.

3. Wait on this screen until notifications are imported. You should see Notifications Center populate with notifications. Back out of the screen and revisit Notifications Center to trigger a refresh call. The screen should still display imported notifications, with only the bottom toolbar indicating that the notifications are refreshing (bottom toolbar will say "Checking for notifications..."). Before this PR, the list would switch to an empty state while it was refreshing.

4. Scroll down for a while in Notifications Center. Open the Filters modal. Close the filters modal without making changes. Notifications Center scroll position should remain in place without any reloading. Before this PR, the user would be reloaded and scrolled to the top.
5. Repeat step 4, only this time make a change, then revert the change before dismissing the modal. Confirm the list still doesn't reload.
6. Repeat step 4, only this time make a change, then dismiss the modal. Confirm list reloads, and list is scrolled to the top.

7. Repeat steps 4-6 for the Inbox modal.


mazevedofs (approved):
Works great!

PR #4130, opened 2022-02-17

translatewiki:
Translation updates


mcleinman (approved):
Major update for [Uzbek](https://en.wikipedia.org/wiki/Uzbek_language) - now has 3x as many strings.

PR #4144, opened 2022-03-07

translatewiki:
Translation updates


mcleinman:
Two tests failed. I've updated in TranslateWiki, the next transfer should work. 

Side note: Huge expansion of Telugu translation in this set. Really awesome to see.

PR #4145, opened 2022-03-07
Add per theme Notifications Center and Settings bar button images

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T302454

### Notes
Updates the Explore and Settings bar button items to use new blue colored per theme images. This also moves the Explore View Controller's bar button construction and updating out of App View Controller and into the Explore controller itself.

### Test Steps
1. With the explore feed both on and off, switch the app's theme and confirm the icons update and render as expected.
2. Mark all notifications as read/unread, and confirm the notifications bell button switches between the indicated and default (no red indicator) variants.


tonisevener (approved):
Looks great! Nice to see the Explore button logic in a place more consistent with Settings too. I pushed one small commit removing a couple of methods that we were no longer using, as far as I could tell.

PR #4146, opened 2022-03-08
Fix checkEchoV1 content in notification service extension

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T302295

### Notes
We received a complaint from a diary study user of a few instances of `checkEchoV1` content in their push notification. I found a case where we are not properly falling back to the generic string if we receive an error from the API call. My hope is this is the source of the bug.

### Test Steps
1. I had difficulty reproducing this without forcing an error via code. To force this error I replaced the entire guard at [this line](https://github.com/wikimedia/wikipedia-ios/blob/f20f474d1a30ba933f414ce31b9cc939d6e7c740/WMF%20Framework/Remote%20Notifications/RemoteNotificationsAPIController.swift#L226) with something like:

```
completion([], RequestError.http(500))
return
```
2. From `main`, fresh install on a test device and enable push notifications via Settings. Use our push notification utility tool to push a notification to your test device. You can reach out to me  for more details here if needed.
3. You should see a push notification appear with the content "checkEchoV1".
4. Now repeat the steps in this PR branch. At step 3 you should see a push notification appear with the content "New activity on Wikipedia".


mazevedofs (approved):
Looks great!

PR #4148, opened 2022-03-09
Add `UserDefaults` bool to indicate if user has onboarded to Notifications Center feature

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T302633

### Notes
This boolean should only be set by the Notifications Center controller itself after the user has completed onboarding to the feature (that is, user has both seen the first run modal and has made an explicit decision about turning on Push in the Notifications Center itself).


mcleinman (approved):
🚀

PR #4149, opened 2022-03-10
Replace `willLogOutNotification` with `authenticationManagerWillLogOut` delegate method

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T288688 (in part)

### Notes
This PR addresses issues when unsubscribing the user's device from echo subscriptions when logging out. The previous `willLogOutNotification` approach was not sufficient to always unsubscribe the user from Echo. The echo unsubscription call requires an authenticated account (with a valid CSRF token), and the old notification approach would sometimes fail trying to unsubscribe them after their device was no longer authenticated, leaving the user as subscribed to echo as far as the app was concerned. This became evident when following the test steps below, but seeing during step 3 that the Push switch was still enabled on a fresh log in, which should not happen.

This approach allows the Authentication Manager's delegate to perform any authenticated actions (like unsubscribing the device from Echo notifications) before it performs log out related actions.

We should wait to merge it until after https://github.com/wikimedia/wikipedia-ios/pull/4138 is merged.

### Test Steps
1. Log in to an account on device
2. Enroll in Push notifications (Settings > Push notifications)
3. Log out, then log back in and confirm the Push notifications switch (Settings > Push notifications) is off


tonisevener:
@staykids can you update this PR for merging into `main` when you get a chance?

staykids:
@tonisevener Updated!

staykids:
Closing in favor of https://github.com/wikimedia/wikipedia-ios/pull/4174.

PR #4151, opened 2022-03-10
Add accessibility label, custom actions to notifications center cell

mazevedofs:
**Phabricator:**  https://phabricator.wikimedia.org/T296779 and partially addresses https://phabricator.wikimedia.org/T288776 and 

### Notes
* The intention is to make each cell an accessible element, providing context through an accessibility label and making the swipe actions available to VoiceOver users. 
* Note: the idea is for the VoiceOver behavior to be similar to the Apple Mail app.

### Test Steps
1. With VoicerOver on, go to the notifications center
2. You should be able to navigate through the items on the screen by swiping right to left.
3. When selecting a notification, verify the content read. It should be:
- the type of the notification
- its project
- title
- read/unread status 
- Actions available
- Example: `"Talk page message notification. English Wikipedia. Topic title. Read. Actions available"`
4. Tapping twice on a selected notification should take it to it
5. Swiping up and down on the notification should present the actions available: 
- Mark as Read or Unread, 
- More menu
-  Open notification.

6. Tapping twice on the screen should execute the selected action.
Marking items as unread makes a subtle sound as it changes the screen.
7. Then, Select the Edit button.
8. Swipe left to right to navigate through the notifications, and tap twice to select.
9. When selecting an item, you should hear "Selected"
10. The default VoiceOver behavior announces only the selected items
11. Navigate to the bottom bar buttons and make sure you can read and activate them.
12. Make sure you can access the Mark button and its sub-menu
13. Make sure you can, when available, mark all as read


Wikipedia/Code/NotificationsCenterCellViewModel.swift
```diff
@@ -47,6 +47,24 @@ final class NotificationsCenterCellViewModel {
         
     }
     
+    func accessibilityText() -> NSAttributedString? {
```
mcleinman, line 50: This is super minor, but is there any reason this should be a function instead of a computed var? 
```suggestion
    var accessibilityText: NSAttributedString? {
```

Wikipedia/Code/NotificationsCenterCellViewModel.swift
```diff
@@ -47,6 +47,24 @@ final class NotificationsCenterCellViewModel {
...
+        let readAccessibilityText = WMFLocalizedString("notifications-center-cell-read-accessibility-label", value: "Read", comment: "Accessibility text for indicating that a notification's contents have been read.")
+        let unreadAccessibilityText = WMFLocalizedString("notifications-center-cell-unread-accessibility-label", value: "Unread", comment: "Accessibility text for indicating that a notification's contents have not been read.")
+        let readPronounciationAccessibilityAttribute = WMFLocalizedString("notifications-center-cell-read-ipa-accessibility-attribute", value: "rɛd", comment: "Accessibility ipa pronounciation for indicating that a notification's contents have been read.")
```
mcleinman, line 53: I'm curious, why do we need the phonetic for these two? I'm concerned it won't be translated well.

Wikipedia/Code/NotificationsCenterCell.swift
```diff
@@ -466,6 +466,19 @@ final class NotificationsCenterCell: UICollectionViewCell {
...
+    fileprivate func setupAccessibility(_ viewModel: NotificationsCenterCellViewModel) {
+        accessibilityAttributedLabel = viewModel.accessibilityText()
+        isAccessibilityElement = true
+        
+        let moreActionAccessibilityActionLabel = viewModel.isRead ? CommonStrings.notificationsCenterMarkAsUnread : CommonStrings.notificationsCenterMarkAsRead
+        let moreAction = UIAccessibilityCustomAction(name: "More", target: self, selector: #selector(tappedMoreAction))
+        let markasReadorUnreadAction = UIAccessibilityCustomAction(name: moreActionAccessibilityActionLabel, target: self, selector: #selector(tappedReadUnreadAction))
+        let isEditing = viewModel.displayState.isEditing
+        
+        accessibilityCustomActions = isEditing ? nil : [moreAction, markasReadorUnreadAction]
```
mcleinman, line 478: Slightly more performant to not set up those labels if we don't need them. (I think we still need to set as `nil` so we don't get outdated custom actions due to cell reuse, but I'm not 100% sure.)

```suggestion
        if !viewModel.displayState.isEditing {
                let moreActionAccessibilityActionLabel = viewModel.isRead ? CommonStrings.notificationsCenterMarkAsUnread : CommonStrings.notificationsCenterMarkAsRead
                let moreAction = UIAccessibilityCustomAction(name: "More", target: self, selector: #selector(tappedMoreAction))
                let markasReadorUnreadAction = UIAccessibilityCustomAction(name: moreActionAccessibilityActionLabel, target: self, selector: #selector(tappedReadUnreadAction))
        
                accessibilityCustomActions = [moreAction, markasReadorUnreadAction]
        } else {
                accessibilityCustomActions =  nil
        }
```

Wikipedia/Code/NotificationsCenterCell.swift
```diff
@@ -466,6 +466,19 @@ final class NotificationsCenterCell: UICollectionViewCell {
 
     // MARK: - Public
 
+    fileprivate func setupAccessibility(_ viewModel: NotificationsCenterCellViewModel) {
+        accessibilityAttributedLabel = viewModel.accessibilityText()
+        isAccessibilityElement = true
+        
+        let moreActionAccessibilityActionLabel = viewModel.isRead ? CommonStrings.notificationsCenterMarkAsUnread : CommonStrings.notificationsCenterMarkAsRead
+        let moreAction = UIAccessibilityCustomAction(name: "More", target: self, selector: #selector(tappedMoreAction))
```
mcleinman, line 474: Is this what VoiceOver reads out? If so, instead of "More", should this be a translated string?

mcleinman (commented):
This is sounding great! Just a couple small things/questions.

Wikipedia/Code/NotificationsCenterCell.swift
mazevedofs: Yes, great catch!

Wikipedia/Code/NotificationsCenterCellViewModel.swift
mazevedofs: Great catch - I guess I initially thought it would require more complex operations to build this string

Wikipedia/Code/NotificationsCenterCellViewModel.swift
mazevedofs: In English VoiceOver reads it as "riːd" instead of "rɛd" which I thought was confusing given the context. 
In our `TalkPageTopicCell` we use the same solution, and in Portuguese, the translation works well. But I'm ok with removing it if it causes issues for any language.

mcleinman: I didn't realize this was a strategy we used elsewhere, sorry! In that case, could we re-use the same localization strings? We could move them into `CommonStrings`, and refer to the same one in each spot. 

We could probably update the translation's `comment` to explain that it's now used in a couple places. While it would be good if could update the name of the string, that would cause us to lose all the translations that have been created for this one, so probably best to keep it as `talk-page-discussion-unread-ipa-accessibility-attribute` and `talk-page-discussion-read-ipa-accessibility-attribute`.

mcleinman:
Apologies for the delay in looking at this again. After we resolve the pronunciation stuff, this should be good to go!

mcleinman:
A test is failing on this one. @mazevedofs let me know if you want to look at this together on Friday.

mcleinman:
Thanks for that fix!

PR #4152, opened 2022-03-11
Notification Center: iPad adjustments

mcleinman:
**Phabricator:** https://phabricator.wikimedia.org/T286607

### Notes
* In phab ticket, asked for approval on margins being a bit different than proposed - using `readableContentGuide` instead of requested exact numbers. Olga and Carolyn are comfortable with this.
* iPad cells now smaller. For larger font sizes, cell size will grow.
* Changed the `sourceRect` for the popover so it doesn't block the content. Current behavior was impact of giving a source size of zero.
* As discussed in meeting, we don't have control over the font size or truncation style of pop overs or action sheets, the fourth part of this ticket. If we want to control this, we'd need to switch to another type of control.

### Test Steps
1. Run app, look at Notifications Center and its submenus.
2. Check on both iPad and iPhone, and in various colors

### Screenshots/Videos
Filter margins (before/after)
Simulator Screen Shot - iPad Pro (9 7-inch) - 2022-03-11 at 13 17 31 Simulator Screen Shot - iPad Pro (9 7-inch) - 2022-03-11 at 13 28 41

Inbox margins (before/after)
Simulator Screen Shot - iPad Pro (9 7-inch) - 2022-03-11 at 13 30 18 Simulator Screen Shot - iPad Pro (9 7-inch) - 2022-03-11 at 14 02 33

Updated anchor for popover menu
Simulator Screen Shot - iPad Pro (9 7-inch) - 2022-03-11 at 14 33 46

Main screen - margins and cell height (before/after)
Simulator Screen Shot - iPad Pro (9 7-inch) - 2022-03-11 at 14 23 33 Simulator Screen Shot - iPad Pro (9 7-inch) - 2022-03-11 at 14 23 36

Large fonts
Simulator Screen Shot - iPad Pro (9 7-inch) - 2022-03-11 at 14 35 45
Simulator Screen Shot - iPad Pro (9 7-inch) - 2022-03-11 at 14 35 54 - 2022-03-11 at 14 36 16](https://user-images.githubusercontent.com/9295855/157983875-576775ed-c61a-4e67-86ab-417c3b5f4d28.png)
be122cac94ec.png)


Wikipedia/Code/NotificationsCenterCell.swift
```diff
@@ -356,18 +357,18 @@ final class NotificationsCenterCell: UICollectionViewCell {
         // Foreground and Background Container Constraints
 
         NSLayoutConstraint.activate([
-            swipeBackgroundFillView.leadingAnchor.constraint(equalTo: contentView.leadingAnchor, constant: NotificationsCenterCell.swipeEdgeBuffer * 2.0),
+            swipeBackgroundFillView.leadingAnchor.constraint(equalTo: contentView.readableContentGuide.leadingAnchor, constant: NotificationsCenterCell.swipeEdgeBuffer * 2.0),
```
tonisevener, line 360: I wonder if we could adjust the placement of this `readableContentGuide` check elsewhere so that the swipe actions go all the way to the edge of the screen? This would line up better with how the saved articles cells act. Also it would be good if the background color when selected / highlighted can extend to the very edges too. We might want to confirm with @carolynlimadeo on this first.

**Current PR:**
IMG_1249 2
IMG_1248
IMG_1247

**Ideally:**
IMG_1246
IMG_1245
IMG_1244

Wikipedia/Code/NotificationsCenterFilterView.swift
```diff
@@ -65,6 +66,7 @@ struct NotificationsCenterFilterItemView: View {
                 }
             }
         }
+        .padding(.horizontal, horizontalSizeClass == .regular ? 16 : 0)
```
tonisevener, line 69: I wonder if SwiftUI or the accessibility APIs give us a better constant for this that is more dependent on their dynamic type sizing. This is working fine for me though, so I would say looking into this is optional.

Edit: Maybe [this API](https://developer.apple.com/documentation/uikit/uifontmetrics/2877388-scaledvalue) would help.

Wikipedia/Code/NotificationsCenterCell.swift
```diff
@@ -337,12 +337,13 @@ final class NotificationsCenterCell: UICollectionViewCell {
         projectSourceContainer.addSubview(projectSourceLabel)
         projectSourceContainer.addSubview(projectSourceImage)
 
+        let minimumSummaryHeight = (traitCollection.horizontalSizeClass == .regular) ? 40.0 : 64.0
```
tonisevener, line 340: One issue I've noticed is that the 2nd label gets squished when are on Notifications Center and change iPad to split screen (see screenshot), but it works when loading this screen fresh while in a split screen.

IMG_1243

I think this is related to our cell height caching. I was able to fix it by making this change [here](https://github.com/wikimedia/wikipedia-ios/blob/40291b2d8b28d4cb35537c93675e1ad3d2972e6d/Wikipedia/Code/NotificationsCenterView.swift#L74):

```
     override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
        super.traitCollectionDidChange(previousTraitCollection)

        if previousTraitCollection?.preferredContentSizeCategory != traitCollection.preferredContentSizeCategory {
            emptyOverlayHeaderLabel.font = UIFont.wmf_font(.mediumBody, compatibleWithTraitCollection: traitCollection)
            emptyOverlaySubheaderLabel.font = UIFont.wmf_font(.subheadline, compatibleWithTraitCollection: traitCollection)
        }
        
        if previousTraitCollection?.preferredContentSizeCategory != traitCollection.preferredContentSizeCategory || previousTraitCollection?.horizontalSizeClass != traitCollection.horizontalSizeClass {
            calculatedCellHeight = nil
        }
    }
```

tonisevener (changes requested):
@mcleinman thanks for helping us out with this! I have a few suggestions, let me know what you think.

Wikipedia/Code/NotificationsCenterCell.swift
mcleinman: This has been fixed, thanks.

Wikipedia/Code/NotificationsCenterFilterView.swift
mcleinman: Done, thanks!

Wikipedia/Code/NotificationsCenterCell.swift
mcleinman: Done

mcleinman:
I believe this is ready for another review. Thanks to @tonisevener for her help with parts of this!

Wikipedia/Code/NotificationsCenterCell.swift
tonisevener: @mcleinman I'm seeing something odd when I first load the screen - the `messageSummaryLabel` appears squished at first, then expands out to the proper width after 1 second. I didn't notice it in `main`, so I think this is related to the `readableContentGuide` changes. I'm testing this on an iPad Pro 11-inch, iOS 15.2. This was the weird thing I remember seeing when getting those screenshots too.

out

mcleinman: Good catch. I couldn't isolate why this was happening - tried a few different things. Thus, I went to your way of making the readable content guide work - that seems to work great, as well as fix this problem. Thank you!

tonisevener:
@mcleinman Thanks - this is looking good! Just a couple of last things - hopefully they're an easy fix:

It looks like the non-Wikipedia project icons are still aligned to the edge. Screenshot from iOS15.2:

IMG_1251

This summary label bleeds too far into the project icon area. Oddly I don't see this behavior in the compact horizontal size class, it wraps well before the project icon. Which is good, but surprising since I don't think we're doing anything differently via code for that. Screenshot from iOS15.2:

IMG_C9EBA9EEA314-1

mcleinman:
@tonisevener I think this is finally set. Fingers crossed. That last set was sloppy, this one should finally fix the readable content.

tonisevener (approved):
🎉 Working great!

PR #4154, opened 2022-03-15
fix a crash after the first launch

ngn999:
**Phabricator:** 

### Notes
*  swift `try catch` cannot catch objc exception
*  every time I reinstall wikipedia-ios on emulator, it crashs after *first* launch.

### Test Steps
1. remove app from emulator
2. reinstall it( ^+⌘+R ) 
3. just click some tab
4. it crash


tonisevener (approved):
@ngn999 This looks great - thank you! I managed to find the area where this json is getting malformed in the first place. I'll make a PR for that, but this is good safety to have in case this happens again.

If you'd like, feel free to add your name to the contributors' section of [AboutViewController.plist](https://github.com/wikimedia/wikipedia-ios/blob/main/Wikipedia/Code/AboutViewController.plist) in a followup PR.

PR #4156, opened 2022-03-15
Fix scroll to left side of the scrollview when end dragging

MaximKorobov:
**Phabricator:** 

https://phabricator.wikimedia.org/T303881

### Notes
I found a problem in scrollViewWillEndDragging callback in NavigationBarHider class.
According to [Apple docs](https://developer.apple.com/documentation/uikit/uiscrollviewdelegate/1619385-scrollviewwillenddragging) developer could override targetContentOffset value. Some calculations for Y in this method ignores targetContentOffset.pointee.x value by overwriting it with 0 value.

### Test Steps

1. Open SVG file from any article. For example, with maps
2. Zoom in
3. Scroll to bottom-right corner to hide navigation bar
4. Scroll to top or vertically only. When end dragging ScrollView also change X position to zero.

### Screenshots/Videos

Before: http://e.pc.cd/Fd9otalK
After: http://e.pc.cd/Yd9otalK


mazevedofs (approved):
Looks great, thank you for your contribution!

PR #4157, opened 2022-03-15
Fix cold builds on main

OscarSwanros:
**Phabricator:** 

### Notes
* `main` was not compiling due to a dependency cycle between the WFM and ContinueReadingWidget targets.
* Also, removed a SwiftUI warning

### Test Steps
1.


tonisevener (approved):
Thanks for your contribution @OscarSwanros! Sorry this took so long to review - we were deep into another release. I can't seem to reproduce your compile failure. It could be that a subsequent Xcode release fixed it. I reverted that particular commit, but I do want to keep your warning fix.

Feel free to open another PR if you're still seeing compile failures, and let us know which simulator or device you're seeing it on. Thanks again!

PR #4158, opened 2022-03-17
NotificationsCenterCellViewModel tests

tonisevener:
**Phabricator:** 
Pre-work for https://phabricator.wikimedia.org/T302453

### Notes
This is some cleanup work I needed to do for the upcoming detail screen view model work. Most of this is sample JSON and unit tests for `NotificationsCenterCellViewModel`. A lot of the cell view model extensions will be reused and tweaked with the detail view model, so these unit tests will help us catch regressions in the next PR. This testing setup will also help us write tests for the detail view model.

This PR will be easiest to review commit-by-commit.

### Test Steps
1. Just some very light regression testing on Notifications Center cells, to be sure the data still looks okay.
2. Ensure unit tests pass.


Wikipedia/Code/NotificationsCenterCellViewModel+SheetActionExtensions.swift
```diff
@@ -4,66 +4,92 @@ import WMF
 
 extension NotificationsCenterCellViewModel {
     
-    enum SheetAction {
+    enum SheetAction: Equatable {
         case markAsReadOrUnread(SheetActionData)
         case custom(SheetActionData)
         case notificationSubscriptionSettings(SheetActionData)
+        
+        static func == (lhs: SheetAction, rhs: SheetAction) -> Bool {
```
staykids, line 12: Is this custom Equatable implementation required over just using the automatically synthesized conformance?

tonisevener: @staykids good point, I think that was a misplaced assumption of mine with enums like this. This is removed. Hope you don't mind but I noticed another custom implementation on `RemoteNotificationType`. I didn't notice any regressions after removing it so I added that to the commit as well.

staykids (commented):
I'm seeing some inconsistent test failures when testing locally (Sim and device):

- One related to `SectionEditorFindAndReplaceTests`
- And these related to the date formatting inconsistencies with the notifications

Any ideas?

test_failures

tonisevener:
@staykids Thanks for pointing these out! I think I have fixed them properly now. I think the timestamp failures were due to differences in time zones on the machines running them. I've been trying to paper over it (in this PR and later PRs) by picking different dates in the json that are more likely to line up on our local machines and the CircleCI executors, but I think a better fix is to set the default `NSTimeZone` in the setup method. Hopefully it's more consistent now.

For the find and replace failures - It only crashed for me if my simulator was already logged in, which could explain why CircleCI wasn't running into it. Ultimately the problem was that the remote notifications Core Data stack was set up twice in quick succession, because previously we were instantantiating two instances of `MWKDataStore` to run these tests - one temporary testing one within the test class itself and another inadvertently was instantiated via the `SectionFetcher` superclass init, which calls the `MWKDataStore` singleton. These tests automatically log the user out, which caused a batch delete call for the `RemoteNotifications` database, and that call was complaining about double entities. By tweaking the way we instantiate `SectionFetcher`, I don't see the errors anymore.

I think this tweak is pretty safe, and is one less `MWKDataStore` singleton call for us to deal with, but let me know if you have concerns.

staykids (approved):
👍🏽 Sticking to the logged out state is something I'll remember for future test related work.

PR #4160, opened 2022-03-17
Add Push permissions opt in to Notifications Center onboarding

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T302633 (partially)

### Notes
This is a follow on to https://github.com/wikimedia/wikipedia-ios/pull/4155 that adds the actual device-level system permissions request for Notifications. When the user taps the "Turn on push notifications" button during onboarding, we explicitly ask for system permissions if necessary then silently subscribe to Echo subscriptions without interrupting the user experience.

### Test Steps
This can be tested on Simulator or physical device, but of course Echo subscription/Push will only work on a real device build deployed via TestFlight/the App Store.

1. Uninstall/reinstall app and reset privacy permissions if necessary.
2. When logged in, tap the Notifications Center bell. Confirm you see the onboarding flow.
3. Tap "Turn on push notifications" in Notifications Center panel.
4. Confirm you're asked for system permissions for Notifications. If the user has already authorized this permission, they will not be prompted.


Wikipedia/Code/NotificationsCenterViewController.swift
```diff
@@ -831,3 +835,32 @@ extension NotificationsCenterViewController {
...
 }
+
+
+// MARK: - Device push permissions and silent Echo subscription
+
+extension NotificationsCenterViewController {
+
+    /// Mimics the Settings Push opt in approach, but Echo subscription fails silently to prevent interrupting the experience
+    func requestPushPermissionsAndSilentlySubscribeToEchoNotifications() {
+        deviceTokenRetryTask = RetryBlockTask { [weak self] in
```
tonisevener, line 846: After testing this a bit, do you think we should exit early in `presentingOnboardingPushOptInIfNecessary` (don't show any prompt) if the user has permissions denied or `UserDefaults.standard.wmf_isSubscribedToEchoNotifications = true`? Both situations can happen if the user already interacted with app push settings before going to notifications center for the first time, and as a result nothing ends up happening when the user taps "Turn on push notifications", which could be confusing.

I know we discussed failing silently which is still good here in case of network failures or something, but if we can predict that it will fail here in these two cases, maybe it's better to not show the prompt at all.

Feel free to push back if this feels like too much of an edge case, or if it will over-complicate things too much. It was just something that jumped out at me.

tonisevener (commented):
@staykids This looks good! Thanks for breaking this off into a separate PR - it helped with isolated testing. Just one potential improvement, but I don't think it's a must.

tonisevener:
@staykids FYI I pushed one minor fix I found, wrapping some `completionHandler` calls in WMFNotificationsController.m with a `nil` check. I was seeing a crash when attempting to turn on push without it, because the new call to subscribe has a `nil` completion handler.

Wikipedia/Code/NotificationsCenterViewController.swift
staykids: > do you think we should exit early in `presentingOnboardingPushOptInIfNecessary` (don't show any prompt) if the user has permissions denied or `UserDefaults.standard.wmf_isSubscribedToEchoNotifications = true`?

Good idea - updated!

Also, I missed that I forgot to subscribe to echo in the this flow if the device token is actually available 😅. Added that commit too!

PR #4163, opened 2022-03-22
Explore feed notifications onboarding card

mazevedofs:
**Phabricator:**  https://phabricator.wikimedia.org/T288693

### Notes
* The explore feed notification announcement card was updated to b used to alert users about the new notifications center feature. This PR updates the UI and the actions available from this card.

### Test Steps
1. Log in to your account
2. (It's necessary to wait for the data part to be done, a PR is coming up soon with these changes)
3. Check if the three-dot button actions work properly on the card (as it works on every other explore feed card)
4. Check if the "Turn on notifications" button takes you to the Notifications Center settings and  if the "dismiss" button removes the card from the feed


tonisevener (approved):
This looks good! Nice work hooking into the existing setup we have - I didn't notice any regressions. I pushed one minor string key tweak, and updated with `main` while I was in there. I'll merge once the tests pass.

There is one thing that jumps out to me, but it seems like a data thing so you can look into it in the next PR:

Can you remove the 2nd option in the action sheet only this notifications card? The one that says "Hide all 'Editing is now available' cards".

IMG_0658

mazevedofs:
Sure! I'll address it on this PR I'm working on

> This looks good! Nice work hooking into the existing setup we have - I didn't notice any regressions. I pushed one minor string key tweak, and updated with `main` while I was in there. I'll merge once the tests pass.
> 
> There is one thing that jumps out to me, but it seems like a data thing so you can look into it in the next PR:
> 
> Can you remove the 2nd option in the action sheet only this notifications card? The one that says "Hide all 'Editing is now available' cards".
> 
> IMG_0658

PR #4166, opened 2022-03-24
Add Notifications Center notification detail view

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T302453

### Notes
This PR adds the notification detail view UI. Previous PRs address the data layers.

### Test Steps
To access the detail view, tap a notification in your Notifications Center.

1. Change theme, device, and accessibility size to confirm view renders as expected.

### DIFF
--- a/Wikipedia/Code/NotificationsCenterAction.swift
+++ b/Wikipedia/Code/NotificationsCenterAction.swift
@@ -4,6 +4,13 @@ enum NotificationsCenterAction: Equatable {
     case markAsReadOrUnread(NotificationsCenterActionData)
     case custom(NotificationsCenterActionData)
     case notificationSubscriptionSettings(NotificationsCenterActionData)
+
+    var actionData: NotificationsCenterActionData? {
+        switch self {
+        case .notificationSubscriptionSettings(let data), .markAsReadOrUnread(let data), .custom(let data):
+            return data
+        }
+    }
 }
 
 struct NotificationsCenterActionData: Equatable {
--- a/Wikipedia/Code/NotificationsCenterDetailActionCell.swift
+++ b/Wikipedia/Code/NotificationsCenterDetailActionCell.swift
@@ -1,0 +1,62 @@
+import UIKit
+
+class NotificationsCenterDetailActionCell: UITableViewCell, ReusableCell {
+
+    // MARK: - Properties
+
+    var action: NotificationsCenterAction?
+    var theme: Theme = .light
+
+    // MARK: - Lifecycle
+
+    override init(style: UITableViewCell.CellStyle, reuseIdentifier: String?) {
+        super.init(style: .value1, reuseIdentifier: reuseIdentifier)
+        setup()
+    }
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    override func prepareForReuse() {
+        super.prepareForReuse()
+        action = nil
+    }
+
+    func setup() {
+        detailTextLabel?.font = UIFont.wmf_scaledSystemFont(forTextStyle: .footnote, weight: .regular, size: 13, maximumPointSize: 64)
+    }
+
+    // MARK: - Configuration
+
+    func configure(action: NotificationsCenterAction?, theme: Theme) {
+        self.theme = theme
+
+        backgroundColor = theme.colors.paperBackground
+        selectedBackgroundView = UIView()
+        selectedBackgroundView?.backgroundColor = theme.colors.midBackground
+
+        imageView?.tintColor = theme.colors.link
+        textLabel?.textColor = theme.colors.link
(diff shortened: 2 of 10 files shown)


Wikipedia/Code/NotificationsCenterDetailActionCell.swift
```diff
@@ -0,0 +1,62 @@
...
+
+        if let actionData = action.actionData {
+            let imageType = actionData.iconType
+            textLabel?.text = actionData.text
+            detailTextLabel?.text = actionData.destinationText
+            switch imageType {
+            case .custom(let name):
+                imageView?.image = UIImage(named: name)
+            case .system(let name):
+                imageView?.image = UIImage(systemName: name)
```
tonisevener, line 55: Optional change - we could use a symbol configuration here with a font to have the action icons scale up. But since this is capable of displaying mixed SF Symbols and custom icons, I'm not sure if we should - mixed scaling would look horrible.

Wikipedia/Code/NotificationsCenterDetailView.swift
```diff
@@ -0,0 +1,44 @@
...
+
+}
+
+// MARK: - Themeable
+
+extension NotificationsCenterDetailView: Themeable {
+
+    func apply(theme: Theme) {
+        backgroundColor = theme.colors.baseBackground
+        tableView.backgroundColor = theme.colors.baseBackground
```
tonisevener, line 41: Might want to `reloadData()` here or in the view controller so a system theme change on this screen will refresh the cells.

Wikipedia/Code/ReusableCell.swift
```diff
@@ -0,0 +1,11 @@
+import UIKit
+
+protocol ReusableCell {
+    static var reuseIdentifier: String { get }
+}
+
+extension ReusableCell {
+    static var reuseIdentifier: String {
```
tonisevener, line 8: Just an FYI, this felt familiar so I did some digging - we do have something similar in UIView+Identifier.swift for views and cells. But I do like that this feels both more specific to cells and more flexible as a protocol, so I'm open to keeping this here.

Wikipedia/Code/NotificationsCenterAction.swift
```diff
@@ -30,6 +30,13 @@ enum NotificationsCenterAction: Equatable {
             }
         }
     }
+
+    var actionData: NotificationsCenterActionData? {
+        switch self {
+        case .notificationSubscriptionSettings(let data), .markAsReadOrUnread(let data), .custom(let data):
```
tonisevener, line 36: Sorry you had to do this - I'm hating this enum that I made more and more as time goes on 😁.

Wikipedia/Code/NotificationsCenterDetailHeaderCell.swift
```diff
@@ -0,0 +1,152 @@
...
+        leadingImageView.tintColor = theme.colors.paperBackground
+        leadingImageView.backgroundColor = viewModel.headerImageBackgroundColorWithTheme(theme)
+
+        configureDetailLabelStackFor(sizeCategory: traitCollection.preferredContentSizeCategory)
+    }
+
+    // MARK: - Private
+
+    fileprivate func configureDetailLabelStackFor(sizeCategory: UIContentSizeCategory) {
+        if sizeCategory < .accessibilityMedium {
```
tonisevener, line 143: Good call 👍

tonisevener (changes requested):
Awesome to see this in place! Just one must-fix with cell reloading upon system theme change, the other stuff is optional or just ramblings.

Wikipedia/Code/ReusableCell.swift
staykids: Ah I missed this extension. Thanks for pointing it out! I have a minor quibble with the naming of it - `identifier` to me indicates something unique, whereas `reuseIdentifier` at least indicates it's not something globally unique, intended to be reused. Because you also see the value in this flexible protocol/specific to cell approach, I'd rather we eventually use _this_ extension for the current instances using the `UIView` one rather than using that one here.

Wikipedia/Code/NotificationsCenterDetailActionCell.swift
staykids: Yeah I agree - mixed image scaling would look bad. I think we could definitely solve this by using a wholly custom cell instead of relying on `UITableViewCell`'s `imageView`, but since we have this existing issue of table view cell image scaling elsewhere (like Settings), we should tackle this later.

Wikipedia/Code/NotificationsCenterDetailView.swift
staykids: Good catch - updated! Also added in extra commit that adjusts some font sizes per design request.

tonisevener (approved):
Looks good! We'll merge after the base PRs are in.

PR #4167, opened 2022-03-28
Update displaying logic for the explore feed notifications card

mazevedofs:
**Phabricator:**  https://phabricator.wikimedia.org/T288693

### Notes
* Business logic for displaying the new Notifications Center feed card

### Test Steps
1. Log into your account (updating from a version pre-notification center)
2. You should see the old notifications feed card - the new card UI is on a separate PR


Wikipedia/Code/MWKDataStore.m
```diff
@@ -461,6 +461,8 @@ - (void)performUpdatesFromLibraryVersion:(NSUInteger)currentLibraryVersion inMan
     
     if (currentLibraryVersion < 14) {
         [self.remoteNotificationsController deleteLegacyDatabaseFilesAndReturnError:nil];
+        NSUserDefaults *userDefaults = [NSUserDefaults standardUserDefaults];
+        userDefaults.wmf_shouldShowNotificationsExploreFeedCard = YES;
         [NSHTTPCookieStorage migrateCookiesToSharedStorage];
```
tonisevener, line 466: As mentioned below, let's also add this line to delete any old notifications cards from their database, to keep our content source logic simple:

```
[moc removeAllContentGroupsOfKind:WMFContentGroupKindNotification];
```

tonisevener (changes requested):
Thanks @mazevedofs! I have a some thoughts around making this logic refresh more often, while also hopefully keeping it fairly simple. I tried all of this out in my `T288693-feed-card-data-suggestions` [branch](https://github.com/wikimedia/wikipedia-ios/compare/T288693-feed-card-data...T288693-feed-card-data-suggestions), just to make sure I wasn't leading you down an unsolvable rabbit hole. Feel free to cherry-pick from that if you'd like. The git commit messages are hopefully descriptive enough.

A few of other things I discovered that I found handy to add:
1. A change to the [card sort order](https://github.com/wikimedia/wikipedia-ios/commit/1c2caf70ea75bd04ccae07fbcfa13351c6366e29), which helps push this thing up to the top.

2. I found and fixed [bug](https://github.com/wikimedia/wikipedia-ios/commit/a00c9b47e6e191e7ee5d652efa4bd8c424bcd061) where this card had its `isVisible` flag set to NO elsewhere after I moved logic over to the `WMFAnnouncementsContentSource.m`. That may have been the bug you were running into earlier. I'm not wild about this fix but it seems to help this card show more reliably for me.

3. I'm asking the `WMFAnnouncementsContentSource.m` to [update](https://github.com/wikimedia/wikipedia-ios/commit/0426c155b74439220b7fd9769520c1e923580273) upon user login/logout while they are using the app, so that the card hides or shows immediately rather than waiting until the next launch. It doesn't seem like we've been this reactive to login status before with Explore feed cards, but this update is pretty lightweight so I think it's worthwhile.

Wikipedia/Code/WMFFeedContentSource.m
```diff
@@ -335,15 +336,24 @@ - (void)saveGroupForNews:(NSArray<WMFFeedNewsStory *> *)news pageViews:(NSDictio
...
+    }
+}
+
+- (void)saveNotificationsGroupInManagedObjectContext:(NSManagedObjectContext *)moc date:(NSDate *)date {
     NSUserDefaults *userDefaults = [NSUserDefaults standardUserDefaults];
-    if (newsGroup && newsGroup.isVisible && ![userDefaults wmf_inTheNewsNotificationsEnabled] && ![userDefaults wmf_didShowNewsNotificationCardInFeed]) {
-        NSURL *URL = [WMFContentGroup notificationContentGroupURLWithLanguageVariantCode:self.siteURL.wmf_languageVariantCode];
-        [moc fetchOrCreateGroupForURL:URL ofKind:WMFContentGroupKindNotification forDate:newsGroup.date withSiteURL:self.siteURL associatedContent:nil customizationBlock:NULL];
+
+    if (userDefaults.wmf_shouldShowNotificationsExploreFeedCard && ![userDefaults wmf_didShowNewsNotificationCardInFeed] && self.fetcher.session.isAuthenticated ){
```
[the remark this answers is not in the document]
mazevedofs: Thanks for the deep dive into my PR! It shouldn't use the old card userDefaults flag. I guess repurposing the card led me to several assumptions I shouldn't be making. I also agree that keeping this card's logic in `WMFAnnouncementsContentSource` makes more sense. It also makes it easier for future maintenance since it's the same type of card and no longer tied to all the news logic. I cherry-picked the commits from your branch so we keep the commit history.

Wikipedia/iOS Native Localizations/en.lproj/.dat.nosync71be.xLCM1b
```diff
@@ -0,0 +1,828 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
tonisevener, line 1: hmm I'm not sure where this file came from - can you remove it?

Wikipedia/Code/WMFAnnouncementsContentSource.m
```diff
@@ -123,6 +126,33 @@ - (void)updateVisibilityOfNotificationAnnouncementsInManagedObjectContext:(NSMan
...
+
+        WMFContentGroup *currentNotificationsCardGroup = [moc newestGroupOfKind:WMFContentGroupKindNotification];
+
+        if (isLoggedIn) {
+            if (currentNotificationsCardGroup) {
+                if (!currentNotificationsCardGroup.isVisible && !currentNotificationsCardGroup.wasDismissed) {
+                    currentNotificationsCardGroup.isVisible = YES;
+                }
+            } else {
+                WMFContentGroup *newNotificationsCardGroup = [moc createGroupOfKind:WMFContentGroupKindNotification forDate:[NSDate date] withSiteURL:self.siteURL associatedContent:nil];
```
tonisevener, line 143: My mistake on this one - we should probably just use the `date` parameter here instead of creating a new `[NSDate date]` here:

```suggestion
                WMFContentGroup *newNotificationsCardGroup = [moc createGroupOfKind:WMFContentGroupKindNotification forDate:date withSiteURL:self.siteURL associatedContent:nil];
```

Wikipedia/Code/WMFAnnouncementsContentSource.m
```diff
@@ -67,6 +67,7 @@ - (void)loadContentForDate:(NSDate *)date inManagedObjectContext:(NSManagedObjec
```
tonisevener, line 67: **Note:** this fix is not in this file, but I just needed to leave a comment somewhere that I can resolve later. This is a bug I missed in the UI PR.

I think the date formatter for the cell subtitle label might be off. It's not noticeable when the section title says "Today", but once the card is on screen, if you terminate and move your device date up a few days, then relaunch and and scroll down until you see the card again, you'll notice that it's off from the section title:

IMG_0660

Can you update it so it matches with however the section title is calculating its full date string? I think it's [this line](https://github.com/wikimedia/wikipedia-ios/blob/40291b2d8b28d4cb35537c93675e1ad3d2972e6d/Wikipedia/Code/WMFContentGroup%2BWMFFeedContentDisplaying.m#L104).

tonisevener (changes requested):
Thanks @mazevedofs! Just a few more things, then this will be good to merge.

tonisevener (approved):
🎉 Thanks! Looks great!

PR #4170, opened 2022-03-31
Reorder messages about talk pages

amire80:
"Talk page" is the most common name, so it should
appear first to help translators be in the right
context.

**Phabricator:** 

### Notes
* 

### Test Steps
1.


tonisevener:
@amire80 Update from design - since this seems to be an English-only thing, we'll just use the term talk pages for all of these instead of making the collaboration and discussion distinction. I'll post a PR cleaning these extra strings out. Thanks for keeping an eye on this!

PR #4179, opened 2022-04-07
Address and remove notification TODO comments

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T291383

### Notes
This cleans up any TODO comments I had sprinkled throughout the notifications feature. TODOs I addressed are a service extension version check for the push content and adding a couple of missing `didLogIn` notifications to the authentication manager. I didn't feel like the others needed a change, but I'm happy to discuss further.

### Test Steps

Some light regression testing on the service extension is needed to be sure the content is still populating:

1. Enable push notifications via app settings on a device - take the device token sent in the echo subscription API call and use it in our push utility in step 3.
2. Do something on Desktop Wikipedia to trigger a notification.
3. Trigger notification via our push utility.
4. Confirm push content received on device populates with specific notification data, and doesn't just fall back to "New activity on Wikipedia" every time.

Also test that the authentication manager still works:

1. Regression test logging in and out. Be sure notifications bell icon still hides after logout, and shows after login.
2. Test new functionality: Enter airplane mode (and turn off WiFi) before launching the app while logged in. Before the bell icon would not display, even though the app is technically in a logged in state. Now it does show. This is good because there is still useful persisted notifications to display to the user on this screen. Further offline error handling will occur in https://phabricator.wikimedia.org/T287766.


WMF Framework/Remote Notifications/Model/RemoteNotificationType.swift
```diff
@@ -102,7 +102,7 @@ public extension RemoteNotificationType {
                 self = .userRightsChange
             case "editReverted":
                 self = .editReverted
-            case "loginFailKnownDevice", //for filters this represents any login-related notification (i.e. also loginFailUnknownDevice, loginSuccessUnknownDevice, etc.). todo: clean this up. todo: split up into login attempts vs login success?
+            case "loginFailKnownDevice", //for filters this represents any login-related notification (i.e. also loginFailUnknownDevice, loginSuccessUnknownDevice, etc.).
```
tonisevener, line 105: These login filter comments are already documented in https://phabricator.wikimedia.org/T301129.

WMF Framework/Remote Notifications/Model/RemoteNotificationType.swift
```diff
@@ -169,13 +169,13 @@ public extension RemoteNotificationType {
             .pageLinked,
             .connectionWithWikidata,
             .emailFromOtherUser,
-            .mentionInTalkPage, //todo: combine this and edit summary mention to "received mention"?
```
tonisevener, line 172: Already confirmed that these do not need to be combined in https://phabricator.wikimedia.org/T301129.

Wikipedia/Code/NotificationsCenterViewModel.swift
```diff
@@ -157,7 +155,6 @@ final class NotificationsCenterViewModel: NSObject {
             switch result {
             case .failure(let error):
                 DDLogError("Error refreshing notifications: \(error)")
-                //TODO: show some sort of error state
```
tonisevener, line 160: These will be addressed in https://phabricator.wikimedia.org/T288773 and https://phabricator.wikimedia.org/T287766.

staykids (approved):
~~// TODO: Merge PR~~

PR #4181, opened 2022-04-07
Remove default primary button border from `ScrollableEducationPanelViewController`

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T302633#7832964 (partially)

### Notes
This PR addresses design feedback on the  push opt-in panel. It removes the primary button border from this panel and all other panels.

### Test Steps
1. Test `ScrollableEducationPanelViewController` panel presentations (like the Notifications Center push opt-in panel) and confirm there is no border on the primary button.

panel_button


tonisevener (commented):
Looking good! My only feedback is that the Sepia primary button seems to be missing a background color - maybe that was intentional back when these panels were made? @carolynlimadeo 

IMG_0661


Here's the current bordered panel on Sepia:

IMG_0662

carolynlimadeo:
Good catch! Can we use the Sepia 'base' color (#E1DAD1) as the fill for this button? Thanks!

staykids:
> Good catch! Can we use the Sepia 'base' color (#E1DAD1) as the fill for this button? Thanks!

👍🏽 

@tonisevener Updated and ready for another look!

PR #4185, opened 2022-04-12
Truncate `NotificationsCenterDetailActionCell` primary label text

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T302453

### Notes
The default `value1` `UITableViewCell` configuration appears to favor completely discarding the detail text label if the primary text label is sufficiently long. In order to support the design intention of truncating the title label but still preserving the destination label, this PR switches to custom layout for this cell type.

### Test Steps
1. Do a light regression test against `main`, confirming these action cells now truncate the title text without discarding the destination text. Confirm it truncates long text as shown below.

truncate


tonisevener (approved):
Working well!

PR #4186, opened 2022-04-13
Notifications - displaying error banners

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T288773 (partial, no connection screenshot)
https://phabricator.wikimedia.org/T287766

### Notes
This PR presents a banner over Notifications Center if some sort of error occurred while it first loads, or when the user tries to mark something as read/unread. I am using our existing `WMFAlertManager` to display the banner, as well as existing error strings for descriptions.

My goal was to keep changes light touch. I tried (for probably too long) to display an identifiable code in the banner too, so that we can seek it out in the source code, but I found it difficult to refactor and it felt like too much of a lift at this point. I think it would be better to tackle that in a holistic way later, when these banners go through a redesign.

### Test Steps
1. Run the app in airplane mode with no Wifi connection. Tap into Notifications Center. You should see a "No internet connection" banner at the top of Notifications Center (but note, purposefully not on the onboarding modal, if that appears).
2. Try marking as read/unread via the swipe menu button, the more action sheet, the edit contextual menu, and marking all as read. You should see a "No internet connection" banner at the top when performing these actions. 
3. Try changing the http response of a notifications-related call to 500 Internal Server Error (I did this by mapping local through a proxy), then perform the related action in the app. You should see a "The app received an unexpected response from the server. Please try again later." banner at the top.
5. Try forcing one of our custom errors via code (for example, [this](https://github.com/wikimedia/wikipedia-ios/blob/ca991a804bb454a7b8b8130481bcb5ef41bc4fc4/WMF%20Framework/Remote%20Notifications/RemoteNotificationsController.swift#L127) `databaseUnavailable` error).  When performing the action, you should see a generic "Something went wrong. Please try again later." banner at the top.

IMG_0690
IMG_0689
IMG_0692


mazevedofs (approved):
Working great!

PR #4189, opened 2022-04-18
Add interruption levels and relevance scores to incoming notifications

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T285343

### Notes
Per the Echo Notification mapping spreadsheet in the description of the linked Phab ticket, adds passive, active, and time sensitive notification priorities to our supported types.

### Test Steps
You can manually assign these interruption levels and relevance scores to the `bestAttemptContent` in `NotificationService.swift` and use our Push utility to test these. Otherwise you can build to device, trigger the notifications types below, then use our Push utility to cause the push notification to fire on device:

1. Confirm you are logged into an account in the app with Settings > Push notifications enabled.
2. Tap your device's sleep/wake button to put it in sleep mode with the display off. Find an `active` notification type in our mapping doc, and trigger that notification type for the account in step 1. Confirm your device screen lights up with that notification displayed on the lock screen.
3. Tap your device's sleep/wake button to put it in sleep mode with the display off. Find a `passive` notification and trigger it on the same account. Confirm that your device screen _does not_ light up, but when you wake your device and visit iOS's notification center that this notification is visible.
4. Tap your device's sleep/wake button to put it in sleep mode with the display off. Find a `time sensitive` notification and trigger it on the same account. Confirm your device lights up displaying that notification, and that `Time Sensitive` is displayed in a label above the notification's title.
5. Enable a [Focus mode](https://support.apple.com/en-us/HT212608) and allow `Time Sensitive` notifications for that focus mode. Repeat Step 4 and confirm the time sensitive notification "breaks through" your focus mode and is displayed on your lock screen. Repeat Steps 2 and 3 and confirm the active and passive notifications _do not_ break through your focus mode and aren't displayed on your lock screen but are displayed in iOS's notification center.
6. Using that same Focus mode, disallow `Time Sensitive` notifications. Repeat Steps 2, 3, and 4 and confirm none of the notifications "break through" your focus mode on the lock screen. Confirm they are visible when your navigate to iOS's notification center.

time_sensitive


NotificationServiceExtension/NotificationService.swift
```diff
@@ -67,6 +67,18 @@ class NotificationService: UNNotificationServiceExtension {
...
                 }
 
+                // Assigning interruption level and relevance score only available starting on iOS 15
+                if #available(iOS 15.0, *) {
+                    if finalNotifications.notificationsToDisplay.count == 1, let notification = finalNotifications.notificationsToDisplay.first {
+                        let priority = RemoteNotification.typeFrom(notification: notification).priority
+                        bestAttemptContent.interruptionLevel = priority.interruptionLevel
+                        bestAttemptContent.relevanceScore = priority.relevanceScore
+                    } else {
+                        bestAttemptContent.interruptionLevel = RemoteNotificationType.bulkPriority.interruptionLevel
```
tonisevener, line 77: There is a chance that there are multiple notifications that all come from the same talk page (this is caught within the conditional above on line 61), in which case it could get more of a bulk talk page message treatment ("N new messages on {title})". These types would get the lower bulkPriority of .passive/0.2, but I think they should match the "Multiple messages" row in the spreadsheet. Maybe within here add something like this?

```
if NotificationServiceHelper.allNotificationsAreForSameTalkPage(notifications: finalNotifications.notificationsToDisplay) {
    bestAttemptContent.interruptionLevel = RemoteNotificationType.bulkTalkPriority.interruptionLevel //.timeSensitive
    bestAttemptContent. relevanceScore = RemoteNotificationType.bulkTalkPriority. relevanceScore //0.9
} else {
    bestAttemptContent.interruptionLevel = RemoteNotificationType.bulkPriority.interruptionLevel //passive
    bestAttemptContent. relevanceScore = RemoteNotificationType.bulkPriority.relevanceScore //0.2
}
```

tonisevener (commented):
Nice @staykids! Just a couple of things from code review. I'll pull it down and test on device tomorrow.

WMF Framework/Remote Notifications/Model/RemoteNotificationType.swift
```diff
@@ -230,3 +230,52 @@ public extension RemoteNotificationType {
...
+            return (.passive, 0.5)
+        case .editMilestone:
+            return (.passive, 0.4)
+        case .failedMention:
+            return (.active, 0.5)
+        case .successfulMention:
+            return (.passive, 0.65)
+        case .welcome:
+            return (.active, 0.8)
+        case .loginFailUnknownDevice, .loginFailKnownDevice, .loginSuccessUnknownDevice:
```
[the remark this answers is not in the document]
staykids: Good catch! I made the change – we can revert if it's not the case, but sure seems like you're right after reading the descriptions in the sheet.

NotificationServiceExtension/NotificationService.swift
staykids: Whoops, I missed that sorry. Rationale to treat them with same priority as one makes sense – updated!

tonisevener (approved):
Working well for me!

PR #4190, opened 2022-04-18
Use readable width for cell layout in `NotificationsCenterDetailView`

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T286607 (partially)

### Notes
Lays out detail view table view to mimic Notifications Center's readable content margins.

### Test Steps
1. Open Notifications Center on iPad and tap a notification. Confirm that, similar to the Notifications Center itself, the Notification detail view cells are layout to respect the readable content margins (has leading and trailing margin space to those device edges).


tonisevener (approved):
Working great!

PR #4195, opened 2022-04-20
Notifications - Split up 'Login issues' filter into 'Login attempts' and 'Login success'

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T301129

### Notes
This breaks out the "Login issues" filter into two separate filters - Login attempts and Login success - to match the mocks. These were grouped together in a haphazard way before (which we pulled in via the prototype merge). I spent a while trying to refactor it to a cleaner way, but ultimately felt like it was too big of a change at this point.

### Test Steps
This is a bit difficult to test because I have only ever triggered one login notification type. For this task I faked it a bit by changing the `categoryString` and `typeString` values when we [save](https://github.com/wikimedia/wikipedia-ios/blob/0ccf042bc0eb47a04811b3308f39363ef982f11b/WMF%20Framework/Remote%20Notifications/Model/RemoteNotificationsModelController.swift#L359-L380) notifications in Core Data:

```
        let fakeCategoryString: String
        let fakeTypeString: String
        switch Int.random(in: 1...3) {
        case 1:
            fakeCategoryString = "login-fail"
            fakeTypeString = "login-fail-new"
        case 2:
            fakeCategoryString = "login-fail"
            fakeTypeString = "login-fail-known"
        case 3:
            fakeCategoryString = "login-success"
            fakeTypeString = "login-success"
        default:
            fakeCategoryString = ""
            fakeTypeString = ""
        }

        let isRead = notification.readString == nil ? NSNumber(booleanLiteral: false) : NSNumber(booleanLiteral: true)
        moc.wmf_create(entityNamed: "RemoteNotification",
                                                withKeysAndValues: [
                                                    "wiki": notification.wiki,
                                                    "id": notification.id,
                                                    "key": notification.key,
                                                    "typeString": fakeTypeString,
                                                    "categoryString" : fakeCategoryString,
                                                    "section" : notification.section,
                                                    "date": date,
                                                    "utcUnixString": notification.timestamp.utcunix,
                                                    "titleFull": notification.title?.full,
                                                    "titleNamespace": notification.title?.namespace,
                                                    "titleNamespaceKey": notification.title?.namespaceKey,
                                                    "titleText": notification.title?.text,
                                                    "agentId": notification.agent?.id,
                                                    "agentName": notification.agent?.name,
                                                    "isRead" : isRead,
                                                    "revisionID": notification.revisionID,
                                                    "messageHeader": notification.message?.header,
                                                    "messageBody": notification.message?.body,
                                                    "messageLinks": notification.message?.links])
```

But, note this can cause issues when revisiting a Notifications Center, because the randomized assignment will change the type for an existing notifications ID (something that would never happen in production), causing the filter to look broken. So only test via this way on a fresh install, without revisiting Notifications Center. This way will also still use the API values for the notification message content, which makes things look a bit off.

1. Visit Notifications Center.
2. Open Filters modal, confirm the Login issues type is now split up into Login attempts and Login success. Confirm subtitles appear correctly, theme correctly, and scale correctly.
3. Turn off all notification types, and turn on only Login success. Confirm Notifications Center only shows types that say "Login from an unfamiliar device"
4. Go back to the filters modal, turn off the Login success toggle and turn on the Login attempts toggle. Confirm Notifications Center only shows types that say "Multiple failed log in attempts" OR "Failed log in attempt".

### Screenshots/Videos

IMG_4A93A1A09909-1


staykids (approved):
👍🏽 looks good using your test code and steps. Just a merge conflict to resolve before merging.

PR #4197, opened 2022-04-21
Reorder swipe actions to match documentation

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T288687 (partially)

### Notes
Updates the order of the secondary actions in the swipe `More` action sheet to match our current documentation.

### Test Steps
I did not test every notification type, but instead verified the order of construction of the secondary actions matched the spreadsheet. The naming of things between the spreadsheet and implementation is not identical, so hopefully I identified things correctly in this update.

1. Open Notifications Center. Swipe a notification, and tap the More action.
2. Confirm the order of secondary actions displayed in the action sheet matches the spreadsheet linked in the Phab task description. The first action should be `Mark as read/unread` and the last should be `Notifications settings`.


tonisevener (approved):
Looks good to me!

PR #4198, opened 2022-04-22
Automatically select search bar when opening with search

mcleinman:
**Phabricator:** https://phabricator.wikimedia.org/T305892

### Notes
* One side effect: If a user doesn't have the setting enabled AND has a search view controller open AND hard quits the app (or it's closed in the background), when the app next opens it will open with the search box focused (instead of unfocused, as it is now). That seems okay, but wanted to flag it.

### Test Steps
1. Run app. Go to search screen. Hard quit app.
2. Re-open app. You should be on search screen, but search box is not focused - to type something in, you need to tap in the box.
3. In Settings, under Search, turn on the "always open to search" setting.
4. Repeat steps 1 and 2 - the search box should now be focused when you re-open the app.


tonisevener (approved):
Looks good, thanks!

PR #4200, opened 2022-04-25
Track notifications interactions

mazevedofs:
**Phabricator:** 

### Notes
* This PR implements the analytics tracking for the notifications. We are using the new analytics service to track the new iOS Notifications schema.

### Test Steps
1. Launch the app, and log into your account to access your notifications.
2. Click on the notification, and from the detail view, click on some of the actions.
3. From the swipe action, choose either mark as read/unread or one of the `More` actions from the action sheet
4. Select mark as read/unread from the Edit option
5. You should be able to check (with a network inspector) the calls to our API, be sure they're successful, and are sending the correct data for each action


Wikipedia/Code/NotificationsCenterCommonViewModel+ActionExtensions.swift
```diff
@@ -168,11 +168,22 @@ extension NotificationsCenterCommonViewModel {
...
+        if notification.type == .userTalkPageMessage {
+            return .userTalk
+        } else if namespace == .userTalk {
+            return .senderPage
+        } else if namespace == .talk {
+            return .articleTalk
+        } else {
+            return .article
+        }
+    }
```
tonisevener, line 181: Yeah, this is the trickiest action type since that `titleAction` method could create several different types of actions. I think let's purely only look at the namespace here to simplify, and test heavily. If we're creating an action URL with the namespace "User_talk:", then that should cover your first conditional, no need to look at the notification type.

```suggestion
    private func getActionType(namespace: PageNamespace) -> RemoteNotificationActionType? {
       if namespace == .userTalk {
            return .userTalk
        } else if namespace == .talk {
            return .articleTalk
        } else if namespace == .user {
            return .senderPage
        } else if namespace == .main {
            return .article
        } else {
            return nil
        }
    }
```

Wikipedia/Code/NotificationsCenterCommonViewModel+ActionExtensions.swift
```diff
@@ -229,7 +240,7 @@ extension NotificationsCenterCommonViewModel {
             return nil
         }
 
-        let data = NotificationsCenterActionData(text: title, url: url, iconType: .document, destinationText: destinationText(for: url))
+        let data = NotificationsCenterActionData(text: title, url: url, iconType: .document, destinationText: destinationText(for: url), actionType: nil)
```
tonisevener, line 243: I'll confirm to be sure they don't want any action strings for these remaining types - it should be easy to add them since the action takes any `String`. I'll update when I hear back!

tonisevener (changes requested):
@mazevedofs This is looking great! Some initial comments from code review - I'll confirm a couple of things with analytics and will let you know what I hear. I'll pull down and test after that.

Wikipedia/Code/NotificationsCenterAction.swift
```diff
@@ -18,4 +18,19 @@ struct NotificationsCenterActionData: Equatable {
     let url: URL?
     let iconType: NotificationsCenterIconType?
     let destinationText: String?
+    let actionType: RemoteNotificationActionType?
 }
+
+public enum RemoteNotificationActionType: String {
```
tonisevener, line 24: I think it might be a bit cleaner if this was nested within `NotificationsCenterActionData` and called `LoggingLabel` or something like that, to make it clear that this is only used for analytics and doesn't power other logic in the app.

Wikipedia/Code/NotificationsCenterCommonViewModel+ActionExtensions.swift
mazevedofs: Thanks for checking that! I was sending `nil` for the actions the schema didn't cover in case it could be somehow useful, but I can see that it might not be the case

tonisevener (changes requested):
A few more items I found after testing.

I also made this [suggestion commit](https://github.com/wikimedia/wikipedia-ios/commit/a7dbf4d1578812886739fbad1ef914073370cb04) for the trickiest new `link_` action, feel free to use it if you'd like. The other new actions should be more simple - hopefully just replacing those `nil` action strings that you were sending before.

mazevedofs:
> 

Thanks for that! I cherry-picked it into this PR, I guess it's ready for another look.

Wikipedia/Code/NotificationsCenterViewController.swift
```diff
@@ -803,7 +803,8 @@ extension NotificationsCenterViewController: NotificationsCenterCellDelegate {
             switch action {
             case .markAsReadOrUnread(let data):
                 alertAction = UIAlertAction(title: data.text, style: .default, handler: { alertAction in
-                    let shouldMarkRead = cellViewModel.isRead ? false : true
+                    let shouldMarkRead = data.actionType == .markRead
+                    print(shouldMarkRead)
```
tonisevener, line 807: Can you remove this print?

Wikipedia/Code/NotificationsCenterCommonViewModel+ActionExtensions.swift
```diff
@@ -168,22 +168,35 @@ extension NotificationsCenterCommonViewModel {
...
     // [Article where link was made]
-    var pageLinkToAction: NotificationsCenterAction? {
+    var pageLinkFromAction: NotificationsCenterAction? {
         guard let url = pageLinkToURL,
               let title = url.wmf_title else {
             return nil
         }
 
-        let data = NotificationsCenterActionData(text: title, url: url, iconType: .document, destinationText: destinationText(for: url))
+        let data = NotificationsCenterActionData(text: title, url: url, iconType: .document, destinationText: destinationText(for: url), actionType: .linkedFromArticle) //?????
```
tonisevener, line 199: Can you remove the question marks?

tonisevener (changes requested):
Thanks @mazevedofs! Just a few more things, also I think there are still a couple of unresolved comments of the UUID() within a `for` loop.

mazevedofs:
> Thanks @mazevedofs! Just a few more things, also I think there are still a couple of unresolved comments of the UUID() within a `for` loop.

I guess I got all of them now!

PR #4204, opened 2022-04-29
Fix notification operation memory leaks

tonisevener:
**Phabricator:** 
N/A

### Notes
I noticed a lot of new memory leaks coming from the notifications operations when digging into #4203. This fixes them.

### Test Steps
1. Run leaks instrument on device. Log in via Settings (if you aren't already) so that notifications begin importing.
2. At this point in `main` you would start to see many leaks of foundation objects and RemoteNotification operations. This clears most of those out (I only see 2 now and they seem unrelated to notifications).

**Confirm cross wiki and reauth operations still work after changes:**
1. Fresh install app, log into an account that has an unread notification on a Wiki that is **not** an app language, wikidata or commons. You should still see that notification imported and displaying in notifications center.
2. While app is logged in on app, log into your same account on Desktop. Then log out on Desktop. Revisit Notifications Center on app. If you watch at this point via a Proxy, you will see that the initial notification refresh calls return a logged out error, but then it should automatically refresh the token and reattempt the notifications calls.


mazevedofs (approved):
Looks great, no leaks related to Notifications!

PR #4209, opened 2022-05-06
Add pull to refresh to Notifications Center

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T305657

### Notes
Adds pull to refresh to the Notifications Center. Also prevents the `attemptingToRefreshBeforeDeadline` error from ever displaying to the user in the Notifications Center.

Similar to the article view controller, it was a bit indirect to get to a working solution that didn't experience severe offset jittering when refreshing. With the initial approach, when pulling to refresh the center, the offset of the collection view would animate wildly before settling and ending refresh. I had tracked this down to the snapshotting API in `reconfigureCells(...)` and resolved it with the new API (with old behavior) in iOS 15 `applySnapshotUsingReloadData`, but this caused a regression in the redisplay of cells when switching to edit mode. To resolve this, I removed the iOS 15 snapshotting and just use the fallback iOS 14 behavior. So far, I haven't noticed a regression when using this approach.

### Test Steps

Generally confirm you aren't noticing any regressions in the Notifications Center itself.

1. Open the Notifications Center, let the initial load complete
2. Trigger a notification on your account from elsewhere (like a user talk page message from Desktop)
3. Pull to refresh the Notifications Center and confirm your new notification is inserted into the list

Specifically related to Edit mode, would be good to confirm:

- Switching to edit mode removes the user's ability to pull to refresh, and returning to normal display enables it again
- While in edit mode, new elements that are displayed as you scroll are configured to display in the expected edit state (and not the normal cell display state)
- As you select some cells in edit mode, scroll those cells off screen and return to them, they are still selected as expected


tonisevener (approved):
Looks good! All the pull to refresh functionality works well across the 3 OSes in my testing. I pushed one minor indentation change. 

I am seeing that edit mode cell state weirdness on iOS13 & 14 simulators, but it's looking like I can reproduce that in the 6.9.0 branch too. I think it's unrelated to these changes. I'll spin off another Phab task for that.

PR #4211, opened 2022-05-09
Attempt to fix crashes with background tasks

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T298216
https://phabricator.wikimedia.org/T291299

### Notes
https://phabricator.wikimedia.org/T298216
This PR reworks the way our `WMFDatabaseHouseKeeper` class calls are made, in hopes of fixing the crashes related to `WMFDatabaseHouseKeeper.deleteStaleUnreferencedArticles` calls. I noticed some of these crash with timeout errors such as:

```
<RBSTerminateContext| domain:10 code:0x8BADF00D explanation:scene-update watchdog transgression: application<org.wikimedia.wikipedia>:35350 exhausted real (wall clock) time allowance of 10.00 seconds
```

To buy us more time, I think we should try doing database maintenance work via the `BackgroundTasks` framework instead of what we're doing now. The first note in [this documentation](https://developer.apple.com/documentation/uikit/app_and_environment/scenes/preparing_your_ui_to_run_in_the_background/extending_your_app_s_background_execution_time?language=objc) says to avoid the background extension API as the app enters the background (which we're doing), and [this documentation](https://developer.apple.com/documentation/backgroundtasks/choosing_background_strategies_for_your_app?language=objc) suggests a `BGProcessingTask` is a good candidate for database maintenance work.

https://phabricator.wikimedia.org/T291299
The other fix I did was a single line change to avoid scheduling these Background Tasks when terminating. Because we called `applicationDidEnterBackground` from `applicationWillTerminate`, I suspect scheduling the background refresh task at the termination point in the app lifecycle wasn't intentional. Now I'm breaking off the logic so that `applicationWillTerminate` no longer calls into `applicationDidEnterBackground`.

Background task testing can only occur on device. To confirm the database is cleaned out, I created a testing branch at `AppContainerTesting` which allows us to export the app container on device and inspect the sqlite tables.

### Test Steps
1. First lower a couple of our maximums to make testing easier. Set the [WMFExploreFeedBaximumNumberOfDays](https://github.com/wikimedia/wikipedia-ios/blob/e0a6a7f45cf4e42ab30d84549946dc3f3ace6075/WMF%20Framework/WMFExploreFeedContentController.m#L13) to 3, and set this [talk page fetch offset](https://github.com/wikimedia/wikipedia-ios/blob/e0a6a7f45cf4e42ab30d84549946dc3f3ace6075/Wikipedia/Code/WMFDatabaseHouseKeeper.swift#L43) to 3.
2. Change device time to one week earlier than today, then fresh install and run on device. 
3. Populate a week's work of Explore feed data. Ensure Explore feed is populated with data from 1 week ago. Background and change device time to the next day. Foreground and pull to refresh to load the next day. Do this for the next several days until you reach today's date.
4. Populate several talk pages of data. Search User_talk: and visit several talk pages.
5. Now export your app container by going to App Settings > Help and Feedback > Export User Data" and airdrop to your machine. Open up the zip file, then inspect Wikipedia.sqlite (I inspect with the Liya Mac app).
6. Open the ZWMFContentGroup table, sort by ZMidnightUTCDate and observe the oldest date - it should be about 1 week ago. Also open the ZTalkPage table and observe how many records there are - it should be the same as the number of pages you visited in step 4.
7. NOW execute the background database cleanup task. Set a breakpoint right after task scheduling [here](https://github.com/wikimedia/wikipedia-ios/blob/b2a693672852457a39763681820f7f3abd761b97/Wikipedia/Code/AppDelegate.m#L263). Background the app, then when you hit the breakpoint print this in the console:

```
e -l objc -- (void)[[BGTaskScheduler sharedScheduler] _simulateLaunchForTaskWithIdentifier:@"org.wikimedia.wikipedia.databaseHousekeeper"]
```

After a little bit you should see a printout in the console saying:

```
Simulating launch for task with identifier org.wikimedia.wikipedia.databaseHousekeeper
```

When you see this, hit the play button in the debugger to continue executing the program. The task will run and then the console will print out:

```
Marking simulated task complete: <BGProcessingTask: org.wikimedia.wikipedia.databaseHousekeeper>
```

When this happens, export the app container again like step 5, and inspect the sqlite tables again like step 6. You will see that there are now only 3 days of data for ZWMFContentGroup, and there should now only be 3 records for ZTalkPage.

### Screenshots/Videos

Before cleanup:

Screen Shot 2022-05-09 at 9 11 25 AM

Screen Shot 2022-05-09 at 9 11 00 AM

After cleanup:
Screen Shot 2022-05-09 at 9 14 56 AM

Screen Shot 2022-05-09 at 9 14 41 AM


staykids (approved):
Very cool – I haven’t messed with this API much myself so was fun to re-read about it while actually playing with an implementation. 

Yeah I agree on the termination delegate method explicitly calling the background delegate method looking fishy. Thanks for that helpful branch to simplify exporting the App Group data. My testing matches your test steps, where the tables appear to be pruning data as expected. In my testing, it seemed like the time between executing the background task and it completing were ~1 second of execution time.

PR #4215, opened 2022-05-10
Update/fix RemoteConfig URL.

dbrant:
The URL for fetching our RemoteConfig parameters has been updated recently (or a long time ago, and we hadn't noticed?) and is now returning 404.
This is the new URL.


tonisevener (approved):
Thanks for catching this! 😅

PR #4216, opened 2022-05-10
Save identifiers of Notifications to display to `userInfo`

staykids:
**Phabricator:** 
- https://phabricator.wikimedia.org/T301127 (partially) 
- https://phabricator.wikimedia.org/T291019 (partially)

### Notes
This PR adds some architectural/clean up work to help support upcoming notifications refresh work and upcoming work related to identifying and clearing out the OS's Notification Center.

- Adds a "model" to represent the Echo notifier's `checkEchoV1` message
- Adds a lightweight model to persist notification keys and dates of notifications about to be displayed for later use in the app's Notifications Center 
- Only pushes a tapped notification to the Notifications Center if it meets the criteria of being a `threadIdentifier` identified Echo message

The strangeness about type choice (class vs. enum) for `EchoModelVersion` is purely for simplicity with Objective-C interoperability. I'm happy to explore some alternates if you can think of a less gross approach. I just wanted to safely namespace the identifier because it's being used more than one place now, but didn't want to over-engineer a solution for something that will likely not change in the near future.

### Test Steps
I typically disable the relevance score work in `NotificationService` in order for quickly trigger displayed notifications with our Push utility.

1. Lightly regression test push notifications to confirm you can a. still receive them and b. tapping them opens the Notifications Center in app. Our Push utility serves as a good tool to trigger cases of no notifications, 1 notification, and multiple notifications of multiple types to help confirm the service extension does not crash.
2. Confirm the received push notifications on device don't display "checkEchoV1" text or anything else out of the ordinary in the body.

### DIFF
--- a/NotificationServiceExtension/NotificationService.swift
+++ b/NotificationServiceExtension/NotificationService.swift
@@ -27,7 +27,7 @@ class NotificationService: UNNotificationServiceExtension {
         
         self.bestAttemptContent = bestAttemptContent
         
-        guard bestAttemptContent.body == "checkEchoV1" else {
+        guard bestAttemptContent.body == EchoModelVersion.current else {
             bestAttemptContent.body = fallbackPushContent
             contentHandler(bestAttemptContent)
             return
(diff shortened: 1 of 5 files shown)


tonisevener (approved):
Looks good to me!

PR #4218, opened 2022-05-11
Attempt to fix performBatchUpdates() crashes

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T253762

### Notes
This PR attempts to reduce the potential number of invalid collection view updates happening within CollectionViewUpdater by adding a little extra logic to force a `reloadData()` call. It tries to do this with two approaches:

d5d8857c5071aaf78d384878af719fd477c81b70 - I noticed `CollectionViewUpdater` [already](https://github.com/wikimedia/wikipedia-ios/blob/fd697a4b762c5d3a7b37c2893539d3bac8565b55/Wikipedia/Code/CollectionViewUpdater.swift#L126) tries to fall back to `reloadData` if the number of sections pulled from the fetchResultsController match the number of sections from the previous update, plus or minus the new section delta value. Here I thought I would also check against `collectionView.numberOfSections`, in case it's somehow changing when we don't expect.
61b7f6804994c49917095f811610020eabddd859 - I also noticed `ExploreViewController` sets `isGranularPushEnabled` flag within it's  `viewDidDisappear` and `viewDidAppear` methods, which also forces a `reloadData()` in `CollectionViewUpdater()` when it's disabled. To match this functionality I thought I would also set this flag when the app is backgrounded. There's no need for us to have insertion/deletion animations here if the app is backgrounded. 

### Test Steps
I couldn't reproduce this, so most of this testing will just be regression testing.
1. Load the Explore feed on a fresh install. Page back several days and be sure older dates animate in.
2. Background the app, change the device date to the next day, then foreground. Pull to refresh Explore. Be sure new date data animates in at the top. 
3. Try hiding some cards, be sure those animate out well.
4. Terminate the app. Change the device date to 1 week later. Relaunch the app. Be sure new data animates in at the top.

### DIFF
--- a/Wikipedia/Code/CollectionViewUpdater.swift
+++ b/Wikipedia/Code/CollectionViewUpdater.swift
@@ -124,19 +124,21 @@ class CollectionViewUpdater<T: NSFetchRequestResult>: NSObject, NSFetchedResults
         }
         
         let sectionCountsMatch = (previousSectionCounts.count + sectionDelta) == sectionCounts.count
-        guard !forceReload, sectionCountsMatch, objectChanges.count < 1000 && sectionChanges.count < 10 else { // reload data for invalid changes & larger changes
+        let currentNumberOfSections = collectionView.numberOfSections
+        let previousSectionCountsEqualCurrentNumberOfSections = previousSectionCounts.count == currentNumberOfSections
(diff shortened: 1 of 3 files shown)


mazevedofs (approved):
Looks good, no crashes from testing on the device

PR #4221, opened 2022-05-12
Notifications - fix importing instead of refreshing paging bug

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T308293

### Notes
I came across this bug when testing another PR. Occasionally our Notifications Center refresh calls would continue paging through to the end, as if it were importing. I tracked it down to the logic we use to pick the last notification received in a response. We use that last notification to check whether or not it's already persisted. If it's already persisted, we tell the refresh operation operation to [stop paging](https://github.com/wikimedia/wikipedia-ios/blob/44b930324d8fb0715102d361ff64beb87742b905/WMF%20Framework/Remote%20Notifications/Operations/RemoteNotificationsRefreshOperation.swift#L6).

Sometimes we unintentionally chose the cross wiki summary type of notification as a last notification, which we never persist, resulting in refresh calls to continue paging. (Sidenote: The cross wiki summary notification type is the object that tells us which other wikis have unread notifications to fetch from). This PR fixes that logic.

### Test Steps
1. Fresh install on a device, with an account that has > 50 notifications on its primary language. Watch the notifications calls as the app launches and you log in.
2. Notifications should automatically page upon initial launch, because it's importing.
3. Visit Notifications Center. Notifications should only fetch the first page, because it's refreshing.


staykids:
@tonisevener Since the bulk of my notifications are spread out over multiple accounts rather than centralized in one, my testing approach has been to change the `getAllNotifications` fetch limit from `max` to a low `numeric` (10) and update `fetchNotifications(fetchLimit...)` (to 10) in order to clearly see the paging and single page refresh attempts.

Does this seem like a reasonable approach to hit the code paths you're hoping or am I missing something you're expecting?

staykids (approved):
👍🏽 wasn't able to reproduce the original issue myself, but your explanation was helpful in helping me understand. Using your test steps and the changes I mentioned, was able to confirm I'm only seeing the single page refresh per project when opening the Notifications Center rather than full paging imports.

PR #4224, opened 2022-05-16
Prevent SettingsViewController from instantiating more than once

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T303334

### Notes
* This PR fixes a bug where the Settings View Controller was being instantiated twice, preventing the app from setting the theme on the correct instance when updating the themes based on the trait collection

### Test Steps
1. On Settings, set the Reading Preferences to default
2. Turn the Explore feed Off
3. Put the app in the background and on the device, go to Settings -> Display & Brightness -> alternate between Light and Dark mode
4. Go back to the app, the Settings view should match the app theme
5. Turn the explore feed on and repeat the step 3


tonisevener (approved):
Working well for me!

PR #4225, opened 2022-05-17
Separate filter types from Remote Notification types

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T308503

### Notes
* This PR separates the logic for RemoteNotificationType from RemoteNotificationFilterType, so the filters on the Notifications Center are not tied to Remote Notifications types

### Test Steps
1. Basic regression testing on the Notifications Center filter view. Filters should work as always


tonisevener (approved):
Working great! I pushed one minor casing fix in a variable name, but that's all I found.

PR #4226, opened 2022-05-17
fix bug from iOS overlay

mcleinman:
**Phabricator:** https://phabricator.wikimedia.org/T305112

### Notes
* When a user was in our app, selected a word/phrase, then tapped "Look Up" (and entered the iOS overlay), then tapped the "Siri Knowledge" cell (if available), then tapped "Wikipedia": iOS's overlay would open the link in the app, but currentNavigationController would present the new article VC on the iOS Look Up VC, not the original article VC (as expected). Since the iOS Look Up VC would deinit nearly immediately, the new article would close after a split second.
* I tested this on iPad and iOS

### Test Steps (please test on iPhone and iPad)
1. In our app, select a word/phrase
2. then tap "Look Up" (and enter the iOS overlay)
3. then tap the "Siri Knowledge" cell (if available - it should be for most places/concepts/etc)
4. then tap "Wikipedia"
5. Confirm: the link opens in the app, and the new article is showing as expected
6. Confirm: The back button goes to the prior article


tonisevener (approved):
Works well for me!

PR #4232, opened 2022-05-24
adding selected text to share sheet

mcleinman:
**Phabricator:** https://phabricator.wikimedia.org/T277351

### Notes
* I also improved this code in a few different ways. 
* One non-intuitive tweak: `shareMenuItemTapped` didn't need `webView.wmf_getSelectedText` because `shareArticle()` will call that when it does not get passed any text.
* I also fixed a bug: When *selecting text* and tapping "share", the "Add to reading list" option is *not* our app's, but iOS's (and goes to Safari?). I've rewritten the share code to always include our reading list option (and suppress iOS's default implementation), in addition to Share-a-fact (if available). Furthermore, I believe I've suppressed the iOS system "add to reading list" everywhere in the app - while we had been doing this in some places, we hadn't been doing it universally.

### Test Steps
1. Share things via all methods (peek and pop context menu, selected text context menu, article tool bar) both with text selected and not selected. Ensure it pastes as expected. 
2. Test share a fact card and "add to reading list" and ensure they still work.


mcleinman:
This breaks "add to reading list" via share sheet. Putting this on hold until I fix that.

mcleinman:
Taken off hold, ready for review - it turns out I hadn't broken "add to reading list" via share sheet, it was an existing bug. I fixed it as part of this PR.

tonisevener (approved):
Looks good! Great catch on the confusing Safari Add to Reading List option.

I found one bug with saving an article via the share sheet, but it looks like it's existing too. I created a bug for it so we can tackle it separately - [T311238](https://phabricator.wikimedia.org/T311238).

PR #4244, opened 2022-05-31
Linter: Automatically address rules (3 of 3)

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T308566

### Notes
This PR resolves SwiftLint rules that can be automatically corrected by SwiftLint. While the number of line changes is large, it's best to go commit by commit as you'll notice each commit are changes of the exact same type.

### Test Steps
Confirm the app targets build with no SwiftLint errors/warnings.


tonisevener (commented):
@staykids Looks good! I think the last thing we should do is update the README.md to mention SwiftLint, since ClangFormat is mentioned multiple times there. You're welcome to do it in a followup PR or add it to this PR. Thanks!

staykids:
> @staykids Looks good! I think the last thing we should do is update the README.md to mention SwiftLint, since ClangFormat is mentioned multiple times there. You're welcome to do it in a followup PR or add it to this PR. Thanks!

👍🏽 I've got that one sitting in my TODO list – I'll add it in a follow-up PR.

tonisevener (approved):

PR #4257, opened 2022-06-22
Navigation bar documentation and some code tweaks

mcleinman:
**Phabricator:** N/A

### Notes
* This started as documenting code, but I also did a couple code tweaks. While most/some swift code can be self-documenting, this class has long been a thorn in our side, and I figured it could use some more clarity.
* IMO, the custom nav bar has a lot of features, but it's not clear what they all are. I tried to pull all public properties to a long 50+ line block of comments at the top of the class, and tried to organize them in a logical way.
* I've explained what most of the properties are, but I don't actually know about some of them and left those ones blank.
* Please let me know if there are parts of this I could write more clearly. Or if there are parts of this class I should document more.
* Documentation is a journey. Please feel free to continue this work when digging into this class.

### Test Steps
1. Run app, ensure it compiles.


mcleinman:
I'm going to add comments to NavigationBarHider tomorrow - putting this on hold until I add those.

WMF Framework/NavigationBar.swift
```diff
@@ -436,7 +493,6 @@ public class NavigationBar: SetupView, FakeProgressReceiving, FakeProgressDelega
     }
     
     @objc dynamic public var visibleHeight: CGFloat = 0
-    @objc dynamic public var insetTop: CGFloat = 0
```
mcleinman, line 439: This variable was written to in the class, but was never read.

Wikipedia/Code/NavigationBarHider.swift
```diff
@@ -33,8 +34,8 @@ public class NavigationBarHider: NSObject {
     fileprivate var isUserScrolling: Bool = false
     fileprivate var isScrollingToTop: Bool = false
     fileprivate var isProgramaticallyScrolling: Bool = false
-    var initialScrollY: CGFloat = 0
-    var initialNavigationBarPercentHidden: CGFloat = 0
+    fileprivate var initialScrollY: CGFloat = 0
+    fileprivate var initialNavigationBarPercentHidden: CGFloat = 0
```
mcleinman, line 38: For consistency

mcleinman:
I believe this is ready for review.

tonisevener (approved):
Thanks for doing this!

PR #4259, opened 2022-06-23
Fetch data from new Talk Page API

mazevedofs:

### Notes
* This PR adds a class to fetch data from the new mobile Talk Page API
* It also renames the existing TalkPage API-related files

### Test Steps
1. You can add a call to this class from the Explore feed to trigger this method
2. check with article talk pages and user talk pages; check if the response does not throw any errors and parses correctly


Wikipedia/Code/TalkPageFetcher.swift
```diff
@@ -1,228 +1,55 @@
...
+        performDecodableMediaWikiAPIGET(for: url, with: params) { (result: Result<TalkPageAPIResponse?, Error>) in
+            switch result {
+            case let .success(talk):
+                completion(.success(talk?.threads))
                 
-                for (replyIndex, reply) in topic.replies.enumerated() {
-                    reply.sort = replyIndex
-                }
+            case .failure:
+                completion(.failure(RequestError.invalidParameters))
```
tonisevener, line 50: In these fetcher classes I like to send up the error in case we want to react to it at a higher level (for example, if it's a network failure, we'll send it back and whoever is calling it can check for that particular error, and fall back to some persisted object for offline mode).

```
            case .failure(let error):
                completion(.failure(error))
```

Wikipedia/Code/TalkPageFetcher.swift
```diff
@@ -1,228 +1,55 @@
...
-        assert(title.contains(":"), "Title must already be prefixed with namespace.")
-        return title.denormalizedPageTitle
+        case threadItems = "threaditemshtml"
     }
 }
 
-enum TalkPageFetcherError: Error {
-    case talkPageDoesNotExist
+struct TalkPageItem: Codable {
+    let type: String?
```
tonisevener, line 21: Let's try removing some optionals in this struct.

1. Turn type into a non-optional String enum, with the raw values "comment" or "heading".
2. Let's remove `level` property entirely for now until we're sure we need it. As of now it just corresponds to the nesting level, so I'm not sure that we need it yet. It's also easy to confuse it with headingLevel (which I think of as more of a heading font size) so it would help to get rid of this one if we can.
3. Make `replies` non-optional.
4. It looks like the API sometimes spits out an `othercontent` String property here, so let's add an optional property for that. [Here's](https://test.wikipedia.org/w/api.php?action=discussiontoolspageinfo&format=json&page=Talk:Bird2&prop=threaditemshtml&formatversion=2) an example of one that does that. It's a more recent addition.

Unfortunately it seems like the other stuff could be optional (even `id`) from some of my API testing :(

tonisevener (changes requested):
@mazevedofs Nice work! It's working well, just some comments mainly around trying to reduce the number of optionals.

mazevedofs:
@tonisevener I think it's ready for another look. Let me know if the error propagation in these calls is making sense

tonisevener:
@mazevedofs it's looking good! Going to do a last little bit of testing / response documentation digging tomorrow, then I'll approve.

Wikipedia/Code/TalkPageFetcher.swift
```diff
@@ -1,227 +1,68 @@
...
-            }
-            
-            completion(.success(result))
-        }
-    }
-    
-    func addReply(to topic: TalkPageTopic, title: String, siteURL: URL, body: String, completion: @escaping (Result<[AnyHashable : Any], Error>) -> Void) {
-        
-        guard let url = postURL(for: title, siteURL: siteURL) else {
+    func fetchTalkPageContent(url: URL, completion: @escaping (Result<[TalkPageItem], Error>) -> Void) {
```
tonisevener, line 44: Since we're moving towards the empty view controller taking in a `siteURL` and `talkPageTitle` in the feature flag PR, change this url into two params : `talkPageTitle: String, siteURL: URL`. Then you can remove this guard and use them when building the request below:

```
...
        let params = ["action" : "discussiontoolspageinfo",
                      "page" : talkPageTitle.denormalizedPageTitle,
                      "format": "json",
                      "prop" : "threaditemshtml",
                      "fomatversion" : "2"
        ]

        performDecodableMediaWikiAPIGET(for: siteURL, with: params) { (result: Result<TalkPageAPIResponse, Error>) in
...
```

`denormalizedPageTitle` is just our helper extension to replace spaces with an underscore - that's how the API wants titles.

Wikipedia/Code/TalkPageFetcher.swift
```diff
@@ -1,227 +1,68 @@
...
-import WMF
+struct TalkPageItem: Codable {
+    let type: TalkPageItemType
+    let level: Int?
+    let id: String?
+    let html: String?
+    let headingLevel: Int?
+    let placeholderHeading: Bool?
+    let replies: [TalkPageItem]
+    let otherContent: String?
```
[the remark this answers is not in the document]
tonisevener: Update - they answered quickly. Looks like we can make `id` non-optional, and remove `placeholderHeading`. Thanks!

tonisevener (approved):
Looks good! I pushed one minor fix to make `name` optional again. Even though it should never be null, I think there's still a chance it could not exist at all for comment types, whereas `id` should always be there. We can keep an eye on the endpoint and change it later if we need to - it looks like those changes haven't been deployed in the API yet.

PR #4262, opened 2022-06-27
Add feature flag for new talk pages work

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T311313

### Notes
* This PR creates a hard-coded feature flag triggered on our staging build to test the new Talk Pages work
* It also adds a new blank TalkPageViewController 
* It changes the fastfile to deploy both production and staging builds

### Test Steps
1. Run the app on the staging target
2. Go to a user talk page
3. Verify that the blank view controller is pushed
4. Build the main build, go to a user talk page
5. Verify that the current talk page shows
6. Check if the nightly build triggers the staging build correctly


Wikipedia/Code/AccountViewController.swift
```diff
@@ -122,7 +122,13 @@ class AccountViewController: SubSettingsViewController {
...
                 let loadingFlowController = TalkPageContainerViewController.talkPageContainer(title: title, siteURL: siteURL,  type: .user, dataStore: dataStore, theme: theme)
                 
-                self.navigationController?.pushViewController(loadingFlowController, animated: true)
+                let newTalkPage = TalkPageViewController(theme: theme)
+                if FeatureFlags.needsNewTalkPage {
+                    self.navigationController?.pushViewController(newTalkPage, animated: true)
+                } else {
+                    self.navigationController?.pushViewController(loadingFlowController, animated: true)
+                }
+
```
tonisevener, line 131: We should only instantiate the appropriate view controllers within the conditionals, so we're not spinning up a view controller that we'll never display:
```
if FeatureFlags.needsNewTalkPage {
    let newTalkPage = TalkPageViewController(theme: theme)
    self.navigationController?.pushViewController(newTalkPage, animated: true)
} else {
    let oldTalkPage = TalkPageContainerViewController.talkPageContainer(title: title, siteURL: siteURL,  type: .user, dataStore: dataStore, theme: theme)
    self.navigationController?.pushViewController(oldTalkPage, animated: true)
}
```

Wikipedia/Code/ViewControllerRouter.swift
```diff
@@ -94,7 +94,13 @@ class ViewControllerRouter: NSObject {
...
                 return false
             }
-            return presentOrPush(talkPageVC, with: completion)
+            let newTalkPage = TalkPageViewController(theme: theme)
+            if FeatureFlags.needsNewTalkPage {
+                return presentOrPush(newTalkPage, with: completion)
+            } else {
+                return presentOrPush(talkPageVC, with: completion)
+            }
+
```
tonisevener, line 103: This whole case can now become something like:

```
            if FeatureFlags.needsNewTalkPage {
                guard let newTalkPage = TalkPageViewController(url: linkURL, theme: theme) else {
                    completion()
                    return false
                }
                return presentOrPush(newTalkPage, with: completion)
            } else {
                guard let oldTalkPage = TalkPageContainerViewController.userTalkPageContainer(url: linkURL, dataStore: appViewController.dataStore, theme: theme) else {
                    completion()
                    return false
                }
                return presentOrPush(oldTalkPage, with: completion)
            }
```

tonisevener (changes requested):
Looking good! I think while we're here let's give your new view controller properties it will need for your fetcher, so mainly just some comments around that.

tonisevener (approved):

PR #4263, opened 2022-06-27
When saving to specific reading list, ensure articles are also added to default reading list

mcleinman:
**Phabricator:** https://phabricator.wikimedia.org/T311238

### Notes
* Fairly straightforward fix - thanks to lots of delegates, this spot should catch it for all routes to saving an article to a reading list.

### Test Steps
1. On Explore feed, in an article, or anywhere else, use share sheet to add an unsaved article to a specific reading list.
2. Check the main "Saved articles" list, ensure articles is on this AND on the specific reading list selected.


staykids (approved):
Nice – appears to be working well on my end. Thanks for this fix!

PR #4266, opened 2022-06-30
Add feature flag logic for the article talk page

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T311313

### Notes
I forgot about the article talk page portion of this! This adds a new check in our routing logic for article talk pages, which we didn't have before. 

I also adjusted what endpoints our Staging scheme points to, so that it mostly hits production. I think this will make development go more smoothly, because the production articles and talk pages will have more robust data to work with. `.deploymentLabsForEventLogging` just means for our legacy events in the app will post to development endpoints, everything else looks at production. This was just the lightest weight staging configuration I could set up - it won't hurt anything setting it to this just for the Staging app. Now you can use your usual main app credentials to log in for the Staging scheme.

### Test Steps
1. Switch to Staging scheme
2. Go to an article and scroll down and tap "View talk page"
3. App should navigate to our new talk page view controller.
4. Run the app in the Main scheme
5. Go to an article and scroll down and tap "View talk page"
6. App should navigate to an in-app web view, showing the mobile view talk page.

### DIFF
--- a/WMF Framework/Configuration.swift
+++ b/WMF Framework/Configuration.swift
@@ -50,7 +50,7 @@ public class Configuration: NSObject {
 			.deploymentLabsForEventLogging = labs instance for testing event logging endpoints
 			All other endpoints would point to production */
 		
-        return Configuration.staging(options: [.betaCluster])
+        return Configuration.staging(options: [.deploymentLabsForEventLogging])
         #else
         return .production
         #endif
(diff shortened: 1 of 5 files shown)


mazevedofs (approved):
Works great! Thanks for catching this. There's just a project file conflict and we can merge it

PR #4267, opened 2022-07-01
Fix bug where keyboard popped up when opening app

mcleinman:
**Phabricator:** https://phabricator.wikimedia.org/T310798

### Notes
* `shouldShowExploreScreenOnLaunch` didn't consider whether the user wanted to always launch into the search tab - so it was opening to Search, then switching to Settings and causing the bug

### Test Steps
1. set `WMFTimeBeforeShowingExploreScreenOnLaunch` to 10 seconds.
2. Turn off Explore feed (so first tab is Settings), and set option to always open on Search screen.
3. Close app, wait 10 seconds, re-open app, ensure bug no longer exists.

(Run steps 1-3 on main branch to see bug in action.)


tonisevener (approved):
Working well - thank you for finding and fixing this @mcleinman!

PR #4271, opened 2022-07-06
Add methods to post topic and post reply to new Talk Page API

mazevedofs:

### Notes
* This PR adds two methods for posting topics and posting replies to talk pages 

### Test Steps
1. Run the app on staging 
2. Add a call to the methods to our new [TalkPageViewController](https://github.com/wikimedia/wikipedia-ios/blob/4b927a509a147ec02ae323bf728746620a818089/Wikipedia/Code/TalkPageViewController.swift#L31)
3. Run and check on the web if it was successful (a topic or reply was posted)


tonisevener (changes requested):
It's working great! Just a few minor requests.

mazevedofs:
@tonisevener Hi Toni, I guess it's ready for another look!

Wikipedia/Code/TalkPageFetcher.swift
```diff
@@ -115,19 +104,28 @@ class TalkPageFetcher: Fetcher {
...
+            return
+        }
+        
+        if let resultError = result?["error"] as? [String: Any],
+           let info = resultError["info"] as? String {
+            completion(.failure(RequestError.api(info)))
+            return
+        }
+        
+        if let resultSuccess = result?["discussiontoolsedit"] as? [String: Any] {
```
tonisevener, line 123: I think it'd still be good to check for the "success" value here, like you were before. Also I don't think we'll need to pass back / access the dictionary afterwards, just a basic success flag should be enough. You can use `Void` in Swift's `Result` success value to avoid needing to pass something back:

```
fileprivate func performTalkPagePost(_ error: Error?, _ result: [String : Any]?, completion: @escaping(Result<Void, Error>) -> Void) {
```

Then here it could be something like (using `guard` because I think that's a bit more clear):

```
        guard let discussionToolsEdit = result?["discussiontoolsedit"] as? [String: Any],
              let discussionToolsEditResult = discussionToolsEdit["result"] as? String,
              discussionToolsEditResult == "success" else {
                completion(.failure(RequestError.unexpectedResponse))
            return
        }
        
        completion(.success(()))
```

Wikipedia/Code/TalkPageFetcher.swift
```diff
@@ -115,19 +104,28 @@ class TalkPageFetcher: Fetcher {
...
-            }
-            
-            if let resultSuccess = result?["success"] as? [String: Any] {
-                completion(.success(resultSuccess))
-            }
+            self.performTalkPagePost(error, result, completion: completion)
         }
     }
+    
+    fileprivate func performTalkPagePost(_ error: Error?, _ result: [String : Any]?, completion: @escaping(Result<[AnyHashable: Any], Error>) -> Void) {
```
tonisevener, line 111: Maybe rename this to something like `evaluateResponse(_ error: Error?, _ result: [String : Any]?...)`? `performTalkPagePost` makes me think this method will make another POST call.

tonisevener (changes requested):
Thanks! Looking good, just a couple of minor cleanup things.

PR #4273, opened 2022-07-11
Add subscription methods to TalkPageFetcher

mazevedofs:

### Notes
* This PR adds methods for subscribing/unsubscribing to topics and getting all the topics currently subscribed

### Test Steps
1. Run the app on staging
2. Add a call to the methods to our new [TalkPageViewController](https://github.com/wikimedia/wikipedia-ios/blob/4b927a509a147ec02ae323bf728746620a818089/Wikipedia/Code/TalkPageViewController.swift#L31)
3. Run and check on the web if it was successful (subscribing/unsubscribing and getting all the current subscribed topics)


Wikipedia/Code/TalkPageFetcher.swift
```diff
@@ -64,5 +64,79 @@ class TalkPageFetcher: Fetcher {
...
+            
+            if let resultError = result?["error"] as? [String: Any],
+               let info = resultError["info"] as? String {
+                completion(.failure(RequestError.api(info)))
+                return
+            }
+            
+            if let resultSuccess = result?["subscriptions"] as? [String: Any] {
+                var subscribedTopics = [String]()
+                for (topicId, _) in resultSuccess {
```
tonisevener, line 131: I came across a situation with the API where the response was this:

```
{
  "subscriptions": {
    "h-Tsevener-20220720150400": 0,
    "h-TSevener_(WMF)-2022-06-24T18:39:00.000Z": 1
  }
}
```

I had to first make the **unsubscribe** call to `h-Tsevener-20220720150400`, then it started returning with a `0` value in this response. So I think you should check the dictionary value here as well to make sure it's a `1` value, before appending to `subscribedTopics`.

tonisevener (changes requested):
Looks good! Just one small tweak is needed I think after doing some API testing.

Wikipedia/Code/TalkPageFetcher.swift
```diff
@@ -64,5 +64,80 @@ class TalkPageFetcher: Fetcher {
             }
         }
     }
-
+    
+    func subscribeToTopic(talkPageTitle: String, siteURL: URL, topic: String, shouldSubscribe: Bool, completion: @escaping (Result<Bool, Error>) -> Void) {
```
tonisevener, line 68: Can you add a comment before this method and the next one, explaining that `topic: String` expects the `name` value from `TalkPageItem` `heading` types? I mistakenly tested with the `id` values and I think a comment may help.

Wikipedia/Code/TalkPageFetcher.swift
```diff
@@ -64,5 +64,80 @@ class TalkPageFetcher: Fetcher {
...
+            if let resultSuccess = result?["discussiontoolssubscribe"] as? [String: Any],
+                let didSubscribe = resultSuccess["subscribe"] as? Bool {
+                completion(.success(didSubscribe))
+                return
+            }
+            completion(.failure(RequestError.unexpectedResponse))
+        }
+    }
+    
+    func getAllSubscriptions(siteURL: URL, topics: [String], completion: @escaping (Result<[String], Error>) -> Void) {
```
tonisevener, line 107: Actually now that we know we'll only be sending back the subscribed topics, maybe we should rename this to something like:

```
func getSubscribedTopics(from topics: [String], siteURL: URL, completion: @escaping (Result<[String], Error>) -> Void) {
```

tonisevener (changes requested):
Working well, just a couple of very minor naming things, thanks!

tonisevener (approved):
Looks good! Can you fix the merge conflicts?

mazevedofs:
> Looks good! Can you fix the merge conflicts?

Fixed!

PR #4274, opened 2022-07-19
Rename more old Talk Pages classes

staykids:
**Phabricator:** N/A

### Notes
In order to prevent namespace clashes with new incoming Talk Pages work, rename more old Talk Pages classes and delegates appending the `Old` prefix already established.

### Test Steps
1. Confirm Talk Pages still load
2. Confirm topics/sections are tappable and load when tapped
3. Confirm tapping reply shows the reply cell


tonisevener (approved):
Looks good!

PR #4280, opened 2022-07-26
Apple Pay foundational UI work

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T288285

### Notes
This is some foundational UI work for Apple Pay. It adds the Settings option and pushes on a view controller with an embedded SwiftUI view. For now I have it behind a feature flag, which I'll remove once this is further along.

Originally I added a wrapper class that subclassed `ViewController` to see if I could have a hybrid SwiftUI screen with our custom UIKit navigation bar, but it was too buggy on iOS 14. I'm keeping it in the PR history because it may be useful to reference later, but note that commit 18ec4e5 basically undoes 12ab203 and e91683b if you're reviewing commit-by-commit.

### Test Steps
1. Run app on main scheme, confirm Settings options look the same (only a link to web).
2. Run app on Staging scheme, confirm Settings now has an additional link to Apple Pay (a better icon will come in a future PR).
3. Confirm scratch Apple Pay UI acts okay on various devices and iOS versions. It scrolls when it gets too long. Textfield can focus and unfocus. When textfield focuses, a Done button appears for keyboard dismissal (this will be needed when we have a numberpad keyboard). Textfield is not covered up by the keyboard when focused (except for on iOS 13, which I marked as a TODO).


tonisevener:
Going to close this one for now - I'll reopen when Apple Pay work comes back.

PR #4282, opened 2022-07-29
Fix notifications links to talk page topics with non-Latin titles

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T307604

### Notes
Notifications talk page links sometimes come to us in this form:

```
//es.wikipedia.org/wiki/Usuario_discusi%C3%B3n:Tsevener?markasread=45718198&markasreadwiki=eswiki#આભાર
```

It has an encoded path, but an unencoded fragment. This url fails to instantiate in Swift, and after some difficulty getting this changed within the API, we decided to do it manually here.

I also decided to remove a comment while I was digging around in here (fe78b13). I feel like this comment was too confusing now that we have the Other filter handling. It was an old comment from back before we added the Other filter. Happy to revert if you disagree.

### Test Steps
1. Trigger a notification that has one of these links. Try sending yourself a user talk page topic with a title containing non-Latin characters. ([Here's](https://es.wikipedia.org/wiki/Usuario_discusi%C3%B3n:Tsevener) a talk page for inspiration).
2. On the notification detail screen, as well as the swipe More menu from the cell, go to the talk page.
3. Ensure it deep links to the correct topic reply thread. Currently on `main` it does not and stops at the topic list screen. This is the bug.


staykids (approved):
Nice! Much more favorable than that potential approach of completely changing our URL handling with that third party framework 😅

I've no clue if this change may negatively affect some specific combination/set of languages I haven't tested, but so far testing non-latin and mixed character sets seems to be working as expected taking the user directly to the new topic.

PR #4286, opened 2022-08-11

translatewiki:
Translation updates


tonisevener (approved):
Had to do some additional work to get the new Lombard language translations imported into the project. I updated the documentation to reflect what I did.

I still don't see a way to test this. I do not see this language as an option for me in iOS Settings. But if they ever do add it, these translations should now be set up to display.

staykids:
@tonisevener Did you happen to see the App Store build warning email from this morning? It looks like it's related to this change.

PR #4287, opened 2022-08-11
Add new header view for article and user talk pages

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T310273

### Notes
Adds article and user talk page header to the new talk pages view controller. This single new header supports all the configurations I'm aware of for both article and user talk pages. 

- Need to actually populate this with real data once the middle layer talk pages data controller object is constructed
- Once the color style guide is finalized (https://phabricator.wikimedia.org/T299793), need to update the coffee roll background color to use the appropriate theme element.

### Test Steps
1. On a staging build, access the new talk pages view controller.
2. In `TalkPageViewModel`, change the `// TODO` sample data in order to play with the different configuration permutations. See the Phabricator ticket for some possible states.


Wikipedia/Code/TalkPageViewController.swift
```diff
@@ -35,9 +36,33 @@ class TalkPageViewController: ViewController {
...
+    }
+
+    private func setupHeaderView() {
+        let headerView = TalkPageHeaderView()
+        self.headerView = headerView
+
+        headerView.configure(viewModel: viewModel)
+        navigationBar.isBarHidingEnabled = false
+        navigationBar.isUnderBarViewHidingEnabled = true
+        navigationBar.allowsUnderbarHitsFallThrough = true
```
tonisevener, line 53: Looks like you'll still want to override `override func point(inside point: CGPoint, with event: UIEvent?) -> Bool {` in the `TalkPageHeaderView`, and return true if they tapped inside anything tappable (I think just the coffee roll Read More button?), and otherwise return false.

You can reference the `OldTalkPageHeaderView` to see what we did there.

tonisevener (commented):
Looking great! Just one thing with the fallthrough tap handling, then this will be good to merge.

Wikipedia/Code/TalkPageViewController.swift
staykids: Great catch - thanks! Updated.

tonisevener (approved):
Nice work!

PR #4288, opened 2022-08-13
Talk Pages - Reply Compose Overlay (Part 1)

tonisevener:
**Phabricator:** 

### Notes
This is the beginning of the talk page reply compose overlay. It presents an empty view over the keyboard, set to 1/5 height of the parent view, and expands to nearly full screen in vertical compact trait collections.

I tried to keep as much as I could out of the view controller just so we can try to keep that class size small, so most of this is managed in a separate controller class. I'm fine moving the logic back into the view controller or into the view model if it doesn't make sense here, though.

### Test Steps
1. Load a talk page and tap a cell.
2. Confirm you see an empty reply compose view appear at the bottom in 1/5 the parent view height.
3. Tap empty reply compose view, keyboard should appear. Confirm reply compose view moves up with the keyboard.
4. Confirm you can still scroll the cells behind it per requirements. Confirm you can scroll all the way to the bottom and fully see the last cell.
5. Rotate to iPhone landscape, confirm it expands to full screen, over the navigation bar per mocks.
6. Confirm rotating back goes back to 1/5 height.
7. Test on iPad, confirm rotating works as expected, they should all keep the 1/5 height.
8. Confirm no autolayout complaints appear in the console.

Draggable handle and the rest of the reply compose content view will come in a follow up PR.


mazevedofs (approved):
Looking great 🤩

PR #4291, opened 2022-08-19
Talk pages - Populate header with live data (Part 1)

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T315330

### Notes
This PR builds a data controller layer for TalkPageViewModel to pull its data from. So far it is performing two parallel calls to fetch talk page content - one to fetch the article summary for populating the header UI, and another to get the topic and thread content from TalkPageFetcher. The other methods in this data layer are just passthrough methods to the other TalkPageFetcher methods.

One big TODO item on here is populating the project icon. It required a larger refactor so I will put that up in a follow up PR once this is merged.

The commits here are pretty linear, so it may be easier to review commit-by-commit.

### Test Steps
Load a few article talk pages and user talk pages. Confirm the header looks good.

### Screenshots/Videos
Screen Shot 2022-08-19 at 4 25 40 PM
Screen Shot 2022-08-19 at 4 26 23 PM


mazevedofs (approved):
Looking great!

PR #4292, opened 2022-08-19
Add initial talk page topic cell UI elements

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T311637

### Notes
Adds the initial implementation of the UI elements for the topic cell. There's a lot here I'm currently unhappy with, but I hope this is sufficient to unblock hooking up live data and more navigation. This work presents the basic UI hierarchy of the topic cells with the ability to expand and collapse the cells to show replies. They should support all our different themes and display fonts that match our designs.

Some unfinished elements include: 
- The reply depth view (just a temporary label for now)
- Refactoring a lot of the variable names/hierarchy/removing redundancy
- Removing unnecessary UI elements 
- Dialing in the view model hierarchy 
- Getting the thread animation to feel nice
- Adding cell tapping for expanding/collapsing
- RTL and accessibility support
- Refactoring the configuration of `TalkPageCell`'s to avoid adding/removing the comment views on every configuration

The biggest unknown to me (and worth exploring while I'm out if there's interest) is how to fine tune the topic cell expanding/collapsing animation. I feel pretty confident that there's a known path forward to all the other stuff though.


tonisevener (approved):
Thanks @staykids! I pushed a minor commit disabling animations for now - we can tackle that in follow up PRs.

> The biggest unknown to me (and worth exploring while I'm out if there's interest) is how to fine tune the topic cell expanding/collapsing animation.

Yeah, agreed on this part being a big unknown. I think it's worth prioritizing how to do this next, just in case we determine we need to change the way we do these cells.

I explored expanding & collapsing a bit this week - I'll put up a PR of something to consider.

PR #4296, opened 2022-08-26
Remove duplicated strings, add more information on parameters 

mazevedofs:
**Phabricator:**   No phab ticket for this

### Notes
* Removing some duplicate strings, moving identical strings to `CommonStrings`, adding more info on parameters that do not need translation

### Test Steps
1. Duplicated strings should be removed from the translation tool and the documentation on `qqq` should be updated.
2. No changes to the text that is used on the vanishing processes views.


Wikipedia/Code/VanishAccountPopUpAlertView.swift
```diff
@@ -11,10 +11,6 @@ struct VanishAccountPopUpAlertView: View {
     
     enum LocalizedStrings {
         static let title = WMFLocalizedString("vanish-modal-title", value: "Vanishing request", comment: "Title text fot the vanish request modal")
-        static let firstItem = WMFLocalizedString("vanish-modal-item", value: "If you completed your vanishing request, please allow a couple of days for the request to be processed by an administrator.", comment: "Text indicating that the process of vanishing might take days to be completed")
-        static let secondItem = WMFLocalizedString("vanish-modal-item-2", value: "If you are unsure if your request went through please check your Mail app", comment: "Text indicating that the user should check if their email was sent in the Mail app used to send the message")
```
tonisevener, line 15: @mazevedofs Are these needed?

mazevedofs: No, I moved them to the ListView struct and they were duplicated as well

tonisevener: Nevermind! I didn't look close enough.

tonisevener (approved):

PR #4297, opened 2022-08-26
Replace back button with arrow for Vanishing screen

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T300080#8190210

### Test Steps
1. Run app, make sure you now only see a back arrow instead of the word "Back".


mazevedofs (approved):
Looking great!

PR #4299, opened 2022-08-29
TalkPageCell expand / collapse animation

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T310663

### Notes
This is an approach to consider for a smooth expand and collapse animation for our talk page cells. I tried a few different things before resorting to this, but I think the hitches came down to these issues:
1. The cell content stack view being pinned to the cell's content view bottom. It seemed like during animation, the stack view elements would exist, pinned to the bottom, but the cell size would not yet be the full height. This caused the stack view to squish its elements during animation and look bad. 
2. Dynamically sized cells. I noticed that the height would always seem to briefly resort back to the layout's estimated height, before the new proper height was calculated. On my smaller iPhone 7 device upon expand, the cell would briefly _shrink_ to be even smaller, then expand to its full height.

I did manage to account for 1 & 2 to smooth over the those quirks while keeping the dynamically sized heights (this work is in the `talk-pages/cell-design-animations-2` branch if interested) , but it was still jerky during the animation itself, and sizing still sometimes goes haywire. I'm not sure the reasons for that, but figuring out the cell heights up front seemed to fix it. This approach may come at a performance cost when expanding all the cells from the overflow menu, or rotating. We may be able to improve some of that with some simple caching of the heights.

This approach is similar to how we built the revision diff cells that expand, FYI. On that screen we are calculating the cell heights up front, but doing more manually by traversing each string in the view model and determining it's height via the string [boundingRect]((https://developer.apple.com/documentation/foundation/nsattributedstring/1529154-boundingrect) methods 😱. This approach with a sizing view at least lets us mostly lean on the autolayout setup of the root container view.

There are a couple of bugs here that I still need to look into - there are layout console complaints upon rotation (I'm probably missing an inset consideration in the cell size calculations), and iPad seems to have more of a margin that gets lost upon expand due to the width I set for the cell heights. Both seem fixable though.

### DIFF
--- a/Wikipedia/Code/TalkPageCell.swift
+++ b/Wikipedia/Code/TalkPageCell.swift
@@ -27,7 +27,6 @@ final class TalkPageCellRootContainerView: SetupView, Themeable {
         
         NSLayoutConstraint.activate([
             stackView.topAnchor.constraint(equalTo: topAnchor, constant: Self.padding.top),
-            stackView.bottomAnchor.constraint(equalTo: bottomAnchor, constant: -Self.padding.bottom),
             stackView.leadingAnchor.constraint(equalTo: leadingAnchor, constant: Self.padding.leading),
             stackView.trailingAnchor.constraint(equalTo: trailingAnchor, constant: -Self.padding.trailing)
         ])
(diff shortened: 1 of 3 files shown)


tonisevener:
Closing - we'll address this later in the release cycle if there's time.

PR #4300, opened 2022-08-29
Redirect users to meta-wiki for more info on vanishing accounts

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T316351

### Notes
* Removing confusing punctuation from the strings
* Redirecting users to https://meta.wikimedia.org/wiki/Right_to_vanish for more info about vanishing on Wikipedia and sister projects.

### Test Steps
1. On the vanish account view (settings -> account -> vanish account) make sure that the bottom link redirects to the above-mentioned meta-wiki page
2. On the modal, ensure the text is updated and there's no `.` after the link
3. It should also redirect to the Meta-wiki page


tonisevener (approved):
Looks great! FYI I noticed there were some automatically built strings changes when I tested, so I pushed them up.

PR #4301, opened 2022-08-30
Add `learn more` option for vanish page and modal on older iOS versions

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T300080 (follow up)

### Notes
* On iOS versions before iOS 15, markdown links don't work properly on SwiftUI. We are adding these buttons to allow users of these OS versions to have easy access to information on vanishing pages.

### Test Steps
1. Run the app on a simulator or device running iOS 13 or 14
2. Verify that on the vanishing page, under the "Send request" button, another button appears with the title "learn more"
3. Click the button and check that it redirects to the meta page on vanishing
4. Trigger the modal (click send button, then return to the app. On the simulator this step will not work) 
5. Verify that under the last bullet point, a "learn more" link appears, it should also redirect to the meta page.
6. Test the app on a device running iOS 15+, it should look as the figma mocks


Wikipedia/Code/VanishAccountPopUpAlertView.swift
```diff
@@ -47,17 +69,32 @@ struct VanishAccountPopUpAlertView: View {
                                     .frame(maxWidth: .infinity)
                                     .frame(height: 43, alignment: .center)
                                     .foregroundColor(Color(theme.colors.link))
-                                    .font(Font(titleFont))
+                                    .font(Font(boldFont))
                             }).buttonStyle(PlainButtonStyle())
+                                .frame(height: 43)
                         }
-                        .frame(width: 300)
+                        .frame(width: 300, height: 470)
```
tonisevener, line 76: I notice the text cuts off with dynamic type increased now on the popup. I suspect it might be this line, when you limit it to a particular height. This is a screenshot from iPhone 7, iOS 15:

IMG_0756

tonisevener (changes requested):
Thanks for the quick work! The buttons are working great and the spacing is much better. Just one thing with labels truncating on the popup - hopefully it's an easy fix.

tonisevener (approved):
Working great - thanks for fixing!

PR #4303, opened 2022-08-31
Talk pages - populate cells with live data

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T315330 (Partial)
https://phabricator.wikimedia.org/T310275 (Partial)
https://phabricator.wikimedia.org/T310279 (Partial)

### Notes
This connects the talk page cell view models with live data from the API. It transforms the array of `TalkPageItem`s into a flat structure, then instantiates `TalkPageCellViewModel`s off of that.

**Note:** As of now timestamps are blocked on the deployment of [T315400](https://phabricator.wikimedia.org/T315400) in MediaWiki. To get around this I pushed a commit marking the timestamp as optional in the view models.

### Test Steps
1. Load a talk page, confirm data generally looks okay besides missing timestamps and html tags in the labels. We will need to strip html out later by setting up NSAttributedStrings in the cell layer.


mazevedofs (approved):
Looking good!

PR #4304, opened 2022-09-01
Talk Pages - Reply Compose Overlay (Part 2)

tonisevener:
**Phabricator:** 

### Notes
This is part 2 of the reply compose overlay. It fills out the content view UI elements, as well as makes connections to the cell reply buttons.

There are still ~~a few~~ several remaining parts to this task. These will be done in later PRs:
- Filling out the reply text field placeholder, once live data to cells are hooked up (#4303)
- Publishing flow (publish reply text to API, dismiss reply view, and reload the talk page cells)
- Logged out confirmation flow
- Tapping another reply button while composing flow
- Fine print link handling.
- Accessibility tweaks.

### Test Steps
1. Tap a reply button on the new talk page cells. Confirm reply view appears. Confirm user can expand reply to full screen. Confirm user can type in text field.
2. Tap close button on reply compose view. Confirm reply view disappears.


mazevedofs (approved):
Looking great!

PR #4305, opened 2022-09-02
Add toolbar to new talk page

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T310288

### Notes
* This PR adds a toolbar to the new talk pages. Note that only the share button is active at this moment

### Test Steps
1. Run the Staging build
2. Go to the new talk pages
3. Only the share button has a working action at this moment


tonisevener (approved):
Great find on the automatic toolbar! I pushed a minor commit adjusting the selectors that get called for each button.

After pulling in the changes from `main`, there is now an extra bottom space when tapping the reply buttons. I hadn't considered the toolbar back when I originally added the content inset. We can fix that when we revisit replying for posting to the API.

PR #4308, opened 2022-09-07
Talk Pages - New topic screen (Part 1)

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T310285

### Notes
This provides the basics of the new topic screen. It includes:

1. Layout, with dynamic type support
2. Theming
3. Text input handling, including placeholders and the disabling Publish button logic
4. Roundtrip posting (publishing to the API, and reloading the talk page upon success)
5. Dismiss confirmation action sheet if user tries to dismiss while text is filled out.

It does not include:
1. The onboarding screen
2. Logged out confirmation / flow
3. Fine print link handling
4. Any POST failure error handling

### Test Steps
Lightly test the screen against the included list above. Note that reloading the topic list after publishing is a little jarring. We can smooth it out later.


Wikipedia/Code/TalkPageViewController.swift
```diff
@@ -339,6 +365,56 @@ extension TalkPageViewController: TalkPageReplyComposeDelegate {
...
+                    // TODO: Display success banner
+                }
+                
+                // Try to refresh page
+                self?.viewModel.fetchTalkPage { [weak self] result in
+                    switch result {
+                    case .success:
+                        self?.talkPageView.collectionView.reloadData()
+                        self?.scrollToLastTopic()
+                        self?.handleNewTopicAlert()
```
mazevedofs, line 386: For some reason, this banner does not show consistently when adding topics. I didn't notice any constraint errors on the console; it's not clear why it's behaving like this. I don't think it's a deal breaker for user testing, but it's worth looking at next.

mazevedofs (approved):
Looks great! The comment I left about the alert does not look like a blocker

Wikipedia/Code/TalkPageViewController.swift
tonisevener: Thanks! I was seeing that too, I'll work on it in a follow up PR.

PR #4310, opened 2022-09-08
Talk Pages - Handle cell links

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T310275

### Notes
This PR turns the talk page html (title, lead reply, other replies) into attributed strings, and routes any link taps within those strings. Since we are actively working on the sticks in these cells, I tried to keep the changes low-touch in case we run into merge conflicts. Any spacing issues, etc can be looked into once the sticks are in place.

### Test Steps
1. Load talk page topic cells & reply threads.
2. Confirm you now see links in topic cells and reply threads. Confirm links can be tapped and the appropriate new screen is pushed on.
3. Confirm you no longer see html tags in the topic cells.


Wikipedia/Code/TalkPageCellTopicView.swift
```diff
@@ -179,9 +192,15 @@ final class TalkPageCellTopicView: SetupView {
...
     func apply(theme: Theme) {
-        topicLabel.textColor = theme.colors.primaryText
+        
+        topicTitleTextView.attributedText = viewModel?.topicTitle.byAttributingHTML(with: .headline, boldWeight: .semibold, matching: traitCollection, color: theme.colors.primaryText, linkColor: theme.colors.link, handlingLists: false, handlingSuperSubscripts: true)
+        topicTitleTextView.backgroundColor = theme.colors.paperBackground
+        
         timestampLabel.textColor = theme.colors.secondaryText
-        topicCommentLabel.textColor = theme.colors.primaryText
+        
+        let commentColor = (viewModel?.isThreadExpanded ?? false) ? theme.colors.primaryText : theme.colors.secondaryText
```
staykids, line 201: Just a personal note, I understand it's as designed but it's kind of visually unexpected to see this change as the expand/collapse is toggled.

staykids (approved):
Awesome - so cool to see this coming together.

PR #4311, opened 2022-09-09
Talk Pages - Post reply to API

tonisevener:
**Phabricator:** 

### Notes
This PR implements the round trip POST to the API and reloading of the collection view upon success. It also properly populates the reply compose placeholder label.

Even with a simple `reloadData` for user testing, I'm running into issues with the cell automatic height estimation. `reloadData()` after the API call weirdly is causing the height to revert back to the estimated cell height and UI elements to appear squished again, similar to issues we ran into earlier with expand and collapse. I put in a hacky fix for now (after looking at [this thread](https://github.com/airbnb/MagazineLayout/issues/49#issuecomment-1022909541)), but it's not something I think should remain for v1, plus it has some jumpy offset issues. I'll continue to tinker with it, but I don't think we should rework things yet until the remaining user testing PRs are in. (**Update**: my work in #4299 improved things dramatically, so I think soon we'll be back to considering that solution.)

This PR is branched off of the new topic PR (#4308), and should not be merged until that PR is merged and the branch deleted.

### Test Steps
1. Tap a reply button. Confirm the placeholder text view represents the user you are replying to.
2. Enter text, tap publish. Upon successful post, the reply compose view should dismiss, and the list will reload with your new reply in place.


Wikipedia/Code/TalkPageViewController.swift
```diff
@@ -361,7 +373,30 @@ extension TalkPageViewController: TalkPageReplyComposeDelegate {
     }
     
     func tappedPublish(text: String, commentViewModel: TalkPageCellCommentViewModel) {
-        // TODO: Publish reply once live data is connected to commentViewModels
+        viewModel.postReply(commentId: commentViewModel.commentId, comment: text) { [weak self] result in
+
+            switch result {
+            case .success:
+                // TODO: Display success banner
```
mazevedofs, line 380: Is this comment still needed?

tonisevener: Nope, good catch! This is removed. I notice there's another one from new topic as well. I'll fix it in my followup PR.

mazevedofs (approved):
Looking great!

PR #4312, opened 2022-09-09
Add banner to subscription action

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T311639

### Notes
* This task adds a banner to the subscription action
* We are not calling the API at this moment (but methods were added to the view model)
* The layout differs from figma, as we are using an existing component at the moment

### Test Steps
1. Go to a talk page, and click the subscribe button. Make sure the alert shows at the bottom.
2. Click unsubscribe, verify the banner


Wikipedia/Code/TalkPageViewController.swift
```diff
@@ -97,21 +98,21 @@ class TalkPageViewController: ViewController {
...
         let talkPageView = TalkPageView(frame: UIScreen.main.bounds)
         view = talkPageView
         scrollView = talkPageView.collectionView
     }
-
+    
     override func viewDidLoad() {
         super.viewDidLoad()
 
-        navigationItem.title = WMFLocalizedString("talk-pages-view-title", value: "Talk", comment: "Title of user and article talk pages view.")
```
tonisevener, line 114: Thanks for cleaning these up!

tonisevener (approved):
Looking good! I didn't realize this view was so broken with bottom alignment and icons, thanks for working through it. I made a couple of minor tweaks in 966837e, hope that's okay:

1. There were some console constraint complaints, and after some digging it looks like there already was a leading constraint in the xib that your new `titleViewLeading` constraint conflicted with. So I exposed that xib constraint and am setting it to 50 instead.
2. I aligned some views to the superview `safeAreaLayoutGuide` instead of the extra safe area padding number. Both of these changes should only affect bottom and icon banners, so I think it's pretty safe.

PR #4313, opened 2022-09-09
[WIP] Add reply depth "sticks" for user testing

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T310279

### Notes
Adds the initial reply depth "sticks" for indicating a reply's depth level. It doesn't work as well as I was hoping yet, and I had to pull out some of the more "correct" work I was doing today because I just wasn't getting the results I was hoping for and wanted to get some eyes on this sooner rather than later.

Some interesting areas to explore:

- Something is askew with `TalkPageCellCommentSeparator` and it's been spewing some Auto Layout noise that was throwing off some cell layout, so I temporarily commented out the `setup()` code. (related, I wonder if there's a better approach to constructing these replies in here: https://github.com/wikimedia/wikipedia-ios/blob/7a396ff81c7a92dd93a7f41a3e31fd94071d10d9/Wikipedia/Code/TalkPageCell.swift#L102)

- It doesn't feel like the other views in the cell are yet properly constrained/prioritized, as there are cases I'm noticing where entire `CommentView`'s missing or incorrectly vertically squished.

- I feel pretty confident the frame calculation approaches I've taken so far for the sticks aren't right yet. Conceptually, I need to determine the number of lines that are drawable given **both** the width _and_ height of the view, while respecting the stick width, horizontal spacing, and minimum delta in height between sticks.


tonisevener (commented):
Thanks! It's looking good, just a few of comments. In regards to layout weirdness:

> Something is askew with `TalkPageCellCommentSeparator` and it's been spewing some Auto Layout noise that was throwing off some cell layout
> There are cases I'm noticing where entire CommentView's missing or incorrectly vertically squished.

I'm feeling like a lot of this is stemming from dynamic sizing of these cells - I noticed the layout getting thrown off when I reloadData() in response to replying, just like I did with expanding and collapsing previously. I wonder if that system is not able to handle such dramatic cell subview changes that we're doing.

When I pre-calculate the cell heights, everything works as I would have expected it to, so I've been exploring that. Once this round of PRs are in I'll put this work up for us to consider. So for now I wouldn't worry about it until that is up.

tonisevener (approved):
Looks good! Just added one small change restoring the comment separators. I think something must have been merged in the interim that caused reply heights to get thrown off without them. 

As discussed, this looks good for user testing to me. You can use a follow up PR for any remaining work on this.

PR #4316, opened 2022-09-12
Add basic progressive loading bar

tonisevener:
**Phabricator:** 
N/A

### Notes
This adds our basic blue progressive fake loading bar to the main talk page. I'll add the other POST loading states to the other PRs. 

### Test Steps
1. Load talk page. Confirm you see blue loading bar as it's fetching data.


Wikipedia/Code/TalkPageViewController.swift
```diff
@@ -318,6 +325,7 @@ extension TalkPageViewController: TalkPageCellDelegate {
 
 extension TalkPageViewController: TalkPageViewModelDelegate {
     func talkPageDataDidUpdate() {
+        fakeProgressController.stop()
```
staykids, line 328: What do you think about also calling the `stop()` method on data fetch failure? Otherwise the progress bar just hangs.

Depending on how much refactoring you're interested in, could add a new delegate method for the failure case, or just pass an error parameter in `talkPageDataDidUpdate()`. Probably the simple solution might just be passing some kind of success `BOOL` in `talkPageDataDidUpdate()`, and in the delegate implementation doing something like:

```swift
func talkPageDataDidUpdate(success: Bool) {
    fakeProgressController.stop()

    if success {
        setupHeaderView()
        talkPageView.collectionView.reloadData()
    }
}
```

tonisevener: @staykids Good catch! This is updated. I ended up moving towards a completion block in  #4308 rather than `TalkPageViewModelDelegate`, due to the different followups that are needed in each context we call `fetchTalkPage`:

**Upon first load:**
- reloadData
- stopping the loading indicator
- setting up the header

**Upon posting new topic:**
- reloadData
- scrollToBottom
- present new topic alert banner

I added stopping in the load completion block whether it's success or failure. Let me know what you think, there may be a cleaner way to do it.

staykids (approved):

PR #4318, opened 2022-09-14
Talk pages - small bug fixes

tonisevener:
**Phabricator:** 
N/A

### Notes
This fixes a few small talk page bugs:

1. The new topic banner display was unreliable - moving it to after the compose view controller is fully dismissed seems to help.
2. Tapping subscribe threw off the threading cell layout. Theming the cell after configuration fixed this, because the text views sets their themed attributed strings there.
3. Switching to another app while composing a reply caused the compose screen to disappear and a stuck state when returning to the app.

### Test Steps
1. Confirm the new topic banner displays after adding a new topic.
2. Confirm tapping subscribe or unsubscribe button while a cell is expanded doesn't throw off the layout.
3. Confirm when composing a reply, switching to another app, then returning, you are able to continue typing in the reply compose view.


mazevedofs:
All the other issues were fixed; my guess with the doubled replies is that we're calling `cell.configure(viewModel: cellViewModel, linkDelegate: self)` when subscribing, and that might be causing this new layout issue

tonisevener:
> All the other issues were fixed; my guess with the doubled replies is that we're calling `cell.configure(viewModel: cellViewModel, linkDelegate: self)` when subscribing, and that might be causing this new layout issue

@mazevedofs I think I fixed the double-reply bug you were seeing. Let me know what you think!

mazevedofs (approved):
Works great! Thanks for tackling this :)

PR #4320, opened 2022-09-14
Update coffee roll background colors to be legible

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T310275#8234596

### Notes
Previously, the coffee roll was not legible in themes other than light. This adds theme specific colors for it. Like the page background color, we'll need to update these values once the color theme updates are implemented. I made a note of that [here](https://phabricator.wikimedia.org/T299793#8236964).

### Test Steps
1. Confirm talk pages coffee roll is legible using each theme.


tonisevener (approved):
Looks good!

PR #4321, opened 2022-09-14
Check and remove new line characters from coffee roll text

mazevedofs:
**Phabricator:** N/A (see figma)

### Notes
* This PR verifies the existence of undesired new line characters `\n` at the beginning of coffee roll text to remove it, so the text aligns with the layout according to the design

### Test Steps
1. Go to a talk page that has a coffee roll (a yellow-ish table)
2. Verify that the text aligns according to the design

### DIFF
--- a/Wikipedia/Code/TalkPageHeaderView.swift
+++ b/Wikipedia/Code/TalkPageHeaderView.swift
@@ -335,7 +335,21 @@ final class TalkPageHeaderView: SetupView {
         
         let theme = viewModel.theme
         
-        coffeeRollLabel.attributedText = coffeeRollText.byAttributingHTML(with: .callout, boldWeight: .semibold, matching: traitCollection, color: theme.colors.primaryText, linkColor: theme.colors.link, handlingLists: true, handlingSuperSubscripts: true, tagMapping: ["a": "b"])
+        var attributedText: NSMutableAttributedString = coffeeRollText.byAttributingHTML(with: .callout, boldWeight: .semibold, matching: traitCollection, color: theme.colors.primaryText, linkColor: theme.colors.link, handlingLists: true, handlingSuperSubscripts: true, tagMapping: ["a": "b"])
+        if attributedText.string.first == "\n" {
+            attributedText = removeInitialCharactersFrom(&attributedText)
+        }
+        coffeeRollLabel.attributedText = attributedText
+    }
+    
+    // Coffee roll attributed strigs were sometimes created with extra escape characters (\n) as the initial characters, so we remove them to properly render the text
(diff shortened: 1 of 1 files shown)


Wikipedia/Code/TalkPageHeaderView.swift
```diff
@@ -335,7 +335,21 @@ final class TalkPageHeaderView: SetupView {
         
         let theme = viewModel.theme
         
-        coffeeRollLabel.attributedText = coffeeRollText.byAttributingHTML(with: .callout, boldWeight: .semibold, matching: traitCollection, color: theme.colors.primaryText, linkColor: theme.colors.link, handlingLists: true, handlingSuperSubscripts: true, tagMapping: ["a": "b"])
+        var attributedText: NSMutableAttributedString = coffeeRollText.byAttributingHTML(with: .callout, boldWeight: .semibold, matching: traitCollection, color: theme.colors.primaryText, linkColor: theme.colors.link, handlingLists: true, handlingSuperSubscripts: true, tagMapping: ["a": "b"])
+        if attributedText.string.first == "\n" {
```
tonisevener, line 339: I think this could move into the helper method too, just to simplify the call site. Treat it as a guard early in the method.

Wikipedia/Code/TalkPageHeaderView.swift
```diff
@@ -335,7 +335,21 @@ final class TalkPageHeaderView: SetupView {
...
+            attributedText = removeInitialCharactersFrom(&attributedText)
+        }
+        coffeeRollLabel.attributedText = attributedText
+    }
+    
+    // Coffee roll attributed strigs were sometimes created with extra new line characters (\n) as the initial characters, so we remove them to properly render the text
+    func removeInitialCharactersFrom(_ attributedString: inout NSMutableAttributedString) -> NSMutableAttributedString {
```
tonisevener, line 346: Can you change this into an NSAttributedString extension instead? That way it will be easier to use in the talk page cells. It would be nice if the callsite could look something like:

```
let coffeeRollAttributedText = coffeeRollText.byAttributingHTML(with: .callout, boldWeight: .semibold, matching: traitCollection, color: theme.colors.primaryText, linkColor: theme.colors.link, handlingLists: true, handlingSuperSubscripts: true, tagMapping: ["a": "b"])
cofeeRollLabel.attributedText = coffeeRollAttributedText.removingInitialNewlineCharacters()
```

tonisevener (commented):
Awesome, great find! Just a couple of cleanup suggestions for future use. Not a must-have for this PR though, it can be in a follow up. I'll approve once I confirm on device.

tonisevener (approved):

PR #4323, opened 2022-09-15
Project changes for Xcode 14

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T310091

### Notes
These are project changes needed for moving to Xcode 14. Each commit message is fairly descriptive - I'm happy to chime in if there's any questions during PR review. I also spun off a couple of follow up tasks based on some of these warnings - [T317901](https://phabricator.wikimedia.org/T317901) & [T317905](https://phabricator.wikimedia.org/T317905).

### Test Steps
1. Run app on a few devices, some with iOS 16, some with lower OSes if possible.
2. Confirm there are no build warnings.

Build server has been set up with Xcode 14, but we'll test deployment with future builds and fix any issues in a follow up PR.


Wikipedia/Code/ArticleViewController.swift
```diff
@@ -114,7 +114,6 @@ class ArticleViewController: ViewController, HintPresenting {
     }
     
     deinit {
-        NotificationCenter.default.removeObserver(self)
```
staykids, line 117: I could be wrong but I believe that strange build warning is related to some future Swift 6 `actor` isolation-related [work](https://github.com/apple/swift-evolution/blob/main/proposals/0371-isolated-synchronous-deinit.md).

Would you be comfortable adding `NotificationCenter.default.removeObserver(self)` back to the end of this function (which prevents the build warnings from displaying) or would you prefer to keep it as-is? My reasoning for suggesting we keep it is mostly preventative – if someone in the future _does_ for whatever reason decide to use the block-based API, we're opening ourselves up to an easy to miss potential memory leak in our most used view. In my mind, no harm is being overly protective here.

staykids (approved):
Nice – thanks for getting us on to Team Xcode 14.

One note below about the `deinit` change, but if you're comfortable with it as-is I'm happy to merge.

Wikipedia/Code/ArticleViewController.swift
tonisevener: Good idea! I didn't think about just swapping the order around. This is done.

It's strange to me that the call in the beginning would cause those other warnings to pop up. We're referencing properties in other view controller `deinit`s without warning, which is what clued me to something weird going on specifically in this one. Maybe your linked documentation will clear things up for me.

PR #4324, opened 2022-09-16
Subscribe and unsubscribe to talk page topics

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T311639

### Notes
* Add calls to the subscription APIs to get all active subscriptions and subscribe/unsubscribe to a topic
* Update the talk page cells UI accordingly

### Test Steps
1. Go to a talk page, tap the button to subscribe/unsubscribe to a topic
2. Leave the page and when returning, check if your subscriptions are correctly updated


Wikipedia/Code/TalkPageViewModel.swift
```diff
@@ -160,8 +173,13 @@ final class TalkPageViewModel {
             
             let activeUsersCount = activeUsersCount(topic: topic)
             
-            let topicViewModel = TalkPageCellViewModel(id: topic.id, topicTitle: topicTitle, timestamp: firstReply.timestamp, leadComment: leadCommentViewModel, replies: remainingCommentViewModels, activeUsersCount: activeUsersCount)
+            guard let topicName = topic.name else {
+                DDLogError("Unable to parse topic name")
+                return
```
tonisevener, line 178: Let's do a `continue` here rather than `return`, so we try to turn the rest of the topics in the for loop into view models.

Wikipedia/Code/TalkPageViewModel.swift
```diff
@@ -78,17 +79,21 @@ final class TalkPageViewModel {
...
+    
 
-    func updateSubscriptionToTopic(topic: String, shouldSubscribe: Bool, completion: @escaping (Result<Bool, Error>) -> Void) {
+    func subscribe(to topic: String, shouldSubscribe: Bool, completion: @escaping (Result<Bool, Error>) -> Void) {
         dataController.subscribeToTopic(topicName: topic, shouldSubscribe: shouldSubscribe) { [self] result in
             switch result {
             case let .success(result) :
-                let topicUpdated = topics.filter { $0.topicTitle == topic}
+                let topicUpdated = topics.filter { $0.topicName == topic}
                 topicUpdated[0].isSubscribed = result
```
tonisevener, line 97: Let's not actually flip the flag here, and instead do it optimistically before calling subscribe and update the cell state (see comment above). So these lines can be removed.

Wikipedia/Code/TalkPageViewController.swift
```diff
@@ -355,9 +355,19 @@ extension TalkPageViewController: TalkPageCellDelegate {
...
+            case let .success(didSubscribe):
+                cell.configure(viewModel: configuredCellViewModel, linkDelegate: self)
+                self.handleSubscriptionAlert(isSubscribedToTopic: didSubscribe)
+            case let .failure(error):
+                DDLogError("Error subscribing to topic: \(error)")
+                //TODO: Error handling
+            }
+        }
+ 
     }
```
tonisevener, line 371: This is my thinking for this method. Without this approach, the button won't change states immediately with a slow Internet connection.

```
        guard let cellViewModel = cellViewModel else {
            return
        }
        
        let shouldSubscribe = cellViewModel.isSubscribed
        
        // Optimistically switching cell state. It will revert later if the call fails.
        cellViewModel.isSubscribed.toggle()
        cell.configure(viewModel: cellViewModel, linkDelegate: self)
        
        viewModel.subscribe(to: cellViewModel.topicName, shouldSubscribe: shouldSubscribe) { result in
            switch result {
            case let .success(didSubscribe):
                self.handleSubscriptionAlert(isSubscribedToTopic: didSubscribe)
            case let .failure(error):
                DDLogError("Error subscribing to topic: \(error)")
                
                // Revert isSubscribed state back
                cellViewModel.isSubscribed.toggle()
                
                // Reconfigure cell, if it already hasn't been reconfigured with another cell view model
                if cell.viewModel?.topicName == cellViewModel.topicName {
                    cell.configure(viewModel: cellViewModel, linkDelegate: self)
                }
                
                // TODO: Error banner
            }
        }
```

tonisevener (changes requested):
Nice work! It's working well in my (limited) testing. I have mostly some small tweaks, and one request to optimistically flip the button state instead of waiting for the fetch to go through. Thanks!

Wikipedia/Code/TalkPageViewController.swift
mazevedofs: Good point; it does make it smoother! I'm just curious about this check `if cell.viewModel?.topicName == cellViewModel.topicName {..}`, do you think there's a risk to configuring it with the incorrect view model?

tonisevener: Yep - I'm worried here that the cell could already be reused elsewhere in the collection view once the POST call completes, so I added that check. Although that thinking could be outdated with collection view / table view changes in recent years - I did some testing this morning on a similar concern elsewhere and never ran into it.

A user could subscribe, the button state immediately changes to subscribed, then they scroll down for a while as the API call goes through on a slow connection. If they have scrolled down far enough for this cell to get reused and receive a different topic view model in this interim period, we want to NOT reconfigure it with this old failed subscription view model. Without this check, they would suddenly see that same failed subscription topic from above appear at the bottom of the collection view. Hope that makes sense!

mazevedofs: It makes sense, thanks for clarifying that! I added this verification, I think it's ready for another look

PR #4326, opened 2022-09-16
Add `authenticationManager` to `TalkPageViewModel`

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T311637 (partially)

### Notes
This adds the shared `WMFAuthenticationManager` as a passed property on `TalkPageViewModel`. I went this route over passing the entire `MWKDataStore` as a dependency just to keep things leaner and more focused. I'm happy to close this though if I'm missing a benefit we'd have for this feature in passing the entire data store as a dependency here or if someone else had plans for how to pass this data along.


tonisevener (approved):
Looks good! Agreed, I generally prefer to not pass the whole `MWKDataStore` to keep things leaner. Thanks for checking!

PR #4328, opened 2022-09-20
Talk Pages - Cell performance improvements

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T311637

### Notes
This is an attempt at improving the performance and layout quirks of the talk page cells that we've been seeing. The goal is to get this to an acceptable place without having to rework too much of what we already have.

I could not figure out how to get smooth layout and smooth reloading of cells with the dynamic cell layout heights. Even without expanded cells, the layout appeared jumpy when scrolling quickly, because it defaults to the estimated size very often. Once I switched to precalculated heights using a sizing view, everything worked smoothly. I am heavily caching these sizes to try and reduce the additional layout burden this precalculation causes. To help with testing, also implemented the "Open all threads" functionality.

There's additional precalculation and caching I could do up front for other orientations, but I wanted to go ahead and get this up before too long for feedback. I think if this still doesn't work for us, we may want to consider reworking the view model comments into their own cells, or dipping into manually sized frames everywhere. I would like to avoid that if we can though.

### Test Steps
1. Load a large talk page (Queen Elizabeth II is a good one). Confirm scrolling is much smoother than it was in `main`.
2. Confirm cells expand and collapse correctly.
3. Confirm rotation works correctly (it should be faster after the first rotation, once the new sizes are cached).
4. Confirm open all threads works correctly. I added a spinner here since this could take a while to lay out.

### DIFF
--- a/Wikipedia/Code/TalkPageCell.swift
+++ b/Wikipedia/Code/TalkPageCell.swift
@@ -11,28 +11,7 @@ protocol TalkPageCellReplyDelegate: AnyObject {
     func tappedReply(commentViewModel: TalkPageCellCommentViewModel)
 }
 
-final class TalkPageCell: UICollectionViewCell {
-
-    // MARK: - Properties
-
-    static let reuseIdentifier = "TalkPageCell"
-
-    weak var viewModel: TalkPageCellViewModel?
-    weak var delegate: TalkPageCellDelegate?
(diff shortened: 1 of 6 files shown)


tonisevener:
Closing this one in preference of #4331.

PR #4329, opened 2022-09-20
Fix deployment & CircleCI tests for Xcode 14

tonisevener:
**Phabricator:** 
N/A

### Notes
This fixes issues with deployment and our failing tests. 

754be93 fixed a transporter error we were seeing upon attempting to upload: 

> Could not find transporter at /Applications/Xcode.app/Contents/Developer/. Please make sure you set the correct path to your Xcode installation.

c34377d fixed our failing CircleCI tests. It removes the `xcversion` check step in the pull requests fastlane code. This code references the value in our .xcversion file for which Xcode version to run against. It's really only beneficial for deployment from our build server at this point - CircleCI uses its own Xcode version number in our .circleci/config.yml file. I removed this so that they are no longer linked.

As to why CircleCI suddenly started failing (whereas it ran fine a few days ago on the Xcode 14 PR), I suspect it's something to do with their [Xcode update policy](https://circleci.com/docs/xcode-policy).

> When Xcode 13.2.1 was released, we removed the previous patch version, 13.2.0, and automatically redirected all requests for 13.2.0 to 13.2.1.

Perhaps version 14.0.0 no longer exists, and 14.0.0 is redirecting to a later patch on CircleCI's side. So our explicit `xcversion` check for 14.0.0 failed.

### Test Steps
1. Confirm PR tests passed. Deployment has already been tested.


staykids (approved):
👍🏽

PR #4331, opened 2022-09-21
Talk Pages - Cell performance improvements (alternative)

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T311637

### Notes
This is a smaller alternative approach to #4328 to improve cell performance. Some of the commits I kept, but I was able to throw out all of the gross work (the sizing view portion). I'll go into the commits that I think had the most effect on improving things here:

7fcccb3 - This was a hacky thing that I added a while back to re-layout the cell after posting a reply. Originally I needed to add this because it seemed to default to the estimated cell height (225) upon reload. It's no longer needed now after 4d47941.
53ac73a - This reduces the amount of stack view movement as the cell is configured, which improves the scrolling frame rate.
4d47941 - This fixed the issue of the collection view occasionally using the estimated cell size when scrolling quickly. It also smoothed out the appearance of `reloadData()` after posting a reply.
392ef58 - I still seemed to need this commit, otherwise occasionally the cell would size incorrectly upon expanding + rotating. I mainly reworked the comment view away from stack views and am setting up the text, reply button, and depth indicator view constraints directly. I'm still not sure why this helps though, the stack view setup looked fine to me before.

One last remaining issue that I notice is some content offset jumping on iOS14 upon cell expanding and collapsing. I'm not sure why this is, but I see it even after subclassing `UIViewController` directly, so I don't think it's something buried in our custom navigation bar / header view setup. I'll spin off a separate bug task for this to keep track of.

### Test Steps
1. Load a large talk page (Queen Elizabeth II is a good one). Confirm scrolling is much smoother than it was in main.
2. Confirm cells expand and collapse correctly.
3. Confirm rotation works correctly.
4. Confirm open all threads works correctly.
5. Confirm posting an inline reply reloads the cell correctly with proper cell sizing.
6. Confirm adding a new topic still appears at the bottom of the list correctly.


Wikipedia/Code/TalkPageCell.swift
```diff
@@ -79,10 +79,13 @@ final class TalkPageCell: UICollectionViewCell {
     }
 
     override func prepareForReuse() {
+        super.prepareForReuse()
```
staykids, line 82: Nice catch 😅

Wikipedia/Code/TalkPageCell.swift
```diff
@@ -117,37 +118,38 @@ final class TalkPageCell: UICollectionViewCell {
...
-        leadReplySpacer.isHidden = !viewModel.isThreadExpanded
-        leadReplyButton.isHidden = !viewModel.isThreadExpanded
-
-        let comments: [UIView] = stackView.arrangedSubviews.filter { view in view is TalkPageCellCommentView || view is TalkPageCellCommentSeparator }
-        stackView.arrangedSubviews.forEach { view in
-            if comments.contains(view) {
-                view.removeFromSuperview()
-            }
-        }
+        if viewModel.isThreadExpanded {
```
staykids, line 121: I'm so glad you refactored my original approach here – I was never satisfied with it and it felt overly wasteful.

staykids (approved):
Nice - scrolling performance feels quite wonderful on iOS 15 for me. I noticed some oddness with the reply depth indicator view not being sized correctly, but it's not worth diagnosing and treating that here. Better to address it if it remains after the sticks refactoring's complete.

PR #4333, opened 2022-09-23
Talk pages - Simplify reply depth view

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T310279

Specifically:
https://phabricator.wikimedia.org/T310279#8256483
https://phabricator.wikimedia.org/T310279#8263680

### Notes
This is an attempt at the simple approach to the reply depth indicator as discussed. It maxes out at 10 lines, then begins to show the reply depth indicator.

### Test Steps
1. Expand various talk page threads on a device/simulator capable of showing both size classes. Confirm it stops at 10 lines, and displays an additional depth indicator label beyond that. The indicator count does not consider the lines it covers up, which was ok'ed by design.
2. Confirm it displays correctly in RTL.
3. Confirm the depth label wraps with a giant number, or with a large accessibility type.


staykids (approved):
Working wonderfully in my testing so far! Thanks for picking up on this work where I left off. Glad we landed on something that works for both design and engineering.

PR #4335, opened 2022-09-23
Add full screen coffee roll text view

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T310275 (partially)

### Notes
Adds full screen coffee roll view. I reused the text attribution and link handling logic and pattern from the talk page comment view – please let me know if this is unexpected.

### Test Steps
1. Open a talk page with a coffee roll
2. Confirm tapping read more in coffee roll opens new full screen view controller
3. Confirm it renders as expected regardless of the theme

### DIFF
--- a/Wikipedia/Code/TalkPageCoffeeRollView.swift
+++ b/Wikipedia/Code/TalkPageCoffeeRollView.swift
@@ -1,0 +1,76 @@
+import UIKit
+
+final class TalkPageCoffeeRollView: SetupView {
+
+    // MARK: - Properties
+
+    var theme: Theme!
+    var viewModel: TalkPageCoffeeRollViewModel!
+    
+    weak var linkDelegate: TalkPageTextViewLinkHandling?
+
+    // MARK: - UI Elements
+
+    lazy var textView: UITextView = {
+        let textView = UITextView()
+        textView.translatesAutoresizingMaskIntoConstraints = false
+        textView.isScrollEnabled = true
+        textView.isEditable = false
+        textView.textContainerInset = UIEdgeInsets(top: 16, left: 0, bottom: 16, right: 16)
+        textView.delegate = self
+        return textView
+    }()
+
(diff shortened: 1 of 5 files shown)


Wikipedia/Code/TalkPageCoffeeRollView.swift
```diff
@@ -0,0 +1,76 @@
+import UIKit
+
+final class TalkPageCoffeeRollView: SetupView {
+
+    // MARK: - Properties
+
+    var theme: Theme!
```
tonisevener, line 7: I think you can remove the `!`s here if you're setting them in the `init`.

Wikipedia/Code/TalkPageCoffeeRollViewController.swift
```diff
@@ -0,0 +1,58 @@
...
+        super.apply(theme: theme)
+        coffeeRollView.apply(theme: theme)
+    }
+
+}
+
+extension TalkPageCoffeeRollViewController: TalkPageTextViewLinkHandling {
+
+    func tappedLink(_ url: URL, sourceTextView: UITextView) {
+        guard let url = URL(string: url.absoluteString, relativeTo: viewModel.talkPageURL) else {
```
tonisevener, line 52: I'm starting to think this relative link may not be entirely correct here and in the other spot. Links to talk page archives from the coffee roll (which open natively! 😎) causes _their_ links to get screwy. But I think hashing that out is beyond the scope of this task. I'll make a bug note of it.

tonisevener (approved):
Awesome, looks good! Just one minor note that you're free to change or not, and another thing with relative links that we can tackle separately.

Wikipedia/Code/TalkPageCoffeeRollView.swift
staykids: Good catch, sorry – a holdover from before I switched my approach to requiring them at `init`. Updated!

PR #4338, opened 2022-09-26
Fix double back button on iOS16 Vanish view

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T318625

### Notes
iOS16 shows a double back navigation bar button in our Vanishing flow. This seemed unique to UIHostingControllers that are a UINavigationController's root view controller for iOS16. Embedding it in it's own container view controller seems to fix it.

### Test Steps
1. Load the Vanishing screen in iOS16.
2. Confirm you only see a single arrow back button.
3. Tap back button, confirm it goes back.
4. Go back to Vanishing screen. Fill out additional information.
5. Tap back button, confirm you see a confirmation action sheet for discarding a request before it navigates you back.
6. Confirm tapping Send Request still sends you to email client as before.
7. Repeat steps 1-6 with iOS 15, 14, and 13 (note you will not be navigated to email client on simulator in step 6 - this is expected).

### DIFF
--- a/Wikipedia/Code/AccountViewController.swift
+++ b/Wikipedia/Code/AccountViewController.swift
@@ -145,8 +145,8 @@ class AccountViewController: SubSettingsViewController {
                 return
             }
 
-            let hostingController = VanishAccountCustomUIHostingController(title: CommonStrings.vanishAccount.localizedCapitalized, theme: theme, username: username)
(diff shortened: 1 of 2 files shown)


mazevedofs (approved):
Working great in all OS versions!

PR #4340, opened 2022-09-28
Talk page cache

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T315330

### Notes
* This PR creates caching for the new talk pages, using the `SharedContainerCache` to persist .json files. It also adds a housekeeping task to remove older cached pages.

### Test Steps
1. Go to a talk page
2. leave the page, turn off the internet connection and return to the same page. It should load the cached topics
3. To test the housekeeping, export your cached data to your computer. Go To Settings >Help and Feedback > Export User data. Check how many items you have under Talk Page Cache. 
4. Lower the max cached items on `deleteStaleCachedItems` so you don't have to save 50 pages 
5. Follow the steps described here (test step 7) https://github.com/wikimedia/wikipedia-ios/pull/4211

Execute the background database cleanup task. Set a breakpoint right after task scheduling [here](https://github.com/wikimedia/wikipedia-ios/blob/b2a693672852457a39763681820f7f3abd761b97/Wikipedia/Code/AppDelegate.m#L263). Background the app, then when you hit the breakpoint print this in the console:

> e -l objc -- (void)[[BGTaskScheduler sharedScheduler] _simulateLaunchForTaskWithIdentifier:@"org.wikimedia.wikipedia.databaseHousekeeper"]

After a little bit you should see a printout in the console saying:

> Simulating launch for task with identifier org.wikimedia.wikipedia.databaseHousekeeper

When you see this, hit the play button in the debugger to continue executing the program. The task will run and then the console will print out:

> Marking simulated task complete: <BGProcessingTask: org.wikimedia.wikipedia.databaseHousekeeper>

6. Now export your data again, and check if the older entries were deleted


WMF Framework/SharedContainerCache.swift
```diff
@@ -1,20 +1,23 @@
...
         case userDataExportSyncInfo = "User Data Export Sync Info"
+        case talkPageCache = "Talk Page Cache"
     }
     
     private let pathComponent: PathComponent
     private let defaultCache: () -> T
+    private let fileFolder: String
     
-    public init(pathComponent: PathComponent, defaultCache: @escaping () -> T) {
+    public init(pathComponent: PathComponent, fileFolder: String = String(), defaultCache: @escaping () -> T) {
```
tonisevener, line 17: I realize this is a larger refactor, but I think this class would be much easier to read if we reworked these parameters into:

```
private let fileName: String
private let subdirectoryPathComponent: String?

public init(fileName: String, subdirectoryPathComponent: String?, defaultCache: @escaping () -> T) {
    self.fileName = fileName
    self.subdirectoryPathComponent = subdirectoryPathComponent
    ...
}
```

A class that needs to cache a file in the root directory inits without a subdirectory, like:

```
let sharedCache = SharedContainerCache<PushNotificationsCache>.init(fileName: "Push Notifications Cache", defaultCache: {...})
```

and talk pages can init like:

```
let sharedCache = SharedContainerCache<TalkPageCache>.init(fileName: cachedFileName, subdirectory: "Talk Pages Cache", defaultCache: {
    TalkPageCache(talkPages: [])
})
```

I just think working with the `PathComponent` enum is hurting us more than helping, at this point. Let me know what you think! Most of the rest of my comments will have this refactor in mind.

WMF Framework/SharedContainerCache.swift
```diff
@@ -25,20 +28,79 @@ public final class SharedContainerCache<T: Codable> {
         return cacheDirectoryContainerURL.appendingPathComponent(pathComponent.rawValue).appendingPathExtension("json")
     }
     
+    private func subdirectoryFileURL(to subfolder: String) -> URL {
```
tonisevener, line 31: This could be reworked to:

```
private func subdirectoryURL() -> URL? {
    guard let subdirectoryPathComponent = subdirectoryPathComponent else {
        return nil
    }

    return cacheDirectoryContainerURL
            .appendingPathComponent(subdirectoryPathComponent, isDirectory: true)
}
```

WMF Framework/SharedContainerCache.swift
```diff
@@ -25,20 +28,79 @@ public final class SharedContainerCache<T: Codable> {
         return cacheDirectoryContainerURL.appendingPathComponent(pathComponent.rawValue).appendingPathExtension("json")
```
tonisevener, line 28: This could be reworked to:

```
private var cacheDataFileURL: URL {
    let baseURL = subdirectoryURL ?? cacheDirectoryContainerURL
    return baseURL.appendingPathComponent(fileName).appendingPathExtension("json")
}
```

WMF Framework/SharedContainerCache.swift
```diff
@@ -25,20 +28,79 @@ public final class SharedContainerCache<T: Codable> {
...
     }
     
+    private func subdirectoryFileURL(to subfolder: String) -> URL {
+        return cacheDirectoryContainerURL
+            .appendingPathComponent(pathComponent.rawValue, isDirectory: true)
+    }
+
+    func cacheFileURL(to subfolder: String) -> URL {
```
tonisevener, line 36: This could be deleted.

WMF Framework/SharedContainerCache.swift
```diff
@@ -25,20 +28,79 @@ public final class SharedContainerCache<T: Codable> {
...
+        return cacheDirectoryContainerURL
+            .appendingPathComponent(pathComponent.rawValue, isDirectory: true)
+            .appendingPathComponent(subfolder)
+            .appendingPathExtension("json")
+    }
+    
     public func loadCache() -> T {
-        if let data = try? Data(contentsOf: cacheDataFileURL), let decodedCache = try? JSONDecoder().decode(T.self, from: data) {
+        let filePath = fileFolder.isEmpty ? cacheDataFileURL : cacheFileURL(to: fileFolder)
```
tonisevener, line 44: This method could go back to how it was before, because `cacheDataFileURL` would be capable of making URLs with or without the subdirectory.

WMF Framework/SharedContainerCache.swift
```diff
@@ -25,20 +28,79 @@ public final class SharedContainerCache<T: Codable> {
...
+        if let data = try? Data(contentsOf: filePath), let decodedCache = try? JSONDecoder().decode(T.self, from: data) {
             return decodedCache
         }
 
         return defaultCache()
     }
 
     public func saveCache(_ cache: T) {
+        if fileFolder.isEmpty {
```
tonisevener, line 53: This method could go back to how it was before. But before writing, create the subdirectory if needed:

```
public func saveCache(_ cache: T) {

        let encoder = JSONEncoder()
        guard let encodedCache = try? encoder.encode(cache) else {
            return
        }

        if let subdirectoryURL = subdirectoryURL {
            try? FileManager.default.createDirectory(at: subdirectoryURL, withIntermediateDirectories: true, attributes: nil)
        }

        try? encodedCache.write(to: cacheDataFileURL)
```

WMF Framework/SharedContainerCache.swift
```diff
@@ -25,20 +28,79 @@ public final class SharedContainerCache<T: Codable> {
...
+            guard let encodedCache = try? encoder.encode(cache) else {
+                return
+            }
+            try? encodedCache.write(to: cacheDataFileURL)
+        } else {
+            saveCache(to: fileFolder, cache)
+        }
+    }
+    
+    private func saveCache(to folder: String, _ cache: T) {
```
tonisevener, line 64: This new method could be deleted.

tonisevener (changes requested):
Thanks for working through this! I appreciate you trying to keep the changes low-touch, that's a good instinct to have. I'm finding the way subfolders and path components are used inside and outside the class confusing, so I have a refactor to propose which will affect the `init` call sites everywhere. Happy to discuss it further if you disagree!

Wikipedia/Code/TalkPageViewModel.swift
```diff
@@ -25,6 +25,10 @@ final class TalkPageViewModel {
     var theme: Theme = .light
     private(set) var topics: [TalkPageCellViewModel] = []
 
+    private let sharedCache = SharedContainerCache<TalkPageCache>.init(pathComponent: .talkPageCache, defaultCache: {
```
tonisevener, line 28: Let me know if I'm missing something, but I don't think you need to use the cache in this `TalkPageViewModel` if you're already using it in the `TalkPageDataController`.

mazevedofs:
Thanks @tonisevener for the thought you put into this review, it made the `SharedContainerCache` class much more readable.  I guess it's ready for another review, let me know what you think :)

Wikipedia/Code/TalkPageDataController.swift
```diff
@@ -117,9 +111,18 @@ class TalkPageDataController {
         }
     }
 
+    private func cachedFileName() -> String {
+        guard let languageCode = siteURL.wmf_contentLanguageCode else {
+            return "en"
```
mazevedofs, line 116: I'm not sure if defaulting to "en" is the best approach, but I am happy to discuss better options

Wikipedia/Code/TalkPageDataController.swift
```diff
@@ -125,9 +137,11 @@ class TalkPageDataController {
                 
                 switch result {
                 case .success(let items):
+                    cache.talkPages = items
```
tonisevener, line 140: Can you name this property `talkPageItems` instead of `talkPages`? So it will be:

`cache.talkPageItems = items`. I think it will be a little more clear.

WMF Framework/SharedContainerCache.swift
```diff
@@ -1,19 +1,14 @@
...
-    
-    // values must all be distinct
-    public enum PathComponent: String {
-        case widgetCache = "Widget Cache"
-        case pushNotificationsCache = "Push Notifications Cache"
-        case userDataExportSyncInfo = "User Data Export Sync Info"
-    }
-    
-    private let pathComponent: PathComponent
+public final class SharedContainerCache<T: Codable>: SharedContainerCacheHousekeepingProtocol {
```
tonisevener, line 3: Optional suggestion: For common file names, you could set up a struct (separate from the class unfortunately, because it's generic):

```
struct SharedContainerCacheCommonFileNames {
    static let pushNotificationsCache = "Push Notifications Cache"
}

public final class SharedContainerCache<T: Codable>: ...
```

This way you don't have multiple string literals peppered throughout the codebase. A typo on any of them could risk throwing things off with caching.

```
let sharedCache = SharedContainerCache<PushNotificationsCache>(fileName: SharedContainerCacheCommonFileNames.pushNotificationsCache, defaultCache: { PushNotificationsCache(settings: .default, notifications: []) })
```

tonisevener (changes requested):
Thanks for jumping on this! It's looking good. I think it will be ready after this next round of changes.

WMF Framework/SharedContainerCache.swift
```diff
@@ -39,6 +41,38 @@ public final class SharedContainerCache<T: Codable> {
...
+        if let subdirectoryURL = subdirectoryURL() {
+            try? FileManager.default.createDirectory(at: subdirectoryURL, withIntermediateDirectories: true, attributes: nil)
+
+        }
+
         try? encodedCache.write(to: cacheDataFileURL)
     }
+
+    /// Persist only the last 50 visited talk pages
+    @objc public func deleteStaleCachedItems(for folder: String) {
```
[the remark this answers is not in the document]
mazevedofs: Good catch, the static method works great with the housekeeping!

Wikipedia/Code/TalkPageViewController.swift
```diff
@@ -155,9 +155,11 @@ class TalkPageViewController: ViewController {
                 self.talkPageView.emptyView.actionButton.addTarget(self, action: #selector(self.userDidTapAddTopicButton), for: .primaryActionTriggered)
                 self.updateEmptyStateVisibility()
                 self.talkPageView.collectionView.reloadData()
```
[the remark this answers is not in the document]
mazevedofs: Missed that!

Wikipedia/Code/TalkPageFetcher.swift
```diff
@@ -17,7 +17,7 @@ struct TalkPageThreadItems: Codable {
     }
 }
 
-struct TalkPageItem: Codable {
+struct TalkPageItem: Hashable, Codable {
```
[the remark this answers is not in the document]
mazevedofs: It's not being used for the caching, removed!

WMF Framework/SharedContainerCache.swift
mazevedofs: Done!

Wikipedia/Code/WMFAppViewController.m
```diff
@@ -142,6 +142,8 @@ - (void)viewDidLoad {
     self.theme = [[NSUserDefaults standardUserDefaults] themeCompatibleWith:self.traitCollection];
 
     self.backgroundTasks = [NSMutableDictionary dictionaryWithCapacity:5];
+    [SharedContainerCacheHousekeeping deleteStaleCachedItemsIn: @"Talk Page Cache"];
```
tonisevener, line 145: I think an extra `deleteStaleCachedItems` call slipped in here.

Wikipedia/Code/WMFAppViewController.m
```diff
@@ -632,7 +635,10 @@ - (void)performDatabaseHousekeepingWithCompletion:(void (^)(NSError *))completio
         DDLogError(@"Error on cleanup: %@", housekeepingError);
         housekeepingError = nil;
     }
-    
+
+    /// Housekeeping for the new talk page cache
+    [SharedContainerCacheHousekeeping deleteStaleCachedItemsIn: @"Talk Page Cache"];
```
tonisevener, line 640: Go ahead and use your new SharedContainerCacheCommonFileNames.talkPageCache struct here.

WMF Framework/SharedContainerCache.swift
```diff
@@ -1,35 +1,42 @@
...
-    
-    // values must all be distinct
-    public enum PathComponent: String {
-        case widgetCache = "Widget Cache"
-        case pushNotificationsCache = "Push Notifications Cache"
-        case userDataExportSyncInfo = "User Data Export Sync Info"
-    }
-    
-    private let pathComponent: PathComponent
+public struct SharedContainerCacheCommonFileNames {
```
tonisevener, line 3: Actually since `talkPageCache` technically becomes a folder, maybe just name this `SharedContainerCacheCommonNames` instead of `SharedContainerCacheCommonFileNames`.

tonisevener (changes requested):
Just a few more minor things!

tonisevener (approved):
Working great! Great job on this!

PR #4342, opened 2022-09-29
Talk pages - Reply Compose action sheets

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T310291#8131100
https://phabricator.wikimedia.org/T318438

### Notes
This PR adds a confirmation action sheet to the reply compose view in two situations:

- The user has added text and attempts to dismiss.
- The user has added text and then taps reply on a different comment.

In both situations we ask the user to confirm that they want to discard their reply, before following through with their action.

The sizing of the reply compose view was buggy during this action sheet display, so I reworked the constraints a bit. Since I was in there I also fixed these issues:

1. Added a display and dismiss animation to the view so it doesn't feel so sudden.
2. Added dismissal if the user is swiping down fast enough, per the task requirements.
3. Fixed the bottom content inset issues in the collection view. Originally I tried to fix this in #4318, but I like this fix more (I will revert that old commit).
4. Increased the height of the view and added additional spacing on larger text sizes, to fix https://phabricator.wikimedia.org/T318438.

### Test Steps
1. Trigger reply view. Confirm it is now larger upon first appearance to improve https://phabricator.wikimedia.org/T318438.
2. Type text, then tap the X to dismiss. Confirm you now see an action sheet to confirm first.
3. Repeat step 2, but this time swipe down to dismiss. Confirm you see an action sheet.
4. Trigger a reply view. While it's open, tap another reply button. You should see the action sheet confirmation. After confirming, the reply view should dismiss, then a new reply view will appear.


mazevedofs (approved):
Working great!

PR #4347, opened 2022-10-05
Show user education onboarding view on first new topic or comment reply

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T319075

### Notes
Displays user education onboarding modal on first tap of the new topic button or tapping reply in a thread. This view only shows once. I avoided the attributed string complexity within onboarding body text, but if it's a design requirement after review, I'll adjust to use an iOS 15+ only approach akin to the account vanishing view.

### Test Steps
1. Open any article or user talk page
2. Tap either new topic or reply to a comment
3. Confirm you see user education view. Pulling to dismiss or tapping dismiss button will set the user defaults key to prevent showing this view ever again.
4. Reset user defaults (or just add `UserDefaults.standard.wmf_userHasOnboardedToContributingToTalkPages = false` to the talk page view controller's `viewDidLoad`) and repeat step 2 with the alternate reply type.


Wikipedia/Code/TalkPageTopicReplyOnboardingView.swift
```diff
@@ -0,0 +1,77 @@
...
+                    Spacer(minLength: 64)
+                    Text(LocalizedStrings.title)
+                        .font(Font(titleFont))
+                        .multilineTextAlignment(.center)
+                        .foregroundColor(Color(theme.colors.primaryText))
+                    Spacer(minLength: 44)
+                    Image("talk-pages-empty-view-image")
+                    Spacer(minLength: 31)
+                    Text(LocalizedStrings.body)
+                        .font(.callout)
```
tonisevener, line 51: This label disappears on me in dark and black themes.

Wikipedia/Code/TalkPageViewController.swift
```diff
@@ -276,6 +278,7 @@ class TalkPageViewController: ViewController {
         navVC.modalPresentationStyle = .pageSheet
         navVC.presentationController?.delegate = self
         present(navVC, animated: true, completion: nil)
+        presentTopicReplyOnboardingIfNecessary()
```
tonisevener, line 281: I wonder if we should call this in the previous line's completion block, because you check for `presentedViewController` in this method. Is there a chance `presentedViewController` could not yet be assigned?

This is minor/optional, since it's working perfectly fine as-is with my testing.

tonisevener (changes requested):
Looks great! Just noticed one fix needed with theming, and another optional note.

Wikipedia/Code/TalkPageTopicReplyOnboardingView.swift
staykids: Whoops - updated!

Wikipedia/Code/TalkPageViewController.swift
staykids: I recall having this same thought when originally working on this, and at the time left it afterwards just because it was working as expected. But I agree, it feels more _right_ having it in the completion block as if it's respecting some kind of order of operations. Updated!

staykids:
Should be ready for a final look!

tonisevener (approved):
Working great!

PR #4348, opened 2022-10-05
Talk pages - Deep linking

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T310275 - The "Entry points" section

### Notes
This work is mainly to add deep linking ability from Notifications Center, to scroll to a particular talk page topic or reply after tapping an action from those screens. Scrolling to a particular topic should also work from any talk page universal link with a section fragment.

### Test Steps
1. Tap various "Talk page message" notifications (I suggest using our handy filter) from Notifications Center. Try some that appear at the top of the talk page, and some that appear at the bottom. Confirm the talk page scrolls so that the new reply is visible after the talk page loads.
2. Try loading a talk page url with a topic title fragment in Safari (for example: https://test.wikipedia.org/wiki/User_talk:Tsevener#why), and tap the "Open in app" banner if necessary. Confirm it scrolls to the correct topic cell and expands it after loading.

Note: When the notifications API doesn't give us any talk page title or reply text information, we fall back to generic Notifications Center text like "Talk page message / Message on your talk page". We will not be able to scroll to a particular talk page topic or comment on those.


mazevedofs (approved):
Looks great!

PR #4357, opened 2022-10-11
Talk pages - Prevent disclosure button from horizontally compressing

staykids:
**Phabricator:** N/A

### Notes
This fixes a bug I noticed where a long title would sometimes make the disclosure button compress horizontally when the user was viewing a talk page without being logged in.

### Test Steps
1. Replace the `TopicTitleTextView` text at line 265 in `TalkPageCellTopicView` to something long, like "This is a title that is long enough to make the disclosure button shrink when the text is at two lines"
2. Log out, and go to any text page.
3. Confirm you don't see any horizontally compressed disclosure buttons like in the before screenshot below. In `main`, it should be easy to see the bug.

### Screenshot

ba


tonisevener (approved):
Good catch!

PR #4359, opened 2022-10-12
Talk pages - New topic & reply logged out flows

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T310285

### Notes
This PR adds a login/signup prompt when the user publishes a new topic or reply while logged out.

I tried for a while to automatically publish once a successful login or account creation went through, but tracking that through the login and create account view controllers was pretty gross. I also explored responding to the`WMFAuthenticationManager.didLogInNotification` notification - but I worry that this could fire in unexpected ways and publish a user's half-finished talk page message (say the user backgrounds while composing, then returns later, and the app automatically logs them in again).

So to try and keep things a little cleaner and safer, this approach requires the user to tap "Publish" again after their log in and sign up modals dismiss. I looked a bit at Desktop Wikipedia and it acts the same way - requiring the user to tap the Add Topic and Reply button after log in.

Marking as Draft while I wait for any feedback on https://phabricator.wikimedia.org/T310285#8312935.

### Test Steps
1. Try to add a new topic while logged out. Confirm you see the log in prompt after tapping Publish.
2. Choose "Edit without logging in". Confirm prompt disappears and new topic is published anonymously.
3. Repeat steps 1 - 2 for an inline reply.
4. Repeat steps 1 - 2 while logged out, but this time tap "Log in or create account" on the prompt. Confirm prompt disappears after logging in. Tap Publish again. Confirm new topic is added with logged in user signature.
5. Repeat step 4 for an inline reply.
6. (Optional) Repeat step 4 while logged out, but this time on the Log in screen, tap the "Join Wikipedia" link. Confirm prompt disappears after signing up. Tap Publish again. Confirm new topic is added with logged in user signature. **Note**: I think this is pretty safe as long as we're only publishing on Test Wikipedia. But note that I was eventually blocked here from creating too many accounts on the same IP within a day. So given that, I think it's fine to leave these Sign up steps to QA.
7. (Optional) Repeat step 6 for an inline reply.


mazevedofs (approved):
Looking great!

mazevedofs:
@tonisevener there are just some conflicts that need to be fixed then we can merge this PR!

PR #4362, opened 2022-10-13
Talk pages - Topic cell edge cases

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T310275#8244655

### Notes
This adjusts our cell view model and cell view to handle some topic edge cases.

### Test Steps
Visit User_talk:Brion_VIBBER on EN Wikipedia - confirm you can now see two new cells at the top. The first is a thread of replies (just one reply in this example) without a topic title, with the subscription button hidden. The second cell should just be an informative, expandable cell. It hides the subscription button, as well as the metadata row and reply button.


mazevedofs (approved):
Looking great!

PR #4364, opened 2022-10-14
Talk pages - iPad tweaks

tonisevener:
**Phabricator:** 
N/A, but called this PR out in https://phabricator.wikimedia.org/T310275#8317922.

### Notes
This is a PR to align a few more things to the `readableContentGuide`, so that iPad content lines up with the header. 

I also added a final commit to adjust the full-screen coffee roll so that the scroll indicator appears on the edge. This is more of a personal preference - happy to revert if there's pushback.

### Test Steps
1. Review topic cells, full screen coffee roll and empty state on iPad and iPhone. Confirm there are no layout issues.

### Screenshots/Videos

Before:
B4bID
After:
HQn7e

Before:
tRdEm
After:
763YE

Before:
ODQKg
After:
aD1GR

Before:
QBP1s
After:
OnpdE


mazevedofs:
Looking good, just confirming that this is the intended layout for the button (or should the text have as much padding as the button?)

IMG_0010

mazevedofs (changes requested):
Looking good, just confirming that this is the intended layout for the button (or should the text have as much padding as the button?)

IMG_0010

tonisevener:
> Looking good, just confirming that this is the intended layout for the button (or should the text have as much padding as the button?)
> 
> IMG_0010

Good catch! I'm not sure on this one. I asked design in the Phab task so we'll see what they say.

mazevedofs (approved):

PR #4365, opened 2022-10-17

translatewiki:
Translation updates


tonisevener:
Tests were failing due to incorrect replacement token in the translation (`%1` instead of `$1`). Fixed at translatewiki [here](https://translatewiki.net/wiki/Wikimedia:Wikipedia-ios-notifications-center-go-to-user-page/sl). I'll close this now and wait to see if the next PR passes.

PR #4368, opened 2022-10-19
Talk pages - cell design review tweaks

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T310279#8307471

### Notes
This PR adds a time to the timestamp label as well as adjust the colors of the reply depth indicators. To match Android, we're going with showing the timestamp in the user's local timezone. Note that this does mean it may be a different time than what is displayed in their signature, you can see the comment thread starting at https://phabricator.wikimedia.org/T310279#8330402 for more details.

For the format - I'm trying to match the mocks as close as I can, but used `setLocalizedDateFormatFromTemplate` to allow the system to adjust it to locale conventions. Here's an example of how it looks on an EN Wiki talk page vs. a DE Wiki talk page:

EN:
IMG_0839
DE:
IMG_0840

**Test Steps:**

1. Confirm the time displays in your device timezone local time. Confirm the format changes depending on the Wiki you are looking at. Note: RTL layout changes have not been applied yet. That will be in a followup PR.
2. Confirm reply depth indicators have changed according to Phab comment.


mazevedofs (approved):
Looking great!

PR #4371, opened 2022-10-21
Add vanish account warning view

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T320758

### Notes
Adds a warning modal when a user taps to vanish their account. This presents every time this Settings row is tapped.

### Test Steps
1. Log into an account
2. Switching between each theme, tap Vanish account and confirm the modal view renders correctly
3. Tapping Continue should then present the vanish account form. Tapping Cancel or pulling to dismiss should just return you to the Account view.


tonisevener (approved):
Working great!

PR #4367, opened 2022-10-19
Talk pages other error states

mazevedofs:
**Phabricator:** Partially https://phabricator.wikimedia.org/T312314

### Notes
* <strike>This PR was based on the other error states PR branch, PR https://github.com/wikimedia/wikipedia-ios/pull/4361 should be merged before merging this one, and the base branch should be updated to main </strike>
* This PR adds more error messages to other error states on the new Talk Pages
* It does not cover errors when creating new topics, this will be a follow-up PR

### Test Steps
1. Go to the new talk pages, try to subscribe or post a reply with no internet connection
2. Make sure an alert shows notifying the top of the screen as soon as the connection is out
3. Make sure an alert shows when trying to subscribe or post a reply
4. You can meddle with the API call when replying to force an unexpected server error to show

PR #4374, opened 2022-10-24
Talk page new topic error

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T312314

### Notes
* This PR is based on https://github.com/wikimedia/wikipedia-ios/pull/4367, do not merge before 
* We're adding some more error banners when posting topics and when some menu actions fail
* There's an alert view adjustment made specifically for the `TalkPageTopicComposeViewController` since it was causing the alert to display incorrectly

### Test Steps
1. Go to a talk page, create a new topic, and with the internet connection interrupted, try to post
2. A banner should display on the top of the screen
3. Do the same with the menu actions, this will display an alert


Wikipedia/Code/TalkPageViewController.swift
```diff
@@ -360,10 +364,17 @@ class TalkPageViewController: ViewController {
...
+        let alert = UIAlertController(title: TalkPageLocalizedStrings.unexpectedErrorAlertTitle, message: TalkPageLocalizedStrings.unexpectedErrorAlertSubtitle, preferredStyle: .alert)
+        let action = UIAlertAction(title: CommonStrings.okTitle, style: .default)
+        alert.addAction(action)
+        present(alert, animated: true)
+    }
     
     fileprivate func pushToDesktopWeb() {
         guard let url = viewModel.siteURL.wmf_URL(withPath: "/wiki/\(viewModel.pageTitle)", isMobile: false) else {
-            // TODO: Error banner
+            showUnexpectedErrorAlert()
```
mazevedofs, line 377: I'm adding the "Unexpected error" alert here, if it's something that only a connection error could cause, I'm happy to replace it with a "No connection" alert

Wikipedia/Code/TalkPageViewController.swift
```diff
@@ -854,7 +862,14 @@ extension TalkPageViewController: TalkPageTopicComposeViewControllerDelegate {
             case .failure(let error):
                 DDLogError("Failure publishing topic: \(error)")
                 composeViewController.setupNavigationBar(isPublishing: false)
-                // TODO: Display failure banner on topic compose VC
+                let title = TalkPageLocalizedStrings.newTopicFailedAlertTitle
```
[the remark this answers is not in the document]
tonisevener: Update: it looks like you might want to pass along the `composeViewController` parameter and present the server error on that, to avoid the console error:

```
    fileprivate func showUnexpectedErrorAlert(on viewController: UIViewController) {
        let alert = UIAlertController(title: TalkPageLocalizedStrings.unexpectedErrorAlertTitle, message: TalkPageLocalizedStrings.unexpectedErrorAlertSubtitle, preferredStyle: .alert)
        let action = UIAlertAction(title: CommonStrings.okTitle, style: .default)
        alert.addAction(action)
        viewController.present(alert, animated: true)
    }
```

Call for replies with: ` self.showUnexpectedErrorAlert(on: self)` and topics with `self?.showUnexpectedErrorAlert(on: composeViewController)`

Wikipedia/Code/TalkPageViewController.swift
tonisevener: Thanks, I forgot about these. I don't think either error you've done so far quite fits here, since it doesn't really have anything to do with an internet connection or a server error. This is more a fallback in case for some reason it cannot create a URL to push the user to (like say, maybe there's a special character that causes the URL init to fail).

Can you call `showGenericError()` instead here? It's a basic error method in the superclass that displays a "Something went wrong, please try again later." banner.

tonisevener (changes requested):
It's looking good! I just have a couple of suggestions.

Wikipedia/Code/TalkPageViewController.swift
mazevedofs: Good idea, done!

Wikipedia/Code/TalkPageViewController.swift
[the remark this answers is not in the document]
mazevedofs: Great catch, these are also done!

tonisevener (approved):
Looks good! I fixed a few things that trickled in from merge conflicts (project warning from RTL PR, hide Settings button from #4367). I also shifted a couple of lines in df65773 so that the new topic spinner stops if the server error displays. Let me know if you disagree, thanks!

PR #4376, opened 2022-10-27
Add find in page to talk pages

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T319470 (partially)

### Notes
Adds find in page to talk pages. 

More refinement work in a forthcoming PR remains, including:
- More precise scrolling to active search results when in a long blob of text
- Dismissing the find in page bar when the user performs a competing action (like tapping to post a reply)

### Test Steps
1. Open a talk page and tap the find in page tab bar item
2. Try searching for a term, then tapping the keyboard bar actions (next, previous, clear, etc.)
3. Change themes and try again, confirming there are no render issues


Wikipedia/Code/TalkPageViewController+FindInPage.swift
```diff
@@ -0,0 +1,104 @@
...
+        talkPageView.collectionView.reloadData()
+
+        // TODO: - Fine tune position when scrolling
+        switch result.location {
+        case .topicTitle(topicIndex: let index, topicIdentifier: _):
+            let indexPath = IndexPath(row: index, section: 0)
+            talkPageView.collectionView.scrollToItem(at: indexPath, at: .top, animated: true)
+        case .topicLeadComment(topicIndex: let index, replyIdentifier: _):
+            let commentViewModel = viewModel.topics[index].leadComment
+            scrollToComment(commentViewModel: commentViewModel, animated: true)
```
tonisevener, line 55: Just a heads up, unfortunately my `scrollToComment` method I made only considers comment views beneath the lead comment. So you might need to figure out an alternative way to scroll to the lead comment, or modify `scrollToComment` to consider the lead comment text view.

Wikipedia/Code/TalkPageViewController+FindInPage.swift
```diff
@@ -0,0 +1,104 @@
...
+        case .reply(topicIndex: let topicIndex, topicIdentifier: _, replyIndex: let replyIndex, replyIdentifier: _):
+            let topicViewModel = viewModel.topics[topicIndex]
+            topicViewModel.isThreadExpanded = true
```
tonisevener, line 58: Okay, actually you might need a reloadData call after this for the cell to expand 😅 I think it might be more simple at this point to just open all the threads (however the overflow menu button is working) as soon as the Find in Page is button is tapped instead of disclosing them as they navigate through the results, since `reloadData()` here seems to mess with the scrolling.

Wikipedia/Code/TalkPageViewController+FindInPage.swift
```diff
@@ -0,0 +1,104 @@
...
+
+    // MARK: - Overrides
+
+    override var canBecomeFirstResponder: Bool {
+        return findInPageState.keyboardBar != nil
+    }
+
+    override var inputAccessoryView: UIView? {
+        return findInPageState.keyboardBar
+    }
```
tonisevener, line 13: I have a few suggestions here to ensure the find in page bar plays nicely with the reply compose view.

Here I'm trying to keep the input accessory view from appearing while user is typing a reply:

```
    override var canBecomeFirstResponder: Bool {
        
        if replyComposeController.contentView == nil {
            return findInPageState.keyboardBar != nil
        }
        
        return false
    }
    override var inputAccessoryView: UIView? {
        if replyComposeController.contentView == nil {
            return findInPageState.keyboardBar
        }
        
        return nil
    }
```

Somewhat related - TalkPageViewController overrides a method to fix the content inset so that cells do not display beneath the reply compose view. But this is messing with the find in page content inset, so can you update that method to be:

```
override var keyboardIsIncludedInBottomContentInset: Bool {
    if replyComposeController.contentView != nil {
        return false
    } else {
        return true
    }
}
```

And feel free to refactor `replyComposeController.contentView != nil` into something like `replyComposeController.isShowing`, for code clarity.

tonisevener (changes requested):
This is great to see! Nice job working with the existing search view.

I have a few actual code review comments, and then the rest are suggestions I tinkered with for getting the scrolling to work a bit more predictably. Let me know what you think!

Wikipedia/Code/NSMutableAttributedString+Highlight.swift
```diff
@@ -0,0 +1,31 @@
...
+    ///   - backgroundColor: the color to use in the background for matches
+    ///   - foregroundColor: the color to use as the foreground for matches
+    ///   - targetRange: if nil, highlight all matches in `self`. If non-nil, only highlight matches in the target `NSRange`
+    /// - Returns: the attributed string with highlight attributes added
+    func highlight(_ text: String?, backgroundColor: UIColor, foregroundColor: UIColor? = nil, targetRange: NSRange? = nil) -> NSMutableAttributedString {
+        guard let text = text else { return self }
+        
+        let attributedString = self
+
+        var ranges = (string as NSString).ranges(of: text)
```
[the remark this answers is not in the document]
staykids: I agree there are more performant ways to construct this. The reason I went with this particular approach though is to prevent misuse of the target range. I didn't want to allow a engineer to accidentally provide a `NSRange` that is potentially out of range for the string, or just provide an arbitrary range that doesn't actually contain the search text.

But I acknowledge this may be poor or confusing method design on my part trying to consolidate this highlight code in one place rather than potentially breaking it up more. I don't mind changing this or breaking it up somehow if you'd like as long as we're still providing some of that in-built range safety.

What do you think?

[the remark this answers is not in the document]
tonisevener: @staykids Ah! That makes sense, I didn't think about potential misuse like that. I think it's fine to leave as-is then, and revisit if we notice any performance problems.

Wikipedia/Code/TalkPageCellCommentView.swift
```diff
@@ -97,16 +97,47 @@ final class TalkPageCellCommentView: SetupView {
         replyDelegate?.tappedReply(commentViewModel: viewModel)
     }
 
+    // MARK: - Find in page
+
+    private func applyTextHighlightIfNecessary(theme: Theme) {
+        let activeHighlightBackgroundColor: UIColor = .yellow50
+        let backgroundHighlightColor: UIColor!
```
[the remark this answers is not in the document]
staykids: Updated - thanks!

Wikipedia/Code/TalkPageFindInPageSearchController.swift
```diff
@@ -1,21 +1,21 @@
...
+final class TalkPageFindInPageSearchController {
 
     // MARK: - Nested Types
 
     struct SearchResult {
 
         enum Location {
             // Term was located within the topic's title for topic at associated index
-            case topicTitle(topicIndex: Int)
+            case topicTitle(topicIndex: Int, topicIdentifier: ObjectIdentifier)
```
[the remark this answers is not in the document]
staykids: I added them because I had a hunch they'd be useful for scrolling to specific positions. They aren't really used much now and I'm happy to remove them if they don't end up being necessary, but I'd prefer to wait until this feature is closer to completion before doing so to avoid repeated work if they are useful.

Wikipedia/Code/TalkPageViewController+FindInPage.swift
```diff
@@ -0,0 +1,104 @@
...
+        talkPageView.collectionView.reloadData()
+        resignFirstResponder()
+    }
+
+    // MARK: - Scroll to Element
+
+    func scrollToFindInPageResult(_ result: TalkPageFindInPageSearchController.SearchResult?) {
+        guard let result = result else { return }
+
+        talkPageView.collectionView.reloadData()
```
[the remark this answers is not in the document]
staykids: Great idea - feels much nicer to use the scalpel that is retheming the visible cells over the hammer that is `reloadData` here. Updated.

Wikipedia/Code/TalkPageViewController+FindInPage.swift
staykids: Thanks for the heads up. Personally I view my logic here as in progress. Added that TODO previously as I intended to come back to refining this scrolling in a future PR.

Wikipedia/Code/TalkPageViewController+FindInPage.swift
staykids: Got it – updated to open all threads when find in page is first tapped. Personally, I think this behavior of expanding _all_ threads when entering find in page is a little heavy, but it's nice to have better scrolling. In a future improvement to this, I think it might be better if we can move this thread expansion earlier in the process after the user has completed a search and keep the behavior to only opening the threads with matches.

Wikipedia/Code/TalkPageViewController+FindInPage.swift
staykids: > I have a few suggestions here to ensure the find in page bar plays nicely with the reply compose view.

Unless I'm missing something, I think it may just be better to completely dismiss the find in page view in cases where the user interacts with view actions/controls that take over the page (like a reply, or pushing to a new view). I'd originally intended to do this in a follow-up PR, but I went ahead and added it here. Let me know what you think!

> TalkPageViewController overrides a method to fix the content inset so that cells do not display beneath the reply compose view.

Ahhh thank you, this was irritating me and I was incorrectly blaming `ViewController` for doing something wrong without digging into it. Updated!

staykids:
@tonisevener Thanks for the great suggestions - updated and ready for another look!

Wikipedia/Code/TalkPageViewController+FindInPage.swift
tonisevener: > Unless I'm missing something, I think it may just be better to completely dismiss the find in page view in cases where the user interacts with view actions/controls that take over the page.

Ah sorry, didn't catch that it would have been a part of that functionality. It's working well, thanks for doing it early!

Wikipedia/Code/TalkPageViewController+FindInPage.swift
tonisevener: > In a future improvement to this, I think it might be better if we can move this thread expansion earlier in the process after the user has completed a search and keep the behavior to only opening the threads with matches.

Definitely agreed!

tonisevener (approved):
Looking good! I had to adjust a little bit from some merge conflicts from `main`. We are now letting talk page items display that have no replies, so that caused some adjustments in the view models. `main` is also utilizing the `removingInitialNewlineCharacters` method in more places, which threw off some of the ranges for the active highlight.

I patched these up as best I could, but it could probably use more attention with the next PR. I'm going to go ahead and merge to get the current functionality in though.

PR #4378, opened 2022-10-28
Talk pages - fix various bugs

tonisevener:
**Phabricator:** 
N/A

### Notes
These fix a few bugs that I found when testing. Descriptions are in test steps.

### Test Steps
1. 32f8599 With the reply compose view open, background the app and increase the dynamic type font size in iOS Settings. Foreground. Confirm text size changes (it doesn't in `main`).
2. 71e6d5c With the full screen coffee roll view open, background the app and increase the dynamic type font size in iOS Settings. Foreground. Confirm text size changes (it doesn't in `main`).
3. c23294d & 58615ff Sometimes the cells would spit out console constraint errors and size incorrectly, usually upon rotation and when they were configured in an alternative way to handle cells without replies or subscribe buttons. This seems mostly fixed. I sometimes still see constraint errors but it seems to happen less often, and I don't see any more layout issues.
4. 58f83a4 My talk page at en.wikipedia.org/wiki/User_talk:Tsevener would strangely show a blank cell at the top (because the API gives us essentially an empty item). This filters those items out.
5. bf71d70 There's a waiting period after posting reply or a topic when the talk page would refresh. I added the progress loading bar during this so that the app doesn't feel stuck.


mazevedofs (approved):
Looking great!

PR #4382, opened 2022-10-31
Address vanishing view design feedback

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T320758#8357720

### Notes
* Updates element spacing to be 16 pts
* Adjusts leading/trailing padding to be 16 pts
* Updates font size for form elements to 16 pt
* Reduces height of additional information text view
* Updates `Send request` button stack to be pinned at bottom of screen (I confirmed with design that this change is ok)

I removed the `GeometryReader` as it didn't seem necessary anymore, but I didn't originally create this view so please let me know if it's actually still necessary for some reason. Also, GitHub's diff isn't particularly helpful here, so may be work using Xcode's code review feature to more precisely see the differences.

### Test Steps
1. Open vanish account view, confirm spacing/sizing changes
2. Change theme, confirm vanish account view still renders as expected

### Screenshots
vanishing_view_updates

### DIFF
--- a/Wikipedia/Code/VanishAccountContentView.swift
+++ b/Wikipedia/Code/VanishAccountContentView.swift
@@ -82,7 +82,7 @@ struct VanishAccountContentView: View {
                                 .padding([.top], 5)
                             TextView(placeholder: LocalizedStrings.additionalInformationFieldPlaceholder, theme: theme, text: $userInput.text)
                                 .padding([.leading, .trailing], 20)
-                                .frame(maxWidth: .infinity, minHeight: 100)
+                                .frame(maxWidth: .infinity, maxHeight: 50)
                             Spacer()
                                 .frame(height: 12)
                         }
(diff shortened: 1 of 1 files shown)


staykids:
@mazevedofs Design has approved the visual changes and this is ready for code review. Lemme know if you have any questions!

mazevedofs:
Looking great! Just one thing - when using the app in landscape mode, the button overlaps the content. The view functionality is not affected, so maybe it's worth checking with design.

IMG_0069

staykids:
> Looking great! Just one thing - when using the app in landscape mode, the button overlaps the content. The view functionality is not affected, so maybe it's worth checking with design.

@mazevedofs Thanks for looking at this! As long as you're able to 1. scroll the view to reveal the content hidden behind the button and 2. when tapping the additional information field to reveal the keyboard the button disappears, this has been approved by design.

PR #4384, opened 2022-11-02
Talk pages - banner design review feedback

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T312314#8350922
https://phabricator.wikimedia.org/T311639#8280839

### Notes
These are design review tweaks for the alerts presented on talk pages. They dipped into wider-reaching changes in e139b2e and 9997c9a. One is to add a shadow to all banners, which would be easy to revert if design decides against it. The other is reworking the presentation constraints and view controller picking to better consider safe area. I could not figure out a low-touch way around some of the layout bugs we were seeing, and this probably is more stable approach in the medium-term (though, I think we should rewrite it entirely in the long term). I tested a few banners throughout the app and haven't noticed any glaring issues. I'll also call this change out in the tickets so that QA can keep an eye out for regressions here.

### Test Steps
1. Confirm subscription banner text is no longer blue for the light theme.
2. Confirm there's some bottom padding with subscription banners on devices that have a home button (I tested on iPhone 7).
3. Confirm you now see a bottom shadow on top banners, and a top shadow on bottom banners.

### DIFF
--- a/Wikipedia/Code/TalkPageViewController.swift
+++ b/Wikipedia/Code/TalkPageViewController.swift
@@ -490,7 +490,7 @@ class TalkPageViewController: ViewController {
         if UIAccessibility.isVoiceOverRunning {
             UIAccessibility.post(notification: UIAccessibility.Notification.announcement, argument: title)
         } else {
-            WMFAlertManager.sharedInstance.showBottomAlertWithMessage(title, subtitle: nil, image: UIImage(systemName: "exclamationmark.circle"), type: .custom, customTypeName: "subscription", dismissPreviousAlerts: true)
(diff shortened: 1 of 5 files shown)


mazevedofs (commented):
This is looking great, the weird safety area layout spacing is gone! Just one question, the sepia theme the banner has the default coloring, is this intended? (It is like this in main, so I'm just double checking)

![Simulator Screen Shot - iPhone 14 Pro - 2022-11-03 at 12 39 42]
(https://user-images.githubusercontent.com/31116510/199767286-b245af53-24fb-423e-9690-5ccbbc5c9aec.png)

mazevedofs (approved):

tonisevener:
> This is looking great, the weird safety area layout spacing is gone! Just one question, in the sepia theme the banner has the default coloring, is this intended? (It is like this in main, so I'm just double checking)
> 
> Simulator Screen Shot - iPhone 14 Pro - 2022-11-03 at 12 39 42

Good catch on that - I'll ask! I think it's always been a white background for Sepia, strangely. But this seems like I good time to update it while we're here.

PR #4385, opened 2022-11-02
Talk page voiceover improvements part deux

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T318611

### Notes
These are some follow-ons to the original talk page PR (#4383). Please merge after that PR is approved and merged.

5d548a0 - Adjusted ordering so that the topic title is read first instead of the subscription button. This felt more natural to me.
d3da4b4 - Added the .header trait to some elements. This will allow users to navigate quickly to each topic using the Rotor headings item.
fa6dfbab & 249c5d6 - There's a two-finger "Z" pattern swipe that users can use to back out of things. Here I'm adding support for that in the reply compose and topic compose modals. There's also some additional work here to highlight a better element after closing these modals, so the user doesn't lose their place.
5ce3555 - I'm adding a label to the reply depth indicator view. I don't love this solution because it doubles-up items they have to swipe through, so I'm interested to hear any better thoughts on this!
f76c8cd - The project name on user talk pages wasn't accessible, due to a layout issue (the project container has a width of zero). I struggled to actually fix the layout problem, so instead I'm embedding the project name in the type label. I decided to keep it this way for article talk too, just for the extra info and consistency, but happy to remove it if that feels odd.
6ab8739 - Just some slight adjustments for the placeholder label to be read when the reply compose and topic compose text views are highlighted.
695beb7 - The overflow menu button mysteriously says "More" (maybe that's built into SFSymbols?), which isn't bad, but I decided to update it to be a bit more specific.

Also, feel free to suggest better accessibility labels! Thanks!

### Test steps:
I would follow along with the commit descriptions above to test.


mazevedofs (approved):
Working great! The "z" like gesture is a great one, there's so much I need to learn about the available accessibility features!

PR #4393, opened 2022-11-09
Talk pages - find in page adjustments

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T319470

### Notes
This is our first round of find in page adjustments. The goal for this PR is to enable more precise scrolling to the particular rect of the search result.

The main work of this ability is in 52cd406. The other commits I added were to fix additional surrounding bugs, but mostly not changing the original approach. A couple of notes on the larger commits:

bd62f42 - I think one of the reasons we would occasionally have missed active highlights is because we were using different base text for the searching vs. the highlighting. The text views used our `byAttributingHTML` string extension to strip html, and the searching methods used our `removingHTML` string extension method. This commit refactors much of this so that both the views and search logic lean on the `byAttributingHTML` extension to strip html before displaying.
eda9313 - This adjusts the logic slightly from the original 52cd406 WIP commit to restore animations. I tried to tack it on at the end so it would be easily reversible, in case @staykids wants to try a different approach in a followup PR.

### Test Steps
1. Open find in page, confirm traversing results scrolls to the specific highlighted rect.

**Note:** I still occasionally see some incorrect scrolling rects on smaller screens, I suspect due to a layout timing thing. I think this can be handled in a followup PR.


mazevedofs (approved):
Working great!

PR #4395, opened 2022-11-10
Talk pages - add source property for future logging

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T321841

### Notes
This PR adds the plumbing for eventually tracking the source for talk pages. Since TalkPageViewController isn't usually directly instantiated in the context (besides in Account), I had to add it as a part of the `userInfo` dictionary for our NSUserActivities. So far it's only getting passed into talk pages through our router, but now it should be easy to hook into this and do it for other screens in the future like ArticleViewController.

### Test Steps
1. I suggest adding some console logging to print out the source, like `print("‼️‼️‼️source:\(viewModel.source.rawValue)‼️‼️‼️")` in TalkPageViewController's viewDidLoad method.
2. Visit talk pages through various methods (Notifications Center, Account, Article Footer, Search, a talk link in talk pages, a talk link in user pages, as well as loading a talk page link in Safari and tapping the "Open in App" banner (if necessary) to deep link. Confirm the source value printed matches up with what the source screen was.


staykids (approved):
👍🏽

PR #4396, opened 2022-11-15
Fix build server

tonisevener:
**Phabricator:** 
N/A, but caused by https://phabricator.wikimedia.org/T319206

### Notes
Fixes the build server error:

> Invalid Info.plist value. The value for key UIApplicationShortcutWidget must be the bundle identifier of a Today extension in the app.

I missed a change needed when working on #4391. We removed two of our old-style redundant widgets because we started getting deprecation warnings, and suppressed the warnings on the continue reading widget. Updating this flag to our remaining Continue Reading widget ID fixes it.

### Test Steps
Build has already been deployed from this branch, so testing is done.

### DIFF
--- a/Wikipedia/Experimental-Info.plist
+++ b/Wikipedia/Experimental-Info.plist
@@ -73,6 +73,8 @@
 		<string>org.wikimedia.wikipedia.article</string>
 		<string>org.wikimedia.wikipedia.searchresults</string>
 	</array>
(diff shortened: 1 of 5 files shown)


staykids (approved):
👍🏽 Phab task to eventually remove this key completely: https://phabricator.wikimedia.org/T323167

PR #4398, opened 2022-11-16
Import Shared Reading Lists

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T322158

### Notes
This PR imports a shared reading lists via a deep link. There are still a few remaining things up in the air (marked with a TODO comment), but this is close enough to begin PR review.

It's probably easiest to review this work commit-by-commit. Events logging will be added in a follow up PR.

### Test Steps

Note that this implementation isn't exactly like the task description. It has been demo'ed to Product & has the go-ahead.

1. Run PR on device. Visit the example [import landing page] (https://en.m.wikipedia.beta.wmflabs.org/wiki/Special:ReadingLists?limport=eyJsaXN0Ijp7ImVuIjpbNTk4NzQsMzE4ODMsMjQ4NjgsMTQzODFdLCJydSI6WzU5ODc0LDMxODgzLDI0ODY4LDE0MzgxXX19) (taken from [T316822#8366987](https://phabricator.wikimedia.org/T316822#8366987)) link in Safari. The app should automatically open, or you should see the universal links banner at the top with a button to open the Wikipedia app. If you don't, on your test device go to iOS Settings > Developer > Toggle "Associated Domains Development", then try again.
2. Once you get the app successfully opened, confirm you see the "Import shared reading list" modal (Mock 7 in the task).
3. Enter a reading list name and description, then tap "Create reading list".
4. Confirm you are navigated to the reading list detail screen. (Mock 9)
5. After 5 seconds, confirm a survey prompt appears. (Mock 10)


WMF Framework/Router.swift
```diff
@@ -69,6 +76,15 @@ public class Router: NSObject {
             return project.supportsNativeUserTalkPages ? .userTalk(url) : nil
         case .special:
             
+            if title == "ReadingLists",
```
mazevedofs, line 79: More of a curiosity, will this behave like other namespaces (that are language-specific)?

mazevedofs (approved):
It's looking great!

WMF Framework/Router.swift
tonisevener: Good eye - I think line 65  above (`namespaceAndTitleOfWikiResourcePath`) should be flexible enough to handle the translated "Special" namespace, so if links come through as `es.wikipedia.org/wiki/Especial:ReadingLists?limport=nnn` it should still flow through this router correctly. I haven't tested a full end-to-end flow on a non-EN deep link yet and was planning on waiting till the landing page was in production first.

With "ReadingLists" being the title - I am unsure if that will stay as this translation or not - I'll check with Android. This (and probably other areas) will break if not.

tonisevener: Update - we can expect the "ReadingLists" title to remain static for all wikis.

tonisevener:
Going to go ahead and merge this - we can handle the TODOs in a follow up PR.

PR #4401, opened 2022-11-16
Improve find in page scroll accuracy with topic comments

staykids:
**Phabricator:** 
- https://phabricator.wikimedia.org/T319470 (partially)
- https://phabricator.wikimedia.org/T310288#8389567 (partially)

### Notes
Previously, sometimes using find in page for a term deeply nested in a reply would not accurately scroll the user to that element. I believe this occurred because `talkPageCell(indexPath: IndexPath)` was sometimes returning nil (as expected per the API if the cell isn't yet constructed) for that deeply nested element's topic cell. So far, the changes here seem to most effectively improve the find in page next/previous scroll accuracy with the least amount of change. It sacrifices some animation, and I wasn't able to see noticable improvements to partial header scroll weirdness when using the programmatic scrolling flag technique seen elsewhere, but in the cases I've tried so far the actual next/prev element scrolling has seemed accurate.

Certainly happy to continue iterating on this if it seems to be missing a case.

### Test Steps
1. Find a talk page that has a mix of thread types, where some contain no replies and some contain multiple replies to a topic (the `World War II` talk page is one I've used for some testing)
2. Search for terms that may appear multiple times in the page (`war` is a good one on that talk page), confirm next/previous scrolls the element to be visible
3. Search for a term or phrase that only appears in a reply to a topic, and confirm search and or next/previous scrolls the element to be visible


tonisevener (approved):
Awesome, this is working great! I'm glad we were able to keep some animations, if not all. I have a couple of follow up things that would be good to look into. I'll spin off a couple of tasks for it.

1. I notice if there's only one search result, I can't scroll to it.
2. This has nothing to do with find-in-page, but we should officially have a ticket to investigate the sticky header in our custom navigation bar code. If I recall correctly, that `isProgramaticallyScrolling` flag only seemed to help me when I _didn't_ use animations, so it's possible your partial animations are conflicting with it. Still, I'd rather us have a different fix that didn't require disabling animated scrolling.

PR #4403, opened 2022-11-17
Attempt to fix RMessageView crash

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T323163

### Notes
This is an attempt to fix our RMessageView crashes. I cannot reproduce it, but I suspect the [superview reference](https://github.com/wikimedia/wikipedia-ios/blob/main/Wikipedia/Third%20Party%20Code/RMessage/RMessageView.m#L461) when setting up the constraints is occasionally nil, because when I replace that parameter with nil I see the same crash and call stack.

I think the better fix would be to have the superview set up all of these constraints, but that's too big of a change at this point in the release. So for now I'm just adding some defensive work to stop the display entirely if the superview is nil to avoid the crash.

### Test Steps
1. Regression test these toasts - subscribe/unsubscribe to a talk page topic, go to airplane mode, and confirm messages display fine as before.
2. Force a situation where the [new code](https://github.com/wikimedia/wikipedia-ios/blob/T323163/Wikipedia/Third%20Party%20Code/RMessage/RMessageView.m#L261-L266) is run with comments:

```
      //if (self.superview) {
      //    [self setupLayout];
      //} else {
          [self removeFromSuperview];
          return nil;
      //}
```

and repeat Step 1. Confirm toasts no longer display, and there are no crashes.

### DIFF
--- a/Wikipedia/Third Party Code/RMessage/RMessageView.m
+++ b/Wikipedia/Third Party Code/RMessage/RMessageView.m
@@ -253,7 +253,18 @@ static NSMutableDictionary *globalDesignDictionary;
     if (designError) return nil;
 
     [self setupDesign];
-    [self setupLayout];
+    self.translatesAutoresizingMaskIntoConstraints = NO;
+    [self addToSuperviewForPresentation];
+      
+      // This check fixes an NSLayoutConstraint crash
+      // https://phabricator.wikimedia.org/T323163
+      if (self.superview) {
+          [self setupLayout];
+      } else {
+          [self removeFromSuperview];
+          return nil;
+      }
+      
     if (dismissingEnabled) {
       [self setupGestureRecognizers];
     } else {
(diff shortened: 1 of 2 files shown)


Wikipedia/Third Party Code/RMessage/RMessageView.m
```diff
@@ -451,13 +462,7 @@ - (void)setupDesign
 
 - (void)setupLayout
 {
-  self.translatesAutoresizingMaskIntoConstraints = NO;
-
-  // Add RMessage to superview
-  [self addToSuperviewForPresentation];
-
   if (self.messagePosition != RMessagePositionBottom) {
-    [self layoutIfNeeded];
```
tonisevener, line 460: This is an unrelated cleanup thing - I think it was only needed with the way the old constraints were set up, which were modified in #4384. I don't see a need for this call at this point, plus we're calling `layoutIfNeeded` it again further down in the method after the constraints are set up. Removing it seems to cause no regressions in my testing.

staykids (approved):
Nice – this bit of defensive programming doesn't seem unreasonable in an attempt to resolve the crash. Haven't yet noticed regressions on iPhone or iPad. Looking at this `RMessage` library again makes me want to pick up my toast library work again the next time it makes sense to...

PR #4406, opened 2022-11-18
Base survey prompt URLs and translations off of primary app language

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T322158

### Notes
This PR is to change the survey URL, survey prompt translations, and survey footer privacy policy link to be based on the user's primary app language, per task requirements.

For translations - I added them manually into the strings files in the Localizations subdirectory (and ran the import script so they copied into the iOS Native Localizations subdirectory), based on the spreadsheet mentioned in [this comment](https://phabricator.wikimedia.org/T322158#8421741). I also made those changes on translatewiki's side, so the next PR from them won't strip them out.

### Test Steps
1. Trigger survey prompt while various languages are the primary (see "Feedback Form Links" section in Phab task). Confirm they go to the correct privacy policy link and survey link. Confirm translations on the survey prompt are in the correct language. It may help to comment out the user defaults [check](https://github.com/wikimedia/wikipedia-ios/commit/0dabe51994250ee8a95d596664a60651c7606596#diff-e3c2665359c4b1974fe1ea867c37d077062051b92f0cdb08a9e44d1a1e5e183eR355) while testing.
2. Also trigger it with a primary language not in this list. Confirm it takes user to the English survey link and privacy policy, and displays strings in English.

### Screenshots
See https://phabricator.wikimedia.org/T322158#8426040.


staykids (commented):
👍🏽 Looks good! I double checked that the form links resolved, were publicly available, and tried a few of them in app after switching my primary language.

Currently this is still just blocked by Spanish and Urdu translates forms. If we choose to move forward without those, we need to the remove the two `importSurveyURLs` dictionary entries  for those language codes otherwise the survey action button won't properly fall back to an English form for those two languages.

Wikipedia/Localizations/ar.lproj/Localizable.strings
```diff
@@ -506,6 +506,11 @@
 "icon-shortcut-random-title" = "مقالة عشوائية";
 "icon-shortcut-search-title" = "البحث في ويكيبيديا";
 "image-gallery-unknown-owner" = "المؤلف غير معروف.";
+"import-shared-reading-list-survey-prompt-button-cancel" = "ليس الآن";
+"import-shared-reading-list-survey-prompt-button-take-survey" = "المشاركة في الاستطلاع";
+"import-shared-reading-list-survey-prompt-footer" = "\"اطلع على $1بيان الخصوصية$2 الخاص بنا. الاستطلاع مدعوم من قِبل جهة خارجية. اطلع على $3سياسة الخصوصية$4 الخاصة بهم.\"";
```
staykids, line 511: Here's an example of the included in string quotation marks.

staykids (approved):
Nice! Tested with a few languages and works as expected with LTR and RTL languages. Thanks so much for those screenshots in Phab.

I'm a little confused about the surrounding quotation marks included in some translated `import-shared-reading-list-survey-prompt-footer` strings (see example below) – maybe these were just extraneously pasted into the sheet to begin with? – but they match the translation sheet so I'm happy to merge as-is.

Wikipedia/Localizations/ar.lproj/Localizable.strings
tonisevener: Yeah I know what you mean - originally I stripped them out as I was copying them over, then second-guessed myself. The French and German marks made me think it was deliberate, but I'll confirm. Thanks for checking!

tonisevener: Update - I confirmed with Android that we should remove these.

staykids: 👍🏽 will merge when tests pass!

PR #4409, opened 2022-12-01

translatewiki:
Translation updates


tonisevener (approved):
This is the first update after manually adding translations in #4406.

Some "Fuzzy" flags were added, because I tweaked the English string in the project after adding the latest translations in translatewiki, which is fine. Someone also changed a couple of the translations for French and German afterwards, which I think is also fine.

PR #4410, opened 2022-12-01
Talk pages - Add logging

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T321841

### Notes
This is the client-side portion of our talk pages logging. We cannot fully test end-to-end until the server-side patches are merged, but this is enough to confirm our client-side events. Any fixes that are needed after patches are merged can be made in a follow-up PR.

### Test Steps
1. Launch app, log in and enable logging toggle in app Settings or onboarding.
2. Visit a user talk page that you can post to (your own user talk page, for example).
3. Expand a topic, tap new topic + icon, tap an inline reply button, post a new topic or inline reply, and change the talk page language. After each event, you will see a logging in the console that looks like:

> EPC: Scheduling event to be sent to https://intake-analytics.wikimedia.org/v1/events with POST body: {payload}

Confirm `action`, `page_ns`, `primary_language`, `wiki_db`, `is_anon`, and `time_spent` in payload looks correct. Some additional notes:

- `action` type descriptions are:

```
          open_topic:
            User expanded topic
          new_topic_click:
            User selected new topic
          reply_click:
            User selected reply
          refresh (Note, this is used so far):
            User selected refresh
          lang_change:
            User changed Talk Page language
          submit:
            User selected submit new topic or inline reply to a thread
```

- `page_ns` should be "1" for article talk, "3" for user talk.
- `wiki_db` will be the identifier of the wiki that the talk page is on.
- `time_spent` is the number of seconds the user has been on the talk page - I have the timestamp for this measurement starting when viewDidAppear on the talk page is triggered.  

4. Log out, go to an article talk page on Test wiki, and repeat step 3. Confirm console values look okay.

**Note**: Instead of any network call going out, you will see another console message: `Event submitted to 'talkPagesInteraction' but only the following streams are configured: ios.edit_history_compare, ios.notification_interaction`. This is expected until the patches are merged.


Wikipedia/Code/TalkPagesFunnel.swift
```diff
@@ -0,0 +1,87 @@
...
+        }
+    }
+    
+    private struct Event: EventInterface {
+        static let schema: EventPlatformClient.Schema = .talkPages
+        let action: Action
+        let source: Source
+        let page_ns: String
+        let time_spent: Int
+        let wiki_db: String
```
mazevedofs, line 46: Just noting that this prop was changed to wiki_id as suggested on the schema

mazevedofs (changes requested):
Looking good, we just need to update the wiki_id prop

Wikipedia/Code/TalkPagesFunnel.swift
tonisevener: Done!

mazevedofs (approved):

PR #4411, opened 2022-12-01
Talk pages -  add formatting toolbar 

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T315606

### Notes
* This PR adds a simple formatting toolbar to the talk page reply compose view and to the topic compose view controller
* The toolbar is a simple UIView containing the actions to add bold, italics, and link _wikitext_ formatting to the UITextView
* The add image action was not considered for this PR
* The link editing was also not considered

### Test Steps
1. On a talk page, tap reply to topic or create new topic
2. For bold and italics, you can select previously typed text or select the button before typing
3. For link insertion, it will display a new view, search for article and insert it
4. Post your topic and reply, and verify that the formatting works
5. To remove the formatting, select the typed word and click the same button - also works if nothing was typed
6. Make sure the find in page toolbar was not affected


Wikipedia/Code/TalkPageTopicComposeViewController.swift
```diff
@@ -125,6 +125,14 @@ class TalkPageTopicComposeViewController: ViewController {
     weak var delegate: TalkPageTopicComposeViewControllerDelegate?
     
     private weak var authenticationManager: WMFAuthenticationManager?
+
+    private let textFormattingPlainToolbarView = TextFormattingPlainToolbarView.wmf_viewFromClassNib()
+
+    private(set) var inputAccessoryViewType: InputAccessoryViewType?
```
tonisevener: Is this needed? Seems like it might always be type .format.

Wikipedia/Code/TalkPageFormattingToolbarView.swift
```diff
@@ -0,0 +1,155 @@
...
+
+    @objc func didTapItalicsButton() {
+        delegate?.didSelectItalics()
+    }
+
+    @objc func didTapInsertLinkButton() {
+        delegate?.didSelectInsertLink()
+    }
+
+    func configure() {
```
tonisevener, line 123: I would add private around most of these properties and methods in here if you don't intend to call them from the outside.

Wikipedia/Code/TalkPageTopicComposeViewController+TalkPageFormattingToolbar.swift
```diff
@@ -0,0 +1,42 @@
...
+        dismiss(animated: true)
+    }
+
+    func insertLinkViewController(_ insertLinkViewController: InsertLinkViewController, didInsertLinkFor page: String, withLabel label: String?) {
+        insertOrEditLink(page: page, label: label)
+        dismiss(animated: true)
+    }
+
+    func insertOrEditLink(page: String, label: String?) {
+        if let label {
```
tonisevener, line 36: Nice, I didn't know we could do this shorthand!

Wikipedia/Code/TalkPageTopicComposeViewController.swift
```diff
@@ -7,6 +7,8 @@ protocol TalkPageTopicComposeViewControllerDelegate: AnyObject {
 
 struct TalkPageTopicComposeViewModel {
     let semanticContentAttribute: UISemanticContentAttribute
+    let siteUrl: URL
+    let pageTitle: String
```
tonisevener, line 11: Just a heads up - I don't think this is used anywhere.

Wikipedia/Code/TalkPageViewController+TalkPageFormattingToolbar.swift
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    func didSelectInsertLink() {
+        guard let link = Link(page: "", label: "", exists: false) else {
+            return
+        }
+        let insertLinkViewController = InsertLinkViewController(link: link, siteURL: viewModel.siteURL, dataStore: MWKDataStore.shared())
+        insertLinkViewController.delegate = self
+        let navigationController = WMFThemeableNavigationController(rootViewController: insertLinkViewController, theme: self.theme)
+        self.present(navigationController, animated: true)
```
tonisevener, line 24: I think you can remove the `self` here.

Wikipedia/Code/TalkPageTopicComposeViewController+TalkPageFormattingToolbar.swift
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    func didSelectInsertLink() {
+        guard let link = Link(page: "", label: "", exists: false) else {
+            return
+        }
+        let insertLinkViewController = InsertLinkViewController(link: link, siteURL: viewModel.siteUrl, dataStore: MWKDataStore.shared())
+        insertLinkViewController.delegate = self
+        let navigationController = WMFThemeableNavigationController(rootViewController: insertLinkViewController, theme: self.theme)
+        self.present(navigationController, animated: true)
```
tonisevener, line 20: Same here - remove `self`.

Wikipedia/Code/UITextView+Extensions.swift
```diff
@@ -15,4 +15,46 @@ extension UITextView {
...
+                if let selectedSubstring = text(in: selectedRange) {
+                    replace(selectedTextRange ?? UITextRange(), withText: formattingString + selectedSubstring + formattingString)
+                } else {
+                    replace(selectedTextRange ?? UITextRange(), withText: formattingString + formattingString)
+                }
+            }
+        }
+    }
+
+    func addOrRemoveStringFormattingCharacters(formattingString: String, cursorOffset: Int) {
```
tonisevener, line 42: You might be able to just do a count of the `formattingString` within these methods, to avoid the need to pass in a `cursorOffset`.

tonisevener (changes requested):
This is looking so great! It is super exciting to see us take some first steps away from web-based editing. Mostly minor things in code review and a few experience things I'd love to see get in if possible. Let me know if any of these give you trouble:

1. The formatting buttons on the new topic title textfield are acting a little strange - you can tap a button but the wikitext is added to the body text view. I wonder if we should remove the formatting toolbar for the title entirely for now, especially since we seem to style titles as bold on the native-side by default. Might be a good question for design.
2. I notice on Android's editor, after highlighting text and applying a bold or italic, it automatically highlights the text again. This actually makes for much quicker toggling of the formatting. This also may be more of a design question, so we can wait to address it in design review phase if you want.

out

3. For the link wizard, I know we can't do link editing, but for highlighting non-link text, can we have it act a bit more like our section editor? There, the highlighted text is sent into the wizard as a default search term. Then the final wikitext completely replaces the text they originally highlighted. I think that would be a nice bonus that hopefully won't cause you to run into web view / link parsing blocks, but let me know if you do.

Wikipedia/Code/TalkPageTopicComposeViewController.swift
tonisevener: Ignore that ^. It was an old comment from outdated commit.

Wikipedia/Code/UITextView+Extensions.swift
mazevedofs: Done!

Wikipedia/Code/TalkPageTopicComposeViewController+TalkPageFormattingToolbar.swift
mazevedofs: > 1. The formatting buttons on the new topic title textfield are acting a little strange - you can tap a button but the wikitext is added to the body text view. I wonder if we should remove the formatting toolbar for the title entirely for now, especially since we seem to style titles as bold on the native-side by default. Might be a good question for design.

Great catch, I ignored the title field when testing, I removed it from here since it doesn't make sense to allow formatting. 

> 2. I notice on Android's editor (actually on our section editor as well), after highlighting text and applying a bold or italic, it automatically highlights the text again. This actually makes for much quicker toggling of the formatting. This also may be more of a design question, so we can wait to address it in design review phase if you want.
> 
Another good catch, I added the same behavior here.

> 3. For the link wizard, I know we can't do link editing, but for highlighting non-link text, can we have it act a bit more like our section editor? There, the highlighted text is sent into the wizard as a default search term. Then the final wikitext completely replaces the text they originally highlighted. I think that would be a nice bonus that hopefully won't cause you to run into web view / link parsing issues, but let me know if you do.

I'm sending the search term in now (and replacing it with the link syntax once inserted), so it pre-populates the search field, but it still doesn't trigger a search - it requires the user to type in something else. Do you think it is worth the risk of trying to tweak the search controller?
For the editing, I now used logic similar to the one I used on the bold/italics button. There are some limitations to it, for example, if the user selects the term and the label when editing it won't take them to the editing view, but it is working. Maybe that's a limitation we'll have to deal with because of the limitations with wikitext, but I think it's worth testing.

tonisevener: >  but it still doesn't trigger a search - it requires the user to type in something else. Do you think it is worth the risk of trying to tweak the search controller?

Ah I see - it looks like that bug exists in our section editor too. I would say that fix is beyond the scope of this task. We can spin off a separate bug task for that to fix in both places.

Wikipedia/Code/TalkPageTopicComposeViewController+TalkPageFormattingToolbar.swift
```diff
@@ -0,0 +1,118 @@
...
+
+    func didSelectBold() {
+        bodyTextView.addOrRemoveStringFormattingCharacters(formattingString: "'''")
+    }
+
+    func didSelectItalics() {
+        bodyTextView.addOrRemoveStringFormattingCharacters(formattingString: "''")
+    }
+
+    func didSelectInsertLink() {
```
tonisevener, line 15: I know you already mentioned this so just commenting for tracking purposes. It'd be good to DRY up some of the logic in these delegate methods between new topic and new reply. Let me know if you want any thoughts on approaches for that.

Wikipedia/Code/UITextView+Extensions.swift
```diff
@@ -15,4 +15,55 @@ extension UITextView {
...
+    /// Adds formatting characters around selected text or around cursor
+    /// - Parameters:
+    ///   - formattingString: string used for formatting, will surround selected text or cursor
+    ///   - cursorOffset: offset to adjust cursor position, should be equal to the number of characters
+    func addStringFormattingCharacters(formattingString: String) {
+        let cursorOffset = formattingString.count
+        if let selectedRange = selectedTextRange {
+            let cursorPosition = offset(from: endOfDocument, to: selectedRange.end)
+            if selectedRange.isEmpty {
+                replace(selectedTextRange ?? UITextRange(), withText: formattingString + formattingString)
```
tonisevener, line 28: I can't quite place what happened but I did prefer the previous way toggling on and off formatting worked when it was _only_ a cursor. Now it jumps to the end of the formatting ticks, but before it stayed in the middle for easy toggling.

Toggling when text is highlighted is working great for me now though! Hopefully you can restore the single cursor behavior without affecting that.

Wikipedia/Code/TalkPageTopicComposeViewController+TalkPageFormattingToolbar.swift
```diff
@@ -0,0 +1,118 @@
...
+        dismiss(animated: true)
+    }
+
+    func insertLinkViewController(_ insertLinkViewController: InsertLinkViewController, didInsertLinkFor page: String, withLabel label: String?) {
+        insertLink(page: page)
+        dismiss(animated: true)
+    }
+
+    func insertLink(page: String) {
+        let content = "[[\(page)]]"
```
tonisevener, line 69: What do you think about something like this here (and in the other spot)?

```
        var content: String = "[[\(page)]]"
        
        if let selectedText = bodyTextView.text(in: preselectedTextRange) {
            if page != selectedText {
                content = "[[\(page)|\(selectedText)]]"
            }
        }
```

This way users can get some of the piping wikitext to paste without going through the EditLinkViewController. (This also seems to be how our section editor works). Here's what it does:

out

tonisevener (changes requested):
Looking great! A few more comments for this round, I put them all inline this time.

Parsing and pushing to the edit link controller is working pretty well! I'd say the bugs with the piping logic need to be fixed a little bit more before we release it. I didn't test it enough to get a list of must-fix situations but I can do that after this next round. If we run out of time, I think we can temporarily comment out your progress with it and finish it up in a followup PR.

Wikipedia/Code/TalkPageTopicComposeViewController+TalkPageFormattingToolbar.swift
```diff
@@ -26,20 +28,33 @@ extension TalkPageTopicComposeViewController: TalkPageFormattingToolbarViewDeleg
...
+            let index = text.firstIndex(of: "|") ?? text.endIndex
+            let beggining = text[..<index]
+
+            let ending = text[index...]
+
+            let newSearchTerm = String(beggining)
+            let newLabel = String(ending.dropFirst())
+
+            let linkText = doesLinkExist ? newSearchTerm : text
+            let labelText = doesLinkExist ? newLabel : text
```
tonisevener, line 46: I notice this chunk of logic is missing in the other `didSelectInsertLink` method - so there's a bug in the reply textview that is fixed on this new topic textview here. Can you share some of this logic in your new UITextView+FormattingToolbarExtension.swift class?

tonisevener (changes requested):

tonisevener (approved):
🎉 👏  Great work on this one!

PR #4412, opened 2022-12-02
Measure article page load times

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T322563

### Notes
In this PR, we are keeping track of the time it takes for a request from our `mobile-html` endpoint to complete, and sending the min, max, and average values with our session data. I'm trying to avoid logging in the case of any errors or connection failures, as well as if we had to lean on our local permanent cache for any reason (such as Saved Articles) to try to keep the load times a bit more accurate. 

### Test Steps
1. Launch app, enable logging toggle.
2. Visit a few articles, keeping track of how many you have visited.
3. Add a break point [here](https://github.com/wikimedia/wikipedia-ios/commit/857a8a0bd30dec292fd4464e879eba795d1f063f#diff-a8e66e0b1b432cd980bff55f8a5990ab15748817d90decef53560da66701ac9eR47), then background the app. Confirm the `pageLoadTimes` property reflects the number of articles you visited in this session. Inspect the `event` payload, confirm the page load max, min, and average values look correct against what pageLoadTimes contain. These payload values should be integers of milliseconds.
4. Foreground and repeat step 2-3. Confirm the old page load times were cleared out, and your new page load times are being calculated and sent in the event payload.


Wikipedia/Code/Session.swift
```diff
@@ -59,13 +59,14 @@ public class Session: NSObject {
     }
     
     public struct Callback {
+        public typealias UsedPermanentCache = Bool
         let response: ((URLResponse) -> Void)?
         let data: ((Data) -> Void)?
-        let success: (() -> Void)
+        let success: ((UsedPermanentCache) -> Void)
```
staykids, line 65: I find this `let success: ((UsedPermanentCache) -> Void)` change a little confusing from an in-app API perspective, simply because the `callback.success(false)`/`callback.success(true)` calls themselves convey unclear meaning to me in isolation, but I don't have an alternative to suggest.

staykids (approved):
Nice! Seems to work for me as expected on the app side.

PR #4413, opened 2022-12-02
Scale picture of the day widget to address blank widget issue

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T321061

### Notes
This PR addresses an issue where the POTD widget doesn't render an image. For whatever reason, some OS level change appears to have constrained what's possible to display in a widget's SwiftUI `Image` view. Previously, `WidgetKit`/`SwiftUI` seemed to be more forgiving of the dimensionality of images here.

Originally, I had the belief that unusually large image dimensions were causing the crash because the requested image required too much memory (exceeding what's allotted by `WidgetKit`), but it actually seems to be the case that just the _dimensions_ themselves were what was causing issues, not necessarily the image's file size itself. For example, an image with an unusually large width regardless of file size, which may have rendered without issue before, is now simply causing the widget to blank out. Through some experimentation, resizing the image to draw within 1000x1000 pixels seems to allow the images/devices I've tested to render without blanking out. The negative of this approach though, is that depending on device and widget size, the image may appear blurrier than before (because it's not considering the device's scale).

It's worth noting, I continue to have really poor luck with the debugging `WidgetKit` widgets. I also have additional incomplete work that removes the Core Data dependencies to make the widget more independent (like the feature article widget), but I need to hold that work for a future PR because it's becoming more involved than I had hoped.

### Test Steps
1. Delete existing Wikipedia POTD widgets then build/install and open app.
2. Long pressing the home screen and opening the Wikipedia widget gallery, confirm the POTD widget isn't blank.
3. Add a POTD widget at any size. Confirm it actually renders the expected image on your home screen.
4. Delete that POTD widget and add another at a different size.

Additionally if interested, you can try to manually use some previously problematic images that resulted in blank widgets to ensure they work with the new scaling logic. An example of a previous image that resulted in failure is the image featured here: https://en.wikipedia.org/api/rest_v1/feed/featured/2022/11/12. By adding it to the Widget asset catalog, then replacing `cachedImage.staticImage`, `fetchedImage.image.staticImage`, and `UIImage(named: "PictureOfTheYear_2019")` in `PictureOfTheDayWidget.swift` with that image, you should be able to present it in the POTD widget to confirm it renders and isn't blank.


mazevedofs (commented):
I'm still having issues with the large widget, I'm testing on an iPhone 13 mini and an iPad 9th gen. The small and medium widgets are working well in both devices.

staykids:
> I'm still having issues with the large widget

Thanks for reviewing this! I'm not noticing the same on my end. A few questions:
- To confirm - this is following the steps above of removing any existing POTD widgets before adding a large one? There seems to be some existing issue (even on `main`) with multiple POTD widgets not rendering, but with this branch I haven't been able to replicate an individual widget at any size not rendering.
- Are you noticing this on device or simulator, iOS 15 or 16?
- Does tapping the blank widget to launch the app and returning to the home screen show the picture content?
- Are you encountering this with the current POTD image or a test one I mentioned above?

mazevedofs (approved):

PR #4417, opened 2022-12-07
Update supported Wikipedia project languages

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T317844

### Notes
Adds `ami`, `blk`, `dag`, `guw`, `kcg`, `pcm`, `pwn`, `shi` via the `Update Languages` script.

### Test Steps
We can rely on QA for extensive testing (I updated the original task description), but for a couple of the language codes above you can:
1. Search for a language code/language name from above
2. Confirm you can search for/view an article in that language


tonisevener (approved):
Looks good. I noticed some issues with saving articles on the `blk` language, but it appears to be [this](https://phabricator.wikimedia.org/T277599) existing bug. I called it out in the task description for QA.

PR #4424, opened 2022-12-16
Use language inclusive edit help links

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T320738

### Notes
Updates edit help links to language agnostic links rather than EN links. Note that the Watchlist help link was also updated here even though we don't currently display it in app.

### Test Steps
1. Edit an article, continuing up until the Save changes view
2. Tap the Learn more button, confirm it goes to Meta link for Edit summary help
3. Tap the Learn more about minor edits button, confirm it goes to Meta link for Minor edit


tonisevener (approved):
Looks good!

PR #4426, opened 2022-12-20
Enable scrolling on long title article history pages

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T314809

### Notes
Previously, long titled article history pages would not scroll because the navigation bar in the view didn't enable `allowsUnderbarHitsFallThrough`. This adds a `PageHistoryCountsView` to represent the existing implemented XIB, and adds the point inside override to support `allowsUnderbarHitsFallThrough` and allow scrolling the view when interacting with the header.

If there's a better way to handle this, I'm happy to change this PR!

### Test Steps
From Phabricator:
```
- Go to the article: https://en.wikipedia.org/wiki/Cneoridium_dumosum_(Nuttall)_Hooker_F._Collected_March_26%2C_1960%2C_at_an_Elevation_of_about_1450_Meters_on_Cerro_Quemaz%C3%B3n%2C_15_Miles_South_of_Bah%C3%ADa_de_Los_Angeles%2C_Baja_California%2C_M%C3%A9xico%2C_Apparently_for_a_Southeastward_Range_Extension_of_Some_140_Miles within the iOS app
- Navigate to the article's history page
- Scroll down on the history page
```

Confirm in the case above (and any other article history page) that interacting with the navigation bar to scroll works and allows the user to see the rest of the view.


tonisevener (approved):
Looks good!

PR #4429, opened 2022-12-26
Fix peek/pop bug: References should not peek

mcleinman:
**Phabricator:** N/A

### Notes
* 👋🏻 I realized I introduced this bug when working on peek/pop. Sorry!
* Currently, if you long press on an article reference (or note), a user gets a preview of the article they're currently in, which is not intended. This PR suppresses long press preview for references (and notes).
* Additionally, when a user long presses an element that cannot be previewed (like an external URL or [now] a reference), the text they tap grows as if it can be previewed. I changed the completion handler to `nil` (from `nilConfig`), which tells iOS suppresses this behavior. Please let me know if you'd like to keep the current behavior, and I can remove this from the PR.
* Also, I renamed a variable to `newArticleUrl` as we use `articleUrl` for the current VC's article. While the compiler can handle it, I think this improves readability for humans - but feel free to request that I revert that.

### Test Steps
1. Load an article.
2. Use peek/pop on a reference. It should not work.
3. Use peek/pop on an article link. It should work.


tonisevener (approved):
👋 😀 Working great for me! Thanks for this.

PR #4435, opened 2023-01-02
Let URLComponents setters handle encoding

junholee6a:
**Phabricator: https://phabricator.wikimedia.org/T321064** 

### Notes
* Currently, question marks in `href` are unencoded in case they are query string separators (https://github.com/wikimedia/wikipedia-ios/pull/3710)
* However, if there is a question mark in the path (such as in https://en.wikipedia.org/wiki/Who's_the_Man%3F) it must be percent-encoded, or it is interpreted as a query string separator and excluded from the path
* [The setters of URLComponents handle encoding automatically based on what is allowed in each component](https://developer.apple.com/documentation/foundation/urlcomponents/1780144-path), so we can use them to make sure the URL is valid
* As a side effect, it encodes non-ASCII characters in domains using Punycode instead of percent-encoding, so tests were adjusted
* This change means that `CharacterSet.relativePathAndFragmentAllowed` is no longer used anywhere, but I haven't removed it because the underlying `NSCharacterSet.wmf_relativePathAndFragmentAllowedCharacterSet` is still used

### Test Steps
**Unit Tests**
1. Ensure `URLParsingAndRoutingTests` run successfully

**Manual Test**
1. Navigate to this article in the app: https://en.wikipedia.org/wiki/Party_and_Bullshit
2. Click the "[Who’s the Man?](https://en.wikipedia.org/wiki/Who%27s_the_Man%3F)" link
3. The "Who’s the Man?" article should load


junholee6a:
Tests are failing on CircleCI but passing on my local computer, with MacOS 13.1 and XCode 14.2. I'm lost on why it doesn't work

junholee6a:
I realized it's likely because I am not running fastlane locally, so I will reopen this when I get that working

junholee6a:
I was not able to install Ruby 2.6.1, so I ran fastlane with a more recent version locally, but all unit tests passed

junholee6a:
Looks like the tests fail on iOS 15.5. I'll open a new PR if I find a fix

PR #4438, opened 2023-01-04
Match edit summary content language to article language

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T324942

### Notes
* This PR changes the EditSummaryViewController and EditSaveViewController classes to match their content language to the edited article language

### Test Steps
1. Go to an article in any language other than the app's primary language
2. Edit the article to reach the summary view
3. Make sure the content of the view matches the article's language


Wikipedia/Code/EditSaveViewController.swift
```diff
@@ -161,7 +163,7 @@ class EditSaveViewController: WMFScrollViewController, Themeable, UITextFieldDel
 
         setupButtonsAndTitle()
         mode = .preview
-        
+        setupSemanticContentAttibute()
```
tonisevener, line 166: I notice the fine print text views at the bottom don't seem to take effect for RTL (HE example):

IMG_0161

I think the alignment setting becomes overwritten any time the attributed strings are updated when `updateTextViews()` is called, which is called in a couple of different places. When I move this `setupSemanticContentAttribute()` call to the end of the `updateTextViews` method, that seems to fix it. I'm open to other options though!

mazevedofs: Sounds good to me!

tonisevener (approved):
Thanks!

PR #4439, opened 2023-01-04
Fix Notifications Center theming issue on iOS 16

staykids:
**Phabricator:** N/A

### Notes
As of iOS 16, it appears that `List` is no longer backed by a `UITableView`: https://developer.apple.com/documentation/ios-ipados-release-notes/ios-16-release-notes. Because of this, our previous methodology for painting the background color in the Notifications Center filter and inbox views no longer holds. This PR conditionally uses the new `scrollContentBackground` API to hide the background view and apply the `Color` background view on iOS 16.

### Test Steps
Only needs a light visual check:

1. Open in-app Notifications Center on iOS 16 simulator/device
2. Tap either the Filter or Inbox toolbar icons
3. Confirm the view background renders as expected
4. Change app theme and check again
5. Repeat steps on iOS 15 simulator/device

### Screenshots
ba_filters


mazevedofs (approved):
Looking good!

PR #4442, opened 2023-01-08
Remove invalid references in article cache and explore views

junholee6a:
**Phabricator: https://phabricator.wikimedia.org/T317587** 

### Notes
* Currently, there is a problem where the explore page does not display properly after the user clears the article history.
* When article history is cleared by the user, [`MWKDataStore.viewContext.reset()`](https://developer.apple.com/documentation/coredata/nsmanagedobjectcontext/1506807-reset) is called (see NSManagedObjectContext.swift, method `clearReadHistory`), which invalidates the `NSManagedObjects` fetched from `viewContext`. Thus, references to these must be re-fetched or removed.
* Cached articles are `WMFArticle`s, a subclass of `NSManagedObject`. Even when the cached objects become invalid they are not `nil`, so the code attempts to extract data from them rather than re-fetching (see MWKDataStore.m, method `fetchArticleWithKey`). This causes article titles, descriptions, and images to not be shown on the explore page.
  * This is fixed by clearing the article cache after `viewContext` is reset.
* `ExploreViewController` has child view controllers of type `ExploreCardCollectionViewCell`, which contains a field `contentGroup`. `contentGroup` is of type `WMFContentGroup`, a subclass of `NSManagedObject`. After history is cleared, this seems to cause a problem where some cells nearby the current scroll position on the explore page do not load.
  * To fix this, [`ExploreViewController.collectionView.reloadData()`](https://developer.apple.com/documentation/uikit/uicollectionview/1618078-reloaddata) is called whenever `viewContext` is reset.

### Test Steps
1. Visit at least one article to fill the history
2. Navigate to the history tab, press "Clear", then confirm to delete
3. Navigate back to the Explore page
4. Articles and cells should load properly rather than some being blank


tonisevener (approved):
👏  Works great, I'm very glad to see this weird bug fixed. Thank you for the detailed PR notes!

PR #4446, opened 2023-01-13
Talk Page Archives - Part 2

tonisevener:
**Phabricator:** 

### Notes
This mostly wraps up the Talk Page Archives feature. Most of these commits are feature-specific logic, but a few towards the end dip back into adding abilities to the shifting top views setup (shadow behavior, loading, flicking).

There are a few things for me to look into from [Part 1 PR feedback](https://github.com/wikimedia/wikipedia-ios/pull/4441). I'll look into these in a Part 3 PR.

At the end I added the ability for the navigation bar to reappear when the user flicks back up, to try to build in more functionality from elsewhere in the app. It did turn into some gnarly state code that still doesn't quite feel 100%. I'm also not sure if it even makes sense on this particular screen since the Talk Pages bar doesn't scroll out of view. As of now I feel like the work is worth keeping than scrapping, but I'm open to reverting 1208684 if reviewers disagree.

### Test Steps
1. Load a talk page archives screen.
2. Confirm data loads (Note this isn't picky about Archive pages in particular, just any subpages it seems. This is what Android is doing).
3. Confirm you can tap the row and see an archived talk page.
4. Confirm you can see an empty state on a page with no subpages ("Puppy cat" on EN Wikipedia works).
5. Load archives page while on airplane mode. Confirm you see an error state.
6. Confirm changing themes works.
7. Confirm changing dynamic type works.


mazevedofs (approved):
The shifting view is looking great! I believe you're considering Deepak's feedback for part 3, so I'm not considering it for now

PR #4447, opened 2023-01-14
In article share menu, only copy article URL

junholee6a:
**Phabricator: https://phabricator.wikimedia.org/T304724** 

### Notes
* Currently, there is an issue where, if you go to an article for a location (e.g. [New York City](https://en.wikipedia.org/wiki/New_York_City)), press the Share button, then press Copy, the copied text includes both an Apple Map link and the article link, like so:
"https://en.wikipedia.org/wiki/New_York_City?wprov=sfti1 https://maps.apple.com/?ll=40.712778,-74.006111&q=New%20York%20City"
* If the user has selected text in the article, that may be included in the pasteboard as well.
* Expected behavior: Only the article URL is copied to the pasteboard.
* By default, the `UIActivity.ActivityType.copyToPasteboard` activity concatenates all relevant items (such as strings and URLs) together and puts the combined result into the pasteboard, which may not be what the user expects.
* We add two classes that can wrap around an item (`WMFItemSourceExcludingActivityTypes`) or around a  `UIActivityItemSource` (`WMFItemSourceWrapperExcludingActivityTypes`), and exclude a given list of activity types. When an article's `ShareActivityController` is opened, we exclude the `.copyToPasteboard` activity from all items except for the article URL.
* I chose this implementation because it seemed the least intrusive, but there are alternatives:
  * We could make a custom Copy `UIActivity` to replace the built-in one, but that may require a new localized string for "Copy"
  * We could make a single `UIActivityItemSource` which serves all of the items to `ShareActivityController`. But this would require changing more of the existing code to keep track of pairs of activity types and items, since we'd need to explicitly define which item object is given to which activity type(s). Currently, `UIActivity`s can infer which items are relevant to them based on the types of items.

### Test Steps
**Test 1: Map URL + article URL**
1. Open the [New York City](https://en.wikipedia.org/wiki/New_York_City) article in the app
2. Press the Share button at the bottom
3. Press the Copy button
4. Open Safari
5. Paste into the search bar
6. The pasted text should only have the article URL (https://en.wikipedia.org/wiki/New_York_City?wprov=sfti1)

**Test 2: Selected text + Map URL + article URL**
1. Open the [New York City](https://en.wikipedia.org/wiki/New_York_City) article in the app
2. Long-press on any non-linked word in the article text to select it
3. In the resulting context menu, find and press the "Share..." button (see screenshot below)
4. Press the Copy button
5. Open Safari
6. Paste into the search bar
7. The pasted text should only have the article URL
Simulator Screen Shot - iPhone 14 Pro - 2023-01-13 at 18 49 59


tonisevener (approved):
Looks good, this is working well for me. Thank you for fixing!

PR #4448, opened 2023-01-16
Remove old Talk Pages code

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T319482

### Notes
* This Pr removes the code for the legacy talk pages that is not used anymore
* I'm keeping the `FeatureFlags` struct so we can use it for other features in the future, but I'm happy to remove it if it makes more sense to do so

### Test Steps
1.  Go to the new talk pages
2. Certify it works as expected


tonisevener (changes requested):
Looks good - can you also delete the Core Data entities TalkPage, TalkPageReply, TalkPageTopic and TalkPageTopicContent from the Wikipedia.xcdatamodel file?

I think it'll be easier to test this now before I merge, while it's easy to switch to the old version of Talk Pages. You'll probably need to bump up the data model version number to Wikipedia 7 as well. Once you've removed the entities in this branch:

1. Use the feature flag to switch to the old version of talk pages in `main`.
2. Run the app, visit a few old user talk pages. You can export the user data zip file to confirm you see these items it in the sqlite database file.
3. Then run the app in this branch, confirm it launches without issue. Then export the user data zip file and confirm those items are gone from the database file.

mazevedofs:
@tonisevener I think we can move on with this one, let me know if you'd like to wait until 7.3.0 is released

tonisevener (approved):
Working great - thank you for cleaning this up!

PR #4449, opened 2023-01-17
Article ToC: Fall back to native iOS localization bundle if specific variant is unavailable

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T322160

### Notes
Currently, the article table of contents is constructed using a combination of PCS returned localized item strings (translated in the displayed article's content language) and by appending extra items (About this article, Read more) on device.

On the iOS side, we do not have native localizations for every one of these variants (see `iOS Native Localizations` folder). Additionally, iOS's chosen `Locale` codes for localization bundles do not always match up with with the Wikipedia-level language code. Because of all this, trying to view a ToC for certain language variants sometimes results in the last ToC elements not appearing to be translated.

This PR attempts to add sensible fallbacks to the "root" language to at least try and present these extra ToC items in a language that may be more expected to the user rather than defaulting to an untranslated string. Now, if a user views the ToC for a language variant that is not something we currently have a native iOS localization bundle/string for, we do our best to display something that may be coherent to a native reader.

Here are the languages with variants (per cross-checking `iOS Native Localizations` and `wikipedia-language-variants.json`) that we perform this fall back to the root language with:

- Chinese (zh - our available native localizations are zh-hans, zh-hant)
- Kurdish (ku - our available native localizations are ckb, ku-latn)
- Serbian (sr - our available native localizations are sr-ec, sr-el)
- Tajik (tg - our available native localizations are tg, tg-cyrl)
- Uzbek (uz - our available native localization is uz)

I anticipate this work and behavior may have to change in the future if it doesn't align with the expectations of native readers.

### Test Steps
We should rely on QA for thorough variant testing, but you can see the updated behavior in action with these steps:

1. Open a well-established article like `Dog` on English Wikipedia
2. Using the language selector on the bottom toolbar, select a language variant from the language list above
3. Tap the toolbar button to open the ToC and scroll to the bottom
4. Confirm `About this article` and `Read more` appear to be translated* in line with the other ToC elements

* Note that some localizations may not have translated strings for these final ToC elements, and will fallback to using the iOS device's locale/localization string if available. If these last elements appear untranslated/incorrectly translated, confirm that `article-about-title`, `article-read-more-title` don't appear in the corresponding iOS localization bundle's `Localizable.strings` for that element. There's not much we can do in this case except hope those strings are eventually user translated.


mazevedofs (approved):
Looking good! The Chinese variants don't fallback to english anymore (that's the extent of the testing I can do) and now the app is ready to show them when we get translations for some of the variants

PR #4451, opened 2023-01-21
Update fix-reverse-dollar-signs.sh

aria1991:
Phabricator: N/A

### Notes:
I Used `find -print0` and` xargs -0` instead of `find -print` and `xargs -L 1` allowing to handle filenames with spaces or special characters more safely & I also removed the `-L 1` option from the xargs command, as this is not necessary when using 
`-print0` and` -0 `together.

### Test Steps:

1. Open a terminal or command prompt in the directory where the `'.strings'` files are located.
2. Type the script and press Enter to execute it
3. The script should run and perform the substitutions on all the matching `'.strings'` files
4. You can verify that the script worked by opening the `'.strings' `files and checking if the substitutions have been made.


tonisevener (approved):
Looks good, thank you for improving this! Feel free to open another PR adding your name to the AboutViewController.plist > contributors array so that your name shows up in the app.

PR #4453, opened 2023-01-23
Do not prevent users from viewing talk page if summary call fails

tonisevener:
**Phabricator:** 

### Notes
This fixes a discrepancy we were seeing between Android and iOS on Talk Page Archives. For Elizabeth II > FAQ and Elizabeth II > Archive index, Android showed ann empty state, whereas iOS shows an error state.

This is because the summary fetch fails for these archived pages, and those pages happen to parse as empty. My fix is to not consider a summary fetch as a breaking failure. The header is capable of displaying without it, so we should still allow the user to see an empty state.

### Test Steps
1. Go to Elizabeth II's talk page. Confirm talk page displays fine.
2. Go to archives, tap an archived page. Confirm talk page displays fine.
3. Go to FAQ or Archive index. Confirm talk page displays empty state.

### DIFF
--- a/Wikipedia/Code/TalkPageDataController.swift
+++ b/Wikipedia/Code/TalkPageDataController.swift
@@ -1,5 +1,6 @@
 import Foundation
 import WMF
+import CocoaLumberjackSwift
 
 /// Class that coordinates network fetches for talk pages.
 /// Leans on file persistence for offline mode as-needed.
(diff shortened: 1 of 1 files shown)


mazevedofs (approved):
Working great! Thanks for fixing it so quickly :)

PR #4454, opened 2023-01-24
Add alert when closing the editor if wikitext was modified

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T303218

### Notes
* This PR adds an UIAlertView to the editor close button if the wikitext is altered. To assess that the wikitext was altered, we're checking if the progress button (labeled Next) and the undo button are enabled. The codemirror underlying system disables both buttons if the editor has no changes.
* I chose not to try and tap into codemirror's `change` messaging option to avoid side effects.

### Test Steps
1. Tap the edit button in an article. Make a change, and tap the close button. A system alert should be displayed.
2. Tapping discard should close the editor. Taping cancel only dismisses the alert.
3. Tap the edit button in an article, but make no changes. Tap the close button. The editor should close normally.


tonisevener (approved):
Working great!

PR #4457, opened 2023-02-01
Update copy_code_mirror_resources

aria1991:
**Phabricator:** N/A

### Notes:

This version of the script:

- Uses proper exit codes and error messages to provide more clear feedback in case of an error
- Uses double quotes around variables to ensure they are properly expanded, even if they contain spaces or other special characters
- Adds a success message at the end to indicate that the operation completed successfully.

### Test Steps:

1. Make sure you have the necessary directories and files. The script assumes that the `CodeMirror` and `Wikipedia directories` exist, and that the `wikipedia-ios-codemirror/resources` directory is present within the `CodeMirror` directory.

2. Open a terminal or command prompt and navigate to the directory where the script is located.

3. Make the script executable by running the following command: `chmod +x` `<script_name>.sh` (replace `<script_name>` with the actual name of your script).

4. Run the script by typing `./<script_name>.sh` (again, replace `<script_name>` with the actual name of your script).

5. If the script runs without any errors, you should see a success message indicating that the CodeMirror resources have been copied to the `Wikipedia/assets/codemirror directory`.

6. If there are any errors, the script will output an error message and the exit status will be non-zero, indicating that the operation failed.

7. After the script has run, you can verify that the files have been copied correctly by checking the `Wikipedia/assets/codemirror/resources `directory.


tonisevener (approved):
Looks great, thank you!

PR #4468, opened 2023-02-13
Move Reading Lists funnel to MEP, delete Feed funnel related code

mazevedofs:
**Phabricator:** partially https://phabricator.wikimedia.org/T327341

### Notes
* This PR moves the existing reading lists funnel to MEP, and removes the Feed funnel
* The feed funnel code was removed as it was coupled with some of the reading lists logic

### Test Steps
1. Launch the app, and check if the reading lists events are being sent
2. Do a quick smoke test on the feed cards


mazevedofs:
I'll close this PR and open another one with all the work since it's adding unnecessary complexity at this moment

PR #4469, opened 2023-02-14
Add UIComponents Package

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T329328

### Notes
This PR adds our new UIComponents package (repo [here](https://github.com/wikimedia/wikipedia-ios-ui-components)) to the project. The package is pretty much empty, aside from an empty Demo app project and a `UIComponents` struct (automatically added via New Package flow).

### Test Steps
1. On ExploreViewController.swift, add an `import UIComponents` to the top.
2. At the end of viewDidAppear, add a call to `print(UIComponents().text)`.
3. Run app, confirm you see "Hello World" printed to the console.


staykids (approved):
Great to see this coming together! 

I think an interesting challenge we'll encounter is the friction of trying to create new components while previewing in the app itself. The adding the local package to project/making & pushing changes/deleting local package route covered in the Apple docs you shared previously definitely provides a path forward, but I wonder how cumbersome it will actually be in practice.

PR #4474, opened 2023-02-17
Blocked messages design review feedback

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T201640#8619330

### Notes
This PR addresses design review feedback for blocked panels. I was able to get to most of it except for the gradient view.

For pinned panel buttons, I created a second set of buttons that are pinned to the scroll view frame layout guide. Only the blocked and abuse filter panels enable pinning, so the other panels should remain unchanged.

### Test Steps
1. Block one of your accounts on [Test Wikipedia](https://test.wikipedia.org/wiki/Special:Block) (if you don't have admin privileges, I can block one for you).
2. Log into that account on device. Set up your primary app language as non-Test language.
3. Visit an article on Test Wikipedia. Go to section editor. Confirm PR feedback is addressed:

- Blocked panel shows up more quickly than before (you can compare against the video in the Phab task)
- Confirm panel is smaller in width and image is smaller.
- Confirm Close button and Action buttons are pinned.
- Confirm indicator flashes when blocked message displays (this one is a bit spotty for me).

I was receiving compiler errors from `main`, so this PR is stacked off of #4472.


mazevedofs (approved):
Looking great!

PR #4476, opened 2023-02-22
Color cleanup

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T329559

### Notes
* This PR adds color literals to the Colors class, so we can more easily transition them to the components repo if needed
* System colors such as `.clear,` and `.systemBackground` were not replaced. 
* Edit:  My idea when naming the colors that were moved into the Colors extension was to provide context as to where they are used in the app, so feel free to suggest better names.

### Test Steps
1. Run the app, do a simple smoke test
2. Special attention to the settings screens and attributed strings
3. An extra step would be to search the codebase for UIColor literals  colors instantiated with ` wmf_colorWithHex(_ hex: Int)`


Wikipedia/Code/NSAttributedString+WMFSavedPagesAttributedStrings.m
```diff
@@ -2,14 +2,11 @@
 #import "NSString+FormattedAttributedString.h"
 @import WMF;
 
-static NSInteger const kTitleColor = 0x000000;
```
tonisevener, line 5: Actually I think you might be able to remove this and its header file entirely, it doesn't look like it's used anymore in the app.

tonisevener (changes requested):

tonisevener (approved):

PR #4477, opened 2023-02-22
Address edit notices design feedback

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T201596#8618418

### Notes
Addresses design feedback from above Phab permalink:
- Suppresses toasts from appearing on edit notices modal
- Prevents keyboard from appearing over edit notices modal
- Updates body text size
- Updates margins of the bottom toggle container
- Aligns edit notices modal's body text based on semantic content attribute of content language

Additionally:
- Prevents the edit notices modal from re-appearing when you return to the editor from the preview

(RE: keyboard appearance over the modal – I think there is a _proper_ pathway to cleaning up our keyboard issue (where we wait to perform a bunch of fetch completion logic until the appearance of the view and completion of the task group), but that path would be kind of involved, so I don't recommend we tackle it unless we're still seeing the issue with an approach like the one present here.)

### Test Steps
1. Just light regression testing keeping an eye out to be sure no keyboards/accessory input views/or toasts appear over the edit notices modal
2. Opening an RTL language article with an edit notice (used Arabic featured article from today) to confirm body text of edit modal is using semantic content attribute of content language to set text alignment


mazevedofs (approved):
Looking good! The solution for hiding the keyboard is working great!

PR #4491, opened 2023-03-14
Clear out featured article widget cache if article data changed

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T331902

### Notes
This PR clears out the featured article widget cache from the app when new article data indicates it has changed. This is so that the featured article widget bypasses this [cache check](https://github.com/wikimedia/wikipedia-ios/blob/main/WMF%20Framework/Widget/WidgetController.swift#L216), so that old data isn't stuck on the widget for a day.

### Test Steps
1. With a proxy, locally map the explore feed daily call (https://en.wikipedia.org/api/rest_v1/feed/featured/2023/03/14) with some incorrect local data for the `tfa` object (Changing the `thumbnail.source` and `originalimage.source` values is a good test).
2. Run the app from this branch. Add the featured article home screen widget. Confirm you see the incorrect image in the widget.
3. Now uncheck the map local flag from your proxy so that the real data is fetched.
4. Tap your home screen widget so that you are taken to article view.
5. Background and go back to your home screen widget. Confirm it refreshes with the correct data.
6. Delete the app, repeat steps 1 - 3.
7. At this point, instead of tapping the home screen widget, foreground the app, go to the Explore tab and pull to refresh the Explore feed.
8. Background and go back to your home screen widget. Confirm it refreshes with the correct data.

### Screenshots
out

### DIFF
--- a/WMF Framework/SharedContainerCache.swift
+++ b/WMF Framework/SharedContainerCache.swift
@@ -81,3 +81,12 @@ public final class SharedContainerCache<T: Codable>: SharedContainerCacheHouseke
 @objc public protocol SharedContainerCacheHousekeepingProtocol: AnyObject {
     static func deleteStaleCachedItems(in subdirectoryPathComponent: String)
 }
+
+@objc public class SharedContainerCacheClearFeaturedArticleWrapper: NSObject {
+    @objc public static func clearOutFeaturedArticleWidgetCache() {
+        let sharedCache = SharedContainerCache<WidgetCache>(fileName: "Widget Cache", defaultCache: { WidgetCache(settings: .default, featuredContent: nil) })
+        var updatedCache = sharedCache.loadCache()
+        updatedCache.featuredContent = nil
+        sharedCache.saveCache(updatedCache)
+    }
+}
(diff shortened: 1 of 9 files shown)


WMF Framework/SharedContainerCache.swift
```diff
@@ -81,3 +81,12 @@ public final class SharedContainerCache<T: Codable>: SharedContainerCacheHouseke
 @objc public protocol SharedContainerCacheHousekeepingProtocol: AnyObject {
     static func deleteStaleCachedItems(in subdirectoryPathComponent: String)
 }
+
+@objc public class SharedContainerCacheClearFeaturedArticleWrapper: NSObject {
+    @objc public static func clearOutFeaturedArticleWidgetCache() {
+        let sharedCache = SharedContainerCache<WidgetCache>(fileName: "Widget Cache", defaultCache: { WidgetCache(settings: .default, featuredContent: nil) })
```
staykids, line 87: With our multiple uses of the "Widget Cache" filename, do you think we'd be better served making it a static property on `WidgetCache` (or leveraging `SharedContainerCacheCommonNames`, though I think the intent of the properties there are slightly different)?

tonisevener: Ah good catch, I think that is exactly what `SharedContainerCacheCommonNames` was made for, if I recall correctly. This is done!

PR #4493, opened 2023-03-22
Table of Contents iPad crash fix

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T280700

### Notes
This is a followup fix to https://github.com/wikimedia/wikipedia-ios/pull/4487. I think it did reduce our table of contents presentation crashes, but we have one that I'm still seeing, mainly for iPad on 16.4. I want to take a similar approach and try some more defensive coding targeting some issues that are possible when presenting. I added the specific comments of UIKit errors I'm trying to target for code clarity.

### Test Steps
1. I have no repro steps, but please regression test table of contents display on iPad and iPhone.

### DIFF
--- a/Wikipedia/Code/ArticleTableOfContentsDisplayController.swift
+++ b/Wikipedia/Code/ArticleTableOfContentsDisplayController.swift
@@ -92,8 +92,14 @@ class ArticleTableOfContentsDisplayController: Themeable {
             self.viewController.isVisible = true
             self.selectAndScroll(to: sectionId, animated: false)
             
-            // Attempt to fix TOC presentation crashes.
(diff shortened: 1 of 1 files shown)


staykids (approved):
I love how hyper-precise these guard statements are 😆, but I suppose especially when we're chasing a problem we can't reproduce it's totally reasonable putting more trust in what the crash reports offer. If the duplicated `delegate?.presentedViewController == nil` here (from the start of the function) results in less crashes of that type, maybe that'll shed a little more light on what's up.

I tried to induce some weird states with the network link conditioner, but so far I haven't noticed any TOC appearance/disappearance regressions in my normal iPhone/iPad usage.

PR #4495, opened 2023-03-28
Protected page warning

mazevedofs:
**Phabricator:**  https://phabricator.wikimedia.org/T313772

### Notes
* This PR adds a warning panel on protected pages when the user doesn't have the necessary rights to perform and edit
* The same logic from the block messages was used: the protected information is received upon clicking on the editing button and loading the wikitext. If the user can't edit, the editor will be displayed as read-only, and a modal with the protection information will be shown.
* Edit notices are not displayed as a modal in this case.

### Test Steps
1. Go to a protected or semi-protected page using an account that does not have the correct permission to edit. Confirm that the modal is shown with the protection info. 
2. Note that, at this moment, the icon is not correct. It will be updated soon.


Wikipedia/Code/SectionEditorViewController.swift
```diff
@@ -458,22 +476,28 @@ class SectionEditorViewController: ViewController {
...
         wmf_showBlockedPanel(messageHtml: blockedError.messageHtml, linkBaseURL: blockedError.linkBaseURL, currentTitle: currentTitle, theme: theme)
     }
+
+    private func presentProtectedPageWarning(error: MediaWikiAPIDisplayError) {
+        guard let currentTitle = articleURL.wmf_title else {
+            return
+        }
+
+        wmf_showBlockedPanel(messageHtml: error.messageHtml, linkBaseURL: error.linkBaseURL, currentTitle: currentTitle, theme: theme, image: UIImage(named: "error-icon"))
+
```
mazevedofs, line 486: Icon to be updated

Wikipedia/Code/SectionEditorViewController.swift
```diff
@@ -95,19 +97,30 @@ class SectionEditorViewController: ViewController {
...
-                self.presentErrorMessage(blockedError: blockedError)
-                
-                DispatchQueue.main.asyncAfter(deadline: DispatchTime.now() + 0.2) { // helps prevent flash as wikitext is loaded
-                    self.configureWebView(readOnly: true)
+            if let error = resultError {
+                if error.code.contains("protectedpage") {
+                    self.presentProtectedPageWarning(error: error)
+                    DispatchQueue.main.asyncAfter(deadline: DispatchTime.now() + 0.2) { // helps prevent flash as wikitext is loaded
+
+                        self.configureWebView(readOnly: true)
```
tonisevener, line 111: Looking at the task description, it seems like there's a chance that the user could have `canUserEdit = true` (meaning they have elevated permissions to edit a semi-protected page) and need an enabled editor, but design still expects the modal to present informing them of protection. (see "Editor decides to edit a semi-protected page" section in the Phab task).

So I think it may make more sense to combine the checks into one conditional section:
```
if {we have an error with a code of type protectedpage}  {
    {present warning}
    self.configureWebView(readOnly: !self.canUserEdit)
}
```

Wikipedia/Code/SectionEditorViewController.swift
```diff
@@ -95,19 +97,30 @@ class SectionEditorViewController: ViewController {
...
+                    }
+                } else {
+                    self.presentErrorMessage(blockedError: error)
```
tonisevener, line 114: I haven't tested, but you might need to move this check to before checking protection levels - it looks like from the task description they want blocked messages to take precedence.

Wikipedia/Code/SectionEditorViewController.swift
```diff
@@ -434,15 +447,20 @@ class SectionEditorViewController: ViewController {
...
-                    self.handle(protection: response.protection)
-                    
-                    if let blockedError = response.blockedError {
-                        self.lastBlockedDisplayError = blockedError
-                        completion(blockedError)
-                    } else {
-                        self.lastBlockedDisplayError = nil
-                        completion(nil)
-                    }
+                    self.canUserEdit = self.handle(protection: response.protection, userInfo: response.userInfo?.groups ?? [])
```
tonisevener, line 450: This naming feels a bit generic for something that's very page protection specific, maybe we could adjust the property and method name so that a reader can tell that it has nothing to do with blocked messages? I don't love this alternative but maybe something like:

```
self.userGroupLevelCanEdit = self.pageProtectionLevelsContainUserGroup(protection: response.protection, userInfo: response.userInfo?.groups ?? [])
```

tonisevener (commented):
Glad you were able to get these displaying without too much extra work!

Here are my notes just from code review. I noticed a a couple of things that might cause a difference in behavior from the task description, but this is without running the branch and confirming, so I may be wrong. Just wanted to call it out as an area you might want to check on.

Wikipedia/Code/SectionEditorViewController.swift
tonisevener: Actually, this advice may have been wrong (I should have tested!) 😅 I don't think it's possible to get into a situation where the editor is enabled and a page protection error displays, because it seems like if the user has the right permissions, we don't receive any error text from the API for the modal.

No need to change anything though, this logic is all working fine. I may just leave a comment on the Phab task as an FYI.

tonisevener (approved):
Looks good! There are a couple of improvements that I found that sent me down a routing rabbit hole. They are all add-ons to this code, so I'm going to open up a follow-up PR.

PR #4501, opened 2023-04-11
Fixing build issues faced and silencing 1 warning

mrSandeepKr:
**Phabricator:** 

### Notes
* I am unable to build on my machine with the following changes.

#### _Error_:
**OnThisDayView** - description doesn't have to unwrapped as it already of type String
**NotificationCenterFilterView** - imageView is not required to be unwrapped
**WikimediaProject** -  condition 114 and 116 are exactly the same

#### _Warning_ 
**WikidataFetcher** : Warning is that type should be explicitly called out.

### Test Steps
1. Worked with different interfaces of the application to check all scenarios work.


tonisevener (approved):
Looks good, thank you for this!

PR #4506, opened 2023-04-11
Fix insert link search bar tint color

mrSandeepKr:
**Phabricator:** 
Task - [T326207](https://phabricator.wikimedia.org/T326207)

### Notes
* The placeholder text color for the search bar is not visible when the application theme is dark. This only happens when the system theme is light because the user style is coming in light.
* The bug only mentions the link insertion view but it is true for the media insertion view and for any other views which directly depend on system color.
* To solve this on the core level, we need to override the interface styles.

### Test Steps
Verified that these issues do not reproduce.


mrSandeepKr:
Hey, any suggestions on who should be reviewing this?

tonisevener (approved):
Working great, thank you! If you'd like, feel free to open another PR with your name in [AboutViewController.plist](https://github.com/wikimedia/wikipedia-ios/blob/main/Wikipedia/Code/AboutViewController.plist) so that it shows up on our Settings > About the app > Contributors screen.

PR #4509, opened 2023-04-19
Limit explore feed fetch to 30 days

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T335075

### Notes
See https://phabricator.wikimedia.org/T334759#8794589 for why I think this will fix our app launch times.

### Test Steps
I think we could lean on QA to do longer test steps, but general regression on the Explore feed. Make sure it loads each day and loads more data once scrolled to the bottom.


Wikipedia/Code/ExploreViewController.swift
```diff
@@ -320,7 +320,9 @@ class ExploreViewController: ColumnarCollectionViewController, ExploreCardViewCo
 
     private func setupFetchedResultsController() {
         let fetchRequest: NSFetchRequest<WMFContentGroup> = WMFContentGroup.fetchRequest()
-        fetchRequest.predicate = NSPredicate(format: "isVisible == YES && (placement == NULL || placement == %@)", "feed")
+        let now: Date = (Date() as NSDate).wmf_midnightUTCDateFromLocal
+        let oldestDate = now.addingTimeInterval(-(60 * 60 * 24 * Double(WMFExploreFeedMaximumNumberOfDays)))
```
staykids, line 324: Certainly nothing wrong with this approach and fine as is, since we're just trying to roughly filter the displayed results, but here's an alternative if you're looking to clean up some of magic numbers and have it read a little more directly:

```suggestion
let today = NSDate().wmf_midnightUTCDateFromLocal as Date
let oldestDate = Calendar.current.date(byAdding: .day, value: -WMFExploreFeedMaximumNumberOfDays, to: today) ?? today
```

Wikipedia/Code/ExploreViewController.swift
```diff
@@ -320,7 +320,9 @@ class ExploreViewController: ColumnarCollectionViewController, ExploreCardViewCo
...
+        fetchRequest.predicate = NSPredicate(format: "isVisible == YES && (placement == NULL || placement == %@) && contentMidnightUTCDate >= %@", "feed", oldestDate as NSDate)
```
staykids, line 325: It looks like I'm only seeing top read items in the explore feed. Inspecting the sqlite db, it seems `contentMidnightUTCDate` is only being set on top read items. I could be misinterpreting something, but should we leverage `midnightUTCDate` here instead, which seems to be set for every element?

tonisevener: Yikes, good catch! 😅 I changed to `midnightUTCDate`. I also inspected some of the libraries I've received and it seems to always be populated, so I think we're safe with that.

Wikipedia/Code/ExploreViewController.swift
tonisevener: Agreed, that does feel much better. Updated!

PR #4510, opened 2023-04-20
Color adjustments

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T299793

### Notes
* Color adjustments for some views

### Test Steps
1. Run the app, and make sure the colors follow the design guideline


Wikipedia/Code/TalkPageCell.swift
```diff
@@ -229,6 +229,10 @@ extension TalkPageCell: Themeable {
         rootContainer.backgroundColor = theme.colors.paperBackground
         rootContainer.layer.borderColor = theme.colors.border.cgColor
 
+        if theme == Theme.light {
```
tonisevener, line 232: I'm sure we're not consistent with this, but I think we should avoid changes at the UI level like this. If there is a new combination of colors needed for this border, we should spin up a new semantic color group name and use that instead.

mazevedofs: Good point, I created a new one for this use case

tonisevener (approved):

PR #4473, opened 2023-02-16
Configure iOS schemas to send data to MEP, remove legacy schemas

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T327341 and https://phabricator.wikimedia.org/T331481

### Notes
* This PR sets our app to send analytics data to the Modern Events Platform.
* It removes the base classes that served the legacy system, removes some funnels that were not to be moved into MEP, and updates the existing funnels to work with MEP
* I tried to merge most of the deletion commits, but the ones that were too far apart on the timeline were causing conflicts,  so there are some separate commits with deletions

### Test Steps
1. Run the app. The schemas for the explore feed, funnel, settings, login, readings lists, and user history (that captures a broader picture of the app's current state) were updated, so please interact with all these areas.
2. Check if info is being sent on a network debugger or the events platform.
3. A lot of code was removed, so a broader smoke test would be good while exploring the app

PR #4512, opened 2023-04-24
MEP session tweaks

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T327341

### Notes
This is a follow-on from #4473 that attempts to fix some bugs I was seeing with session handling.

I changed a few things here:
1. To match the communication flow of the other funnels, I tried to have any UI (i.e. Settings or Onboarding toggle) or app lifecycle (foreground, background) event call `SessionsFunnel`, which in turn calls `EventPlatformClient` as-needed. I found this easier to understand, but I'm open changes if others disagree. `UserSession` still exists, but only `EventPlatformClient` communicates with it.
2. After testing, I noticed logging the user's last session needs to be called with a completion block, before resetting their sessionID or turning off analytics. Without this, there were times when the `app_session` event sent with the new session ID instead of the previous one, or didn't send at all because `wmf_sendUsageReports` was set to off before we tried to save the last `app_session` event.
3. Before we generated session ID as-needed, which meant even when toggling on their settings toggle, the session ID wouldn't be generated and the session length wouldn't start counting until they triggered the first analytics event. Usually that happened right away, but not always. In these situations, toggling off analytics caused crashes (because the session start date was never determined, so the `app_session` event didn't know what to do). So now to keep everything more predictable, I am generating the session ID and setting the session start date as soon as they toggle on. This feels more accurate to what we're trying to measure in `app_session` anyway.

### Test Steps
1. Fresh install app. Turn on analytics in onboarding screen.
2. Trigger various events, and watch the console log. Confirm the session ID stays the same for these events.
3. Background the app, then foreground. Trigger more events. Confirm the sessionID stays the same for these events.
4. Terminate the app, then relaunch. Trigger more events. Confirm the session ID stays the same for these events.
5. Background the app, then change the device time to 30+ minutes into the future. Foreground the app and confirm you see an `app_session` event in the console with your session ID. Trigger more events. Confirm these events send with a new session ID.
6. Go to Settings > toggle off analytics. Confirm you see an `app_session` event in the console with your new session ID.

### DIFF
--- a/WMF Framework/Event Platform/EventPlatformClient.swift
+++ b/WMF Framework/Event Platform/EventPlatformClient.swift
@@ -83,9 +83,12 @@ import CocoaLumberjackSwift
         userSession.reset()
     }
     
-    public typealias DidResetSession = Bool
-    public func appDidBecomeActive() -> DidResetSession {
-        return userSession.appDidBecomeActive()
+    public func generateSessionID() {
+        userSession.generateSessionID()
+    }
+    
+    public func needsReset() -> Bool {
+        return userSession.needsReset()
     }
     
     public func appDidBackground() {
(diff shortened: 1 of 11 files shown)


mazevedofs (approved):
Working great!

PR #4514, opened 2023-04-25
Add Edit Attempt Schema Funnel 

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T313561

### Notes
* This PR adds logging for edit actions on the edit attempt schema - can be found [here](https://github.com/wikimedia/schemas-event-secondary/blob/master/jsonschema/analytics/legacy/editattemptstep/current.yaml) 
* Our implementation was based on Android -  [for reference](https://github.com/wikimedia/apps-android-wikipedia/blob/main/app/src/main/java/org/wikipedia/analytics/eventplatform/EditAttemptStepEvent.kt) 
* Also to maintain some parity with android, we're not sending `user_editcount`
* Some required fields, like MediaWiki version, are not available for us, so we're sending empty literals in this case

### Test Steps
1. Go to any article, and click the edit button. 
2. Go all the way to confirming your edit
3. Start another edit attempt, but this time cancel it 
4. Check if the events are being sent


Wikipedia/Code/ArticleViewController+Editing.swift
```diff
@@ -30,6 +31,7 @@ extension ArticleViewController {
         } else {
             present(navigationController, animated: true)
         }
+        EditAttemptFunnel.shared.logInit(articleURL: articleURL)
```
staykids, line 34: Apologies if I'm missing something, but do we need to call this in both `showEditorForSectionOrTitleDescription` and here in `showEditorForSection`? Would it suffice to just have the call in `showEditorForSectionOrTitleDescription`?

Wikipedia/Code/SectionEditorViewController.swift
```diff
@@ -550,15 +551,21 @@ class SectionEditorViewController: ViewController {
...
         present(alert, animated: true)
     }
 
+    fileprivate func logCancelEdit() {
+        EditAttemptFunnel.shared.logAbort(articleURL: articleURL)
+    }
+
     fileprivate func closeEditor(navigateToURL url: URL? = nil) {
         delegate?.sectionEditorDidCancelEditing(self, navigateToURL: url)
+        logCancelEdit()
```
staykids, line 568: Is this event needed here, as `ArticleViewController`'s implementation of `sectionEditorDidCancelEditing` itself sends the same `logAbort` event leading to duplicate calls?

Wikipedia/Code/SectionEditorViewController.swift
```diff
@@ -595,6 +603,7 @@ extension SectionEditorViewController: SectionEditorNavigationItemControllerDele
             showDestructiveDismissAlert()
         } else {
             closeEditor()
+            logCancelEdit()
```
staykids, line 606: Same note as in `closeEditor` about duplicate event logging. It appears closing an unedited editor view results in several duplicate `logAbort` calls.

Wikipedia/Code/EditSaveViewController.swift
```diff
@@ -348,6 +349,9 @@ class EditSaveViewController: WMFScrollViewController, Themeable, UITextFieldDel
         default:
             WMFAlertManager.sharedInstance.showErrorAlert(nsError, sticky: true, dismissPreviousAlerts: true, tapCallBack: nil)
         }
+        if let articleURL {
+            EditAttemptFunnel.shared.logSaveFailure(articleURL: articleURL)
+        }
```
staykids, line 354: Would it make sense to move this block before the `switch` statement to ensure it's called regardless if one of the `switch` cases returns early?

staykids (commented):
Apologies for the review delay. Nice – looking good! I traced some happy path events up through to `EventPlatformClient.shared.submit`. Some notes below on things I was unclear about.

Wikipedia/Code/ArticleViewController+Editing.swift
mazevedofs: You're right, it would duplicate the event, good catch!

Wikipedia/Code/EditSaveViewController.swift
mazevedofs: Yes! Thanks for catching all those :) I guess it's good for another review, let me know if I can adjust something else

staykids (commented):
Looks great! Have tried both happy path and aborting the flow editing on Test wiki and it seems like the correct events are constructed.

One small suggestion: when editing the introduction paragraph, a start event is published, but when presented with the article description/introduction modal, tapping cancel does not publish an abort event. What do you think about replacing this [cancel action](https://github.com/wikimedia/wikipedia-ios/blob/e2e4da97e9f78e2be5e1f66ac8daef394fbcd607/Wikipedia/Code/ArticleViewController%2BEditing.swift#L87) with: 

```swift
sheet.addAction(UIAlertAction(title: CommonStrings.cancelActionTitle, style: .cancel) { _ in
    EditAttemptFunnel.shared.logAbort(articleURL: self.articleURL)
})
```

If we don't need matching start/abort events though just lemme know and I'll merge as-is.

mazevedofs:
> 

I'm not sure the events need to match, but this one was missing; thanks for catching this one as well!

staykids (approved):

PR #4515, opened 2023-04-25
Improve background task handling

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T335077
https://phabricator.wikimedia.org/T335134

### Notes
This PR attempts to reduce our background crashes. It mainly audits and improves our use of UIApplication background task handling. Notes are below on the more important commits.

41b88af - This is probably unrelated, but this fixes a console error when loading an article, which may help with some of our issues:

> API MISUSE: dataTask:didReceiveResponse:completionHandler: completion handler not called

8e4ac41 - This fixes some background crashes. If a user's database contained an article object that needed conversion from our pre-PCS format, but the user didn't contain article legacy html in the filesystem, it would get stuck in an unending loop trying to convert that article. This is because [at this line](https://github.com/wikimedia/wikipedia-ios/commit/8e4ac41208f6a27dac2f23289ea7b560a538e048#diff-0a5f34cb96a3fef7e51d00485445149a9de621d3c21ddaea5bf12fb956598bc7R46), we didn't unset the article object to prevent it from being picked up by the converter again. Because we wrap up this migrating code in a background task, it never completed enough to call `endBackgroundTask`, so it crashed.

f694e9f, 8eb674276, 74797e6 - This calls `endBackgroundTask` in the expiration handler of various background tasks that we're starting. This matches the way our other background tasks were set up, and [this documentation](https://developer.apple.com/documentation/uikit/uiapplication/1623031-beginbackgroundtask) suggests it:

> If you provide a block object in the handler parameter, the system calls your handler before time expires to give you a chance to end the task.

d04d971, eaf1394 - This attempts to shorten the lifespan of some of our background tasks, so they don't stay open during long processes. `MobileViewToMobileHTMLMigrationController` and `SavedArticlesFetcher` loop through articles one by one and converts or downloads/deletes them, respectively. Before these commits the background tasks stayed open as long as there were articles to work against. For migrating or saving a large amount of articles at once, this could cause the background task to last longer than 30 seconds, causing a background crash. This modifies it so that tasks are opened and closed during each article update iteration.

5b71918 - I saw a collection view crash in Explore Feed when artificially faking a long library migration and backgrounding during the process. This basically disables animation updates on that screen when the app is backgrounded to fix the crash (we are setting the same flag for `viewDidAppear` and `viewWillDisappear` in `ExploreViewController`). It may have been something only caused by my testing, but this change felt pretty low risk and may help with any [future library migrations](https://phabricator.wikimedia.org/T335076) that we decide to pick up.

### Test Steps
1. Regression-test downloading a lot of saved articles at once and backgrounding during the download process (maybe log into an account with many saved articles synced). Wait 30+ seconds. Ensure when you return to the app you do not see the Wikipedia splash screen and downloading resumes, indicating that it didn't crash in the background.


mazevedofs (approved):
Working great! Looking at these changes it makes sense that it might affect users with older data on their devices, but it's looking pretty good. I tested with saving a bunch of articles on one device and then fresh installing the app in another one.

PR #4517, opened 2023-04-27

translatewiki:
Translation updates


tonisevener:
Fixed in [Translatewiki](https://translatewiki.net/w/i.php?title=Wikimedia%3AWikipedia-ios-reading-list-entry-limit-exceeded-title%2Fnqo&diff=11562777&oldid=11558339).

PR #4521, opened 2023-05-05
Add Native Source Editor - Part 1

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T331936

### Notes
This PR adds the native editor feature flag and implements the initial pieces of it within `PageEditorViewController`. I also added a full source edit button in `ArticleViewController`, so that we can test performance against a large set of wikitext as we build this. The entry point of this full source edit button is likely to change before release, though.

Note I found a bug in the default automatic theme switching in this branch. Marking as draft while I look into that.

### Test Steps
1. In the Staging scheme, load the article editor, either by tapping the full source "Edit" button at the top of an article or any edit pencil. 
2. Text formatting toolbars and input views should display like our Code Mirror editor, though state logic in the Find and Replace toolbar view will act wrong. I will fix this in a separate find & replace PR later on.


tonisevener:
> Note I found a bug in the default automatic theme switching in this branch. Marking as draft while I look into that.

I can reproduce this bug in `main` as well, so it's unrelated to these changes. Marking as ready to review.

Wikipedia/Code/WMFAppViewController.m
```diff
@@ -1853,6 +1854,7 @@ - (void)updateAppThemeIfNecessary {
     if (self.theme != theme) {
         [self applyTheme:theme];
         [self.settingsViewController loadSections];
+        [self setWKAppEnvironmentThemeWithTheme: theme];
```
staykids, line 1857: Just a note for forthcoming PRs that we'd eventually need to also pass along the trait collection changes so the font scaling respects our design intents.

Wikipedia/Code/PageEditorViewController.swift
```diff
@@ -0,0 +1,236 @@
+import UIKit
+import Components
+
+protocol PageEditorViewControllerDelegate: AnyObject {
+    func pageEditorDidCancelEditing(_ pageEditor: PageEditorViewController, navigateToURL: URL?)
+}
+
+class PageEditorViewController: UIViewController {
```
staykids, line 8: Should we mark this as `final`?

Wikipedia/Code/PageEditorViewController.swift
```diff
@@ -0,0 +1,236 @@
...
+        
+        navigationItemController.apply(theme: theme)
+        view.backgroundColor = theme.colors.paperBackground
+    }
+}
+
+// MARK: - WKSourceEditorViewControllerDelegate
+
+extension PageEditorViewController: WKSourceEditorViewControllerDelegate {
+    func sourceEditorViewControllerDidTapFind(sourceEditorViewController: Components.WKSourceEditorViewController) {
```
staykids, line 160: The `Components.` namespace prefix is unneeded here.

staykids (commented):
Soooo cool seeing this come together! I ignored text theming stuff as I assume it's coming in forthcoming PRs. Just a couple of minor notes in `PageEditorViewController`, but happy to merge as-is.

Wikipedia/Code/WMFAppViewController.m
tonisevener: Thanks - this is updated!

Wikipedia/Code/PageEditorViewController.swift
tonisevener: Agreed, done!

Wikipedia/Code/PageEditorViewController.swift
tonisevener: Done!

Wikipedia/Code/WMFAppViewController.m
JoeWarwick: good stuff

mazevedofs (approved):
looking great!

PR #4525, opened 2023-05-11
Theme not applied to PageHistoryViewController [T330320]

YuriSolodkin:
Theme was not applied when opening PageHistoryViewController from actions menu [T330320]


tonisevener (approved):
Working great - thanks for fixing!

PR #4526, opened 2023-05-12
Add NavigationEventsFunnel to MEP, update ReadingListsFunnel

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T336508 & https://phabricator.wikimedia.org/T328487

### Notes
* We're adding the NavigationEventsFunnel to MEP. It existed in the legacy platform but was left out of the initial iteration of the migration
* We're enhancing this funnel with new fields to observe interaction with the article view toolbar
* We're adding new fields so we better understand interaction on the saved reading lists view

### Test Steps
1. Run the app, interact with the article view toolbar, and navigate around settings and the whole app
2. Interact with the reading lists view
3. Confirm that you see the events being sent with an HTTP request debugger

### DIFF
--- a/Wikipedia/Code/SavedViewController.swift
+++ b/Wikipedia/Code/SavedViewController.swift
@@ -326,7 +326,6 @@ extension SavedViewController: CollectionViewEditControllerNavigationDelegate {
     func didChangeEditingState(from oldEditingState: EditingState, to newEditingState: EditingState, rightBarButton: UIBarButtonItem?, leftBarButton: UIBarButtonItem?) {
         defer {
             navigationBar.updateNavigationItems()
-            ReadingListsFunnel.shared.logTappedEditButton()
         }
         navigationItem.rightBarButtonItem = rightBarButton
         navigationItem.rightBarButtonItem?.tintColor = theme.colors.link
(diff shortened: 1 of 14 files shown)


Wikipedia/Code/SavedViewController.swift
```diff
@@ -314,6 +326,7 @@ extension SavedViewController: CollectionViewEditControllerNavigationDelegate {
     func didChangeEditingState(from oldEditingState: EditingState, to newEditingState: EditingState, rightBarButton: UIBarButtonItem?, leftBarButton: UIBarButtonItem?) {
         defer {
             navigationBar.updateNavigationItems()
+            ReadingListsFunnel.shared.logTappedEditButton()
```
tonisevener, line 329: I'm not sure when all this method gets called, but I notice `edit_tab` is getting logged here even when I only tap the all articles or reading lists tab.

tonisevener (changes requested):
Looking good! I found a few issues with some of the events either not logging at all or too much, see the comments.

My only other comment is that I feel like the `_success` events for the article toolbar actions may not be too useful, but I think they're good where they are at for now. We can adjust later if needed after testing.

Wikipedia/Code/SavedViewController.swift
mazevedofs: Great catch, I left it at the wrong place; now it's only called when the `isEditing` flag is true

Wikipedia/Code/WMFSettingsViewController.m
```diff
@@ -203,6 +203,62 @@ - (void)updateStateForMenuItemType:(WMFSettingsMenuItemType)type isSwitchOnValue
     }
 }
 
+- (void)logNavigationEventsForMenuType:(WMFSettingsMenuItemType)type {
```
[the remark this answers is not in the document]
mazevedofs: Thanks for catching all these! It's looking better now; let me know if there's something else missing :)

tonisevener (approved):
Looks great, thanks! Can you resolve the merge conflicts?

PR #4529, opened 2023-05-19
Clean out mobile-html migration code

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T336804

### Notes
This PR removes code related to local migration of saved articles to the mobile-html endpoint format. Upon app resume, it looked up `WMFArticle` core data objects flagged with `isConversionFromMobileViewNeeded`, then pushed their legacy persisted saved article data through some bundled javascript to transform it into modern persisted saved article data, which our `ArticleViewController` is capable of displaying.

This PR cleans out that process so that this local conversion no longer happens. A regression that this will cause is that users migrating from 6.5.1 and earlier will no longer be able to read their old articles while offline upon initial launch. Once they resume an internet connection, our `SavedArticlesFetcher` will automatically download and persist their saved article data from the network. After that point they can then read them while offline as usual.

While I was in this area of the app, I also deleted some unused code in `WMFAppViewController` and `MWKDataStore`.

### Test Steps

I pushed up an old 6.5.1 version of the app in our Experimental build for testing this flow.

1. On a test device, download and run Wikipedia Experimental, version 6.5.1 (21). Since the old `mobileview` article endpoint is gone, this build fakes out that endpoint with some old locally bundled `mobileview` json, and has saved some articles with these fake responses. This should be enough to test this PR.
2. Disable device internet connection. Go to app Settings > Clear cache, then install the new app version by running the Experimental scheme in Xcode from this PR branch.
3. Launch while offline. Confirm list of saved articles remains, but they will probably not load now that the migration code is removed.
- This is okay, as the next two steps illustrate that this will just be a fleeting state for users.
4. Re-enable internet connection. Confirm you can revisit articles and they load. Wait until "Article queued to be downloaded" labels disappear, indicating that all articles have saved for offline.
5. Disable internet connection again. Clear cache again via Settings. Relaunch while offline. Confirm you can revisit saved articles and they load.

**Regression Test**
Confirm that saving a new article for offline in the main Wikipedia scheme still works.


staykids:
Holding until after 7.3.0 is released.

mazevedofs (approved):
Looking great!

PR #4530, opened 2023-05-19
Add last scrolled anchor to the share link

revolter:
**Phabricator:** 

### Notes
* This enables sharing a specific section instead of only the general article URL.

### Test Steps
1. Open the app
2. Open an article
3. Open the table of contents sidebar
4. Select a header
5. Share the article

The shared link will include the anchor of the selected header.

### DIFF
--- a/Wikipedia/Code/ArticleViewController+TableOfContents.swift
+++ b/Wikipedia/Code/ArticleViewController+TableOfContents.swift
@@ -66,6 +66,8 @@ extension ArticleViewController : ArticleTableOfContentsDisplayControllerDelegat
                 }
             }
         }
+
+        article.viewedFragment = item.anchor
     }
 
     public func tableOfContentsControllerDidCancel(_ controller: TableOfContentsViewController) {
--- a/Wikipedia/Code/ShareActivityController.swift
+++ b/Wikipedia/Code/ShareActivityController.swift
@@ -144,7 +144,7 @@ class ShareActivityController: UIActivityViewController {
             items.append(text)
         }
         
-        if let shareURL = article.url?.wmf_URLForTextSharing {
+        if let shareURL = article.url?.wmf_URL(withOptionalFragment: article.viewedFragment)?.wmf_URLForTextSharing {
             items.append(shareURL)
         }
 
@@ -162,7 +162,7 @@ class ShareActivityController: UIActivityViewController {
         items.append(WMFItemSourceWrapperExcludingActivityTypes(itemSource: textActivitySource, excludedActivityTypes: [.copyToPasteboard]))
 
         // shareURL is the only item that should be included in the UIActivity.ActivityType.copyToPasteboard activity.
-        if let shareURL = article.url?.wmf_URLForTextSharing {
+        if let shareURL = article.url?.wmf_URL(withOptionalFragment: article.viewedFragment)?.wmf_URLForTextSharing {
             items.append(shareURL)
         }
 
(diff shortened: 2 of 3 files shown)


revolter:
Bump

revolter:
@tonisevener, know that I replaced your merge with a rebase, since I prefer it for clarity

Wikipedia/Code/ArticleViewController+TableOfContents.swift
```diff
@@ -66,6 +66,8 @@ extension ArticleViewController : ArticleTableOfContentsDisplayControllerDelegat
                 }
             }
         }
+
+        article.viewedFragment = item.anchor
```
tonisevener, line 70: One thing I notice is that a reader who naturally reaches a section via scrolling only, without tapping the table of contents, does not have their section shared in the URL. The inconsistency might throw users off a bit. Something you could do instead is call `getVisibleSection` in [shareArticle](https://github.com/wikimedia/wikipedia-ios/blob/8e31893464930229bff58763d6aa110651332715/Wikipedia/Code/ArticleViewController%2BSharing.swift#L2), which should cover both the Table of Contents tap and scrolling:

```
    func shareArticle() {
        themesPresenter.dismissReadingThemesPopoverIfActive(from: self)
        
        let group = DispatchGroup()
        
        group.enter()
        getVisibleSection { [weak self] (sectionId, anchor) in
            self?.article.viewedFragment = anchor
            group.leave()
        }
        
        group.enter()
        var selectedText: String?
        webView.wmf_getSelectedText({ webViewSelectedText in
            selectedText = webViewSelectedText
            group.leave()
        })
        
        group.notify(queue: .main) { [weak self] in
            guard let self else {
                return
            }
            
            self.shareArticle(with: selectedText)
            NavigationEventsFunnel.shared.logEvent(action: .articleToolbarShareSuccess)
        }
    }
```

Note though is that `getVisibleSection` is a bit flaky - sometimes it chooses the section before the one that is clearly on screen. You can see this surface randomly with these test steps:

1. Load an article (I'm testing on iPhone portrait), open table of contents and pick a section further down.
2. Article will scroll down to the section
3. Open table of contents again, you will see that the section _before_ the one you just selected is in a highlighted state.

I think this may be a problem within the web-side of the article, perhaps a loading layout race condition or a bug in the `getFirstOnScreenSection` javascript. So it's maybe not the best approach, but I just wanted to float an alternative.

tonisevener (commented):
> @tonisevener, know that I replaced your merge with a rebase, since I prefer it for clarity

@revolter No problem - thank you for your patience on this one. I created a Phabricator [task](https://phabricator.wikimedia.org/T345436) for the wider team to approve this change in behavior. Feel free to watch that for updates. In the meantime, there's one thing during PR review that I wanted to call out. Thanks!

Wikipedia/Code/ArticleViewController+TableOfContents.swift
revolter: Well, I actually did think of this use case (I remember writing it down somewhere, but I might be mistaken 🤔), but I decided to leave it like this because it closely matches the experience on the web.

Meaning that, if you just scroll the webpage in a browser on a computer, and then share the webpage or manually copy the link from the address bar, it will NOT contain the currently visible section.

Thoughts?

tonisevener: That is a great point! I agree, it does feel expected when comparing it to web. Others are good with this behavior change so I'm going to resolve and merge. Thanks again!

tonisevener (approved):

Wikipedia/Code/ArticleViewController+TableOfContents.swift
revolter: Great! And thank you for merging ❤️

PR #4531, opened 2023-05-23
Improve state restoration performance

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T337073

### Notes
This PR simplifies state restoration to only the user's last read article, to reduce the chance of prolonged state restoration times upon launch. It also fixes a couple of bugs that originally caused a situation where the user was stuck in a Reading List Detail modal state upon restoration. Test steps for each commit are below.

6485b92 - This fixed a bug where there was a missing close button upon state restoration to a modal Reading List Detail screen. This area of code isn't actually hit anymore due to later commits in this PR, but I decided to fix it anyway in case we bring full view controller hierarchy restoration back in the future. You must git checkout this SHA if you want to test it. Test steps for this are:

1. Go to an article, tap saved button in bottom toolbar
2. In the toast, tap to save to a reading list
3. In the next toast, tap to view reading list
4. On reading list detail modal screen, background the app, then terminate. Re-launch the app. Confirm Close button shows when the Reading List Detail modal is restored. It does not show in `main`.

92c22db and 0b9740a - This prevents pushing on an article from a Reading List Detail modal. Instead it dismisses the modal upon cell tap, then pushes on the article onto the stack below. Test steps are:

1. Go to an article, tap saved button in bottom toolbar
2. In the toast, tap to save to a reading list
3. In the next toast, tap to view reading list
4. On reading list detail modal screen, tap an article. Confirm modal dismisses and article is pushed onto navigation stack beneath.
5. Regression: Go to Saved tab > Reading Lists > Reading lists detail. Tap an article, confirm article is pushed on as before.
6. Regression: Go to Saved tab > All articles. Tap an article, confirm article is pushed on as before.

b4c0bcf - This is the bigger change with simplifying state restoration. I kept the way we persist navigation state as-is, but now upon state restoration, this only seeks out the top-most article, and pushes that onto the root navigation stack.

1. Create a simple rabbit hole from tapping on the Explore feed featured article card. Then background and terminate via the app switcher. Relaunch the app, confirm your last article is restored. Confirm tapping "Back" takes you back to the home screen (i.e. no other parts of the stack are restored).
2. Create a rabbit hole in a modal from tapping on the Explore feed > Top read card > All top read articles. Background and terminate via the app switcher. Relaunch the app, confirm your last article is restored. Confirm tapping "Back" takes you back to the home screen.
3. Create another rabbit hole from app Settings > Your talk page. If possible, try to reach an article from this flow. Terminate the app and confirm only the article restores like the previous test steps.
4. Create a navigation stack where there is no article view. Confirm it launches on the home screen.

Marking as Draft while we confirm this direction with product / design.


Wikipedia/Code/NavigationStateController.swift
```diff
@@ -21,6 +21,78 @@ final class NavigationStateController: NSObject {
     private typealias Presentation = ViewController.Presentation
     private typealias Info = ViewController.Info
 
+    /// Finds the topmost article from persisted NavigationState and pushes it onto navigationController
+    @objc func restoreLastArticle(for navigationController: UINavigationController, in moc: NSManagedObjectContext, with theme: Theme, completion: @escaping () -> Void) {
+        guard let tabBarController = navigationController.viewControllers.first as? UITabBarController else {
```
staykids, line 26: I think this warning can be alleviated with:
```swift
guard navigationController.viewControllers.first is UITabBarController else { ...
```

(Also very cool to see the Xcode Cloud warnings propagating up to the code review layer here in GitHub!)

tonisevener: ‼️ That is very cool! This is fixed.

Wikipedia/Code/NavigationStateController.swift
```diff
@@ -21,6 +21,78 @@ final class NavigationStateController: NSObject {
...
+        
+        var childArticleViewController: NavigationState.ViewController?
+        for child in viewController.children {
+            childArticleViewController = topmostArticleViewControllerFromViewController(child)
+        }
+        
+        return childArticleViewController ?? articleViewController
+    }
+    
+    /// Attempts to restore user's entire heirarchy, included presented modal navigation stacks. Currently not in use in preference of restoreLastArticle to improve loading times.
```
staykids, line 95: @tonisevener Super minor: typos heirarchy → hierarchy, included → including.

tonisevener: Ah sorry about that, this is fixed (also in other places I found).

staykids (approved):
Awesome - working as expected for me so far. Hoping it makes an impact for our users!

PR #4537, opened 2023-06-01

translatewiki:
Translation updates


tonisevener:
Fixed unit test failure manually and in [Translatewiki](https://translatewiki.net/w/i.php?title=Wikimedia%3AWikipedia-ios-aaald-article-insert-last-updated%2Fba&diff=11684500&oldid=11636088).

PR #4542, opened 2023-06-08
Add support for gur(Frafra), guc (Wayuu), fat (Fanti), anp (Angika) languages

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T333596 and https://phabricator.wikimedia.org/T331550

### Notes
* Adding support for gur(Frafra), guc (Wayuu), fat (Fanti), anp (Angika) languages on the iOS app
* ~~Adding support for the Tachelhit script variant shi-tfng~~

### Test Steps
1. On settings > my languages, search for and add the new languages.
2. On search, try to see if you can search for articles in the newly added languages


Wikipedia/Code/wikipedia-language-variants.json
```diff
@@ -127,6 +127,15 @@
             "languageCode": "ku"
         }
     ],
+    "shi" : [
```
tonisevener, line 130: @mazevedofs This file is not auto-generated, so you are correct to edit it directly.

This is the first time we've changed a language to support variants since 6.8.0, and I found some documentation that indicates it'll be a larger thing than adjusting this file. So I suggest backing out this change from this PR and making it it's own separate effort. Below are some of my notes for first places to start if you decide to tackle it:

### Add the Latin option to this file

Looks like this list will take over for all `shi` language options, so you will lose the original Latin option if you only have one here. Adding it in this array should fix it:

```
{
       "languageVariantCode": "shi-latn",
       "languageName": "Taclḥit",
       "localName": "Tachelhit (Latin)",
       "languageCode": "shi"
}
```

Note: I used [this json](https://www.mediawiki.org/w/api.php?action=query&format=json&meta=siteinfo&formatversion=2&siprop=languages%7Clanguagevariants&siinlanguagecode=en) to get the code value.

### Update `MediaWikiAcceptLanguageMapping.json`
 
Here's some nice [documentation](https://github.com/wikimedia/wikipedia-ios/blob/main/docs/languages.md) about this.

### Add a data migration step converting old persisted non-variant data to variants. Present onboarding modal informing them.

Here's some more documentation I found in this [file](https://github.com/wikimedia/wikipedia-ios/blob/main/WMF%20Framework/MWKDataStore%2BLanguageVariantMigration.swift) about this (see comment at the top).

Wikipedia/Code/wikipedia-language-variants.json
```diff
@@ -127,6 +127,15 @@
...
+        {
+            "languageVariantCode": "shi-tfng",
+            "languageName": "ⵜⴰⵛⵍⵃⵉⵜ ",
+            "localName": "Tachelhit (Tifinagh/ⵜⴰⵛⵍⵃⵉⵜ)",
```
tonisevener, line 134: This is minor, but in the app languages list, `ⵜⴰⵛⵍⵃⵉⵜ` is already displayed as the main title of the cell. I suggest changing `localName` to `Tachelhit (Tifinagh)` to remove the duplicate names.

Screenshot 2023-06-14 at 4 16 50 PM

tonisevener (changes requested):
@mazevedofs Thank you for picking this one up 🙏 I didn't intend for it to sit on our board for as long as it did. I would love to one day automate these PRs so that we don't have to remember to add these manually.

Comments are inline, but happy to discuss further if you have any questions.

Wikipedia/Code/wikipedia-language-variants.json
tonisevener: FYI - this file looked good but I think more will be needed from this comment to get these new variants working, so I'm backing this one out so that we can at least get [T331550](https://phabricator.wikimedia.org/T331550) in. Thanks!

tonisevener (approved):

Wikipedia/Code/wikipedia-language-variants.json
mazevedofs: Oh I assumed that the core data migration would happen `automagically` with the corrected json, thanks for pulling it so we can get the other languages in!

tonisevener: @mazevedofs Ah I'm sorry, you could be right, I assumed it wouldn't happen automatically. Happy to test this out in another PR. Note you'll probably also need to add the correct default variant choices for `shi` in `MediaWikiAcceptLanguageMapping.json` as well.

PR #4543, opened 2023-06-09
Refactored all unit tests that were failing due to language localizati…

FilippoZazzeroni:
**Phabricator:https://phabricator.wikimedia.org/T259859** 

### Notes

Added localizable in the test folder for each supported language and implemented for italian. 
Adjusted all unit tests that were failing due to language issue except for MWKLanguageLinkControllerTests that is built to check the en device language.
In order to have unit tests passing in all the supported languages is needed to implement the other localizable files. 

### Test Steps

In order to test just set the simulator language to italian and run the whole unit tests

### Screenshots/Videos

Screenshot 2023-06-09 at 12 49 17
Screenshot 2023-06-09 at 12 49 24


FilippoZazzeroni:
@tonisevener  when you can you please review it, thanks in advance.

tonisevener:
Hi @FilippoZazzeroni. Thanks for looking into this one, and my apologies for the long delay in review. Instead of creating a new set of Localizable.strings that translators will need to keep up with, I wonder if we could just lean on the localized strings that the app is already using? I think that would reduce the setup burden for all languages.

I made a commit [here](https://github.com/wikimedia/wikipedia-ios/commit/ba2b40dd9d92017a2e890e70f0d60dc3ba72d6d9) to hopefully illustrate what I mean. The NotificationsCenterCellViewModelGenericTests I believe work across all languages with this commit. For these:

1. I moved shared strings between the test class and the app into CommonStrings.swift, for easy reference. Then I used those strings as the expected string value in the tests. Note that for this particular view model, it is expected that some of these strings will be populated directly from the server (or the case of this unit test, the fixture json `notifications-generic.json`). This is why I was able to leave some expected English strings in here like "Thank" - this is how my faked server fixture is set up.
2. For dates, I dug into the cell view model `dateText` logic, to determine what date formatter it will ultimately use. Then I used that formatter to translate the expected date string in the test.

Alternatively, since this is a heavier lift, perhaps a quick alternative to all of this would be to simply update our schemes to run tests under English / United States:

screenshot

Let me know what you think - thank you!

nishith142 (commented):
[]()

tonisevener:
Hi @FilippoZazzeroni! We merged a [PR](https://github.com/wikimedia/wikipedia-ios/pull/5069) earlier today which specifies the language and region in the test plan configuration. My hope is that this will stabilize these tests across system regions / languages. If you run into further issues, feel free to reopen this PR and Phabricator task. Thanks!

PR #4544, opened 2023-06-13
Send revision ID on save success event log

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T337331

### Notes
* We missed sending the revision ID on the `saveSuccess()` event log

### Test Steps
1. On an article, create an edit and save it. Check either on a network traffic debugger on the analytics platform that the events contain a `revision_id` field.

### DIFF
--- a/Wikipedia/Code/ArticleViewController+Editing.swift
+++ b/Wikipedia/Code/ArticleViewController+Editing.swift
@@ -184,7 +184,7 @@ extension ArticleViewController: SectionEditorViewControllerDelegate {
         case .success(let changes):
             dismiss(animated: true)
             waitForNewContentAndRefresh(changes.newRevisionID)
-            EditAttemptFunnel.shared.logSaveSuccess(articleURL: self.articleURL)
+            EditAttemptFunnel.shared.logSaveSuccess(articleURL: self.articleURL, revisionId: Int(changes.newRevisionID))
         }
     }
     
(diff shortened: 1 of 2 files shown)


Wikipedia/Code/ArticleViewController+Editing.swift
```diff
@@ -184,7 +184,7 @@ extension ArticleViewController: SectionEditorViewControllerDelegate {
         case .success(let changes):
             dismiss(animated: true)
             waitForNewContentAndRefresh(changes.newRevisionID)
-            EditAttemptFunnel.shared.logSaveSuccess(articleURL: self.articleURL)
+            EditAttemptFunnel.shared.logSaveSuccess(articleURL: self.articleURL, revisionId: Int(changes.newRevisionID))
```
staykids, line 187: No change needed here, and it looks like we already use this pattern elsewhere, but just wanted to mention: generally for important casting situations where there's possible precision loss, I'll just future proof and use the higher level `NSNumber` and allow it to handle doing the right thing, so something like `NSNumber(value: changes.newRevisionID).intValue` here. But my understanding of this specific case is that we're ok without that, with `Int` on the devices that can run our app itself being a `Int64` and it being unlikely we hit the max value revision id.

staykids (approved):
Nice! Made a Test wiki change and confirmed via inspecting traffic that the `saveSuccess` action payload sent along a revision id and was a successful 201.

PR #4546, opened 2023-06-14
Fix search events funnel bugs

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T335544

### Notes
* Fixing some bugs that were preventing us to correctly log the source of the search event, also adding one missing event for `did you mean` on the places search view

### Test Steps
1. Try search from the various entry points in the app, including the places search
2. Also, open an article via deeplink (clicking from a web link that opens the app) and search from that article
3. Check on a network debugger that the events are being sent with the correct source


tonisevener (approved):
@mazevedofs Working well, thanks!

I did find one issue, but it exists in the old version so these fixes are all worth merging. You may want to make a followup PR with adjustments for the insert link flow from the section editor. It leans on `SearchViewController` and I notice it is still sending the source as `search_tab`.

PR #4547, opened 2023-06-26
Fix automatic theme switching bug

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T336093

### Notes
This was a regression from merging https://github.com/wikimedia/wikipedia-ios/pull/4506. Per findings in https://github.com/wikimedia/wikipedia-ios-components/pull/10#pullrequestreview-1454170194, setting this property while in Default theme can cause `traitCollectionDidChange` to not fire when the system mode changes. Keeping the style to `.unspecified` in default theme fixes it. 

### Test Steps
1. Go over test steps in https://phabricator.wikimedia.org/T336093 - confirm mode now correctly switches when in Default theme.
2. Go over test steps in https://phabricator.wikimedia.org/T326207 - confirm placeholder color fix still works.

### DIFF
--- a/Wikipedia/Code/WMFThemeableNavigationController.m
+++ b/Wikipedia/Code/WMFThemeableNavigationController.m
@@ -43,11 +43,11 @@
     
     NSString *themeName = [[NSUserDefaults standardUserDefaults] themeName];
     if ([WMFTheme isDefaultThemeName:themeName]) {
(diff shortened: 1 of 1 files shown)


staykids (approved):
Nice – change seems to work as expected with the cases you provided. Just made a minor commit fixing some of the indentation.

PR #4549, opened 2023-06-27
Universal links handling for places

jayesh15111988:
**Phabricator:** 

### Notes
* Added support to open the universal link with place name and show the place details on map

### Test Steps
Added universal link from test app and verified that Wikipedia app successfully open the map view with specified place name. Verified locally.

### Screenshots/Videos

https://github.com/wikimedia/wikipedia-ios/assets/6687735/edcc5777-1b0b-421a-aa80-b31666f56ec1


tonisevener (changes requested):
@jayesh15111988 Thanks for fixing this! I found one bug, let me know what you think. The other changes look good to me.

PR #4550, opened 2023-06-27
Add Tachelhit Tifinagh script variant, update variants to use BCP47 formatted codes

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T333596 & https://phabricator.wikimedia.org/T338079

### Notes
* This PR adds the Tachelhit Tifinagh script variant to the app and updates our language variant codes to the BCP47 format as the mobileapps service 

### Test Steps
- To test the new Tachelhit script
1.  Run the app on main, and add the Tachelhit language to your languages
2. Change to this branch. You should see this panel when first opening the app.
3. Add the new variant to your languages and confirm the explore feed correctly displays the Tifinagh script

- To test the new variant codes
1. On settings > My languages, add the languages with variants (Chinese, Tachelhit, Serbian, Crimean Tatar, Gan, Inuktitut, Kazakh, Kurdish, Tajik, Uzbek)
4. Confirm that all variants are correctly displayed on the explore feed
PS: I had some persistence issues on the explore feed when removing/adding languages. Deleting the app and then installing it again resolved it, I'm not sure what was causing it


WMF Framework/MediaWikiAcceptLanguageMapping.json
```diff
@@ -1,88 +1,99 @@
...
         }
     },
     "nb": {
         "default": {
             "default": "no",
         }
+    },
+    "shi": {
+        "default": {
+            "default": "shi",
```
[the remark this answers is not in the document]
tonisevener: Update: I went ahead and pushed this change since I was already in there.

I'm still a little foggy on how this file is used. I looked at the variants [documentation](https://github.com/wikimedia/wikipedia-ios/blob/main/docs/languages.md), and it seems to indicate that this allows onboarding to suggest the correct initial wiki language based on device settings. My onboarding screen always suggests EN, even if my device language and region is Chinese/mainland China, for example. I see this behavior in `main` too, so I don't think this is anything you broke - just something I wanted to point out as odd. Maybe I'm testing it wrong.

Wikipedia/Code/wikipedia-language-variants.json
```diff
@@ -1,155 +1,170 @@
 {
```
tonisevener, line 1: This file looks good!

One unfortunate must-fix that I didn't foresee are all the old persisted language variant codes. If you export the user library and dig around in the `sqlite` database, you'll notice some entities (WMFArticle, WMFContentGroup, etc.) that still have the old variant codes in their records (in the `ZVARIANT` column). I think this manifests in weird bugs like:

1. In `main`, add some ZH or SR variant app languages.
2. Switch to this branch, go to app Settings > My Languages.
3. Notice how ZH and SR languages you chose magically disappeared. If you only had variant languages selected, this screen is in a weird broken empty state.

I think the easiest way to migrate this over is to dig into how these four migration methods work in the language variant migration file (see [here](https://github.com/wikimedia/wikipedia-ios/blob/add-shi-tfng/WMF%20Framework/MWKDataStore%2BLanguageVariantMigration.swift#L59)). Of course these methods were originally written to add language variant codes to persisted items that previously had none. But maybe they could be reused to change these old persisted variant codes as well?

tonisevener (changes requested):
Looking good! Just one must-fix I found with old persisted variant codes.

Wikipedia/Code/wikipedia-language-variants.json
mazevedofs: I hadn't noticed that; I mainly checked the explore feed and thought it was fine 🤣. I'm manually adding the variants to the migration - as it was previously done for _nb_ to _no_. It's not super elegant, but not every language got a new code that was just a small change in capitalization so we would need an explicit way to do that either way.

mazevedofs:
@tonisevener I also changed the base branch to 7.3.1; let me know if we should ship this with the next release instead

Wikipedia/Code/wikipedia-language-variants.json
tonisevener: @mazevedofs Looks like this mapping is working for `migratePreferredLanguages` and `migrateExploreFeedSettings`, but doesn't seem to work for `migrateSearchLanguageSetting` and `migrateWikipediaEntities`.

Test steps for `migrateWikipediaEntities`:

1. From `main`, add Traditional Chinese (Hong Kong) as an app language
2. In Explore feed Chinese cards, save some articles.
3. Switch to this branch, export user library
4. In `sqlite`, look up ZWMFContentGroup, ZWMFArticle, and ZReadingListEntry tables. Notice ZVariant column still has records with `zh-hk` - they should be migrated to `zh-Hant-HK`. 

Fix:

In the [migrateWikipediaEntities](https://github.com/wikimedia/wikipedia-ios/blob/add-shi-tfng/WMF%20Framework/MWKDataStore%2BLanguageVariantMigration.swift#L99) method, for example, it's assuming the key in your mapping is a `languageCode` looking it up in that way and updating the variant. I think you'll need to write a new method that takes a different mapping - one that takes an old variant as the key, and updates to a new variant. Something like:

```
    private func migrateWikipediaEntitiesFromOldVariantCodesToNewVariantCodes(_ mapping: [String:String], in moc: NSManagedObjectContext) {
    for (oldLanguageVariantCode, newLanguageVariantCode) in mapping {
            do {
                // Update ContentGroups
                let contentGroupFetchRequest: NSFetchRequest<WMFContentGroup> = WMFContentGroup.fetchRequest()
                contentGroupFetchRequest.predicate = NSPredicate(format: "variant == %@", oldLanguageVariantCode)
                let groups = try moc.fetch(contentGroupFetchRequest)
                for group in groups {
                    group.variant = newLanguageVariantCode
                }

                {rewrite for WMFArticle and ReadingListEntry here as well}
```

Test steps for `migrateSearchLanguageSetting`:

1. From `main`, add Traditional Chinese (Hong Kong) as a secondary app language and EN as primary
2. In app settings search, toggle on "Show languages on search"
3. Search from Explore feed and select ZH tab for searching
4. Switch to this branch and run
6. Try searching from Explore feed again, notice both EN and ZH are grayed out - search language should have persisted and showed ZH as selected.

Fix:

[migrateSearchLanguageSetting](https://github.com/wikimedia/wikipedia-ios/blob/add-shi-tfng/WMF%20Framework/MWKDataStore%2BLanguageVariantMigration.swift#L89) is looking up an old user defaults key, one we had when we did the original language variant release. Instead you'll want to write a new method with this logic tweaked to look up and change the search content language code:

```
    private func migrateSearchLanguageSettingFromOldVariantCodeToNewVariantCode(_ mapping: [String:String]) {
        let defaults = UserDefaults.standard
        let currentSearchLanguageCode = UserDefaults.standard.wmf_currentSearchContentLanguageCode {
            let newSearchLanguageCode = mapping[currentSearchLanguageCode] ?? currentSearchLanguageCode
            defaults.wmf_setCurrentSearchContentLanguageCode(newSearchLanguageCode)
        }
    }
```

Let me know if you have any questions, thanks!

tonisevener:
> I also changed the base branch to 7.3.1; let me know if we should ship this with the next release instead

@mazevedofs yeah, I think so, please go ahead and switch the base branch back to `main` for the next release. Sorry for the back & forth!

tonisevener (approved):
Working well for me!

PR #4553, opened 2023-06-30
Load main Core Data store asynchronously

tonisevener:
**Phabricator:**
https://phabricator.wikimedia.org/T340886

### Notes
7.3.1 showed a significant slowdown in `didFinishLaunching` when I tested by upgrading from 7.3.0 with saved articles. I suspect it may be due to bumping up the Core Data model version number when cleaning out old talk pages (my [suggestion](https://github.com/wikimedia/wikipedia-ios/pull/4448#pullrequestreview-1283735464)).

Our old way of setting up the Core Data stack loaded persistent stores on the main thread and was triggered (indirectly) from `didFinishLaunching`.

For this PR I reworked the Core Data stack setup to use a `NSPersistentContainer` and `NSPersistentStoreDescription`. `NSPersistentStoreDescription` has a `shouldAddStoreAsynchronously` which allows us to set up the stack asynchronously,  which allows this automatic migration to occur asynchronously as well.

### Test Steps
1. Run the App Store version 7.3.0. Save some articles.
2. Run from this branch. Confirm app launches and saved articles still work.
3. Run again from this branch. Confirm app launches and saved articles still work.

### Screenshots

Before (note green `didFinishLaunching` time):
before

After:
after

### DIFF
--- a/WMF Framework/Widget/WidgetController.swift
+++ b/WMF Framework/Widget/WidgetController.swift
@@ -128,8 +128,6 @@ public final class WidgetController: NSObject {
                 self.completions.forEach { $0(dataStore) }
                 self.completions.removeAll()
             }
-        } needsMigrateBlock: {
-            DDLogDebug("Needed a migration from the widgets")
         }
     }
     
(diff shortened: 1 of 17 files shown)


mazevedofs (approved):
Working great! The launch time has improved significantly, and the saved articles and reading lists load quickly with no issues.

PR #4560, opened 2023-07-10
Diffs - Connect Undo and Rollback buttons to API

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T335579

### Notes
This PR connects the undo and rollback buttons with the appropriate API calls. For the success banners, I just used our standard `WMFAlertManager` view, with the copy I observed from Android. I suggest we do tweaks to it at a later design review stage if needed. Since undo and rollbacks are edits, I am also parsing and presenting any MediaWiki error messages and presenting them in a panel, like we did with our recent custom blocked and page protection work.

### Test Steps
1. From any account (or anonymously), go to a recent diff revision on Test Wikipedia and tap through the undo flow. Confirm revision is undone and screen is refreshed to newest revision (and user-entered summary is also displayed in the TO: header).
2. From a non-admin Test Wikipedia account, go to an article's diff screen. Confirm Rollback is not an option. Then enter editor and make a few edits back-to-back on an article.
4. From a Test Wikipedia _admin_ account, go to one of those edits, confirm you can see rollback, and tap through the rollback flow. Confirm all user revisions from step 2 are reverted and screen is refreshed to newest revision.
5. Try doing both flows while in airplane mode. Confirm you see a "No internet connection" banner.
6. (Optional, we can lean on QA for this) Try the undo flow from a blocked account. Confirm you see an error panel and links are tappable.


Wikipedia/Code/DiffContainerViewController.swift
```diff
@@ -613,6 +623,35 @@ private extension DiffContainerViewController {
...
+                guard let self else {
+                    return
+                }
+                
+                switch result {
+                case .success(let status):
+                    
+                    let needsWatchButton = !status.watched && FeatureFlags.watchlistEnabled
+                    let needsUnwatchButton = status.watched && FeatureFlags.watchlistEnabled
+                    let needsArticleEditHistoryButton = !(revisionRetrievingDelegate is PageHistoryViewController) && FeatureFlags.watchlistEnabled
```
tonisevener, line 644: I took liberties in hiding these buttons here if the watchlist feature is disabled. Also, I don't think it makes sense to have an article edit history button if they just came from the article edit history view (they could simply tap Back). My thinking is that this button should only appear if the user came from watchlist. Happy to change it if I'm wrong here though.

staykids:
String merge conflict in this one.

tonisevener:
> String merge conflict in this one.

Should be fixed now!

Wikipedia/Code/DiffContainerViewController.swift
tonisevener: Update: After looking at https://github.com/wikimedia/wikipedia-ios/pull/4568, I changed this to just show the article edit history button at all times here - my previous logic here caused it to hide when coming from talk > revision history > diff.

Wikipedia/Code/DiffContainerViewController.swift
```diff
@@ -1148,43 +1187,118 @@ extension DiffContainerViewController: DiffToolbarViewDelegate {
...
+        
+        fakeProgressController.start()
+        WKWatchlistService().rollback(title: title, project: wkProject, username: username) { [weak self] result in
+            DispatchQueue.main.async {
+                self?.completeRollbackOrUndo(result: result, isRollback: true)
+            }
+        }
+    }
+    
+    private func completeRollbackOrUndo(result: Result<WKUndoOrRollbackResult, Error>, isRollback: Bool) {
```
staykids, line 1275: Let's **not** make any changes here (because I believe I understand the tradeoffs made in this case and it's working well as-is 😀), but just wanted to make a suggestion for thinking about future API design decisions.

When reviewing, this declaration stood out to me as slightly confusing:
`private func completeRollbackOrUndo(result: Result<WKUndoOrRollbackResult, Error>, isRollback: Bool)`

and as something that could be simplified and made more sustainable as something like:
`private func completeRevert(result: Result<WKRevertResult, Error>)`

My recommendation would be that the `enum` type would be more suited than `bool` as an underlying type indicator in cases like this. `Bool` is certainly useful in some cases when we're sure there are and will only be two discrete states like on/off or enabled/disabled, but even though there are _currently_ only two options in this case, `isRollback = false` doesn't immediately and precisely tell me that it's an undo the same way `isUndo = false` wouldn't precisely tell me that the action is a rollback. I personally dunno if "revert" is exactly the right grouping word for these actions or not, but I think having some parent type to represent these actions is more sustainable than using the `Bool` type here.

A fuller `WKData`/client app sketch:

```swift
struct WKRevertResult {
	let action: WKRevertAction
	let newRevisionID: Int
	let oldRevisionID: Int
}

enum WKRevertAction {
	case rollback
	case undo
}

public func rollback(title: String, project: WKProject, username: String, completion: @escaping (Result<WKRevertResult, Error>) -> Void) {
	...
	completion(.success(WKRevertResult(action: .rollback, newRevisionID: newRevisionID, oldRevisionID: oldRevisionID)))
}

public func undo(title: String, revisionID: UInt, summary: String, username: String, project: WKProject, completion: @escaping (Result<WKUndoOrRollbackResult, Error>) -> Void) {
	...
	completion(.success(WKRevertResult(action: .undo, newRevisionID: newRevisionID, oldRevisionID: oldRevisionID)))
}


private func completeRevert(result: Result<WKRevertResult, Error>) {
	...
}
```

Totally open to different perspectives here, but to me benefits of the modified approach:

1. if our "revert" action types grow beyond just undo and rollback, our method signature doesn't need to change
2. all the interesting details (the action itself and the revisions we're interested in) stay contained in the result and don't leak out

staykids (approved):
So rad seeing this all come together – great stuff! Did happy path testing with admin and non-admin accounts up to step 4. Let's rely on QA to exhaustively test this.

Only a minor suggestion with `undoAlertUndoAction`/`isEmpty`, but happy to merge as-is if you disagree.

Wikipedia/Code/DiffContainerViewController.swift
```diff
@@ -1148,43 +1187,118 @@ extension DiffContainerViewController: DiffToolbarViewDelegate {
...
+        undoAlertUndoAction = undo
 
         alertController.addAction(cancel)
         alertController.addAction(undo)
 
         present(alertController, animated: true)
     }
+    
+    @objc private func undoSummaryTextfieldDidChange() {
+        undoAlertUndoAction?.isEnabled = (undoAlertSummaryTextField?.text?.count ?? 0) > 0
```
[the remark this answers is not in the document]
tonisevener: Sounds good, I committed it!

Wikipedia/Code/DiffContainerViewController.swift
tonisevener: @staykids thanks for the write-up! I definitely agree, your approach does feel more clean. Leaving as-is per your request, but I'll keep an eye out for this habit of mine in the future.

PR #4562, opened 2023-07-10
Diffs - Add localized strings and icons for diff user button menu actions

staykids:

### Notes
Adds localized strings and icons to the user button menu actions in the updated diff view header.

### Test Steps
1. Open article history and tap a revision.
2. Tap the from and to user buttons in the header.
3. Confirm icons appear for each menu action.
4. Confirm tapping the actions in the menu route the user to the appropriate views in app.


tonisevener (approved):
Not a blocker at this point but something to look at in a future PR - I notice that the ordering of the menu items change between To and From.

staykids:
> Not a blocker at this point but something to look at in a future PR - I notice that the ordering of the menu items change between To and From.

@tonisevener Funny you mention this – I actually spent some time looking into this yesterday because I was really puzzled by it myself, and it seems to actually be the expected behavior of these contextual `UIMenu`s. I believe Apple's design intention is for the ordering of the menu items to be such that the first/most important element is always closest to your finger's touchdown point given the space available to present the menu. Looking at the screenshot, you can see that `User page` (the first in the array of `UIAction`s)[^1] is always closest to the touchdown point on the button and `User contributions` is furthest away. Note that in iOS 16, there is a way to change the `preferredMenuElementOrder` to be `fixed` and present in the same order regardless of touchdown point, but I think we should roll with the default behavior for the time being.

[^1]: Caveat: Eagle eyed code reviewers will note that `User contributions` is actually the first action added to the menu, not `User page`. When I create the `WKMenuButton`, I didn't have the same understanding of these contextual menus that I have now, so I actually [reversed the menu items](https://github.com/wikimedia/wikipedia-ios-components/blob/30953d67abc1d315d0c080ffb2a4a53010d77536/Sources/Components/Components/Shared/Buttons/WKMenuButton.swift#L118) to get what I felt was the expected presentation 🥴. This is on my list to correct when I touch the `Components` framework next.

tonisevener:
@staykids aah, interesting, that makes sense!  Agreed on leaving it with the default behavior.

PR #4563, opened 2023-07-11
Add asynchronous setup call to fix the on this day widget

tonisevener:
**Phabricator:** 
N/A

### Notes
There was a regression after https://github.com/wikimedia/wikipedia-ios/pull/4553 where my On This Day widget stopped loading. It showed the "Your app language does not support On This Day" error message despite having it set to EN.

I missed a `finishSetup` call at this point to ensure the Core Data stack was loaded asynchronously for the On This Day widget. I think this change is fairly low risk, being that On This Day is the last home screen widget that uses this method of data loading.

### Test Steps
1. Change device time to 1 week ago, add On This Day widget to home screen.
2. Move device time 1 day forward, confirm widget refreshes.
3. Repeat step 2 until current day.


staykids (approved):
Still looking forward to the _On this Day_ when the soup of Core Data in widgets is gone. Works per your test steps! Just had to manually nuke my local Core Data persistent store beforehand because it was presenting cached results for today until doing so.

PR #4564, opened 2023-07-12
Diffs - Add share functionality back into diff view

staykids:

### Notes
- Adds share functionality back into the diff view, via the more > share button
- Also adds edit history icon and delegate method (still need to push actual `PageHistoryViewController` when tapped)

### Test Steps
1. View edit history for an article, and tap a diff
2. Tap the more > share button in the toolbar
3. Confirm shared link opens expected revision(s) when viewed on web


tonisevener (approved):
Looks good! I did one minor fix - I noticed when there are intermediate revisions (i.e. going through the Compare button flow from history and tapping two revisions far apart) the Share url was messed up. I think that's a path I never needed to consider before since the old compare design didn't offer a share button. Should be fixed now.

PR #4566, opened 2023-07-13
[T300321] Fix freeze after pull to refresh

Serguitar:
**Phabricator:** 
https://phabricator.wikimedia.org/T300321

### Notes
Here is the fix for the case when the user releases the drag after pulling to refresh with zero velocity. In this scenario, the content doesn't decelerate and remains dimmed, causing the app to become frozen. In this pull request, this behaviour is handled in the **scrollViewWillEndDragging** function.

### Test Steps
1. Open any article
2. Pull to refresh and scroll down without releasing
3. Release after top content will be under the status bar

### Screenshots/Videos
before the fix
https://drive.google.com/file/d/1MMVbLLhDkBcQpUrN3Crn7l0MMXaCZOxE/view?usp=drive_link

after the fix
https://drive.google.com/file/d/1UCmSGluwlEbNFbGFItBXOis7LoCIAy7o/view?usp=drive_link


tonisevener (approved):
Working great - thank you for the fix!

PR #4567, opened 2023-07-13
Diffs - Edit attempt schema updates

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T337331

### Notes
* This PR logs edit actions on article descriptions and talk pages to the EditAttemptSchema

### Test Steps
1. Edit an article description, add a talk page topic, reply to a talk page topic, undo or rollback diffs, and submit these actions.
2. Try it  again, but now close the edit views before submitting
3. Verify that the logs are correctly sent with no errors


Wikipedia/Code/DescriptionEditViewController.swift
```diff
@@ -84,6 +84,9 @@ protocol DescriptionEditViewControllerDelegate: AnyObject {
         descriptionTextView.textContainerInset = .zero
         
         updateFonts()
+        if let articleURL = articleDescriptionController.article.url {
+            EditAttemptFunnel.shared.logInit(articleURL: articleURL)
```
tonisevener, line 88: I notice `EditAttemptFunnel.shared.logInit(articleURL: articleURL)` is called in ArticleViewController+Editing in `showEditorForSectionOrTitleDescription`, with this you might see double `init` logs when going into the article description editor.

tonisevener (commented):
Here are some thoughts from a code-only check. Use your best judgement on them though as I may be a bit out of the loop on how these should be logged. 

I do feel a bit confused on the difference between the `init`, `intent`, and `attempt` actions, and when each should be logged. I notice `intent` in the old section editor flow seems to occur when the user taps the progress button on the editor screen. Since these other contexts don't really have a multi-view flow, should we maybe not log `intent` at all?

mazevedofs:
> 

@tonisevener  thanks for catching those! I kept all the logAbort calls and removed the log intent call from the diff view controller. I had to add compliance to the new `WKMenuButtonDelegate`since I merged the components PR, so I think we should hold off merging this one until https://github.com/wikimedia/wikipedia-ios/pull/4581 is merged.

Wikipedia/Code/TalkPageViewController.swift
```diff
@@ -426,10 +415,13 @@ class TalkPageViewController: ViewController {
         if let lastViewDidAppearDate = lastViewDidAppearDate {
             TalkPagesFunnel.shared.logTappedNewTopic(routingSource: viewModel.source, project: viewModel.project, talkPageType: viewModel.pageType, lastViewDidAppearDate: lastViewDidAppearDate)
         }
-        let topicComposeViewModel = TalkPageTopicComposeViewModel(semanticContentAttribute: viewModel.semanticContentAttribute, siteURL: viewModel.siteURL)
+        let topicComposeViewModel = TalkPageTopicComposeViewModel(semanticContentAttribute: viewModel.semanticContentAttribute, siteURL: viewModel.siteURL, pageLink: viewModel.getTalkPageURL(encoded: false))
         let topicComposeVC = TalkPageTopicComposeViewController(viewModel: topicComposeViewModel, authenticationManager: viewModel.authenticationManager, theme: theme)
         topicComposeVC.delegate = self
         inputAccessoryViewType = .format
+        if let url = viewModel.getTalkPageURL(encoded: false) {
+            EditAttemptFunnel.shared.logInit(articleURL: url)
```
tonisevener, line 423: Looks good for the new topic button - we're missing `logInit` when a user taps "Reply". Can you add it to the `tappedReply` method in this file?

Wikipedia/Code/TalkPageViewController.swift
```diff
@@ -1023,6 +1026,9 @@ extension TalkPageViewController: TalkPageTopicComposeViewControllerDelegate {
                         self?.updateEmptyStateVisibility()
                         self?.talkPageView.collectionView.reloadData()
                         self?.scrollToLastTopic()
+                        if let viewModel = self?.viewModel, let pageURL = viewModel.getTalkPageURL(encoded: false) {
+                            EditAttemptFunnel.shared.logSaveSuccess(articleURL: pageURL, revisionId: viewModel.latestRevisionID)
```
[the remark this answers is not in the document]
mazevedofs: 🤦‍♀️  updated!

Wikipedia/Code/DiffContainerViewController.swift
```diff
@@ -1264,10 +1280,13 @@ extension DiffContainerViewController: DiffToolbarViewDelegate {
               let username = toModel?.user else {
             return
         }
-        
+
         fakeProgressController.start()
-        WKWatchlistService().rollback(title: title, project: wkProject, username: username) { [weak self] result in
-            DispatchQueue.main.async {
+        DispatchQueue.main.async {
```
tonisevener, line 1285: I'm getting a threading error after you moved this dispatch queue call. Can you move it to only directly surround the `self?.completeRollbackOrUndo` call (line 1290)?

The rollback call to the service sends it to a different thread, so even if you have it above on line 1285 it won't remain on the main thread in the service completion handler.

Wikipedia/Code/DiffContainerViewController.swift
```diff
@@ -1227,34 +1236,41 @@ extension DiffContainerViewController: DiffToolbarViewDelegate {
...
+              let summary = undoAlertSummaryTextField?.text else {
             return
         }
-        
         fakeProgressController.start()
-        WKWatchlistService().undo(title: title, revisionID: UInt(revisionID), summary: summary, username: username, project: wkProject) { [weak self] result in
-
-            DispatchQueue.main.async {
-                self?.completeRollbackOrUndo(result: result, isRollback: false)
+        DispatchQueue.main.async {
```
tonisevener, line 1245: Threading error upon this undo as well - please move this `DispatchQueue.main.async` to only surround `self?.completeRollbackOrUndo` (line 1251).

tonisevener (changes requested):
Here are some additional things I found while testing. Thanks!

Wikipedia/Code/DiffContainerViewController.swift
mazevedofs: Updated all those!

tonisevener (approved):
Looks good! I found a couple of more minor things while testing Wikidata description editing and undo/rollback and pushed a fix. I'm good to merge whenever you're ready.

mazevedofs:
> Looks good! I found a couple of more minor things while testing Wikidata description editing and undo/rollback and pushed a fix. I'm good to merge whenever you're ready.

Thank you so much! The context-switching these couple of days caused me to overlook some stuff, thank goodness for code review! I updated it with main and ran tests locally, I'm good with it being merged now (maybe some more local unit testing to double-check) @tonisevener :)

PR #4570, opened 2023-07-14
Diffs - Update width of user menu buttons

staykids:

### Notes
Update width of user menu buttons in diff view to match designs.

### Test Steps
- Confirm user menu buttons in diff header don't fill entire screen's width but only fill width needed to display the user name within them as designed in Figma doc


tonisevener (approved):
Working great!

PR #4573, opened 2023-07-17
Watchlist - Add watch / unwatch flow to article and diff toolbars

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T335816

### Notes
This PR connects the watch/unwatch/update expiry actions to the article toolbar and diff toolbar.

This PR is dependent on https://github.com/wikimedia/wikipedia-ios-components/pull/22 & https://github.com/wikimedia/wikipedia-ios-data/pull/7.

### Test Steps
1. Go to an article while logged out of the app on the Staging scheme. Confirm tapping "Watch" more menu option prompts you to log in. Confirm you see a choose expiry action sheet after logging in.
2. Chose an expiry. Confirm you see a success toast indicating the new expiry time, with a button to view your watchlist.
3. Tap view watchlist. Confirm the watchlist view is pushed on.
4. Go into airplane mode and repeat some previous actions. Confirm you see a "No Internet connection" banner.
5. Go to a diff screen and repeat previous steps. Confirm watch / unwatch works as expected.

**Not-so-edge case where app is in login state but server rejects cookies**
6. Login on app, go to article. Login to same account on desktop. Then log out on desktop. Go back to app article and tap watch/unwatch. Confirm you see login prompt, and are able to complete the watch/unwatch flow after logging in.


WMF Framework/FeatureFlags.swift
```diff
@@ -7,11 +7,11 @@ public struct FeatureFlags {
...
     public static var watchlistEnabled: Bool {
-        #if WMF_STAGING
+        //#if WMF_STAGING
         return true
-        #else
-        return false
-        #endif
+        //#else
+        //return false
+        //#endif
```
staykids, line 14: I think this `FeatureFlag` commenting out got in here by accident.

staykids (changes requested):
Soooo cool seeing this come together – still looking at it but wanted to point out one small thing I noticed immediately.

Wikipedia/Code/ArticleToolbarController.swift
```diff
@@ -71,6 +76,56 @@ class ArticleToolbarController: Themeable {
         item.apply(theme: theme)
         return item
     }()
+    
+    lazy var moreButton: IconBarButtonItem = {
+        return createMoreButton()
+    }()
+    
+    private func createMoreButton(needsWatchButton: Bool = false, needsUnwatchHalfButton: Bool = false, needsUnwatchFullButton: Bool = false) -> IconBarButtonItem {
```
staykids, line 84: Not suggesting we change this here, but wanted to mention something for consideration in the future.

```swift 
private func createMoreButton(needsWatchButton: Bool = false, needsUnwatchHalfButton: Bool = false, needsUnwatchFullButton: Bool)
```

I think this mutually exclusive boolean pattern is one of those internal API choices that potentially seeps unexpected behavioral expectations to callers down the line. I know in the actual logic there's some mutual exclusivity to do the right thing, but that doesn't prevent outside callers from using it in an unexpected way and getting unexpected (to them) results. For example, a caller can `createMoreButton(needsWatchButton: true, needsUnwatchHalfButton: true, needsUnwatchFullButton: true)`.

Though it's a little more lift (especially for minor buttons or whatever), I generally think wrapping these kinds of things in an `enum` is better API design: dramatically cleans up the function parameters and any call sites and makes it painfully obvious the only valid button configuration possible. Something like:

```swift

func createMoreButton(watchButtonConfiguration: WatchButtonConfiguration) { ... }

enum WatchButtonConfiguration {
	case watch
	case unwatchHalfFilled
	case unwatchFilled
	
	var image: UIImage {
		switch self {
		case .watch:
			return UIImage(systemName: "star") 
		case .unwatchHalfFilled:
			return UIImage(systemName: "star.leadinghalf.filled") 
		case .unwatchFilled:
			return UIImage(systemName: "star.fill") 		
		}
	}
}

moreButton = createMoreButton(watchButtonConfiguration: .unwatchHalfFilled)
```

staykids (commented):
👍🏽 - just a minor typo in `revisionHistory` string.

WMF Framework/CommonStrings.swift
```diff
@@ -267,7 +267,9 @@ public class CommonStrings: NSObject {
     }
     @objc public static let talkPageReplyAccessibilityText = WMFLocalizedString("talk-page-reply-button-accessibility-label", value: "Reply to %@", comment: "Accessibility text for reply button. The %@ will be replaced with the name of the user whose comment is being responded")
 
-    public static let revisionHistory = WMFLocalizedString("talk-page-revision-history", value: "Revision history", comment: "Title for option that leads to talk pages revision history. Please prioritize for de, ar and zh wikis.")
+    public static let revisionHistory = WMFLocalizedString("talk-page-revision-history", value: "Revision history", comment: "Title for for menu option that leads to page revision history.")
```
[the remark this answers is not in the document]
tonisevener: Done!

Wikipedia/Code/ArticleToolbarController.swift
tonisevener: Agh I agree, sorry about that! I went ahead and made this change, but didn't push it in case you weren't keen on looking at it again at this time. I could put up a followup PR if you'd like.

staykids: @tonisevener Oh cool – no worries. Yeah if you don't mind follow-up PR'ing it, I'd like to go ahead and merge this.

PR #4576, opened 2023-07-19
Add article image to diff header view if available

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T335579#9007269

### Notes
* This PR adds a new DiffHeaderView that wraps the existing DiffHeaderTitleView, to add an image view if necessary, with little edits to the existing .xib file
* If the article has an image, it will be displayed on the header of the diff view

### Test Steps
1. Go to a diff on an article with a lead image
2. Confirm that it shows on the diff view according to the designs
3. Go to an article that has no lead image, then make sure its diff view is shown correctly


Wikipedia/Code/DiffHeaderView.swift
```diff
@@ -0,0 +1,89 @@
...
+
+final class DiffHeaderView: SetupView {
+
+    var viewModel: DiffHeaderViewModel?
+
+    lazy var horizontalStackView: UIStackView = {
+        let stackView = UIStackView()
+        stackView.translatesAutoresizingMaskIntoConstraints = false
+        stackView.axis = .horizontal
+        stackView.alignment = .center
```
staykids, line 12: Very cool seeing this come together! One thing I noticed is that as the header/subtitle text increases, the article image displays centered in the stack rather than pinned to the top (like how it functions on talk pages). You might want to test with a few different length strings to see how it looks in different conditions, but resolving this and pinning the image to the top may be as straightforward as updating this alignment to `.top`.

mazevedofs: Good catch, updated!

Wikipedia/Code/DiffHeaderView.swift
```diff
@@ -0,0 +1,89 @@
...
+            })
+        } else {
+            imageView.isHidden = true
+        }
+    }
+
+    func updateTitleView(with viewModel: DiffHeaderTitleViewModel) {
+        headerTitleView.update(viewModel)
+    }
+
```
staykids, line 82: To support the underbar hit fallthrough to allow a user to scroll the view when interacting with the header itself, we need to add something like this (which we also use in other views): 

```suggestion
override func point(inside point: CGPoint, with event: UIEvent?) -> Bool {
    guard !UIAccessibility.isVoiceOverRunning else {
        return super.point(inside: point, with: event)
    }

    return false
}
```

staykids (changes requested):
Looks great! Just one change needed below to support scrolling when interacting with the header view itself.

I hope you don't mind, but I committed a change to `DiffHeaderTitleView.xib` directly in order to resolve a layout issue on iPad.

Wikipedia/Code/DiffHeaderView.swift
mazevedofs: Ohh nice, I didn't know about that 😄 it looks like it's working as expected now!

staykids (approved):
Nice – thanks for making those changes!

PR #4579, opened 2023-07-21
Diffs - Diff view accessibility update

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T342359

* This PR is built from https://github.com/wikimedia/wikipedia-ios/pull/4576, so make sure that PR was merged before merging this

### Notes
* This PR adds some VoiceOver adjustments to the Diff view
* We're grouping some labels and adding some context strings to make context clearer
* For now, we're skipping the actual diff content, as the wikitext diff was confusing and sometimes long

### Test Steps
1. Go to a diff screen with VoiceOver on
2. Make sure you can have it read back the contents of the header and the extended header (with user info), and then moves onto the toolbar.
3. Make sure you can access the user button menu


Wikipedia/Code/DiffHeaderSummaryView.swift
```diff
@@ -19,21 +19,23 @@ class DiffHeaderSummaryView: UIView, Themeable {
...
+
+                summaryLabel.attributedText = attributedText
+            } else {
+                summaryLabel.text = viewModel.summary
             }
-            
-            summaryLabel.attributedText = attributedText
-        } else {
-            summaryLabel.text = viewModel.summary
+            summaryLabel.accessibilityLabel = " SUMMARY \(summary)" // TODO LOCALIZE IT
```
tonisevener, line 38: Reminder about this TODO, although you may not need the additional prefix with the "Edit summary" label above it.

Wikipedia/Code/DiffListChangeCell.swift
```diff
@@ -35,6 +35,9 @@ class DiffListChangeCell: UICollectionViewCell {
     weak var delegate: DiffListChangeCellDelegate?
     
     func update(_ viewModel: DiffListChangeViewModel) {
+
+        textStackView.isAccessibilityElement = false
```
tonisevener, line 39: I also notice there's also a `DiffListUneditedCell` that could maybe use some accessibility suppression. I can't remember how that cell is used though.

tonisevener (commented):
A couple of things from code review, plus these additions:

Can you broadcast an announcement whenever a thank, undo, and rollback succeeds or fails? Just having it speak whatever the WMFAlertManager displays should suffice. I notice when I go through those flows it goes through but the user doesn't have any indication that it worked.

Wikipedia/Code/DiffHeaderSummaryView.swift
mazevedofs: done!

Wikipedia/Code/DiffListChangeCell.swift
mazevedofs: I'm also not sure where it is used, but I suppressed it anyway

mazevedofs:
> A couple of things from code review, plus these additions:
> 
> 1. Can you broadcast an announcement whenever a thank, undo, and rollback succeeds or fails? Just having it speak whatever the WMFAlertManager displays should suffice. I notice when I go through those flows it goes through but the user doesn't have any indication that it worked.
> 2. (Minor) I notice the More button in the toolbar is navigated to first - it would feel better to me if it is the last toolbar item read. Maybe some ordering adjustments need to be made.

hi @tonisevener !
Added VoiceOver support for the banners on thanks and undo/rollback. 
On the toolbar, it always works in the left to right order of the buttons to me (from prev diff -> more). Is there any other way of testing so I can check? It's hard not to be biased when testing it since I already know where all the elements are on the view

tonisevener (approved):
Looks good! I bumped up the announcement delay to 2.0 seconds - let me know if that's okay. I found that it kept getting interrupted by Voice Over reading the back button on my devices, since we refresh the screen at that point.

Also, it looks like I can only reproduce the ordering issue on my iOS 15 device, so I don't think that issue is a blocker.

mazevedofs:
> 

That's great thanks! If ordering the toolbar elements becomes an issue on more versions, I'm happy to make a follow-up PR to fix it. I ran tests locally, but I had to reconcile some conflicts in case you'd like to take another look at the DiffContainerViewController, before merging.

PR #4581, opened 2023-07-24
Diffs - Added logging for Diff changes

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T341896

### Notes
This PR adds logging in various places for our recent diff changes. Although many of these events are under the new watchlist schema, I tried to stick to logging only the diff-related actions through it. We can add all of the watchlist actions once we are further along with that feature. 

This PR is dependent on https://github.com/wikimedia/wikipedia-ios-components/pull/16. Please point the Components package here back to `main` after that one is merged.

### Test Steps
No need to be super comprehensive with this so that we can lean more on QA and Data, but do some general testing of the new diff actions on top of code review. Confirm by checking the console json.


mazevedofs (approved):
Working great!

PR #4598, opened 2023-08-17

translatewiki:
Translation updates


tonisevener:
Our Import translations GitHub action failed to complete last time. It failed with:

```
fatal: unable to access 'https://github.com/wikimedia/wikipedia-ios/': Failed to connect to github.com port 443 after 75016 ms: Couldn't connect to server
```

For now I manually ran it by running the Update Localizations scheme and committing the changes. I think another set of translations should come through tomorrow, so we can check there to see if there's some sort of actions access issue or if it was just a temporary hiccup.

PR #4602, opened 2023-08-25
Drop iOS 14, tweaks for Xcode 15

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T343038
https://phabricator.wikimedia.org/T344005

### Notes
These are changes needed to drop support for iOS 14. This is best reviewed commit-by-commit.

A lot of warnings appeared after bumping the deployment target. I think I fixed 27a2bdf properly. The other two commits I described as suppressions, as they felt like more of a workaround rather than actually fixing. Most of the warnings were in a0a4837 - in my mind properly fixing these buttons with `UIButtonConfiguration` would have been a significant amount of rework that we already want to do with Components rewrites, so I resorted to some warning suppressions. Happy to discuss that further if there's disagreement on this though.

The project mostly already worked in Xcode 15, so there are some minor commits towards the end for that. I think Xcode 14.3.1 should still run fine with these (which our PR unit tests are still set against). I also kicked off an Experimental build against Xcode 15 beta to confirm deployment will work.

### Test Steps
1. Just general regression. Run on simulators, test devices, in both iOS 15, 16, and 17. Ensure no warnings appear.


Wikipedia.xcodeproj/xcshareddata/xcschemes/Wikipedia.xcscheme
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Scheme
-   LastUpgradeVersion = "1430"
-   version = "1.7">
+   LastUpgradeVersion = "1500"
```
staykids, line 3: Curious: any reason for minimum targeting 15.0 instead of a later iOS 15 version? I suppose a 15.0 minimum provides a wider install base, though maybe only marginally as anyone on iOS 15.0 should be on devices that support updating to later versions of 15. I think a lower iOS 15 minimum vs. a later one increases our support burden (though also maybe marginally), as we allow the potential for unaddressed API fixes/changes (https://developer.apple.com/documentation/ios-ipados-release-notes/ios-ipados-15_1-release-notes etc).

Wikipedia/Code/IconBarButtonItem.swift
```diff
@@ -12,8 +14,9 @@ class IconBarButtonItem: UIBarButtonItem {
         let customView = UIButton(type: .system)
         customView.setImage(image, for: .normal)
         customView.addTarget(target, action: action, for: controlEvents)
-        customView.adjustsImageWhenDisabled = false
-        if iconInsets != .zero { customView.imageEdgeInsets = iconInsets }
+        var deprecatedCustomView = customView as DeprecatedButton
```
staykids, line 17: Looks like the `customView` below needs this `as DeprecatedButton` sorcery as well. (Sorry GitHub wasn't letting me comment on that line)

Wikipedia.xcodeproj/xcshareddata/xcschemes/Wikipedia.xcscheme
tonisevener: These `LastUpgradeVersion` changes are the Xcode version checks - this basically prevents that "update with project recommended settings" suggestion on Xcode 15. But I did set the deployment target to [15.0](https://github.com/wikimedia/wikipedia-ios/pull/4602/files#diff-ef4712bf7dff5e1a57a3fc2a39c0397c9de3c33796d6fbfa20b9ef9d7de302ca), so your comment still stands :)

I've generally always thought that if it isn't (greatly) limiting us with what APIs we can use, we should go with the even number to increase our user base for free. But I'm open to us revisiting this! I'll follow up.

Wikipedia/Code/IconBarButtonItem.swift
tonisevener: Fixed! 🧙

Wikipedia.xcodeproj/xcshareddata/xcschemes/Wikipedia.xcscheme
tonisevener: @staykids Updated to 15.4.

staykids (changes requested):
Nice! Some iOS 15 availability/unavailability checks remaining to address: 

- Line 39: `NotificationsCenterIconType`
- Line 31: `VanishAccountContentView`
- Line 129: `VanishAccountContentView`
- Line 43: `VanishAccountPopUpAlertView`
- Line 70: `NotificationService`

tonisevener:
> Nice! Some iOS 15 availability/unavailability checks remaining to address:
> 
> * Line 39: `NotificationsCenterIconType`
> * Line 31: `VanishAccountContentView`
> * Line 129: `VanishAccountContentView`
> * Line 43: `VanishAccountPopUpAlertView`
> * Line 70: `NotificationService`

Thanks for catching those - they are updated.

I'm seeing a crash in Xcode 15 / iOS 17 when I tap a link in an article view, related to the navigation bar. Testing this branch in Xcode 14.3.1 still seems to work, so I don't think merging will introduce anything new. I can look into it as a part of this PR or in a followup PR, just wanted to let you know in case you saw it @staykids.

staykids:
Thanks for calling that out, noticed it too. Xcode 14.3.1 deployed to iOS 16/17 all good but Xcode 15 beta 8 deployed to iOS 17 crashes there. And this was on `main` so yeah seems unrelated to new changes here. I'm happy to finish review on this or you're welcome to look at it here if you were hoping to keep Xcode 15 related stuff in one PR, totally up to you.

tonisevener:
@staykids I think I have fixed the crashes, details are in the commit message.

I also noticed a bug on iOS 17 where if you scroll down on an article until the navigation bar disappears > tap a link > tap back, the navigation bar would appear blank. The navigation bar mysteriously didn't have any subviews during this point in the layout cycle, so it wouldn't apply the transform to display the subviews again. A quick fix I found is to bypass any inner subview transformation here for these shrinking navigation bars on iOS 17. It is a subtle design change though, so let me know if you disagree with this (or we can confirm during design review).

I think this also likely fixes https://phabricator.wikimedia.org/T344005.

WMF Framework/DeprecatedButton.swift
```diff
@@ -0,0 +1,64 @@
+import Foundation
+
+// https://forums.swift.org/t/suppressing-deprecated-warnings/53970/6
+
+public protocol DeprecatedButton {
```
staykids, line 5: I really appreciate your approach here to manage our warnings!

Wikipedia/Code/NotificationsCenterDetailViewController.swift
```diff
@@ -31,6 +31,18 @@ final class NotificationsCenterDetailViewController: ViewController {
...
     }
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        setupNavigationBar()
+    }
+    
+    private func setupNavigationBar() {
+        
+        let detailTitle = WMFLocalizedString("notifications-center-detail-title", value: "Notification Detail", comment: "Title of notification detail view, displayed after tapping a notification in Notifications Center.")
```
staykids, line 42: Just wanted to note, the long press previous view menu item title on pushed views changes to Notification Detail, but the back button title itself remains Back. Not suggesting any change, as this is an improvement from existing functionality which just displays "Back" as both the long press option and the back button title.
detail

staykids (approved):
👍🏽 nice – psyched you found solutions for the Xcode 15/iOS 17 issues! I think this would make a nice candidate for more exhaustive QA, because of `NavigationBar`'s pervasiveness.

PR #4608, opened 2023-08-31
Remove akan (ak)

srish:
- Delete `ak.json`
- Remove ak from `wikipedia-languages.json`

Related task: https://phabricator.wikimedia.org/T333765


tonisevener:
Thanks for the PR @srish! I pushed a commit removing another AK config file. I also added a one-time language settings migration from  AK > TW. I noticed that without this, users upgrading with only AK as their app language had some buggy states in app Settings.

srish:
> https://phabricator.wikimedia.org/T333765

@ tonisevener Thank you for making the necessary changes.

srish:
@tonisevener Also from @dbrant recently in a chat "Both iOS and Android apps have scripts that run periodically and update our list of supported wikis. If a certain wiki has been sunset, it will be picked up automatically and removed". For future sake, I am wondering are there any recommendations for making PRs - is this even needed, should I do exactly what I did this time in the PR or something else?

tonisevener:
> @tonisevener Also from @dbrant recently in a chat "Both iOS and Android apps have scripts that run periodically and update our list of supported wikis. If a certain wiki has been sunset, it will be picked up automatically and removed". For future sake, I am wondering are there any recommendations for making PRs - is this even needed, should I do exactly what I did this time in the PR or something else?

For iOS, we have a command line script that we manually run occasionally to populate languages from the sitematrix config. You can run it in Xcode by selecting the "Update Localizations" scheme. I just now ran it myself to see what happens - it does update `wikipedia-languages.json` to remove AK (which removes it from the Wiki language options in the app Settings). It looks like it doesn't remove the namespace file or the code mirror file. This isn't a big deal but this script could be improved to do that.

screenshot

The only thing that concerns me is our persisted data -  for users that only had AK set up as an app language, once it disappeared from `wikipedia-languages.json` the app acted strangely (blank language options in Settings). That's why I added that migration step to TW. I'll create a bug task for us to look into it - maybe we can provide a better fallback mechanism for users with old persisted language codes that no longer exist in `wikipedia-languages.json`. Until that's fixed, I think doing what you did this time is the way to do it for iOS, to give us an opportunity to double-check things like that. Thanks!

dbrant:
> If a certain wiki has been sunset, it will be picked up automatically and removed

Just for maximum transparency, according to our (Android's) git logs, the `ak` language was removed in April, via our periodic scripts. Users who had the ak language selected in their app would automatically fall back to the default language of their device.

tonisevener (approved):

PR #4615, opened 2023-09-07
Change Components and WKData to local packages

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T345645 (First item)

### Notes
This PR re-establishes our monorepo setup, to remove juggling between 3 repositories. I think I have managed to remove any concerns with local package revision IDs as well, because there are no more references to them in the package manifest files.

Most of this setup came from following instructions in https://developer.apple.com/documentation/xcode/organizing-your-code-with-local-packages. Then I copied over the Source and Test files into the local packages.

Note: So far PR tests only run against the main app unit tests, but I think I can modify the workflow to include Components and WKData tests in there as well. I don't want to change that workflow until this is merged.

### Test Steps
1. Run app.
2. Make a breaking change in WKData and Components. Confirm compile errors bubble up to app call sites.
3. Confirm you can fix errors in app call sites, and see changes reflected across all 3 areas in git.


staykids:
@tonisevener Apologies for the delay on this one as we headed towards the Watchlist TestFlight. Would it be much trouble to update this to represent the current state of `wikipedia-ios`, `wikipedia-ios-components`, and `wikipedia-ios-data`? From there, I can prioritize reviewing it over everything else to get it in as soon as possible, and I can also take responsibility for rebasing/reworking any Watchlist related work that's approved while I'm reviewing this to cleanly merge into this monorepo.

tonisevener:
@staykids This is updated, thanks!

staykids:
@tonisevener Thank you! Will prep that updated Watchlist build then dive in.

staykids (approved):
Our long nightmare is over – thank you for taking on this work!

Just some notes on other next steps that I'll add as a subtask to your original ticket:

- We should probably bring your `DeprecatedButton` approach into the Components framework as well – the Warnings we have to deal with until then are worth it.

- Nice that we can leverage our existing SwiftLint setup here. I was hoping housing the Packages in the main project would also let us use Xcode's Text Settings prefs for per-file indentation settings (visible when selecting a file and viewing the File Inspector), but it seems like Xcode is still (rightly) treating those packages independently.

- I'm a little disappointed losing our lil' Demo project (mainly because it built so blazingly fast compared to the main app 😭), but we can always bring it alongside the main Xcode Project and Xcode Workspace 'em together in the future.

PR #4616, opened 2023-09-08
Add empty view to watchlist feature

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T342552

### Notes
* This PR adds the empty view for the watchlist feature
* On the filter empty view, only some of the subtitle is a button. That was because we can't use attributed strings in SwittUI yet, so I had to break down the sentence and create a small button in the middle.

### Test Steps
1. On an account with no watchlist items, make sure you see the empty view, and it redirects to the app's search
2. Add some watchlist items, and make sure you see them. Then activate all the filters to see the filter option for the empty view, and that it redirects to the filter view.


Wikipedia/Code/ViewControllerRouter.swift
```diff
@@ -194,13 +192,25 @@ class ViewControllerRouter: NSObject {
                 )
             }
 
+            let localizedNumberOfFilters: (Int) -> String = { filters in
+                String.localizedStringWithFormat(
+                    WMFLocalizedString("watchlist-number-filters", value:"{{PLURAL:%1$d|%1$d filter|%1$d filters}}", comment: "Amount of filters active in watchlist - %1$@ is replaced with the number of filters."),
```
tonisevener, line 197: It actually gives translators much more control if you bury the whole string in here:

```
                String.localizedStringWithFormat(
                    WMFLocalizedString("watchlist-number-filters", value:"Modify {{PLURAL:%1$d|%1$d filter|%1$d filters}} to see more Watchlist items", comment: "Amount of filters active in watchlist - %1$@ is replaced with the number of filters."),
                    filters
                )
            }
```

Assuming the order of the translations and splicing it together could cause translation bugs. Plus we don't have to do as much on our side.

I've made progress on this in this Components [commit](https://github.com/wikimedia/wikipedia-ios-components/commit/6262e97bb5d9387ba470b0ff435dfbfe0655c9ce) and app-side [commit](https://github.com/wikimedia/wikipedia-ios/commit/a1ca7f65ef67e04051b59f162a875ab0d492876f). Unfortunately I've lost the "n filters" coloring though. I know we managed to properly theme AttributedStrings in on the Account Vanishing footer view - maybe you can reference that to finish this up? Let me know, thanks.

tonisevener (approved):

PR #4619, opened 2023-09-09
Apple Pay - Donate form API calls

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T345847

### Notes
This PR adds the ability to do a few API calls for the donate form:

1. Fetching the [donate config file](https://donate.wikimedia.org/wiki/MediaWiki:AppsDonationConfig.json?action=raw).
2. Calling FR Tech's [payment networks](https://phabricator.wikimedia.org/T344860) endpoint. (Still in development)
3. Calling FRTech's [submit payment](https://phabricator.wikimedia.org/T344860) endpoint. (Still in development)

Since these are pretty basic calls*, I thought this was a good chance to spin up a new service in `WKData` and avoid the loop back to the app for networking. This new service is called `WKBasicService`.

_*FR Tech endpoints may eventually need an API key, but I don't have those details yet._

I am also fetching the config and payment methods endpoints at the same time as our current announcements endpoint fetch. I may modify it slightly in the future to trigger once our future [campaign config call](https://phabricator.wikimedia.org/T344024) goes through, to reduce unnecessary calls.

Note: https://github.com/wikimedia/wikipedia-ios/pull/4619/commits/b3aea1bdc197f820c6adf8e7bf5f24ece9467053 will come in handy while the FR Tech endpoints are still in development. So no need to revert before merging, but the goal is to eventually revert this commit.

### Test Steps

1. In the app and trigger an announcements fetch (generally just a background/foreground, then pull to refresh is needed on Explore). Confirm config values are printed out.
5. Confirm unit tests pass.


WKData/Sources/WKData/Extensions/Int+HttpStatusCodes.swift
```diff
@@ -0,0 +1,7 @@
+import Foundation
+
+extension Int {
+    var isHttpSuccess: Bool {
```
staykids, line 4: I like the intention, and appreciate it only internal to `WKData`, but I wonder if this really needs to extend _all_ `Int`s over just being in some generalized HTTP status code utility.

WKData/Sources/WKData/Extensions/URL+API.swift
```diff
@@ -0,0 +1,68 @@
+import Foundation
+
+extension URL {
+    static func mediaWikiAPIURL(project: WKProject) -> URL? {
+        var components = URLComponents()
+        components.scheme = "https"
+        components.path = "/w/api.php"
```
staykids, line 7: Would it make sense to generalize out these repeated constructions into a `WMFAPIPath`-ish from `NSURL+WMFLinkParsing` thing (`BaseMediaWikiAPIPathComponents`?)?

WKData/Sources/WKData/Models/Donate/WKDonateConfig.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+
+public struct WKDonateConfig: Codable {
+    let version: Int
+    public let currencyMinimums: [String: Decimal]
+    public let currencyMaximums: [String: Decimal]
+    public let currencyAmounts7: [String: [Decimal]]
```
staykids, line 7: I understand these currently match the API, but just wanted to mention the same note as in Slack about potentially more precise naming here if those suggestions still agree with you.

WKData/Sources/WKData/Models/Donate/WKPaymentSubmissionResponse.swift
```diff
@@ -0,0 +1,13 @@
...
+    public class Response: Codable {
+        let status: String
+        let gatewayTransactionID: String?
+        let orderID: String?
+    }
+    
+    let response: Response
+}
+
+/// {"response":{"status":"Success","gateway_transaction_id":"PVLSL6JZ4NK2WN82","order_id":"1235.1"}}
```
staykids, line 13: Extraneous

WKData/Sources/WKData/Service/WKBasicService.swift
```diff
@@ -0,0 +1,224 @@
+import Foundation
+
+/// Use this service for the most basic networking service calls. It does not handle authentication.
+///
```
staykids, line 4: Extraneous

WMF Framework/FeatureFlags.swift
```diff
@@ -10,4 +10,11 @@ public struct FeatureFlags {
         return true
     }
     
+    public static var applePayEnabled: Bool {
```
staykids, line 13: Minor, but what do you think about something like `applePayDonationEnabled` instead? On first read `applePayEnabled` connoted the availability of Apple Pay _itself_ to me, rather than our intent of Apple Pay as a donation option.

staykids (commented):
There's a big chunk of work to review here. Here's just very superficial stuff from my first pass. Will dig deeper in next pass.

WKData/Sources/WKData/Extensions/Int+HttpStatusCodes.swift
tonisevener: Good idea - I moved this into a `HTTPURLResponse` extension.

WKData/Sources/WKData/Extensions/URL+API.swift
tonisevener: Done (I think) - let me know if it's what you had in mind!

WKData/Sources/WKData/Models/Donate/WKDonateConfig.swift
tonisevener: Done! I updated the mock json and the config at donate.wikimedia.org...so far I haven't changed the config at test.wikipedia.org. I will EOD today once I know FR Tech isn't testing against it.

WMF Framework/FeatureFlags.swift
tonisevener: Done (twice actually, sorry). I went with the more general `donorExperienceImrovementsEnabled` - I'm hoping to be able to wrap up all of our improvements in the next few weeks around that for easy rollback, not just the native form.

WKData/Sources/WKData/Extensions/URL+API.swift
staykids: To clarify, I think even pulling the repeated scheme would be useful. Like:

```swift
private static var baseMediaWikiAPIPathComponents: URLComponents {
	var components = URLComponents()
	components.scheme = "https"
	components.path = "/w/api.php"
	return components
}

static func mediaWikiAPIURL(project: WKProject) -> URL? {
    var components = baseMediaWikiAPIPathComponents

    switch project {
    ...
}
```

WMF Framework/FeatureFlags.swift
staykids: 👍🏽 new name works great, much more precise about what it is

WKData/Sources/WKData/Data Controllers/WKDonateDataController.swift
```diff
@@ -0,0 +1,132 @@
...
+        }
+        
+        // TODO: Send in API key
+            
+        let request = WKBasicServiceRequest(url: donatePaymentSubmissionURL, method: .POST, parameters: parameters, bodyContentType: .json)
+        service?.performDecodablePOST(request: request, completion: { (result: Result<WKPaymentSubmissionResponse, Error>) in
+            switch result {
+            case .success(let response):
+                guard response.response.status == "Success" else {
+                    return
```
staykids, line 123: Missing a `completion` handler call in this failure case.

WKData/Sources/WKData/Extensions/Dictionary+API.swift
```diff
@@ -0,0 +1,21 @@
...
+    
+    func encodedForAPI() -> [String: Any] {
+        var returnDict: [String: Any] = [:]
+        
+        for (key, value) in self {
+            
+            guard
+                let encodedName = key.addingPercentEncoding(withAllowedCharacters: CharacterSet.urlQueryComponentAllowed),
+                let encodedValue = (value as? String)?.addingPercentEncoding(withAllowedCharacters: CharacterSet.urlQueryComponentAllowed) else {
+                    continue
```
staykids, line 13: Unsure if we want to continue our pattern of `assert(...)` in the new frameworks, but this may be a good candidate. Any key/value parameter pair failing to be properly URL encoded means the entire parameter dictionary is invalid, right?

WKData/Sources/WKData/Extensions/HTTPURLResponse+Extensions.swift
```diff
@@ -0,0 +1,7 @@
+import Foundation
+
+extension HTTPURLResponse {
```
staykids, line 3: 👍🏽

WKData/Sources/WKData/Service/Protocol/WKURLSession.swift
```diff
@@ -0,0 +1,19 @@
+import Foundation
+
+public protocol WKURLSession {
+    func wkDataTask(with request: URLRequest, completionHandler: @escaping @Sendable (Data?, URLResponse?, Error?) -> Void) -> WKURLSessionDataTask
```
staykids, line 4: My limited understanding of some of the new Swift concurrency changes (actors/actor isolation/et al) makes it unclear to me if the `@Sendable` here and elsewhere is needed or not.

WKData/Sources/WKData/Service/WKBasicService.swift
```diff
@@ -0,0 +1,223 @@
+import Foundation
+
+/// Use this service for the most basic networking service calls. It does not handle authentication.
+public final class WKBasicService: WKService {
```
staykids, line 4: Are you intending `WKService` to serve some more general purpose than effectively performing network requests? If not, `WKNetworkService`, `WKBasicNetworkService`, and `WKNetworkServiceEnvironment` read as far more precise type identifiers to me.

staykids (commented):
@tonisevener More notes below

WKData/Sources/WKData/Data Controllers/WKDonateDataController.swift
tonisevener: Great catch! Done.

WKData/Sources/WKData/Extensions/Dictionary+API.swift
tonisevener: Fixed!

WKData/Sources/WKData/Service/WKBasicService.swift
staykids: I can certainly appreciate the desire to keep it short (`WKService` over `WKNetworkService`) but from my first read of this work it didn't quite click for me. I think I mentally expect `Service` as something more generalized, and there being specializations of it – like what if there's some local data service, or some string localization service? That's just my personal take!

For time sake let's leave this as-is and pick this conversation back up in the future as `WKData` expands in functionality – sound good?

staykids (approved):

WKData/Sources/WKData/Extensions/URL+API.swift
tonisevener: Sorry I missed this one - will make a note to put it in a later PR. Thanks!

PR #4628, opened 2023-09-18
Diff design review feedback updates

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T344453

### Notes
* This PR adds some adjustments to the diff screen based on the design feedback described on the ticket

### Test Steps
1. Run the app and make sure the issues were addressed
2. If possible, test in different sizes of iPhones and iPads, rotate the screen 

###Screenshots


WMF Framework/CommonStrings.swift
```diff
@@ -355,7 +355,7 @@ public class CommonStrings: NSObject {
 
     public static let diffToHeading = WMFLocalizedString("diff-compare-header-to-info-heading", value: "Displayed Edit", comment: "Heading label in info box for current revision when comparing two revisions.")
 
-    public static let diffArticleEditHistory = WMFLocalizedString("diff-article-edit-history", value: "Article Edit History", comment: "Label for article edit history menu item in diff more menu.")
+    public static let diffArticleEditHistory = WMFLocalizedString("diff-article-edit-history", value: "Article edit history", comment: "Label for article edit history menu item in diff more menu.")
```
staykids, line 358: Looking at the Phab ticket, is this supposed to be `Article revision history`?

Wikipedia/Code/DiffListContextViewModel.swift
```diff
@@ -183,7 +184,7 @@ final class DiffListContextViewModel: DiffListGroupViewModel {
             }
             
             height += itemTextHeight
-            height += contextItemPadding.bottom
+//            height += contextItemPadding.bottom
```
staykids, line 187: Commented element.

Wikipedia/Code/DiffListUneditedCell.swift
```diff
@@ -5,22 +5,19 @@ class DiffListUneditedCell: UICollectionViewCell {
...
     func update(_ viewModel: DiffListUneditedViewModel) {
         innerLeadingConstraint.constant = viewModel.innerPadding.leading
-        innerTrailingConstraint.constant = viewModel.innerPadding.trailing
         innerTopConstraint.constant = viewModel.innerPadding.top
         innerBottomConstraint.constant = viewModel.innerPadding.bottom
         
         textLabel.font = viewModel.font
-        textLabel.text = viewModel.text
-        
+        textLabel.text = viewModel.text.capitalized(with: Locale.current)
```
staykids, line 19: An alternative here is just `.localizedCapitalized` which we also use elsewhere.

Wikipedia/Code/DiffHeaderCompareView.swift
```diff
@@ -142,15 +144,32 @@ class DiffHeaderCompareView: SetupView {
...
 
+        setupStackView()
+
+        stackView.setNeedsLayout()
+        stackView.layoutIfNeeded()
+
+        updateFonts(with: traitCollection)
+    }
+
+    func setupStackView() {
```
staykids, line 155: One issue exposed by this change is the metadata/button stack when there's no change message is present doesn't render correctly. I'd expect perhaps that the stack should top aligned in this case, instead of filling in extra space to match the height of its sibling stack, but this may require design input to confirm what the intention should be.

stack_spacing

Wikipedia/Code/DiffListContextViewModel.swift
```diff
@@ -107,10 +107,11 @@ final class DiffListContextViewModel: DiffListGroupViewModel {
     private(set) var expandedHeight: CGFloat = 0
     private(set) var innerPadding: NSDirectionalEdgeInsets
     
-    static let contextItemTextPadding = NSDirectionalEdgeInsets(top: 3, leading: 8, bottom: 8, trailing: 8)
+    static let contextItemTextPadding = NSDirectionalEdgeInsets(top: 15, leading: 8, bottom: 0, trailing: 8)
     static let contextItemStackSpacing: CGFloat = 5
     static let containerStackSpacing: CGFloat = 15
-    
+    private(set) var headerHeight: CGFloat = 35
```
staykids, line 113: I'm curious about the intention of this property – it may be contributing to clipping I'm occasionally noticing with the first compressed element.

clip

staykids (changes requested):
Nice updates - some notes below.

Has this particular branch been design reviewed after these changes (in an experimental build or something)?

mazevedofs:
> Nice updates - some notes below.
> 
> Has this particular branch been design reviewed after these changes (in an experimental build or something)?

Not yet, Robin requested one more small change in the diff's colors, after that, I'll trigger an experimental build

Wikipedia/Code/DiffHeaderCompareView.swift
mazevedofs: I'll get Robin's feedback on this case before updating

Wikipedia/Code/DiffListContextViewModel.swift
mazevedofs: Setting a header height was needed so the div view had the correct spacing to the text, adding it through only a constraint and not adjusting this height caused the cell height to be calculated incorrectly and hide the disclosure button. That new cell height might have thrown things off on this view, I'll take a look at that as well

mazevedofs: Made an adjustment to that text clipping

Wikipedia/Code/DiffHeaderCompareView.swift
mazevedofs: This one will be handled in a separate ticket: https://phabricator.wikimedia.org/T347425

staykids (approved):
Thank you for these updates and for making that separate ticket!

PR #4630, opened 2023-09-18
Watchlist - Allow thanking user from Watchlist view

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T335606 (partially)

### Notes
Enables thanking a user from the Watchlist view. Note this lightly depends on https://github.com/wikimedia/wikipedia-ios-components/pull/35 which removes the thank button for changes by anonymous and bot users.

### Test Steps
1. Delete your app (or reset UserDefaults) to reset onboarding
2. Open Watchlist, and tap a username and thank them
3. Confirm the Thank onboarding modal is displayed and that when Send Thanks is tapped, a success toast shows at bottom of screen
4. Confirm subsequent thank attempts don't show onboarding


Wikipedia/Code/WMFAppViewController+Extensions.swift
```diff
@@ -231,9 +231,33 @@ extension WMFAppViewController: WKWatchlistDelegate {
             navigate(to: siteURL.wmf_URL(withPath: "/wiki/User_talk:\(username)", isMobile: true))
         case .userContributions:
             navigate(to: siteURL.wmf_URL(withPath: "/wiki/Special:Contributions/\(username)", isMobile: true))
-        case .thank:
-            // TODO
-            break
+        case .thank(let revisionID):
```
tonisevener, line 249: @staykids Do you think we could lean on [this protocol](https://github.com/wikimedia/wikipedia-ios/blob/main/Wikipedia/Code/ThanksGiving.swift) that diffs use? It looks like it's already set up with a login and onboarding prompt if needed, plus some logging. I don't think there's a reason it needs to be limited to `ViewController` classes, but I could be wrong.

staykids: 👍🏽 sounds good – thanks for bringing it up to begin with!

tonisevener (approved):
Looks good!

PR #4635, opened 2023-09-21
Watchlist - fix next/prev diff animation

staykids:
### Notes
This fixes an issue where the diff next/prev animation (when navigated to from Settings > Watchlist) didn't animate as expected. It occurred because the settings navigation controller wasn't set up to the use the same delegate as elsewhere, so it didn't have knowledge about the custom transition animator and defaulted to the navigation push animation. Thanks to Toni for the assist in identifying the probable originating issue.

I'm unclear about spiral regressions from this change, but so far navigating the top level settings pages after this change hasn't resulted in any regressions.

### Test Steps
1. Open Settings > Watchlist
2. Tap a change to get pushed to the revision comparison view
3. Tap the next/prev diff navigation buttons in the bottom toolbar
4. Confirm the standard fade animation between diffs occurs (previously, it'd use a navigation stack push animation in this context)


tonisevener (approved):
Working well for me!

PR #4636, opened 2023-09-21
Watchlist - use HTML removal logic from client app for displayed comments

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T335606#9188177 (see item 2)

### Notes
Uses the HTML removal logic from the main client app instead of the simplified Components framework regex. Two things I'm unhappy about here are the name and transiting it via `LocalizedStrings` (which isn't _technically_ the correct place but it's living beside the similarly functioning client closure for bytes and saved an extra initializer to `WKWatchlistViewModel`) – can leave as-is or happy to consider any alternatives you all may have.

### Test Steps
1. Add an article that has changes with special characters (like quotation marks) in the revision comments to your watchlist – today's featured article is a good candidate
2. Confirm viewing the comments in your Watchlist doesn't show HTML escaped characters/tags (see Phab ticket comment for screenshots)


tonisevener (approved):
Working well for me, and I think the approach is fine as an interim thing.

PR #4637, opened 2023-09-21
Snap to parent `WKTheme` on dim theme selection

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T346449

### Notes
Snaps to closest `WKTheme` for `WKAppEnvironment` when a dimmed client theme is selected. I hadn't realized the dim images option constructs a whole different theme name, which is why in watchlist selecting a dim dark/dim black theme defaulted to our default option light. 

The approach here is probably fine for the Components work we have now, but is likely not sustainable in the future depending on the expectation of theme options like dim in future work, so if we need to hold merging this until further thinking I understand.

### Test Steps
1. Open Settings > Theme, select the dark or black theme with dim images off.
2. Open Settings > Watchlist, confirm it is themed correctly.
3. Open Setttings > Theme, select the dark of black theme with dim images off.
4. Open Watchlist again and confirm it is themed correctly.


tonisevener (approved):
Good catch - I hadn't realized this either.

PR #4644, opened 2023-09-28
Watchlist Onboarding Fixes

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T335609

### Notes
These are a couple of fixes I wanted to get in for Watchlist Onboarding.

63f3827 changes the primary button tap target so that the blue area registers as a tap. Before only the text triggered the tap.
a593d29 moves the presentation logic into `ViewControllerRouter` so that tapping "View Watchlist" from the article and diff view toasts also trigger onboarding if needed.

### Test Steps
1. Trigger onboarding from Account. Confirm you can tap the blue area on the primary button and go to your Watchlist.
2. Reinstall app. From an article, watch an article from the More menu, then tap the "View Watchlist" button from the success toast. Confirm you see onboarding. Confirm repeating this step does not show onboarding.
3. Repeat step 2 from the diff view.


mazevedofs (approved):
Working great!

PR #4647, opened 2023-10-01
Feature: Check username validity on text field return

b-onc:
Hello 👋 

I've recently downloaded and the iOS app and when trying to create a wikipedia account, I've gotten too many "username not available" errors after typing every field (especially Captcha is annoying) and tapping "Create". This frustrated me so I went to the web and tried to create an account: to my surprise, web UI checked username validity on text field return, which didn't require me to input all the fields like password.

So I've inspected the API required and implemented it in the iOS app. I couldn't find any official documentation on this API so I've just tested with 3 username types: taken, not available, and available. I'm opening this PR as a draft since I'm sure it needs some changes before being mergeable. I'm not sure the response struct I've implemented can handle all responses this API could return.

Video of the feature:

https://github.com/wikimedia/wikipedia-ios/assets/770464/0dba5e61-4fb4-4ecf-bac1-5216f6f9e090


tonisevener:
@b-onc Hello! Thank you for this PR. The team has agreed that this is a good improvement to pull into a near-future release. We are currently wrapping up a couple of releases in the next two weeks, but after that we should have some time to take a look at this. In the meantime I'll try to find some answers for you on the expected API responses here.

tonisevener:
@b-onc Thanks for your patience on this. I found the web-side handling of this [here](https://gerrit.wikimedia.org/r/plugins/gitiles/mediawiki/core/+/refs/heads/master/resources/src/mediawiki.special.createaccount/signup.js#62). I think this logic is close enough to our need to check your implementation against. Note here that the web adds `errorformat=html&errorsuselocal=true` to the API call, which changes the `cancreateerror` structure slightly.

1. If no users return or "invalid" flag is true, display "You have not specified a valid username."

2. Else If `userid` is populated, display "Username entered already in use. Please choose a different name."
3. Else If `cancreate=false`, display error text in the `cancreateerror.html` property.
4. Else If the username returned does not equal the username input, display info label "Your username will be adjusted to {new username} due to technical restrictions."


I think your simplification of checking against the `canCreate` flag and displaying the generic "not available" error works well enough here. If you want to go the extra mile and add the different handling for points #1 and #4, feel free, but it's not a blocker for approval.

I also have a collection of usernames [here](https://en.wikipedia.org/w/api.php?action=query&errorformat=html&errorsuselocal=true&formatversion=2&list=users&usprop=cancreate&ususers=1.2.3.4%7CCatrope%7CVandal01%7Cgrassisgreen2&format=json) that you can test against. One thing that threw me is that there can be both existing users with no flags (Catrope), as well as users with `missing=true` but an error object that indicates it is an existing user (Vandal01). I'm not sure why there is a difference here, but as long as both names show an error I think you're good here.

I hope this helps! Let us know if you have any other questions.

tonisevener (approved):

tonisevener:
@b-onc This looks good to me. I will post a followup PR with some minor PR feedback and design feedback. Thank you for your contribution!

PR #4657, opened 2023-10-11
Add missing user agent and json param to fundraising endpoints

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T345847

### Notes
These are a couple of final things needed for payment submission. I populated the User-Agent header in the requests with a custom user agent that we are populating elsewhere in the app. I used a `userAgentUtility` workaround in `WKDataEnvironment`. I don't love this since I was trying to keep `WKBasicService` free of this communication pattern, but it did allow me to spin it up quickly.

I also added a missing json format parameter expected on `submitPayment`.

### Test Steps
It works now in production! If you want to submit a donation, add a working card to Apple Wallet. A quick workaround before testing is to comment out [this conditional check](https://github.com/wikimedia/wikipedia-ios/blob/main/Wikipedia/Code/WMFSettingsViewController.m#L298) so that you can get to the payment method prompt in Settings.

1. Trigger payment method prompt in Settings. Tap to get to the native form.
2. Donate a small amount. Payment should succeed. If watching traffic you should see the User Agent passed along in the header.


staykids (approved):
👍🏽 I don't have a personal card on my test device so I didn't test an actual transaction.

PR #4658, opened 2023-10-12
Log preset_selected in the applepay_ui_confirm event

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T346086

Link to more discussion: https://docs.google.com/presentation/d/1vZZwXb0rOGFlqvSy5PxxqRVy83ORPBAwYVHisfp4_o0/edit?disco=AAAA6dNQPAg

### Notes
Analytics wants to be able to track if a user entered a custom amount in the donate form. I had difficulty hooking into this from the SwiftUI currency textfield without a lot of programmatic misfires, so instead we are sending an additional flag upon authorization of the payment sheet indicating if they have an amount preset button selected. 

```
{
  "app_session_id" : "066fe4b6c88c3891e8bf",
  "primary_language" : "en",
  "active_interface" : "applepay_processed",
  "action" : "applepay_ui_confirm",
  "action_data" : "...preset_selected:true...",
  "platform" : "ios",
  "is_anon" : false,
...
```

### Test Steps
Local-only changes: Add `WKDataEnvironment.current.basicService = WKMockBasicService()` to [this area](https://github.com/wikimedia/wikipedia-ios/blob/main/Wikipedia/Code/WMFAppViewController%2BExtensions.swift#L512) to fake all the payment endpoints.

1. Log into iOS Sandbox Tester Apple ID (see 1Password) in test device iOS Settings. Do not enable 2 factor auth.
In Wallet app, add [Test Apple Pay card](https://developer.apple.com/apple-pay/sandbox-testing/).
2. Change device region to NL or IT.
3. On Staging scheme, fresh install and launch app.
4. In app Settings, turn on logging.
5. Go through Donate with Apple Pay form, after authorizing on payment sheet you should see an payload printed out to the console similar to the one above. Confirm `action_data` has `preset_selected:true` if an amount button is selected, and `preset_selected:false` if no amount button is selected.


staykids (approved):
👍🏽 looks like `preset_selected` is populating as expected in the `actionData` property during logging event initialization.

PR #4666, opened 2023-10-26
Fix Simplified Chinese issues

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T348723

### Notes
I think back when we worked on #4550, we should have switched to a `zh-Hans` variant code instead of `zh-Hans-CN`. Looking at the [API call](https://www.mediawiki.org/w/api.php?action=query&format=json&meta=siteinfo&formatversion=2&siprop=languages%7Clanguagevariants&siinlanguagecode=en) referenced in the task, that seems to be the expected bcp47 code.

### Test Steps
1. Change device language to Simplified Chinese
2. Fresh install app
3. Confirm the you see a language selected in onboarding, the Explore feed loads, and you can login, logout and select the Support Wikipedia cell without crashing.
4. Confirm loading an article shows Simplified Chinese characters.


mazevedofs (approved):
Working great!

PR #4672, opened 2023-11-08
Native Source Editor - Base styling + popover theme and font size support

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T347925 (popover theme and font size parts only)

### Notes
This PR sets up some foundational classes for the native text editor work. This includes some initial text storage handling and formatters. The goal of the formatters will be to provide styling and syntax highlighting to the attributed strings. In the future I will build extensions off of them to also support insert / delete logic in response to the keyboard buttons.

So far this only has one base formatter to provide the base font size, color, and paragraph style (line height). In the next PR we will start to see this come together with bold / italic support.

As I was working on this I decided to build out whatever TextKit 2 support I could. One of the major concerns I had earlier this year was the poor scrolling performance with TextKit 2 on iOS 16. Scrolling has improved on iOS 17. I have moved all text-framework-specific communication logic into a `WKSourceEditorTextFrameworkMediator` class (open to name changes here). The idea with this is to make `WKSourceEditorView` unaware of which TextKit framework it's dealing with. Once we move to TextKit 2 we should only need to change the Mediator class.

The scrolling performance issue is why you see the "needsTextKit2" boolean set to true for iOS 17+. Feel free to change it to 16 to see the scrolling issue.

### Test Steps
1. Launch app on Staging iOS 17 (TextKit 2). Go to a large article. Tap "Edit" in the navigation bar to edit the full article wikitext.
2. Confirm text scrolls quickly. Confirm you can change the theme and font size in the popover.
3. Repeat steps 1 & 2 on iOS 15 or 16 (TextKit 1).


mazevedofs (approved):
Working great! Nice to see the improvements from TextKit2!

PR #4678, opened 2023-11-15
Log watchlist filter button tap event

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T343040

### Notes
Adds missing call to log filter button tap event.

### Test Steps
1. Set a breakpoint here: https://github.com/wikimedia/wikipedia-ios/blob/38acaf337b5867458b4ac74e29bc8c3851db7b12/Wikipedia/Code/WatchlistFunnel.swift#L264
2. Open Watchlist, tap Filter button
3. Confirm breakpoint for event is hit


tonisevener (approved):
Thanks for fixing this!

PR #4679, opened 2023-11-15
[HOLD] Native Source Editor - Bold/italic syntax highlighting support

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T347925 (bold/italic/bolditalic syntax highlighting, syntax highlighting toggle support)

### Notes
_Note: This PR is dependent on #4672. Marking as hold while we wait on that._

This PR adds the bold / italics syntax highlighting to the text editor. It also adds support to the "show syntax highlighting" toggle in the popover.

### Test Steps
1. Launch app on Staging iOS 17 (TextKit 2). Go to a large article. Tap "Edit" in the navigation bar to edit the full article wikitext.
2. Confirm you can see bold and italic syntax highlighting as you scroll.
3. Add bold (''') or italic ('') wikitext around words. Confirm they update properly.
4. Turn syntax highlighting off from toggle. Confirm syntax highlighting disappears.
5. Repeat steps 1 - 4 on iOS 15 or 16 (TextKit 1).

PR #4681, opened 2023-11-17
[HOLD] Native Source Editor - Bold/italic button selection, insertion & deletion support

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T347925 (button selection states for bold and italic, formatter inserts / deletions for bold & italic)

### Notes
_Note: This PR is dependent on #4679. Marking as hold while we wait on that._

This PR automatically sets the bold and italic button in selected or unselected states, depending on where the text view selection or cursor is. It also adds the bold / italic wikitext insertion and deletion logic upon button tap. 

### Test Steps
1. Go to editor on Staging.
2. Scroll down and select some already-formatted bold text. Confirm bold buttons in keyboard menus show as selected.
3. Tap bold button to unselect. Confirm bold formatting is automatically removed around selection.
4. Tap bold button again to select. Confirm bold formatting is automatically added around selection.
5. Place cursor somewhere not within a bold range. Tap the bold button. Confirm you see bold wikitext inserted around cursor.
6. Repeat steps 1 - 5 for italics.


mazevedofs (approved):
I'm not sure this was intended so I'm not holding the PR for this (especially as it's not live yet), but when tapping the bold button outside a bold range (on a zero length range) and then clicking again now inside this new bold range, it doesn't remove the bold characters, it adds more (same for italics). It doesn't happen when a range of 1 char or more is selected.
Screenshot 2023-11-23 at 13 39 15

PR #4683, opened 2023-11-28
Perform updated SwiftLint autocorrections

staykids:
**Phabricator:** N/A

### Notes
A recent version of SwiftLint has added more automatic corrections. This PR adds those changes. Mostly, it removes empty subclass overrides and unchanged initializers where the memberwise initializer suffices.

### Test Steps
1. Confirm unit tests pass
2. Very lightly regression test Explore feed and the Article view


tonisevener (approved):
Looks good!

PR #4684, opened 2023-11-29
Diffs: top align from/to container stack view

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T347425

### Notes
Fixes an alignment issue where the from/to revision metadata was not rendering as expected.

### Test Steps
1. Open a diff comparison view between two revisions where one revision does not contain an edit description. (Here's an example I tested with: https://en.wikipedia.org/w/index.php?title=YouTube&diff=1187392832&oldid=1187244209)
2. In the header from/to edit metadata stacks, confirm the labels appear top aligned and are not filled with extraneous inter-item space.

diff_header


tonisevener (approved):
Looks good, thanks!

PR #4685, opened 2023-11-30

translatewiki:
Translation updates


tonisevener (approved):
Approving, however I don't think the `dga` translations will be accessible. I do not see iOS support for it yet.

PR #4688, opened 2023-12-01
Native Editor - Template support

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T348074

### Notes
This PR adds template syntax highlighting and button action support. It highlights all text within curly braces {{ }} as purple. This is different than our current code mirror editor, which is able to keep the parameter values black. I can do a followup to try and match this support, but I decided to match Android at this stage for simplicity.

I am trying to highlight both inline templates and templates that span across multiple lines. It seems to work pretty well, though I'm sure I missed some cases. Please note any egregious bugs that you find with this.

Deselecting a template (thus removing the surrounding {{ }}) works on inline templates, but I avoided it on templates that span across multiple lines. From what I could tell our current editor doesn't support that either.

This PR is dependent on #4687, do not merge until that is in.

Note: Template syntax color was taken from https://phabricator.wikimedia.org/T344508, which now a very light red color on Dark and Black themes

### Test Steps
1. In Staging scheme, load any article via "Edit" button in article nav bar.
2. Confirm templates are syntax highlighted, both inline and multiline. 
3. Change theme and font size, confirm text changes as expected
4. Toggle off and on syntax highlighting in popover, confirm text changes as expected
5. Put a cursor inside an _inline_ template. Confirm template buttons are selected. Confirm deselecting template button removes surrounding {{ }}
6. Pur a cursor inside a _multiline_ template. Confirm template buttons are not selected. Confirm selecting template button adds new {{ }} at cursor point.
7. Select a random word in the introduction. Confirm template button is not selected. Confirm selecting button adds surrounding {{ }}.
8. On a new line, confirm template button is not selected. Confirm selecting button adds surrounding {{ }} and a space where your cursor is. Confirm deselecting template button removes it.

### Screenshots/Videos
Note: we do not yet have link syntax highlighting, so screenshots differ a bit with Android.

Android:
android
android2

iOS:
ios
ios2


mazevedofs (approved):
Working great 🎉

mazevedofs:
UITests actually passed; I was editing the flow at the same time, so it might have caused it to fail. I'm trying to prevent them from running every time

PR #4696, opened 2023-12-07
Bug fix - Native Editor new input view bug

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T344508

### Notes
My iPad mini (device-only) would show an empty input view for the native editor. I couldn't figure out why this was happening, but I suspect it had something to do with the embedded view controllers or navigation controller, since simple test views displayed fine for me.

I rewrote and simplified these views to be a single input view, to more closely match what I've seen proposed from design. The UI is not complete, but I think we can merge this first and fine-tune in a followup PR once T344508 is updated with final mocks.

### Test Steps
1. On Staging, load editor.
2. Notice there's only one format button now rather than two in the input accessory view. Tap it.
3. Confirm the keyboard input view displays.

### Screenshots/Videos

Before (iPad mini bug):
out2

After:
out1


mazevedofs (approved):
Looking good on my devices/simulators 🎉

PR #4701, opened 2023-12-12
Fix widget presentation on iOS 17

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T346810

### Notes
Conforms to iOS 17's container background API for widgets to fix the presentation of widgets on the home screen when built with Xcode 15. As a result, also opts into [StandBy](https://support.apple.com/guide/iphone/use-standby-iph878d77632/ios) on iPhone.

### Test Steps
1. Build this branch using Xcode 15, and run on iOS 17 device or simulator.
2. Add one of each type of widget to the home screen and confirm they render without the margins seen in Phab task (can also confirm they render without margins in the widget gallery when adding widgets)
3. Optionally, enable [StandBy](https://support.apple.com/guide/iphone/use-standby-iph878d77632/ios) and confirm each of the small variant widgets render as expected.

### Screenshots

Here're screenshots of the widgets in [StandBy](https://support.apple.com/guide/iphone/use-standby-iph878d77632/ios) mode:

standby


tonisevener (approved):
Looks good so far, I'd like to hold merging until we're past releasing 7.4.6, since we'll have to change our Xcode version.

PR #4715, opened 2024-01-18
Native Editor - add find

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T348091

### Notes
This adds find functionality to the native editor. I will put up the replace pieces in a followup PR. 

Note:
I notice that iOS 17 / TextKit 2 doesn't always reliably scroll to the correct selected match. I haven't found a quick fix for this yet, but if it becomes too big of a problem we can easily switch to the TextKit 1 implementation.

### Test Steps
1. Launch app in Staging scheme, navigate to the editor.
2. Tap the find formatting button
3.  Type in an existing word in the find text field. Confirm you see find results highlighted and counts displayed. 
4. Tap the previous and next buttons, confirm you can traverse matches.

### DIFF
--- a/Components/Sources/Components/Components/Editors/Common Views/Input Accessory Views/Find and Replace/WKFindAndReplaceView.swift
+++ b/Components/Sources/Components/Components/Editors/Common Views/Input Accessory Views/Find and Replace/WKFindAndReplaceView.swift
@@ -137,7 +137,7 @@ class WKFindAndReplaceView: WKComponentView {
     
     @IBAction private func tappedFindClear() {
         findTextField.text = ""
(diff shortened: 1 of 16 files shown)


mazevedofs (approved):
It's looking pretty good! I did notice the issue you mentioned on iOS 17, but most times it actually scrolls correctly to the highlighted match (I tested searching COVID on this [article](https://en.wikipedia.org/wiki/COVID-19_pandemic_in_the_United_States) and it wasn't so bad. (But I'm biased towards textKit 2 as it's sooo much better on most stuff 😅 )

PR #4716, opened 2024-01-19
Native Editor - add replace

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T348091

### Notes
This adds replace functionality to the native editor. I will put up the replace pieces in a followup PR. Note that this PR is dependent on #4715, so that PR should be merged first.

### Test Steps
1. Launch app in Staging scheme, navigate to the editor.
2. Tap the find formatting button
3. Type in an existing word in the find text field. Confirm you see find results highlighted and counts displayed.
4. Add replace text and tap the replace button (magnifying glass). Confirm the match is replaced with the new word and is highlighted blue. The next match should then be highlighted dark yellow.
5. Repeat steps 1 - 3.
6. Add replace text, then tap the 3 dots and choose "Replace all". Confirm all of your matches are replaced with your new word. Confirm they are highlighted blue.


mazevedofs (approved):
Looking 🔍  great!

PR #4721, opened 2024-01-22
Native Editor - Allow full article editing (part 1)

tonisevener:
**Phabricator:**
https://phabricator.wikimedia.org/T331935

### Notes
This PR moves the full article entry point and allows publishing through the rest of the edit flow. There are other requirements in T331935, such as error handling and the destructive alert view that I will add in a separate PR.

### Test Steps
1. On Staging, load an article on Test Wiki.
2. Tap the overflow menu button in the bottom article toolbar, tap Edit.
3. Add or edit text, then tap Next to go through the rest of the edit flow. Tap Publish to publish your edit.
4. Confirm you land on article, and article reloads to reflect your changes. Confirm you see "Your edit was published" toast.

Note: I have noticed Test Wiki sometimes does not reflect the latest edit. I haven't noticed this with EN wiki edits, so I think this is an issue in the PCS caching layer. I'll ask about that. This task is also still in design's column, so marking as Draft until that is confirmed.


mazevedofs (approved):
Looks good!

PR #4719, opened 2024-01-22
Native Editor - Add basic undo / redo support

tonisevener:
**Phabricator:**
https://phabricator.wikimedia.org/T348088

### Notes
This PR connects the undo / redo buttons to the editor's undo manager. It also disables the progress button until the text has changed.

### Test Steps
1. On Staging, open an article editor. Undo, Redo, and Next buttons should be disabled.
2. Make a change in the text view. Confirm undo button enables.
3. Tap undo button. Confirm change is undone.
4. Tap redo button. Confirm change is redone.
5. Confirm Next button only enables when there are changes to publish.

PR #4723, opened 2024-01-23
Native Editor - Add close confirmation

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T348092

### Notes
This is a quick copy of our destructive alert from the legacy editor.

### Test Steps
1. Open editor on Staging
2. Close, confirm you see alert confirmation and it works as expected.


Wikipedia/Code/PageEditorViewController.swift
```diff
@@ -290,7 +301,17 @@ extension PageEditorViewController: SectionEditorNavigationItemControllerDelegat
     }
     
     func sectionEditorNavigationItemController(_ sectionEditorNavigationItemController: SectionEditorNavigationItemController, didTapCloseButton closeButton: UIBarButtonItem) {
-        delegate?.pageEditorDidCancelEditing(self, navigateToURL: nil)
+        
+        if navigationItemController.progressButton.isEnabled {
```
tonisevener, line 305: I added better enabled button handling in https://github.com/wikimedia/wikipedia-ios/pull/4719, so this conditional should keep the alert limited to only a textview with changes once that PR is in.

mazevedofs (approved):
Looking good!

PR #4724, opened 2024-01-23
Native Editor - crash fixes

tonisevener:
**Phabricator:** 
N/A

### Notes
This PR contains a crash and bug fix.

I came across the crash when selecting multiple lines on iOS 17. This spooked me so I added a defensive check any time we are evaluating an attributed string against a range across the formatters. I also removed any redundant conditionals checking for the same thing.

I also fixed a highlighting bug when undoing after find and replace. 

### Test Steps
1. On iOS 17, go to the first bold text and drag to select multiple lines of wikitext.
2. Confirm there's no crashes.
3. Light regression test, confirm syntax highlighting and selected state generally works as it did before.
-------
1. On editor, find a word and tap the replace icon to replace it.
2. Close out of Find and Replace mode, then tap the Undo button.
3. Confirm your replace is undone, and the background color is not highlighted.


mazevedofs (approved):
Working great 🎉

PR #4729, opened 2024-01-29
Native Editor - Input view design + icon work

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T344508

### Notes
This wraps up the task with finalizing the new input view design as requested in [this comment](https://phabricator.wikimedia.org/T344508#9400570). I also did another pass at icons in the toolbars, find and replace, and navigation bar to improve dynamic type support.

### Test Steps
1. Open an article editor on Staging. Do light regression testing on buttons.
2. Tap the new text format icon to show the new input view. Do light regression testing on buttons.
3. Tap the reading themes navigation button. Do light regression testing on controls.
4. Confirm theming works on buttons and new input view (note there may be further tweaks on colors once this goes through design review).
5.  Change dynamic type. Confirm toolbar buttons, navigation bar buttons and new input view buttons expand or shrink as needed.


Components/Sources/Components/Style/WKIcon.swift
```diff
@@ -66,85 +67,93 @@ public enum WKSFSymbolIcon {
...
         case .pencil:
-            return UIImage(systemName: "pencil", withConfiguration: configuration)?.withRenderingMode(.alwaysTemplate)
+            image = UIImage(systemName: "pencil", withConfiguration: configuration)
         case .plusCircleFill:
-            return UIImage(systemName: "plus.circle.fill", withConfiguration: configuration)?.withRenderingMode(.alwaysTemplate)
-
+            image = UIImage(systemName: "plus.circle.fill", withConfiguration: configuration)
+        }
+        
+        image = image?.withRenderingMode(.alwaysTemplate)
```
mazevedofs, line 164: Was the idea to use the `renderingMode` method argument here?

Components/Sources/Components/Components/Editors/Common Views/Base/WKEditorToolbarButton.swift
```diff
@@ -52,49 +55,87 @@ class WKEditorToolbarButton: WKComponentView {
...
     override func appEnvironmentDidChange() {
-        updateColors()
+        
+        guard let button else {
+            return
+        }
+        
+        let buttonConfig = createButtonConfig(image: image)
+        button.configuration = buttonConfig
+        button.configurationUpdateHandler = buttonConfigurationUpdateHandler(button:)
```
mazevedofs, line 65: Cool! I didn't know about this!

mazevedofs (approved):
Looking great! (the comment I left is a minor detail so I'm merging it)

Components/Sources/Components/Style/WKIcon.swift
tonisevener: @mazevedofs Good catch! Originally I thought I might need to be able to pass different rendering modes to support both palette and monochrome icons, but it turns out `.alwaysTemplate` worked for both. I forgot to remove the parameter after that. I'll add that cleanup to my next PR.

Components/Sources/Components/Components/Editors/Common Views/Base/WKEditorToolbarButton.swift
tonisevener: @mazevedofs yep, it's the newer style of button configuration for UIKit. I can't remember if this was required for this particular button, but I needed it to add padding to the buttons in `WKEditorMultiButton` without triggering project warnings. We'll need to use a similar approach in `WKSmallMenuButton` to remove those warnings.

PR #4735, opened 2024-02-05
Native Editor - scroll to article selected text

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T348291

### Notes
This PR ports the javascript logic located in [https://github.com/wikimedia/wikipedia-ios/blob/main/Wikipedia/assets/codemirror/codemirror-editTextSelection.js](codemirror-editTextSelection.js) to determine the range of wikitext to highlight once a user lands on editor after tapping "Edit" from the article selection context menu. There are still occasionally times when it doesn't find text, but I think in general it works as well as it did before if not a little better for non-Latin scripts.

I put this logic in a WKData utility struct called `WKWikitextUtils`, in hopes that we can reuse it in other contexts in the future.

### Test Steps
1. Go to an article. Select some text somewhere and tap "Edit" from the context menu that appears.
2. Confirm when editor loads you are scrolled to the correct location.


mazevedofs (approved):
Looking good!

PR #4737, opened 2024-02-05
Native Editor - VoiceOver improvements

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T347870#9500666

### Notes
This PR attempts to smooth over some of the editor VoiceOver handling.

### Test Steps
1. Navigate the editor via voice over.
2. Confirm focusing the text view reads out "Wikitext editor"
3. Double-tap to activate. Depending on how fast your device is, you may hear "Loading editor text" before the text view is activated. This attempts to smooth over the complaint in the Phab task comment.
4. Confirm opening the image wizard via VoiceOver focuses on the Back button first (Before focus seemed stuck on editor text view underneath).
5. Confirm an edit notice focuses on the "Edit notices" title first.


Components/Sources/Components/Components/Editors/Source Editor/WKSourceEditorTextFrameworkMediator.swift
```diff
@@ -13,6 +13,15 @@ fileprivate var needsTextKit2: Bool {
     }
 }
 
+class WKSourceEditorTextView: UITextView {
+    override func accessibilityActivate() -> Bool {
+
+        UIAccessibility.post(notification: .announcement, argument: "Loading editor text.")
```
staykids, line 19: Patterning usage elsewhere, should we localize this argument string?

Components/Sources/Components/Components/Editors/Source Editor/WKSourceEditorLocalizedStrings.swift
```diff
@@ -66,7 +67,7 @@ public struct WKSourceEditorLocalizedStrings {
...
...lity: String, keyboardItalicsButtonAccessibility: String, keyboardUnderlineButtonAccessibility: String, keyboardStrikethroughButtonAccessibility: String, keyboardReferenceButtonAccessibility: String, keyboardLinkButtonAccessibility: String, keyboardListUnorderedButtonAccessibility: String, keyboardListOrderedButtonAccessibility: String, keyboardIndentIncreaseButtonAccessibility: String, keyboardIndentDecreaseButtonAccessibility: String, keyboardSuperscriptButtonAccessibility: String, keyboardSubscriptButtonAccessibility: String, keyboardTemplateButtonAccessibility: String, keyboardCommentButtonAccessibility: String, wikitextEditorAccessibility: String, findTextFieldAccessibility: String, findClearButtonAccessibility: String, findCurrentMatchInfoFormatAccessibility: String, findCurrentMatchInfoZeroResultsAccessibility: String, findCloseButtonAccessibility: String, findNextButtonAccessibility: String, findPreviousButtonAccessibility: String, replaceTextFieldAccessibility: String, replaceClearButtonAccessibility: String, replaceButtonAccessibilityFormat: String, replaceTypeButtonAccessibilityFormat: String, replaceTypeSingleAccessibility: String, replaceTypeAllAccessibility: String) {
```
staykids, line 70: It's an interesting situation we've backed ourselves into transiting localizations into Components through these structs and this pattern because of, in practice, how verbose and exploded our call sites become. 🤠 To be clear, I'm ok with this pattern because even though it's verbose, it's at least 100% clear about the components data needs over something like an arbitrary `Dictionary`. But it may be worth brainstorming a little during an eng. sync as a group to see if we can think of any cleaner alternatives (for the sake of the client call sites) in the future.

staykids (changes requested):
Nice – one note below about localizing a string. I didn't run into that "Loading editor text" VoiceOver because I suppose my test device isn't slow enough.

Components/Sources/Components/Components/Editors/Source Editor/WKSourceEditorTextFrameworkMediator.swift
tonisevener: 👍 Good catch!

Components/Sources/Components/Components/Editors/Source Editor/WKSourceEditorLocalizedStrings.swift
tonisevener: Agreed

staykids (approved):

PR #4742, opened 2024-02-09
Native Editor - Fix iOS 16 crash

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T357138

### Notes
This adds a little bit of extra safety to the formatter range check method to fix a crash I saw in TestFlight.

### Test Steps
1. Open editor on an iOS 16 device (I don't think you need an iPad, but this crash occurred on an iPad).
2. Scroll all the way down to the last line of text. Select text in that last line. Also try to put cursor at the very end. Confirm no crashes occur.


staykids (approved):
Repro'd and confirmed fix on 16.7 iPhone mini.

PR #4751, opened 2024-02-12
Native Editor - naming cleanup

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T357073

### Notes
This PR contains some class name changes for native editor code. For anything that was reused from our old web editor, I removed the word `Section`, since the new editor is capable of editing the full source now. I also decided to rename `PageEditorViewController` (which was the native-editor app-side container view controller) to just `EditorViewController`. `Page` felt a bit superfluous at this stage of the feature.

Note: The web editor does exist in this branch, so you will see some class names with `Section` in them. That code will be removed entirely in #4743.

### Test Steps
1. Light regression testing on editor.


mazevedofs (approved):
Looking good!

PR #4752, opened 2024-02-14
Native Editor - fix reading themes button crash

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T357541

### Notes
Just a reminder to myself that it's still good to avoid implicitly unwrapped properties. 🙂

### Test Steps
1. Force slow connection so that the editor shows a spinner when it appears. One way to do this is by adding a `sleep(5)` before the `DispatchQueue.main.async` [here](https://github.com/wikimedia/wikipedia-ios/blob/main/Wikipedia/Code/PageEditorViewController.swift#L304).
2. Reading themes button should be disabled as editor loads.
3. Upon editor load, reading themes button should enable. Tap it, ensure there's no crash.


staykids (approved):
👍🏽 Induced crash on `main`, confirmed delayed enabling in this branch resolves it.

PR #4754, opened 2024-02-16
Fix disappearing randomizer dice

spatel:
**Phabricator:** 
https://phabricator.wikimedia.org/T350601

### Notes
* I observed that the randomizer dice was disappearing whenever 3 - 6 was thrown (1 & 2 always work fine)
* The randomizer dice is a web view so I tested `WMFRandomDiceButton.html` (and `WMFRandomDiceButtonRoll.js`)  in isolation on Safari & Chrome. Chrome works absolutely fine but Safari always fails when rolling 3 - 6 (so this is clearly a Safari CSS bug).
* The CSS `rotateX` and `rotateY` values seemed to be the root cause of the issue and it seems that if you over-rotate (or under-rotate) by a fractional amount (e.g. 0.0001), it fixes the Safari animation glitch. For example, when a rotation value was set to exactly 90 degrees the dice would disappear (but if you set it set to 89.9999 or 90.0001, it would appear fine 🤷).
* `WMFRandomDiceButton.html` is a code-generated file (by `generate_dice_roll_html.rb`) so I applied the `0.0001` rotation fix there and re-generated the HTML file from that.
* Tested on iOS16 & 17 (iPhone & iPad) and it worked in all cases

### Test Steps
1. Go to "Explore" tab and select "Random article"
2. Tap the randomizer dice multiple times

#### ❌ Before

https://github.com/wikimedia/wikipedia-ios/assets/478681/9df4e4fa-d40b-4228-807a-d02f538449eb

### ✅ After

https://github.com/wikimedia/wikipedia-ios/assets/478681/eac6374a-e78d-42ec-b77a-7d44edda9783


scripts/generate_dice_roll_html.rb
```diff
@@ -121,11 +121,14 @@
 mxes = [1, 1, -1, -1]
 myes = [1, -1, -1, 1]
 
+# https://phabricator.wikimedia.org/T350601
+rotation_offset = 0.0001 # Fixes a Safari CSS animation bug where the dice disappears
```
spatel, line 125: N.B. This is the main fix - applying a small rotation offset here appears to fix the final animation keyframe. I then ran this Ruby code to re-generate the `WMFRandomDiceButton.html` file.

**Disclaimer: I am neither a Ruby nor a CSS developer (I am a Swift developer) so this is really just based on what I was able to find out. It seems these kinds of workarounds are commonplace with Safari anyway:**
* https://stackoverflow.com/questions/18146511/bug-in-css3-rotatey-transition-on-safari
* https://stackoverflow.com/questions/35186768/safari-css-bug-animation-rotation-direction-incorrect

PR #4755, opened 2024-02-16
Watchlist – [3] Update filter view when new language is added

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T356794 (partially)

### Notes

This is part of a series of PRs that adds the functionality to add new languages from the Watchlist Filter view. This PR replaces the previously presented language view controller, allows a user to select a new preferred language to add to the app/Watchlist, and reloads the Filter view's Wikipedias section.

### Test Steps

1. Open Watchlist
2. Open the Filter View
3. Tap the Add Language accessory row
4. Select a new language
5. Confirm that language is presented in the Wikipedia sections view
6. Confirm dismissing the filter view shows the articles in that new language you're watching (since my watchlist items are primarily English, I watched an article in a different language, removed that language from my preferred languages, tested adding that language via this Filter view to confirm those articles displayed in the main watchlist view)


tonisevener:
These are looking good 👍

Components/Sources/Components/Components/Watchlist/Filter/WKWatchlistFilterViewModel.swift
```diff
@@ -285,6 +275,30 @@ public final class WKWatchlistFilterViewModel {
...
+			Self.section1(projectViewModels: Array(projectViewModels.prefix(2)), strings: localizedStrings),
+			Self.section2(projectViewModels: Array(projectViewModels.suffix(from: 2)), strings: localizedStrings, addLanguageAction: addLanguageAction),
+			Self.section3(strings: localizedStrings, filterSettings: filterSettings),
+			Self.section4(strings: localizedStrings, filterSettings: filterSettings),
+			Self.section5(strings: localizedStrings, filterSettings: filterSettings),
+			Self.section6(strings: localizedStrings, filterSettings: filterSettings),
+			Self.section7(strings: localizedStrings, filterSettings: filterSettings),
+			Self.section8(allChangeTypes: allChangeTypes, offChangeTypes: offChangeTypes, strings: localizedStrings, filterSettings: filterSettings)
+		])
+	}
```
tonisevener, line 300: I think you only really need to reload section 2 here, to update it with the latest `addLanguagesAction` from that `didSet` call. You could replace this method with

```
formViewModel.sections[1] = Self.section2(projectViewModels: Array(projectViewModels.suffix(from: 2)), strings: localizedStrings, addLanguageAction: addLanguageAction)
```

and it still works.

Wikipedia/Code/WMFAppViewController+Extensions.swift
```diff
@@ -283,29 +283,45 @@ extension WMFAppViewController: WKWatchlistDelegate {
...
+            let appLanguages = dataStore.languageLinkController.preferredLanguages
+            var localizedProjectNames = appLanguages.reduce(into: [WKProject: String]()) { result, language in
+                guard let wikimediaProject = WikimediaProject(siteURL: language.siteURL, languageLinkController: dataStore.languageLinkController), let wkProject = wikimediaProject.wkProject else {
+                    return
+                }
+
+                result[wkProject] = wikimediaProject.projectName(shouldReturnCodedFormat: false)
+            }
+            localizedProjectNames[.wikidata] = WikimediaProject.wikidata.projectName(shouldReturnCodedFormat: false)
+            localizedProjectNames[.commons] = WikimediaProject.commons.projectName(shouldReturnCodedFormat: false)
```
tonisevener, line 308: An alternative you could do here is to change [this line](https://github.com/wikimedia/wikipedia-ios/blob/main/Wikipedia/Code/ViewControllerRouter.swift#L237) to `dataStore.languageLinkController.allLanguages`. The filter localized strings will be populated with every possible language name upon init, and then you won't have to bother with updating it at this point after they add a language.

I understand the preference to keep localized strings limited to only will display on screen though, so not a must.

Components/Sources/Components/Components/Shared/Forms/View Models/Base/WKFormSectionViewModel.swift
```diff
@@ -1,6 +1,6 @@
 import Foundation
 
-public class WKFormSectionViewModel: Identifiable {
+public class WKFormSectionViewModel: Identifiable, ObservableObject {
```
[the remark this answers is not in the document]
staykids: 👍🏽 – removed!

Components/Sources/Components/Components/Watchlist/Filter/WKWatchlistFilterViewModel.swift
staykids: My intent with this approach that mimics the initializer was to capture the side effects of the user's filter settings. For example, with this suggested single line change approach, opening the filter settings, turning off a language, hitting done, and returning to the filter view still shows the language as on.

Wikipedia/Code/WMFAppViewController+Extensions.swift
tonisevener: No strong preference for me, just thought I'd mention it in case you were drawn more to it. I'm going to go ahead and approve since it's working well and there's good reason for the current approach.

tonisevener (approved):

PR #4759, opened 2024-02-22
Image recommendations - Add utility class to style HTML

tonisevener:
**Phabricator:**
https://phabricator.wikimedia.org/T356816

### Notes
This PR creates a utility class for converting html strings into NSAttributedString and AttributedString. We need the AttributedString conversion to support attributed SwiftUI text views in the Components framework.

I believe I have covered all of the major needs from our old class in String+Html.swift, though note that I didn't strictly follow it. This utility supports:

- Bold
- Italics
- BoldItalics
- Links
- Lists
- Superscript
- Subscript
- Underline
- Strikethrough

I created some dummy UI to see both UI frameworks in action. Followup PRs for this task will be reusing the article intro view from the `T347121/prototype`, and connecting it to live data.

### Test Steps
1. On Staging or Experimental scheme, launch app.
2. Tap Notifications Center bell icon.
3. The top view is SwiftUI, and the bottom view is UIKit. Confirm styles look correct. Feel free to play with the text in WKImageRecommendationsViewController.swift to test different stylings.

### Screenshots
screenshot2


Components/Sources/Components/Utility/HTMLUtils.swift
```diff
@@ -0,0 +1,513 @@
...
+    private struct TagAndContentRemoveData {
+        let range: NSRange
+    }
+    
+    // MARK: - NSAttributedString - Public
+    
+    public static func nsAttributedStringFromHtml(_ html: String, styles: Styles) throws -> NSAttributedString {
+        let attributedString = NSMutableAttributedString(string: html)
+        let paragraphStyle = NSMutableParagraphStyle()
+        paragraphStyle.lineSpacing = 3
```
staykids, line 58: What do you think about propagating this value through `Styles` as well?

Components/Sources/Components/Utility/HTMLUtils.swift
```diff
@@ -0,0 +1,513 @@
...
+                    return
+                }
+                
+                if case .ordered = currentListType {
+                    if let lastOrderedListCount = orderedListCounts.popLast() {
+                        orderedListCounts.append(lastOrderedListCount + 1)
+                    }
+                }
+                
+                let spaces = String(repeating: "    ", count: indent)
```
staykids, line 382: Same note as `lineSpacing` above – I wonder if it makes sense to future-proof this with an indentation style element in `Styles` (either a `String` or `enum` to represent different indentation styles), and a static property on `HTMLUtils` that represents this string of spaces as a default.

staykids (commented):
Working nicely! Some minor optional stuff below from my first pass for future proofing if it also resonates with you, but happy to continue reviewing and approve as-is.

Components/Sources/Components/Utility/HTMLUtils.swift
tonisevener: Done! One thing that's still a bit awkward here is that SwiftUI adds line spacing as a modifier, outside of the utility. I couldn't find any line spacing key while making AttributedString. I guess we could consider making a `WKHtmlText` component or something, just for SwiftUI, with line spacing applied within so that we don't have to remember to add this modifier every time. Let me know what you think.

staykids: Yeah a component that itself encapsulates those custom modifiers makes sense in the future to me too.

staykids (approved):

PR #4760, opened 2024-02-23
Image recommendations - Add view model for article intro view

tonisevener:
**Phabricator:**
https://phabricator.wikimedia.org/T356816 (View model)

### Notes
This PR adds a view model for the image recommendations article intro view. In order to get this data, I had to add the ability to fetch an article summary in WKData. As I was testing I discovered that the app-side MediaWikiFetcher wasn't flexible enough for this endpoint, due to the way the urls need to be constructed (it was stripping any paths I added from WKData). Rather than adjust the app-side code, I found it easiest to lean instead on WKBasicService for this call. Ultimately we want all calls to originate from WKData so this fits our future plans.

This PR is dependent on https://github.com/wikimedia/wikipedia-ios/pull/4759.

### Test Steps
1. Open app in Staging
2. Set up your app language to be one with image recommendations (Czech)
3. Tap Notifications Center icon.
4. Confirm you see an article title. Tap next button and confirm you can cycle through titles.
5. Go to app languages and change primary to a Chinese variant.
6. Begin proxying and visit an article. Make a note of the Accept_Language and X-WMF-UUID request header values for the `mobile-html` call.
7. Tap Notifications Center.
8. Confirm you see a call go out to the `page/summary` endpoint, and confirm you headers match what you saw in step 6.


mazevedofs (approved):
Looking great!

PR #4764, opened 2024-02-28
Diffs: Enable VoiceOver for diff list cells

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T343582 (partially)

### Notes
This PR re-introduces VoiceOver support for diff list cells. It also auto-expands the cells if the user has VoiceOver enabled.

### Test Steps
1. Enable VoiceOver on device
2. Navigate to a diff page change for an article of your choice
3. Confirm (because VoiceOver is on) that the diff list cells start as expanded
4. Confirm navigating elements from top down navigates in order (previously, the collapse/expand button was read out of order)


tonisevener (approved):
Looks good. I believe the unit test failures are an unrelated bug - https://phabricator.wikimedia.org/T358668.

PR #4769, opened 2024-02-29
Fix PDF image generation error

spatel:
**Phabricator:** 
https://phabricator.wikimedia.org/T357266

### Notes
* When tapping on some PDF files, an error alert appears as the PDF to image conversion fails
* I found several articles with PDFs where this problem occurs (there are potentially many more):
  * https://en.wikipedia.org/wiki/1717–1718_Acts_of_Grace (*🚨good for testing - the first file is a PDF 🚨*)
  * https://en.wikipedia.org/wiki/American_Civil_Liberties_Union_v._Ashcroft (mentioned in original ticket)
  * https://en.wikipedia.org/wiki/Illinois_Freedom_of_Information_Act
  * https://en.wikipedia.org/wiki/United_States_House_Select_Committee_on_the_January_6_Attack

### Test Steps
1. Navigate to any of the pages above
2. Tap on any PDF file within the page

#### ❌ Before
https://github.com/wikimedia/wikipedia-ios/assets/478681/dc8889ee-09ab-4e30-874d-8c4409006f4c

#### ✅ After
https://github.com/wikimedia/wikipedia-ios/assets/478681/4d303b23-63c3-4305-a563-c30df7bae4dd


WMF Framework/ImageCacheController.swift
```diff
@@ -248,7 +248,7 @@ public final class ImageCacheController: CacheController {
         }
         let options = [kCGImageSourceShouldCache as String: NSNumber(value: true)] as CFDictionary
         guard let cgImage = CGImageSourceCreateImageAtIndex(source, 0, options) else {
-            return nil
+            return createImageForRejectedCGImage(with: data, mimeType: mimeType)
```
spatel, line 251: I observed that when trying to create an image from the PDF file, `CGImageSourceCreateImageAtIndex` was failing (and returning `nil` here). My suspicion is that `CGImageSourceCreateImageAtIndex` probably worked fine in the past for PDFs (and possibly still does for some) but that Apple may well have changed something under the hood. So I added a recovery method to try and generate an image from the first page of the PDF (i.e. so it will ONLY fall through to this behaviour as a last resort).

tonisevener (approved):
Working great - thank you for the fix!

PR #4770, opened 2024-02-29
Use web form for Account Vanishing flow

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T336494 &
https://phabricator.wikimedia.org/T356210

### Notes
This PR deletes our old native vanishing form and replaces it with the web form. Initially I had some odd logged out states when switching accounts and reloading the form. I was able to fix it with [this line](https://github.com/wikimedia/wikipedia-ios/compare/vanishing/web-view-flow?expand=1#diff-fa94d2fed583475541c9cce141bc15abd17d399a0b8674536af6655c585a3325R57), which essentially turns our web view into a private browsing session. I don't think this will cause major ill effects, since we're copying the session cookies over upon load. We'll probably drop cookies saved from other in-app web views though. I tested some of the donation payment methods and didn't see any issues.

### Test Steps
1. Log into app
2. Go to a user page. You can get there through article revision history > user name button > user page
3. This will load the mobile web view, tap the bottom Desktop button in the footer.
4. On this new web view, confirm you see a logged in state in the top right corner.
5. Now go to the Vanishing view under Account.
6. After warning, confirm web view loads and the correct username is populated.
7. Feel free to submit the form for a test account if you want (I already tested this, and QA can test it more).
8. Go back to Account and log out. Go to another user page > Desktop web view and confirm you are in a logged out state.
9. Log into a different account.
10. Go to vanishing form. Confirm correct username is populated.


mazevedofs (approved):
Looks good!

PR #4771, opened 2024-03-02
Diffs: Update cell accessibility labels

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T343582 (partially)

### Notes
This PR improves the accessibility of the diff view with targeted improvements when VoiceOver is enabled.

Changes include:
- Adds an "accessibility text" element per diff view model to customize the VoiceOver experience
- Changes to the diff change cell labels to "source code" context to read as many characters as possible
- Updates the VoiceOver experience for addition/removal of content (in order to communicate the red/green deletion and insertion styling)
- Updates contextual cell lines to read "Contextual line, unchanged"
- Updates line/section moving changes to indicate distance from current element
- Updates the collapse/expand buttons to indicate line numbers they will collapse/expand

It would have been nice to leverage some of the attributed string [accessibility keys](https://developer.apple.com/documentation/foundation/nsattributedstring/key) to directly apply accessibility traits on the constructed attributed string without having to construct separate strings for VoiceOver purposes.
But unfortunately, [accessibilityTextCustom](https://developer.apple.com/documentation/foundation/nsattributedstring/key/2865778-accessibilitytextcustom) doesn't appear to work as expected (or I'm misinterpreting its intent).

### Test Steps

I'd recommend we rely on QA for extensive testing, but you can lightly check the changes above in any article diff view: 

1. Open an Article
2. Scroll to article history
3. Confirm VoiceOver is on, then tap a change
4. Confirm the contained diff change area can be navigated via VoiceOver
5. Confirm VoiceOver changes from above (if present) in the particular diff you're on


Wikipedia/Code/DiffListChangeItemViewModel.swift
```diff
@@ -222,6 +225,69 @@ final class DiffListChangeItemViewModel {
...
+                    } else {
+                        modifiedAccessibilityText = String.localizedStringWithFormat(diffMoveSourceSectionsDown, amount)
+                    }
+                }
+
+                modifiedAccessibilityText += text
+            }
+        case .moveDestination:
+            let diffParagraphMoved = WMFLocalizedString("diff-change-paragraph-moved", value: "Paragraph moved. Go to previous location.", comment: "Accessibility text announced in VoiceOver to indicate content has moved.")
+            modifiedAccessibilityText = diffParagraphMoved + text
```
tonisevener, line 283: I notice this "text" can have some integrated deletions and additions. Maybe you could run this text through [this logic](https://github.com/wikimedia/wikipedia-ios/pull/4771/files#diff-3d5f1b06bbd8cbd1d63ac39e6999b4267584e6d5a1918f039e1bd5810635caa8R240-R255) from above to get those "removal" / "addition" bits?

IMG_0336

Wikipedia/Code/DiffListChangeCell.swift
tonisevener, line 1: I think adding the tap gesture recognizer is causing all these change labels to adopt the "Button" trait, which is a bit confusing because it reads out "button" for things that aren't really buttons. In [these lines](https://github.com/wikimedia/wikipedia-ios/pull/4771/files#diff-f5f9d93a4f4f6ee19724923211e8e56c2b97ecfe0c339c9f9ec7d6c2d5f41f3dR108-R112), if you only added the tap gesture recognizer when item.diffItemType == .moveSource or .moveDestination, that may fix it.

tonisevener (commented):
Looking good! Just noted a couple of things that jumped out at me when testing. I also tried to get the focus to shift on paragraph moved cells, because I noticed it wouldn't switch to the new cell that it tries to scroll to. It sort of works, but not perfectly. Feel free to pull it in if you want:

https://github.com/wikimedia/wikipedia-ios/commit/32443a21b62b2f40ea43d366e31293fd29d62667

Wikipedia/Code/DiffListChangeItemViewModel.swift
staykids: Good catch – updated all the diff change types except add/removal lines to use that logic.

Wikipedia/Code/DiffListChangeCell.swift
staykids: Good call – pulled in your commit and made this change.

tonisevener (approved):
Sounds good, thanks!

PR #4775, opened 2024-03-08
Image recommendations - Explore feed card (Part 2)

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T356811

### Notes
This PR wraps up the image recommendations explore feed card. It finalizes the UI, ensures the card is always displayed second, and only displays it if:

1. The user is logged in
2. The user has 50+ edits on the app language wiki
3. The user is not blocked site-wide and
4. The app language wiki has image recommendations to display

I would recommend leaning on QA for every test scenario, but here are some light happy path steps:

### Test Steps
1. Launch app, login with a user that has 50+ edits.
2. Confirm suggested edits card displays 2nd. Try changing theme. Confirm suggested edits card theming looks good.
3. Confirm you can tap card and are pushed to the image recommendations view.
4. Log out
5. Confirm suggested edits card disappears.


mazevedofs (approved):
Looking great! Thanks for also handling the RTL stuff :)

PR #4778, opened 2024-03-14
Image recommendations: Add onboarding modal

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T358896

### Notes
This PR adds the images recommendations onboarding modal. It also updates the `WKOnboardViewCell` to be able to support filling the background state of an icon if required by design. I also added a `WKData` data controller for image recs (similar to how Watchlist was patterned), but if this isn't gonna be our desired way to read/manipulate data (like the user defaults here) we can figure something else out.

### Test Steps
1. Optionally comment out `guard !dataController.hasPresentedOnboardingModal else ...` for repeated presentation of the modal
2. Run on Staging
3. Tap the Add button on the explore feed card
4. Confirm the modal is displayed


WKData/Sources/WKData/Data Controllers/Image Recommendations/WKImageRecommendationsDataController.swift
```diff
@@ -0,0 +1,42 @@
...
+
+	}
+
+	// MARK: - Onboarding
+
+	private var onboardingStatus: OnboardingStatus {
+		return (try? userDefaultsStore?.load(key: WKUserDefaultsKey.imageRecommendationsOnboarding.rawValue)) ?? OnboardingStatus.default
+	}
+
+	public var hasPresentedOnboardingModal: Bool {
```
tonisevener, line 32: The user defaults store can save any Codable type, so you could just store a simple Bool without having to wrap it up in the `OnboardingStatus` struct:

```
    public var hasPresentedOnboardingModal: Bool {
        get {
            return (try? userDefaultsStore?.load(key: WKUserDefaultsKey.imageRecommendationsOnboarding.rawValue)) ?? false
        } set {
            try? userDefaultsStore?.save(key: WKUserDefaultsKey.imageRecommendationsOnboarding.rawValue, value: newValue)
        }
    }
```

The struct is fine if you plan on adding more stuff to it in the future, but if not I think this simplification is a little easier to read at a glance.

tonisevener (commented):
Looking good!

> I also added a WKData data controller for image recs (similar to how Watchlist was patterned), but if this isn't gonna be our desired way to read/manipulate data (like the user defaults here) we can figure something else out.

I noticed watchlist deals with the onboarding flag app-side (but filter struct within WKData), but I would prefer that we store onboarding through WKData like you did here. We already use another data controller called `WKGrowthTasksDataController` which so far only fetches image recommendations. I also prefer that we use your naming of `WKImageRecommendationsDataController` as well, so maybe we can consolidate them both if @mazevedofs is good with it.

WKData/Sources/WKData/Data Controllers/Image Recommendations/WKImageRecommendationsDataController.swift
staykids: Though not explicitly mentioned on the Phab task, I anticipated we'd potentially need to also track the progress of other feature-related onboarding activities (like whether the nice to have tooltips have displayed yet or not), but I'm happy to adjust it if that's not the case, or in whatever manner @mazevedofs feels is best for the feature.

tonisevener (approved):

WKData/Sources/WKData/Data Controllers/Image Recommendations/WKImageRecommendationsDataController.swift
mazevedofs: I think we can keep the struct since it paves the way for the tooltips implementation, we'll eventually get to implement the nice to haves in a future release.

mazevedofs:
> Looking good!
> 
> > I also added a WKData data controller for image recs (similar to how Watchlist was patterned), but if this isn't gonna be our desired way to read/manipulate data (like the user defaults here) we can figure something else out.
> 
> I noticed watchlist deals with the onboarding flag app-side (but filter struct within WKData), but I would prefer that we store onboarding through WKData like you did here. We already use another data controller called `WKGrowthTasksDataController` which so far only fetches image recommendations. I also prefer that we use your naming of `WKImageRecommendationsDataController` as well, so maybe we can consolidate them both if @mazevedofs is good with it.

I think we can consolidate both into the same `WKImageRecommendationsDataController.` I named `WKGrowthTasksDataController` thinking of sharing some basic calls from several tasks, but it doesn't seem to be the case for these APIs and just adds some confusion.

PR #4779, opened 2024-03-15
Updates localized string keys

mazevedofs:
**Phabricator:** N/A

### Notes
* Updates two identical localized string keys (talk-pages-topic-compose-close-confirmation-discard) representing different strings. I thought it should have broken our build, but somehow it slipped through a while ago

### Test Steps
1. Build the app, and make sure no related errors or warnings occur.
2. Start drafting and discard a topic on a talk page. Confirm the alert has the correct title.
3. Repeat the above for a talk page reply.


Wikipedia/Code/TalkPageReplyComposeController.swift
```diff
@@ -12,7 +12,7 @@ class TalkPageReplyComposeController {
     
     enum ActionSheetStrings {
         static let closeConfirmationTitle = WMFLocalizedString("talk-pages-reply-compose-close-confirmation-title", value: "Are you sure you want to discard this new reply?", comment: "Title of confirmation alert displayed to user when they attempt to close the new reply view after entering text. Please prioritize for de, ar and zh wikis.")
-        static let closeConfirmationDiscard = WMFLocalizedString("talk-pages-topic-compose-close-confirmation-discard", value: "Discard Reply", comment: "Title of discard action, displayed within a confirmation alert to user when they attempt to close the new topic view after entering title or body text. Please prioritize for de, ar and zh wikis.")
+        static let closeConfirmationDiscard = WMFLocalizedString("talk-pages-topic-compose-close-confirmation-discard-reply", value: "Discard Reply", comment: "Title of discard action, displayed within a confirmation alert to user when they attempt to close the new topic view after entering title or body text. Please prioritize for de, ar and zh wikis.")
```
tonisevener, line 15: Comment should probably change here.

Maybe something like:
`Title of discard action, displayed within a confirmation alert to user when they attempt to close the reply view after entering reply text.`

tonisevener (commented):
@mazevedofs Great catch! Just one minor thing.

tonisevener (approved):

PR #4787, opened 2024-03-22
Image recommendations: Add survey view

staykids:
**Phabricator:** https://phabricator.wikimedia.org/T358924

### Notes
Adds survey view presented to user when they tap "No" to indicate that they don't recommend the suggestion. This has been design reviewed, but includes requested design changes that have not yet been re-reviewed and confirmed.

### Test Steps
1. Open Suggested edits view
2. Tap "No" on bottom modal sheet suggestion
3. Confirm Survey view is displayed
4. Submit button should be disabled until the user selects an option
5. Confirm typing in the other field enables the checkmark for that field, and tapping a different reason while the field is active resigns the responder status for the textfield and dismisses the keyboard
6. Optionally, you can inspect the submit button's action to confirm the user selected reasons are computed properly when the user taps submit


tonisevener (changes requested):
Looks great! Just a couple of minor things you can remove. Also, if there's a quick modifier you could add somewhere to avoid this iPad split view, I think that would be good to fit in here.

test

Larger iPad improvements will be handled separately in https://phabricator.wikimedia.org/T360036.

staykids:
@tonisevener Apologies – removed that unused WIP code that slipped in before design review. Also updated the navigation view presentation style so it doesn't appear in that sidebar style on iPad.

tonisevener (approved):

PR #4788, opened 2024-03-22
Image recommendations - go to Add Image Details

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T358917

### Notes
* This PR adds the ability to redirect the user to our existing "InsertMediaSettingsViewController" when tapping "Yes" and accepting an image recommendation for an article
* Handling the wikitext is covered in another ticket, so this PR does not go beyond adding info, and does not add any actions to the insert button  
* We still don't have the image displaying here, as handling the image download is yet to be done on WKData
* Note: since both the image title and image description text lengths are unpredictable, I'm setting a hard-coded number of lines for both to limit the size of the header view. With very large texts, it could cause the header to take over most of the view, which I considered to be a bad experience.

### Test Steps
1. On an image recommendation, tap "Yes" on the toolbar
2. Make sure the layout corresponds to Figma and that the "Insert button" is only blue when there is text on the caption fields.
3. Make sure the button is clickable even when greyed out - no action happens but it should be clicked.
4. Go to the Editor and tap Add an image on an article.
5. Make sure it still works as usual.


Wikipedia/Code/InsertMediaSettingsViewController.swift
```diff
@@ -208,8 +218,50 @@ final class InsertMediaSettingsViewController: ViewController {
         tableView.separatorStyle = .none
         tableView.tableHeaderView = imageView
         tableView.tableFooterView = buttonView
+
+        if FeatureFlags.needsImageRecommendations {
```
tonisevener, line 222: > Make sure it still works as usual.

Editor flow is broken here on Staging - the "Next" button doesn't do anything.

Instead of the feature flag, you could set this as a view controller property instead (`fromImageRecommendations`, or something like that), default it to false and set it from the outside at the when you instantiate this from the image recommendations flow. This will leave the editor flow unchanged in the Staging scheme.

Wikipedia/Code/InsertMediaSettingsViewController.swift
```diff
@@ -208,8 +218,50 @@ final class InsertMediaSettingsViewController: ViewController {
...
+            title = WMFLocalizedString("insert-media-add-image-details-title", value: "Add image details", comment: "Title for add image details view")
+            nextButton = UIBarButtonItem(title: WMFLocalizedString("insert-action-title", value: "Next", comment: "Title for insert action indicating the user can go to the next step"), style: .done, target: self, action: #selector(insertMedia))
```
tonisevener, line 224: I see a duplicate keys warning with this.

Wikipedia/Code/InsertMediaSettingsViewController.swift
```diff
@@ -208,8 +218,50 @@ final class InsertMediaSettingsViewController: ViewController {
...
+            nextButton?.tintColor = theme.colors.secondaryText
+            navigationItem.rightBarButtonItem = nextButton
+            navigationController?.navigationBar.topItem?.title = String()
+            self.apply(theme: theme)
+        }
     }
-    
+
+    @objc private func insertMedia(_ sender: UIBarButtonItem) {
+        guard let mediaSettingsTableViewController = navigationController?.topViewController as? InsertMediaSettingsViewController else {
```
tonisevener, line 233: I think you can remove this guard and use the self properties below.

Wikipedia/Code/InsertMediaSettingsViewController.swift
```diff
@@ -161,6 +168,7 @@ final class InsertMediaSettingsViewController: ViewController {
         let headerText: String
         let placeholder: String
         let footerText: String
+        let url: String
```
tonisevener, line 171: Maybe name this `learnMoreUrl`?

tonisevener (changes requested):

Wikipedia/Code/InsertMediaSettingsViewController.swift
```diff
@@ -217,8 +217,51 @@ final class InsertMediaSettingsViewController: ViewController {
...
         tableView.tableHeaderView = imageView
         tableView.tableFooterView = buttonView
+
+        if FeatureFlags.needsImageRecommendations {
+            title = WMFLocalizedString("insert-media-media-settings-title", value: "Media settings", comment: "Title for media settings view")
+
+            let insertButton = UIBarButtonItem(title: WMFLocalizedString("insert-action-title", value: "Insert", comment: "Title for insert action"), style: .done, target: self, action: #selector(insertMedia))
+            insertButton.tintColor = theme.colors.link
+            navigationItem.rightBarButtonItem = insertButton
+            navigationController?.navigationBar.topItem?.title = String()
```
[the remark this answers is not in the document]
mazevedofs: It removes the title of the back button. On the designs, the back button doesn't have a title.

tonisevener (approved):

PR #4796, opened 2024-03-27
Image recommendations - Add feedback API calls

tonisevener:
**Phabricator:**
https://phabricator.wikimedia.org/T360574

### Notes
This PR adds the ability to send feedback API calls when approving or rejecting an image suggestion. This is the first WKData API call against the [MediaWiki Rest API](https://www.mediawiki.org/wiki/API:REST_API). I wasn't able to lean on some of the existing calls in MediaWikiFetcher, so I had to extend it a bit to support PUT and these new url styles.

There's not really an easy way to test this without the connections to the survey and edit summary views, so we can skip testing until that connection is made. I tested these calls by updating this method in ExploreViewController, tapping Yes to the toolbar button and watching the http traffic:

```
    func imageRecommendationsUserDidTapInsertImage(with imageData: WKImageRecommendationsViewModel.WKImageRecommendationData) {

//        let image = UIImage() // TODO: fetch image properly
//        if let imageURL = URL(string: imageData.descriptionURL),
//           let thumbURL = URL(string: imageData.thumbUrl) {
//            let searchResult = InsertMediaSearchResult(fileTitle: imageData.filename, displayTitle: imageData.filename, thumbnailURL: thumbURL, imageDescription: imageData.description,  filePageURL: imageURL)
//            let insertMediaViewController = InsertMediaSettingsViewController(image: image, searchResult: searchResult, fromImageRecommendations: true)
//            navigationController?.pushViewController(insertMediaViewController, animated: true)
//        }

// TEMP CODE
        let language = WKLanguage(languageCode: "en", languageVariantCode: nil)
        let dataController = WKImageRecommendationsDataController()
        dataController.sendFeedback(project: .wikipedia(language), pageTitle: imageData.pageTitle, editRevId: nil, fileName: imageData.filename, accepted: true, caption: nil) { result in
            switch result {
            case .success(let dict):
                print(dict)
            case .failure(let error):
                print(error)
            }
        }
    }
```


WMF Framework/MediaWikiFetcher.swift
```diff
@@ -83,9 +83,51 @@ public final class MediaWikiFetcher: Fetcher, WKService {
...
+        case (.PUT, .mediaWikiREST):
+            if let tokenType = mediaWikiRequest.tokenType {
+                requestMediaWikiAPIAuthToken(for: url, type: tokenType.wmfTokenType) { [weak self] result in
+                    switch result {
+                    case .success(let token):
+                        
+                        var tokenizedParams = request.parameters ?? [:]
+                        tokenizedParams["token"] = token.value
+                        
+                        self?.performPut(url: url, parameters: tokenizedParams, completion: completion)
```
staykids, line 95: In the event that self is nil here, this call could leave the completion block uncalled. Would it make sense to call the completion block with some kind of failure here in the case that self is nil in order to prevent the caller from waiting on a completion that never occurs?

staykids (commented):
Nice – one note below.

WMF Framework/MediaWikiFetcher.swift
tonisevener: @staykids Good catch - this is done!

PR #4798, opened 2024-03-28
`MediaListItem`: Use explicit `MediaListItemType` enum over `String` 

staykids:
**Phabricator:** N/A

### Notes
From a PR I reviewed yesterday, just noticed this use of strings over the type safe enum. This PR updates that struct to use the enum directly.

### Test Steps
Explore articles with multiple images to confirm the gallery experience hasn't regressed.


tonisevener (approved):
Looks good! Thanks for the cleanup.

PR #4800, opened 2024-03-28
Image recommendations -  Data transition to preview article

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T359225

### Notes
* This PR adds a fetch wikitext call to the insert media settings VC when coming from the image recs feature
* It then uses the wikitext utility to compose the new updated wikitext
* No error handling is being done, as there are no checks for user permissions - assuming only unblocked users can access the feature and no locked articles are made available by the API

### Test Steps
1. Go through the image recs feature on Staging
2. Verify the wikitext being printed is correct
3. Article preview is being handled in another ticket
4. Go through the regular add image edit flow, it should work as usual


WKData/Sources/WKData/Data Controllers/WKGrowthTasks/WKGrowthTasksDataController.swift
```diff
@@ -104,7 +104,7 @@ import Foundation
     }
 
     fileprivate func getMetadataObject(from image: WKImageRecommendationAPIResponse.ImageMetadata) -> WKImageRecommendation.ImageMetadata {
-        let metadata = WKImageRecommendation.ImageMetadata(descriptionUrl: image.descriptionUrl, thumbUrl: image.thumbUrl, fullUrl: image.fullUrl, originalWidth: image.originalWidth, originalHeight: image.originalHeight, mediaType: image.mediaType, description: image.description, author: image.author, license: image.license, date: image.date, caption: image.caption, categories: image.categories, reason: image.reason, contentLanguageName: image.contentLanguageName)
+        let metadata = WKImageRecommendation.ImageMetadata(descriptionUrl: image.descriptionUrl, thumbUrl: image.thumbUrl, fullUrl: image.fullUrl, originalWidth: image.originalWidth, originalHeight: image.originalHeight, mediaType: image.mediaType, description: image.description, author: image.author, license: image.license, date: image.date, caption: image.caption, categories: image.categories, reason: image.reason, contentLanguageName: image.contentLanguageName, sectionNumber: image.sectionNumber)
```
tonisevener, line 107: I've been looking at this data across a couple of languages, and it seems like the `sectionNumber` is always null. The call I took from Android already returns the first section's wikitext, under `revisions.slots.main.content`. I think we should just past that through to avoid the extra call and pass in section 0 when needed. Let me know what you think!

Wikipedia/Code/ExploreViewController.swift
```diff
@@ -1126,13 +1126,28 @@ extension ExploreViewController: WKImageRecommendationsDelegate {
...
+
+        var components = URLComponents()
+        components.scheme = "https"
+
+        switch project {
+        case .wikipedia(let language):
+            components.host = "\(language.languageCode).wikipedia.org"
+        default:
+            assertionFailure("Unexpected project for image recommendations")
+        }
```
tonisevener, line 1139: I recently made a WKProject.siteURL convenience method that will save you some lines here.

Wikipedia/Code/InsertMediaSettingsViewController.swift
```diff
@@ -224,10 +235,20 @@ final class InsertMediaSettingsViewController: ViewController {
...
             self.apply(theme: theme)
+
+            loadWikitext { result in
+                switch result {
+                case let .success(wikitext):
+                    self.wikitext = wikitext
+                case let .failure(error):
+                    print("error \(error)")
+                }
+            }
```
tonisevener, line 246: If you do decide to keep this - can you wrap it up in a `fromImageRecommendations` conditional so we aren't fetching it in the editor context?

tonisevener (commented):
Looking good! A couple of suggestions to limit (or remove entirely) the wikitext call.

tonisevener (approved):
Looks good!

PR #4802, opened 2024-03-29
Image recommendations - data improvements

tonisevener:
**Phabricator:** N/A

### Notes
This PR contains some improvements to image recommendations data fetching.

1. The explore feed card already fetched image recommendations, because it needs that data to know whether or not to display. However, when we pushed on image recommendations, it fetched a whole new set. It's wasteful to do it twice, especially since this call seems to take a while to load. Now I am caching the image recommendations within the data controller.
2. At the end of the cycle, I am attempting to fetch more image recommendations. This matches Android.
3. I added underscores to the article summary title before fetching. Without them, the article summary call redirected to another endpoint with underscores.
4. The faster initial call caused onboarding to intercept the half sheet modal. The last commit fixes that.

### Test Steps
1. Fresh install on Staging. Tap Explore feed card.
2. Dismiss onboarding. Confirm image recommendations appear after dismissal.
3. Tap Not Sure until the end. Confirm you see a loading state after exhausting recommendations, and then new recommendations load.

### DIFF
--- a/Components/Sources/Components/Components/Onboarding/WKOnboardingViewController.swift
+++ b/Components/Sources/Components/Components/Onboarding/WKOnboardingViewController.swift
@@ -4,9 +4,16 @@ import SwiftUI
 public protocol WKOnboardingViewDelegate: AnyObject {
     func didClickPrimaryButton()
     func didClickSecondaryButton()
+    func willSwipeToDismiss()
 }
 
 public class WKOnboardingViewController: WKCanvasViewController {
+    
+    public weak var delegate: WKOnboardingViewDelegate? {
+        didSet {
+            hostingController.delegate = delegate
+        }
+    }
     
    // MARK: - Properties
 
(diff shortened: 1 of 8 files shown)


Components/Sources/Components/Components/Onboarding/WKOnboardingViewController.swift
```diff
@@ -4,10 +4,17 @@ import SwiftUI
 public protocol WKOnboardingViewDelegate: AnyObject {
     func didClickPrimaryButton()
     func didClickSecondaryButton()
+    func willSwipeToDismiss()
```
staykids, line 7: One recommendation with these delegate methods would be to prefix them with the delegating type, so in this case something like `onboardingViewWillSwipeToDismiss` or something (akin to how table view or collection view delegate methods might be be named). As we see later on with the `ViewControllerRouter` and `WKImageRecommendationsViewController` conformance, the general `willSwipeToDismiss` or `didClickPrimaryButton` etc. naming makes it confusing as to the function's intent.

staykids (commented):
Nice – one minor note below, but happy to merge as-is if you prefer.

Components/Sources/Components/Components/Onboarding/WKOnboardingViewController.swift
tonisevener: Agreed - Done!

staykids (approved):
👍🏽 thanks for that change! Just a merge conflict to resolve.

PR #4804, opened 2024-04-02
Image recommendations - Go to article preview 

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T358920

### Notes
* This PR adds the ability to go from the InsertMediaSettingViewController to the EditPreviewViewController when coming from the image recommendations feature
* It depends on https://github.com/wikimedia/wikipedia-ios/pull/4800

### Test Steps
1. Run the experimental app
2. Tap the Image Recommendation card
3. Tap "Yes" for a recommendation
4. On the Add Image Details view, add some caption and tap next
5. You should preview the first section of the article with the image and caption
6. Go to the page editor, add an image and preview the edit, make sure nothing breaks or looks off


Components/Sources/Components/Components/Suggested Edits/Image Recommendations/WKImageRecommendationsViewModel.swift
```diff
@@ -247,7 +249,8 @@ public final class WKImageRecommendationsViewModel: ObservableObject {
                     description: metadata.description,
                     descriptionURL: metadata.descriptionUrl,
                     reason: metadata.reason,
-                    wikitext: page.revisions.first?.wikitext
+                    wikitext: page.revisions.first?.wikitext,
+                    sectionNumber: Int(metadata.sectionNumber ?? String())
```
tonisevener, line 253: I still don't really trust the metadata's `sectionNumber`, because it's always `null` when I inspect the API response. I think you should hardcode 0 here, because we are explicitly [fetching section 0 wikitext](https://github.com/wikimedia/wikipedia-ios/blob/main/WKData/Sources/WKData/Data%20Controllers/WKGrowthTasks/WKGrowthTasksDataController.swift#L30) in the image recommendations call.

We want this section number to stay in sync with what wikitext we're using to inject image wikitext into. I suspect that when `sectionNumber` is null, once we connect the posting to `EditSaveViewController`, you will be updating the entire article wikitext with only the intro text. If we use section 0 and pass that along into `EditSaveViewController`'s `sectionID` property, it should post the update to the correct section.

Wikipedia/Code/InsertMediaSettingsViewController.swift
```diff
@@ -270,14 +273,26 @@ final class InsertMediaSettingsViewController: ViewController {
...
+        }
+
+        let editPreviewViewController = EditPreviewViewController(pageURL: articleURL)
+        editPreviewViewController.theme = theme
+        editPreviewViewController.sectionID = sectionNumber
+        editPreviewViewController.languageCode = articleURL.wmf_languageCode
+        editPreviewViewController.wikitext = wikitext
+        editPreviewViewController.delegate = self
+
+        navigationController?.pushViewController(editPreviewViewController, animated: true)
```
tonisevener, line 295: I notice the preview content is hidden beneath the navigation bar. I'm happy to merge now and we fix as a followup, just wanted to call it out. It's probably our custom navigation bar up to its old tricks again.

screenshot23

tonisevener (commented):
Looking good, just a couple of things that jumped out at me. Happy to merge as-is if you want to work on those in later PRs.

Components/Sources/Components/Components/Suggested Edits/Image Recommendations/WKImageRecommendationsViewModel.swift
mazevedofs: Done!

Wikipedia/Code/InsertMediaSettingsViewController.swift
mazevedofs: If you don't mind merging as-is, I'll take a look on it after wrapping up the next task

tonisevener (approved):

PR #4816, opened 2024-04-21
Fix bolding of italicized titles in search

joewalsh:
### Test Steps
1. Search for an article with an italicized title, using a term in that title

Expected: Search term is bold and italic
Actual: Search term is just bold

### Screenshots/Videos
Before:

After:

### DIFF
--- a/WMF Framework/UIFont+WMFDynamicType.swift
+++ b/WMF Framework/UIFont+WMFDynamicType.swift
@@ -143,10 +143,9 @@ public extension UIFont {
     }
     
     func with(traits: UIFontDescriptor.SymbolicTraits) -> UIFont {
-        guard let descriptor = self.fontDescriptor.withSymbolicTraits(traits) else {
+        guard let descriptor = fontDescriptor.withSymbolicTraits(fontDescriptor.symbolicTraits.union(traits)) else {
             return self
         }
-        
         return UIFont(descriptor: descriptor, size: 0)
     }
 }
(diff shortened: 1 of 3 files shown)


Wikipedia/Code/ArticleCollectionViewCell.swift
```diff
@@ -16,11 +16,11 @@ open class ArticleCollectionViewCell: CollectionViewCell, SwipeableCell, BatchEd
...
     
     private func updateTitleLabel() {
-        if let titleHTML = _titleHTML {
-            let attributedTitle = titleHTML.byAttributingHTML(with: titleTextStyle, matching: traitCollection)
+        if var titleHTML = _titleHTML {
             if let boldString = _titleBoldedString {
-                attributedTitle.applyBoldFont(to: boldString, textStyle: titleTextStyle, matching: traitCollection)
+                titleHTML.applyBoldTag(to: boldString)
             }
+            let attributedTitle = titleHTML.byAttributingHTML(with: titleTextStyle, matching: traitCollection)
```
joewalsh, line 23: This would previously override the font applied by `byAttributingHTML`. Now, it adds a bold tag so that `byAttributingHTML` will apply the bold (or bold + italic) font

WMF Framework/UIFont+WMFDynamicType.swift
```diff
@@ -143,10 +143,9 @@ public extension UIFont {
     }
     
     func with(traits: UIFontDescriptor.SymbolicTraits) -> UIFont {
-        guard let descriptor = self.fontDescriptor.withSymbolicTraits(traits) else {
+        guard let descriptor = fontDescriptor.withSymbolicTraits(fontDescriptor.symbolicTraits.union(traits)) else {
             return self
         }
-        
         return UIFont(descriptor: descriptor, size: 0)
```
joewalsh, line 149: This `with(traits:)` call is used to apply the italic trait to the bold font, but it wasn't working. The issue was not adding the new `traits` to the existing traits, previously only the new traits were applied instead of new + old

joewalsh:
Hm, closing for now. Need to make sure the bold tag isn't applied to the content of other tags and will add tests 😅

PR #4817, opened 2024-04-22
Fix unsubscribe from push notifications bug

spatel:
**Phabricator:** 
https://phabricator.wikimedia.org/T361221

### Notes

When unsubscribing from push notifications, the code eventually reaches [`EchoSubscriptionFetcher.unsubscribe(siteURL:deviceToken:completion)`](https://github.com/wikimedia/wikipedia-ios/blob/88622cb57f7110276e4234c32ac2bb5673b5cc84/Wikipedia/Code/EchoSubscriptionFetcher.swift#L51), however this throws an error because the required device token is always `nil`. 

Here's why:

   * In the app delegate, `[[UIApplication sharedApplication] registerForRemoteNotifications]` is called at app startup ✅
   * This triggers the following sequence of events:
```
    ✅ -(void)application:(UIApplication *)application didRegisterForRemoteNotificationsWithDeviceToken:(NSData *)deviceToken
    ✅   -> [self.appViewController setRemoteNotificationRegistrationStatusWithDeviceToken:deviceToken error:nil];
    ❌      -> [self.notificationsController setRemoteNotificationRegistrationStatusWithDeviceToken:error:error] 
```
   * `self.notificationsController` is actually always `nil` at this point (i.e. hence the token never gets saved). This is because `self.notificationsController` is only initialised towards the very end of [`[MWKDataStore finishSetup]`](https://github.com/wikimedia/wikipedia-ios/blob/88622cb57f7110276e4234c32ac2bb5673b5cc84/Wikipedia/Code/MWKDataStore.m#L118) (i.e. the final step in the `[WMFAppViewController launchAppInWindow:waitToResumeApp:]` flow).

**Note: there are multiple ways to fix this. At this stage, I've just moved `[UIApplication registerForRemoteNotifications]` so that it happens AFTER `WMFAppViewController` is launched (with all its dependencies fully initialised). But I'm really just highlighting one possible solution - it's completely up to you if you have a preferred / better way of addressing this 😀**

> FYI - in case you're curious why this only affects turning OFF push notifications, it's because when you turn them ON, you're explicitly calling `[[UIApplication sharedApplication] registerForRemoteNotifications]` on subscription, not just at startup.

### Test Steps
1. While logged in, go to Settings > Push Notifications > Toggle "ON"
2. Kill / restart the app
3. Go back to Push Notifications > Toggle "OFF"


tonisevener (approved):
@spatel Nice investigative work! It's working well for me after your change.

I do think there's a chance that the bug will remain in some instances, due to the Core Data stack loading [asynchronously](https://github.com/wikimedia/wikipedia-ios/blob/main/Wikipedia/Code/MWKDataStore.m#L213) during `finishSetup`. It's possible `notificationsController` may still be nil if that loading takes a while.

I'm willing to wait and see if it's an actual issue before adjusting though, since this change so far feels low risk. Thanks for fixing!

tonisevener (approved):

PR #4819, opened 2024-04-24
Clear Xcode warnings

mazevedofs:
**Phabricator:** N/A

### Notes
This PR:
* Removes some unused props that were causing warnings, probably after merge conflicts
* Adds a warning suppression for the swiftlint rule `unused_capture_list` that was causing false warnings in some cases where `weak self` is captured but not referenced directly after unwrapping it with `guard let`.  The syntax is correct, but the rule is apparently outdated.

Let me know if you disagree with this approach.

### Test Steps
1. Build the app, confirm the warnings are gone


Wikipedia/Code/EditSaveViewController.swift
```diff
@@ -489,9 +489,9 @@ class EditSaveViewController: WMFScrollViewController, Themeable, UITextFieldDel
                   let currentTitle = pageURL?.wmf_title else {
                 return
             }
-            
+            // swiftlint:disable unused_capture_list
```
tonisevener, line 492: Hmm I'm not seeing this warning - can you try updating swiftlint? The latest I'm able to update to is `0.53.0`.

mazevedofs: > Hmm I'm not seeing this warning - can you try updating swiftlint? The latest I'm able to update to is `0.53.0`.

I just updated, and they went away! I'll close this PR as the other ones seem to have been fixed as well

PR #4822, opened 2024-04-25
Image Recommendations - Fix assert failure for ZH Wiki

tonisevener:
**Phabricator:** N/A

### Notes
I noticed this assert failure when testing something unrelated in ZH Wiki.

### Test Steps
1. Fresh install experimental app.
2. Set primary language to a ZH Wiki variant in onboarding
3. Wait until the app logs the user in and tries to insert the suggested edits card.
4. Confirm card is successfully inserted without triggering assert, and you can tap on it to get into the feature.


mazevedofs (approved):
🎉

PR #4825, opened 2024-04-26
Image Recommendations - Fix explore card toggle bug

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T363601

### Notes
The Explore settings persist the global card preference ON toggles upon fresh install. This means those preferences were originally saved without the suggested edits card type. When updating the app with the latest version, the Explore feed only considers those older types as on, so suggested edits defaults to off.

To fix this I am adding a one-time migration step in MWKDataStore to turn the suggested edits card preference on.

### Test Steps
1. Through TestFlight, install version 7.4.9 (3332).
2. Launch app and ensure you are logged in.
3. Run the app in Xcode from this branch.
4. Confirm the suggested edits card eventually appears in the Explore feed, and tapping into Explore settings shows the suggested edits card as ON.


mazevedofs (approved):
:)

PR #4828, opened 2024-04-29
Image Recommendations - Error Handling

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T358928

### Notes
This PR adds some error handling to the image recommendations feature. In summary, it delegates out any loading error to `ExploreViewController`, which then displays it using `WMFAlertController`. It also changes out the content view of image recommendations to a new error state component.

`EditSaveViewController` (Edit Summary view) already had a lot of working error handling from being in the editor flow, so I left that as-is.

### Test Steps
The only error state that is easy to test is a network connection. Other things you can try is to temporarily mess up the url creation methods in the data controllers, like the [article summary endpoint](https://github.com/wikimedia/wikipedia-ios/blob/main/WKData/Sources/WKData/Data%20Controllers/Shared/WKArticleSummaryDataController.swift#L18). The server should respond with some sort of error if you're missing the `summary` in the url path, for example.

1. Launch app while having a connection. Wait for suggested edits card to appear.
2. Before tapping card, enter airplane mode to ensure your connection is dropped.
3. Tap card.
4. Confirm you see "No internet connection" error banner and new error state component.
5. Leave airplane mode.
6. Once connection returns, tap "Try again".
7. Confirm image recommendations appear.


mazevedofs (approved):
Looks good!

PR #4836, opened 2024-05-02
Picture Of The day bug fixes

spatel:
**Phabricator:** 
https://phabricator.wikimedia.org/T356255

---

### Test Steps

1. Fresh install of the app (primary language English)
2. Background the app > add POTD widget to home screen
3. Tap the widget (it deeplinks to the POTD gallery view) ✅
4. Settings > My Languages > Add language (e.g. Spanish)
5. Change the primary language (e.g. to Spanish)
6. Close settings > refresh the Explore tab (pull-to-refresh)
7. Background the app > tap POTD widget
8. The app launches but the deeplink does nothing (it actually only works against the original primary language i.e. English) ❌

**Expected behaviour**: the app should launch and open the POTD image gallery view

---

### Notes

I split this into 2 commits (one of which is just clang-format changes), so you only really need to review [this commit](https://github.com/wikimedia/wikipedia-ios/commit/200efbf9f1bc6af7fb56ffcdae3d4e1ecf8d8824)

#### Possible root cause

I think this could be due to the fact the POTD image only gets saved once in the data store against the app's initial site URL (i.e. initial primary language). After that, the same record gets continually modified. This is deliberate (to avoid duplicate images from appearing across each language feed) and there was a comment in the code to explain why this was originally done. However, I think you may be running into issues with content URL deep-links from widgets etc because they appear to be querying data against site URL (even for POTD).

#### Worked example

This is based on the test steps mentioned above:

1. The "English" feed is initially loaded and it inserts a new POTD content record in the data store (against the "EN" site URL)
2. The user later switches primary language to "Spanish" and launches the app from a widget deeplink like this: `wikipedia://content/picture-of-the-day/wikipedia.org/es/2024/05/01`
3. The app tries processes the deeplink (and reaches [`[WMFAppViewController processUserActivity:animated:completion]`](https://github.com/wikimedia/wikipedia-ios/blob/6182dd8cc2717655dceaed4b5650f5afae5e2124/Wikipedia/Code/WMFAppViewController.m#L1185))
4. It [queries the data store for a POTD (_against the **ES** site URL_)](https://github.com/wikimedia/wikipedia-ios/blob/6182dd8cc2717655dceaed4b5650f5afae5e2124/Wikipedia/Code/WMFAppViewController.m#L1191) but finds no match (because it was originally stored against "EN")
5. It then [falls through to here](https://github.com/wikimedia/wikipedia-ios/blob/6182dd8cc2717655dceaed4b5650f5afae5e2124/Wikipedia/Code/WMFAppViewController.m#L1206) where it eventually tries to [refresh all the feeds and save the POTD again](https://github.com/wikimedia/wikipedia-ios/blob/6182dd8cc2717655dceaed4b5650f5afae5e2124/Wikipedia/Code/WMFFeedContentSource.m#L233)
6. At this point it [**DOESN'T** query against the site URL](https://github.com/wikimedia/wikipedia-ios/blob/6182dd8cc2717655dceaed4b5650f5afae5e2124/Wikipedia/Code/WMFFeedContentSource.m#L332), so it finds the existing "EN" record and updates it.
7. Once everything is refreshed, a completion block runs & queries for the POTD (i.e. similar to step 4, but it now expects the **ES** record to exist) - but since it doesn't exist, nothing happens

### Other observations

* At point 6 above, this is interesting. Because ALL language feeds end up overwriting the same record, the image description is overwritten (so whichever language is saved last "wins"). As a result you can end up with the language text randomly changing language (see attached video).
* Although this might fix some quirks with POTD, it's unlikely to fix all of them. After switching languages, I still found it occasionally necessary to do a manual refresh of the Explore tab before the POTD deeplinks fix themselves. I'm not sure why this was (but hopefully someone with a better knowledge of the code base may know!).
* I also noticed that if the user is offline when they originally install the widget, it's possible for the placeholder image to get cached on their home screen (which can affect deeplinking to the POTD image gallery). This is something that may need fixing separately (for example, you might want to pass something like a `?forceReloadWidget=<widget.id>` query parameter with the widget URL) - but this seems to be quite "edge-casey" anyway.

---

### Screenshots/Videos

As described above, here is the Explore feed's POTD image description (see bottom part of the image) flipping languages:

https://github.com/wikimedia/wikipedia-ios/assets/478681/5dc97b15-637a-4109-acb2-1eb12270e500


Wikipedia/Code/WMFFeedContentSource.m
```diff
@@ -232,10 +232,28 @@ - (void)saveGroupForPictureOfTheDay:(WMFFeedImage *)image date:(NSDate *)date in
...
+            group = [moc createGroupOfKind:WMFContentGroupKindPictureOfTheDay forDate:date withSiteURL:self.siteURL associatedContent:nil];
+        } else {
+            DDLogDebug(@"Updating POTD content group for primary feed language [%@]", feedLanguage);
+        }
+        group.contentPreview = image;
+    } else {
+        if (group != nil) {
+            DDLogDebug(@"Removing POTD content group for OLD secondary feed language [%@].", feedLanguage);
+            [moc removeContentGroup:group];
+        }
```
spatel, line 255: This is the main logic change (1 of 2). 

Here's what I've done:

- The original logic is still the same (it's just moved inside the `if (isPrimaryLanguageFeed) { }` statement)
- The `else` part is new and prevents duplicates by removing any POTD records found against non-primary languages.
- The logging statements are new and were really just to help me troubleshoot (please let me know if you'd like me to remove them).

Wikipedia/Code/WMFFeedContentSource.m
```diff
@@ -328,8 +346,7 @@ - (nullable WMFContentGroup *)featuredForDate:(NSDate *)date inManagedObjectCont
 }
 
 - (nullable WMFContentGroup *)pictureOfTheDayForDate:(NSDate *)date inManagedObjectContext:(NSManagedObjectContext *)moc {
-    //NOTE: POTDs are the same across languages so we do not not want to constrain our search by site URL as this will cause duplicates
-    return (id)[moc groupOfKind:WMFContentGroupKindPictureOfTheDay forDate:date];
+    return (id)[moc groupOfKind:WMFContentGroupKindPictureOfTheDay forDate:date siteURL:self.siteURL];
```
spatel, line 349: This is the other main logic change (part 2 of 2). We use the site URL (in the same way as other content) and the duplicate logic is now handled as part of the `save` method.

_N.B. All other changes to this file are just clang-format changes_

tonisevener (changes requested):
@spatel Sorry for the delay on this review. Excellent investigative work, as always. 🙂

I found a couple of bugs, but overall I'm good with your approach in this method. Let us know if these fixes give you trouble and we can discuss alternatives. Thanks!

Wikipedia/Code/WMFFeedContentSource.m
```diff
@@ -229,15 +229,49 @@ - (void)saveGroupForPictureOfTheDay:(WMFFeedImage *)image date:(NSDate *)date in
...
+    // T356255 - the POTD is the same across all feeds (i.e. same image, different localized description).
+    // To prevent the same image from showing on each language feed, we save the POTD for the primary language only
+    // (and clean up any secondary / duplicate ones - secondary ones may exist when the user has removed a language
+    // or changed the app primary language).
+    
+    if (isPrimaryLanguageFeed) {
+        [self insertOrUpdatePictureOfTheDayForImage:image date:date inManagedObjectContext:moc];
+        [self removeDuplicatePicturesOfTheDayForDate:date inManagedObjectContext:moc];
+    }
+}
```
spatel, line 246: **Change 2 of 3**

This now inserts (or updates) POTD against the primary language only (and removes any duplicate ones found against secondary languages).

PR #4841, opened 2024-05-03
Image recommendations -  hide settings toggle

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T364038

### Notes
* Checks edit count before adding Suggested Edit to the explore feed settings toggles

### Test Steps
1. On an account with less than 50 edits (or logged out), go to explore feed settings
2. Should not see the Suggested Edits toggle
3. Repeat on an account with more than 50 edits, the toggle should appear


tonisevener (approved):

Wikipedia/Code/ExploreFeedSettingsViewController.swift
```diff
@@ -189,6 +189,22 @@ class ExploreFeedSettingsViewController: BaseExploreFeedSettingsViewController {
         dismiss(animated: true)
     }
 
+    var editCount: Int {
+        var count: Int = 0
+        if let language = self.dataStore?.languageLinkController.appLanguage?.siteURL {
+            self.dataStore?.authenticationManager.getLoggedInUser(for: language, completion: { result in
```
tonisevener, line 195: Just note that this is an async call, so if it needs to make a network request this count would return 0 while it's fetching. I think because we have good caching within `getLoggedInUser`, it rarely needs to make a call here. Very rarely I could reproduce some weirdness where if I slowed down the connection speed, launched the app, then went straight to the Explore feed settings the toggle would be missing. But returning a little bit later (once the user was fetched) allowed it to appear again.

Not a blocker IMO but something to remember if we hear complaints.

mazevedofs: Good point, should we create a ticket to improve this? Since we're doing some tech debt work soon?

tonisevener: Sounds good! Created a task here - https://phabricator.wikimedia.org/T364133

PR #4847, opened 2024-05-07
Add campaign ID to button taps on campaign modal

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T352722

### Notes
* This PR adds the `campaign_id` in the `action_data` field for campaign button events.

### Test Steps
1. Change device region to Japan and device date to May 29th.
2. Launch app on Staging. Background app on Explore feed. Foreground, then pull to refresh to fetch campaigns.
3. Visit an article. Tap the Close button on campaign modal and confirm you see `campaign_id` listed out in the console for button tap event.
4. Repeat steps 2-3 for other campaign modal buttons (donate, maybe later, I already donated), confirm `campaign_id` is now sent with event.


mazevedofs (approved):
Working great! I tried to fix the conflicts but I was messing something up, so just need to fix that and it can be merged

PR #4848, opened 2024-05-07
Use particular campaign ID format for fundraising metrics

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T352717

### Notes
This PR modifies the `campaign_id` format when sending fundraising metrics. Note that we do still need to retain the old format and send to the `submitPayment` endpoint. I kept that logic as-is (generally called `bannerID` and is sourced from the config asset's `utm_source` value), but I updated all the metrics call sites to accept a new `metricsID`. 

### Test Steps (Optional)
1. Change device region to Japan, device date to May 29th. Add card to Apple Wallet.
2. Fresh install on Staging. Background, foreground, pull to refresh to fetch campaign configs.
3. Tap to article to trigger campaign modal.
4. Add breakpoint [here](https://github.com/wikimedia/wikipedia-ios/blob/7f117a3cfc807d213d0f3e1966976ee614df886f/Wikipedia/Code/AppInteractionFunnel.swift#L101), [here](https://github.com/wikimedia/wikipedia-ios/blob/7f117a3cfc807d213d0f3e1966976ee614df886f/Wikipedia/Code/AppInteractionFunnel.swift#L208) and [here](https://github.com/wikimedia/wikipedia-ios/blob/7f117a3cfc807d213d0f3e1966976ee614df886f/WKData/Sources/WKData/Data%20Controllers/Donor%20Experience/WKDonateDataController.swift#L177).
5. Tap Donate button. Confirm first breakpoint is hit and campaign ID follows format `enJP_2024_05_iOS`.
6. Tap Apple Pay method. Submit payment. Confirm second breakpoint is hit and campaign ID follows format `enJP_2024_05_iOS`.
7. Continue debugging. Confirm third breakpoint is hit and banner ID follows format `app_2024_JP_iOS_control`. Stop Xcode now to avoid charges.


mazevedofs (approved):
Looking good!

PR #4868, opened 2024-06-24
Add time warning to image recs

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T364049

### Notes
* This PR adds a timed warning to the "Yes" button

### Test Steps
1. On an image recommendation, tap the yes button quickly as the image appears
2. Verify the warning is shown
3. Try again. The warning should not appear as probably 5 seconds have already passed
4. Verify the event is being logged correctly


Components/Sources/Components/Components/Suggested Edits/Image Recommendations/WKImageRecommendationsBottomSheetViewController.swift
```diff
@@ -33,10 +33,12 @@ final public class WKImageRecommendationsBottomSheetViewController: WKCanvasView
             self.bottomSheetView = bottomSheetView
         }
     }
-    
+
+    private var startTime: Date?
+
     public override func viewDidAppear(_ animated: Bool) {
```
tonisevener, line 39: I think we might need some more logic to align with Android better. If I tap View Article and spend plenty of time reviewing, then go back, this timer resets and the toast appears again, which is rather annoying. Android doesn't seem to have this issue.

Maybe better would be to assign a start time property in `WKImageRecommendationsViewModel` whenever its `currentRecommendation` value is changed. That way it's more tied to whenever the recommendation is changed on screen rather the bottom sheet appearance. Then in `didTapYesButton` here you can reference `if let startTime = viewModel.startTime` in the first conditional.

tonisevener (changes requested):

Components/Sources/Components/Components/Suggested Edits/Image Recommendations/WKImageRecommendationsBottomSheetViewController.swift
mazevedofs: Good point; I'm starting the counter after populating the article and image info, let me know what you think.

tonisevener (approved):
Looks good!

PR #4876, opened 2024-07-08
Remove Root Navigation Controller

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T369567

### Notes
This PR does a few things:

- Previously our tab bar controller was the root of a single navigation controller, and that navigation controller was the root of the window. This PR changes it so that the tab bar controller is the root of the window, with 5 navigation controllers embedded within each tab. This feels a bit more standard to what iOS documentation suggests and could help split up any future coordinators that we build.
- Previously the splash screen view controller was embedded within our single navigation controller, and we triggered it from the outside. The splash screen logic is now removed from the navigation controller and is presented from the tab bar controller during launch, similar to onboarding.
- In testing the article state restoration, the code felt too complicated for what it was doing. Previously, upon save, it walked the entire stack and persisted structs representing the entire view controller hierarchy, but upon restore it _only_ looked for the topmost article and pushed it on. Now we are simplifying the save portion and are only preserving the top most article view controller.

### Test Steps
1. Do some general regression testing on navigation. Confirm things still act as they did before.
2. Confirm onboarding still appears as it did before.
3. Test migrating splash screen animation. Change current library version to 19 [here](https://github.com/wikimedia/wikipedia-ios/blob/88b5d4afff28c53b7e4d74a3db2bf12faa4a213b/Wikipedia/Code/MWKDataStore.m#L17), then add this code [here](https://github.com/wikimedia/wikipedia-ios/blob/88b5d4afff28c53b7e4d74a3db2bf12faa4a213b/Wikipedia/Code/MWKDataStore.m#L514):

```
    if (currentLibraryVersion < 19) {
        sleep(10);
        [moc wmf_setValue:@(19) forKey:WMFLibraryVersionKey];
        if ([moc hasChanges] && ![moc save:&migrationError]) {
            DDLogError(@"Error saving during migration: %@", migrationError);
            return;
        }
    }
```

Run the app and confirm you see the "Updating" animation. Once the 10 seconds are up the app should show you the Explore feed.
4. Confirm deep linking to an article still works.
5. Test state restoration. Visit an article. Background the app to trigger navigation state save. Terminate the app. Relaunch the app and confirm article is pushed on.


mazevedofs (approved):
Looking great!

PR #4877, opened 2024-07-10
Rewrite AppDelegate in Swift, use SceneDelegate

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T369725

### Notes
This PR rewrites our AppDelegate into Swift, and adds a SceneDelegate. Note that there will still be quite a few issues to fix before we can fully support multiple windows on iPad, due to assumptions our codebase makes. By setting a "Supports multiple windows" to false in Info.plist, the app is able to act as before while still using the new APIs.

This PR is dependent on https://github.com/wikimedia/wikipedia-ios/pull/4876.

### Test Steps
1. Confirm app launches as it did before. Test onboarding as well.
2. Confirm deep linking to an article works as it did before.
3. Confirm you _cannot_ create multiple windows on iPad, as it did before.


mazevedofs (approved):
Looking great! So good to see all the old objc go 👋

PR #4881, opened 2024-07-10
Add JS confirm support for Account Vanishing

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T369628

### Notes
The new account vanishing form will have a confirmation popup before submitting. We need to add extra handling for this to work.

### Test Steps
1. For a test account, go to the new Vanishing form. You should see a native confirmation modal when you submit the form. Tapping OK submits the form, tapping cancel backs out.


mazevedofs (approved):
Looks good!

PR #4882, opened 2024-07-11
Add developer settings options

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T334848

### Notes
This PR moves our old FeatureFlags struct into persisted User Defaults flags through WKData. It also adds a Developer Settings configuration modal for adjusting these.

As a part of these I cleaned out older Feature Flags that I felt we no longer need. The only one left is for suppressing image recommendation edit post calls, which should still be handy for us to avoid production edits when testing the alt text feature.

### Test Steps
1. Go to Settings > About the App > Tab the navigation bar title's "About" title ~7 times.
2. A modal should appear for adjusting developer settings. The one toggle is for suppressing image recommendation edit posts. Confirm it is off initially. Toggle it on.
3. Terminate and relaunch the app. Go back to developer settings view. Confirm toggle is still on.
4. Post an edit through image recommendations. Check on Desktop Wikipedia article history and confirm edit was NOT posted.
5. Go back to developer settings and toggle off the flag.
6. Post an edit through image recommendations. Check on Desktop Wikipedia article history and confirm edit was posted.


mazevedofs (approved):
Looking great!

PR #4883, opened 2024-07-12
Fix warnings and logs

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T369871

### Notes
These are some tweaks to improve our logging and fix easy project warnings.

### Test Steps
1. Observe user buttons on diff view, confirm no regressions occurred.
2. Confirm there are fewer Cocoalumberjack logs in the console.
3. Temporarily change debug level to all [here](https://github.com/wikimedia/wikipedia-ios/blob/main/Wikipedia/Code/WMFLogging.h#L6).
4. Confirm Cocoalumberjack debug console logs return.

### DIFF
--- a/Components/Sources/Components/Components/Shared/Buttons/WKButton.swift
+++ b/Components/Sources/Components/Components/Shared/Buttons/WKButton.swift
@@ -1,3 +1,0 @@
-import UIKit
-
-public class WKButton: UIButton {}
--- a/Components/Sources/Components/Components/Shared/Buttons/WKSmallMenuButton.swift
+++ b/Components/Sources/Components/Components/Shared/Buttons/WKSmallMenuButton.swift
@@ -52,10 +52,14 @@ public class WKSmallMenuButton: WKComponentView {
 
     // MARK: - UI Elements
 
-    private lazy var button: WKButton = {
-        let button = WKButton(type: .custom)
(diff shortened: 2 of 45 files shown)


Wikipedia/Code/WMFBarButtonItemPopoverMessageViewController.h
```diff
@@ -1,10 +1,13 @@
 @import UIKit;
 @import WMF.Swift;
 
-@interface WMFBarButtonItemPopoverMessageViewController : UIViewController <UIPopoverPresentationControllerDelegate, WMFThemeable>
+@interface WMFBarButtonItemPopoverMessageViewController : UIViewController <UIPopoverPresentationControllerDelegate>
```
mazevedofs, line 5: I'm getting a crash with this message `Thread 1: Could not find a storyboard named WMFBarButtonItemPopoverMessageViewController' in bundle NSBundle` maybe when the popover tries to instantiate.
I did a fresh install of the app, logged out (on simulator)

tonisevener: @mazevedofs Good catch 😅 I think I have fixed this, let me know if it's working for you.

mazevedofs (approved):
Looks good!

PR #4885, opened 2024-07-16
T365306

Mashtaleralia:
**Phabricator:** 

### Notes
* 

### Test Steps
1.


tonisevener (changes requested):
Hi @Mashtaleralia, thanks for working on this one!

I notice the transaction fee label doesn't update correctly. Here are some tips for fixing.

Even though you are updating the `transactionFeeAmount` in the `finalAmount` setter, the `transactionFeeOptInViewModel` (which powers the label text) is not updated. I would try two things:

1. [Here](https://github.com/Mashtaleralia/wikipedia-ios/blob/T365306/Wikipedia/Code/UIViewController%2BDonateHelpers.swift#L62-L63) on the outside we are determining the final transaction fee text and passing it into the localized strings struct further down in this method. The assumption at the time was that once this fee text was passed in, it would not need to change. For here I would recommend ONLY passing in the format string. We will then resolve the final text later within the `WMFDonateViewModel` as the `transactionFeeAmount` changes.

2. Adding something like this code [here](https://github.com/Mashtaleralia/wikipedia-ios/blob/T365306/WMFComponents/Sources/WMFComponents/Components/Donate%20Form/WMFDonateViewModel.swift#L189) in `didSet` will update the label. We are assigning a new view model which should trigger SwiftUI to redraw that area of the form.

```
        let formatter = NumberFormatter.wmfCurrencyFormatter
        formatter.currencyCode = currencyCode
        let transactionFeeString = formatter.string(from: transactionFeeAmount as NSNumber) ?? ""
        let text = String.localizedStringWithFormat(localizedStrings.transactionFeeOptInTextFormat, transactionFeeString)
        let isOldModelSelected = transactionFeeOptInViewModel.isSelected
        transactionFeeOptInViewModel = OptInViewModel(localizedStrings: OptInViewModel.LocalizedStrings(text: text, accessibilityHint: localizedStrings.accessibilityTransactionFeeHint), isSelected: isOldModelSelected)
        addTransactionFeeSelectionListener()
```

You'll also need to update the [init](https://github.com/Mashtaleralia/wikipedia-ios/blob/T365306/WMFComponents/Sources/WMFComponents/Components/Donate%20Form/WMFDonateViewModel.swift#L122) for `OptInViewModel` to allow an `isSelected` value to be passed in.

tonisevener:
@Mashtaleralia We plan on releasing this piece soon, so I have built off of this work and moved it to https://github.com/wikimedia/wikipedia-ios/pull/4964. Thank you for your contribution!

PR #4890, opened 2024-07-23
[ALT TEXT]  Bypass edit count for target wikis on suggested edits explore feed card

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T370224

### Notes
* This PR updates the explore feed card and settings logic to bypass the edit count for target wikis on the alt text experiment

### Test Steps
1. Use an account with less than 50 edits
2. With the Developer settings toggle on, add one of the target wikis as the primary app language (pt, es, fr, or zh)
3. Verify you can see the explore feed card and setting toggle
4. Turn off the developer settings toggle, verify both are not shown
5. WIth the toggle on again, add another primary language that is not target 
6. Verify you cannot see the explore feed card and explore settings


tonisevener (commented):
@mazevedofs Looking good, a couple of things that came to mind during code review:

1. I know it's not mentioned in the engineering notes but maybe we should bypass the new logic if the end date has passed? (October 21st)
2. Should we still have a requirement that the user is logged in before showing the card / settings option on the target wikis? I haven't tested but since we aren't checking user or editCount anymore, there may be a side-effect of the card now appearing for anonymous users. 
3. Just FYI I noticed you aren't checking iPad and iOS 15, but I think that might be correct in this case and my original thinking when adding that in the engineering notes was wrong. Feel free to remove that in the task description if you agree.

mazevedofs:
> @mazevedofs Looking good, a couple of things that came to mind during code review:
> 
> 1. I know it's not mentioned in the engineering notes but maybe we should bypass the new logic if the end date has passed? (October 21st)
> 2. Should we still have a requirement that the user is logged in before showing the card / settings option on the target wikis? I haven't tested but since we aren't checking user or editCount anymore, there may be a side-effect of the card now appearing for anonymous users.
> 3. Just FYI I noticed you aren't checking iPad and iOS 15, but I think that might be correct in this case and my original thinking when adding that in the engineering notes was wrong. Feel free to remove that in the task description if you agree.

Yeah I forgot about those checks 🤦‍♀️ ! Just added, let me know what you think!

tonisevener (approved):
Looks good!

PR #4891, opened 2024-07-24
[ALT TEXT] Assign experiments after edit publish

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T370228

### Notes
This PR adds a couple of data controllers to WKData to allow us to assign experiments for alt text. Our old ABTestsController was previously set up to store within our old metrics database, so it wasn't immediately usable. I copied its logic over to into WKData.WKExperimentsDataController, leaning on our shared cache store for storage. I also made an WKAltTextDataController to handle assignment and additional business logic.

### Test Steps
1. First, temporarily change the WMFLogging.h logging level to DDLogLevelAll to see some additional informational logs for confirming assignments.
2. You may also need to relax the 50+ edit count, because this PR is not built against https://github.com/wikimedia/wikipedia-ios/pull/4890.
3. Fresh install, set app languages to ES Wikipedia (primary language) and Test Wikipedia.
4. Go to Developer Settings and turn on both toggles (suppress image rec edit posting, enable alt text feature).

#### Image Recommendations First Test Steps:
5. Go through image recommendations flow and publish an edit. After publishing, you should see a printout of the experiment assignments. Image recommendations group should be `A` or `B`, article editor group should be `nil`. You should also see a "TODO: Present modal" printout IF the assignment was `B`.
6. Repeat step 5. Your experiment group printouts should be the same. You should also NOT see a "TODO: Present modal" printout, even if your assignment is `B`, because you already saw the modal once.
7. Publish an edit on Test Wiki. Your experiment group will print out. If your image recommendations group in step 5 was `B`, then article editor group should remain `nil`. If your image recommendations group in step 5 was `A`, then your article editor group should be assigned `C` or `D`. You should also see a "TODO: Present modal" printout IF the assignment was `C`.
8. Repeat step 7. Your experiment group printouts should be the same. You should also NOT see a "TODO: Present modal" printout, even if your assignment is `C`, because you already saw the modal once.

#### Article Editor First Test Steps:
9. Fresh install the app. Repeat steps 3-4.
10. Go to an article on Test Wiki and publish an edit. After publishing, you should see a printout of the experiment assignments. Image recommendations group should be `nil`, article editor group should be `C` or `D`. You should also see a "TODO: Present modal" printout IF the assignment was `C`.
11. Repeat step 10. Your experiment group printouts should be the same. You should also NOT see a "TODO: Present modal" printout, even if your assignment is `C`, because you already saw the modal once.
[...]

PR #4893, opened 2024-07-25
[ALT TEXT] Alt text trigger panel

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T370229

### Notes
* Some of the verification logic in this PR depends on https://github.com/wikimedia/wikipedia-ios/pull/4891, so it'll need to be updated after it is merged.
* But some of the other logic can be verified and reviewed before that if necessary
* This PR adds some verifications to trigger the Alt Text task panel
* A new panel was added according to the designs

### Test Steps
1. With the dev settings toggle on, go into image recommendations on a target language
2. Add a recommendation, add alt text
3. You should not see the panel
4. Add another recommendation, **do not add** alt text
6. After publishing, you should see the panel
7. Note that the panel has no actions, and the bottom sheet is not shown underneath (the bottom sheet can't push/present). Bringing it back on after hitting "do not add" should be done with the survey task
8. Go back to the home page, enter image recs again
9. Add a recommendation without alt text
10. You should not see the panel


tonisevener:
It's looking good to me! I noticed some console complaints about presenting over already presenting view controllers, so I made some tweaks that hopefully fixed those.

tonisevener (approved):
Looks good!

PR #4896, opened 2024-07-25
Image Recs - Allow any license for pageimages

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T371055

### Notes
This parameter might allow us to strip out more articles from the image recommendations pool that already have an image.

### Test Steps
1. Add a breakpoint [here](https://github.com/wikimedia/wikipedia-ios/blob/main/WKData/Sources/WKData/Data%20Controllers/WKGrowthTasks/WKGrowthTasksDataController.swift#L84).
2. Run the app and go to the image recommendations flow. Confirm that we still are able to see some recommendations. If breakpoint is hit, that may be a sign that we are stripping out more invalid recommendations.


mazevedofs (approved):
Looking good!

PR #4897, opened 2024-07-25
Fix crash after publishing last suggestion

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T371127

### Notes
This is a crash I noticed when testing https://github.com/wikimedia/wikipedia-ios/pull/4893. I was able to reproduce it on `main`.

### Test Steps
1. Fresh install app, go to developer settings and enable both toggles.
2. You might need to temporarily remove the 50+ edit count to see the Explore feed card.
3. Go to image recommendations. The key here is to tap "Not sure" until you are on the last image recommendation for that group. On the last recommendation, tap Yes and publish an edit.
4. At this point, `main` crashes. This PR should not crash and load the next batch of image recommendations.


mazevedofs (approved):
Looks good!

PR #4901, opened 2024-07-29
[ALT TEXT] Display Article (Part 1)

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T370231

### Notes
This PR pushes on an ArticleViewController after tapping the primary button on the alt text experiment modal from https://github.com/wikimedia/wikipedia-ios/pull/4893.

After testing, I realized we will need to publish an actual edit before we can test scrolling to the correct image on the article view. To make testing easier, I am adding an additional developer settings toggle to show the alt text experiment for EN users, per product's request. Note that _both_ the overall "enable alt text experiment" toggle and the more specific "enable alt text experiment for EN" developer setting will need to be on in order to test the alt text flow on EN Wiki.

Note: This is just Part 1 of this task. Updating the navigation bar and scrolling to the correct image will come in a followup PR.

### Test Steps
1. Fresh install the app
2. In onboarding, select EN Wikipedia
3. Go to developer settings, turn on all 3 toggles.
4. Go through image recommendations flow, publish an edit without alt text.
5. (If you were in the lucky 50%) You will see a modal here about alt text. Tap the primary button.
6. You should see the article view controller pushed on.


mazevedofs (approved):
Looking good!

PR #4903, opened 2024-07-30
Provisional JS shared lib wrapping alt text logic

bvibber:
Missing alt text links detection for experiment C (finding files on the page that are used as images without alt text after you've done an edit), implemented with regexes in common JS code with an eye to experimenting with the ability to share this sort of common logic with the Android side later.

The JS library files are included in the output bundle like the assets for the webkit, but loaded directly into JavaScriptCore for calling.

A Swift wrapper class, AltText, wraps the alt-text.js library and converts the returned JSON structure to a Swift array of structs listing any image links that didn't have alt text.

A CLI node test is included in the sharedlib bundle

A Swift unit test is included on the iOS side to confirm the wrapper function works as expected.

Possible problems:

* JSC doesn't JIT-compile except in the special WKWebView space, so performance might be poorer that expected. If it turns out to be a problem, we can consider wrapping in a WKWebView or just rewriting the code to Swift with common unit tests.

* Maintaining a second copy might be difficult. No attempt has yet been made to make a separate repo, as we may scrap this.

* there might yet be bugs in the code

* Still need to port the internationalization support for the alt= and File: keywords in the logic! The English hardcoded names are used at the moment. This can be a hardcoded list for now of the wider set of languages but should be re-fetchable with some kind of script.

**Phabricator:** 
[T367908](https://phabricator.wikimedia.org/T367908)

### Test Steps
1. Unit tests should run automatically
2. Manual CLI tests in `sharedlib`: run `npm test`


tonisevener (approved):
This looks great, thank you @bvibber!

I have been working against this in https://github.com/wikimedia/wikipedia-ios/pull/4924 to connect the full flow. So far I haven't noticed any performance problems significant enough to halt an experiment.

PR #4906, opened 2024-07-31
[ALT TEXT] Bottom sheet - part 1

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T370232

### Notes
* This creates a barebones version of the bottom sheet
* It doesn't match the designs yet
* I'm still waiting on design's update on the icon we should use [see ticket](https://phabricator.wikimedia.org/T370232#10028171)
* Using a text view for multiline purposes created the need to add the placeholder as a label
* I'm updating ArticleViewController init method to receive the alt text view model, let me know if this is not desired (I can create a separate init, adding the VM after creating the VC was causing to be nil when needed)
* The VM property, when kept weak, did not hold the value correctly. It was left strong and nulled on the existing deinit method. 

### Test Steps
1. Go through all the steps to get into alt text - tweaking some values might be necessary
2. When prompted to add alt text, tap "add"
3. Verify the basics of the design, dark mode, dynamic type
4. The placeholder should disappear when typing
5. The button should enable only with text = there was no character minimum defined


Wikipedia/Code/ArticleViewController.swift
```diff
@@ -119,6 +126,7 @@ class ArticleViewController: ViewController, HintPresenting {
         contentSizeObservation?.invalidate()
         messagingController.removeScriptMessageHandler()
         articleLoadWaitGroup = nil
+        altTextBottomSheetViewModel = nil
```
tonisevener, line 129: @mazevedofs Were you seeing some memory issues without this line? I don't think it hurts to have it but I thought it might be unnecessary.

tonisevener (approved):
Looks good! One minor question but nothing worth holding up the merge.

Wikipedia/Code/ArticleViewController.swift
mazevedofs: No issues so far, but I feared I could be missing something since I had to keep the reference strong. We can remove it in future PRs for the bottom sheet.

PR #4910, opened 2024-08-01
Update HtmlUtils to handle apostrophes and ampersands regression bugs

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T371148 & https://phabricator.wikimedia.org/T371149

### Notes
* This PR updates the HtmlUtils to search for another HTML apostrophe tag and to handle HTML tags for characters better when stripping HTML from strings

### Test Steps
1. Run the app, find an article on the explore feed that has an apostrophe on the title. It should display correctly
2. Go to the top read widget (or any place that uses the removingHTML String extension)
3. Confirm it handles & correctly (and other tags like <>, etc, etc)

### DIFF
--- a/Components/Sources/Components/Utility/HTMLUtils.swift
+++ b/Components/Sources/Components/Utility/HTMLUtils.swift
@@ -644,7 +644,7 @@ public struct HtmlUtils {
                 replaceText = ">"
             case "&lt;":
                 replaceText = "<"
-            case "&apos;":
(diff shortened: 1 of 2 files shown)


tonisevener (approved):
Working great! I noticed a problem with the featured article widget stripping an entity entirely (July 30th "Hundred Years War"), but looks like it was caused by the old stripping method. I pushed a commit using the new one.

PR #4923, opened 2024-08-14
[ALT TEXT] Alt text preview step

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T370359

### Notes
* This PR adds the review step to the alt text flow
* Publishing alt text now works from this view
* I'm downloading the full image at the modal to prevent downloading two images, even if the modal only would need the thumbnail

### Test Steps
1. Go through the alt text flow
2. Verify the preview matches the designs
3. Verify you can return to the modal sheet and that the alt text is pre-filled in the text field
4. Publish and verify your edit
5. You should go back to the image recs flow, to the next recommendation


tonisevener (changes requested):
A couple of minor things from code review, also I'm seeing some bugs:

1. The preview image covers the label UI
2. After publishing it goes back to the article view and continues to display the half sheet modal.

out

Wikipedia/Localizations/es.lproj/Localizable.strings
```diff
@@ -163,15 +163,23 @@
 "advanced-settings-title" = "Ajustes avanzados";
 "alt-text-add-button-title" = "Añadir";
 "alt-text-do-not-add-button-title" = "No añadir";
+<<<<<<< HEAD
```
tonisevener, line 166: @mazevedofs Seeing some merge conflicts in this file.

PR #4924, opened 2024-08-14
[ALT TEXT] Connect Flow C end-to-end

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T370305, https://phabricator.wikimedia.org/T372008

### Notes
This PR does several things:

1. Immediately after we publish an article edit, for valid users, we are fetching the article's full wikitext. We need this step for users that only edit an article section, because there is a chance that they are editing a section without any images, but other sections _do_ have valid images. The editor flow so far only had the section wikitext, so I had to add an extra step to fetch the full article wikitext for evaluation later. 
2. Once the publish is complete and we have their full article wikitext, we run the wikitext through @bvibber's javascript library to detect images missing alt text.
3. Once that image wikitext is detected, we try to extract a simplified version of the image caption. This logic probably should eventually live in the JS library, but I wrote this in Swift for speed.
4. We present the alt text prompt/panel.
5. After a user adds alt text and taps Next, we insert the new alt text before the caption, then post the changes.
6. For this flow I had to pull down additional magic words to help with localizing alt text logic in these previous steps.
7. I also did some cleanup by moving the JS library into WMFData.

### Test Steps
1. Fresh install app. Go to developer settings menu, enable alt text feature flag and set experiment to 100% feature flag.
2. Go to an article with an image on Test Wiki.
3. Make any kind of edit to the wikitext. Confirm after edit you see the alt text prompt.
4. Tap into the alt text flow and add alt text.
5. After alt text publishes, check the diffs on Desktop Wikipedia and confirm alt text is added correctly.

Note: Sometimes I am able to insert alt text before the caption, but if that fails for whatever reason, I am inserting it at the very end of the image wikitext. We have a follow-up task (https://phabricator.wikimedia.org/T372517) to fix this later if we do a permanent version.


mazevedofs (approved):
Looking great! 
I didn't expect all this logic to get this complex 🤯

PR #4927, opened 2024-08-16
[ALT TEXT] Enable always showing alt text entry point

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T372098

### Notes
I think I fixed the strangeness I was seeing earlier with alt-text prompts. This logic bypasses the persisted experiment and prompt states so that we can always see the entry point after publishing an image recommendation.

### Test Steps
1. Fresh install. Enable alt text feature flags.
2. Go through image recommendations flow. Confirm you see alt text prompt after first publish.
3. Publish alt text.
4. Go through next image recommendation flow. Confirm you see prompt again after image recommendation publish.


mazevedofs (approved):
Looking good!

PR #4932, opened 2024-08-20
Remove double space from a message

amire80:


mazevedofs (approved):
Thank you for the fix!

PR #4933, opened 2024-08-20
[ALT TEXT] Present gallery after tapping image in half sheet

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T370338

### Notes
This PR shows the gallery after tapping the image in the alt text half sheet. It also logs an impression event when the gallery is presented.

### Test Steps
1. Go through alt text flow, tap image in half sheet. Confirm gallery is presented. Confirm when dismissing gallery the half sheet appears again.


mazevedofs (approved):
Looks good!

PR #4934, opened 2024-08-21
[ALT TEXT] Add feedback alerts

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T370360

### Notes
* This PR adds the feedback alerts with post-edit surveys

### Test Steps
1. Go through flow B to the end
2. After publishing alt text, you should get a Yes/No survey
3. Note that the alert has the buutons side-by-side. This is the default with two options for this type of alert.
4. Go through flow C to the end.
5. You should get a survey with three options. After clicking one option, the same Yes/No survey above displays.
6. Check the event logging and translation
7. Also look out for the toasts!


tonisevener (changes requested):
@mazevedofs Looking good!

I pushed one minor commit because I wasn't seeing the alert when suppressing image recommendation posts. That code will go away when we remove that feature flag.

A couple of other things I noticed:

The Flow C survey 2 should have different question copy: "Would you be interested in receiving similar notifications about suggested edits on articles after you edit them?" instead of "Would you use a feature that provided a feed of images in need of alt text?".

There's also a "Feedback submitted" toast after the alerts are complete, and a `feedback_submit_toast` logging action.

tonisevener (approved):
Working well for me! I merged with `main` and fixed a bug that was brought in from my old PR (I forgot to translate "Feedback submitted"). Also fixed a minor var name misspelling.

PR #4938, opened 2024-08-23
Fix editor dismissal bugs

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T372301 & https://phabricator.wikimedia.org/T372839

### Notes
**Background:** In 7.5.6, we reworked our single root navigation controller (which contained the tab bar controller) into 5 root navigation controllers (that serve as the root view controllers of the tab bar controller). `EditorViewController` wraps up `InsertLinkViewController` in a navigation controller and presents it [here](https://github.com/wikimedia/wikipedia-ios/blob/main/Wikipedia/Code/EditorViewController.swift#L712). This logic was not changed 7.5.6.

Inspecting the `presentedViewController` and `presentedViewController` properties upon dismissal [here](https://github.com/wikimedia/wikipedia-ios/blob/main/Wikipedia/Code/EditorViewController.swift#L909) for EditorViewController and InsertLinkViewController shows some odd things:

- EditorViewController.presentedViewController = nil
- InsertLinkViewController.presentingViewController = WMFAppViewController (i.e. the tab bar VC)

The 7.5.5 release looked like this:

- EditorViewController.presentedViewController = InsertLinkViewController*
- InsertLinkViewController.presentingViewController = EditorViewController*

_*Actually the containing UINavigationController, but that's not important._

The [documentation](https://developer.apple.com/documentation/uikit/uiviewcontroller/1621380-present) for presenting a view controller says:

> The object on which you call this method may not always be the one that handles the presentation. Each presentation style has different rules governing its behavior. For example, a full-screen presentation must be made by a view controller that itself covers the entire screen. If the current view controller is unable to fulfill a request, it forwards the request up the view controller hierarchy to its nearest parent, which can then handle or forward the request.

So it seems like something about our 7.5.6 navigation changes caused the system to no longer think the EditorViewController counted as full screen, so the responsible presenter logic walked up and landed on WMFAppViewController. The dismiss call in EditorViewController failed silently, because it was not presenting the link view controller anymore.

I looked and oddly the navigation controller that contains the EditorViewController sets it's modalPresentationStyle to `overCurrentContext`. Changing it to `overFullScreen` fixed the bugs.

### Test Steps
1. Practice the actions on the Insert Link view and Edit Link flows from the Editor. Confirm the screens now dismiss and return to Editor.
2. Do the same for the insert media flow.


mazevedofs (approved):
Working great! 🎉

PR #4939, opened 2024-08-23
Fix `stringFromHTML` crash

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T372119

### Notes
A rule of thumb is to be careful any time we manipulate a string during a loop. We are first analyzing the entire string for html entities and are saving an array of NSRanges in `entityReplaceData`.

The first iteration of this replace loop on line 411 works fine, but after that iteration the string is no longer as long as it was when we created `entityReplaceData`. So the next iteration of the loop will have an invalid NSRange. In some cases the line at 411 fails because we are asking it to update at an out-of-range index.

One very easy way around this (one I lean on all the time as you can see elsewhere in this file) is to iterate through the entityReplaceData in reverse. If we update the string starting from the back, you will not run into out-of-index issues because the next iteration range will be replacing characters earlier in the string. This solution assumes the entityReplaceData NSRanges are in order from front to back.

I think this crash only occurs when we are replacing a string that contains two or more html entities, which could be why our original dev testing didn't catch it (I only tested those with one entity).

### Test Steps
1. Go to article revision history (the Donald Trump article on EN Wikipedia is one that is currently crashing), confirm no crash occurs.


mazevedofs (approved):
Thanks for fixing it! I'll definitely use the `.reversed()` from now on when working with these ranges.

PR #4942, opened 2024-08-26
Refactor out ArticleViewController into extensions

l-olson1214:
**Phabricator:**  N/A

### Notes
* Refactor ArticleViewController to have extension files where needed, clean up warnings of unused variables

### Test Steps
1.

### DIFF
--- a/Wikipedia/Code/ArticleViewController+ArticleAsLivingDoc.swift
+++ b/Wikipedia/Code/ArticleViewController+ArticleAsLivingDoc.swift
@@ -1,0 +1,125 @@
+import WMFComponents
+import WMF
+import CocoaLumberjackSwift
+import WMFData
+
+// MARK: Article As Living Doc Protocols
+extension ArticleViewController: ArticleAsLivingDocViewControllerDelegate {
+    func livingDocViewWillPush() {
+        surveyTimerController?.livingDocViewWillPush(withState: state)
+    }
+    
+    func livingDocViewWillAppear() {
+        surveyTimerController?.livingDocViewWillAppear(withState: state)
+    }
+    
+    var articleAsLivingDocViewModel: ArticleAsLivingDocViewModel? {
+        return articleAsLivingDocController.articleAsLivingDocViewModel
+    }
+    
+    func fetchNextPage(nextRvStartId: UInt, theme: Theme) {
+        articleAsLivingDocController.fetchNextPage(nextRvStartId: nextRvStartId, traitCollection: traitCollection, theme: theme)
+    }
+
+    var isFetchingAdditionalPages: Bool {
+        return articleAsLivingDocController.isFetchingAdditionalPages
+    }
+}
+
+extension ArticleViewController: ArticleAsLivingDocControllerDelegate {
+    var abTestsController: ABTestsController {
+        return dataStore.abTestsController
+    }
+    
+    var isInValidSurveyCampaignAndArticleList: Bool {
+        surveyAnnouncementResult != nil
+    }
+    
+    func extendTimerForPresentingModal() {
+        surveyTimerController?.extendTimer()
+    }
+}
+
+extension ArticleViewController: ArticleSurveyTimerControllerDelegate {
+    var displayDelay: TimeInterval? {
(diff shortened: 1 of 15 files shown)


Wikipedia/Code/ArticleViewController+Notifications.swift
```diff
@@ -0,0 +1,182 @@
...
+
+        messagingController.setup(with: webView, languageCode: articleLanguageCode, theme: theme, layoutMargins: articleMargins, leadImageHeight: leadImageHeight, areTablesInitiallyExpanded: areTablesInitiallyExpanded, userGroups: userGroups)
+    }
+    
+    func setupToolbar() {
+        enableToolbar()
+        toolbarController.apply(theme: theme)
+        toolbarController.setSavedState(isSaved: article.isAnyVariantSaved)
+        setToolbarHidden(false, animated: false)
+    }
```
tonisevener, line 132: There are several `setup`-y method names, here and elsewhere. Maybe they could live in their own +Setup extensions file?

Wikipedia/Code/ArticleViewController+ArticleStateRestoration.swift
```diff
@@ -0,0 +1,67 @@
+//~~~**DELETE THIS HEADER**~~~
```
tonisevener, line 1: Need this header deleted. At the time this default was made, Xcode didn't support the ability to have an empty header when a new file was created. It's possible that has since changed.

Right now it adds `DELETE THIS HEADER`, and there's a git commit hook that checks if this exists in any file before  allowing the commit.

tonisevener (changes requested):
@l-olson1214 Thanks for taking this on! I'm totally on board to move related methods out into extensions to more closely align with the original way ArticleViewController was built.

I commented on some methods in your extensions that I think should be shuffled around, and maybe some new extensions could be made - let me know what you think. One other thing:

screenshot2

For any new file that lives in the Wikipedia target, we should also check the Staging and Experimental targets in this screenshot. Otherwise the Experimental and Staging schemes won't compile. Thanks!

Wikipedia/Code/ArticleViewController+ArticleAsLivingDoc.swift
```diff
@@ -0,0 +1,125 @@
...
+        
+        guard let identifier = surveyAnnouncementResult?.campaignIdentifier else {
+            return false
+        }
+        
+        return SurveyAnnouncementsController.shared.userHasSeenSurveyPrompt(forCampaignIdentifier: identifier)
+    }
+    
+    var shouldShowArticleAsLivingDoc: Bool {
+        return articleAsLivingDocController.shouldShowArticleAsLivingDoc
```
ankushkushwaha, line 62:
```suggestion
        articleAsLivingDocController.shouldShowArticleAsLivingDoc
```

Wikipedia/Code/ArticleViewController+ArticleAsLivingDoc.swift
```diff
@@ -0,0 +1,125 @@
...
+    }
+    
+    var livingDocSurveyLinkState: ArticleAsLivingDocSurveyLinkState {
+        return articleAsLivingDocController.surveyLinkState
```
ankushkushwaha, line 66:
```suggestion
        articleAsLivingDocController.surveyLinkState
```

Wikipedia/Code/ArticleViewController+ArticleLoad.swift
```diff
@@ -0,0 +1,232 @@
...
+            bottomSheetViewController.isModalInPresentation = true
+            self.altTextBottomSheetViewController = bottomSheetViewController
+            
+            present(bottomSheetViewController, animated: true) { [weak self] in
+                self?.presentAltTextTooltipsIfNecessary(force: false)
+            }
+        }
+    }
+    
+    internal func presentAltTextTooltipsIfNecessary(force: Bool = false) {
```
ankushkushwaha, line 102: Do we need to specify internal explicitly? As swift has all method as internal by default.

PR #4943, opened 2024-08-26
[ALT TEXT] Fix stub article content inset bug

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T372583

### Notes
I was updating the content inset in various places within ArticleViewController, to ensure enough spacing was added to the bottom of article for the half sheet modal to display.

It seems like when an _image_ is loaded in the article, that resets the content inset and removes the bottom spacing. This causes the half sheet modal to display on top of the content. In this PR I am adding a callback closure to SchemeHandler (which handles all of our article calls). When an image successfully loads, I am setting the content inset to the correct spacing again.

### Test Steps
1. Enable Alt Text feature flag in dev settings. Go through image recommendations flow. Note that you will have to publish a live edit here, because we need the article to contain an image for proper testing of this. It also helps to check out the image recommendation article first, and be sure it is a very short (stub) article. If the article is too long, tap "Not sure" to find another image recommendation for a short article. 
2. Tap add on the alt text prompt.
3. When the article appears, confirm you can scroll down to the bottom and see all of the content.


mazevedofs (approved):
Looks good!

PR #4945, opened 2024-08-26
[ALT TEXT] Publish button tweak

mazevedofs:
**Phabricator:** 

### Notes
* Disables the publish button when waiting for a response from posting the publish
* Enables it when an error occurs - here I'm verifying if `error != nil`, I'm assuming that is not possible to have a successful publish that returns a valid revID and an error 
* I added a generic error toast so the user knows what happened

### Test Steps
1. Go through alt text with a valid edit
2. Simulate an error before publishing (turning off wifi for example)
3. Tap publish - the button should disable
4. The button should be enabled when publishing returns an error
5. Try publishing normally (with wifi on); verify it works as normal


WMFComponents/Sources/WMFComponents/Components/Suggested Edits/Alt Text Experiment/Preview/WMFAltTextExperimentPreviewViewController.swift
```diff
@@ -21,6 +21,7 @@ final public class WMFAltTextExperimentPreviewViewController: WMFCanvasViewContr
     private let hostingViewController: WMFAltTextExperimentPreviewHostingViewController
     private var viewModel: WMFAltTextExperimentPreviewViewModel
     public weak var delegate: WMFAltTextPreviewDelegate?
+    public var publishButton: UIBarButtonItem?
```
tonisevener, line 24: (Minor) I would rather us not expose the button itself if we can avoid it. Can you make this property private, and make a public method that only toggles the publish button enabled state?

tonisevener (commented):
@mazevedofs Working well! Just one minor request.

WMFComponents/Sources/WMFComponents/Components/Suggested Edits/Alt Text Experiment/Preview/WMFAltTextExperimentPreviewViewController.swift
mazevedofs: Done!

tonisevener (approved):

PR #4953, opened 2024-08-30
T370205: Update to Xcode 16/ iOS 16 (and fix iPad transition issues)

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T370205
https://phabricator.wikimedia.org/T370967

### Notes
* Tested locally regarding not updated CICD 
* Tested white logo, WMFdata, and WMFcomponennts, all passed

### Test Steps
1.


tonisevener (changes requested):
@l-olson1214 Thanks for working on this! It's looking promising.

I pushed one minor commit turning user sandboxing off - my builds weren't working due to SwiftLint with it on. There may be a better fix but this works for now.

Mainly minor things, and one regression I found with the article highlight context menu. I'm not totally sure of how to fix it but happy to pair if you want to on it.

Wikipedia/Code/ArticleViewController+Editing.swift
```diff
@@ -559,9 +559,7 @@ extension ArticleViewController: WMFAltTextPreviewDelegate {
                        navigationController.viewControllers.count > 2 {
                         // pop back two view controllers.
                         let index = (navigationController.viewControllers.count-1) - 2
-                        if let _ = navigationController.viewControllers[index] as? ArticleViewController {
-                            navigationController.popToViewController(navigationController.viewControllers[index], animated: true)
-                        }
+                        navigationController.popToViewController(navigationController.viewControllers[index], animated: true)
```
tonisevener, line 562: Actually we do want to make sure this view controller is an ArticleViewController before popping. What did the SwiftLint error say?

tonisevener (changes requested):

Wikipedia/Code/ArticleViewController+Editing.swift
l-olson1214: Unused Optional Binding Violation: Prefer `!= nil` over `let _ =` (unused_optional_binding)

I could do something like 

`if navigationController.viewControllers[index] is ArticleViewController {
    navigationController.popToViewController(navigationController.viewControllers[index], animated: true)
}
`

tonisevener: That sounds good!

Wikipedia/Code/ArticleViewController+ContextMenu.swift
```diff
@@ -1,11 +1,15 @@
...
+        }
         let editMenuItemTitle = CommonStrings.editContextMenuTitle
-        let editMenuItem = UIMenuItem(title: editMenuItemTitle, action: #selector(editMenuItemTapped))
-        
-        UIMenuController.shared.menuItems = [editMenuItem, shareMenuItem]
+        let editAction = UIAction(title: editMenuItemTitle) { _ in
+            self.editMenuItemTapped()
+        }
+
+        let menu = UIMenu(title: "", children: [editAction, shareAction])
```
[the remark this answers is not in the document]
tonisevener: @l-olson1214 I think I found a fix for this - I pushed a commit. Found some [suggestions](https://stackoverflow.com/a/75667352) that overriding `buildMenu` might work, and so far I haven't noticed regressions.

Also it looks like our old implementation had "Share" twice (according to screenshot above) 🤦‍♀️ I removed the system share in preference of our custom share action.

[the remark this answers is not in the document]
l-olson1214: Interesting! Thanks!!

tonisevener (approved):
👏 Looks great so far! I pushed some minor things, happy to go over them whenever if you'd like. I will merge once the new RC finishes downloading for me and I get tests to pass locally.

PR #4955, opened 2024-09-02
Calculate maximum amount donation for currencies other than USD 

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T365043

### Notes
* Use the USD max/min values to calculate the max value for other currencies based on the USD max of 25000
* The calculation is as follows: 
Ratio the current currency minimum donation to the USD minimum donation, then multiply this value by the max amount in USD

### Test Steps
1.  Run the app with the region set as US
2. Try to donate more than 25000 with Apple Pay
3. Verify error message
4. Set the region as any other existing in the Donate Config
5. Try to donate more than the equivalent value in USD
6. Check error message


WMFData/Sources/WMFData/Models/Donor Experience/WMFDonateConfig.swift
```diff
@@ -12,4 +12,16 @@ public struct WMFDonateConfig: Codable {
...
         return currencyTransactionFees[currencyCode] ?? currencyTransactionFees["default"]
     }
+
+    public func getMaxAmount(for currencyCode: String) -> Decimal {
+        var max = currencyMaximumDonation[currencyCode] ?? Decimal()
+
+        if max.isZero {
+            if let defaultMin = currencyMinimumDonation["USD"] {
+                max = (currencyMinimumDonation[currencyCode] ?? 0.0) / defaultMin *
+                (currencyMaximumDonation["USD"] ?? 0.0)
```
l-olson1214, line 22: Out of curiosity, does defaulting to 0.0 cause any issues? I would worry about it trying to divide by zero and causing an error/crashing.

mazevedofs: Good catch, updated

l-olson1214 (approved):

WMFData/Sources/WMFData/Models/Donor Experience/WMFDonateConfig.swift
```diff
@@ -13,4 +13,15 @@ public struct WMFDonateConfig: Codable {
     public func transactionFee(for currencyCode: String) -> Decimal? {
         return currencyTransactionFees[currencyCode] ?? currencyTransactionFees["default"]
     }
+
+    public func getMaxAmount(for currencyCode: String) -> Decimal {
+        var max = currencyMaximumDonation[currencyCode] ?? Decimal()
+
+        if max.isZero {
+            if let defaultMin = currencyMinimumDonation["USD"], let defaultMax = currencyMaximumDonation["USD"], let currencyMax = currencyMinimumDonation[currencyCode] {
```
tonisevener, line 21: Minor - I think the final var here should be named `let currencyMin` instead of `let currencyMax`.

mazevedofs: I opened another PR to fix that https://github.com/wikimedia/wikipedia-ios/pull/4961

PR #4959, opened 2024-09-05
Temp accounts fixes

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T339378

### Notes
This PR refactors our current `WMFAuthenticationManager` to support temporary accounts. It fixes some underlying bugs we had with automatic logout after a temporary account was created. It also attempts to clean up and update the general naming of methods and properties to fit the new set of auth states (which I am calling `ip`, `temporary`, or `permanent`).

High-level, I discovered that there is a handy [userinfo API call](https://www.mediawiki.org/wiki/API:Userinfo) that the app was already occasionally calling. This method returns a user object with one of the 3 states we need, depending on what underlying auth cookies we are sending when making that call. So I am now treating the result of this call as the source-of-truth for auth states and am fetching it after a user successfully logs in.

Keep in mind that we could have different auth states depending on whether we are working against a pilot wiki or not, so having a single auth state property within `WMFAuthenticationManager` would not be accurate. To deal with this nuance I am keeping track of these userinfo API objects (`WMFCurrentUser`) in a dictionary property called `currentUserCache`, keyed against the wiki `siteURL`. So potentially in the future, if someone is editing on a pilot wiki, we just need to pull the user from that `siteURL` and check their `authState` enum to see if they are `temporary` or `ip` on that particular wiki, to display the temporary account UI banners.

I have a couple of convenience properties (`authStatePermanentUsername` & `authStateIsPermanent`) that comb through the entire `currentUserCache` and return if any one of them is in a permanent auth state. This may be a bit confusing but in general, if the user is considered permanent on any wiki, then they should be considered permanent on all of them, because we copy the underlying auth cookies over to all domains. These two methods were easier to write than adding 300+ duplicate permanent user objects to currentUserCache for each language siteURL. I don't expect the number of items saved in currentUserCache will ever get very big, so I don't think looping through those whenever we call `authStatePermanentUsername` & `authStateIsPermanent` will be much of a performance hit.

`authStatePermanentUsername` & `authStateIsPermanent` are then used mostly used outside this class throughout the app. We use it to adjust the UI for a logged in state, as well as for displaying the logged in user UI. In general I tried to avoid the phrasing "logged in" and instead stick with "permanent", just to be very clear, though there are still some areas of the app that reference being "logged in".

On top of this, I made a couple of minor improvements. I ensured "Test" is displayed in the explore feed cards when needed ("On Test Wikipedia", etc.). I also added the Test 2 language so we can test temporary accounts across multiple domains.

### Test Steps
1. Launch app in a logged out state. To do this you may need to run the app from this branch and log out in the app first. Then fresh install.
2. Edit an article on Test Language and publish. After editing, go to revision history and confirm you see the temporary account name.
3. Terminate the app. Run again, then edit an article on Test 2 Language and publish. After editing, go to revision history and confirm you see _same_ temporary account name as you did in step 2.
4. Now edit something on EN Wikipedia. You can edit your own user talk page if you don't want to edit an article (search User_talk:TSevener (WMF) for example). After editing, go to revision history and confirm you see an IP address as your edit.
5. Go back and edit a Test Language article. Go to article history, confirm you see the same temporary account name as you did in step 2.
6. Log into the app. Go back to Test, Test 2, and EN Wiki articles (or talk pages) that you edited earlier. Edit them again. Go to article history and confirm you see your username as the edit.
7. As you click around the app, note if anything feels like a regression with authentication (unexpected logged out panels, Settings displaying an unexpected logged in our out state, etc.).


mazevedofs (approved):
Looking great. The app seems to be working as expected based on my testing—I was able to edit and get the temp accounts. 
I also love the clarity that `isPermanent` brings to this context.

PR #4960, opened 2024-09-09
Locally save native donations

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T372343

### Notes
* This PR adds logic and some UI  for saving locally a history of Apple Pay donations
* It doesn't address web donations
* I created a new `SharedContainerCacheRemover` that doesn't require a specific value, to allow for the whole directory deletion. I wanted to bypass the need to pass in a value <T> when not needed - in this case, I'm going for the whole file. Let me know if there's a better way to handle this case.

* Note: pending Design review for now - there was no Figma for this, so I created a simple Ui, and it's now under design review, but the saving and deleting logic can be reviewed independently of the UI code

### Test Steps
1. Make a donation 
2. Go to Settings -> Account -> Donation History
3. Make sure you see your donation with date and amount
4. Click delete.
5. You should see an empty view with a message
6. Background and terminate the app. Go to donation history again. Make sure the history is still empty
7. You can check the local storage, look for Donor Experience/AppLocalDonationHistory.json


tonisevener (commented):
This is working really well! 

> I wanted to bypass the need to pass in a value when not needed - in this case, I'm going for the whole file. Let me know if there's a better way to handle this case.

One thing you could try is to remove the class generic from SharedContainerCacheStore, and only have the necessary methods be generic. You can look at [this commit](https://github.com/wikimedia/wikipedia-ios/commit/e87686cb746e2e2e6efe083e90c71da6fa147f30) to see what I mean. It could use more testing to be sure I didn't break anything with the other users of SharedContainerCache. 😅

One bug I noticed when inspecting the underlying json file is that it seems `isFirstDonation` is saved as `false` for the first donation and `true` for the following donations. So probably just a missing boolean somewhere.

WMFData/Sources/WMFData/Data Controllers/Donor Experience/WMFDonateLocalHistory.swift
```diff
@@ -0,0 +1,26 @@
+import Foundation
+
+public final class WMFDonateLocalHistory: Codable {
+
+    public let donationTimestamp: String
+    public let donationType: DonationType
+    public let donationAmount: Decimal
```
tonisevener, line 7: I think we should save the currencyCode here as well - it'll help us aggregate them and display properly in the future.

WMFComponents/Sources/WMFComponents/Components/Donate Form/WMFDonateViewModel.swift
```diff
@@ -192,10 +192,12 @@ public final class WMFDonateViewModel: NSObject, ObservableObject {
     
     private(set) weak var coordinatorDelegate: DonateCoordinatorDelegate?
     private(set) weak var loggingDelegate: WMFDonateLoggingDelegate?
-    
+
+    private let userDefaultsStore: WMFKeyValueStore?
```
tonisevener, line 196: It doesn't look like this property is used.

tonisevener (commented):
Thanks! I found nothing show-stopping, but these should be quick fixes.

Wikipedia/Code/AccountViewController.swift
```diff
@@ -177,18 +177,43 @@ class AccountViewController: SubSettingsViewController {
...
+            alertController.addAction(logoutAction)
+            alertController.addAction(cancelAction)
+            self.navigationController?.present(alertController, animated: true, completion: nil)
         default:
             break
         }
     }
 
+    private func deleteLocalHistory(completion: @escaping () -> Void) {
+        donateDataController.deleteLocalDonationHistory()
```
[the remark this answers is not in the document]
mazevedofs: Yeah not sure what I tried to accomplish here 😂 deleted!

tonisevener (approved):

PR #4962, opened 2024-09-11
T373126: Fix picture of the day not being shown

christianbilodeau:
**Phabricator:** https://phabricator.wikimedia.org/T373126

This is a suggestion to fix the "picture of the day" image not being presented when the user deep-links to the app by tapping on the widget, when the app is in a backgrounded state.

The issue seems to have been introduced in [this PR](https://github.com/wikimedia/wikipedia-ios/pull/4876).  As the `presentViewController` API was now used to display the image from `WMFAppViewController`, the system was trying to display two controllers at once:

- the photo gallery
- the splash screen (which is shown as the user deep-links into the app via `appViewController.showSplashView()`)

The issue is not present when opening the app from a terminated state because the splash screen is not shown manually in that case.

The fix suggested here is to avoid presenting the splash screen with the `presentViewController` API and go back to using `add/remove subview`, like it was done before in `SplashScreenViewController`. This way the splash screen can't interfere with other controllers.

### Notes

- Although there is a call to `dismissPresentedViewControllers` before presenting the gallery, it's executed asynchronously, so the conflict still occurs. A completion block could be added to ensure dismissing is finished, but it seemed less elegant as a fix
- It looks like [the PR mentioned above](https://github.com/wikimedia/wikipedia-ios/pull/4876) removed animation for dismissing the splash screen. Assuming this was intentional, I went ahead and removed some dead code which was ensuring the smoother transition

### Test Steps
- Add the picture of the day widget and tap it to open the app, the picture should be displayed
- Make sure the splash screen works as before
- Make sure migration splash screen/animation works as before

### Screenshots/Videos

Without the fix, the following is logged:

Screenshot 2024-09-11 at 16 04 32

### DIFF
--- a/Wikipedia/Code/WMFAppViewController.m
+++ b/Wikipedia/Code/WMFAppViewController.m
@@ -63,7 +63,7 @@ NSString *const WMFLanguageVariantAlertsLibraryVersion = @"WMFLanguageVariantAle
 @property (nonatomic, strong, readonly) WMFPlacesViewController *placesViewController;
 @property (nonatomic, strong, readonly) WMFHistoryViewController *recentArticlesViewController;
 
-@property (nonatomic, strong, readonly) WMFSplashScreenViewController *splashScreenViewController;
+@property (nonatomic, strong) WMFSplashScreenViewController *splashScreenViewController;
 
 @property (nonatomic, strong) WMFSavedArticlesFetcher *savedArticlesFetcher;
 
@@ -1578,7 +1578,7 @@ static NSString *const WMFDidShowOnboarding = @"DidShowOnboarding5.3";
 #pragma mark - Splash
 
 - (void)showSplashView {
-    if (_splashScreenViewController) {
+    if (self.splashScreenViewController) {
         return;
     }
     WMFSplashScreenViewController *vc = [[WMFSplashScreenViewController alloc] initWithNibName:nil bundle:nil];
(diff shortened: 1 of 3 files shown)


Wikipedia/Code/WMFAppViewController.m
```diff
@@ -1576,22 +1578,31 @@ - (void)presentOnboardingIfNeededWithCompletion:(void (^)(BOOL didShowOnboarding
 #pragma mark - Splash
 
 - (void)showSplashView {
+    if (_splashScreenViewController) {
```
tonisevener, line 1581: Would you be opposed to using self notation `self.splashScreenViewController` instead of `_splashScreenViewController`? It would be a little more consistent with how the other properties (`self.periodicWorkerController`, etc.) are called in here, and I think it will still function the same. I generally try to avoid accessing the instance variable directly outside of getters and setters for Obj-C.

tonisevener (changes requested):
Thanks for the fix @christianbilodeau! Nice investigative skills on determining the source of the bug here - this was more advanced than I anticipated for a "good first task" tag.

For https://github.com/wikimedia/wikipedia-ios/pull/4876, I thought the `beginAppearanceTransition` calls felt a little hacky since the comment indicated it was only used to bypass some underlying UINavigationController behaviors. At the time it looked like I might get away with the presentation API without regressions. 😆 Anyways, I'm good with bringing this back - we know it works.

> It looks like https://github.com/wikimedia/wikipedia-ios/pull/4876 removed animation for dismissing the splash screen. Assuming this was intentional, I went ahead and removed some dead code which was ensuring the smoother transition

Yeah, my guess is `ensureMinimumShowDuration` was used to prevent the splash from hiding too quickly, thus just showing a flash of the splash view. Personally I don't really notice a flash, so maybe it was something only noticeable with the old navigation controller setup. I'm good with cleaning it out.

I have one ask for self notation (see comment). Let me know what you think!

Wikipedia/Code/WMFAppViewController.m
christianbilodeau: I agree! Code updated, thank you

christianbilodeau: @tonisevener I'm a stranger to this process, should the phabricator ticket be moved back to "Needs code review" or should it be left as is? 🙏

tonisevener: @christianbilodeau You can move it back into "Needs code review". Thanks!

tonisevener (approved):
Looks great, thanks @christianbilodeau!

PR #4963, opened 2024-09-11
Fundraising cleanup - use metricsID consistently

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T367360

### Notes
This is partially a follow-on to https://github.com/wikimedia/wikipedia-ios/pull/4848. In that PR we updated the campaign identifier for fundraising events sent via MEP, using a new identifier called `metricsID`. There were several spots still sending the old format, so those have been updated.

We also now want to send this same identifier format to the `submitPayment` endpoint under the "banner" parameter. This removes our dependence on the `utm_source` parameter from the campaign config (see linked Phabricator task).

### Test Steps
Since there's no active campaign going, there's not an easy way to confirm this. I suggest we lean on data validation for testing this one.


mazevedofs (approved):
Looks good!

PR #4966, opened 2024-09-12
Fixed typos in file paths breaking compilation on case-sensitive file systems

tladesignz:


tonisevener:
@tladesignz Thanks for the fix! I have created an associated Phabricator task so that we can track it's status on our workboard - https://phabricator.wikimedia.org/T377345.

We will post feedback here in GitHub if any changes are needed from code review.

tonisevener:
@tladesignz Thanks for the PR. Can you provide test steps for us? I can spin up a case-sensitive volume on my machine and compile from there, but I wanted to make sure that matches your setup.

tladesignz:
To reproduce:
- Create case sensitive APFS volume
- Check out project into volume
- Open Xcode project
- Check project navigator pane: You will find some files marked red, because they cannot be found
- Try to compile project: won't work, files could not be found
- Apply PR fixing file name casing
- Red marking will go away
- Project compiles

tonisevener (approved):
Working for me, thanks!

PR #4969, opened 2024-09-13
T374160: Profile View + Entry Point (pt 1)

l-olson1214:
**Phabricator:** 

https://phabricator.wikimedia.org/T374160
https://phabricator.wikimedia.org/T374162

### Notes
* Not UI finished 100% for the profile view - UI for the person with notifications in the upper right hand corner is finished
* Some logic not done ("Done" button, translations, etc)

### Test Steps
1. 

### Screenshots/Videos

Screenshot 2024-09-13 at 1 35 58 PM

Screenshot 2024-09-13 at 1 40 00 PM


Wikipedia/Code/ProfileView.swift
```diff
@@ -0,0 +1,233 @@
+import SwiftUI
+import WMFComponents
+
+struct ProfileView: View {
```
tonisevener, line 4: @l-olson1214 It looks like there's some duplicate classes here, probably as a result of us moving code files around earlier this week. I would like to remove the ProfileView.swift that lives here in the Wikipedia/Code/ directory, and keep the one in WMFComponents/Components/Profile directory. Same goes for ProfileViewController. Can you delete this file (you may need to copy it's logic over into the WMFComponents class).

Wikipedia/Code/ProfileViewController.swift
```diff
@@ -0,0 +1,29 @@
+import SwiftUI
+import UIKit
+
+class ProfileViewController: UIViewController {
```
tonisevener, line 4: Delete this class too (see previous comment).

WMFComponents/Sources/WMFComponents/Components/Profile/ProfileView.swift
```diff
@@ -0,0 +1,48 @@
+import SwiftUI
+
+struct ProfileView: View {
```
tonisevener, line 3: @l-olson1214 Can you rename this to WMFProfileView? We've been trying to prepend any public name that lives in WMFComponents with "WMF". This is just as an indication from callers in the app-side code (like ExploreViewController) that this component lives in one of the packages/frameworks.

WMFComponents/Sources/WMFComponents/Components/Profile/ProfileViewController.swift
```diff
@@ -0,0 +1,25 @@
+import SwiftUI
+import UIKit
+
+class ProfileViewController: UIViewController {
```
tonisevener, line 4: Similar note here - name this "WMFProfileViewController" since it lives in WMFComponents.

tonisevener (changes requested):
Thanks for getting this up @l-olson1214! I will get into more granular reviewing in your followup PRs. For now I just have some high level feedback on file placement and names. Let me know if you have any questions or want to pair.

WMFComponents/Sources/WMFComponents/Components/Profile/ProfileView.swift
l-olson1214: Can do!

Wikipedia/Code/ProfileView.swift
l-olson1214: Good catch!

Wikipedia/Code/ExploreViewController.swift
```diff
@@ -192,11 +200,20 @@ class ExploreViewController: ColumnarCollectionViewController, ExploreCardViewCo
...
+        let profileView = WMFProfileView(isLoggedIn: true)
+        let hostingController = UIHostingController(rootView: profileView)
+        hostingController.modalPresentationStyle = .pageSheet
+
+        if let sheetPresentationController = hostingController.sheetPresentationController {
+            sheetPresentationController.detents = [.large()]
+            sheetPresentationController.prefersGrabberVisible = true
+        }
 
+        present(hostingController, animated: true, completion: nil)
```
tonisevener, line 216: @l-olson1214 One more thing - I notice WMFProfileViewController isn't used yet. You could replace these lines with this to use it:

```
let profileViewController = WMFProfileViewController()
present(profileViewController, animated: true, completion: nil)
```

WMFComponents/Package.swift
```diff
@@ -5,7 +5,7 @@ import PackageDescription
 
 let package = Package(
     name: "WMFComponents",
-    platforms: [.iOS(.v15)],
+    platforms: [.iOS(.v16)],
```
tonisevener, line 8: Good catch on this!

tonisevener (approved):

PR #4974, opened 2024-09-18
Fix Core Data threading complaints

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T375091

### Notes
This fixes a couple of threading complaints thrown when setting the `-com.apple.CoreData.ConcurrencyDebug 1` run argument in the Wikipedia scheme. My hope is that this fixes some of our Core Data crashes that we've been seeing.

### Test Steps
1. Regression test the Suggested Edits card appearance.
2. Go to Notifications, confirm it loads. Log out of the app and log into a different account. Confirm notifications still loads.

### DIFF
--- a/Wikipedia/Code/WMFSuggestedEditsContentSource.m
+++ b/Wikipedia/Code/WMFSuggestedEditsContentSource.m
@@ -24,42 +24,45 @@
 
 - (void)loadNewContentInManagedObjectContext:(nonnull NSManagedObjectContext *)moc force:(BOOL)force completion:(nullable dispatch_block_t)completion {
 
-    // First delete old card
-    [self removeAllContentInManagedObjectContext:moc];
-
-    NSURL *appLanguageSiteURL = self.dataStore.languageLinkController.appLanguage.siteURL;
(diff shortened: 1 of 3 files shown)


Wikipedia/Code/WMFSuggestedEditsContentSource.m
```diff
@@ -24,42 +24,45 @@ - (instancetype)initWithDataStore:(MWKDataStore *)dataStore {
...
-    }
+            return;
+        }
+        
+        if (!authManager.authStateIsPermanent) {
+            completion();
+            return;
+        }
+            
+        WMFCurrentUser *user = [self.dataStore.authenticationManager userWithSiteURL:appLanguageSiteURL];
```
mazevedofs, line 44: On a fresh install, the user returns nil when I change language at this point. If I log out and then log in again, the user is successfully retrieved, and then the card appears. 
What I did:
- Fresh install
- App auto login happens.
- English as primary language - user is not nil
- Add PT as primary language (I've also tested with other languages that are on the alt text experiment, such as FR and ES)
- User returns nil from authenticationManager userWithSiteURL

Let me know if you reproduce that.

PR #4977, opened 2024-09-18
Profile/Explore nav fixes

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T374162

### Notes
These were a couple of bugs that I noticed when testing profile:

1. The profile icon badge isn't always in sync with having unread notifications. When I change a read state in Notifications Center, the bell icon will show a badge, but the profile icon does not. I found a few more places in the codebase where the Explore bell icon update was called. I made sure to call the profile update method in the same place.
2. According to task description, we should hide the notifications center bell icon. I commented the icon out in this PR. I'd rather not fully deleting the code just yet until we're sure we won't have to bring it back as a part of https://phabricator.wikimedia.org/T374161 (A/B test control experience).

Note that some of this diff includes automatic linter changes.

### Test Steps
1. Confirm notifications bell icon is hidden.
2. We can't really test the badge update until our Notifications Center navigation is in place.

### Screenshots/Videos
Video of bug:

https://github.com/user-attachments/assets/879261b5-355c-4bb5-8241-f7f245d828db

### DIFF
--- a/Wikipedia/Code/ExploreViewController.swift
+++ b/Wikipedia/Code/ExploreViewController.swift
@@ -28,7 +28,7 @@ class ExploreViewController: ColumnarCollectionViewController, ExploreCardViewCo
         navigationBar.shouldTransformUnderBarViewWithBar = true
         navigationBar.isShadowHidingEnabled = true
 
-        //updateNotificationsCenterButton()
+        // updateNotificationsCenterButton()
         updateProfileViewButton()
         updateNavigationBarVisibility()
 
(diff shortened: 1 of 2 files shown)


Wikipedia/Code/ExploreViewController.swift
```diff
@@ -28,7 +28,7 @@ class ExploreViewController: ColumnarCollectionViewController, ExploreCardViewCo
         navigationBar.shouldTransformUnderBarViewWithBar = true
         navigationBar.isShadowHidingEnabled = true
 
-        updateNotificationsCenterButton()
+        // updateNotificationsCenterButton()
```
l-olson1214, line 31: Here and elsewhere, did you want to keep these commented out for future or remove them?

l-olson1214 (approved):

PR #4983, opened 2024-09-20
Implement Coordinator-based navigation for donate

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T374166 (Donate link)

### Notes
This PR adds navigation into the donate flow from the profile view. Here is a high-level list of the logic:

1. The ProfileListItem model struct now has an `isDonate` and `isLoadingDonateConfigs` flag. When a profile item view is created, if the model indicates it is a donate item, I am storing that button's frame into a `targetRects` environment object on WMFProfileView. That environment object is instantiated and owned by the ProfileCoordinator. I am also checking the `isLoadingDonateConfigs` flag to determine if a spinner should appear when the profile item view is created.
2. From ProfileCoordinator, when the donate action is triggered, I create and handle the navigation logic in a DonateCoordinator.
3. DonateCoordinator takes in the target rect from ProfileCoordinator, as well as a block for setting the spinner.
4. Upon start(), the DonateCoordinator fetches the donate configs from the network. As it does this it flips the flag for showing the spinner in the donate profile item view.
5. From the config response, the donate coordinator determines if the user needs to see the payment method choice action sheet. If it does, it uses the targetRect passed in from the profile coordinator for displaying the action sheet. This target rect is required for the iPad popover arrow to point in the correct location. For iPhone, the target rect is not needed.
6. From the action sheet, if you tap "Donate Apple Pay", the profile dismisses and the native Apple Pay form is pushed on. If you tap "Other payment method", the profile dismisses and the donate web view is pushed on. This logic is determined in the DonateCoordinator.
7. If on step 6 the donate coordinator determines the user cannot display the native Apple Pay form, the action sheet is skipped, profile dismisses and the donate web view is pushed on immediately.
8. While I was here, I refactored the article campaign modal presentation flow to use DonateCoordinator.

### Test Steps
1. Fresh install the app on a device that has Apple Pay set up in Wallet. Tap profile from Explore tab.
2. Confirm the payment method action sheet is presented when tapping donate. Confirm you are navigated to the native Apple Pay form or the web view depending on your answer.

_Optional steps, we can lean on QA for these_
4. (We can lean on QA for this) Donate a small amount, confirm payment goes through and navigation pops you back to Explore.
5. (We can lean on QA for this) Temporarily comment out this date logic in `WMFFundraisingCampaignDataController`:

```
//            guard (firstAsset.startDate...firstAsset.endDate).contains(currentDate) else {
//                return
//            }
```
and
```
//            guard (startDate...endDate).contains(currentDate) else {
//                return
//            }
```

6. Change app region to Spain in iOS Settings.
7. Fresh install app. On Explore, background, then foreground the app. Pull to refresh on Explore to fetch campaign data. Tap on an EN Wikipedia article. Note, you may need to visit an article twice for the campaign modal to appear.
8. You should see a campaign modal. Tap "Donate". Confirm you see action sheet for payment methods after tapping. Tap one of the options and confirm the correct view is pushed on (native Apple Pay form or web view).
9. (We can lean on QA for this) Donate a small amount, (my native payment didn't work here, but could be due to the app region code or outdated bannerID). Confirm navigation pops you back to Article if payment goes through.


Wikipedia/Code/ProfileCoordinator.swift
```diff
@@ -104,8 +117,21 @@ class ProfileCoordinator: NSObject, Coordinator, ProfileCoordinatorDelegate {
...
+        }
+        
+        let donateCoordinator = DonateCoordinator(navigationController: navigationController, donateButtonGlobalRect: targetRects.donateButtonFrame, source: donateSouce, dataStore: dataStore, theme: theme, setLoadingBlock: { isLoading in
+            viewModel.isLoadingDonateConfigs = isLoading
+        })
+        
+        donateCoordinator.start()
+        
+        // Note: DonateCoordinator needs to handle a lot of delayed logic (fetch configs, present payment method action sheet, present native donate form and handle delegate callbacks from native donate form) as opposed to a fleeting navigation call with the other actions. For this reason we need to save it in a property so it isn't deallocated before this logic runs.
+        self.donateCoordinator = donateCoordinator
```
mazevedofs, line 132: If, in the future, we need to do this treatment to more coordinators, it could be worth doing "child coordinators." Here it works really well.

mazevedofs (approved):

PR #4986, opened 2024-09-23
Deep Link support added for places

raihan:
#### Background Context:
I have added deep link support for places in the Wikipedia app. So, when the Wikipedia app is opened through the deep link, it will directly go to the ‘Places’ tab and show the location specified by the caller app.

#### In this PR:
* Added deep link support for location coordinates (latitude & longitude) and name.
* Used existing logic to search places with coordinates.
* Used existing logic to search places with names.

#### Testing Steps:
Please use the following deep link patterns 
* With coordinates: `wikipedia://places?WMFArticleURL=https://en.wikipedia.org/wiki?place=Amsterdam&latitude=52.3547498&longitude=4.8339215`
* With location name: `wikipedia://places?WMFArticleURL=https://en.wikipedia.org/wiki?place=Berlin`

#### Screenshot/demo:
https://github.com/user-attachments/assets/3546f134-7a07-4e86-ab53-19cc5c77a07e


Wikipedia/Code/WMFAppViewController.m
```diff
@@ -1215,9 +1215,21 @@ - (BOOL)processUserActivity:(NSUserActivity *)activity animated:(BOOL)animated c
...
-                // For "View on a map" action to succeed, view mode has to be set to map.
-                [[self placesViewController] updateViewModeToMap];
-                [[self placesViewController] showArticleURL:articleURL];
+                double latitude = [[articleURL wmf_valueForQueryKey:@"latitude"] doubleValue];
+                double longitude = [[articleURL wmf_valueForQueryKey:@"longitude"] doubleValue];
+                NSString *place = [articleURL wmf_valueForQueryKey:@"place"];
+
+                if (latitude && longitude) {
+                    // If the place coordinate is provided in the URL, show it in the map
+                    [[self placesViewController] searchPlaceWithCoordinateWithLatitude:latitude longitude:longitude];
```
tonisevener, line 1224: I am not able to hit this condition - I think the latitude and longitude values are getting dropped somewhere.

Wikipedia/Code/WMFAppViewController.m
```diff
@@ -1215,9 +1215,21 @@ - (BOOL)processUserActivity:(NSUserActivity *)activity animated:(BOOL)animated c
             [self.currentTabNavigationController popToRootViewControllerAnimated:animated];
             NSURL *articleURL = activity.wmf_linkURL;
             if (articleURL) {
```
tonisevener, line 1217: (Optional)

This may be tricky to fix, but I would prefer if we removed the whole `WMFArticleURL=` part of these places deep links:

```
wikipedia://places?WMFArticleURL=https://en.wikipedia.org/wiki?latitude=52.3547498&longitude=4.8339215
wikipedia://places?WMFArticleURL=https://en.wikipedia.org/wiki?place=Berlin
```

It is unnecessary information, and indicates we are searching against the subdomain wiki (`en.wikipedia.org`), when in reality the app searches against the primary app language wiki in Settings.

If you could find a way to get these two links to work instead:

```
wikipedia://places?latitude=52.3547498&longitude=4.8339215
wikipedia://places?place=Berlin
```

I think that would feel better.

tonisevener (changes requested):
@raihan Thanks for the improvement! I have one issue with getting longitude and latitude to work. If you have another deep link I could test with let me know.

mazevedofs:
Hi @raihan, just checking in to see if you were able to take a look at Toni's comments. Let us know if you need assistance.

tonisevener:
Hi @raihan, I'm going to close this to clear up our board. Feel free to reopen with the feedback addressed and we will take another look. Thank you!

PR #4992, opened 2024-09-27
T375416 add an image

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T375416

### Notes
* 

### Test Steps
1. On an account with more than 50 edits (or go into the code and change it) - 
2. Set language to one of the target languages ["es", "pt", "fr", "zh"]
3. Fresh install
4. Launch app, go to a different tab, return to explore
5. See pop up for add an image
6. Navigate to a different tab, return to explore
7. See pop up for add an image again

### DIFF
--- a/WMFData/Sources/WMFData/Data Controllers/Image Recommendations/WMFImageRecommendationsDataController.swift
+++ b/WMFData/Sources/WMFData/Data Controllers/Image Recommendations/WMFImageRecommendationsDataController.swift
@@ -9,10 +9,10 @@ public class WMFImageRecommendationsDataController {
 		var hasPresentedOnboardingModal: Bool
         var hasPresentedOnboardingTooltips: Bool
         var hasPresentedFeatureAnnouncementModal: Bool
-        var hasPresentedAddAnImageFeatureAnnouncemnt: Bool
+        var hasPresentedFeatureAnnouncementModalAgainForAltTextTargetWikis: Bool
 
 		static var `default`: OnboardingStatus {
-            return OnboardingStatus(hasPresentedOnboardingModal: false, hasPresentedOnboardingTooltips: false, hasPresentedFeatureAnnouncementModal: false, hasPresentedAddAnImageFeatureAnnouncemnt: false)
+            return OnboardingStatus(hasPresentedOnboardingModal: false, hasPresentedOnboardingTooltips: false, hasPresentedFeatureAnnouncementModal: false, hasPresentedFeatureAnnouncementModalAgainForAltTextTargetWikis: false)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


Wikipedia/Code/ExploreViewController.swift
```diff
@@ -1242,6 +1312,12 @@ extension ExploreViewController {
         if !UIAccessibility.isVoiceOverRunning {
             presentImageRecommendationsFeatureAnnouncementIfNeeded()
         }
+        
+        let imageRecommendationsDataController = WMFImageRecommendationsDataController()
+        
+        if imageRecommendationsDataController.hasPresentedFeatureAnnouncementModal {
+            presentImageRecommendationsAnnouncementAltText()
+        }
```
tonisevener, line 1320: @l-olson1214 Can you move this bit up into the !isVoiceOverRunning conditional?

WMFData/Sources/WMFData/Data Controllers/Image Recommendations/WMFImageRecommendationsDataController.swift
```diff
@@ -9,9 +9,10 @@ public class WMFImageRecommendationsDataController {
 		var hasPresentedOnboardingModal: Bool
         var hasPresentedOnboardingTooltips: Bool
         var hasPresentedFeatureAnnouncementModal: Bool
+        var hasPresentedFeatureAnnouncementModalAgainForAltTextTargetWikis: Bool
```
tonisevener, line 12: I found an interesting bug. I tried an alternative test by running on the 7.5.9 release first, triggering the feature announcement modal, then running again from this branch. When I do those steps, it was presenting the feature announcement a _third_ time when revisiting the Explore feed, when really it should only happen twice: displayed once when running from 7.5.9, then a 2nd time in this branch for alt text wiki folks based on this `hasPresentedFeatureAnnouncementModalAgainForAltTextTargetWikis` flag.

I think what is happening is that because we are adding this property to the struct, the old persistence is failing to deserialize, and we lose the status struct from 7.5.9 entirely. So when running from this branch it's starting fresh and presenting the announcement modal two more times.

One small thing that seems to work is changing this type to a Bool?. The optional allows deserialization to default this property false if the old persisted struct doesn't contain this property, rather than failing entirely.

Here are the test steps:
1. Fresh install 7.5.9 from TestFlight on device.
2. In onboarding, choose Spanish Wikipedia as your primary language.
3. If feature announcement hasn't displayed already, ensure you are logged in, maybe wait a bit to ensure the Suggested Edits card appears. Leave the tab and come back to Explore to trigger feature announcement.
4. Now run on device from this PR branch. Feature announcement should pop up _again_ when you land on Explore.
5. Leave the tab then come back to Explore. Feature announcement should _not_ show anymore.

tonisevener (commented):
@l-olson1214 Just a couple of little things, thanks!

Wikipedia/Code/ExploreViewController.swift
l-olson1214: Got it!

WMFData/Sources/WMFData/Data Controllers/Image Recommendations/WMFImageRecommendationsDataController.swift
l-olson1214: Awesome catch! Fixed :)

tonisevener (approved):

PR #4996, opened 2024-09-30
Fix Core Data threading complaints + Auth suggested edits card appearance

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T375091

### Notes
This is a repeat PR of https://github.com/wikimedia/wikipedia-ios/pull/4974 with the PR feedback addressed.

### Test Steps
1. Regression test the Suggested Edits card appearance.
2. Go to Notifications, confirm it loads. Log out of the app and log into a different account. Confirm notifications still loads.

af85b00528d61ce4bfe2ceee2709ce70069c59ca - this commit fixes the regression found in the PR feedback on https://github.com/wikimedia/wikipedia-ios/pull/4974. Now when changing the app primary language, the authentication manager will fetch the user's information against their app primary language and cache it. Then the Explore feed will refresh and re-evaluate the suggested edits card visibility.

### DIFF
--- a/WMF Framework/Remote Notifications/Model/RemoteNotificationsModelController.swift
+++ b/WMF Framework/Remote Notifications/Model/RemoteNotificationsModelController.swift
@@ -113,7 +113,7 @@ final class RemoteNotificationsModelController {
         try FileManager.default.removeItem(at: legecyJournalWalUrl)
     }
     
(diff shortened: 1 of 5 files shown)


mazevedofs (approved):
Working great! The suggested edits card appearance is even smoother than before (I usually had to refresh the explore feed manually)

PR #4999, opened 2024-10-01
Add Core Data stack to WMFData and persist page views

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T370216

### Notes
This PR sets up a new Core Data stack in WMFData and persists page views. There are several high-level pieces here:

1. The Core Data stack is set up [here](https://github.com/wikimedia/wikipedia-ios/compare/viewed-articles-coredata?expand=1#diff-53892ba35877905bfabb1572472302074e7251ac44ee8ac623d2b0be877d4953R886) after the old data store has finished it's setup.
2. During the first launch, in this [data migration area](https://github.com/wikimedia/wikipedia-ios/compare/viewed-articles-coredata?expand=1#diff-9c1c83ec9a4c64f14828322c5f3b7e271f466ad8e71bb6fa0f948aea01af2b3aR514), I am migrating all old article views into the new page views table. This migration is handy so that we will only have one data store to reference when determining their most-viewed article. We won't have to reference the old app-side core data stack.
3. When an article has finished loading, I am adding a page view record [here](https://github.com/wikimedia/wikipedia-ios/compare/viewed-articles-coredata?expand=1#diff-9f5148d285fb05e36234bacc9dc8faff0de5a47dda466a0e5f9bd440936028ebR77).
4. I updated `HistoryViewController` to clear all page views or individual page views when things are deleted on the History screen.
5. I created a housekeeping method for WMFData, called in the same [location](https://github.com/wikimedia/wikipedia-ios/compare/viewed-articles-coredata?expand=1#diff-53892ba35877905bfabb1572472302074e7251ac44ee8ac623d2b0be877d4953R650) as the other housekeeping calls, which clears out any page or pageviews in WMFData older than January 1st, one year ago (2023 as of today). 

One area I want to highlight is WMFData's Core Data model file. I am setting up the entity names as WMFPage and WMFPageView, but their underlying class names are CDPage and CDPageView. This may feel a bit strange, but this is the recommended pattern as displayed in the https://developer.apple.com/documentation/coredata/adopting_swiftdata_for_a_core_data_app Coexistence sample app. In the future when our deployment target is iOS17+, we should be able add `@Model` to WMFPage and WMFPageView, and it will still pull from this underlying data store.

### Test Steps
1. In Simulator, fresh install on 7.6.0 branch. Tap to visit several articles. Open location of "Simulator container directory" printed out in the console.
2. Switch to this branch. Confirm app launches. You should see a WMFData.sqlite file in the simulator container directory. Open file (I use Liya). You should see a `ZWMFPAGEVIEW` table and a `ZWMFPAGE` table. Each table should show imported data from the 7.6.0 branch views. Confirm the `ZWMFPAGEVIEW` has a populated `ZWMFPAGE` column to confirm the connection worked.
3. Visit one of the articles again. Confirm an additional page view record appears in the `ZWMFPAGEVIEW` table, so now you have multiple page views associated with the same `ZWMFPAGE`.
4. Go to History tab. Swipe to delete one of the article views. Confirm article data disappears from both `ZWMFPAGEVIEW` and `ZWMFPAGE` tables.
5. Tap "Clear" button at the top of history. Confirm all article data disappears from both `ZWMFPAGEVIEW` and `ZWMFPAGE` tables.


mazevedofs (approved):
Looks great! There's just a warning for the "sevenDaysAgoDate" prop in the WMFCoreDataStore, but that is very minor

PR #5015, opened 2024-10-10
T376044 paging components

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T376044

### Notes
* Translations are waiting for any slides with data in them until we get the finalized numbers.

### Test Steps
Open build #173 (or manually uncomment out the profile entry point)
Navigate to Year in Review entry point
Begin slideshow
Test that next will bring you to the next slide, done will close
Note: data is hardcoded and not translated yet, we don't want to cause double translations

### Screenshots/Videos

https://github.com/user-attachments/assets/dd5ae31f-01d7-4d6e-b039-e5046d3acc86


tonisevener (approved):
Looks great, thanks! I pushed a minor tweak to your code comments - the build script was still exporting those translations so I changed it to WMFFakeLocalizedString to prevent that from happening.

PR #5016, opened 2024-10-10
Add YiR dev settings

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T376041 (partially)

### Notes
* Adds YiR dev settings
* Not wired into any flows at the moment

### Test Steps
1. Go to dev settings, make sure you can see the toggle


WMFData/Sources/WMFData/Data Controllers/YearInReview/YearInReviewDataController.swift
```diff
@@ -0,0 +1,5 @@
+import Foundation
+
+final public class YearInReviewDataController {
```
tonisevener, line 3: @mazevedofs I actually do want this empty class here, just in case we need to hook into it soon, thanks. Can you rename to WMFYearInReviewDataController?

tonisevener (approved):

PR #5017, opened 2024-10-11
Fix retain cycle in ArticleViewController

spatel:
**Phabricator:** 
https://phabricator.wikimedia.org/T374331

### Notes
* I am not 100% certain this is the root cause of the above issue, but it's more of an educated guess.
* I think there's a memory leak in `ArticleViewController` (if you put a breakpoint in the `deinit` method [here](https://github.com/wikimedia/wikipedia-ios/blob/07ed2c7fb8de51005cfd57a66e3b7b322705e702/Wikipedia/Code/ArticleViewController.swift#L165), you'll see that it never gets called even after you dismiss the article). So, for example, if you browse through 10 random articles (then tap back 10 times), all 10 articles are still in memory.
* A quick check of the memory graph demonstrates this:
Screenshot 2024-10-11 at 15 13 25

* The reason I suspect the image loading could perhaps just be due to a memory issue is because I *was* eventually able to reproduce the stated behaviour on 2 very different devices - an older iPad Pro (10.5 inch, 1st gen) and iPhone 15 Pro. After tapping back and forth through random articles, I observed images failing to load after approx. 50 articles on iPad but on iPhone 15 Pro, it was closer to 100 articles.
* Even if this isn't the root cause of the issue, it's nice to fix the memory leak anyway 😀

### Test Steps
N.B. See notes above, it's VERY difficult to reproduce this consistently:
1. Go to a random article and tap the randomiser button
2. Repeat step 1 **a LOT of times**
3. Eventually images may stop loading or the app may crash

### Screenshots/Videos


Wikipedia/Code/ArticleViewController.swift
```diff
@@ -152,7 +152,9 @@ class ArticleViewController: ViewController, HintPresenting {
 
         super.init(theme: theme)
         
-        self.schemeHandler.imageDidSuccessfullyLoad = imageDidSuccessfullyLoad
```
spatel, line 155: There's a retain cycle here is because:

`self.schemeHandler.imageDidSuccessfullyLoad = imageDidSuccessfullyLoad`

is basically equivalent to:

`self.schemeHandler.imageDidSuccessfullyLoad = { self.imageDidSuccessfullyLoad() }`

In other words, strong `self` is being implicitly captured, which creates the retain cycle. It's a real shame though because I generally prefer the shorter syntax too (when it doesn't create a retain cycle 😀).

mazevedofs (approved):
Hi @spatel! Thanks for taking a look at this. While I'm also not sure what is the root cause of the image loading issue, your fix did solve the retain cycle issue: I no longer see WMFRandomArticleViewController objects on the stack after dismissing them.

PR #5025, opened 2024-10-18
[YiR] Feature announcement

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T376316

### Notes
* This PR adds logic for presenting the YiR feature
* It timeboxes the presentation to a deadline, code to be removed after this date
* For presenting the iPad popover pointing to the profile button, we're creating a CGRect and not using the UIBArButtonItem. That was causing a crash, probably related to our custom nav bar

### Test Steps
1. Set your device to a valid region (Italy or France)
2. Fresh install the app and login
3. Set the language to Italian or French
4. Read articles and generate 6+ article views in the History tab.
5. Go to developer settings and toggle on Year in Review feature flag.
6. Run the app and background it on the explore feed
7. Open the app again. You should see the modal. Tapping the "Continue" button should take you to the feature
8. Fresh install the app again, repeat the previous steps
9. Go to an article
10. You should see the modal
11. Go back to another article - it should not appear again
12. Repeat all the steps on iPad


Wikipedia/Code/ExploreViewController.swift
```diff
@@ -1070,6 +1113,27 @@ struct ImageRecommendationsFeatureAnnouncementTimeBox {
     }
 }
 
+// MARK: Year In Review Time-box
+// TODO: Remove after expiry date (1 March 2025)
+
+struct YearInReviewFeatureAnnouncementTimeBox {
```
tonisevener, line 1119: Maybe we should put this in the WMFYearInReviewDataController?

Wikipedia/Code/ExploreViewController.swift
```diff
@@ -927,7 +905,71 @@ class ExploreViewController: ColumnarCollectionViewController, ExploreCardViewCo
...
+        let languages = ["fr", "it"]
+        guard YearInReviewFeatureAnnouncementTimeBox.isAnnouncementActive() else {
+            return
+        }
+
+        guard let appLanguage = dataStore.languageLinkController.appLanguage else {
+            return
+        }
+
+        guard languages.contains(appLanguage.languageCode) else {
```
tonisevener, line 925: I think `shouldShowYearInReviewEntryPoint` below should already cover checking for this.

Wikipedia/Code/ExploreViewController.swift
```diff
@@ -927,7 +905,71 @@ class ExploreViewController: ColumnarCollectionViewController, ExploreCardViewCo
...
+        guard let yirDataController = try? WMFYearInReviewDataController() else {
+            return
+        }
+
+        let project = WMFProject.wikipedia(WMFLanguage(languageCode: appLanguage.languageCode, languageVariantCode: nil))
+        guard yirDataController.shouldShowYearInReviewEntryPoint(countryCode: Locale.current.region?.identifier, primaryAppLanguageProject: project) else {
+            return
+        }
+
+        guard !yirDataController.hasPresentedYiRFeatureAnnouncementModel else {
```
tonisevener, line 938: We might be able to go a bit further and create a `shouldShowYearInReviewFeatureAnnouncement` method in the data controller, which references this flag, the `YearInReviewFeatureAnnouncementTimeBox` value and the `shouldShowYearInReviewEntryPoint` response. Just to try reduce the amount of YiR code that has to live in ExploreViewController.

Wikipedia/Code/ExploreViewController.swift
```diff
@@ -1319,13 +1383,16 @@ extension ExploreViewController {
 
     @objc func applicationDidBecomeActive() {
         if !UIAccessibility.isVoiceOverRunning {
+            presentYearInReviewAnnouncement()
```
tonisevener, line 1386: I worry about this line and the next interacting with each other, or causing two feature announcement presentations back-to-back for different features. If you move a boolean method into the data controller maybe you could do something here like:

```
if yirDataController.shouldShowYearInReviewFeatureAnnouncement() {
   presentYearInReviewAnnouncement()
} else {
   presentImageRecommendationsFeatureAnnouncementIfNeeded
}
```

Also note that we probably want this one to show if VoiceOver is running.

This call presents it after foregrounding, but Explore's `viewDidAppear` also calls `presentImageRecommendationsFeatureAnnouncementIfNeeded`. Do we need `presentYearInReviewAnnouncement` there as well?

Wikipedia/Code/ArticleViewController.swift
```diff
@@ -459,6 +460,7 @@ class ArticleViewController: ViewController, HintPresenting {
             return
         }
         showAnnouncementIfNeeded()
```
tonisevener, line 462: Careful with this area - this line will call the fundraising banner. I'm sure we don't want the YiR feature announcement to mess with it's presentation. You should only present YiR feature announcement if a fundraising banner will not present, IMO (be sure to confirm that with Product).

tonisevener (changes requested):
Looking good! I'll pull and test after the next round of code changes.

WMFData/Sources/WMFData/Data Controllers/Year In Review/WMFYearInReviewDataController.swift
```diff
@@ -2,23 +2,82 @@ import Foundation
...
+        guard isAnnouncementActive() else {
+            return false
+        }
+
+
+        guard shouldShowYearInReviewEntryPoint(countryCode: Locale.current.region?.identifier, primaryAppLanguageProject: primaryAppLanguageProject) else {
+            return false
+        }
+
+        guard hasPresentedYiRFeatureAnnouncementModel else {
```
tonisevener, line 74: I had to switch this to `!hasPresentedYiRFeatureAnnouncementModel` in order to get it to present.

tonisevener (changes requested):
Looks great! A few more things found after testing.

I also notice the "Done" button doesn't dismiss when the YiR flow is presented from this announcement Continue button.

tonisevener (approved):
Working great!

PR #5027, opened 2024-10-21

translatewiki:
Translation updates


tonisevener:
This was a failure in [testAllSupportedTranslatedLanguagesWereAddedToProjectLocalizations](https://github.com/wikimedia/wikipedia-ios/blob/main/WikipediaUnitTests/Code/TWNStringsTests.m#L357), which has some good fix steps listed as a comment. What I had to do:

1. In project editor Info tab, remove "Azerbaijani" localization option (which said 0 files localized).
2. In "Wikipedia > iOS Native Localizations", delete az.lproj subdirectory.
3. In project editor > Info tab, add back in "Azerbaijani" as a localization. Keep InfoPlist.strings, Localizable.strings, and Localizable.stringsdict (Reference Language English) files checked when prompted.

The other test fixes were more self-explanatory. One new AZ translation used html tags, and two more added extra curly braces. Fixed both here and in Translatewiki.

tonisevener (approved):

PR #5030, opened 2024-10-23
T375320 

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T376320#10219967

### Notes
* 

### Test Steps
1. Put app in one of the accepted locations (Italy, Mexico) and an accepted language (Italian, Spanish, or English)
2. Get an account with edits (or hardcode it for testing) (test more than 500 for 500+, 0 for no personalized slide, or any amount in between to see regular personalized slide) 
3. Go through slideshow as normal, comparing the third slide to desired outcome

### Screenshots/Videos
Simulator Screenshot - iPhone 15 Plus - 2024-10-23 at 13 59 39

PR #5033, opened 2024-10-25
T376320 Clean up, optionals

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T376320

### Notes
* Clean up from https://github.com/wikimedia/wikipedia-ios/pull/5030

### Test Steps
1.

### DIFF
--- a/WMFData/Sources/WMFData/Data Controllers/Year In Review/WMFYearInReviewDataController.swift
+++ b/WMFData/Sources/WMFData/Data Controllers/Year In Review/WMFYearInReviewDataController.swift
@@ -180,7 +180,7 @@ public class WMFYearInReviewDataController {
     }
     
     @discardableResult
-    public func populateYearInReviewReportData(for year: Int, countryCode: String, primaryAppLanguageProject: WMFProject?, username: String) async throws -> WMFYearInReviewReport? {
+    public func populateYearInReviewReportData(for year: Int, countryCode: String, primaryAppLanguageProject: WMFProject?, username: String?) async throws -> WMFYearInReviewReport? {
         
         guard shouldPopulateYearInReviewReportData(countryCode: countryCode, primaryAppLanguageProject: primaryAppLanguageProject) else {
             return nil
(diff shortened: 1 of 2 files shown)


WMFData/Sources/WMFData/Data Controllers/Year In Review/WMFYearInReviewDataController.swift
```diff
@@ -555,19 +554,19 @@ public class WMFYearInReviewDataController {
...
-        
+
         guard let project = project else {
             completion(.failure(WMFDataControllerError.mediaWikiServiceUnavailable))
             return
         }
         
-        let ucStartDate = "2024-11-01T22:20:13.000Z"
-        let ucEndDate = "2024-01-01T22:20:13.000Z"
+        let ucStartDate = startDate
```
tonisevener, line 568: Note the API accepts these backwards - use `endDate` here and `startDate` on the next line.

WMFData/Sources/WMFData/Data Controllers/Year In Review/WMFYearInReviewDataController.swift
```diff
@@ -205,16 +205,18 @@ public class WMFYearInReviewDataController {
         }
         
         if result.needsEditingPopulation == true {
-            let edits = try await fetchEditCount(username: username, project: primaryAppLanguageProject)
-            try await backgroundContext.perform { [weak self] in
-                try self?.populateEditingSlide(edits: edits, report: report, backgroundContext: backgroundContext)
+            if let username = username {
```
tonisevener, line 208: You can shorten this to `if let username`.

tonisevener (changes requested):

WMFData/Sources/WMFData/Data Controllers/Year In Review/WMFYearInReviewDataController.swift
tonisevener: Please flip these like this, otherwise the API is not returning results for me. It's counter-intuitive (IMO) but the API expects the later date to be added to the `ucstart` parameter.

```
let ucStartDate = endDate
let ucEndDate = startDate
```

tonisevener (approved):

PR #5037, opened 2024-10-29
T376073 donate slide: pt1

l-olson1214:
**Phabricator:** 

### Notes
* Created switches between hasDonated/hasn't, isEdgeSlide, etc.

### Test Steps
1.


tonisevener (changes requested):
This is dipping a bit into early design review, which is not my area, but I want to avoid reverting these changes if I merge now.

My thinking is that this is final view should be a part of `WMFSlideShow`, so that this slide represents a new dot, the dots remain visible (only Share and Next might disappear) and a user can swipe to and swipe back from it. Right now if I swipe to it, the dots disappear and it rather abruptly changes to the final view:

https://github.com/user-attachments/assets/794d813c-5407-4f0d-8cd6-466121c61eda

I'll wait on more granular feedback in case this causes a bit of rework. Also feel free to check with design if you need clarity here, I may have missed some back & forth agreements between y'all.

PR #5040, opened 2024-10-30
T378491: Update Slides

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T378491

### Notes
* Numbers may be wrong, they will be updated before launch

### Test Steps
1. Navigate to YIR (usual set up to enable)
2. Scroll through slides, ensure each one has a gradient, and the proper item associated with it (image vs text)

### Screenshots/Videos

https://github.com/user-attachments/assets/cb475213-22ad-4b1e-a695-9068abb81e67


mazevedofs (changes requested):
Looks great! You can delete the old YiR illustrations since we won't need it.

mazevedofs (approved):

PR #5045, opened 2024-11-01
Invalid widget cache when fetching feature content failed.

ngn999:
**Phabricator:** 

### Notes
* When `widgetCache.featureContent.topRead` is nil, Top Read widget always get placeholder content.
The Featured Article widget and Picture of the Day widget have similar problems.
### Test Steps
1.


tonisevener:
Hi @ngn999!

Thank you for the PR. Can you describe the bug you are seeing in more detail, and test steps demonstrating how this PR fixes it? We may want to lean on the last successful cached content in the case of a failure, to smooth over server-side delays in data.

ngn999:
> Hi @ngn999!
> 
> Thank you for the PR. Can you describe the bug you are seeing in more detail, and test steps demonstrating how this PR fixes it? We may want to lean on the last successful cached content in the case of a failure, to smooth over server-side delays in data.

After closing the app for a couple of hours, the cache will become invalid, and a placeholder like this will be displayed:

tonisevener (approved):
Thanks for the PR!

I think it makes sense for us to clear the cache when there's invalid content. Potentially, if we return a failure in these completions, the widget will try to refresh again in [two hours](https://github.com/wikimedia/wikipedia-ios/blob/c43d37daae7f3fc8418a7ef816625d0422477e9d/Widgets/Widgets/TopReadWidget.swift#L127). Upon refresh it will [check the cache](https://github.com/wikimedia/wikipedia-ios/blob/c43d37daae7f3fc8418a7ef816625d0422477e9d/WMF%20Framework/Widget/WidgetController.swift#L250C53-L250C77), pull the invalid content and become stuck on this blank state.

I could not enter a situation where it gets stuck on a blank top read or featured article, but I was able to see an improvement for picture of the day.

PR #5044, opened 2024-10-31
Improve Core Data error handling in WMFData tests

tonisevener:
**Phabricator:** N/A

### Notes
This PR attempts to improve our flaky Core Data tests in WMFData. [For example](https://github.com/wikimedia/wikipedia-ios/runs/32358820485).

### Test Steps
1. Ensure unit tests pass

PR #5059, opened 2024-11-07
[YiR] Fix Explore profile button

tonisevener:
**Phabricator:** N/A

### Notes
This PR fixes a race condition that I think only appeared once https://github.com/wikimedia/wikipedia-ios/pull/5044 and https://github.com/wikimedia/wikipedia-ios/pull/5056 were merged.

I noticed that ExploreViewController's lazy yirCoordinator was stuck with nil, because it's `yirDataController` would throw a `coreDataStoreUnavailable` error upon instantiation. Likely when a YearInReviewDataController is instantiated early on in the app lifecycle, the core data store isn't ready yet. Adjusting these a bit to make the `yirDataController` computed on Explore and Article (so that it tries again each time), and relaxing the laziness a bit for `yirCoordinator`, seems to help.

### Test Steps
1. Fresh install app in correct device region with. correct device language. Confirm Explore profile button tap works.
2. Enable YiR feature flag. Background, foreground. Leave Explore tab and come back. Confirm feature announcement and badge appear. Confirm tapping into intro slide clears out badge as it did before. Confirm survey still presents after the last slide.
3. Repeat steps 1-2 for Article view.


mazevedofs (approved):
Looks good! This bug makes me wonder if we might need to rethink our coordinator lifecycle management soon

PR #5060, opened 2024-11-07
[YiR] Allow option to hide YiR in Settings, remove developer settings flag

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T376310

### Notes
This PR adds a sub settings menu with a toggle to hide year in review. When the toggle is turned off, we call the (existing) yir data controller method to clear out all the reports. I am also removing the developer feature flag.

To save time with building, I followed the same pattern as the push notifications settings submenu which had a pretty similar look. This means it lives app-side instead of within the WMFComponents package, and is built with UIKit (sorry!).

### Test Steps
1. Launch app NOT in the proper device region or primary app language wiki. Confirm settings Year in Review option doesn't show.
2. Launch app in the correct device region and primary app language, while logged into an account with edits. Confirm you see a "Year in Review" option in Settings. Confirm you see a "Year in Review" entry point in Profile.
3. Go back to settings, tap "Year in Review" and disable toggle.
4. (Optional, if testing on simulator) Open up simulator app container directory, WMFData.sqlite database. Confirm the year in review tables are now empty.
5. Confirm entry point is gone.
6. Go back to settings, tap "Year in Review" and enable toggle.
7. Background/foreground to trigger report popuation.
8. Confirm entry point is back in Profile.


mazevedofs (approved):
Looks good!

PR #5065, opened 2024-11-11

translatewiki:
Translation updates


mazevedofs (approved):
Something weird happened when I updated it with main - it's showing commits that are already in main. I'm merging it to check what is going on.

PR #5067, opened 2024-11-13
T375135 Lost location text in Places

rajparad:
**Phabricator:** 
https://phabricator.wikimedia.org/T375135
### Notes
* Issue was enable location dialog appears without the text below the enable button.
* The issue was not that footerText was missing but footerText was not changing the styles based on theme due to `updateFooterHTML()` was not called.
* After updating it, Footer text works with all Themes.

### Test Steps
1. Navigate to places
2. View the enable location dialog

### Screenshots/Videos
iPhone 12 white:
image

iPhone 12 Black theme: 
image

iPad Pro Dark theme: 
image


tonisevener:
Hi @rajparad,

Sorry for the delay. We can no longer reproduce this, so we think an iOS update may have fixed this. Feel free to re-open this PR if you can reproduce the bug. Thanks.

PR #5082, opened 2024-11-18
Fix Article Web View Link Bug

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T378546

### Notes
* When hiding find in page, it's also necessary to set the find in page view as nil. This forces the view to stop being the first responder as it updates the `canBecomeFirstResponder` prop in articleVC to false.
* This allows the webview to take over as first responder instead of the VC

### Test Steps
1. Click find in page in the Article view. 
2. Do to type anything, and tap a link. It should go to the new page
3. Go back, click find in page again. 
4. Type something. In the highlighted state, tap a link. It should work.
5. Return. The toolbar should be active on the view (not sure if this is a bug, it's the behavior as-is in prod)
6. Clear the term. Close the toolbar. Click on a link.
7. Open the toolbar and close it. Click on a link.
8. We can rely on QA for other scenarios.

### DIFF
--- a/Wikipedia/Code/ArticleViewController+FindInPage.swift
+++ b/Wikipedia/Code/ArticleViewController+FindInPage.swift
@@ -71,6 +71,10 @@ extension ArticleViewController {
             self.resignFirstResponder()
             completion?()
         }
+
+        self.findInPage.view?.removeFromSuperview()
+        self.view.layoutSubviews()
+        self.findInPage.view = nil
     }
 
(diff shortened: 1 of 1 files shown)


l-olson1214 (approved):

swrobel:
Will there be a new release with this fix?

swrobel:
@mazevedofs i just installed the latest Testflight build 7.6.3 (4683) and the same bug still exists. It does not appear to be fixed.

PR #5084, opened 2024-11-20
T376069

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T376069

### Notes
* Still has slide in it - do not merge until cleared by Haley/Robin

### Test Steps
* Fresh install
1. In goal region/language, populate history with 6+ articles
2. Background and foreground app
3. Navigate to year in review
4. On the second slide that's usually the 1.5 billion slide, it should tell you what day of the week you've read most on if you've read more than 1 article on a given day

### Screenshots/Videos

Simulator Screenshot - iPhone 15 Plus - 2024-11-20 at 09 44 55


WMFData/Sources/WMFData/Data Controllers/Shared/WMFPageViewsDataController.swift
```diff
@@ -178,4 +202,45 @@ public final class WMFPageViewsDataController {
...
+                }
+                
+                let dayFormatter = Date.FormatStyle()
+                    .weekday(.wide)
+                let dayOfWeek = timestamp.formatted(dayFormatter)
+                
+                pageViewDays.append(WMFPageViewDay(day: dayOfWeek, viewCount: count))
+            }
+            
+            return pageViewDays
```
tonisevener, line 241: I don't think this will work the way we want - what the existing `fetchPageViewCounts` method does above is take all CDPageViews, group them by _page_, and count up how many are in each group. So you end up with a [page: count] dictionary like:

[{"page": "Cat", "count": 1},
{"page": "Dog", "count": 3},
{"page": "Bird", "count": 2}]*

_*The "page" value type here is actually an `NSManagedObjectID` type, but I'm using the article title here for clarity._

Meaning Cat was viewed once, Dog was viewed 3 times, etc.

But what you're doing here is similar but you're grouping by timestamp. Since timestamp is hyper-specific, you're likely not going to get more than one count per grouping, since I don't expect it's possible to view multiple articles at the exact same time:

[{"timestamp": "2024-10-09T14:20:58Z", "count": 1},
{"timestamp": "2024-10-09T14:21:32Z", "count": 1},
{"timestamp": "2024-10-09T14:22:46Z", "count": 1}]

Meaning the number of articles viewed at exactly `2024-10-09T14:20:58Z` was one, etc.

So I think going the `fetchedGrouped` route doesn't make sense here. What I suggest doing is pulling an array of CDPageViews from Core Data, then looping through and manually adding them to your own counts dictionary, like this:

```
            let predicate = NSPredicate(format: "timestamp >= %@ && timestamp <= %@", startDate as CVarArg, endDate as CVarArg)
            
            guard let cdPageViews = try self.coreDataStore.fetch(entityType: CDPageView.self, predicate: predicate, fetchLimit: nil, in: context) else {
                return []
            }
            
            // [dayofWeek: count]
            var countsDictionary: [String: Int] = [:]
            
            for cdPageView in cdPageViews {
                
                // Figure out cdPageView's dayOfWeek here from cdPageView.timestamp
                
                // Add or increment counts dictionary here
            }
            
            // turn counts dictionary into array of WMFPageViewDay, return that
```

Wikipedia/Code/YearInReviewCoordinator.swift
```diff
@@ -181,6 +185,77 @@ final class YearInReviewCoordinator: NSObject, Coordinator {
...
+            value: "You read the most articles on %1$@. It's clear that %1$@ are your prime day for exploring new content. Thanks for making the most of your reading time!",
+            comment: "Year in review, personalized slide subtitle for users that displays the weekday they read most. %1$@ is replaced with the weekday."
+        )
+        return String.localizedStringWithFormat(format, getLocalizedDay(day: day))
+    }
+    
+    func getLocalizedDay(day: String) -> String {
+        let localizedDay: String
+        switch day {
+        case "Sunday":
```
tonisevener, line 211: Your most read day (`WMFPageViewDay`)'s `day` value are these strings only if the user's device is in English. When my device language is in Spanish:

```
let dayFormatter = Date.FormatStyle().weekday(.wide)
let dayOfWeek = timestamp.formatted(dayFormatter) //value from my testing today is "jueves" 
```

Instead of that code in WMFPageViewsDataController, you should find a way to convert the timestamp into a day of the week that is the same regardless of device language (like a value 1 - 7).

tonisevener (changes requested):

Wikipedia/Code/YearInReviewCoordinator.swift
l-olson1214: Would it be a problem that some people start weeks on Sundays and some start on Mondays?

tonisevener: Yeah could be, good catch! Maybe we could check Calendar.current.firstWeekday at this point? Something like:

```
func getLocalizedDay(day: Int) -> String {
        let localizedDay: String
        
        switch (Calendar.current.firstWeekday, day) {
        case (1, 1), (2, 7): // Sunday
            localizedDay = WMFLocalizedString(
                "year-in-review-day-sunday",
                value: "Sundays",
                comment: "Localized name for Sunday in plural form."
            )
        case (1, 2), (2, 1): // Monday
            localizedDay = WMFLocalizedString(
                "year-in-review-day-monday",
                value: "Mondays",
                comment: "Localized name for Monday in plural form."
            )
       ...etc
```

So if the Calendar's weekday value for the timestamp is 1, and the Calendar's firstWeekday value is 2 (meaning they start counting the week on Mondays), I think you need to display "Monday".

WMFData/Sources/WMFData/Data Controllers/Shared/WMFPageViewsDataController.swift
l-olson1214: Done!

WMFData/Sources/WMFData/Data Controllers/Shared/WMFPageViewsDataController.swift
```diff
@@ -28,6 +28,30 @@ public final class WMFPageViewCount: Identifiable {
...
+public final class WMFPageViewDay: Decodable, Encodable {
+    public let day: String
+    public let viewCount: Int
+    
+    public init(day: String, viewCount: Int) {
+        self.day = day
+        self.viewCount = viewCount
+    }
+    
+    public init(from decoder: any Decoder) throws {
```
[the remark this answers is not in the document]
l-olson1214: Done!

tonisevener (approved):
Working great!

PR #5094, opened 2024-11-22
[YiR] Populate initial most read slide if needed

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T376069

### Notes
This is a bug I found when testing around on Year in Review. In this situation, I had an older app version 7.6.2 on device, displaying Year in Review with the article count. When then running against the `main` branch, I noticed the most read day slide didn't show as expected. 

The reason is [here](https://github.com/wikimedia/wikipedia-ios/compare/most-read-fix?expand=1#diff-1c01e90351b11ab077c5dff447720470dad032889b5206d5e9c5e4a69065649cL337). Since we already had a list of slides persisted, the most read initial slide is never added and this conditional is skipped. So in users who upgrade from 7.6.2 to 7.6.3, they would never see the new slides.

### Test Steps
1. Download TestFlight 7.6.2 build. Ensure region is as expected, build up an article count. Go to app Settings and toggle on and off year in review to display the entry point.
2. Confirm you see year in review article read count slide.
3. In this branch, temporarily set `WMFRemoteAppConfigCheckInterval = 1` [here](https://github.com/wikimedia/wikipedia-ios/blob/main/Wikipedia/Code/WMFAppViewController.m#L45) to force the app to fetch the new remote config.
4. Then run from this branch, confirm most read day slide appears.


l-olson1214 (changes requested):
LGTM just some test failures!

tonisevener:
> LGTM just some test failures!

@l-olson1214 Sorry, missed that they failed before logging off last time. These are fixed now!

l-olson1214 (approved):

PR #5085, opened 2024-11-20
[YiR] Update flow to show collective for logged out users

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T379796

### Notes
This PR removes the requirement that a user must be eligible to see at least one personalized slide for the entry point to appear. It also defaults to all collective slides from the coordinator if the user is logged out.

### Test Steps
1. Ensure device/simulator is in target region. Log out of app if needed, then fresh install app to ensure a logout state upon first launch.
2. Year in Review feature announcement should display when landing on Explore.
3. Go through Year in Review flow, you should see all collective slides.
4. Once you are done with the flow, go to Profile, tap Year in Review button.
5. You should see login/account creation prompt (from https://phabricator.wikimedia.org/T379794 work).
6. Log in, you should now be able to enter the flow through Profile.
7. Depending on your account, you will see the personalized editing slide if you have edits, or the collective slide. Other slides should still be collective since you didn't build up an article history.
8. Once complete with YiR flow, go to app Settings and turn off Year in Review.
9. Go through articles and build up a viewed history. Save some articles as well.
10. Go back to app Settings and turn on Year in Review.
11. Go through Year in Review flow through Profile, you should now see reading personalized slides.

PR #5098, opened 2024-11-28
[YiR] V2 - Saved Articles slide

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T376067

### Notes
* Adds the saved count slide for eligible users (logged-in users with more than 3 saved articles)
* Shows collective slide for the rest of users
* Saved count and the random saved article titles are retrieved from the main app core data and passed through an async call, using a delegate
* These new core data methods live in MWKSavedPageList
* Copy and assets are still outdated, no V2 updates were made

### Test Steps
1. Logged in, device set to Italy or Mexico, set main language to en, es or it
2. Browse 7+ articles
3. Background the app 
4. You should see the personalized saved article slide if you have more than 3 saved articles, plus the names of 3 random articles you saved
5. Remove saved article, leave 2 or less
6. Trigger the regeneration of the report (you can disable and re-enable the YiR repot in settings)
7. You should now see the collective slide
8. Logged out, in a device with any number of saved slides - you should always see the collective slide


mazevedofs:
Opening as a draft so I can get feedback on some issues I'm having

Wikipedia/Code/YearInReviewCoordinator.swift
```diff
@@ -266,69 +286,81 @@ final class YearInReviewCoordinator: NSObject, Coordinator {
...
     }
-    
+
+    func personalizedSavedCountSlideTitle(savedCount: String) -> String {
+        let format = WMFLocalizedString("year-in-review-personalized-saved-title-format", value: "You saved %1$@ articles", comment: "Year in review, personalized saved articles slide subtitle. %1$@ is replaced with the number of articles the user saved.")
+        return String.localizedStringWithFormat(format, savedCount)
+    }
+
+    func personalizedSavedCountSlideSubtitle(savedCount: String, articleNames: [String]) -> String {
+        let format = WMFLocalizedString("year-in-review-personalized-saved-subtitle-format", value: "You saved %1$@ articles this year, including \"%2$@\", \"%3$@\" and \"%4$@\". Each ", comment: "Year in review, personalized saved articles slide subtitle. %1$@ is replaced with the number of articles the user saved, %2$@, %3$@ and %4$@ are replaced with the names  three random articles the user saved.")
```
tonisevener, line 339: Also maybe consider a plural format. What if they have less than 3 saved articles, is this shown? How should the article titles act?

Update: it does look like we only display this slide if 3 or more articles are saved. For safety maybe we should be strict about what we pass in via the parameters (so an article1Title, article2Title, and article3Title param). That way there's no risk of a future dev passing in a short articleNames array and causing an out of index crash.

Wikipedia/Code/YearInReviewCoordinator.swift
```diff
@@ -382,6 +414,25 @@ final class YearInReviewCoordinator: NSObject, Coordinator {
                             hideDonateButton: true)
                     }
                 }
+            case .savedCount:
+                if slide.display == true,
+                   let data = slide.data {
+                    let decoder = JSONDecoder()
+                    if let savedSlidedata = try? decoder.decode(SavedArticleSlideData.self, from: data),
+                       savedSlidedata.savedArticlesCount > 3,
+                       savedSlidedata.articleTitles.count > 2 {
```
tonisevener, line 423: Should this be `> 3`? I'm curious about the difference here. Also minor casing issue: `savedSlideData`.

Wikipedia/Code/YearInReviewSettingsViewController.swift
```diff
@@ -197,3 +198,14 @@ final class YearInReviewSettingsViewController: SubSettingsViewController {
     }
 
 }
+
+extension YearInReviewSettingsViewController: SavedArticleSlideDataDelegate {
+    func getSavedArticleSlideData(from startDate: Date, to endDate: Date) async -> SavedArticleSlideData {
+        await MainActor.run {
```
tonisevener, line 204: Would be nice to find a way to avoid the duplicate code with WMFAppViewController+Extensions. Maybe a `SavedArticleSlideDataDelegate` default protocol implementation?

tonisevener (changes requested):
Just some little things from code review. I will push these PR feedback tweaks.

Wikipedia/Code/WMFAppViewController.m
```diff
@@ -2214,6 +2216,7 @@ - (void)userWasLoggedOut:(NSNotification *)note {
     });
     
     [self deleteYearInReviewPersonalizedEditingData];
+    // TODO - delete saved slide
```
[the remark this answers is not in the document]
tonisevener: Update: per Product, no need to clear out slide data in those situations, specific clearing instructions are covered in the FAQ. I'll remove this todo comment.

WMFData/Sources/WMFData/Data Controllers/Year In Review/WMFYearInReviewDataController.swift
```diff
@@ -369,6 +392,10 @@ import CoreData
                 if slide.evaluated == false && yirConfig.personalizedSlides.donateCount.isEnabled {
                     needsDonatingPopulation = true
                 }
+            case WMFYearInReviewPersonalizedSlideID.savedCount.rawValue:
+                if slide.evaluated == false && yirConfig.personalizedSlides.saveCount.isEnabled && username != nil {
```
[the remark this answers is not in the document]
tonisevener: Update: ok, I see now that these personalized slides are only shown if they are logged in. I will work on adding this logic to https://github.com/wikimedia/wikipedia-ios/pull/5085.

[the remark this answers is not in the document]
tonisevener: Update again: I removed this username check, since technically the saved article data is available to us whether they are logged in or not, whereas with the edit counts we must have a username before we can populate personalized data.

So the local personalized data will still be populated even if logged out, but we will hide the personalized slides at the entry point / coordinator level. I will double-check this when I wrap up https://github.com/wikimedia/wikipedia-ios/pull/5085.

I checked with Product and this is acceptable.

Wikipedia/Code/YearInReviewSettingsViewController.swift
tonisevener: I deduped a bit by moving the SavedArticleSlideDataDelegate conformance into MWKSavedPageList.

tonisevener (approved):

PR #5100, opened 2024-11-28
[YiR] Hide donate buttons from countries we cannot fundraise

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T378842

### Notes
* Check the remote config to verify if we should hide the donate button in countries where we cannot fundraise

### Test Steps
1. You might need to bypass the locale verifications we have in place that limit the feature currently to Italy and Mexico
2. Set the device region to one of the countries in this [list](https://donate.wikimedia.org/wiki/MediaWiki:AppsFeatureConfig.json)
3. Go to YiR. Make sure you can't see the donate button on the slides and on the learn more web view

1. With the current verifications in place, set the device location to Italy or Mexico
2. Make sure you can see the donate buttons on the slides and on the learn more web view (if it satisfies the current verifications)


l-olson1214:
Found this bug on testing - will dig to find problem 

https://github.com/user-attachments/assets/280d4088-7bfa-410c-9421-920d688ea321

l-olson1214:
Fixed! Needed to have the shouldHideDonateButton use shouldHideDonate() instead of true

l-olson1214 (approved):

PR #5102, opened 2024-12-02
Make "this year" less repetitive in a message

amire80:
"This year" was repeated three times, one of which was definitely unnecessary. I reduced it to one, and tried to make it generally more readable. Feel free to adjust it further if the designers or the product managers want something different.


mazevedofs:
Just noting that our PM is fine with the update!

l-olson1214 (approved):

PR #5106, opened 2024-12-02
Update Insert Image wizard mediawiki links to consider device language

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T377887

### Notes
* Adds Special:MyLanguage and uselang= query items to mediawiki URLS to redirect to translated versions if available
* It falls back to English if translation is not available

### Test Steps
1. Set device language to Spanish
2. Go to the insert image wizard, and tap the learn more links
3. They should open in the Spanish translation of the page
4. Set device language to German (or other languages without translations)
5. Link should open in English


tonisevener (approved):
I do notice the scroll is slightly too far - you have to scroll a little bit up to see the section title. Since it's a small amount, I suspect it may be related to our custom navigation bar. Going to merge and see if it's fixed in the system nav bar branch.

PR #5109, opened 2024-12-05
T376353

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T376353
https://phabricator.wikimedia.org/T376355

### Notes
* Added 5px of padding to align the info icon as well

### Test Steps
1. In an account with edits - follow the steps to get year in review up
2. Check the fourth slide and see if it shows the number of edits 
3. Repeat steps with account that does not


WMFData/Sources/WMFData/Data Controllers/Year In Review/WMFYearInReviewDataController.swift
```diff
@@ -6,6 +6,7 @@ import CoreData
     public let coreDataStore: WMFCoreDataStore
     private let userDefaultsStore: WMFKeyValueStore?
     private let developerSettingsDataController: WMFDeveloperSettingsDataControlling
+    private let userID: Int?
```
tonisevener, line 9: @l-olson1214 For your other editing slide, you are passing in `username` through the `populateYearInReviewReportData` method instead of setting it as a property like you are here. Maybe we should do the same with `userID`, to keep it consistent? This way you don't have to send in userID upon instantiation, only when the population method is called (which is only when we really need it).

WMFData/Sources/WMFData/Data Controllers/Year In Review/WMFYearInReviewDataController.swift
```diff
@@ -553,6 +579,107 @@ import CoreData
...
+    // TODO: Grey
+    private func populateViewCountSlide(report: CDYearInReviewReport, backgroundContext: NSManagedObjectContext, project: WMFProject?) throws {
+        guard let iosFeatureConfig = developerSettingsDataController.loadFeatureConfig()?.ios.first,
+              let _ = iosFeatureConfig.yir(yearID: targetConfigYearID) else {
+            throw WMFYearInReviewDataControllerError.missingRemoteConfig
+        }
+
+        guard let userID else { return }
+        
+        let encodedUserId = "#" + String(userID)
```
tonisevener, line 592: We're not encoding the "#" anymore, at least not at this point (we may be deeper in the networking stack), so maybe rename to `prefixedUserId`, or something like that. Also, I think it would be better to move this work into the `fetchEditViews(` method below, which makes the network call. This way callers of `fetchEditViews` don't need to remember to prefix a `#` to the userID.

WMFData/Tests/WMFDataTests/WMFYearInReviewDataControllerCreateOrRetrieveTests.swift
```diff
@@ -7,13 +7,14 @@ fileprivate class WMFMockYearInReviewDataController: WMFYearInReviewDataControll
     var shouldCreateOrRetrieve = true
     var mockSlides: [WMFYearInReviewSlide] = []
 
-    override init(coreDataStore: WMFCoreDataStore? = WMFDataEnvironment.current.coreDataStore, userDefaultsStore: (any WMFKeyValueStore)? = WMFDataEnvironment.current.userDefaultsStore, developerSettingsDataController: any WMFDeveloperSettingsDataControlling = WMFDeveloperSettingsDataController.shared) throws {
```
tonisevener, line 10: This might be fine, but just wanted to call out this change out in case it slipped in accidentally.

Wikipedia/Code/ArticleViewController+Announcements.swift
```diff
@@ -132,7 +132,14 @@ extension ArticleViewController {
             return false
         }
 
-        guard let yirDataController = try? WMFYearInReviewDataController() else {
+        var userId: Int?
+
+        if let siteURL = dataStore.languageLinkController.appLanguage?.siteURL,
+           let userID = dataStore.authenticationManager.permanentUser(siteURL: siteURL)?.userID {
+            userId = userID
+        }
```
tonisevener, line 140: (See comment above) You will be able to delete this `userId` setup code if you pass in `userID` as a parameter to `populateYearInReviewReportData`.

Wikipedia/Code/ArticleViewController+Announcements.swift
```diff
@@ -145,8 +152,14 @@ extension ArticleViewController {
     
     // TODO: remove after expiry date (1 March 2025)
     func presentYearInReviewAnnouncement() {
+        var userId: Int?
+
+        if let siteURL = dataStore.languageLinkController.appLanguage?.siteURL,
```
tonisevener, line 157: This can be deleted too.

Wikipedia/Code/ArticleViewController.swift
```diff
@@ -69,7 +69,14 @@ class ArticleViewController: ViewController, HintPresenting {
     }
 
     private var yirDataController: WMFYearInReviewDataController? {
-        return try? WMFYearInReviewDataController()
+        var userId: Int?
+
+        if let siteURL = dataStore.languageLinkController.appLanguage?.siteURL,
```
tonisevener, line 74: This can be deleted too.

Wikipedia/Code/ExploreViewController.swift
```diff
@@ -15,7 +15,13 @@ class ExploreViewController: ColumnarCollectionViewController, ExploreCardViewCo
     private var altTextImageRecommendationsOnboardingPresenter: AltTextImageRecommendationsOnboardingPresenter?
 
     private var yirDataController: WMFYearInReviewDataController? {
-        return try? WMFYearInReviewDataController()
+        var userId: Int?
+
+        if let siteURL = dataStore.languageLinkController.appLanguage?.siteURL,
```
tonisevener, line 20: This can be deleted too.

Wikipedia/Code/ExploreViewController.swift
```diff
@@ -1105,8 +1111,14 @@ extension ExploreViewController {
 
     // TODO: Remove after expiry date (1 March 2025)
     private func presentYearInReviewAnnouncement() {
+        var userId: Int?
+
+        if let siteURL = dataStore.languageLinkController.appLanguage?.siteURL,
```
tonisevener, line 1116: This can be deleted too.

Wikipedia/Code/ProfileCoordinator.swift
```diff
@@ -78,11 +78,17 @@ final class ProfileCoordinator: NSObject, Coordinator, ProfileCoordinatorDelegat
         )
         
         let inboxCount = try? dataStore.remoteNotificationsController.numberOfUnreadNotifications()
+        var userId: Int?
+
+        if let siteURL = dataStore.languageLinkController.appLanguage?.siteURL,
```
tonisevener, line 83: This can be deleted too.

Wikipedia/Code/WMFAppViewController+Extensions.swift
```diff
@@ -629,9 +635,15 @@ extension WMFAppViewController {
     }
     
     @objc func deleteYearInReviewPersonalizedEditingData() {
+        var userId: Int?
+
+        if let siteURL = dataStore.languageLinkController.appLanguage?.siteURL,
```
tonisevener, line 640: This can be deleted too.

Wikipedia/Code/YearInReviewSettingsViewController.swift
```diff
@@ -105,7 +106,15 @@ final class YearInReviewSettingsViewController: SubSettingsViewController {
...
-                let yirDataController = try WMFYearInReviewDataController()
+                
+                var userId: Int?
+
+                if let siteURL = dataStore.languageLinkController.appLanguage?.siteURL,
+                   let userID = dataStore.authenticationManager.permanentUser(siteURL: siteURL)?.userID {
+                    userId = userID
+                }
+                
+                let yirDataController = try WMFYearInReviewDataController(userID: userId)
```
tonisevener, line 117: Here you'll pass `userID` into the next line as a parameter instead of into the init.

Wikipedia/Code/WMFAppViewController+Extensions.swift
```diff
@@ -613,10 +613,16 @@ extension WMFAppViewController {
...
+           let userID = dataStore.authenticationManager.permanentUser(siteURL: siteURL)?.userID {
+            userId = userID
+        }
 
         Task {
             do {
-                let yirDataController = try WMFYearInReviewDataController()
+                let yirDataController = try WMFYearInReviewDataController(userID: userId)
```
tonisevener, line 625: Here you'll pass `userID` into the next line as a parameter instead of into the init.

WMFData/Sources/WMFData/Data Controllers/Year In Review/WMFYearInReviewDataController.swift
```diff
@@ -553,6 +579,107 @@ import CoreData
...
+                    }
+                    
+                    switch slideID {
+                    case WMFYearInReviewPersonalizedSlideID.viewCount.rawValue:
+                        do {
+                            let encoder = JSONEncoder()
+                            slide.data = try encoder.encode(totalViews)
+                            
+                            if totalViews > 0 {
+                                slide.display = true
```
tonisevener, line 614: I think we'll run into threading issues because in this network call completion block, you are no longer on the Core Data thread, and updating Core Data models from that could cause crashes.

When we worked on your edit count slide, we determined it works better when we first make the network call separately to get the count, then ask the Core Data thread to update its report:

```
if result.needsEditingPopulation == true {
            if let username {
                let edits = try await fetchEditCount(username: username, project: primaryAppLanguageProject)
                try await backgroundContext.perform { [weak self] in
                    try self?.populateEditingSlide(edits: edits, report: report, backgroundContext: backgroundContext)
                }
            }
}
```

Can you follow that same pattern for this slide? You'll need to update `fetchEditViews` to be `async` instead of using completion blocks.

Also just a note, when I add `-com.apple.CoreData.ConcurrencyDebug 1` as a launch argument, I get complaints, but I get them even in `main` 😬. I can look into that tomorrow.

tonisevener (changes requested):

WMFData/Sources/WMFData/Data Controllers/Year In Review/WMFYearInReviewDataController.swift
tonisevener: Update: Luckily it was just a small fix for the `main` threading complaint, I pushed a commit fixing that. Now the launch argument complains in this area. Let's see if it reworking it to the edit count slide pattern fixes that, thanks!

Screenshot 2024-12-05 at 7 51 47 PM

WMFData/Tests/WMFDataTests/WMFYearInReviewDataControllerCreateOrRetrieveTests.swift
l-olson1214: It was one of the errors I had gotten, do you think it needs to be changed?

WMFData/Sources/WMFData/Data Controllers/Year In Review/WMFYearInReviewDataController.swift
l-olson1214: Done!

WMFData/Tests/WMFDataTests/WMFYearInReviewDataControllerCreateOrRetrieveTests.swift
tonisevener: I think it just had to do with adding userID in the data controller `init`, it needed to be added here too in order for the override to work. But now that you removed `userID` from the init you should be able to bring back the override.

tonisevener (changes requested):
Most feedback is addressed, thanks! Still need to rework the network call to match the other edit count slide to avoid threading issues. Let me know if you want me to grab that part.

Wikipedia/Code/YearInReviewCoordinator.swift
```diff
@@ -557,8 +597,8 @@ final class YearInReviewCoordinator: NSObject, Coordinator {
             thirdSlide = saveCountSlide
         }
 
-        if let editCountSlide = personalizedSlides.editCount {
-            fourthSlide = editCountSlide
+        if let viewCountSlide = personalizedSlides.viewCount {
+            fourthSlide = viewCountSlide
```
tonisevener, line 601: One last thing! We should keep the editCount slide as the fourth slide, and insert the edits viewed count slide as a fifth slide. It might help to reference the [Figma](https://www.figma.com/design/JRTyGA3rZ0qlQ9eqh0FMhE/iOS---Wikipedia-YiR?node-id=2131-2181&node-type=canvas&t=u9Rh2aPv8fkxySZC-0) for an overview.

tonisevener (approved):
🎉

PR #5111, opened 2024-12-06
[YiR] Fix Read Count

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T381511

### Notes
This PR changes around the way we fetch the data to evaluate the read count and most read day slide. We can't really rely on WMFData's `WMFPageViewsDataController` anymore, because the 2024 page views may not have migrated properly from the old database to the new one.

So as a stopgap measure for 2024 only, I am fetching this data from the old database. I referenced the saved counts slide as inspiration, since it also had to fetch from the old database in a similar way.

Another thing I am trying is wrapping up the report population method into a background task. You can see [this documentation](https://developer.apple.com/documentation/uikit/extending-your-app-s-background-execution-time) for more details there, but essentially this gives us extra time in the background to finish data population, in case a user backgrounds while the report is being generated.

### Test Steps
1. Just regression testing. Ensure the read count and most read day count slides work as before. If you have an older install of the app (< October 2024), you should see a higher article read count than you did in `main`.


l-olson1214 (approved):
LGTM! @tonisevener just some merge conflicts :)

PR #5112, opened 2024-12-11
Periodic update/refresh of languages.

dbrant:
(pardon the drive-by)
It looks like the `Update Languages` scheme hasn't been run in a while, and this will bring in a few new language wikis that have popped up in recent months.

Feel free to disregard and/or close if it's not as simple as running this script 😅


tonisevener (approved):
Thanks for the PR! This looks good to me. You are correct that it is as simple as running this script.

The last time I ran this I thought it sprinkled in incubator projects, but I'm not noticing the same issue.

PR #5120, opened 2024-12-20
Add zh-cn

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T363511

### Notes
* Adds based on Android's settings

### Test Steps
1. Check in languages in settings for Chinese (China)


mazevedofs (changes requested):
It's been a while since I added variants, but we might need to display an alert for users with other zh variants to let them know this new variant is available. Last time we didi it for other language variants was here https://github.com/wikimedia/wikipedia-ios/pull/4550
Let me know if this is not the case for this variant.

mazevedofs (approved):
The alert part will be completed in a separate task per Product direction on https://phabricator.wikimedia.org/T360915

PR #5121, opened 2024-12-20
Remove custom navigation bar

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T363606

### Notes
This PR removes the ViewController, NavigationBar, NavigationBarHider, and WMFThemeableNavigationController superclasses and deals with the fallout. There other other smaller cleanup items, but those are the main ones.

I added a WMComponentNavigationController (subclass of UINavigationController) in WMFComponents for us to switch to. All legacy view controllers and newer WMFComponentViewControllers use this for navigating.

I also added a WMFNavigationBarConfiguring and default protocol implementation of `configureNavigationBar(...)` that all UIViewControllers can (optionally) call. I also tried to comb through most of the app and use this method in every `viewWillAppear`, so that all configuration flows through one method.

 There are still bugs, code cleanup and design review feedback that I need to do, but I wanted to get this up (at least in draft form) so that we can discuss it and potentially start review.

### Test Steps
1. Run the app, bask in the glow of the system navigation bar.


WMFComponents/Sources/WMFComponents/Base/Component Base Classes/WMFComponentHostingController.swift
```diff
@@ -4,49 +4,60 @@ import Combine
...
+    private func subscribeToAppEnvironmentChanges() {
+        WMFAppEnvironment.publisher
+            .sink(receiveValue: { [weak self] _ in self?.appEnvironmentDidChange() })
+            .store(in: &cancellables)
+    }
+
+    // MARK: - Subclass Overrides
+
+    public func appEnvironmentDidChange() {
+    overrideUserInterfaceStyle = appEnvironment.theme.userInterfaceStyle
```
l-olson1214, line 58: Small tiny: can we indent this

WMFComponents/Sources/WMFComponents/Extensions/UIViewController+NavigationBarUtils.swift
```diff
@@ -0,0 +1,333 @@
+// Common methods used for navigation bar display in the app
+import UIKit
+
```
mazevedofs, line 3: Optional: I think WMFNavigationBarConfiguring deserves to be in it's own file, rather than in a Utils file, same as WMFNavigationBarHiding.

Wikipedia/Code/DiffHeaderTitleView.swift
```diff
@@ -104,7 +124,7 @@ extension DiffHeaderTitleView: Themeable {
     func apply(theme: Theme) {
         
         backgroundColor = theme.colors.paperBackground
-        contentView.backgroundColor = theme.colors.paperBackground
+        backgroundColor = theme.colors.paperBackground
```
l-olson1214, line 127: Duplicate of line above

Wikipedia/Code/DiffHeaderView.swift
```diff
@@ -1,7 +1,42 @@
...
+        NSLayoutConstraint.activate([
+            topAnchor.constraint(equalTo: contentView.topAnchor),
+            leadingAnchor.constraint(equalTo: contentView.leadingAnchor),
+            trailingAnchor.constraint(equalTo: contentView.trailingAnchor),
+            bottomAnchor.constraint(equalTo: contentView.bottomAnchor)
+        ])
+    }
+    
+    func configure(with vm: DiffHeaderViewModel, titleViewTapDelegate: DiffHeaderTitleViewTapDelegate? = nil, theme: Theme) {
+        contentView.configure(with: vm, titleViewTapDelegate: titleViewTapDelegate, theme: theme)
```
l-olson1214, line 31: Would this be a good candidate for just passing the function through? I know we talked about that the other day in the badge delegate PR

Wikipedia/Code/DiffListViewController.swift
```diff
@@ -241,8 +233,42 @@ private extension DiffListViewController {
 }
 
 extension DiffListViewController: UICollectionViewDataSource {
+    
+    func numberOfSections(in collectionView: UICollectionView) -> Int {
+        return 3
```
l-olson1214, line 238: Why isn't this just a regular variable / computed variable? Why is it a func?

WMFComponents/Sources/WMFComponents/Components/Suggested Edits/Image Recommendations/WMFImageRecommendationsViewController.swift
```diff
@@ -182,6 +165,21 @@ public final class WMFImageRecommendationsViewController: WMFCanvasViewControlle
         cancellables.removeAll()
     }
     
+    public override func viewDidDisappear(_ animated: Bool) {
+        super.viewDidDisappear(animated)
+        
+        if parent == nil {
+            tappedBack()
```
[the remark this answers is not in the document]
tonisevener: @l-olson1214 I think it's okay unless there's something particular you want me to write out here. Before in the old code, we added a custom navigation button that called `tappedBack` upon tap, which logs analytics that they tapped the back button. From what I could find this is a way to get similar functionality with the system nav back button.

Wikipedia/Code/SearchResultsViewController.swift
```diff
@@ -9,13 +13,11 @@ class SearchResultsViewController: ArticleCollectionViewController {
             reload()
         }
     }
-
-    var delegatesSelection: Bool = false
-    var doesShowArticlePreviews = true
+    
+    weak var delegate: SearchResultsViewControllerDelegate?
```
l-olson1214, line 17: Can we just pass the func here

Wikipedia/Code/DiffListViewController.swift
tonisevener: This one is just conforming to the UIKit [UICollectionDataSource](https://developer.apple.com/documentation/uikit/uicollectionviewdatasource/numberofsections(in:)) protocol, basically it's how we tell the collection view how many sections to display in the list.

It defaults to 1, which is why we didn't have to implement it at all before. With this change, we're taking some header views that were embedded in the custom navigation bar and sticking them in the list as section headers. So now the number of sections went from 1 to 3.

WMFComponents/Sources/WMFComponents/Extensions/UIViewController+NavigationBarUtils.swift
tonisevener: Done!

Wikipedia/Code/DiffHeaderTitleView.swift
tonisevener: Fixed!

Wikipedia/Code/DiffHeaderView.swift
tonisevener: @l-olson1214  Reworked these to use closures, let me know if you'd like more tweaks.

Wikipedia/Code/SearchResultsViewController.swift
tonisevener: @l-olson1214 Reworked these to use closures, let me know if you'd like more tweaks.

Wikipedia/Code/PageHistoryViewController.swift
```diff
@@ -17,7 +17,11 @@ enum SelectionOrder: Int, CaseIterable {
 }
 
 @objc(WMFPageHistoryViewController)
-class PageHistoryViewController: ColumnarCollectionViewController {
+class PageHistoryViewController: ColumnarCollectionViewController, WMFNavigationBarConfiguring {
+    
+    // fileprivate static let headerReuseIdentifier = "TestHeader"
```
mazevedofs, line 22: Minor: I think this comment can deleted

l-olson1214 (approved):
This looks really good! There's nothing in here that I'm concerned about terribly, and I haven't been able to find any massive issues. 

Thank you for the hard work you put into this!

Wikipedia/Code/PageHistoryViewController.swift
tonisevener: Ah good catch. Fixed!

mazevedofs (approved):
LGTM!

PR #5125, opened 2025-01-06
Make view first responder when closing find in page 

mazevedofs:
**Phabricator:**  https://phabricator.wikimedia.org/T378546

### Notes
* Follow up to work done in https://github.com/wikimedia/wikipedia-ios/pull/5082 to fix some other scenarios

### Test Steps
1. Click find in page in the Article view.
2. Don't type anything on the field, and tap a link. It should go to the new page
3. Go back, click find in page again.
4. Type something. In the highlighted state, tap a link. It should work.
5. Return. The toolbar should be active on the view (not sure if this is a bug, it's the behavior as-is in prod)
6. Clear the term. Close the toolbar. Click on a link.
7. Open the toolbar and close it. Click on a link.
8. Try other possible combinations or scenarios if possible


Wikipedia/Code/ArticleViewController+FindInPage.swift
```diff
@@ -75,6 +75,7 @@ extension ArticleViewController {
         self.findInPage.view?.removeFromSuperview()
         self.view.layoutSubviews()
         self.findInPage.view = nil
+        self.view.becomeFirstResponder()
```
tonisevener, line 78: What a weird bug!

I still see an issue when I do this:

1. Tap featured article
2. Tap find, type "the". If in simulator, ensure simulator keyboard appears.
3. Clear search field, close find.
4. Keyboard dismisses. Interestingly tapping a link that was under the keyboard doesn't work, but tapping a link near the top does work.

It does look a bit fishy that you are changing responders here, then above on line 71 you are also messing with responders in the completion block. This line may get run before line 71. It might work better if all responder changes happen within that completion block, but I'm not having any luck so far in my tinkering.

tonisevener: Update: This is working a little better for me, but you should double-check it:

```
    func hideFindInPage(_ completion: (() -> Void)? = nil) {
        resetFindInPage {
            self.findInPage.view?.hide()
            self.findInPage.view?.removeFromSuperview()
            self.findInPage.view = nil
            self.resignFirstResponder()
            completion?()
        }
    }
```

I noticed `showFindInPage()` above has a `becomeFirstResponder()` call, so I'm balancing it out here with a `self.resignFirstResponder()`. I'm not sure if that's helping or not, or if I'm just randomly hitting better scenarios.

mazevedofs: This solution is working better for item #4 you described here. My previous attempt was working for the scenarios I tested, so it was a little tricky to debug. The responder being altered inside and outside that closure might be the cause of this behaviour here. I'll maintain this version and have QA test for other possibilities. Thanks for the help!

tonisevener (approved):

PR #5127, opened 2025-01-06
[YiR] Show YiR after creating account through Profile

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T379794

### Notes
Fixes an issue found in https://phabricator.wikimedia.org/T379794#10428455. After creating an account, there was no completion block assigned to navigate to Year in Review. This PR adds the completion handler (similar to how Login is working).

### Test Steps
1. Launch app while logged out.
2. Tap YiR in Profile.
3. Tap "Join Wikipedia / Log in"
4. Tap "Join Wikipedia"
5. Create a new account. Year in Review should present after successfully creating an account.


mazevedofs (approved):
Looks good! I'll leave testing to QA so we don't create too many test accounts

PR #5133, opened 2025-01-09
Update numbers for EN slides

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T382914

### Notes
* Updated the EN specific slides and all of their already translated languages

### Test Steps
1.


tonisevener (approved):
Looks good! I pushed one fix to Korean that I think might have been missed. I also alphabetized the strings, just because it will make my review of the changes in https://github.com/wikimedia/wikipedia-ios/pull/5130 easier.

PR #5138, opened 2025-01-10
Update badge delegate

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T382932

### Notes
* 

### Test Steps
1. Check on a fresh install the notifications, make sure they go away for logged out users upon clicking


mazevedofs (changes requested):
It's not clearing the badge state for me. 
Other than just calling this upon clicking, we have logic in place [here](https://github.com/wikimedia/wikipedia-ios/blob/b1074dd94ccec1cdf634ac1a8610ee78e1402a7c/WMFData/Sources/WMFData/Data%20Controllers/Year%20In%20Review/WMFYearInReviewDataController.swift#L60) to verify if the user has seen the first slide. 
Perhaps in this situation, you need to create a separate check for logged out users that clears this state after just clicking the item in menu.
We still need the current logic in place to logged-in users, so a different check would work better in this case.

edit: You don't need another delegate method, but work on YiRDataController whether you should clear the badge immediately or still check if the user saw the first slide

WMFComponents/Sources/WMFComponents/Components/Profile/WMFProfileViewModel.swift
```diff
@@ -272,6 +275,10 @@ enum ProfileState {
                 isDonate: false,
                 isLoadingDonateConfigs: false,
                 action: {
+                    if let dataController = try? WMFYearInReviewDataController() {
+                        dataController.hasTappedProfileItem = true
+                        needsYiRNotification = false
+                    }
```
mazevedofs, line 281: Wouldn't it be possible to make this a @Published var so it updates the UI immediately when the status change? No need to leave the view to have it cleared.

mazevedofs (changes requested):

WMFComponents/Sources/WMFComponents/Components/Profile/WMFProfileViewModel.swift
mazevedofs: Now, the badge state in the profile icon never clears

l-olson1214: Looking into it now - I'm not sure if Published would work in the enum or if there's an alternate place to put it!

tonisevener: @l-olson1214 I think if you can manage to call `loadProfileSections()` again, that regenerates the profile sections (which are a @Published array) and may cause the UI to update.

PR #5145, opened 2025-01-10
T383359

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T383359

### Notes
* 

### Test Steps
1. Test as outlined in ticket

### DIFF
--- a/Wikipedia/Code/DonateFunnel.swift
+++ b/Wikipedia/Code/DonateFunnel.swift
@@ -360,6 +360,10 @@ import WMF
         logEvent(activeInterface: .wikiYiR, action: .impression, actionData: ["slide": "new_account_engage"])
     }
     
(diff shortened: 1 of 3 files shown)


tonisevener (changes requested):
@l-olson1214 Definitely better! Testing Turkish looked good. I notice the two links in the text on these slides do not link to the German web view, despite there being German translations in that web view.

Screenshot 2025-01-10 at 6 29 26 PM
Screenshot 2025-01-10 at 6 29 29 PM

It navigates to English, which isn't too bad, the user can just tap German from the web view if they need to (though it will kick them out of the app). Better than a broken page like it was fur Turkish before.

Also can you fix the diff here, looks like some reverts from `main` slipped in. If it's easier to close this PR/branch and open a new one against the `7.6.3` branch, that's fine with me.

PR #5163, opened 2025-01-22
Add profile button to Saved, fix Article profile button

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T383822
https://phabricator.wikimedia.org/T383828

### Notes
This PR does some cleanup work to the navigation config so that we can add a leading and trailing button, if needed. With that I was able to fix the missing Article profile icon in the `needsSearchBar = true` mode.

I also leaned on that work to add the profile button to the Saved tab.

To present profile, I created a profile and year in review coordinator, similar to how Explore and Article do now. I had to provide a `sourcePage` and `donateSource` enum for to instantiate profile coordinator, which then highlighted a bunch of logging areas we may need to log. I'll move T383822 into Blocked while we wait for those answers.

### Test Steps
1. Go to Saved tab, confirm profile icon is working as expected. Confirm tapping More > Edit switches all buttons to Cancel.
2. Go to an Article, confirm profile icon is working as expected (no search bar, profile button, search magnifying glass button).
3. Go to Explore, confirm profile icon is working as expected.
4. Update ArticleViewController, return `true` for `needsSearchBar`, confirm search bar appears as expected in https://phabricator.wikimedia.org/T383828. Confirm profile button appears.

### DIFF
--- a/Wikipedia/Code/SavedViewController.swift
+++ b/Wikipedia/Code/SavedViewController.swift
@@ -318,7 +318,7 @@ class SavedViewController: ThemeableViewController, WMFNavigationBarConfiguring,
         }
         
         let config = self.profileButtonConfig(target: self, action: #selector(userDidTapProfile), dataStore: dataStore, yirDataController: yirDataController, leadingBarButtonItem: nil, trailingBarButtonItem: moreBarButtonItem)
-        updateNavigationBarProfileButton(needsBadge: config.needsBadge)
+        updateNavigationBarProfileButton(needsBadge: config.needsBadge, needsBadgeLabel: CommonStrings.profileButtonBadgeTitle, noBadgeLabel: CommonStrings.profileButtonTitle)
     }
     
     private func setSavedArticlesViewControllerIfNeeded() {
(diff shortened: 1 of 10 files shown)


Wikipedia/Code/SavedViewController.swift
```diff
@@ -225,7 +275,50 @@ class SavedViewController: ThemeableViewController, WMFNavigationBarConfiguring,
...
+            return
+        }
+        
+        guard let languageCode = dataStore.languageLinkController.appLanguage?.languageCode,
+              let metricsID = DonateCoordinator.metricsID(for: .savedProfile, languageCode: languageCode) else {
+            return
+        }
+        
+        // TODO: Do we need logging like this?
+        // DonateFunnel.shared.logExploreProfile(metricsID: metricsID)
```
mazevedofs, line 293: I asked Shay in the data sync yesterday, she said we only log it if we're trying to validate something specific. In this case, we might not log it.

mazevedofs (approved):
Looks great! (The comments are just noting that we'll have some updates when we get the analytics deck)

PR #5165, opened 2025-01-23
Add profile button to places tab

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T383821

### Notes
* Adds profile button to places tab

### Test Steps
1. Run the app, go to places tab
2. Verify the profile button
3. Make sure the button updates the badge accordingly
4. Make sure you can logout from the profile button when you access the profile from this tab, access other items from profile as well


Wikipedia/Code/PlacesViewController.swift
```diff
@@ -378,6 +385,55 @@ class PlacesViewController: ArticleLocationCollectionViewController, UISearchBar
...
+
+    private var yirDataController: WMFYearInReviewDataController? {
+        return try? WMFYearInReviewDataController()
+    }
+
+    @objc private func didTapProfileButton() {
+        profileCoordinator?.start()
+    }
+
+    private func updateProfileButton() {
```
tonisevener, line 432: @mazevedofs Can you try calling this method in applyTheme as well? Minor but the button tint color gets off when the theme is in Default and the system changes to dark or light mode.

Wikipedia/Code/PlacesViewController.swift
```diff
@@ -378,6 +385,55 @@ class PlacesViewController: ArticleLocationCollectionViewController, UISearchBar
...
+    private var _profileCoordinator: ProfileCoordinator?
+    private var profileCoordinator: ProfileCoordinator? {
+
+        guard let navigationController,
+              let yirCoordinator = self.yirCoordinator else {
+            return nil
+        }
+
+        guard let existingProfileCoordinator = _profileCoordinator else {
+            _profileCoordinator = ProfileCoordinator(navigationController: navigationController, theme: theme, dataStore: dataStore, donateSouce: .savedProfile, logoutDelegate: self, sourcePage: ProfileCoordinatorSource.saved, yirCoordinator: yirCoordinator)
```
tonisevener, line 416: Something interesting happens when the system automatically changes to dark or light mode and the app theme is Default.

1. From Places, open Profile and then Settings or Notifications
2. Tap back to Places.
3. Background app, change system light/dark mode
4. Go back to app, open profile, then Settings or Notifications
5. Settings or Notifications is the wrong theme

This doesn't happen with Explore, so maybe there's some sort of theme update there that isn't happening here. The Saved tab profile probably has the same bug.

tonisevener (changes requested):

Wikipedia/Code/PlacesViewController.swift
mazevedofs: It was happening when manually changing the theme as well. I'm updating the theme we send to the profile coordinator; it seemed to fix both situations. Let me know if you catch any other inconsistencies.

tonisevener (approved):

PR #5168, opened 2025-01-24
Update Done/cancel buttons

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T383460

### Notes
* Design review items from the nav bar updates

### Test Steps
1. Run the app. Check all the spots mentioned in Robin's comment liked on the task


Wikipedia/Code/SinglePageWebViewController.swift
```diff
@@ -181,7 +181,7 @@ class SinglePageWebViewController: ThemeableViewController, WMFNavigationBarConf
             let titleConfig = WMFNavigationBarTitleConfig(title: "", customView: nil, alignment: .hidden)
             
             if navigationController?.viewControllers.first === self {
-                closeConfig = WMFNavigationBarCloseButtonConfig(text: CommonStrings.doneTitle, target: self, action: #selector(closeButtonTapped(_:)), alignment: .leading)
+                closeConfig = WMFNavigationBarCloseButtonConfig(text: CommonStrings.cancelActionTitle, target: self, action: #selector(closeButtonTapped(_:)), alignment: .leading)
```
tonisevener, line 184: This fixes the Year in Review > Donate > Other Payment Method modal, but causes a regression with Article > Edit Article Description > Learn More:

> 	12	Article > Article Description Edit > Learn More

Screenshot 2024-12-19 at 4 53 14 PM


> ✅ Keep it as is

Changed to:
Screenshot 2025-01-24 at 3 10 29 PM

tonisevener (changes requested):
Thanks, these mostly look good, found one regression (commented on appropriate file). I also found some that weren't updated:

Screenshot 2025-01-24 at 3 25 44 PM

For the two factor and change temp, I had to fudge some code to even enter that screen.

I also found one double Done button situation that was missed, we should ask design for instruction here:

Settings > My Languages > Add another language so you have multiple > See Edit button, tap it:

Screenshot 2025-01-24 at 3 18 26 PM

Lastly, there's a series of other pieces of feedback, but they seem a bit more involved due to changing existing X buttons into a label. Maybe we can break that off into a separate task? I would like to prioritize them separately at least, for the chance that they will be deprioritized to a later release.

Screenshot 2025-01-24 at 3 22 36 PM

Article Gallery
Article TOC
Article References
Editor Find in Page

Wikipedia/Code/SinglePageWebViewController.swift
mazevedofs: Done!

mazevedofs:
> Thanks, these mostly look good, found one regression (commented on appropriate file). I also found some that weren't updated:
> 
 Updated those!
> 
> Settings > My Languages > Add another language so you have multiple > See Edit button, tap it:
> 
I was a little bold and made the lhs Cancel (since it cancels the actions) and the rhs Done as it applies the changes. I'll run it with Robin on design review as well.

> Lastly, there's a series of other pieces of feedback, but they seem a bit more involved due to changing existing X buttons into a label. Maybe we can break that off into a separate task? I would like to prioritize them separately at least, for the chance that they will be deprioritized to a later release.
> 
I chose not to update those here as it goes a little beyond the scope of the cancel/done updates on the new nav bar. I forgot to note it in the description. I'll create a separate ticket for it.

tonisevener (approved):
Looks good!

PR #5169, opened 2025-01-24
Standardize Headers: History

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T383826

### Notes
* 

### Test Steps
1. Enter several items in search to build up history
2. Navigate to history
3. Make sure scroll behavior works as expected
4. Ensure profile is there and usable
5. Check the same in landscape
6. Repeat in iPad

### DIFF
--- a/Wikipedia/Code/HistoryViewController.swift
+++ b/Wikipedia/Code/HistoryViewController.swift
@@ -9,6 +9,48 @@ class HistoryViewController: ArticleFetchedResultsViewController, WMFNavigationB
     
     var topSafeAreaOverlayHeightConstraint: NSLayoutConstraint?
     var topSafeAreaOverlayView: UIView?
+    
+    // Properties needed for Profile Button
+
+   private var _yirCoordinator: YearInReviewCoordinator?
+   var yirCoordinator: YearInReviewCoordinator? {
+
+       guard let navigationController,
+             let yirDataController,
+             let dataStore else {
+           return nil
+       }
+
+       guard let existingYirCoordinator = _yirCoordinator else {
+           _yirCoordinator = YearInReviewCoordinator(navigationController: navigationController, theme: theme, dataStore: dataStore, dataController: yirDataController)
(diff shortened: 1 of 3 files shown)


Wikipedia/Code/HistoryViewController.swift
```diff
@@ -137,8 +179,42 @@ class HistoryViewController: ArticleFetchedResultsViewController, WMFNavigationB
         }
         
         let hideNavigationBarOnScroll = !isEmpty
+        
+        let profileButtonConfig: WMFNavigationBarProfileButtonConfig?
+        if let dataStore {
+            profileButtonConfig = self.profileButtonConfig(target: self, action: #selector(userDidTapProfile), dataStore: dataStore, yirDataController: yirDataController, leadingBarButtonItem: nil, trailingBarButtonItem: nil)
```
tonisevener, line 185: Code so far looks good - I noticed the Clear button disappeared though, unfortunately. I did a bit of digging and for mysterious reasons of the past that [logic](https://github.com/wikimedia/wikipedia-ios/blob/b2a204ed71df3a33f25bc3e6b5e67cd4a62c9b65/Wikipedia/Code/ArticleFetchedResultsViewController.swift#L64) is buried in the superclass `ArticleFetchedResultsViewController`, but only this subclass takes advantage of it (due to it setting `isDeleteAllVisible` = true).

Can you move the Clear button-related logic out of `ArticleFetchedResultsViewController` and into `HistoryViewController`? Then on this line you can hopefully set the `leadingBarButtonItem` parameter to a clear button lazy property on self.

tonisevener (changes requested):

Wikipedia/Code/HistoryViewController.swift
l-olson1214: Got it!

Went through a small journey to get to the ultimately few line answer, but really awesome catch! Thanks :)

tonisevener (approved):
Working great!

PR #5173, opened 2025-01-27
Remove duplicate languages

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T384073

### Notes
This was a bug I missed in #5112, apparently some of the sites returned in the sites API call are duplicates. I added a temporary workaround to the script to avoid adding these new dupe languages, to get this back to the `7.6.4` state. But eventually we might want to perform a proper user data migration on these if we are expected to switch wiki language codes / subdomains.

When the script ran, a new language bubbled up as well, tig.wikipedia.org. I will note it on the task.

### Test Steps
1. Look and findings in https://phabricator.wikimedia.org/T384073#10493845. Confirm dupe languages no longer appear (except Cantonese, which is an existing bug).


mazevedofs (approved):
Looks good!

PR #5176, opened 2025-01-29
T374010

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T374010

### Notes
* Robin changed design to use "Done" instead of  X button

### Test Steps
1. Either using IP or Temporary Accounts, or switching the statement
2. Navigate to an article
3. Press edit
4. Ensure pop up shows up, repeat with other type of account 

### Screenshots/Videos
Simulator Screenshot - iPhone 15 Pro - 2025-01-29 at 14 32 07
Simulator Screenshot - iPhone 15 Pro - 2025-01-29 at 14 32 47


Wikipedia/Code/EditorViewController.swift
```diff
@@ -120,6 +147,95 @@ final class EditorViewController: UIViewController, WMFNavigationBarConfiguring
     
     // MARK: - Private Helpers
     
+    private func tempEditorSheet() {
+        let vm = WMFTempAccountsSheetViewModel(
+            image: "pageMessage",
+            title: WMFLocalizedString("temp-account-edit-sheet-title", value: "You are using a temporary account", comment: "Temporary account sheet for editors"),
+            subtitle: tempEditorSubtitleString(tempUsername: "fakeUsername"),
```
tonisevener, line 154: Let's pipe in the actual temporary account name here. To do this:

1. Make a new method in Session (you can use [this method](https://github.com/wikimedia/wikipedia-ios/blob/f1b12babb624fa79c14848fa9387cfef690ac303/Wikipedia/Code/Session.swift#L90-L101) as reference) that returns the `centralauth_User` cookie value.
2. In [WMFAuthenticationManager](https://github.com/wikimedia/wikipedia-ios/blob/f1b12babb624fa79c14848fa9387cfef690ac303/Wikipedia/Code/WMFAuthenticationManager.swift#L107), add a `authStateTemporaryUsername` computed var. Unfortunately it looks like the `currentUserCache` isn't always populated for temp accounts, so I think the more reliable implemenation would be something like:

```
public var authStateTemporaryUsername: String? {
        if authStateIsTemporary {
              return session.{your method name from step 1}
       }
}
```

3. Reference that new auth method here in EditorViewController to get the temp username: `authManager.authStateTemporaryUsername`

Wikipedia/Code/EditorViewController.swift
```diff
@@ -103,14 +123,21 @@ final class EditorViewController: UIViewController, WMFNavigationBarConfiguring
...
+        
         setTextSizeInAppEnvironment()
         setupFocusNavigationView()
         setupNavigationItemController()
         setupSpinner()
         apply(theme: theme)
         
         loadContent()
+        
+        let user = authManager.user(siteURL: pageURL)
```
tonisevener, line 135: In testing this, unfortunately it looks like authManager's user cache availability is flakier than I thought for IP and Temp accounts. This user value is often `nil` for me, so your `authState` checking doesn't work. Can you modify to something like this?

```
if authManager.permanentUser(siteURL: pageURL) == nil {
       if authManager.authStateIsTemporary { // This auth state method digs deeper into cookies, which should be more reliable
           presentTempEditorSheet()
       } else {
           presentIPEditorSheet()
       }
}
```

To test:

1. Log out of app. Close app, delete from simulator.
2. Run app again from Xcode
3. Tap edit pencil on a Test Wiki article, confirm IP modal shows.
4. Publish edit.
5. Tap edit pencil again on Test Wiki article. Confirm Temp modal shows.
6. Log in with real account, go back to editor, confirm no modal shows.

Wikipedia/Code/EditorViewController.swift
```diff
@@ -103,14 +123,21 @@ final class EditorViewController: UIViewController, WMFNavigationBarConfiguring
...
-
+        
         setTextSizeInAppEnvironment()
         setupFocusNavigationView()
         setupNavigationItemController()
         setupSpinner()
         apply(theme: theme)
         
         loadContent()
+
```
tonisevener, line 134: There's several modals that can appear when the user lands on editor, so we need to be careful when each one is triggered so that they don't step on each other. All of that logic is located [here](https://github.com/wikimedia/wikipedia-ios/blob/f1b12babb624fa79c14848fa9387cfef690ac303/Wikipedia/Code/EditorViewController.swift#L237-L244).

Can you instead trigger this modal from there? Something like:

```
           ...
           } else if let otherError = wikitextFetchResponse.otherError {
                WMFAlertManager.sharedInstance.showErrorAlertWithMessage(otherError.messageHtml.removingHTML, sticky: false, dismissPreviousAlerts: true)
                isDifferentErrorBannerShown = true
            } else if {conditional check on if temp account modal should show} {
               presentTempEditorSheet()
                isDifferentErrorBannerShown = true
            }  else if {conditional check on if ip account modal should show} {
               presentIPEditorSheet()
                isDifferentErrorBannerShown = true
            }
```

tonisevener (changes requested):
Great work! I have some suggestions on some auth state / modal presentation gotchas. I also noticed we might want to add a scroll view for landscape on iPhone:

Screenshot 2025-01-30 at 4 47 43 PM

Wikipedia/Code/EditorViewController.swift
```diff
@@ -244,6 +355,12 @@ final class EditorViewController: UIViewController, WMFNavigationBarConfiguring
             } else if let otherError = wikitextFetchResponse.otherError {
                 WMFAlertManager.sharedInstance.showErrorAlertWithMessage(otherError.messageHtml.removingHTML, sticky: false, dismissPreviousAlerts: true)
                 isDifferentErrorBannerShown = true
+            } else if authManager.permanentUser(siteURL: pageURL) == nil {
```
tonisevener, line 358: Code here looks great. I came across this complaint in the console:

1. While logged out, set your primary app language to Test Wiki, visit article `Test4/test` (it's in the top read card on the Explore feed). Tap edit pencil on first section. Console displays this error:

```
Attempt to present <Wikipedia.EditNoticesViewController: 0x101669ac0> on <WMFComponents.WMFComponentNavigationController: 0x1020d2800> (from <Wikipedia.EditorViewController: 0x1058ca600>) which is already presenting <_TtGC7SwiftUI19UIHostingControllerV13WMFComponents24WMFTempAccountsSheetView_: 0x1020d8800>.
```

This indicates that multiple modals are trying to present at the same time. What's happening is that this particular article has an "Edit notice", and the presentation code below on line 366 is _also_ getting triggered at the same time as your new modal.

Can you clean up this existing code a bit by adding the edit notice logic (lines 366 - 373) as an `else if` statement above this conditional? This way both aren't triggered at the same time.

Wikipedia/Code/EditorViewController.swift
```diff
@@ -120,6 +140,97 @@ final class EditorViewController: UIViewController, WMFNavigationBarConfiguring
...
+        }
+    }
+    
+    func tempEditorSubtitleString(tempUsername: String) -> String {
+        let openingLink = "<a href=\"\(ipURL)\">"
+        let closingLink = "</a>"
+        let openingBold = "<b>"
+        let closingBold = "</b>"
+        let lineBreaks = "<br/><br/>"
+        let format = WMFLocalizedString("temp-account-edit-sheet-subtitle", value: "Your edit will be attributed to %2$@%1$@%3$@. %4$@IP address%5$@ will be visible to administrators.%6$@If you log in or create an account, your edits will be attributed to a name you choose, among other benefits.",
```
tonisevener, line 184: New temp name insert is working well. I notice a missing "Your" before the IP address link:

Screenshot 2025-02-03 at 12 51 13 PM

tonisevener (changes requested):
Just a couple of more things I found, thanks!

Wikipedia/Code/EditorViewController.swift
l-olson1214: Oh good catch! Thanks!!

Wikipedia/Code/ArticleViewController+Editing.swift
```diff
@@ -77,23 +103,128 @@ extension ArticleViewController {
...
+                
+            } else {
                 self.present(navigationController, animated: true)
             }
-            editingWelcomeViewController.apply(theme: theme)
-            present(editingWelcomeViewController, animated: true) {
-                UserDefaults.standard.didShowEditingOnboarding = true
+        }
+        
+        if authManager.permanentUser(siteURL: articleURL) == nil {
```
tonisevener, line 125: I found one more weird issue with auth state stuff. This method checks for a permanentUser against the specific wiki (in my testing Test Wiki), but I didn't have a permanent state under Test Wiki in the user cache (I logged in under EN Wiki, my primary app language). So even though the app was logged in, I saw a logged out modal.

An easy fix is to just replace this line with `if !authManager.authStateIsPermanent {`. This is a better method in that it checks for a permanent status against ALL the user cache (which includes primary app wiki).

Wikipedia/Code/ArticleViewController+Editing.swift
```diff
@@ -77,23 +103,128 @@ extension ArticleViewController {
...
+            if authManager.authStateIsTemporary {
+                presentTempEditorSheet()
```
tonisevener, line 127: So far in these modal cases, you don't have a callback to run the `presentEditorAction` after they have dismissed the modals (Done or Swipe Down). So merging as it is means temporary and anonymous users can never reach editor.

If you want I can merge as-is since it's going in a feature branch and you can fix in a later PR. We have https://phabricator.wikimedia.org/T374015 which could cover fixing this hole, or you can do it as a part of https://phabricator.wikimedia.org/T374012 (I think testing editor for that task will be blocked unless you fix this navigation).

Wikipedia/Code/EditorViewController.swift
```diff
@@ -79,6 +80,10 @@ final class EditorViewController: UIViewController, WMFNavigationBarConfiguring
         return spinner
     }()
     
+    internal var authManager: WMFAuthenticationManager {
```
tonisevener, line 83: Doesn't look like this is used anymore, but I you'll probably want it for future tasks so I'm good with leaving it here.

tonisevener (changes requested):
One last must-fix and a couple of optional fixes. Thanks!

Wikipedia/Code/ArticleViewController+Editing.swift
```diff
@@ -76,24 +102,142 @@ extension ArticleViewController {
     }
     
     private func presentEditor(editorViewController: UIViewController) {
-        
-        let navigationController = WMFComponentNavigationController(rootViewController: editorViewController, modalPresentationStyle: .overFullScreen)
-        
-        let needsIntro = !UserDefaults.standard.didShowEditingOnboarding
-        if needsIntro {
-            let editingWelcomeViewController = EditingWelcomeViewController(theme: theme) {
+        let presentEditorAction = {
```
tonisevener, line 105: Can you update this to:

```
let presentEditorAction = { [weak self] in
      guard let self else { return }
      ...
}
```

I noticed the new temp accounts views weren't deallocating and this might help.

Wikipedia/Code/ArticleViewController+Editing.swift
```diff
@@ -76,24 +102,142 @@ extension ArticleViewController {
...
+                    
+                    let webVC: SinglePageWebViewController
+                    
+                    let config = SinglePageWebViewController.StandardConfig(url: url, useSimpleNavigationBar: true)
+                    webVC = SinglePageWebViewController(configType: .standard(config), theme: self.theme)
+                    
+                    let newNavigationVC = WMFComponentNavigationController(rootViewController: webVC, modalPresentationStyle: .formSheet)
+                    presentedViewController.present(newNavigationVC, animated: true)
+                },
+                didTapDone: {
```
tonisevener, line 159: also update this to:

```
didTapDone: { [weak self] in
    self?.dismiss(animated: true)
    presentEditorAction()
})
```

Wikipedia/Code/ArticleViewController+Editing.swift
```diff
@@ -76,24 +102,142 @@ extension ArticleViewController {
...
+                    hostingController?.dismiss(animated: true)
+                    presentEditorAction()
```
tonisevener, line 161: Sorry, looking at this with fresh eyes I realize maybe "Done" _shouldn't_ present the editor and only the "Continue without logging in" button should, but we can remove this line later after design review if there's an issue. I do prefer that navigation happens on the UIKit-side at any rate, so that we can (eventually) move navigation calls like this into a coordinator.

tonisevener (changes requested):
Just a couple of tiny tweaks for retain cycles, thanks!

Wikipedia/Code/ArticleViewController+Editing.swift
l-olson1214: Kicking off a build for Robin - we can have him check that prior to merging this :)

tonisevener (approved):

PR #5184, opened 2025-02-04
T383683: Add tests to check for multiple or missing semicolons in strings file.

chandra9302:
**Phabricator:** 
T383683

### Notes
* Added unit tests in TWNStrings.tests.m that:

1. Loop through all the Wikipedia/Localizations/{langcode}.lproj/Localizable.strings files and check for the existence of multiple semicolons at the end of a line. Fail if more than 1 appear on a line.
2. Loop through all the Wikipedia/Localizations/{langcode}.lproj/Localizable.strings files and check for the lack of semicolon at the end of a line. Fail if a semicolon is missing.


WikipediaUnitTests/Code/TWNStringsTests.m
```diff
@@ -426,6 +450,84 @@ - (void)testIncomingTranslationStringForBracketSubstitutionsNotPresentInEN {
...
+                unichar character = [trimmedLine characterAtIndex:i];
+                if (character == ';') {
+                    semicolonCount++;
+                } else {
+                    break; // Stop counting once a non-semicolon character is found
+                }
+            }
+
+            // Fail if multiple semicolons exist at the end of the line
+            XCTAssertLessThanOrEqual(semicolonCount, 1, @"File %@ has a line with multiple semicolons at the end: %@", filePath, trimmedLine);
```
tonisevener, line 489: Hi @chandra9302 

After looking at this, I wonder if we could reduce this to just a single test and change this to `XCTAssertEqual`. I think the additional test might be unnecessary. I updated the Phabricator task description. Thanks!

chandra9302: @tonisevener I updated the tests to have single test to check if there's only one semi-colon at the end.

tonisevener: Looks good, thanks!

tonisevener (approved):

PR #5185, opened 2025-02-04
System Nav Bar - iPad improvements for Explore and Article

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T383471
and https://phabricator.wikimedia.org/T383837

### Notes
This PR adds some improvements for iPadOS 18.

**Left-align Explore logo**
I have less control over alignment of the title for iPadOS 18. I cannot assign it to `leftBarButtonItem` like I am with the other devices, because then the logo will show way up in the top, next to the floating tab bar. So to get around this, for iPadOS18-only, I am using a large leading title with a custom font. That custom font has the logo embedded into a special unicode character, which I am setting as the title. I believe I have dealt with the fallout effects (luckily we can explicitly set a `backButtonTitle`, so that special character doesn't show up there). It's a bit hacky but so far appears to be working.

**Remove custom W icon on Article**
After some collaboration with design, we decided to remove the navigation bar (with the custom W) logo entirely on iPadOS 18. It removes unnecessary spacing, and once we fix https://phabricator.wikimedia.org/T383836, the W icon behavior will become redundant.

### Test Steps
1. Test Explore and Article navigation bar on iPadOS 18. Ensure everything works smoothly.

### Screenshots
Screenshot 2025-02-04 at 5 29 24 PM
Screenshot 2025-02-04 at 5 30 46 PM


mazevedofs (approved):
Looks great!

PR #5187, opened 2025-02-05
Article source logging

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T382032

### Notes
* Adds `source` to ArticleViewController to allow source logging
* Only added for places, history and search
* I'm sending this property to navigate(to:userInfo:) in the user info dictionary
* In some places, we're sending `0` or `undefined`, since we're not covering all sources. In this case, we don't log, the backend has no case for 0

### Test Steps
1. Set log level to `all` on `WMFLogging.h`
2. Browse the app, click on search results, history items and read from places. Verify the logs
3. Verify if the events are posted correctly to MEP
4. Check if the events are not being sent from incorrect sources.


Wikipedia/Code/ArticleCollectionViewController.swift
```diff
@@ -181,8 +182,9 @@ extension ArticleCollectionViewController {
             collectionView.deselectItem(at: indexPath, animated: true)
             return
         }
-
-        navigate(to: articleURL)
+        let userInfo: [AnyHashable:Any]? = [ArticleSourceUserInfoKeys.articleSource: ArticleSource.history.rawValue]
```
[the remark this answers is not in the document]
tonisevener: From testing, looks it causes Explore > More top read > Tap article to log as being from `.history`.

Wikipedia/Code/ArticleLocationCollectionViewController.swift
```diff
@@ -192,8 +192,9 @@ extension ArticleLocationCollectionViewController {
 // MARK: - CollectionViewContextMenuShowing
 extension ArticleLocationCollectionViewController: CollectionViewContextMenuShowing {
     func articleViewController(for indexPath: IndexPath) -> ArticleViewController? {
+        
         let articleURL = articleURL(at: indexPath)
-        let articleViewController = ArticleViewController(articleURL: articleURL, dataStore: dataStore, theme: theme)
+        let articleViewController = ArticleViewController(articleURL: articleURL, dataStore: dataStore, theme: theme, source: .places)
```
tonisevener, line 197: Looks like this method is only called when you peek into an article. I wonder if we should just use .undefined for peeks?

I also wanted to call out that ArticleLocationCollectionViewController is used in places but also in this Explore card > Nearby list view:
Screenshot 2025-02-06 at 4 06 01 PM
Screenshot 2025-02-06 at 4 06 17 PM

So we might be logging as being from places but it's actually from this screen.

Wikipedia/Code/PlacesViewController.swift
```diff
@@ -1736,7 +1736,8 @@ class PlacesViewController: ArticleLocationCollectionViewController, UISearchBar
         }
         switch action {
         case .read:
-            navigate(to: url)
+            let userInfo: [AnyHashable: Any]? = [ArticleSourceUserInfoKeys.articleSource: ArticleSource.places.rawValue]
```
tonisevener, line 1739: This is working when coming from the map. I don't see it working when tapping from the list view though:

Screenshot 2025-02-06 at 4 21 06 PM

For both iPad and iPhone. Also just note that iPhone and iPad list views are two separate views, so be sure to test on both.

Wikipedia/Code/ArticleViewController.swift
```diff
@@ -632,9 +634,9 @@ class ArticleViewController: ThemeableViewController, HintPresenting, UIScrollVi
             let siteURL = self.articleURL.wmf_site,
                   let project = WikimediaProject(siteURL: siteURL) {
                 ArticleLinkInteractionFunnel.shared.logArticleView(pageID: pageID.intValue, project: project)
```
tonisevener, line 636: I'm unsure about how we should treat article peek and pops. I missed this case when logging the original page views.

In testing, this area gets called once when long pressing > article preview is displayed. Tapping further to go into the full screen article does not log this again. Backing out of peek is not ideal either, because we logged it as a page view when really they didn't read the full view of the article.

Maybe a temporary measure (if product is ok with it) is to wrap this up in a conditional that I found:

```
if !isBeingPresentedAsPeek {
    {log page view}
}
```

So that we're only logging when they don't come in through the long press.

tonisevener (changes requested):

Wikipedia/Code/ArticleViewController.swift
```diff
@@ -171,15 +171,20 @@ class ArticleViewController: ThemeableViewController, HintPresenting, UIScrollVi
     private var tocStackViewTopConstraint: NSLayoutConstraint?
     private var searchBarIsAnimating = false
 
-    convenience init?(articleURL: URL, dataStore: MWKDataStore, theme: Theme, schemeHandler: SchemeHandler? = nil, altTextExperimentViewModel: WMFAltTextExperimentViewModel, needsAltTextExperimentSheet: Bool, altTextBottomSheetViewModel: WMFAltTextExperimentModalSheetViewModel?, altTextDelegate: AltTextDelegate?) {
-        self.init(articleURL: articleURL, dataStore: dataStore, theme: theme)
+    private var articleViewSource: ArticleSource
+
+    // Will be populated if needsSearchBar = false
+    private var searchBarButtonItem: UIBarButtonItem?
```
tonisevener, line 177: Might be an unexpected change here?

Wikipedia/Code/PlacesViewController.swift
tonisevener: @mazevedofs Still not seeing the source when coming from the iPhone list view (works on iPad though).

Wikipedia/Code/ArticleViewController.swift
tonisevener: Let's confirm with Product that it's ok for us to suppress this.

Wikipedia/Code/ArticleViewController.swift
```diff
@@ -354,12 +360,6 @@ class ArticleViewController: ThemeableViewController, HintPresenting, UIScrollVi
         return view
     }()
     
-    lazy var searchBarButtonItem: UIBarButtonItem = {
```
tonisevener, line 357: Unexpected change here too.

tonisevener (changes requested):

Wikipedia/Code/PlacesViewController.swift
```diff
@@ -114,7 +114,7 @@ class PlacesViewController: ArticleLocationCollectionViewController, UISearchBar
 
     override func viewDidLoad() {
 
-        listViewController = ArticleLocationCollectionViewController(articleURLs: [], dataStore: dataStore, contentGroup: nil, theme: theme)
+        listViewController = ArticleLocationCollectionViewController(articleURLs: [], dataStore: dataStore, contentGroup: nil, theme: theme, source: .places)
```
tonisevener, line 117: @mazevedofs 

(Reason for this bug - https://github.com/wikimedia/wikipedia-ios/pull/5187/files#r1946990495)

You'll also want to set `self.source = .places` (if you can, may need to set the superclass property as mutable. That's how you can target the iPhone list view.

mazevedofs: Do you mean on ArticleVC? Or from ArticleLocationCollectionVC?

Wikipedia/Code/ArticleViewController.swift
```diff
@@ -663,7 +664,7 @@ class ArticleViewController: ThemeableViewController, HintPresenting, UIScrollVi
               let project = WikimediaProject(siteURL: siteURL) {
             
             if !isBeingPresentedAsPeek {
-                ArticleLinkInteractionFunnel.shared.logArticleView(pageID: pageID.intValue, project: project)
+                ArticleLinkInteractionFunnel.shared.logArticleView(pageID: pageID.intValue, project: project, source: articleViewSource)
```
tonisevener, line 667: Can you add the source below as well, in `logPageViewAfterArticleAppearance`?

tonisevener (changes requested):
I notice some sources are missing when long pressing / peeking and visiting an article from Places, History and Search. I think my comment might fix it a little, but I still see missing sources after making that change.

Wikipedia/Code/ArticleViewController.swift
mazevedofs: Totally missed that, thanks for the catch!

Wikipedia/Code/HistoryViewController.swift
```diff
@@ -165,7 +165,26 @@ class HistoryViewController: ArticleFetchedResultsViewController, WMFNavigationB
...
+            return nil
+        }
+
+        guard let articleURL = articleURL(at: indexPath) else {
+            return nil
+        }
+
+        articleVC.articleViewSource = .history
+        articleVC.articlePreviewingDelegate = self
+        articleVC.wmf_addPeekableChildViewController(for: articleURL, dataStore: dataStore, theme: theme)
```
tonisevener, line 184: I think you should be able to delete these two lines, the superclass call on 170 above should have already set these up.

Wikipedia/Code/SearchResultsViewController.swift
```diff
@@ -88,7 +88,26 @@ class SearchResultsViewController: ArticleCollectionViewController {
...
+            return nil
+        }
+
+        guard let articleURL = articleURL(at: indexPath) else {
+            return nil
+        }
+
+        articleVC.articleViewSource = .search
+        articleVC.articlePreviewingDelegate = self
+        articleVC.wmf_addPeekableChildViewController(for: articleURL, dataStore: dataStore, theme: theme)
```
tonisevener, line 107: Delete these lines if you can.

tonisevener (approved):

PR #5200, opened 2025-02-13
Hide floating tab bar on iPad when scrolling

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T383759

### Notes
* Works on current UIKit implementation

### Test Steps
1. Scroll up and down on Article View, Explore, Saved, Search and History.
2. Make sure it behaves as expected
3. Check if iPhone is not affected


tonisevener (approved):
I notice the tab bar can get "stuck" in a hidden state when the content is short. I faced a similar issue with the navigation bar hiding - you can reference `calculateNavigationBarHiddenState` for how I fixed. Not a must-fix though, in my opinion, as long as design review is good with it.

https://github.com/user-attachments/assets/355dbef4-7cc2-4ec5-b406-f18be4087812

l-olson1214 (approved):

PR #5202, opened 2025-02-15
Replaces "Search" title in tab bar with magnifying glass icon only on  iPad with OS version 18 and above

chandra9302:
**Phabricator:** 
T383831

### Notes
* The change in this PR is specific to iPad with OS version 18 and above.

### Test Steps
1. Launch app on iPad with OS version 18 and above to observe new magnifying icon for search instead of text.

### Screenshots/Videos
Simulator Screen Recording - iPad Pro 13-inch (M4) - 2025-02-14 at 22 09 54

### DIFF
--- a/Wikipedia/Code/WMFAppViewController.m
+++ b/Wikipedia/Code/WMFAppViewController.m
@@ -30,6 +30,23 @@ typedef NS_ENUM(NSUInteger, WMFAppTabType) {
     WMFAppTabTypeSearch = 4
 };
 
+NSString *NSStringFromWMFAppTabType(WMFAppTabType tabType) {
+    switch (tabType) {
+        case WMFAppTabTypeMain:
+            return @"Main";
+        case WMFAppTabTypePlaces:
+            return @"Places";
+        case WMFAppTabTypeSaved:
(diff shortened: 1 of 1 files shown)


tonisevener (changes requested):
@chandra9302 Thanks for the PR! Code looks good to me so far. I will put this through design review.

I did find a couple of bugs when clicking around:

1. Automatic theming. On iPadOS 18, in app Settings choose "Default" theme. Background, in iOS settings switch light or dark appearance mode. When you return the content does not auto-theme:
Screenshot 2025-02-20 at 2 22 49 PM

2. When in split screen mode, the bottom tab bar loses its icons, and when I tap one I see a crash:
Screenshot 2025-02-20 at 2 23 58 PM

tonisevener:
Hi @chandra9302! Updating you that this passed design review. Please address the bugs found during code review (see https://github.com/wikimedia/wikipedia-ios/pull/5202#pullrequestreview-2631074738), then this should be good to merge. Thank you!

mazevedofs:
Hi @chandra9302, were you able to address the comment above? Let us know if you need any help.

PR #5211, opened 2025-02-27
Add coordinators to navigate to articles

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T386373

### Notes
There are two large refactors here.

1. All navigation paths to articles should now occur through a coordinator. We have three new coordinators here - ArticleCoordinator, RandomArticleCoordinator, and LinkCoordinator.

- **ArticleCoordinator:** Pretty simple, spins up an ArticleViewController and tells the navigation controller to push it.
- **RandomArticleCoordinator:** A little more complicated. If instantiated with an `articleURL`, it pushes `RandomArticleViewController`. This was an existing `ArticleViewController` subclass that adds the "Random" button at the bottom. If instantiated with a `siteURL`, it pushes `WMFFirstRandomViewController`. This was an existing view controller that first fetches a random article model, then asks it's navigation controller to replace itself with a RandomArticleViewController. This path is used from the "Another random article" button from the Explore feed, where we don't know what the particular random article will be yet.
- **LinkCoordinator**: This coordinator is meant to take a URL of any kind and determine the best specific coordinator to call. It will be useful when tapping on any link from articles or talk pages, or when deep linking. It will reuse a lot of our old logic from Router.swift and ViewControllerRouter.swift. For now I just put in the bare-minimum to detect an article URL. 

To keep this change from getting too big, I am falling back to the legacy navigation logic if these coordinators fail to navigate. To indicate failure, I am returning a success flag from the coordinators' `start()` method. Callers can choose to reference that flag or not.

2. "All navigation paths" in part 1 above include long pressing and previewing an article, then tapping the preview or the "Read now" context menu item. In the previous behavior of article previewing, an actual ArticleViewController instance was created upon previewing, and a smaller `ArticlePeekPreviewViewController` was instantiated and embedded on top of ArticleViewController as a child view controller. I decoupled this previewing logic from ArticleViewController. Now when you long press an article link or cell, only a `ArticlePeekPreviewViewController` will be instantiated and used as the preview. Then tapping the preview or the "Read now" context menu will kick off an `ArticleCoordinator`.

### Test Steps
1. Tap around the app, confirm all links to articles still push on an article. Ensure long pressing an article cell or link still works as it did before.
2. Confirm deep linking into an article from Safari still works as it did before.


Wikipedia/Code/ExploreViewController.swift
```diff
@@ -600,14 +600,27 @@ class ExploreViewController: ColumnarCollectionViewController, ExploreCardViewCo
...
+        if !useRandomArticlePreviewItem {
             
-            if vc is WMFImageRecommendationsViewController {
-                ImageRecommendationsFunnel.shared.logExploreCardDidTapAddImage()
+            // first try random coordinator
+            if let navigationController,
+               group.contentGroupKind == .random,
+               let randomSiteURL = group.siteURL {
+                
+                // todo: accurate source
```
l-olson1214, line 610: Is this still a todo?

tonisevener: Thanks for catching that! Nope, just an old comment, I removed it.

PR #5225, opened 2025-03-18
Fix date formatting in TalkPageCellViewModel

Ytemiloluwa:
**Phabricator:** 
https://phabricator.wikimedia.org/T377793

### Notes
* Fixed the date formatting in `NSDateFormatter+WMFExtensions` to respect the device's locale settings for date and time. 
@l-olson1214 @mazevedofs 

### Test Steps
1. Open a talk page in the app.
2. Navigate to a specific conversation.
3. Verify that the date displayed below the header matches the device's locale settings.


Wikipedia/Code/TalkPageCellViewModel.swift
```diff
@@ -45,7 +52,7 @@ final class TalkPageCellViewModel: Identifiable {
         self.topicTitleHtml = topicTitleHtml
         self.timestamp = timestamp
         if let timestamp = timestamp {
-            self.timestampDisplay = dateFormatter?.string(from: timestamp)
+            self.timestampDisplay = Self.localizedDateFormatter.string(from: timestamp)
```
mazevedofs, line 55: The dateFormatter used here is passed into each TalkPageCellViewModel at creation, in this spot:
https://github.com/wikimedia/wikipedia-ios/blob/e377daba2412cef246455d2ec7c6e3ffd5edf780/Wikipedia/Code/TalkPageViewModel.swift#L227

It's instantiated on TalkPageViewModel's init here https://github.com/wikimedia/wikipedia-ios/blob/e377daba2412cef246455d2ec7c6e3ffd5edf780/Wikipedia/Code/TalkPageViewModel.swift#L70

If you check that method, it's already trying to retrieve the Locale
https://github.com/wikimedia/wikipedia-ios/blob/e377daba2412cef246455d2ec7c6e3ffd5edf780/Wikipedia/Code/TalkPageViewModel.swift#L173

You could debug this method to understand why it's not returning the correct formatt here - no need to create a new method, we already have an extension that turns this format
https://github.com/wikimedia/wikipedia-ios/blob/e377daba2412cef246455d2ec7c6e3ffd5edf780/Wikipedia/Code/TalkPageViewModel.swift#L178

mazevedofs (changes requested):
Hi @Ytemiloluwa, thanks for your contribution.
I left a comment with a suggestion for another approach that reuses a formatted we already have.

Wikipedia/Code/TalkPageCellViewModel.swift
Ytemiloluwa: Hi @mazevedofs,
Thank you for your feedback and for pointing out the existing dateFormatter that is already being passed into the TalkPageCellViewModel. I appreciate your suggestion to reuse the existing formatter instead of creating a new one.
I will review the dateFormatterForSiteURL method to understand why it may not be returning the correct format and see how I can adjust my implementation to utilize the existing formatter effectively.
Thanks again for your guidance!

Ytemiloluwa:
I have debugged the `wmf_localCustomShortDateFormatterWithTimeForLocale:` method. the issue was that the locale used was based on the wiki article's language. The update now use the device's formatting settings (DMY/24-hour) which is configurable in the app's settings. @mazevedofs 

I have also updated the screenshot as a backup to my claim. I await your feedback, thank you.

Wikipedia/Code/NSDateFormatter+WMFExtensions.m
```diff
@@ -118,8 +118,10 @@ + (NSDateFormatter *)wmf_longDateFormatter {
 + (instancetype)wmf_localCustomShortDateFormatterWithTimeForLocale:(NSLocale *)locale {
     NSDateFormatter *formatter = [[NSDateFormatter alloc] init];
     [formatter setTimeZone:[NSTimeZone localTimeZone]];
-    [formatter setLocale: locale];
-    [formatter setLocalizedDateFormatFromTemplate:@"hh:mm a, dd.MM.yyyy"];
+    [formatter setLocale:[NSLocale currentLocale]];
```
mazevedofs, line 121: This function is used in other parts of the app, and sometimes the locale used is based on the wiki language, like here, for fetching blocks per wiki
https://github.com/wikimedia/wikipedia-ios/blob/11ab7541800a62470c761cae85ae461c8de87ec7/WMF%20Framework/Fetcher.swift#L288

Because of this, we need to use the `locale` argument passed in the function signature, and not use the system's `NSLocale currentLocale`

On this line, please pass `locale`.

Then, on TalkPageViewModel, you'll need to pass `Locale.current` here

https://github.com/wikimedia/wikipedia-ios/blob/11ab7541800a62470c761cae85ae461c8de87ec7/Wikipedia/Code/TalkPageViewModel.swift#L178

So we don't create regression bugs in other parts of the app.

Let me know if you need any help!

mazevedofs (changes requested):

Wikipedia/Code/NSDateFormatter+WMFExtensions.m
Ytemiloluwa: Hello @mazevedofs I have made the requested changes.

mazevedofs (approved):
Looks good!

PR #5229, opened 2025-03-21
Check temp account availability per wiki

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T387131

### Notes
* Adds checks for temp account availability upon app resume for main app language and upon preferred languages update
* Adds check for temp account status after editing
* I wasn't sure about how often should we check for main wiki status, suggestions welcome here

### Test Steps
1. In a logged-out state, before editing, verify UI is only logged out on profile and settings
2. In a pilot wiki, make an edit
3. Go back to profile and settings, make sure UI updates
4. Change the main app language to a non-pilot wiki
5. Verify UI updates to logged out state on profile and settings
6. Change back to pilot wiki
7. Verify that the UI updates accordingly on profile and settings


Wikipedia/Code/EditSaveViewController.swift
```diff
@@ -413,6 +413,16 @@ class EditSaveViewController: WMFScrollViewController, Themeable, UITextFieldDel
     }
     
     private func handleEditSuccess(with result: [AnyHashable: Any]) {
+
+        guard let languageCode else { return }
+        let dataController = WMFTempAccountDataController.shared
+        dataController.checkWikiTempAccountAvailability(language: languageCode, isCheckingPrimaryWiki: false)
```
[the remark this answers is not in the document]
mazevedofs: Thanks for the catch! I moved the check to the moment the user taps "edit". The loading state seems to work, but I guess it could use more testing. 
Let me know what you think!

Wikipedia/Code/ArticleViewController+Editing.swift
```diff
@@ -80,49 +80,60 @@ extension ArticleViewController {
...
-            
-            _ = tempAccountsCoordinator.start()
-        } else {
-            presentEditorAction()
+
+        state = .loading
+
+        Task {
+            let dataController = WMFTempAccountDataController.shared
+            dataController.checkWikiTempAccountAvailability(language: articleLanguageCode, isCheckingPrimaryWiki: false)
```
tonisevener, line 108: I'm still having trouble here. There's no completion block or await, how can you be sure the network call has completed before checking the `wikisWithTempAccountsEnabled ` property?

Try this:
- Ensure app is logged out. Add Test as a second language. Re-run from Xcode.
- Edit an EN Wiki article. Temp accounts sheet should not appear.
- Edit a Test Wiki article. Temp accounts sheet does not appear for me, but it should. This is because the first time the flag is not saved in-memory yet and you are checking property before the network call has finished.
- Edit a Test Wiki article again, temp accounts sheet appears.

tonisevener (changes requested):
I also see this UI on EN Wikipedia but I shouldn't. We should add the check before displaying it.

Screenshot 2025-04-01 at 8 56 48 AM
Screenshot 2025-04-01 at 8 56 57 AM

mazevedofs:
> I also see this UI on EN Wikipedia but I shouldn't. We should add the check before displaying it.
> 
> Screenshot 2025-04-01 at 8 56 48 AM

I added a temp account status check here, but this button is a little confusing to me.
The icon seems to be correct. The icon for IP accounts is this one we see, however, the IP account notice action also triggers the Temp Account Sheet:

https://github.com/wikimedia/wikipedia-ios/blob/d1acbf5e3f1196fbc9924a69977c9be2998d79fe/Wikipedia/Code/EditorViewController.swift#L836-L844

The logic here might work when you're editing a wiki with temp accounts deployed before and after you create a temp account, but this will always display an incorrect option for IP account wikis.
There's no option for IP accounts. I'm not sure what the correct behaviour should be here.

WMFData/Sources/WMFData/Store/WMFSharedCacheDirectoryNames.swift
```diff
@@ -5,4 +5,5 @@ enum WMFSharedCacheDirectoryNames: String {
     case watchlists = "Watchlists"
     case experiments = "Experiments"
     case developerSettings = "Developer Settings"
+    case tempAccountsWikis = "Temp Accounts Wikis"
```
tonisevener, line 8: I think you can remove this now.

WMFData/Sources/WMFData/Store/WMFUserDefaultsKey.swift
```diff
@@ -10,4 +10,5 @@ enum WMFUserDefaultsKey: String {
     case bypassDonation = "bypass-donation"
     case seenYearInReviewIntroSlide = "seen-year-in-review-intro-slide"
     case tappedYIR = "tapped-yir"
+    case primaryWikiHasTempAccounts = "primary-wiki-temp-accounts"
```
tonisevener, line 13: Might be able to remove this too.

tonisevener (approved):
Just a couple of minor things, but looks good to me! Feel free to merge.

PR #5230, opened 2025-03-21
Fix: Prevent notification count overflow when memory is low

anvithagowda098:
**Phabricator:**
https://phabricator.wikimedia.org/T385130 

### Notes
* Introduced a safe upper limit (Int.max / 10) to prevent integer overflow.
* Implemented a fallback mechanism:
* If fetching the unread count fails, it uses the last known valid value instead of displaying an incorrect count.
* Modified updateCacheWithCurrentUnreadNotificationsCount() to handle failures carefully.

### Test Steps
1. Open the Wikipedia app and navigate to the profile screen.
2. Fill up the device storage to simulate low-memory conditions (or use Xcode’s Simulate Memory Warning feature).
3. Check the notification count. It should correctly display unread notifications. No extreme values (e.g., 9223372036854775807) should appear.


mazevedofs (changes requested):
Hi @anvithagowda098,
Thanks for your contribution. I have a few suggestions to improve your current approach:
1. The current use of Int.max / 10 to limit the notification count masks the underlying memory issue. Since a realistic notification count for Wikimedia projects is much lower, displaying such a cap may hide the fact that there's an overflow or an error in the count computation when fetching this info.
2. Updating previousUnreadCount with the fetched count regardless of its correctness means that if the number is incorrect (e.g., due to overflow in this case), it will be stored and used as a fallback. We should update and cache it only if the count is correct.  

Tips for approaching this issue:

- After fetching the count, verify if it is within a realistic threshold (for instance, less than 1000). If the count exceeds this value, it likely indicates an underlying problem.

- If the count is abnormally high, log an error or try to re-check it. This will help identify and diagnose the root cause.

- Only update previousUnreadCount if the count passes the validation check. Otherwise, continue using the last known good value.

- Take a closer look at modelController.numberOfUnreadNotifications() to determine if the overflow is due to low memory conditions, a fetching error, or an NSNumber casting issue. 

Let me know if you have any questions or need any help.

WMF Framework/Remote Notifications/RemoteNotificationsController.swift
```diff
@@ -276,15 +276,38 @@ public enum RemoteNotificationsControllerError: LocalizedError {
     }
     
     /// Fetches a count of unread notifications from the local database. Uses the viewContext and must be called from the main thread
+    var previousUnreadCount: Int = 0
+    let maxNotificationThreshold = 1000
```
mazevedofs, line 280: This prop is not being used

WMF Framework/Remote Notifications/RemoteNotificationsController.swift
```diff
@@ -276,15 +276,38 @@ public enum RemoteNotificationsControllerError: LocalizedError {
...
         }
-        
-        let count = try modelController.numberOfUnreadNotifications()
-        return NSNumber(value: count)
+
+        do {
+            let fetchedCount = try modelController.numberOfUnreadNotifications()
+
+            // Ensure fetchedCount is a valid NSNumber
+            guard let number = fetchedCount as? NSNumber else {
```
mazevedofs, line 291: Casting 'Int' to 'NSNumber' always works, this check is unnecessary.

WMF Framework/Remote Notifications/RemoteNotificationsController.swift
```diff
@@ -276,15 +276,38 @@ public enum RemoteNotificationsControllerError: LocalizedError {
...
+                throw NSError(domain: "NotificationError", code: 500, userInfo: [NSLocalizedDescriptionKey: "Casting issue: fetchedCount is not an NSNumber"])
+            }
+
+            let count = number.intValue
+
+            // Validate the count before updating
+            if count < 0 || count > 1000 {
+                throw NSError(domain: "NotificationError", code: 501, userInfo: [NSLocalizedDescriptionKey: "Invalid notification count detected"])
```
mazevedofs, line 299: We don't use NSError directly like this in the project.
Add another case on the RemoteNotificationsControllerError enum to cover this case 
```suggestion
                throw RemoteNotificationsControllerError.invalidCount
```

WMF Framework/Remote Notifications/RemoteNotificationsController.swift
```diff
@@ -310,11 +333,23 @@ public enum RemoteNotificationsControllerError: LocalizedError {
...
-        let sharedCache = SharedContainerCache(fileName: SharedContainerCacheCommonNames.pushNotificationsCache)
-        var pushCache = sharedCache.loadCache() ?? PushNotificationsCache(settings: .default, notifications: [])
-        pushCache.currentUnreadCount = currentCount
-        sharedCache.saveCache(pushCache)
+        do {
+            let currentCount = try numberOfUnreadNotifications().intValue
+
+            // Same validation before updating cache
+            if currentCount < 0 || currentCount > 1000 {
+                NSLog("Skipping cache update due to unrealistic count: \(currentCount)")
```
mazevedofs, line 341: We use `DDLogError` here
```suggestion
                DDLogError("Skipping cache update due to unrealistic count: \(currentCount)")
```

WMF Framework/Remote Notifications/RemoteNotificationsController.swift
```diff
@@ -310,11 +333,23 @@ public enum RemoteNotificationsControllerError: LocalizedError {
...
+                return // Don't update cache
+            }
+
+            let sharedCache = SharedContainerCache(fileName: SharedContainerCacheCommonNames.pushNotificationsCache)
+            var pushCache = sharedCache.loadCache() ?? PushNotificationsCache(settings: .default, notifications: [])
+            pushCache.currentUnreadCount = currentCount
+            sharedCache.saveCache(pushCache)
+
+        } catch {
+            NSLog("Failed to update cache: \(error.localizedDescription)")
```
mazevedofs, line 351: Same as above about DDLogError

mazevedofs (changes requested):
While the current solution probably prevents erroneous display values, it still doesn’t address the underlying issue causing the overflow. If low memory conditions are the source, I believe it’s worth investigating why we get this int overflow to solve the root cause.

anvithagowda098:
Thanks for the thorough review @mazevedofs !
I have made all the suggested changes.
I’d really like to dig deeper and understand what’s causing the inflated notification count in the first place.
If you have any ideas on how I could start investigating, I’d really appreciate your guidance. I want to make sure we’re not just masking a bigger problem.

anvithagowda098:
@mazevedofs I've made requested changes

mazevedofs:
> @mazevedofs I've made requested changes

Hi @anvithagowda098! Thanks for addressing the changes! Were you able to investigate what caused the integer overflow in the first place? Did simulating a low-memory device help? Let me know if you were able to reproduce the bug in the first place - I understand it's a difficult scenario to simulate.

anvithagowda098:
@mazevedofs I have my university exams going on currently 
I will try to do it as soon as possible

anvithagowda098:
@mazevedofs Hey! I was able to successfully simulate a low memory environment with the help of the tutorial you provided. I couldn't reproduce the issue though. Either it's already resolved or doesn't occur under this condition. That said, this patch ensures that incorrect values are not displayed, offering a safeguard against similar scenarios.

mazevedofs:
@anvithagowda098 Hi! Thanks for trying that out! Were you able to simulate that from the main branch to see if the issues still exist without this fix?

anvithagowda098:
@mazevedofs  yes I did. I simulated it from main branch itself.

mazevedofs:
Hi @anvithagowda098,

Thank you for the effort you put into this task and for submitting a pull request. Since we’ve been unable to consistently reproduce the issue, we can’t confidently confirm whether the proposed solution addresses the problem. And while the changes may be well-intentioned, adding unverified code can introduce unintended side effects.

For now, I’ll be closing this PR. If the issue reappears or becomes easier to reproduce in the future, we’ll gladly revisit it.

Thanks again for your time and contributions!

PR #5235, opened 2025-03-31
Title Edit / Edit Description

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T340790

### Notes
* (Ran out of temp accounts to use for video - sorry)

### Test Steps
1. Fresh install
2. Open article with title
3. Press edit button, ensure toast appears
4. Go through, create temp account
5. Ensure new temp account toast appears after publish
6. Repeat 2-3, ensure different toast with new information

### Screenshots/Videos

https://github.com/user-attachments/assets/4d1f5071-aa9b-4bad-ac85-506d1db316ec


Wikipedia/Code/ArticleViewController+Editing.swift
```diff
@@ -57,26 +57,34 @@ extension ArticleViewController {
...
                 UserDefaults.standard.wmf_setDidShowTitleDescriptionEditingIntro(true)
                 navigationController.view.alpha = 1
             }
         }
+
         present(navigationController, animated: !needsIntro) {
             if needsIntro {
-                showIntro?()
+                showIntro {
+                    editVC.showTempAccountToast()
```
tonisevener, line 80: I notice this toast doesn't appear if you tap "skip" in the welcomeVC.

I think this would be more reliable if you triggered the toast instead from `viewDidAppear` within `DescriptionEditViewController`. That way you don't have to adjust the presentation logic here outside of the description editor view (for the toast at least).

tonisevener (commented):
@l-olson1214 I have a couple of behavioral questions:

My understanding of the requirements are that the temp accounts sheet component should display after tapping "Edit article description":

Screenshot 2025-03-31 at 4 54 13 PM

I also expect this view to display when tapping "Read more" in the toast:

Screenshot 2025-03-31 at 4 54 53 PM

I'm not seeing it act this way in the PR. If these changes already passed design review, then you can ignore this feedback, just be sure QA is aware of the change or else it will bounce back.

Wikipedia/Code/DescriptionEditViewController.swift
```diff
@@ -95,7 +170,7 @@ protocol DescriptionEditViewControllerDelegate: AnyObject {
     
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
-        descriptionTextView.becomeFirstResponder()
+        self.descriptionTextView.becomeFirstResponder()
```
tonisevener, line 173: I suggest removing this `becomeFirstResponder()` line entirely (which focuses the textfield) and replacing it with `showTempAccountToast`, to display the toast at the bottom. If we kept the first responder focus the keyboard would hide the toast.

tonisevener (changes requested):

tonisevener (approved):

PR #5238, opened 2025-04-03
Temp accounts > Main

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T373067

### Notes
* Merging all of temp accounts so far into main

### Test Steps
1.


Wikipedia/Code/EditSaveViewController.swift
```diff
@@ -413,8 +495,19 @@ class EditSaveViewController: WMFScrollViewController, Themeable, UITextFieldDel
...
+        var needsNewTempAccountToast = false
+        guard let dataStore else { return }
+        if !dataStore.authenticationManager.authStateIsPermanent {
+            if dataStore.authenticationManager.authStateIsTemporary {
+                if authState == .ipAccount {
+                    needsNewTempAccountToast = true
+                }
+            }
+        }
+
```
mazevedofs, line 508: I might have missed a pilot wiki status check here, what do you think?

mazevedofs (approved):
Looks good to me! We can rely on Qa for edge case testing!

PR #5241, opened 2025-04-03
Wrap in task, update text

l-olson1214:


Wikipedia/Code/EditSaveViewController.swift
```diff
@@ -495,13 +495,14 @@ class EditSaveViewController: WMFScrollViewController, Themeable, UITextFieldDel
...
     
     private func handleEditSuccess(with result: [AnyHashable: Any]) {
-
         var needsNewTempAccountToast = false
-        guard let dataStore else { return }
-        if !dataStore.authenticationManager.authStateIsPermanent {
-            if dataStore.authenticationManager.authStateIsTemporary {
-                if authState == .ipAccount {
-                    needsNewTempAccountToast = true
+        Task {
```
tonisevener, line 499: I did not see the temp accounts toast anymore when publishing on Test Wiki. I think it's possible that with the new `await`, the flag is not set to true yet before you pass it along on line 511.

At this point @mazevedofs set a `wikiHasTempAccounts` property already in `viewDidLoad`, so you may not need to make the async call again to get the status from the data controller. I would try it without:

```
        var needsNewTempAccountToast = false
        guard let dataStore else { return }
        if let wikiHasTempAccounts, wikiHasTempAccounts, !dataStore.authenticationManager.authStateIsPermanent {
            if dataStore.authenticationManager.authStateIsTemporary {
                if authState == .ipAccount {
                    needsNewTempAccountToast = true
                }
           }
      }
```

mazevedofs (approved):
Looks good!

PR #5249, opened 2025-04-15
Add email authentication 

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T391127

### Notes
* Adds email authentication to the app
* We are now able to verify the API response for email auth
* The 2fa UI was updated to accept a short alphanumeric code and send it back to the login call
* To test, a developer setting option was added - it adds a cookie that forces the email auth flow when on 

### Test Steps
1. Log out of the app
2. Turn on the dev setting to force email auth
3. Log back in - the 2FA UI should show
4. Insert the code you'll get in your email
5. Login should work normally


tonisevener (approved):
Working great!

PR #5269, opened 2025-04-30
T390991- Add Tabs Entry Points

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T390991

### Notes
* 

### Test Steps
1. Open app
2. Navigate to the following places:
    a. Article
    b. Explore
    c. Places
    d. History
    e. Saved
    f. Search
3. Make sure they all have the tabs icon
4. Ensure that tapping the tabs icon successfully brings you to a blank screen (it will say "Tabs" and have a close button)

### Screenshots/Videos

https://github.com/user-attachments/assets/3a84bed9-c32a-4378-b1f6-87aaf261cb1f

https://github.com/user-attachments/assets/c88107d4-e8fa-4023-9b45-6323ad14b410


WMFData/Sources/WMFData/Data Controllers/Article Tabs/ArticleTabsDataController.swift
```diff
@@ -0,0 +1,37 @@
...
+        self.coreDataStore = coreDataStore
+        self.userDefaultsStore = userDefaultsStore
+        self.developerSettingsDataController = developerSettingsDataController
+    }
+    
+    // MARK: Entry point
+
+    @objc public var shouldShowArticleTabs: Bool {
+        get {
+            return (try? userDefaultsStore?.load(key: WMFUserDefaultsKey.developerSettingsArticleTab.rawValue)) ?? true
```
tonisevener, line 32: Try defaulting to false here instead of true - I saw the tabs icon when I first launched I think because of this.

WMFData/Sources/WMFData/Data Controllers/Article Tabs/ArticleTabsDataController.swift
```diff
@@ -0,0 +1,37 @@
...
+            throw WMFDataControllerError.coreDataStoreUnavailable
+        }
+        self.coreDataStore = coreDataStore
+        self.userDefaultsStore = userDefaultsStore
+        self.developerSettingsDataController = developerSettingsDataController
+    }
+    
+    // MARK: Entry point
+
+    @objc public var shouldShowArticleTabs: Bool {
```
tonisevener, line 30: You can remove the @objc here too.

WMFComponents/Sources/WMFComponents/Extensions/Navigation Bar/WMFNavigationBarConfiguring.swift
```diff
@@ -164,27 +183,45 @@ public extension WMFNavigationBarConfiguring where Self: UIViewController {
...
-            if let trailingBarButtonItem = profileButtonConfig.trailingBarButtonItem {
-                rightBarButtonItems.insert(trailingBarButtonItem, at: 0)
+            if let leading = profileButtonConfig.leadingBarButtonItem {
+                rightBarButtonItems.append(leading)
             }
-            
-            navigationItem.rightBarButtonItems = rightBarButtonItems
         }
         
+        if let dataController = try? ArticleTabsDataController() {
```
tonisevener, line 206: You might need to adjust the order that you are building up `rightBarButtonItems` here. I notice that the ordering is off for nav bars that need a variable leading button (Places, Saved, History):

Mock:
Screenshot 2025-04-30 at 4 22 55 PM

PR (Filter should precede tabs):
Screenshot 2025-04-30 at 4 22 25 PM

Wikipedia/Code/ExploreViewController.swift
tonisevener, line 1: There's a little theming bug that can happen with these nav bar bar buttons when the theme switches. From Explore:

Open Profile > Settings > change theme to Dark or Black if you were on Light or Sepia, or vice-versa.

After dismissing settings, the tabs button color is wrong - we need to adjust the tint of it when the theme changes:

Screenshot 2025-04-30 at 4 28 20 PM

I did something similar to fix the profile button, so here's something you could try:

1. When you create your tabs button, add an accessibility identifier like I did here with profile: https://github.com/wikimedia/wikipedia-ios/pull/5269/files#diff-b7428cdc03a8a0d6a101e75fec111a132f82c63ddbf2bc9c0e2a8fba59d3e2c7R193
2. Add a method like this to WMFNavigationBarConfiguring:

```
func themeTabsButton() {
        guard let button = navigationItem.rightBarButtonItems?.filter { $0.accessibilityIdentifier == tabsButtonAccessibilityID }.first else { return }
        button.tintColor = WMFAppEnvironment.current.theme.navigationBarTintColor
}
```

3. Call this method from the apply(theme:) method in Explore, Places, Saved, History, Search, and Article.

```
override func apply(theme: Theme) {
        super.apply(theme: theme)
        ...
        themeTabsButton()
        ...
}
```

tonisevener (changes requested):
Looking good! Just a few things, mostly minor.

tonisevener (approved):

WMFData/Sources/WMFData/Data Controllers/Article Tabs/WMFArticleTabsDataController.swift
```diff
@@ -0,0 +1,30 @@
...
+        }
+        self.coreDataStore = coreDataStore
+        self.userDefaultsStore = userDefaultsStore
+        self.developerSettingsDataController = developerSettingsDataController
+    }
+    
+    // MARK: Entry point
+
+   public var shouldShowArticleTabs: Bool {
+        get {
```
tonisevener, line 24: For this, I think you can just do `return developerSettingsDataController. enableArticleTabs` instead of digging into user defaults here. Sorry I missed this before approval. I can fix it in my next PR.

PR #5277, opened 2025-05-07
Add WMFArticleTabsDataController mock

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T392917

### Notes
This PR wraps up a `WMFArticleTabsDataController` into a protocol, so that we can use a mocked-version when needed. It also creates a `WMFMockArticleTabsDataController` mock in WMFDataMocks. I added an additional `populateWithInitialTabs` method to it so that we can generate initial sample tabs.

### Test Steps
1. Ensure unit tests pass


WMFData/Sources/WMFDataMocks/WMFMockArticleTabsDataController.swift
```diff
@@ -0,0 +1,186 @@
...
+    }
+    
+    public func currentTabIdentifier() async throws -> UUID {
+        for tab in tabs {
+            if tab.isCurrent {
+                return tab.identifier
+            }
+        }
+        
+        assertionFailure("Shouldn't get here.")
```
l-olson1214, line 181: Did we intend to keep this here? Or should we use a more specific message?

tonisevener: @l-olson1214 good catch! Added a commit to improve it - I'm throwing an error now.

l-olson1214 (approved):

PR #5284, opened 2025-05-12
Tabs - Navigation Fixes

tonisevener:
**Phabricator:**

This partially touches a lot of Phabricator tasks. I will audit and document what is remaining on each of these.
https://phabricator.wikimedia.org/T392917
https://phabricator.wikimedia.org/T392398
https://phabricator.wikimedia.org/T392400
https://phabricator.wikimedia.org/T392167
https://phabricator.wikimedia.org/T392653
https://phabricator.wikimedia.org/T392403

### Notes
This PR adds many fixes around article navigation and how that persists as tabs in the database. A lot of these fixes were around incorrect assumptions I had with navigation and database persistence:

1. When tapping a tab in the tabs overview, instead of pushing on all of that tab's articles in the navigation controller, we should only push on the top-most article.
2. I added Main Page as an article to persist in the database. I avoided it up until this point, because I thought Main Page only showed on a (nearly) empty tabs overview screen, so basically when initially launched and an article hasn't been visited yet. But we actually need it to show in the tabs overview for any tab that has either started navigating in a new tab or tapped back to Main Page.

To make sure this is all working as expected, I added navigation logic when tapping a tab and the add tab button. I also made these improvements:

1. I am creating our initial tab as soon as the developer settings tabs toggle is flipped, for easier testing. Now we don't have to re-launch after flipping the developer settings toggle.
2. When a user deletes a tab in the tabs overview, I am firing an NSNotification which eventually clears out articles that were a part of that tab from the navigation stack. This was a clarification from design.
3. I am updating a tab's "current" article (that is, what will show in the tabs overview) on article _appearance_ rather than on tapping the navigation back button via a new `syncTabsOnArticleAppearance` method. This way the tabs overview doesn't actually switch to the new article until the user actually lands on that article. This was another clarification from design.
4. I tweaked NSManagedObjectContext handling to avoid NSMergeConflicts that were printing out to the console.
5. I am guarding out of all of our Year in Review local tracking (page views, number of seconds, categories) for Main Page, since now it will appear on screen a lot more often and is needless information to track.
6. I did some initial work on [T392653](https://phabricator.wikimedia.org/T392653), because it was easier to test my logic if I could see a more descriptive label other than "Back". It could still use smarter truncation.

### Test Steps
1. Fresh install app, toggle on developer settings. Go back to Explore. Oddly you have to switch to a different tab like Saved for the entry point to appear on Explore. Go back to Explore tab and tap tabs overview button. Ensure you see blank "Main Page" grid item.
2. Tap Main Page item. Ensure tabs overview dismisses and you are taken to Main Page. Ensure back button says Explore.
3. Search for "Cat" article on Main Page and select Cat in search results. When landing on new article, tap tabs overview, ensure you see Cat article in first tab.
4. Tap + button
5. Ensure you are taken to a new main page. Ensure back button says Cat. Tap tabs overview. Ensure you see Cat as first grid item, Main Page as second.
6. Dismiss tabs overview. Tap "Wikipedia" link on Main Page. Ensure back button now says Main Page. Tap tabs overview. Ensure you see Wikipedia as second grid item.
7. Delete Wikipedia tab. Dismiss tabs overview. Ensure you see Cat article.
8. Tap Back on Cat. Ensure you see Main Page search results. Cancel search results, ensure you see Main Page. Tap Back on Main Page, ensure you are taken to Explore.
9. Tap tabs overview, ensure you see single Main Page grid item.


Wikipedia/Code/NSUserDefaults+WMFExtensions.swift
```diff
@@ -74,6 +74,8 @@ let WMFYiRSettingsToggleShouldShow = "WMFYiRSettingsToggleShouldShow"
 
     @objc var shouldRestoreNavigationStackOnResume: Bool {
         get {
+            // TODO: Temp code for removing old nav stack restoration
```
l-olson1214, line 77: Did this need to be removed / updated in this or just leftover?

l-olson1214 (approved):

Wikipedia/Code/NSUserDefaults+WMFExtensions.swift
tonisevener: @l-olson1214 We'll revisit this when we pick up https://phabricator.wikimedia.org/T392919, I added a comment with those details.

PR #5289, opened 2025-05-16
Tabs - Add back / forward buttons to article more menu

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T392917

### Notes
Per feedback from Product and Design, we want to add back and forward buttons to the article more menu to allow a user to traverse all articles in a tab.

We now have to actively clean out previous future articles for data integrity. That is, if a user goes back to a previous article in the tab, then navigates elsewhere, we need to clean out the old future articles because they are no longer a part of the original rabbit hole. I am doing this within the data controller for deleting the article tab item, and am also firing an NSNotification (which the WMFAppViewController picks up) to clear out any existing ArticleViewController's with that deleted tab identifier. This way all traces of the old rabbit hole are cleared out.

### Test Steps
1. Create some tabs, some with multiple articles and some without.
2. Terminate the app
3. Open the app again and open a tab from tabs overview.
4. Check the article more menu - confirm you can reload the article view via tapping the new back and forward options. Confirm these back and forward options disable at the appropriate times when you are at the beginning or end of a tab.
5. Navigate to an article in the middle of a tab, then tap a blue link. Confirm old traces of the original rabbit hole are gone.

For example, if you created a Tab with Cat > Dog > Earth articles, then tapped back to Dog and navigated to Ocean, confirm you cannot navigate to Earth anymore. The new tab's articles should read Cat > Dog > Ocean (Earth is deleted).


WMFData/Sources/WMFData/Data Controllers/Article Tabs/WMFArticleTabsDataController.swift
```diff
@@ -122,6 +122,7 @@ public class WMFArticleTabsDataController: WMFArticleTabsDataControlling {
     // MARK: Entry point
 
     public var shouldShowArticleTabs: Bool {
+        return true
```
mazevedofs, line 125: I think it's test code

Wikipedia/Code/ArticleViewController.swift
```diff
@@ -556,41 +565,65 @@ class ArticleViewController: ThemeableViewController, HintPresenting, UIScrollVi
...
+                DDLogError("Error fetching article MediaWiki info: \(error)")
+            }
+        }
+    }
+    
+    func loadNextAndPreviousArticleTabs() {
+        guard let title = articleURL.wmf_title,
+            let siteURL = articleURL.wmf_site,
+            let project = WikimediaProject(siteURL: siteURL)?.wmfProject else {
+                return
```
mazevedofs, line 613: `title` and `project` are not being used, you could use _ here so the warning goes away

mazevedofs (changes requested):
Looks good, just minor details

mazevedofs (approved):
Looks good!

PR #5293, opened 2025-05-20
Improve performance of article summary fetch

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T392918

### Notes
Previously, I had set up the data controller's [fetchAllTabs](https://github.com/wikimedia/wikipedia-ios/compare/T392918?expand=1#diff-468f032d0680cb3596bfe25134d4cffc429c943d231469c2f8456a2789c57af6L585)() to fetch the article summary for all of the user's tabs, which could be up to 500 API calls at once when the tabs overview screen appears. Now I am fetching onAppearance of the particular tab grid item, so that we are only fetching what is necessary for the user at that moment.

I assist with the transition between a loading state and a populated state, I refactored the existing content view of a grid item into a separate SwiftUI view called `WMFArticleTabsViewContent`.

As far as caching goes, I'm letting the system do its default network caching. As I proxy the calls, I can see that followup calls to the same summary endpoint returns 304 Not Modified, which means it is leaning on internal cache already. This will provide a good balance between caching and refreshing in the case of vandalism. The server will control if the article has been updated and will no longer return a Not Modified response, which will cause the app to download again and refresh.

### Test Steps
1. Open a lot of tabs. Watch networking traffic via a proxy.
2. Open tabs overview. Ensure only the tabs seen on screen are fetching the summary endpoint (`https://{lang}.wikipedia.org/api/rest_v1/page/summary/{title}`)
3. As you scroll down, confirm more summary endpoints are fetched as-needed and the grid item shows the data as it did before.

### DIFF
--- a/WMFComponents/Sources/WMFComponents/Components/Article Tabs/WMFArticleTabsViewModel.swift
+++ b/WMFComponents/Sources/WMFComponents/Components/Article Tabs/WMFArticleTabsViewModel.swift
@@ -6,7 +6,6 @@ public class WMFArticleTabsViewModel: NSObject, ObservableObject {
     // articleTab should NEVER be empty - take care of logic of inserting main page in datacontroller/viewcontroller
     @Published var articleTabs: [ArticleTab]
     @Published var shouldShowCloseButton: Bool
-    @Published var count: Int
     
     private let dataController: WMFArticleTabsDataController
     public var updateNavigationBarTitleAction: ((Int) -> Void)?
@@ -26,7 +25,6 @@ public class WMFArticleTabsViewModel: NSObject, ObservableObject {
         self.localizedStrings = localizedStrings
         self.articleTabs = []
         self.shouldShowCloseButton = false
-        self.count = 0
         self.didTapTab = didTapTab
         self.didTapAddTab = didTapAddTab
         self.populateSummary = populateSummary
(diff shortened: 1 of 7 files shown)


WMFComponents/Sources/WMFComponents/Components/Article Tabs/WMFArticleTabsViewModel.swift
```diff
@@ -54,29 +51,25 @@ public class WMFArticleTabsViewModel: NSObject, ObservableObject {
...
             let tabs = try await dataController.fetchAllArticleTabs()
-            self.articleTabs = tabs.map { tab in
+            articleTabs = tabs.map { tab in
                 ArticleTab(
-                    image: tab.articles.last?.imageURL,
                     title: tab.articles.last?.title.underscoresToSpaces ?? "",
-                    subtitle: tab.articles.last?.description,
-                    description: tab.articles.last?.summary,
                     dateCreated: tab.timestamp,
+                    info: nil,
```
l-olson1214, line 58: Why aren't we loading the info here out of curiosity?

l-olson1214 (approved):
Approve once question answered :)

WMFComponents/Sources/WMFComponents/Components/Article Tabs/WMFArticleTabsViewModel.swift
tonisevener: Good question! At this point I'm not sure what grid items are visibly laid out on screen. This map loops through every possible article tab, so if I were to fetch the summary of each tab article here, we're back to the original problem of fetching unnecessary data for the user.

I could  just prefix the first 10 tabs or so, whatever I assume would probably show on screen, and fetch their summaries here. But then when the user scrolls they wouldn't have the later article summaries populated. It's better IMO to have them all initially empty, then listen for the grid item 'onAppear' hook to communicate that this grid item is indeed on screen and we need to fetch its summary data.

l-olson1214 (approved):

PR #5303, opened 2025-05-23
Some fixes from PCS testing

tonisevener:
**Phabricator:** 
https://phabricator.wikimedia.org/T369434
https://phabricator.wikimedia.org/T384845

### Notes
These are some fixes that I worked on while testing the PCS codex patch. They aren't blockers to that patch, but some improvements that have been needed.

- Updating the scheme allowed me to test the local scheme with the local service running from my device.
- The ReferenceShowing change realigns the reference window when tapping on a reference superscript link. This is a bug QA found a while back in https://phabricator.wikimedia.org/T384845#10749096.
- The styleoverrides.css fixes an issue where you could not easily see the current find match in dark and black modes.

### Test Steps
1. Run the app and tap a reference superscript in article. Ensure the window is now more aligned with the tap target. Note this doesn't fix all the issues Anthony found, but I think this partial improvement is still worth merging.
2. Tap the find in page button and find a bit of text. Ensure the current match background color is brighter than the other matches on the dark and black themes.


mazevedofs (approved):
Looks good!

PR #5308, opened 2025-05-27
Remove providence tracker

snan:
PLEASE remove this junk. I know it only takes like five seconds
to remove it when I bookmark or share links but it's the five most
annoying seconds of every day—many times per day even since I link a
lot to Wikipedia. With this line removed, a lot of the other functions
in this file become nops so feel free to squash in some commits that
remove all of them as you merge this. Wikipedia is something we all
created for the public benefit, not to be annoying and painful like
commercial malware is.

### DIFF
--- a/Wikipedia/Code/URL+LinkParsing.swift
+++ b/Wikipedia/Code/URL+LinkParsing.swift
@@ -123,9 +123,7 @@ extension URL {
     }
     
     fileprivate func wmf_URLForSharing(with wprov: String) -> URL {
-        let queryItems = [URLQueryItem(name: "wprov", value: wprov)]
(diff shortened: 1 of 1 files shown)


mazevedofs:
Hi @snan, thanks for your contribution. The provenance values we add to URLs are a valuable tool for tracking sharing from the Wikipedia app to other platforms, helping us understand and demonstrate the impact of our work. We understand it can be inconvenient for some users, so I suggest adding a setting that allows users to opt out of it instead of completely removing it. 
You can update this PR with the settings option or create a Phabricator ticket on our [board](https://phabricator.wikimedia.org/project/board/782/) for this if you don't feel comfortable working on it.

PR #5310, opened 2025-05-28
TABS - Analytics

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T390742

### Notes
* Creates a new funnel class `ArticleTabsFunnel`
* Adds new sources for article link interaction for internal and external links

### Test Steps
1. Enable ALL level logging on WMFLogging.h
2. Set the device to a valid region and language
3. Fresh install the app, verify the group assignement
4. With tabs enabled, verify all the actions according to the deck
5. Verify with a proxy inspector that all the logging calls go through


l-olson1214 (changes requested):
It doesn't seem like we're tracking Open in Background Tab here - is it needed?

PR #5313, opened 2025-06-02
Design feedback, update survey re: Haley

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T392921

### Notes
* Changes appearance of survey to third view as per Haley, as well 

### Test Steps
1.

### DIFF
--- a/WMFComponents/Sources/WMFComponents/Components/Article Tabs/WMFArticleTabsView.swift
+++ b/WMFComponents/Sources/WMFComponents/Components/Article Tabs/WMFArticleTabsView.swift
@@ -251,31 +251,34 @@ fileprivate struct WMFArticleTabsViewContent: View {
     
     private func tabText(tab: ArticleTab) -> some View {
         VStack(alignment: .leading, spacing: 0) {
-            Group {
-                if let subtitle = tab.info?.subtitle {
-                    Text(subtitle)
-                } else {
-                    Text(" ")
-                        .hidden()
-                }
+            if let subtitle = tab.info?.subtitle, !subtitle.isEmpty {
+                Text(subtitle)
+                    .font(Font(WMFFont.for(.caption1)))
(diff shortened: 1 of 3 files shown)


WMFComponents/Sources/WMFComponents/Components/Article Tabs/WMFArticleTabsViewModel.swift
```diff
@@ -101,10 +109,18 @@ public class WMFArticleTabsViewModel: NSObject, ObservableObject {
...
         if UIApplication.shared.preferredContentSizeCategory.isAccessibilityCategory {
+            if UIDevice.current.userInterfaceIdiom == .pad {
+                return 425
+            }
             return 225
         }
+
+        if UIDevice.current.userInterfaceIdiom == .pad {
+            let isLandscape = UIDevice.current.orientation.isLandscape
+            return isLandscape ? 160 : 110
```
mazevedofs, line 122: Using isLandscape here might cause an issue with split screen on iPad, maybe using horizontalSizeClass == .regular to check if it's landscape and full width. I think it's worth testing in this case.

WMFComponents/Sources/WMFComponents/Components/Article Tabs/WMFArticleTabsView.swift
```diff
@@ -269,31 +270,47 @@ fileprivate struct WMFArticleTabsViewContent: View {
...
             } else {
                 Spacer()
             }
         }
         .padding(.horizontal, 10)
     }
 }
+
+struct AspectRatioModifier: ViewModifier {
+    let condition: Bool
```
mazevedofs: "Condition" is too generic, something like `shouldLockAspectRatio` or something else would be clearer.

mazevedofs (approved):

PR #5321, opened 2025-06-09
feat: Add iOS search widget for Homescreen

Ytemiloluwa:
**Phabricator:** 

https://phabricator.wikimedia.org/T352836

### Notes

* This PR adds the small/square widget family size for the homescreen on iOS. A `SearchWidget.swift` file and a `SearchWidget+LocalizedStrings.swift` were added as separation of concerns to match the existing style of other widgets. the `SearchWidget.swift` file consists of the `SearchWidgetEntry`, `SearchWidgetProvider`, `SearchWidgetView` and `SearchWidgetConfiguration`. The other files modified are the `Widgets.swift`, `WMFAppViewController.m`, `Localizable.strings` and the `WidgetController.swift`. and lastly I added the wikipedia image to the assets catalog in the Widget folder.

 * I will appreciate your feedback/review on this PR @mazevedofs and @l-olson1214 

### Test Steps
1. Right click on a free area on the Homescreen to add a new widget.
2. Click on add widget.
3. Type in Wikipedia in the search box that pops up. 
4. Scroll through the widgets that are loaded on the screen until you see `Search`
5. Click on the Add widget blue button to add the search widget. 
6. Click Done on the Homescreen.
7. if you are on a simulator, toggle software keyboard then click on the search widget, it opens the wikipedia app and takes you to the search page with search field and keyboard activated. 

### Screenshots/Videos

https://github.com/user-attachments/assets/7bc80d1a-bfb2-4d7c-94de-50d289ec72a5

PR #5328, opened 2025-06-15
feat: Add iOS search widget for Lockscreen

Ytemiloluwa:
**Phabricator:** 

https://phabricator.wikimedia.org/T352836

### Notes

This PR is a follow up of #5321 

### Test Steps

1.  Long press the lockscreen to add a new lockscreen widget.
2. Click on the blue `customise` button and click on the Lockscreen 
3. Click on the Add widget Box, scroll down and click the Wikipedia button.
4. Tap or drag any of the `.accessoryCircular` or `.accessoryRectangular` widget to add it to the lockscreen. 
5. Click the Done button to add the selected widget to Lockscreen.


Widgets/Widgets/SearchWidget.swift
```diff
@@ -137,7 +154,47 @@ struct SearchWidgetView: View {
...
+    @ViewBuilder
+    private var accessoryInlineView: some View {
+        HStack(spacing: 6) {
+            Image(systemName: "magnifyingglass")
+                .font(.system(size: 11, weight: .medium))
+            
+            Text(SearchWidget.LocalizedStrings.searchWikipedia)
+                .font(.system(size: 13, weight: .medium))
+        }
+        .widgetURL(entry.url)
```
mazevedofs: The widget is not displaying correctly - it also needs to adopt the same solution that was done for the home screen widgets

Screenshot 2025-06-17 at 11 32 03

mazevedofs (changes requested):
When clicking on the widgets, it opens the app, but it doesn't open the search tab

Widgets/Widgets/LockscreenSearchWidget.swift
```diff
@@ -0,0 +1,127 @@
...
+    
+    public var body: some WidgetConfiguration {
+        StaticConfiguration(kind: kind, provider: LockscreenSearchProvider(), content: { entry in
+            LockscreenSearchWidgetView(entry: entry)
+        })
+        .configurationDisplayName(LockscreenSearchWidget.LocalizedStrings.widgetTitle)
+        .description(LockscreenSearchWidget.LocalizedStrings.widgetDescription)
+        .supportedFamilies([.accessoryCircular, .accessoryRectangular, .accessoryInline])
+        .contentMarginsDisabled()
+        .containerBackgroundRemovable(false)
```
mazevedofs: I still get the same background error, you need to use `containerBackground` here and use `contentMargins(.disabled)`

Widgets/Widgets/LockscreenSearchWidget+LocalizedStrings.swift
```diff
@@ -0,0 +1,12 @@
+import Foundation
+import WMF
+
+extension LockscreenSearchWidget {
+
+	enum LocalizedStrings {
+		static let widgetTitle = WMFLocalizedString("lockscreen-search-widget-title", value:"Search", comment: "Text for title of lockscreen search widget.")
+		static let widgetDescription = WMFLocalizedString("lockscreen-search-widget-description", value:"Quick access to search Wikipedia from your lock screen.", comment: "Text for description of lockscreen search widget displayed when adding to lock screen.")
+		static let searchWikipedia = WMFLocalizedString("lockscreen-search-widget-search-wikipedia", value:"Search Wikipedia", comment: "Text displayed on lockscreen search widget.")
+	}
```
mazevedofs: You already added some of the same strings for the Search widget, we can reuse the strings.
You can add them to the CommonStrings.swift file and reuse them whenever you need.

Optional:
The "Search Wikipedia" text already exists in the app, which means that it was translated already. You can reuse it—if it's not yet in CommonStrings, you can move it here.

The catch is keeping the key and comment, do not alter.

mazevedofs (changes requested):

WMF Framework/CommonStrings.swift
```diff
@@ -652,6 +652,10 @@ public class CommonStrings: NSObject {
     
     public static let searchButtonAccessibilityLabel = WMFLocalizedString("search-button-accessibility-label", value: "Search Wikipedia", comment: "Accessibility label for a button that opens a search box to search Wikipedia.")
     
+    // Lockscreen Search Widget strings
+    public static let lockscreenSearchWidgetTitle = WMFLocalizedString("lockscreen-search-widget-title", value: "Search", comment: "Text for title of lockscreen search widget.")
```
l-olson1214: If one already exists, you can use it - so in this case you could use `searchTitle` above in the file (line 42) - what I do is search directly for my string with the quotation marks to see if it's already used. 

If it's already created elsewhere, that means we already have the translations for it. If we recreate it, the volunteers will need to re-translate the same thing again. 

Nice job getting it in the right place!

Widgets/Widgets/LockscreenSearchWidget.swift
```diff
@@ -0,0 +1,140 @@
...
+    var entry: LockscreenSearchEntry
+    
+    var body: some View {
+        switch widgetFamily {
+        case .accessoryCircular:
+            accessoryCircularView
+        case .accessoryRectangular:
+            accessoryRectangularView
+        case .accessoryInline:
+            accessoryInlineView
```
mazevedofs: Is the inline view available? I don't see it when adding the widgets

WMF Framework/CommonStrings.swift
Ytemiloluwa: Okay Grey,  I will remove the duplicated search title and push an update.

Widgets/Widgets/LockscreenSearchWidget.swift
Ytemiloluwa: it is not available, I added it to complete the widget family, I will remove it when I commit the next changes.

Widgets/Widgets/LockscreenSearchWidget.swift
```diff
@@ -0,0 +1,126 @@
...
+struct LockscreenSearchWidget: Widget {
+    private let kind: String = WidgetController.SupportedWidget.lockscreenSearch.identifier
+    
+    public var body: some WidgetConfiguration {
+        StaticConfiguration(kind: kind, provider: LockscreenSearchProvider(), content: { entry in
+            LockscreenSearchWidgetView(entry: entry)
+        })
+        .configurationDisplayName(CommonStrings.searchTitle)
+        .description(CommonStrings.lockscreenSearchWidgetDescription)
+        .supportedFamilies([.accessoryCircular, .accessoryRectangular, .accessoryInline])
```
mazevedofs: You can remove .accessoryInline from here as well.

Ytemiloluwa: I have removed it now.

PR #5331, opened 2025-06-19
T381678: YIR Cleanup pt 1/2

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T381678

### Notes
* Updates out some data controller 
* Creates factory
* Creates a protocol for slides
* Does NOT remove entry points yet, stay tuned for PR 2


WMFData/Sources/WMFData/Data Controllers/Year In Review/WMFYearInReviewSlideProtocol.swift
```diff
@@ -0,0 +1,352 @@
+import Foundation
+import CoreData
+
+public protocol YearInReviewSlideProtocol {
```
mazevedofs: I think the protocol should be in it's own file.
This file is named `YearInReviewSlideProtocol`, but it also contains other models, creates each of the slides, etc.
It'll be hard to find what we need in the future, so maybe splitting the files to make it clearer will help

WMFData/Sources/WMFData/Data Controllers/Year In Review/WMFYearInReviewSlideFactory.swift
```diff
@@ -0,0 +1,90 @@
...
+                year: year,
+                userID: userID,
+                languageCode: lang,
+                project: userInfo.project,
+                fetchViews: fetchEditViews
+            )
+            slides.append(slide)
+        }
+
+        return slides
```
mazevedofs: I'm not sure if t's the moment where this process starts and returns, but none of the slides are returning personalized data, so I'm ending up with only the collective slides.

WMFData/Tests/WMFDataTests/WMFYearInReviewDataControllerTests.swift
```diff
@@ -381,6 +384,7 @@ final class WMFYearInReviewDataControllerTests: XCTestCase {
             let shouldShowEntryPoint = yearInReviewDataController.shouldShowYearInReviewEntryPoint(countryCode: frCountryCode, primaryAppLanguageProject: frProject)
 
             XCTAssertTrue(shouldShowEntryPoint, "Should show entry point when one personalized slide is enabled.")
+            // XCTAssertFalse(shouldShowEntryPoint) // Entry points disabled
```
mazevedofs: Tests should work independently of the implementation so we can test several situations.
Create different tests mocking both situations. That should be executed correctly regardless if the feature is active or not.

Wikipedia/Code/WMFSettingsViewController.m
```diff
@@ -591,13 +582,6 @@ - (WMFSettingsTableViewSection *)section_2 {
     NSMutableArray *items = [NSMutableArray arrayWithArray:commonItems];
     [items addObject:[WMFSettingsMenuItem itemForType:WMFSettingsMenuItemType_ExploreFeed]];
 
-    if ([[WMFYearInReviewDataController dataControllerForObjectiveC] shouldShowYearInReviewSettingsItemWithCountryCode:NSLocale.currentLocale.countryCode primaryAppLanguageCode:self.dataStore.languageLinkController.appLanguage.languageCode]) {
-        [items addObject:[WMFSettingsMenuItem itemForType:WMFSettingsMenuItemType_YearInReview]];
-        [[NSUserDefaults standardUserDefaults] wmf_setShowYirSettingToggle:YES];
-    } else {
-        [[NSUserDefaults standardUserDefaults] wmf_setShowYirSettingToggle:NO];
-    }
-
```
mazevedofs: Instead of removing this (which will be needed in the future, forget we're redoing this VC), it should be tied to the other logic for displaying the entry points.

WMFData/Tests/WMFDataTests/WMFYearInReviewDataControllerCreateOrRetrieveTests.swift
```diff
@@ -154,7 +148,7 @@ final class WMFYearInReviewDataControllerCreateOrRetrieveTests: XCTestCase {
 
         XCTAssertNotNil(report, "Expected a report to be retrieved")
         XCTAssertEqual(report?.year, year)
-        XCTAssertEqual(report?.slides.count, 6)
+        // XCTAssertEqual(report?.slides.count, 6) - Temporarily removed while YIR is not active
```
mazevedofs, line 98: Same comment as for the other tests. They should evaluate all the possible implementations, so no need to comment/alter. Maybe creating another test to evaluate this situation would be better.

WMFData/Sources/WMFData/Data Controllers/Year In Review/WMFYearInReviewDataController.swift
```diff
@@ -815,53 +545,46 @@ import CoreData
...
-              let dataPopulationEndDate = yirConfig.dataPopulationEndDate else {
-            return
-        }
+        try await saveYearInReviewReport(newReport)
+    }
 
-        let donateHistory = WMFDonateDataController.shared.loadLocalDonationHistory(startDate: dataPopulationStartDate, endDate: dataPopulationEndDate)
-        let donateCount = donateHistory?.count ?? 0
+    public func fetchYearInReviewReport(forYear year: Int) throws -> WMFYearInReviewReport? {
+        assert(Thread.isMainThread, "This report must be called from the main thread in order to keep it synchronous")
```
mazevedofs: Optional: In the spirit of making things easier for the Swift 6 migration, we could evaluate if this class should be a main actor.
But only if you feel like exploring Swift concurrency a little, since this code works fine.

mazevedofs (changes requested):
I'm having issues with generating the personalized slides - we need to make sure this cleanup doesn't hurt the current implementation before moving on.
Also, the tests could use some love so they evaluate all the possible situations, and we don't need to comment code.

WMFData/Sources/WMFData/Data Controllers/Year In Review/WMFYearInReviewDataController.swift
```diff
@@ -302,98 +299,99 @@ import CoreData
...
+            },
+            fetchEditViews: { project, userID, language in
+                try await self.fetchEditViews(project: project, userId: userID, language: language)
+            },
+            donationFetcher: { start, end in
+                WMFDonateDataController.shared.loadLocalDonationHistory(startDate: start, endDate: end)?.count
             }
+        )
+
+        let slideModels = slideFactory.makeSlides()
```
tonisevener: (Optional) After looking at this, it feels like these slide models really behave more like little bitty data controllers, unique to a particular slide. Maybe we should rename these objects to slide data controllers instead of slide models, for clarity.

WMFData/Sources/WMFData/Data Controllers/Year In Review/WMFYearInReviewDataController.swift
```diff
@@ -302,98 +299,99 @@ import CoreData
...
-            try await backgroundContext.perform { [weak self] in
-                try self?.populateDonatingSlide(report: report, backgroundContext: backgroundContext)
-            }
-        }
+        let userInfo = YearInReviewUserInfo(
+            username: username,
+            userID: userID,
+            project: primaryAppLanguageProject,
+            legacyPageViews: legacyPageViews,
+            savedArticlesData: savedArticlesData
```
tonisevener: What do you think about moving the `legacyPageViews` and `savedArticlesData` to the factory init, similar to how we're doing `fetchEditCount`? I think from an organizational perspective it would fit better with the other properties, since they're both pulling data from outside sources.

```
        let slideFactory = YearInReviewSlideFactory(
            year: year,
            config: featureConfig,
            userInfo: userInfo,
            fetchLegacyPageViews: {
                try await legacyPageViewsDataDelegate.getLegacyPageViews(from: start, to: end)
            },
            fetchEditCount: { username, project in
                try await self.fetchEditCount(username: username, project: project)
            },
            ...

```

WMFData/Sources/WMFData/Data Controllers/Year In Review/WMFYearInReviewDataController.swift
```diff
@@ -302,98 +299,99 @@ import CoreData
...
         return await backgroundContext.perform {
-            return WMFYearInReviewReport(cdReport: report)
+            let slides: [WMFYearInReviewSlide] = (report.slides as? Set<CDYearInReviewSlide>)?.compactMap { cdSlide in
+                guard let id = self.getSlideId(cdSlide.id) else { return nil }
+                return WMFYearInReviewSlide(year: Int(cdSlide.year), id: id, evaluated: cdSlide.evaluated, data: cdSlide.data)
+            } ?? []
+
+            return WMFYearInReviewReport(year: year, slides: slides)
         }
     }
```
tonisevener: Looks like the method on the next line, `getYearInReviewReportAndDataPopulationFlags`, is no longer called. Can you delete it?

WMFData/Sources/WMFData/Data Controllers/Year In Review/WMFYearInReviewDataController.swift
```diff
@@ -302,98 +299,99 @@ import CoreData
...
+            let predicate = NSPredicate(format: "year == %d", year)
+            let cdReport = try self.coreDataStore.fetchOrCreate(
+                entityType: CDYearInReviewReport.self,
+                predicate: predicate,
+                in: backgroundContext
+            )!
+            cdReport.year = Int32(year)
+
+            let cdSlides = try slideModels.map { try $0.makeCDSlide(in: backgroundContext) }
+            cdReport.slides = Set(cdSlides) as NSSet
```
tonisevener: Note that here, the report might _already_ have all the slides it needs created and populated in core data from previous app sessions. You should make sure the slide doesn't already exist or else you'll overwrite them here. The original requirements were that slide data is fetched once, then frozen. Product does not want counts to change if a user launches the app again later after a report was already generated. If we always make new slides here (and fetch new data every app launch by not first checking the persisted `evaluated` flag), those counts would change.

WMFData/Sources/WMFData/Data Controllers/Year In Review/WMFYearInReviewDataController.swift
```diff
@@ -618,137 +487,6 @@ import CoreData
...
-            }
-
-            switch slideID {
-            case WMFYearInReviewPersonalizedSlideID.editCount.rawValue:
-                let encoder = JSONEncoder()
-                slide.data = try encoder.encode(edits)
-
-                if edits > 0 {
-                    slide.display = true
-                }
```
tonisevener, line 641: Since we're removing the display flag, we should move the `edits > 0` requirement over to the [YearInReviewCoordinator](https://github.com/wikimedia/wikipedia-ios/pull/5331/files#diff-72caf34090dc13194bbfc66eeb449b72948cef8bd240c2a17c71437894632f90R596):

```
case .editCount:
                if slide.display == true,
                if slide.evaluated == true,
                   let data = slide.data {
                    let decoder = JSONDecoder()
                    if let editCount = try? decoder.decode(Int.self, from: data),
                        editCount > 0, { // < new line here
                        editCountSlide = YearInReviewSlideContent(
                        ...
```

There may be other slides that need this too.

WMFData/Sources/WMFData/Resources/WMFData.xcdatamodeld/WMFData 2.xcdatamodel/contents
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
```
tonisevener: To be safe, I think we should increase Core Data model versions instead of modifying the existing ones. We run the risk of triggering a crash without versioning (though I can't seem to trigger one right now). Please revert these data model changes you made here and instead create a new `WMFData 4.xcdatamodel` file, with the display attribute deleted.

I remember this being really buggy in Xcode since WMFData is a Swift Package, so let me know if you run into trouble. Try this:

1. Select the parent WMFData model in the navigator and choose Editor > Add Model Version.
Screenshot 2025-07-07 at 4 26 40 PM

2. Base model off of previous version (WMFData 3).
Screenshot 2025-07-07 at 4 26 53 PM

4. Highlight WMFData 4, delete `display` attribute in CDYearInReviewSlide.
Screenshot 2025-07-07 at 4 27 34 PM

6. Select the parent WMFData model in the navigator and set the current version to WMFData 4.
Screenshot 2025-07-07 at 4 33 14 PM

WMFData/Sources/WMFData/Data Controllers/Year In Review/WMFYearInReviewUniqueSlides.swift
```diff
@@ -0,0 +1,305 @@
...
+        self.year = year
+        self.username = username
+        self.project = project
+        self.isEvaluated = isEvaluated
+        self.fetchEditCount = fetchEditCount
+    }
+
+    public func populateSlideData(in context: NSManagedObjectContext) async throws {
+        guard let username, let project else { return }
+        editCount = try await fetchEditCount(username, project)
```
tonisevener: `CDYearInReviewSlide`'s `evaluated` property should be checked before fetching slide data. This is to avoid unnecessary data calls. Once we successfully fetch slide data and persist it to `CDYearInReviewSlide`'s `data` field, we should set the `CDYearInReviewSlide` `evaluated` flag to true. Here we should pull the associated CDYearInReviewSlide object from the `CDYearInReviewReport` and ensure the evaluated flag = false before fetching this data again.

tonisevener (approved):

josephseddon:
https://phabricator.wikimedia.org/T381678

mazevedofs (approved):

PR #5334, opened 2025-06-23
Remove YIR entry points 

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T383160

### Notes
* WAIT TO MERGE for https://github.com/wikimedia/wikipedia-ios/pull/5331

### Test Steps
1. Ensure entry points are removed

### DIFF
--- a/ContinueReadingWidget/Info.plist
+++ b/ContinueReadingWidget/Info.plist
@@ -17,7 +17,7 @@
 	<key>CFBundlePackageType</key>
 	<string>XPC!</string>
 	<key>CFBundleShortVersionString</key>
-	<string>7.8.0</string>
+	<string>7.8.1</string>
 	<key>CFBundleVersion</key>
 	<string>0</string>
 	<key>NSExtension</key>
--- a/NotificationServiceExtension/Info.plist
+++ b/NotificationServiceExtension/Info.plist
@@ -17,7 +17,7 @@
 	<key>CFBundlePackageType</key>
 	<string>$(PRODUCT_BUNDLE_PACKAGE_TYPE)</string>
 	<key>CFBundleShortVersionString</key>
-	<string>7.8.0</string>
+	<string>7.8.1</string>
 	<key>CFBundleVersion</key>
 	<string>$(CURRENT_PROJECT_VERSION)</string>
 	<key>NSExtension</key>
--- a/WMF Framework/Info.plist
+++ b/WMF Framework/Info.plist
@@ -15,7 +15,7 @@
 	<key>CFBundlePackageType</key>
 	<string>FMWK</string>
 	<key>CFBundleShortVersionString</key>
-	<string>7.8.0</string>
+	<string>7.8.1</string>
 	<key>CFBundleVersion</key>
 	<string>0</string>
 	<key>NSPrincipalClass</key>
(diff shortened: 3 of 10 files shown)


WMFData/Sources/WMFData/Data Controllers/Year In Review/WMFYearInReviewDataController.swift
```diff
@@ -407,54 +401,16 @@ import CoreData
...
-        
         try self.coreDataStore.saveIfNeeded(moc: backgroundContext)
 
         guard let iosFeatureConfig = developerSettingsDataController.loadFeatureConfig()?.ios.first,
-              let yirConfig = iosFeatureConfig.yir(yearID: targetConfigYearID) else {
-            return nil
-        }
-
-        guard let cdSlides = cdReport.slides as? Set<CDYearInReviewSlide> else {
+              let yirConfig = iosFeatureConfig.yir(yearID: targetConfigYearID),
```
mazevedofs: I'm against commenting code because it can lead to us forgetting about this (since it won't execute) or being untested.
My suggestion is:

On the data controller, wrap the remote config in a check like this.
var yirConfig: Bool = false
#if DEBUG
yirConfig = iosFeatureConfig.yir(yearID: targetConfigYearID) // so we still can test it in dev
#endif

This way it'll only see it in debug mode.

mazevedofs (changes requested):
I added one suggestions so we don't need to comment large parts of the code and we're able to easily test in dev.

WMFData/Tests/WMFDataTests/WMFYearInReviewDataControllerTests.swift
```diff
@@ -269,6 +269,7 @@ final class WMFYearInReviewDataControllerTests: XCTestCase {
             let shouldShowEntryPointFR = yearInReviewDataController.shouldShowYearInReviewEntryPoint(countryCode: frCountryCode, primaryAppLanguageProject: frProject)
             
             XCTAssertTrue(shouldShowEntryPointFR, "FR should show entry point for mock YiR config of [FR, IT] country codes.")
+            // XCTAssertFalse(shouldShowEntryPointFR) // Entry points disabled
```
mazevedofs: The tests should validate this even if the actual implementation of the code is different - the idea is that tests are isolated from the implementation and cover several cases.
You might need to create a new test case or test class to mock a version of the implementation with entry points enabled. Let me know if you need help with that.

WMFData/Tests/WMFDataTests/WMFYearInReviewDataControllerCreateOrRetrieveTests.swift
```diff
@@ -154,7 +148,7 @@ final class WMFYearInReviewDataControllerCreateOrRetrieveTests: XCTestCase {
 
         XCTAssertNotNil(report, "Expected a report to be retrieved")
         XCTAssertEqual(report?.year, year)
-        XCTAssertEqual(report?.slides.count, 6)
+        // XCTAssertEqual(report?.slides.count, 6) - Temporarily removed while YIR is not active
```
mazevedofs: Same here: we should be able to mock in the tests a version with a slide count.

PR #5343, opened 2025-06-30
Add new SwiftUI history and recently searched views

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T396409

### Notes
* Reuses the code originally developed to merge the history and search tabs
* From the original work, the search and history tabs were decoupled
* Now both tabs will stay as-is. The new code is done in SwiftUI
* Some small visual details come from the SWiftUI default list styles

### Test Steps
1. Run the app (iPhone and iPad)
2. Both views should look 9pretty much) the same


WMF Framework/CommonStrings.swift
```diff
@@ -651,6 +654,11 @@ public class CommonStrings: NSObject {
     public static let collectiveExploreAccessibilityLabel = WMFLocalizedString("year-in-review-collective-explore", value: "An animated illustration depicting the iconic Wikipedia globe encircled by scaffolding, symbolizing the collaborative effort of people worldwide in building the project. The scene is surrounded by a cityscape, emphasizing its connection to humanity and global culture.", comment: "Accessibility description for the collective explore slide.")
     
     public static let searchButtonAccessibilityLabel = WMFLocalizedString("search-button-accessibility-label", value: "Search Wikipedia", comment: "Accessibility label for a button that opens a search box to search Wikipedia.")
+
+    public static let readNowActionTitle = WMFLocalizedString("button-read-now", value: "Read now", comment: "Read now button text used in various places.")
```
tonisevener: FYI, the wording and icon for the long press menu item has since changed to "Open" from the tabs release.

WMFComponents/Sources/WMFComponents/Components/History/WMFArticlePreviewView.swift
```diff
@@ -0,0 +1,84 @@
+import SwiftUI
+import WMFData
+
+struct WMFArticlePreviewView: View {
+    let item: HistoryItem
```
tonisevener: Maybe we could reuse a more generic model, reference a protocol, or use a `WMFArticlePreviewViewModel` here instead, so that we can convert and use this preview view when tapping non-History items.

WMFComponents/Sources/WMFComponents/Components/History/WMFHistoryView.swift
```diff
@@ -0,0 +1,170 @@
+import SwiftUI
+import WMFData
+
+public struct WMFHistoryView: View {
+
+    // MARK: Properties
+
+    @ObservedObject var appEnvironment = WMFAppEnvironment.current
+    @ObservedObject var viewModel: WMFHistoryViewModel
+    @State private var refreshId = UUID()
```
tonisevener: Is this `refreshId` is still needed? Deleting items still seems to act the same when I comment it out.

WMFComponents/Sources/WMFComponents/Components/History/WMFHistoryView.swift
```diff
@@ -0,0 +1,170 @@
...
+                        title: viewModel.localizedStrings.emptyViewTitle,
+                        subtitle: viewModel.localizedStrings.emptyViewSubtitle,
+                        titleFilter: nil,
+                        buttonTitle: nil,
+                        attributedFilterString: nil
+                    )
+                    let emptyViewModel = WMFEmptyViewModel(
+                        localizedStrings: locStrings,
+                        image: viewModel.emptyViewImage,
+                        imageColor: .blue,
```
tonisevener: Maybe `imageColor: nil` here? I don't think the blue is having any effect.

WMFComponents/Sources/WMFComponents/Components/History/WMFHistoryView.swift
```diff
@@ -0,0 +1,170 @@
...
+        }
+        .listStyle(.plain)
+        .padding(.top, viewModel.topPadding)
+        .scrollContentBackground(.hidden)
+        .background(Color(theme.paperBackground))
+        .ignoresSafeArea(edges: .top)
+        .id(refreshId)
+        .onAppear {
+            refreshId = UUID()
+            viewModel.loadHistory()
```
tonisevener: This might load twice unnecessarily here - in the list view `onAppear` and in the ZStack `onAppear`.

WMFComponents/Sources/WMFComponents/Components/History/WMFHistoryViewModel.swift
```diff
@@ -0,0 +1,137 @@
...
+
+    internal func headerTextForSection(_ section: HistorySection) -> String {
+        let date = section.dateWithoutTime
+        let calendar = Calendar.current
+        if calendar.isDateInToday(date) {
+            return localizedStrings.todayTitle
+        } else if calendar.isDateInYesterday(date) {
+            return localizedStrings.yesterdayTitle
+        } else {
+            return DateFormatter.wmfFullDateFormatter.string(from: date)
```
tonisevener: FYI I have noticed the old behavior strips the year ("Friday, July 4") whereas this includes the date ("Friday, July 4, 2025").

WMFComponents/Sources/WMFComponents/Extensions/Navigation Bar/WMFNavigationBarHiding.swift
```diff
@@ -9,13 +9,17 @@ public protocol WMFNavigationBarHiding: AnyObject {
 public extension WMFNavigationBarHiding where Self:UIViewController {
     
     /// Call from UIViewController's viewDidLoad
-    func setupTopSafeAreaOverlay(scrollView: UIScrollView) {
+    func setupTopSafeAreaOverlay(scrollView: UIScrollView?) {
```
tonisevener: I think the changes in this file aren't used, you can undo them.

WMFComponents/Sources/WMFComponents/Components/History/WMFHistoryView.swift
```diff
@@ -0,0 +1,170 @@
...
+                    ForEach(section.items) { item in
+                        row(for: section, item)
+                            .listRowBackground(Color(theme.paperBackground))
+                    }
+                }
+
+            }
+
+        }
+        .listStyle(.plain)
```
tonisevener, line 140: I notice using a different style like .`grouped` allows the section title to scroll with the content, just FYI. We can hold on changing this until design review though.

tonisevener (commented):
Amazing work! Can't wait to get this cleanup in.

Some stuff that jumped out at me with an initial code review. I still need to play with it more and test it on iPad. A couple of behavioral things I noticed as well, though they may be difficult to fix:

1. The smooth delete row animation is gone - it looks like the whole list reloads now. It would be nice to bring it back - I tried briefly but haven't found anything obvious that you're missing for that.
2. Sometimes the navigation bar gets stuck in a hidden state. This is an issue I knew was coming with our first SwiftUI content rewrite, because right now I'm doing [additional UIKit work](https://github.com/wikimedia/wikipedia-ios/blob/main/Wikipedia/Code/ExploreViewController.swift#L302) on the other views to keep it from staying stuck while hiding. I don't really have any advice on how to do this from SwiftUI though 😕.

WMFComponents/Sources/WMFComponents/Components/History/WMFHistoryView.swift
mazevedofs: At some point I was having issues when deleting a row, I guess as I kep tdeveloping it got sorted out

WMF Framework/CommonStrings.swift
tonisevener: Minor ask - can you change the icon to the arrow instead of the book? Explore feed for reference:

Screenshot 2025-07-12 at 5 09 54 p m

Wikipedia/Code/HistoryViewController.swift
```diff
@@ -1,306 +1,377 @@
...
+        deleteButton = UIBarButtonItem(title: deleteAllButtonText, style: .plain, target: self, action: #selector(deleteButtonPressed(_:)))
+        deleteButton?.isEnabled = !viewModel.isEmpty
+        let hideNavigationBarOnScroll = !viewModel.isEmpty
+
+        let titleConfig: WMFNavigationBarTitleConfig = WMFNavigationBarTitleConfig(title: CommonStrings.historyTabTitle, customView: nil, alignment: .leadingCompact)
         extendedLayoutIncludesOpaqueBars = false
+
         if #available(iOS 18, *) {
             if UIDevice.current.userInterfaceIdiom == .pad && traitCollection.horizontalSizeClass == .regular {
-                titleConfig = WMFNavigationBarTitleConfig(title: CommonStrings.historyTabTitle, customView: nil, alignment: alignment)
```
tonisevener, line 188: I think this might be an existing bug (maybe introduced with activity tab?), but we should have the history title here for iPadOS 18 configured to .leadingLarge, so that it matches the other tabs:

leadingCompact:
IMG_1329

leadingLarge:
IMG_1330

Wikipedia/Code/HistoryViewController.swift
```diff
@@ -1,306 +1,377 @@
...
         }
-        let sectionInfo = sections[section]
-        guard let article = sectionInfo.objects?.first as? WMFArticle, let date = article.viewedDateWithoutTime else {
-            return nil
+    }
+
+    func share(item: HistoryItem, frame: CGRect?) {
+        if let dataStore, let url = item.url {
+            let article = dataStore.fetchArticle(with: url)
+            if let frame {
```
tonisevener: I'm not sure where the bug might be, but my share frames aren't lining up on my iPad mini:

Search 2

Wikipedia/Code/HistoryViewController.swift
```diff
@@ -1,306 +1,377 @@
...
+            return
         }
+
+        showReadingListHint(for: article)
     }
-    
-    override var eventLoggingCategory: EventCategoryMEP {
-        return .history
+
+    private func showReadingListHint(for article: WMFArticle) {
```
tonisevener, line 331: (Optional) A change I noticed is that the reading list hint no longer hides when you scroll, so it feels stuck there until I navigate away from the screen and then return. We were probably listening for some UIScrollView events to hide it before. It would be great if we can find a way to hide upon scroll here.

tonisevener (commented):
A few more notes after testing on iPad.

Wikipedia/Code/HistoryViewController.swift
mazevedofs: I tried intercepting the scrollView and adding a gesture recognizer, but neither worked well. I made the hint self dismiss faster (default time was 13 seconds, now it's 7). It's not ideal, but it looks less stuck.

PR #5349, opened 2025-07-10
T398077: Fix A/B assignment on initial load

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T398077

### Notes
* 

### Test Steps
1. Set Schema system to Aus + En (or another acceptable variant)
2. Remove Wikipedia, fresh install
3. Ensure tabs is available immediately
4. Remove Wikipedia
5. Set Schema system to US + En (or another NON acceptable variant)
6. Ensure tabs is NOT available


Wikipedia/Code/ArticleViewController.swift
```diff
@@ -619,8 +622,7 @@ class ArticleViewController: ThemeableViewController, HintPresenting, UIScrollVi
         Task { [weak self] in
             guard let self else { return }
             let tabDataController = WMFArticleTabsDataController.shared
-            
-            if tabDataController.shouldShowArticleTabs,
+            if tabDataController.shouldShowArticleTabs || articleTabsAssignment == .test,
```
tonisevener, line 625: I'm a bit unsure if this fixes the original bug (but, to be fair, I am also unclear on what the original bug complaint is in the task). It looks to me like `shouldShowArticleTabs` internally already calls `getArticleTabsExperimentAssignment` and returns `true` of the result is `.test`, so I don't think this additional check would do anything. Can you check your test steps against `main` to see if the behavior is different? Keep in mind the test variant is assigned randomly, so it may help to temporarily change the [test percentage to 100](https://github.com/wikimedia/wikipedia-ios/blob/main/WMFData/Sources/WMFData/Data%20Controllers/Article%20Tabs/WMFArticleTabsDataController.swift#L113) so we're not dealing with that uncertainty. Here's how it acts on `main` to me from a fresh install - this branch behaves the same:

https://github.com/user-attachments/assets/541badd0-62f2-4e51-bd0b-bbf193464f99

l-olson1214: ❗ Dang it! During testing I thought I had been able to get the tabs to show up immediately, Haley closed the ticket anyways so I'll be closing this task @tonisevener

PR #5354, opened 2025-07-14
TABS - New Tab View controller

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T398885

### Notes
* Adds a WMFNewArticleTabController, which takes a (for now) blank WMFNewArticleTabView, a WMFNewArticleTabViewModel, and a NewArticleTabCoordinator
* This new view controller already integrates the search controller
* By clicking on any search result, it opens a new tab and sets it as current
* Tapping on a search result removes the new tab VC from the stack
* If the user taps on add new tab in tabs overview when already coming from a new tab, the tabs overview is dismissed. This way, we don't create a stack of empty new tabs

### Test Steps
1. Run the app, enable tabs and more dynamic tabs
2. On tabs overview, click on new tab
3. On the new tab search a term, or click on a recent search, and click on a result
4. Verify you go to the new article, and you see the new tab in overview
5. Click back - you should go back to where you started (not new tab or tabs overview)
6. Click new tab, click overview
7. Click new tab again - verify the overview dismisses
8. Change theme (the view is not theming, not the scope of this task) - verify the navigation bar themes correctly
9. Turn off the dev setting, you should see the old experinece normally

### DIFF
--- a/ContinueReadingWidget/Info.plist
+++ b/ContinueReadingWidget/Info.plist
@@ -17,7 +17,7 @@
 	<key>CFBundlePackageType</key>
 	<string>XPC!</string>
 	<key>CFBundleShortVersionString</key>
-	<string>7.7.8</string>
+	<string>7.8.0</string>
 	<key>CFBundleVersion</key>
 	<string>0</string>
 	<key>NSExtension</key>
--- a/NotificationServiceExtension/Info.plist
+++ b/NotificationServiceExtension/Info.plist
@@ -17,7 +17,7 @@
 	<key>CFBundlePackageType</key>
 	<string>$(PRODUCT_BUNDLE_PACKAGE_TYPE)</string>
 	<key>CFBundleShortVersionString</key>
-	<string>7.7.8</string>
+	<string>7.8.0</string>
 	<key>CFBundleVersion</key>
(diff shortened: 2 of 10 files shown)


WMF Framework/CommonStrings.swift
```diff
@@ -765,6 +767,9 @@ public class CommonStrings: NSObject {
     public static let savedAllTabsViewReadingList = WMFLocalizedString("saved-all-tabs-view-reading-list", value: "Saved all tabs. View reading list.", comment: "Message indicating all tabs were saved with option to view reading list.")
     public static let mainPage = WMFLocalizedString("main-page", value: "Main page", comment: "Title for the Wikipedia main page.")
 
+    public static let tabsButtonAccessibilityHint = WMFLocalizedString("tabs-button-accessibility-hint", value: "Navigates to tabs overview", comment: "Accessibility hint for the tabs navigation bar button")
```
l-olson1214: Should we name this article-tabs-xyz for consistency / translations?

mazevedofs: Yes! Also, on the accessibility label I just added "tabs". Should it be "article tabs"?

PR #5373, opened 2025-08-01
T398890: Did You Know

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T398890

### Notes
* 

### Test Steps
1. Switch the type in the experiment code to be .didYouKnow
2. Navigate to tabs overview
3. Ensure Did You Know shows
4. Ensure fact + article display properly 
5. Ensure tapping link in fact opens new tab
6. Ensure facts change
7. Change language, verify facts load correctly
8. Verify cached facts load correctly

### Screenshots/Videos

https://github.com/user-attachments/assets/f7aea055-8512-451e-a9bb-ca16a9dd68b8

Simulator Screenshot - iPhone 16 Plus - 2025-08-01 at 13 24 52


Wikipedia/Code/NewArticleTabCoordinator.swift
```diff
@@ -135,25 +138,74 @@ final class NewArticleTabCoordinator: Coordinator {
...
+        return String.localizedStringWithFormat(fromLanguageWikipedia, localizedLanguageString)
+    }
+    
+    private func fetchDYK(completion: @escaping ([WMFFeedDidYouKnow]?) -> Void) {
+        guard let languageCode = dataStore.languageLinkController.appLanguage?.languageCode else {
+            return
+        }
+        guard let url = URL(string: "https://\(languageCode).wikipedia.org") else {
+            completion(nil)
+            return
```
mazevedofs: It's preferable to obtain the URL through some of our existing Utils for consistency, such as the methods from URL+LinkParsing.

WMFComponents/Sources/WMFComponents/Components/Article Tabs/New Tab/DYK/WMFNewArticleTabDidYouKnowViewModel.swift
```diff
@@ -0,0 +1,63 @@
...
+    public let dykLocalizedStrings: LocalizedStrings?
+    
+    public init(facts: [String]? = nil, fromSourceDefault: String, languageCode: String?, dykLocalizedStrings: LocalizedStrings?) {
+        self.facts = facts
+        self.fromSourceDefault = fromSourceDefault
+        self.languageCode = languageCode
+        self.dykLocalizedStrings = dykLocalizedStrings
+    }
+
+    public var dyk: String? {
```
mazevedofs: I'm not opposed to abbreviations, but here it is not really clear what this prop is - is it a did you know fact?

WMFComponents/Sources/WMFComponents/Components/Article Tabs/New Tab/DYK/WMFNewArticleTabDidYouKnowViewModel.swift
```diff
@@ -0,0 +1,63 @@
...
+import WMFData
+
+@objc public final class WMFNewArticleTabDidYouKnowViewModel: NSObject, ObservableObject {
+    @Published public var facts: [String]? = nil
+    
+    public let fromSourceDefault: String
+    public let languageCode: String?
+    public let dykLocalizedStrings: LocalizedStrings?
+    
+    public init(facts: [String]? = nil, fromSourceDefault: String, languageCode: String?, dykLocalizedStrings: LocalizedStrings?) {
```
mazevedofs: I don't think it makes sense to allow the view model to be initialized without valid did you know facts.
If the facts are nil for some reason, you won't need to display the did you know view at all.

WMF Framework/Did You Know/WMFFeedDidYouKnow.swift
```diff
@@ -0,0 +1,12 @@
+import Foundation
+
+@objc public class WMFFeedDidYouKnow: NSObject, Decodable {
+    @objc public let html: String
+    @objc public let text: String
+    
+    @objc public init(html: String, text: String) {
+        self.html = html
+        self.text = text
+        super.init()
```
mazevedofs: I'm not sure this super.init call is needed.

WMFComponents/Sources/WMFComponents/Components/Article Tabs/New Tab/DYK/WMFNewArticleTabDidYouKnowViewModel.swift
```diff
@@ -0,0 +1,63 @@
...
+        self.languageCode = languageCode
+        self.dykLocalizedStrings = dykLocalizedStrings
+    }
+
+    public var dyk: String? {
+        guard let randomElement = facts?.randomElement() else { return nil }
+        let cleanedText = replaceEllipsesWithSpace(in: randomElement)
+
+        guard let dykPrefix = dykLocalizedStrings?.dyk, !dykPrefix.isEmpty else {
+            return cleanedText
```
mazevedofs: Is there a reason for the localized strings to be optional?

Wikipedia/Code/NewArticleTabCoordinator.swift
```diff
@@ -135,25 +138,74 @@ final class NewArticleTabCoordinator: Coordinator {
...
                 theme: self.theme,
                 viewModel: viewModel
             )
-            
+
+            self.fetchDYK { facts in
+                DispatchQueue.main.async {
+                    dykVM.facts = facts?.map { $0.html }
+                }
+            }
```
mazevedofs, line 227: You should only create the did you know view model if the call succeeds (whether you're fetching from the API or the cache)
If the facts are nil, you won't need the view model.
So you can first fetch the facts, and if it succeeds, you add the view model to the WMFNewArticleTabViewModel

Wikipedia/Code/NewArticleTabCoordinator.swift
```diff
@@ -135,25 +138,74 @@ final class NewArticleTabCoordinator: Coordinator {
...
+
             self.navigationController.pushViewController(vc, animated: true)
         }
+
         return true
     }
+    
+    // MARK: - DYK
+    
+    let fromLanguageWikipedia = WMFLocalizedString("new-article-tab-from-language-wikipedia", value: "from %1$@ Wikipedia", comment: "Text displayed as Wikipedia source on new tab. %1$@ will be replaced with the language.")
```
mazevedofs: We can check with design on this, but the "from language Wikipedia" format doesn't work in all languages.
Example from Portuguese: 
"da Wikipédia em Português" which roughly translates as "From Wikipedia in Portuguese", and not "From Portuguese Wikipedia".
So I'm not sure how that would be put togheter in other languages.
It might just work out in translate wiki, but I thought about flagging it.

Wikipedia/Code/NewArticleTabCoordinator.swift
```diff
@@ -135,25 +138,74 @@ final class NewArticleTabCoordinator: Coordinator {
                         relatedArticles: relatedRecords
                     )
                     becauseVM?.onTapArticle = onTapArticleAction
-
                 }
             }
-
+            
+            let dykVM = WMFNewArticleTabDidYouKnowViewModel(
```
mazevedofs: Now that we have the ABC group separation in place, we should only fetch did you know if the user was assigned to the group, same for because you read.
Making both calls all the time is making the new tab very slow.

Wikipedia/Code/NewArticleTabCoordinator.swift
```diff
@@ -135,25 +138,74 @@ final class NewArticleTabCoordinator: Coordinator {
...
+            return
+        }
+
+        dykFetcher.fetchDidYouKnow(withSiteURL: url) { [weak self] error, facts in
+            guard error == nil else {
+                completion(nil)
+                return
+            }
+            self?.dykFacts = facts
+            completion(facts)
```
mazevedofs: You can add the caching here; the sharedCache prop is not being used.

mazevedofs (changes requested):

Wikipedia/Code/NewArticleTabCoordinator.swift
```diff
@@ -110,50 +117,146 @@ final class NewArticleTabCoordinator: Coordinator {
...
-                let seedRecord = seed.toHistoryRecord()
-                let relatedRecords = related.compactMap { article in
-                    article.toHistoryRecord()
+            let experiment = try? self.dataController.getMoreDynamicTabsExperimentAssignment()
+            let devSettingsDataController = WMFDeveloperSettingsDataController.shared
+            let enableBYR = devSettingsDataController.enableMoreDynamicTabsBYR
+            let enableDYK = devSettingsDataController.enableMoreDynamicTabsDYK
+
+            if enableDYK || experiment == .didYouKnow {
+                self.fetchDYK { facts in
```
mazevedofs: This call is being made inside the loadNextBatch call, which is only needed for Because You Read - it's causing the new tab to load slowly in both cases. 
Check for `enableBYR || experiment == .becauseYouRead` before making the call loadNextBatch and you can keep all that logic separate, same for did you know.

Wikipedia/Code/NewArticleTabCoordinator.swift
```diff
@@ -110,50 +117,146 @@ final class NewArticleTabCoordinator: Coordinator {
...
+                    DispatchQueue.main.async {
+                        let dykVM = WMFNewArticleTabDidYouKnowViewModel(
+                            facts: facts?.map { $0.html } ?? [],
+                            languageCode: self.dataStore.languageLinkController.appLanguage?.languageCode,
+                            dykLocalizedStrings: WMFNewArticleTabDidYouKnowViewModel.LocalizedStrings.init(
+                                dyk: self.dyk,
```
mazevedofs: Using `dykTitle` or `didYouKnowTitle` for both makes it clearer - using only dyk is confusing as to what this property is

Wikipedia/Code/NewArticleTabCoordinator.swift
```diff
@@ -110,50 +117,146 @@ final class NewArticleTabCoordinator: Coordinator {
...
+        let today = calendar.startOfDay(for: Date())
+        let stringToday = today.formatted()
+        let key = "dyk-last-fetch-date"
+        
+        let lastChecked = try? userDefaultsStore?.load(key: key) ?? ""
+        
+        let wasCheckedToday = stringToday == lastChecked
 
+        let cached = sharedCache.loadCache() ?? DidYouKnowCache()
+        let facts = cached.facts
```
mazevedofs: These are always `nil`, I'm not sure if it's failing to save or load.
So we're calling the endpoint every time.

Wikipedia/Code/NewArticleTabCoordinator.swift
```diff
@@ -110,50 +117,146 @@ final class NewArticleTabCoordinator: Coordinator {
...
+    private func fromLanguageWikipediaTextFor(languageCode: String?) -> String {
+        guard let languageCode = languageCode, let localizedLanguageString = Locale.current.localizedString(forLanguageCode: languageCode) else {
+            return fromWikipediaDefault
+        }
+
+        return String.localizedStringWithFormat(fromLanguageWikipedia, localizedLanguageString)
+    }
+    
+    private func fetchDYK(completion: @escaping ([WMFFeedDidYouKnow]?) -> Void) {
+        guard let url = URL(string: Configuration.current.defaultSiteDomain) else {
```
mazevedofs: Here, the URL you're sending is "wikipedia.org". The call works for en wiki because, `feedContentAPIURLForURL` on the fetcher calls a func that uses our API components builder.
If there's no host - which is the case here, we fallback to en wiki host domain.
You can add a breakpoint  here to check this
https://github.com/wikimedia/wikipedia-ios/blob/61bc79433dbb4feb10a5178a36ec410ab0cb0be1/Wikipedia/Code/APIURLComponentsBuilder.swift#L60

However, it causes the fetcher to only fetch from en wiki.

Screenshot 2025-08-11 at 14 01 20

You could add a method similar to this one that takes a URL instead of NSURL to create the base url with the correct language

https://github.com/wikimedia/wikipedia-ios/blob/61bc79433dbb4feb10a5178a36ec410ab0cb0be1/WMF%20Framework/URLComponents%2BExtensions.swift#L83

WMF Framework/URLComponents+Extensions.swift
```diff
@@ -83,5 +83,8 @@ extension URLComponents {
     public func wmf_URLWithLanguageVariantCode(_ code: String?) -> URL? {
         return (self as NSURLComponents).wmf_URL(withLanguageVariantCode: code)
     }
-
+    
+    public func wmf_URL(_ code: String?) -> URL? {
```
tonisevener, line 87: Is this addition necessary? Seems like the method on line 83 does the same thing.

Wikipedia/Code/NewArticleTabCoordinator.swift
```diff
@@ -2,27 +2,133 @@ import UIKit
...
 
+    private var seed: WMFArticle?
+    private var related: [WMFArticle] = []
+    private var seenSeedKeys = Set<String>()
+    private let contentSource = WMFRelatedPagesContentSource()
+
+    // MARK: - Did you know props
+    private var dykFetcher: WMFFeedDidYouKnowFetcher
+    public var dykFacts: [WMFFeedDidYouKnow]? = nil
+    private let fromLanguageWikipedia = WMFLocalizedString("new-article-tab-from-language-wikipedia", value: "from %1$@ Wikipedia", comment: "Text displayed to indicate Did You Know source displayed on a new tab. %1$@ will be replaced with the Wikipedia language set as the app default")
```
tonisevener, line 24: There is already a string in the Explore feed that uses "From {language} Wikipedia". Might be able to lean on that instead of creating a new string that is missing translations.

Screenshot 2025-08-12 at 5 21 04 PM

Wikipedia/Code/NewArticleTabCoordinator.swift
```diff
@@ -2,27 +2,133 @@ import UIKit
...
+                    dataStore: self.dataStore,
+                    theme: self.theme,
+                    viewModel: viewModel
+                )
+                
+                self.navigationController.pushViewController(vc, animated: true)
+            }
+        } else if enableDYK || experiment == .didYouKnow,
+                  let primaryLanguage = self.dataStore.languageLinkController.appLanguage?.languageCode {
+            self.fetchDYK(for: primaryLanguage) { facts in
```
tonisevener, line 85: (Optional due to experiment, unless complaints cause a must-fix bug to appear)

I don't think coorrdinators should fetch network data like this. After the user taps, the app fetches network data but doesn't push the new screen until the fetch completes. This will cause a hang and will make the app feel broken while the network call completes, especially for users on a slow network.

Ideally, we push on WMFNewArticleTabViewController immediately without any data fetching. Then WMFNewArticleTabViewController displays a spinner while we fetch the data it needs from either the dyk network data or the byr local data.

WMF Framework/Did You Know/WMFFeedDidYouKnowFetcher.swift
```diff
@@ -0,0 +1,64 @@
+import WMFData
+
+public final class WMFFeedDidYouKnowFetcher: Fetcher {
```
tonisevener, line 3: (Optional due to experiment) Putting this in a WMFData data controller is preferred over subclassing the legacy fetcher.

WMFComponents/Sources/WMFComponents/Components/Article Tabs/New Tab/DYK/TextViewWrapper.swift
```diff
@@ -0,0 +1,67 @@
+import UIKit
+import SwiftUI
+
+public struct TextViewWrapper: UIViewRepresentable {
```
tonisevener, line 4: (Optional due to experiment) I still feel like we could have kept this as pure SwiftUI and removed the regex cleanup. See my older `T398890_did-you-know-suggestion` branch.

Wikipedia/Code/SearchViewController.swift
```diff
@@ -646,6 +648,50 @@ class SearchViewController: ThemeableViewController, WMFNavigationBarConfiguring
...
+            navigate(to: url.absoluteURL, userInfo: userInfo)
+        }
+        
+        // first try to navigate using LinkCoordinator. If it fails, use the legacy approach.
+        let navController = customArticleCoordinatorNavigationController
+            ?? self.navigationController
+            ?? self.parent?.navigationController
+
+        if let navController {
+            let linkCoordinator = LinkCoordinator(navigationController: navController, url: url.absoluteURL, dataStore: nil, theme: theme, articleSource: .undefined)
```
tonisevener, line 669: Just checking, I notice tapping a DYK link opens in the old tab instead of the new tab that the user just tried to create:

https://github.com/user-attachments/assets/a822f54c-ea90-45cd-b0f4-abcb0d0b4917

Is that expected? If not might want to try passing in a different tabConfig value in this LinkCoordinator init.

WMF Framework/Did You Know/WMFFeedDidYouKnowFetcher.swift
```diff
@@ -0,0 +1,64 @@
...
+    private let userDefaultsStore = WMFDataEnvironment.current.userDefaultsStore
+
+    public func fetchDidYouKnow(withSiteURL siteURL: URL, completion: @escaping (Error?, [WMFFeedDidYouKnow]?) -> Void) {
+
+        let sharedCache = SharedContainerCache.init(fileName: self.cachedFileName(for: siteURL), subdirectoryPathComponent: SharedContainerCacheCommonNames.didYouKnowCache)
```
tonisevener, line 8: I don't think this cache is working - I see the dyk endpoint attempting to fetch every time.

tonisevener (changes requested):
Noted the bigger cleanup things as optional, since this is meant to be more of a quick experiment. The remaining things felt like they could be small so did not flag as optional. Thanks!

tonisevener (approved):

PR #5381, opened 2025-08-12
Tabs V2 thumbnail

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T399332

### Notes
* Updates copy for the new tab thumbnail
* When in tabs V2, tabs with the main page will become a new tab, and push to the new tab experience
* When cancelling the search, a new "empty tab" is added to the overview 
* The new tab is still a tab with the main page, since tabs have to own one article

### Test Steps
1. Enable tabs V2
2. On the overview, click add new tab
3. Click cancel
4. On the tab overview, verify that a new tab was added
5. Tap this new tab. Verify that you go to the new tab experience.
6. Disable tabs v2
7. Verify the main page tabs work normally


Wikipedia/Code/WMFNewArticleTabViewController.swift
```diff
@@ -160,16 +159,32 @@ final class WMFNewArticleTabViewController: WMFCanvasViewController, WMFNavigati
...
 
-    @objc func userDidTapTabs() {
-        navigationController?.popViewController(animated: true)
-        showTabsOverview?()
+    @MainActor
+    @objc private func goToTabsOverview() {
+        let articleTabsDataController = WMFArticleTabsDataController.shared
+        Task {
+            do {
+                _ = try await articleTabsDataController.createArticleTab(initialArticle: nil, setAsCurrent: true)
```
tonisevener, line 190: I'm not sure if this is the problem area, but I notice that when I create a new tab grid item by cancelling the search bar, then tap that new tab item, it spins up a new tab instead of updating the latest new tab (see the end of this video).



https://github.com/user-attachments/assets/cd050f87-72f5-46e2-b0d3-89d2c421830e

Wikipedia/Code/WMFNewArticleTabViewController.swift
```diff
@@ -160,16 +159,32 @@ final class WMFNewArticleTabViewController: WMFCanvasViewController, WMFNavigati
...
+
+                if let viewContext = articleTabsDataController.coreDataStore?.persistentContainer?.viewContext {
+                     await viewContext.perform {
+                        viewContext.processPendingChanges()
+                    }
+                }
```
tonisevener: I'd rather not expose these Core Data details. What problem is it fixing? It seems to act the same to me with or without it, but maybe I'm missing something.

Wikipedia/Code/WMFNewArticleTabViewController.swift
```diff
@@ -160,16 +159,32 @@ final class WMFNewArticleTabViewController: WMFCanvasViewController, WMFNavigati
...
-    @objc func userDidTapTabs() {
-        navigationController?.popViewController(animated: true)
-        showTabsOverview?()
+    @MainActor
+    @objc private func goToTabsOverview() {
+        let articleTabsDataController = WMFArticleTabsDataController.shared
+        Task {
+            do {
+                _ = try await articleTabsDataController.createArticleTab(initialArticle: nil, setAsCurrent: true)
+
```
tonisevener, line 191: I also notice tapping a new tab, then canceling out of it spins up a second tab:

https://github.com/user-attachments/assets/ac843f53-f96c-4057-8634-878ed831f8ae

tonisevener (commented):
Just one question about the Core Data stuff, and noting some bugs I may have found. If you think the bugs could be a different reason, I'm ok with approving this and us tackling it in followups.

Wikipedia/Code/WMFNewArticleTabViewController.swift
mazevedofs: ~~Without this, when cancelling, sometimes the new tab doesn't show right away. I did this to force the update to propagate to the UI faster. It wasn't consistent.
It was the only solution I found, I'm totally open to alternatives. Maybe burying this deeper into the data controller?~~ 
Edit: adding @\MainActor was enough. Removing this!

PR #5388, opened 2025-08-18
[YiR V3] UI Refactor

tonisevener:
**Phabricator:** N/A

### Notes
These are some UI adjustments that I am doing to allow for some of our more varied slide designs coming up in YiR V3. This restructuring allows me to insert slides that do not fit the gif, title, subtitle pattern. I am inserting a mock location slide at the very end to demonstrate.

The adjustments are:
1. Brought back some of the fallback images that we will need when V3 goes out and deleted some unused strings.
2. Moved some more logic from the `YearInReviewCoordinator` and `WMFYearInReviewView` into the `WMFYearInReviewViewModel`. I'm basically trying to have the view model keep track of its own data a bit more.
3. Modified the view model's slides array to keep track of an array of enums. We can have the specific slide view model of any UI we need attached in as an enum associated type. This way we don't have to have one slide view model that represents all possible different types of slides, since those are starting to vary wildly. This allows us to have different slide view models for different slide designs.
4. Broke off WMFYearInReviewView into smaller subviews, due to some compiler errors I was seeing in the time it takes to evaluate.

I think this is fairly safe to merge whenever, because we blocked off all of the entry points in https://github.com/wikimedia/wikipedia-ios/pull/5334. 

### Test Steps
1. Fresh install Staging app in simulator
2. Log into app.
3. Year in Review option should appear in Profile, tap through. Ensure last slide is a mock location slide that will be free to have any sort of design.


WMFComponents/Sources/WMFComponents/Components/Year in Review/WMFYearInReviewSlideViews.swift
```diff
@@ -0,0 +1,223 @@
+import SwiftUI
```
l-olson1214: Out of curiosity, how did you decide to put these all in one file vs a folder with other files? I struggle determining it sometimes

WMFComponents/Sources/WMFComponents/Components/Year in Review/WMFYearInReviewViewModel.swift
```diff
@@ -86,17 +54,60 @@ public class WMFYearInReviewViewModel: ObservableObject {
...
...g, collectiveSavedArticlesSlideTitle: String, collectiveSavedArticlesSlideSubtitle: String, collectiveAmountEditsSlideTitle: String, collectiveAmountEditsSlideSubtitle: String, collectiveEditsPerMinuteSlideTitle: String, collectiveEditsPerMinuteSlideSubtitle: String, collectiveZeroAdsSlideTitle: String, collectiveZeroAdsSlideSubtitle: @escaping () -> String, personalizedYouReadSlideTitle: @escaping (Int) -> String, personalizedYouReadSlideSubtitle: @escaping (Int) -> String, personalizedDaySlideTitle: @escaping (Int) -> String, personalizedDaySlideSubtitle: @escaping (Int) -> String, personalizedSaveCountSlideTitle: @escaping (Int) -> String, personalizedSaveCountSlideSubtitle: @escaping (Int, [String]) -> String, personalizedUserEditsSlideTitle: @escaping (Int) -> String, personzlizedUserEditsSlideTitle500Plus: String, personzlizedUserEditsSlideSubtitle: @escaping (Int) -> String, personzlizedUserEditsSlideSubtitle500Plus: String, personalizedYourEditsViewedSlideTitle: @escaping (Int) -> String, personalizedYourEditsViewedSlideSubtitle: @escaping (Int) -> String, personalizedThankYouTitle: String, personalizedThankYouSubtitle: @escaping (String) -> String, locationTitle: String) {
```
l-olson1214: This is horrific lol 

An idea for future ones (or maybe a ChatGPT task) - could we separate them and have like ButtonTitlesLocalizedStrings, AccessibilityLocalizedStrings, EnglishSlidesLocalizedStrings, etc. to prevent this massive init?

WMFComponents/Sources/WMFComponents/Components/Year in Review/WMFYearInReviewViewModel.swift
```diff
@@ -114,26 +125,519 @@ public class WMFYearInReviewViewModel: ObservableObject {
...
+        var donateCountSlide: WMFYearInReviewSlideStandardViewModel?
+        var saveCountSlide: WMFYearInReviewSlideStandardViewModel?
+        var mostReadDaySlide: WMFYearInReviewSlideStandardViewModel?
+        var viewCountSlide: WMFYearInReviewSlideStandardViewModel?
+        
+        let dataController = try? WMFYearInReviewDataController()
+        
+        // Fetch YiR report for personalized data, assign to personalized slides
+        if let dataController,
+           let report = try? dataController.fetchYearInReviewReport(forYear: WMFYearInReviewDataController.targetYear) {
```
l-olson1214: Having the data controller directly connected to the view model feels weird to me - what's your take?

tonisevener: I think it's not too unusual - data controllers are a lot like a repository pattern (in fact maybe we should rename them to repository), where the view model calls into a separate object that is responsible for handling the data. The goal of our data controller is to abstract away the details of how the data is stored - a view model doesn't need to know the details of User Defaults, Core Data, or network calls, it just needs data.

- [This link](https://www.avanderlee.com/swift/repository-design-pattern/) describes it a bit more, where you can see the view model calls it directly.
- [This link](https://medium.com/tiendeo-tech/ios-repository-pattern-in-swift-85a8c62bf436) shows how repositories can internally figure out whether it's dealing with local and remote data, while hiding those details from outside callers.

The reason it feels inconsistent in our codebase is because the WMFComponents package has WMFData as a dependency, but nothing else. So if there is a working data controller set up in WMFData, we can freely access and call data controller methods directly from a view model in WMFComponents. Sometimes we have to fetch legacy data (which have not yet been converted and moved to WMFData data controllers), so we do the closure dance where the view model just calls a closure property. I think in an ideal world, once everything is properly moved over to WMFData data controllers, view models in WMFComponents would call data controllers (repositories?) directly.

WMFComponents/Sources/WMFComponents/Components/Year in Review/WMFYearInReviewViewModel.swift
tonisevener: Agreed 😭, our strings situation is getting really tiresome. One quick pro-tip - it is easy to auto-generate this init if you right-click the struct name and tap Refactor > Generate Memberwise Initializer.

Screenshot 2025-08-20 at 12 23 35 PM

I don't like how huge these structs become. It might be nice to someday set up a distinct Swift Package that holds our strings logic. It contains all of our strings files, the WMFLocalizableStrings helper method, and CommonStrings.swift. Then we set it as a dependency of WMFComponents. This way we can freely use these strings directly in WMFComponents rather than having to pass in a separate LocalizableStrings struct. I think we didn't go with this back when we made WMFComponents for 2 reasons:

1. There's some auto-magic set up with our automated strings translations scripts and PRs, and I've been waiting for a nice tech debt window to potentially break that connection and then fix it.
2. We justified that passing in the struct makes the component as a whole more flexible. Any users of the component can pass in whatever strings they want. We may want to revisit this reasoning though if these structs are too cumbersome.

WMFComponents/Sources/WMFComponents/Components/Year in Review/WMFYearInReviewSlideViews.swift
tonisevener: I would chalk this one up to laziness 😆. Good feedback, I think they should be broken out into multiple files.

I guess it's a judgement call - if it's a rather small view that only needs to be called from one view and is quite specific to that one view, I think it can live in the same file. But if starts to get large like this and contains multiple different types, and the file starts to feel hard to scroll through, it should be broken up. I should probably also break up the `WMFYearInReviewSlideDataControllers` file. Will comment when these are done.

WMFComponents/Sources/WMFComponents/Components/Year in Review/WMFYearInReviewViewModel.swift
l-olson1214: Out of curiosity, we talked about that tech debt window earlier - would this be something that I could chip away at? Or how complex do you think something like this would be?

PR #5396, opened 2025-08-25
Tabs V2 - Data cleanup + design feedback

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T402419

### Notes
* The main change on this PR is that we no longer have a New Tab View Controller - we're using SearchViewController alone to address some design feedback points (especially number 1) and to optimize data fetching and add a loading state
* A data controller was created, and fetching and loading are handled by SearchVC
* I'm addressing the following design review items on this PR: 1, 2, 3.
* Item 6 will not be addressed as it breaks navigation logic and causes too much complexity. Also not addressing nuer 10 as it wasn't the quick fix I initially thought it would be
* The logic for adding a new empty tab changed

### Test Steps
1. Run the app - you can do an eligible configuration for the experiment or enable it in dev settings
2. Verify the design review points
3. Verify data loads correctly - we now have a loading state


Wikipedia/Code/SearchViewController.swift
```diff
@@ -127,19 +144,34 @@ class SearchViewController: ThemeableViewController, WMFNavigationBarConfiguring
 
     override func viewDidLoad() {
         super.viewDidLoad()
-        if needsAttachedView {
-            embedNewTab()
+
+        definesPresentationContext = true
```
tonisevener: Is this needed?

Wikipedia/Code/SearchViewController.swift
```diff
@@ -169,6 +201,19 @@ class SearchViewController: ThemeableViewController, WMFNavigationBarConfiguring
         }
     }
 
+    override func viewWillDisappear(_ animated: Bool) {
+        super.viewWillDisappear(animated)
+
+        guard isNewTab, needsCreatingNewEmptyTab, let cameFromNewTab, !cameFromNewTab else { return }
+
+        let leavingNavStack = self.isMovingFromParent || self.isBeingDismissed || (self.navigationController?.isBeingDismissed ?? false)
```
tonisevener: Looks like this isn't used.

Wikipedia/Code/SearchViewController.swift
```diff
@@ -106,19 +106,36 @@ class SearchViewController: ThemeableViewController, WMFNavigationBarConfiguring
...
     public var tabIdentifier: WMFArticleTabsDataController.Identifiers?
 
-    // MARK: - Lifecycle
+    // MARK: - New tab properties
+
+    public let isNewTab: Bool
+    private var newTabDataController: NewArticleTabDataControlling?
+    private var newTabLoadTask: Task<Void, Never>?
+    private var loadingView: UIActivityIndicatorView?
+    private var _newTabHostViewController: UIViewController?
```
tonisevener: I might be wrong but I don't think this one is used.

Wikipedia/Code/SearchViewController.swift
```diff
@@ -892,5 +857,261 @@ extension SearchViewController: WMFNewArticleTabSettingsLoggingDelegate {
...
+
+extension SearchViewController: UITextViewDelegate {
+    func tappedLink(_ url: URL, sourceTextView: UITextView) {
+        guard let url = URL(string: url.absoluteString) else {
+            return
+        }
+
+        let legacyNavigateAction = { [weak self] in
+            guard let self else { return }
+            let userInfo: [AnyHashable : Any] = [RoutingUserInfoKeys.source: RoutingUserInfoSourceValue.talkPage.rawValue]
```
tonisevener: Not sure where this came from, but a `talkPage` userInfo looks wrong.

Wikipedia/Code/SearchViewController.swift
```diff
@@ -169,6 +201,19 @@ class SearchViewController: ThemeableViewController, WMFNavigationBarConfiguring
...
+    override func viewWillDisappear(_ animated: Bool) {
+        super.viewWillDisappear(animated)
+
+        guard isNewTab, needsCreatingNewEmptyTab, let cameFromNewTab, !cameFromNewTab else { return }
+
+        let leavingNavStack = self.isMovingFromParent || self.isBeingDismissed || (self.navigationController?.isBeingDismissed ?? false)
+
+        Task { [weak self] in
+            guard let self else { return }
+            _ = try? await self.tabsDataController.createArticleTab(initialArticle: nil, setAsCurrent: true)
```
tonisevener: (Optional) This does seem to cause the "new tab" item to display at odd moments, see video. I would expect to see it when I have landed on the new tab screen and tapped tabs overview, but I wouldn't expect to see it if I navigate on from there or tapped back. Instead I see the opposite. Marking this as Optional though because this may be more QA / Design Review realm and I know there has been a lot of back & forth there, so you can do a followup PR if they think it's a blocker.


https://github.com/user-attachments/assets/79d64906-7d59-47d3-8682-15a1dc7e1e77

Wikipedia/Code/SearchViewController.swift
```diff
@@ -452,7 +512,7 @@ class SearchViewController: ThemeableViewController, WMFNavigationBarConfiguring
             guard let self else {
```
tonisevener, line 498: Not certain the bug is in this area but I notice that tapping the search result opens in the existing tab rather than a new one. See my last navigation to "Cat":


https://github.com/user-attachments/assets/73687b67-31bc-41f1-a39d-fbe457a94a33

tonisevener (changes requested):
Just a few things that jumped out at me, hopefully we're close! 🤞

Wikipedia/Code/SearchViewController.swift
mazevedofs: This part of the code I just moved to the extension to separate all the new tab stuff. It's from the DYK work, I'll check in with Grey

Wikipedia/Code/SearchViewController.swift
mazevedofs: Adjusted that, agree it makes more sense. It also incidentally fixed the bug from the other video.

PR #5407, opened 2025-09-01
New campaign config

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T386726

### Notes
* Adds support for fundaring A/B test configuration
* Assigns users to a version based on bucket values from the config, and always displays the same version based on installID

### Test Steps
1. Run the app to staging, in a valid config - check [here](https://test.wikipedia.org/wiki/MediaWiki:AppsCampaignConfig.json) 
2. Background and foreground the app, open an article.
3. Make sure you see the banner
4. Backfound the app, close it
5. Open an article again, make sure you see the same version of the banner
6. Mark maybe later, change the date on the device to 24hs in the future
7. Make sure you see the same banner
8. Reinstall and repeat the steps until you are assigned another version of the banner


WMFData/Sources/WMFData/Data Controllers/Donor Experience/WMFFundraisingCampaignDataController.swift
```diff
@@ -354,28 +355,76 @@ import Foundation
...
+                weights[i] = 1.0
+                sum += 1.0
+            }
+            if sum <= 0 {
+                let uniform = 1.0 / Double(assets.count)
+                weights = Array(repeating: uniform, count: assets.count)
+            } else {
+                for i in 0..<weights.count { weights[i] /= sum }
+            }
+        }
```
tonisevener: All this normalization feels a little heavy to me. If we must ensure weight integrity (we can check with Android if they are), do you think we could do some simple weight additions and exit early with a nil asset if they don't add up to 1.0? Above on line 337 we can just guard against a nil randomAsset and `continue` in that case, similar to what we do when the value is empty.

The app wouldn't show a campaign, but I would rather fall back to that and force the config creator to write out the weights correctly than correct for them here.

Wikipedia/Code/WMFAppViewController.m
```diff
@@ -1020,6 +1020,11 @@ - (void)resumeApp:(dispatch_block_t)completion {
             [self hideSplashView];
             done();
         }
+        WMFFundraisingCampaignDataController *controller = [WMFFundraisingCampaignDataController sharedInstance];
+        NSUserDefaults *defaults = [NSUserDefaults standardUserDefaults];
+        [controller setInstallIDProvider:^NSString * _Nonnull{
+            return defaults.wmf_appInstallId;
+        }];
```
tonisevener: There's an `WMFDataEnvironment.current.appInstallIDUtility` closure [here](https://github.com/wikimedia/wikipedia-ios/blob/main/WMFData/Sources/WMFData/Environment/WMFDataEnvironment.swift#L24) that we're already using elsewhere that you should be able to call directly from `WMFFundraisingCampaignDataController`.

tonisevener (commented):
Just a couple of notes from code review, will test it out tomorrow.

tonisevener (changes requested):
Hi, a couple of bugs I think we should look into:

1. When I change my device language / primary app language to Japanese, I would expect to see the "JA" asset chosen (the one without A/B test verbiage) and presented on a Japanese article. I can't trigger that modal though, maybe something is filtering it out in the data controller?
2. (Optional) I notice sometimes that the campaign modal presents first, then the tooltips attempt to present and we get an "already presenting modal" error in the console. It acts fine (I would prefer that we prioritize campaigns over tooltips) but I wonder if we should integrate tooltips into the [Article modal presentation logic](https://github.com/wikimedia/wikipedia-ios/blob/main/Wikipedia/Code/ArticleViewController.swift#L512) so that we are a bit more thoughtful about it. I worry about some race condition where the tooltips happen to present first, then the campaign may be unexpectedly suppressed.

tonisevener (approved):

PR #5408, opened 2025-09-02
No longer explicitly set left/right margin through PCS.

dbrant:
It is no longer necessary to set horizontal margins explicitly through PCS. In fact, explicitly setting left/right margins can now conflict with the margins set by the upstream styles (guided by `@media` queries), and thereby cause presentation issues on different display sizes.

Let's no longer set left/right margin ourselves, and allow the upstream CSS to manage it, as nature intended.

img_0034


tonisevener (approved):
This does feel better, and I found a couple of bugs in our backlog that this might fix. Thank you @dbrant!

PR #5412, opened 2025-09-04
Fix UITraitCollection comparison to fix T386976

bo2themax:
**Phabricator:** 
https://phabricator.wikimedia.org/T386976#11149243

### Notes

The reason I'm using `[UITraitCollection hasDifferentColorAppearanceComparedToTraitCollection:]` is that I think the colour difference is what the app mostly wants here.

### Test Steps
1. Background the app
2. Foreground the app
3. Rotate the phone
4. Toggle light/dark mode

### Screenshots/Videos

https://github.com/user-attachments/assets/e457a076-67e6-405e-a5ab-7266b2f38818

### DIFF
--- a/.github/workflows/run_unit_tests.yml
+++ b/.github/workflows/run_unit_tests.yml
@@ -8,7 +8,7 @@ on:
 
 env:
   XCODE_VERSION: "16.4"
-  IOS_VERSION: "18.6"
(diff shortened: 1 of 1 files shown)


bo2themax:
Font change has issues with cell resizing before this commit, so I think it’s safe to merge.

PR #5417, opened 2025-09-07
Add Debug Xcode Configuration

bo2themax:
### Notes
* This configuration allows developers outside of the Wikipedia developer team to build and run on physical devices.
  > Currently, only the **Debug** is using this configuration.
* Add checks to the `brew_install` script to avoid unwanted upgrades for developers who are strict about tool versions.
* Update `ci_pre_xcodebuild` to set up the default configuration, which should be the same as the previous project settings.

### Test Steps
1. Run `scripts/setup`

### Screenshots/Videos

image

### DIFF
--- a/scripts/setup_bundle_id
+++ b/scripts/setup_bundle_id
@@ -60,7 +60,7 @@ function get_team_info_array() {
             if [ $? -ne 0 ]; then
                 break
             fi
-            if [ "$TEAMID" = "$identifier" ]; then
+            if [ -n "$TEAMID" ]; then
                 NAME=$(/usr/libexec/PlistBuddy -c "Print :IDEProvisioningTeams:$KEY:$i:teamName" "$XCODEPREFS" 2>/dev/null)
                 TEAM_INFO_RESULT="Team: $TEAMID\n    Name: $NAME"
                 export TEAM_INFO_RESULT


l-olson1214:
Hi @bo2themax ! 
Wanted to let you know that we saw this and agree with it, but it may take a bit to merge so we can have enough time to thoroughly test it :) 

Thanks!!

scripts/setup_bundle_id
```diff
@@ -0,0 +1,119 @@
...
+# skip prompt and exit, ususally used in CI
+if [ "${1:-}" = "ci" ]; then
+    exit 0
+fi
+
+TEAM_INFO_RESULT=""
+
+function get_team_info_array() {
+    local XCODEPREFS="$HOME/Library/Preferences/com.apple.dt.Xcode.plist"
+    local team_info
```
tonisevener: Looks like this might be unused.

scripts/setup_bundle_id
```diff
@@ -0,0 +1,119 @@
...
+    local found=false
+    TEAM_KEYS=(`/usr/libexec/PlistBuddy -c "Print :IDEProvisioningTeams" "$XCODEPREFS" | perl -lne 'print $1 if /^    (\S*) =/'`)
+    for KEY in ${TEAM_KEYS[@]}; do
+        i=0
+        while true ; do
+            TEAMID=$(/usr/libexec/PlistBuddy -c "Print :IDEProvisioningTeams:$KEY:$i:teamID" "$XCODEPREFS" 2>/dev/null)
+            if [ $? -ne 0 ]; then
+                break
+            fi
+            if [ "$TEAMID" = "$identifier" ]; then
```
tonisevener: $identifier might also be undefined.

bo2themax:
> Hi @bo2themax! Thanks for working on this one and for your patience with our review. I found a couple of minor unused vars. I also notice when I run `.scripts/setup` from a freshly-cloned state, I see this error:
> 
> `scripts/setup_bundle_id: line 15: scripts/../Configurations/OpenSourceDebug.xcconfig: No such file or directory`
> 
> So maybe we should have the script create the `Configurations` directory in root if it doesn't already exist.

Thank you @tonisevener! I haven't been able to install older virtual machines to test Xcode 15+, now it's working in a freshly cloned state with Xcode 16+. 

**Better to squash this pr if it's going into main.**

PR #5438, opened 2025-09-18
[TABS] - Group C experience

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T398890

### Notes
* This PR depends on #5435 being merged
* Since we're using the dev settings toggle and not assigning the groups at launch, we can't test yet that new users will see the empty view. That should be tested in [Create A/B/C test for more dynamic Tabs](https://phabricator.wikimedia.org/T398878)
* Existing users won't have their tabs deleted 
* I'll add the preview in a separate PR for speed purposes only 

### Test Steps
1. In dev settings, set tabs to group c
2. Close all existing tabs (make sure you can close all)
3. Launch the app again
4. Make sure you still have zero tabs


Wikipedia/Code/TabsOverviewCoordinator.swift
```diff
@@ -148,20 +150,20 @@ final class TabsOverviewCoordinator: Coordinator {
                 displayDeleteAllTabsToast: displayDeleteAllTabsToast
             )
             
-            let articleTabsView = WMFArticleTabsView(viewModel: articleTabsViewModel)
-            let hostingController = WMFArticleTabsHostingController(
+            let articleTabsView = await WMFArticleTabsView(viewModel: articleTabsViewModel)
```
tonisevener, line 152: Maybe in the future we should `@MainActor` around TabsOverviewCoordinator instead. Not a blocker though.

tonisevener (approved):

Wikipedia/Code/TabsOverviewCoordinator.swift
mazevedofs: I can add it to the follow-up PR I'm working on (previews)

PR #5444, opened 2025-09-22
T398883: Overflow Menu

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T398883

### Notes
* Toggles the user defaults key, since we don't actually have the piece to show yet I didn't create that/connect that

### Test Steps
1. Toggle on more dynamic tabs
2. Open multiple tabs (overflow menu shows right now on more dynamic tabs)
3. Toggle between hiding and unhiding the suggested articles section (it will not actually do anything right now except: change the text of the button so if it was hide, it is now show, change the eye icon, and pop up a snack bar)
4. Ensure close all tabs shows upon multiple tabs being open (different flows for tabs do not necessarily work right now)

### Screenshots/Videos

https://github.com/user-attachments/assets/07545b30-f02a-44f6-a2ac-e16d989dcd3d

### DIFF
--- a/.github/workflows/localization.yml
+++ b/.github/workflows/localization.yml
@@ -18,13 +18,17 @@ jobs:
     runs-on: macOS-latest
     steps:
     - uses: actions/checkout@v4
+      with:
+        persist-credentials: false
 
     - name: Update localizations
       continue-on-error: true
+      env:
+        APPS_BOT_TOKEN: ${{ secrets.APPS_BOT_TOKEN }}
       run: |
         $GITHUB_WORKSPACE/scripts/localization $GITHUB_WORKSPACE
-        git config user.name github-actions
-        git config user.email github-actions@github.com
+        git config user.name wmf-apps-ci
+        git config user.email wmf-apps-ci@users.noreply.github.com
         git add .
         git commit -m "Import translations from TranslateWiki"
-        git push
+        git push https://x-access-token:${APPS_BOT_TOKEN}@github.com/${{ github.repository }}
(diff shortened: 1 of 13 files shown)


WMFData/Sources/WMFData/Data Controllers/Article Tabs/WMFArticleTabsDataController.swift
```diff
@@ -135,6 +135,17 @@ public protocol WMFArticleTabsDataControlling {
         return _coreDataStore ?? WMFDataEnvironment.current.coreDataStore
     }
     
+    // Different than should SHOW article suggestions, this is based on user's preferences to hide or not
+    // Should hide = user has chosen to hide the element
+    // ! should hide = user has not chose to hide the element
+    public var shouldHideArticleSuggestions: Bool {
```
tonisevener: (Optional) we could maybe alter the naming here to be a bit more self-explanatory. I think it'll be easy to mix up the two outside of this class.

`public var userHasHiddenArticleSuggestionsTabs` and `public var doesQualifyForArticleSuggestionsExperiment`? Something like that.

WMFData/Sources/WMFData/Data Controllers/Article Tabs/WMFArticleTabsDataController.swift
```diff
@@ -151,6 +162,11 @@ public protocol WMFArticleTabsDataControlling {
     
     // MARK: - Experiment
     
+    // If a user qualifies for article suggestions
+    public var shouldShowArticleSuggestions: Bool {
```
tonisevener: (Optional) This also might overlap in logic with `shouldShowMoreDynamicTabs`, maybe we remove this method entirely?

PR #5446, opened 2025-09-23
Add HCaptcha for Create Account flow

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T412132, https://phabricator.wikimedia.org/T413909

### Notes
This PR adds the HCaptcha SDK to our project and triggers the process when we receive an HCaptcha challenge from the create account endpoint.

### Test Steps
(Optional, I suggest we lean more on QA to test)

I tested with these steps:

1. Try logging in with a wrong password for a few times with EN Wiki primary language, enough to trigger a captcha display there.
2. Go to the create account view, confirm view still displays the classic captcha.
3. Change primary app language to Test 2 Wiki.
4. Go to Create Account view, you should see HCaptcha fine print beneath the "Create Account" button.
5. Create an account. Upon tap of the "Create Account" button you will see a spinner with a black transparent overlay. If HCaptcha decides to challenge further, you may have to answer questions. If there is no challenge, overlay dismisses and account creation succeeds.

### DIFF
--- a/Wikipedia/Code/TalkPageViewModel.swift
+++ b/Wikipedia/Code/TalkPageViewModel.swift
@@ -37,7 +37,7 @@ final class TalkPageViewModel {
     private(set) var latestRevisionID: Int?
     
     static let leadImageSideLength = 80
-    
+
     var theme: Theme = .light
     private(set) var topics: [TalkPageCellViewModel] = []
     private(set) var shouldShowErrorState: Bool = false
@@ -187,7 +187,8 @@ final class TalkPageViewModel {
         headerTitle = pageTitle.namespaceAndTitleOfWikiResourcePath(with: project.languageCode ?? "en").title
         headerDescription = articleSummary?.wikidataDescription
 
-        leadImageURL = articleSummary?.imageURL(forWidth: Self.leadImageSideLength * Int(UIScreen.main.scale))
+        let leadImageWidth = UIScreen.main.wmf_listThumbnailWidthForScale().intValue
+        leadImageURL = articleSummary?.imageURL(forWidth: leadImageWidth)
         
         if let coffeeRollResult = coffeeRollFromItems(items) {
             coffeeRollText = coffeeRollResult.0.otherContent


Wikipedia/Code/WMFHCaptchaViewController.swift
```diff
@@ -0,0 +1,155 @@
...
+        }
+    }
+
+    // MARK: - HCaptcha Setup
+    private func setupHCaptcha() {
+        do {
+            // "f1f21d64-6384-4114-b7d0-d9d23e203b4a" //doesn't always challenge
+            // "45205f58-be1c-40f0-b286-07a4498ea3da" //always challenge
+            
+            guard let baseURL = URL(string: "https://meta.wikimedia.org"),
```
kostajh: for now, we should use `http://localhost` here, or just not set it (if possible). In the longer term, we intend for clients to set the host to the top level domain (excluding sub-domain), so e.g. `wikimedia.org`, `wikipedia.org`, or `mediawiki.org`, etc

tonisevener: Thanks for reviewing @kostajh! We are using our [configuration endpoint](https://en.wikipedia.org/api/rest_v1/feed/configuration) to hold these values now.

Wikipedia/Code/WMFAccountCreationViewController.swift
```diff
@@ -136,6 +145,31 @@ class WMFAccountCreationViewController: WMFScrollViewController, WMFCaptchaViewC
...
+        let color = theme.colors.secondaryText
+        let linkColor = theme.colors.link
+        let styles = HtmlUtils.Styles(font: font, boldFont: font, italicsFont: font, boldItalicsFont: font, color: color, linkColor: linkColor, lineSpacing: 1)
+        if let attributedText = try? HtmlUtils.nsAttributedStringFromHtml(localizedString, styles: styles) {
+            hcaptchaFinePrintTextView.attributedText = attributedText
+        } else {
+            hcaptchaFinePrintTextView.text = localizedString
+        }
+        
+
```
l-olson1214: Little nit picky - could we get rid of the extra spacing here? Also, do we need anything for hcaptcha localization with a special language-esque thing?

PR #5447, opened 2025-09-24
T398888: Group B + Current Tab State

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T398888

### Notes
* 

### Test Steps
1. Fresh install
2. Navigate to developer settings, toggle on tabs group B
3. Add tabs, making sure current tab is always highlighted 
4. Ensure adding a new tab creates focus on search
5. Ensure you can close all tabs, including main page

### Screenshots/Videos

https://github.com/user-attachments/assets/07b88709-b4d3-4bcb-879a-7b7bf34d0c86


WMFData/Sources/WMFData/Data Controllers/Article Tabs/WMFArticleTabsDataController.swift
```diff
@@ -426,7 +426,9 @@ public protocol WMFArticleTabsDataControlling {
             try self.deleteAllTabs(moc: moc)
         }
         
-        _ = try? await self.createArticleTab(initialArticle: nil, setAsCurrent: true)
+        if !moreDynamicTabsGroupBEnabled {
```
mazevedofs: Here it should be `!needsMoreDynamicTabs` and it'll solve the issue with creating a new tab for group C when clicking close all tabs

WMFComponents/Sources/WMFComponents/Components/Article Tabs/WMFArticleTabsView.swift
```diff
@@ -268,6 +268,14 @@ fileprivate struct WMFArticleTabsViewContent: View {
...
         .background(Color(theme.chromeBackground))
         .cornerRadius(12)
+        .overlay {
+            RoundedRectangle(cornerRadius: 12)
+                .inset(by: 0.5)
+                .stroke(viewModel.isCurrentTab(tab) ? Color(uiColor: theme.secondaryText) : Color.clear, lineWidth: 1)
+        }
+        .task {
+            await viewModel.refreshCurrentTab()
+        }
```
mazevedofs, line 302: When closing the current tab, no other tab becomes highlighted as the current.
We actually set another tab as the current, so maybe we have something called `.onChange` for the current tab that forces the UI to highlight the new current tab. Not sure if .onChange will work well with the overlay you created here, maybe watch a `@Published` var with the current tab

https://github.com/wikimedia/wikipedia-ios/blob/d63f134179fc6b858be4a0a6cb963cbed6b2d969/WMFData/Sources/WMFData/Data%20Controllers/Article%20Tabs/WMFArticleTabsDataController.swift#L775-L782

https://github.com/user-attachments/assets/e79b8a63-4829-4783-9958-a3a99506c2e0

l-olson1214: Nice catch! Updated!

WMFData/Sources/WMFData/Data Controllers/Article Tabs/WMFArticleTabsDataController.swift
l-olson1214: Done!

tonisevener:
@mazevedofs 

> When closing the last tab, and then clicking 'Add new tab", when I return to the overview, I have two new tabs.
Is that expected? (that was in both test groups)

I believe I have fixed this, let me know if you still have issues.

mazevedofs (approved):
Looking good! Thanks y'all!

PR #5448, opened 2025-09-25
T402702: Percentile Reading Information

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T402702

### Notes
* 

### Test Steps
1. Go to YIR (with dev settings on, logged in, etc) 
2. Ensure that you see the reading slide first 
3. Ensure that if you have more than 335 read articles (or if you edit the code to be) -
4. You see the slide that says your top x%, and it correctly maps to Shay's notes (see ticket comment)
5. If you do NOT have more than 335 read articles (or are logged out) ensure the general reading slide shows

### Screenshots/Videos
simulator_screenshot_69B69491-3F1F-4CAE-93AA-1539A3197DFE


WMFComponents/Sources/WMFComponents/Components/Year in Review/View Models/WMFYearInReviewViewModel.swift
```diff
@@ -383,17 +383,19 @@ public class WMFYearInReviewViewModel: ObservableObject {
...
-                                altText: localizedStrings.personalizedYouReadAccessibilityLabel,
-                                title: localizedStrings.personalizedYouReadSlideTitleV3(readData.readCount, readData.minutesRead),
-                                subtitle: localizedStrings.personalizedYouReadSlideSubtitleV3(readData.readCount),
-                                subtitleType: .html,
-                                infoURL: aboutYiRURL,
-                                forceHideDonateButton: false,
-                                loggingID: "read_count_custom",
-                                tappedInfo: tappedInfo
-                            )
+                            if readData.readCount > 255 {
```
tonisevener: I think Product wants to only remove the percentage text if they have <= 255 (that is, if they would not be in the top 50% at least). But this conditional would hide the slide entirely in that case. Can you check with them?

Wikipedia/Code/YearInReviewCoordinator.swift
```diff
@@ -370,43 +370,63 @@ final class YearInReviewCoordinator: NSObject, Coordinator {
...
-        let containedReadCount = min(max(readCount, minReadCount), maxReadCount)
-        let percentage = 100 - (Double(containedReadCount) / Double(maxReadCount) * 100)
-        
-        let percentageString: String
-        if readCount > maxReadCount {
-            percentageString = "0.001"
-        } else {
-            percentageString = String(format: "%.3f", percentage)
+    func percentileRange(for readCount: Int) -> String {
+        switch readCount {
```
tonisevener, line 379: @l-olson1214  When I pass in a readCount of 336, I was expecting [50%](https://docs.google.com/spreadsheets/d/1fzdsQbC41Cw9bUH_n7CWx3O7WCLVs4dWaQQ_SXhROf8/edit?gid=542339239#gid=542339239) but it resolves to top 0.01 percent. Can you look into that?

Also to be safe, maybe we should have default and <= 335 return nil, and omit the first sentence if `percentileRange` evaluates to nil.

Wikipedia/Code/YearInReviewCoordinator.swift
```diff
@@ -370,43 +370,63 @@ final class YearInReviewCoordinator: NSObject, Coordinator {
         return String.localizedStringWithFormat(format, readCount, numArticlesString, numLanguagesString)
     }
     
-    func personalizedYouReadSlideTitleV3(readCount: Int, minutesRead: Int) -> String {
+    func personalizedYouReadSlideTitleV3(readCount: Int) -> String {
+        let minutesRead = readCount * 3
```
tonisevener: @l-olson1214 We should keep the `minutesRead` logic as it was, that comes from personalized user data.

Wikipedia/Code/YearInReviewCoordinator.swift
l-olson1214: That was a typo on my part! Did add that logic though - thank you!!

Wikipedia/Code/YearInReviewCoordinator.swift
l-olson1214: @tonisevener Brought it back! Sorry, got mixed up where we had what 😅

tonisevener (approved):

PR #5452, opened 2025-09-26
Detect selectable TOTP module upon login

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T403820

### Notes
This PR adjusts our login logic to detect a `TwoFactorModuleSelectAuthenticationRequest` response from the server upon login, and then makes a followup duplicate login call with a `newModule=totp` parameter. Then the user should be able to login via their two factor code as usual.

### Test Steps
For now, we should regression test normal logins. Ensure you can log into the app on this branch. If you have access to a production account with two factor, ensure you can still log in with it.

(This part maybe optional?)
Once the server-side is updated in the beta cluster:
1. Temporarily undo the [last commit](https://github.com/wikimedia/wikipedia-ios/commit/b7a5fc41f988b1f4f4a870033cee7c9783ed58b5) so that the Staging scheme hits the beta cluster.
2. Log into the beta cluster on desktop, create or set up an existing account with both TOTP and WebAuthn two factor methods.
3. Ensure you can log in from this branch. Ensure your two factor code prompt is presented in the app and that you are able to log in with it.


Wikipedia/Code/WMFAccountLoginLogoutFetcher.swift
```diff
@@ -69,12 +71,20 @@ public class WMFAccountLoginLogoutFetcher: Fetcher {
         if let captchaWord = captchaWord {
             parameters["captchaWord"] = captchaWord
         }
+        
+        if let newModule {
+            parameters["newModule"] = newModule
+            parameters["logincontinue"] = "1"
```
l-olson1214, line 77: Is this just always 1? Just curious how the API works :)

l-olson1214 (approved):

Wikipedia/Code/WMFAccountLoginLogoutFetcher.swift
tonisevener: Yep, although I don't think the value particularly matters here, as long as the parameter exists. Here's the API documentation - https://www.mediawiki.org/wiki/API:Login#Method_2._action=clientlogin

For booleans in the API, it [says](https://www.mediawiki.org/w/api.php?action=help&modules=main#main/datatype/boolean) `Boolean parameters work like HTML checkboxes: if the parameter is specified, regardless of value, it is considered true. For a false value, omit the parameter entirely.`

If the `logincontinue` parameter exists at all it is evaluated to true.

PR #5456, opened 2025-09-29
T405428: Tabs Survey

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T405428

### Notes
* 

### Test Steps
1. Fresh install
2. Choose either group B or C in dev settings
3. Navigate throughout the app, in and out of tabs/ tabs overview, keeping track of how many times it has been shown. On the 4th time opening, you should qualify for the survey
4. Make sure cancel -> tabs overview
5. Make sure submit -> tabs overview with a little pop up
6. Make sure it doesn't pop up again after the initial showing

### DIFF
--- a/WMFComponents/Sources/WMFComponents/Components/Article Tabs/WMFArticleTabsHostingController.swift
+++ b/WMFComponents/Sources/WMFComponents/Components/Article Tabs/WMFArticleTabsHostingController.swift
@@ -124,10 +124,12 @@ public class WMFArticleTabsHostingController<HostedView: View>: WMFComponentHost
     private func presentCloseAllTabsConfirmationDialog() async {
         let button1Title = viewModel.localizedStrings.cancelActionTitle
         let button2Title = viewModel.localizedStrings.closeAllTabs
+        guard let tabsCount = try? await dataController.tabsCount() else { return }
+        
 
         let alert = UIAlertController(
-            title: viewModel.localizedStrings.closeAllTabsTitle,
-            message: viewModel.localizedStrings.closeAllTabsSubtitle,
+            title: viewModel.localizedStrings.closeAllTabsTitle(tabsCount),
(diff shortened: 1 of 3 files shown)


WMFData/Sources/WMFData/Data Controllers/Article Tabs/WMFArticleTabsDataController.swift
```diff
@@ -642,38 +642,43 @@ public protocol WMFArticleTabsDataControlling {
...
         guard let deadline = calendar.date(from: deadlineComponents),
               now <= deadline else {
             return false
         }
         
-        let seenCount: Int = (try? userDefaultsStore?.load(key: WMFUserDefaultsKey.articleTabsOverviewOpenedCount.rawValue)) ?? 0
-        let didTapLongPress = (try? userDefaultsStore?.load(key: WMFUserDefaultsKey.articleTabsDidTapOpenInNewTab.rawValue)) ?? false
-        let seenSurvey = (try? userDefaultsStore?.load(key: WMFUserDefaultsKey.articleTabsDidShowSurvey.rawValue)) ?? false
+        var seenCount: Int = 0
+        if moreDynamicTabsGroupBEnabled || moreDynamicTabsGroupCEnabled {
```
mazevedofs: This could be simplified to `needsMoreDynamicTabs`, it'll also make it easier when you have the actual groups assigned, since it'll check the toggle and the assignemnt

mazevedofs (changes requested):
It didn't work for me.
I added break points here, and in line 133 `shouldShowSurvey` is always nil, so the code exits early and never builds the closure.
https://github.com/wikimedia/wikipedia-ios/blob/d63f134179fc6b858be4a0a6cb963cbed6b2d969/Wikipedia/Code/TabsOverviewCoordinator.swift#L132-L146

PR #5458, opened 2025-09-30
[YIR] T405639: Update login prompting

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T405639

### Notes
* 

### Test Steps
1. Fresh install
2. Navigate to developer settings, toggle on YIR v3
3. Go back home
4. Go to profile, YIR
5. Make sure that dismissing brings up the alert
6. Open YIR again, press continue without logging in, ensure that works

### Screenshots/Videos

https://github.com/user-attachments/assets/d5e6d60c-6c21-4349-995c-b80bd9e26100

### DIFF
--- a/Wikipedia/Code/YearInReviewCoordinator.swift
+++ b/Wikipedia/Code/YearInReviewCoordinator.swift
@@ -19,6 +19,9 @@ final class YearInReviewCoordinator: NSObject, Coordinator {
 
     let yearInReviewDonateText = WMFLocalizedString("year-in-review-donate", value: "Donate", comment: "Year in review donate button")
     weak var badgeDelegate: YearInReviewBadgeDelegate?
+    
+    // When true, presents a toast when they exit the Intro slide. Toast explains that they can access later via Profile menu.
+    public var needsExitFromIntroToast: Bool = false
 
     private var languageCode: String? {
         return dataStore.languageLinkController.appLanguage?.languageCode
@@ -1064,7 +1067,10 @@ extension YearInReviewCoordinator: UIAdaptivePresentationControllerDelegate {
     }
     
     public func presentationControllerDidDismiss(_ presentationController: UIPresentationController) {
-        WMFAlertManager.sharedInstance.showBottomAlertWithMessage(CommonStrings.youCanAccessYIR, subtitle: nil, buttonTitle: nil, image: nil, dismissPreviousAlerts: true)
+        if needsExitFromIntroToast, viewModel?.isShowingIntro ?? false {
+            WMFAlertManager.sharedInstance.showBottomAlertWithMessage(CommonStrings.youCanAccessYIR, subtitle: nil, buttonTitle: nil, image: nil, dismissPreviousAlerts: true)
+            needsExitFromIntroToast = false
+        }
     }
 }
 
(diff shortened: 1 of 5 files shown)


WMF Framework/CommonStrings.swift
```diff
@@ -647,8 +647,11 @@ public class CommonStrings: NSObject {
     public static let joinLoginTitle = WMFLocalizedString("profile-page-join-title", value: "Log in / Join Wikipedia", comment: "Link to sign up or sign in")
     
     public static let noThanksTitle = WMFLocalizedString("variants-alert-dismiss-button", value: "No thanks", comment: "Dismiss button on alert used to inform users about variant support.")
+    public static let continueWithoutLoggingIn = WMFLocalizedString("variants-alert-continue-without-logging-in", value: "Continue without logging in", comment: "Button title to allow users to continue forward without logging in.")
```
tonisevener: I see that we have an existing "Continue without logging in" string in Panels.swift - maybe you could move that to CommonStrings and reference in both places?

Wikipedia/Code/YearInReviewCoordinator.swift
```diff
@@ -697,7 +697,7 @@ final class YearInReviewCoordinator: NSObject, Coordinator {
         let title = WMFLocalizedString("year-in-review-login-title", value: "Improve your Year in Review", comment: "Title of alert that asks user to login. Displayed after they completed the feature for the first time.")
         let subtitle = WMFLocalizedString("year-in-review-login-subtitle", value: "Login or create an account to be eligible for more personalized insights", comment: "Subtitle of alert that asks user to login. Displayed after they completed the feature for the first time.")
         let button1Title = CommonStrings.joinLoginTitle
-        let button2Title = CommonStrings.noThanksTitle
+        let button2Title = CommonStrings.continueWithoutLoggingIn
```
tonisevener: I think this might be a Year in Review V2 method, displayed after the survey presents and isn't supposed to do anything when this button is tapped (see action 2 below, it just logs analytics). Can you revert this so that it still stays "no thanks"?

Wikipedia/Code/YearInReviewCoordinator.swift
```diff
@@ -996,34 +996,11 @@ extension YearInReviewCoordinator: YearInReviewCoordinatorDelegate {
     }
     
     private func showExitConfirmationPromptFromIntroV3Done() {
```
tonisevener: I was extremely descriptive on this method name, so it feels pretty outdated now 😅, maybe just rename to something like `showExitToastFromIntroV3Done`.

There's also a requirement in this task description to show the toast when they swipe down to dismiss the feature, not just tapping the "Done" button.

The `presentationControllerWillDismiss` method in this class is where you can find hooks to react to a user swiping to dismiss.

Wikipedia/Code/YearInReviewCoordinator.swift
```diff
@@ -863,7 +868,7 @@ extension YearInReviewCoordinator: YearInReviewCoordinatorDelegate {
         case .tappedIntroV3GetStartedWhileLoggedOut:
             showLoginPromptFromIntroV3GetStarted()
         case .tappedIntroV3DoneWhileLoggedOut:
-            showExitConfirmationPromptFromIntroV3Done()
+            showExitToastFromIntroV3Done()
```
tonisevener, line 1083: Would it be possible to only show the exit toast ONLY if they reached this feature from the "feature announcement" entry points? That is, from the Explore [presentation](https://github.com/wikimedia/wikipedia-ios/blob/main/Wikipedia/Code/ExploreViewController.swift#L1128) or the Article [presentation](https://github.com/wikimedia/wikipedia-ios/blob/main/Wikipedia/Code/ArticleViewController%2BAnnouncements.swift#L182)?

Feels weird to show this toast when they already reached the feature from Profile, IMO.

Wikipedia/Code/YearInReviewCoordinator.swift
```diff
@@ -855,6 +856,10 @@ extension YearInReviewCoordinator: UIAdaptivePresentationControllerDelegate {
         viewModel?.logYearInReviewDidTapDone()
         (self.navigationController as? WMFComponentNavigationController)?.turnOffForcePortrait()
     }
+    
+    public func presentationControllerDidDismiss(_ presentationController: UIPresentationController) {
+        WMFAlertManager.sharedInstance.showBottomAlertWithMessage(CommonStrings.youCanAccessYIR, subtitle: nil, buttonTitle: nil, image: nil, dismissPreviousAlerts: true)
```
tonisevener: Would it be possible to only show this toast if they swiped down on the _intro_ slide? Right now it shows when you swipe down at any point in the feature, even on the final slide. Swiping down past the first slide can trigger the survey so I'm also worried about the toast displaying on top of that.

tonisevener (approved):

PR #5459, opened 2025-09-30
T402696: Update Donor Slides

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T402696

### Notes
* 

### Test Steps
1. With a fresh install + country set to Indonesia in your schema
2. Go to developer settings, toggle on YIR V3
3. Go back to home page
4. Go to YIR in profile
5. Go through YIR
6. Ensure last slide (donation slide) does NOT show

### Screenshots/Videos

https://github.com/user-attachments/assets/bd9fa294-ffab-4eff-8d7c-3e8d498c8508

### DIFF
--- a/WMFComponents/Sources/WMFComponents/Components/Year in Review/View Models/WMFYearInReviewViewModel.swift
+++ b/WMFComponents/Sources/WMFComponents/Components/Year in Review/View Models/WMFYearInReviewViewModel.swift
@@ -701,7 +701,7 @@ public class WMFYearInReviewViewModel: ObservableObject {
                 slides.append(.standard(personalizedSlides.saveCountSlide ?? (primaryAppLanguage.isEnglishWikipedia ? englishReadingListSlide : collectiveSavedArticlesSlide)))
                 slides.append(.standard(personalizedSlides.editCountSlide ?? (primaryAppLanguage.isEnglishWikipedia ? englishEditsSlide : collectiveAmountEditsSlide)))
                 slides.append(.standard(personalizedSlides.viewCountSlide ?? (primaryAppLanguage.isEnglishWikipedia ? englishEditsBytesSlide : collectiveEditsPerMinuteSlide)))
-                slides.append(.contribution(personalizedSlides.donateCountSlideV3 ?? nonContributorSlide))
+                if shouldShowDonateButton { slides.append(.contribution(personalizedSlides.donateCountSlideV3 ?? nonContributorSlide)) }
                 slides.append(.highlights(getPersonalizedHighlights()))
             } else {
                 slides.append(.standard(primaryAppLanguage.isEnglishWikipedia ? englishHoursReadingSlide : collectiveLanguagesSlide))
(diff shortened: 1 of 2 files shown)


WMFComponents/Sources/WMFComponents/Components/Year in Review/View Models/WMFYearInReviewViewModel.swift
```diff
@@ -701,15 +701,15 @@ public class WMFYearInReviewViewModel: ObservableObject {
                 slides.append(.standard(personalizedSlides.saveCountSlide ?? (primaryAppLanguage.isEnglishWikipedia ? englishReadingListSlide : collectiveSavedArticlesSlide)))
                 slides.append(.standard(personalizedSlides.editCountSlide ?? (primaryAppLanguage.isEnglishWikipedia ? englishEditsSlide : collectiveAmountEditsSlide)))
                 slides.append(.standard(personalizedSlides.viewCountSlide ?? (primaryAppLanguage.isEnglishWikipedia ? englishEditsBytesSlide : collectiveEditsPerMinuteSlide)))
-                slides.append(.contribution(personalizedSlides.donateCountSlideV3 ?? nonContributorSlide))
+                if shouldShowDonateButton { slides.append(.contribution(personalizedSlides.donateCountSlideV3 ?? nonContributorSlide)) }
```
tonisevener: This may have been unclear, but I think if they edited in the last year, we do still want to show them the contributor slide (see [comment](https://phabricator.wikimedia.org/T402696#11221002)). So some sort of pseudocode / test steps like this might work:

```
if editCount > 0 {
   insert donateCountSlideV3
} else if donateCount > 0 and shouldShowDonateButton {
   insert donateCountSlideV3
} else if shouldShowDonateButton {
   // edit count == 0 and donate count == 0 if they made it this far
   insert nonContributorSlide
} else {
   no slide should be inserted
}
```

Wikipedia/Code/YearInReviewCoordinator.swift
```diff
@@ -54,6 +54,22 @@ final class YearInReviewCoordinator: NSObject, Coordinator {
         return "https://www.mediawiki.org/wiki/Special:MyLanguage/Wikimedia_Apps/iOS_FAQ?uselang=\(languageCode ?? "en")#Editing"
     }
     
+    private func noncontributorTitle() -> String {
+        // Check if current date is before the temporary date (January 1, 2026)
+        let calendar = Calendar.current
+        var expiryDateComponents = DateComponents()
```
tonisevener: Minor - might be good to have this expiry date components in a shared computed variable so it's only defined once.

tonisevener (changes requested):

WMFComponents/Sources/WMFComponents/Components/Year in Review/View Models/WMFYearInReviewViewModel.swift
tonisevener: I also think it might be better to break off the config portion of `shouldShowDonateButton` and only check that here. There's other logic within `shouldShowDonateButton` that I worry may mess with this implementation:

```
        ....
        if editCount > 0 {
            insert donateCountSlideV3
        } else if donateCount > 0 and !configShouldHideDonateButton {
            insert donateCountSlideV3
        } else if !configShouldHideDonateButton {
            // edit count == 0 and donate count == 0 if they made it this far
            insert nonContributorSlide
        } else {
            no slide should be inserted
        }
      ....
    }

    var configShouldHideDonateButton: Bool {
        return (try? WMFYearInReviewDataController().shouldHideDonateButton()) ?? false
    }

    var shouldShowDonateButton: Bool {
        if isShowingIntro {
            return false
        }
        
        if configShouldHideDonateButton {
            return false
        }
        
        let slide = currentSlide
        switch slide {
        case .standard(let viewModel):
            if viewModel.forceHideDonateButton {
                return false
            } else {
                return true
            }
        case .contribution(let viewModel):
            if viewModel.contributionStatus == .contributor {
                if viewModel.forceHideDonateButton {
                    return false
                }
                
                return true
            }
            return false
        default:
            return true
        }
    }
```

PR #5484, opened 2025-10-17
T402752: Inserting links when editing - no search results

Krystool:
**Phabricator:** https://phabricator.wikimedia.org/T402752

### Notes
I narrowed down the problem which is that the search function doesn't recognize strings of text pulled from the page when editing as non-whitespace text and therefore will not initiate properly unless the user adds something (even just a space), thereby making the function recognise the string as non-whitespace and then working as intended. 

(reposted PR as correctly formatted)

### DIFF
--- a/WMF Framework/HasText.swift
+++ b/WMF Framework/HasText.swift
@@ -3,7 +3,7 @@ extension String {
         return !isEmpty
     }
     public var wmf_hasNonWhitespaceText: Bool {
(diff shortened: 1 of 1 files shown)


mazevedofs:
Hi @Krystool, since it's been over 6 months since the last update, I'm closing this PR. Feel free to reopen it when you have time to address the code review comments.

Thanks!

PR #5493, opened 2025-10-22
[TABS] Loading fix

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T407912

### Notes
* Now the loading view, in V2,  adds a loding overlay so we wait until all the data is ready
* Should not change behavior for V1

### Test Steps
1. Run the app in tabs group B or C
2. Add several tabs (10+), make a tab from the bottom row current.
3. Open and close tabs overview - the loading state should disappear when the view is ready - tabs are loaded and current tab is in position, the bottom view is loaded. You should not see the flash of scroll.
4. Make other tabs current, check behavior
5. Enable and disable the bottom view, make sure it works properly
6. Close all tabs, make sure the empty state shows correctly

### DIFF
--- a/WMFComponents/Sources/WMFComponents/Components/Article Tabs/WMFArticleTabsView.swift
+++ b/WMFComponents/Sources/WMFComponents/Components/Article Tabs/WMFArticleTabsView.swift
@@ -185,7 +185,7 @@ public struct WMFArticleTabsView: View {
                 }
                 .frame(width: geometry.size.width, height: geometry.size.height, alignment: .center)
             }
-            .background(Color(theme.paperBackground))
+            .background(Color(theme.midBackground))
             .scrollBounceBehavior(.always)
         }
     }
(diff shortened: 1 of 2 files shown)


WMFComponents/Sources/WMFComponents/Components/Article Tabs/WMFArticleTabsView.swift
```diff
@@ -20,57 +20,87 @@ public struct WMFArticleTabsView: View {
...
-            VStack(spacing: 0) {
-                Group {
-                    if !isReady {
-                        loadingView
-                    } else if viewModel.articleTabs.isEmpty {
-                        emptyStateContainer
-                    } else {
-                        if viewModel.shouldShowTabsV2 {
-                            tabsV2Grid(geometry)
+            ZStack {
```
tonisevener, line 31: Still a bit nervous with how much we're changing here, I see a loading spinner on control users which doesn't seem quite right to me. Can you create the ZStack and loading overlay only within `tabsV2Grid`? Then hopefully you can do a simple `viewModel.recLoaded` check before adding the overlay instead of all the checks that `bottomDataReady` is doing.

mazevedofs:
We'll address this before scaling

PR #5716, opened 2026-03-09
Enable sending events to logging intake (logstash).

dbrant:
This expands `EventPlatformClient` to be able to send events to the `logging` intake, in addition to the usual `analytics` intake, which will allow events to be sent directly to Logstash.

This also provides you with a `ClientErrorFunnel` that lets you construct and send arbitrary client-error events to logstash.

(see additional annotations inline)


WMF Framework/Event Platform/EventPlatformClient.swift
```diff
@@ -428,15 +445,6 @@ import WMFData
...
         var meta: Meta
 
-        /**
-         * The top-level field `dt` is for recording the time the event
-         * was generated. EventGate sets `meta.dt` during ingestion, so for
-         * analytics events that field is used as "timestamp of reception" and
-         * is used for partitioning the events in the database. See Phab:T240460
-         * for more information.
-         */
-        let dt: Date
```
dbrant, line 509: The "minimal" event body (i.e. the absolute least common denominator of all analytics and logging events) actually _doesn't_ have a `dt` field.
Since the only other type of event that uses `MinimalEventBody` is `EditAttemptStep`, let's just add an explicit `dt` field to the latter, and remove it from here.

WMF Framework/Event Platform/EventPlatformClient.swift
```diff
@@ -208,12 +218,13 @@ import WMFData
...
      * the returned list of stream configurations).
      */
-    private static let streamConfigsURI = URL(string: "https://meta.wikimedia.org/w/api.php?action=streamconfigs&format=json&constraints=destination_event_service=eventgate-analytics-external")!
+    private static let streamConfigsURI = URL(string: "https://meta.wikimedia.org/w/api.php?action=streamconfigs&format=json")!
 
     /**
      * An individual stream's configuration.
      */
     struct StreamConfiguration: Codable {
+        let destination_event_service: String?
```
dbrant, line 229: Every stream configuration object has a `destination_event_service` field, which actually tells us which event service it "wants" to be streamed into. The service can be either "eventgate-analytics-external" or "eventgate-logging-external".

WMF Framework/Event Platform/EventPlatformClient.swift
```diff
@@ -208,12 +218,13 @@ import WMFData
      * be "eventgate-analytics-external" (to filter out irrelevant streams from
      * the returned list of stream configurations).
      */
-    private static let streamConfigsURI = URL(string: "https://meta.wikimedia.org/w/api.php?action=streamconfigs&format=json&constraints=destination_event_service=eventgate-analytics-external")!
```
dbrant, line 221: This API call was fetching only those stream configurations that fall under the "analytics" event gate. By removing the `constraints` parameter, we can fetch all stream configurations, including those that fall under the "logging" event gate.
This will indeed increase the payload size of this API response, and the size that you've got persisted locally, but since you refresh it relatively infrequently, it shouldn't be very noticeable.

WMF Framework/Event Platform/EventPlatformClient.swift
```diff
@@ -385,7 +396,13 @@ import WMFData
         let group = DispatchGroup()
         for event in events {
             group.enter()
-            httpPost(url: EventPlatformClient.eventIntakeURI, body: event.data) { result in
+
+            var uri = EventPlatformClient.analyticsEventIntakeURI
+            if (streamConfigurations?[event.stream]?.destination_event_service == "eventgate-logging-external") {
```
dbrant: ...We can now pick the appropriate eventgate URL based on what the stream configuration actually asks for.

Wikipedia/Code/ClientErrorFunnel.swift
```diff
@@ -0,0 +1,30 @@
...
+        enum CodingKeys: String, CodingKey {
+            case message = "message"
+            case errorClass = "error_class"
+            case errorContext = "error_context"
+            case stackTrace = "stack_trace"
+            case url = "url"
+        }
+    }
+
+    func logEvent(message: String?) {
```
dbrant, line 26: You can now call this function from anywhere, and send a completely arbitrary `message` to logstash!

Look at the other fields in the event and fill them out optionally.. Things to consider:
- `eventClass` can be the "class name" of an exception that is logged
- `stackTrace` can be a stack trace, but be mindful of character limits of the backend; perhaps log only the top ~2 or 3 lines in the trace.

Wikipedia/Code/ClientErrorFunnel.swift
```diff
@@ -0,0 +1,30 @@
...
+
+    @objc static let shared = ClientErrorFunnel()
+
+    private struct Event: EventInterface {
+        static let schema: EventPlatformClient.Schema = .clientError
+        let message: String?
+        let errorClass: String?
+        let errorContext: String?
+        let stackTrace: String?
+        let url: String?
```
dbrant, line 15: One other note:
For logging HTTP errors, this schema has a `http` object (see [schema definition](https://gitlab.wikimedia.org/repos/data-engineering/schemas-event-primary/-/blob/master/jsonschema/mediawiki/client/error/2.0.0.yaml?ref_type=heads)) that has a `status_code` that you can populate. I simply didn't know how exactly to add a sub-object to this struct, with the appropriate CodingKey logic etc.
The point is, ideally you should put the http error into `http.status_code`, instead of into `errorContext` as a custom string.

tonisevener (approved):
Thanks @dbrant!

PR #5755, opened 2026-03-27
T418950: Join Challenge

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T418950

### Notes
* 

### Test Steps 
1. Oh boy buckle in
2. Okay fresh install, you can go to dev settings, then toggle on the use relative dates
3. Make sure you open the app (explore, activity, article) and the full page sheet shows up
4. Ensure learn more brings you to the learn more page, ensure you can come back
5. Ensure joining allows you to join and that the widget is updated (there's a state of enrolled but not started that looks similar, you can also open an article) 
6. Ensure you can join from the announcement or the widget
7. You can then go back to developer settings and reset the state! This brings your enrollment back to UNENROLLED (yay to not having to reinstall!) 
8. Do your little refresh, make sure you have widget (sometimes widget gets stuck, may need to close app/rerun with the restart) installed, try joining the challenge from the widget state

### Screenshots/Videos

https://github.com/user-attachments/assets/cf4cb1dc-0616-4c60-9840-2cf2d52803d1

### Checklist
-[x] A user may join the challenge at any point between May 1st and May 31st inclusive
-[x] Joining is a one-time action: there is no re-enrollment
-[x] Once joined, the user is enrolled for the remainder of the challenge period
-[x] Only logged-in users can successfully join the challenge.
-[x] Enrollment may be tied to the user’s device, and does not need to be synced across devices at this point

# Full-page announcement
-[x] A full-page announcement is shown to all logged-in users starting May 1st
-[x] Announcement is shown on Article view, Activity, or Explore, on first open of the app on or after May 1
-[x] Users can only see the announcement a maximum of 1 time.
-[x] Displays the challenge name, rules summary, prize details, widget mention, a Join CTA, and a "Learn more" CTA
-[x] "Learn more" leads to mediawiki page https://www.mediawiki.org/wiki/Wikimedia_Apps/Team/25th_Birthday_Reading_Challenge
-[x] Dismissible: the user is not forced to join.
-[x] Must include a brief, plain-language statement clarifying that the only data tracked for the challenge is local data on device.
-[x] Full page announcement should stop showing after May 31

# Widget announcement
-[x] After someone joins the challenge, they should see the widget announcement: it can show immediately or on next app open if they navigate away before it displays
-[x] "Got it" dismisses announcement

# Joining from Widget (State: Not enrolled )
-[x] The widget also renders a Join CTA for logged-in users who have not yet joined
-[x] Tapping "Join challenge" on the Widget:
-[x] Opens app with full-page announcement on top of the activity tab (Any user who has not joined the challenge yet should be eligible to see the announcement when clicking "Join" from the Widget State. They can continue to see the announcement each time they click "Join" until they join the challenge.)
-[x] They must tap “Join challenge” on the announcement to enroll
-[x] They should not see the Widget announcement after the full-page announcement since they already have it installed
-[x] After joining, their Widget should update to State: Enrolled but not started or State: Streak ongoing: read today depending on their reading for the day


Wikipedia/Code/ActivityTabViewController.swift
```diff
@@ -304,9 +306,12 @@ final class WMFActivityTabHostingController: WMFComponentHostingController<WMFAc
                 ActivityTabFunnel.shared.logOnboardingDidAppear()
                 await dataController.setHasSeenActivityTab(true)
             } else {
-                presentSurveyIfNeeded()
+                let didPresentChallenge = await presentReadingChallengeAnnouncementIfNeeded()
+                if !didPresentChallenge {
+                    presentReadingChallengeWidgetAnnouncementIfNeeded()
+                    presentSurveyIfNeeded()
```
tonisevener: Might want to be a little careful here, triggering modals at the same time can cause one or the other to fail. Your discardable success boolean on line 309 is a fine pattern, you could use the same for presentReadingChallengeWidgetAnnouncementIfNeeded.

Also looks like the survey end date is April 15, so maybe we could remove the survey entirely to avoid the complexity?

Wikipedia/Code/ActivityTabViewController.swift
```diff
@@ -318,6 +323,52 @@ final class WMFActivityTabHostingController: WMFComponentHostingController<WMFAc
...
+        guard presentedViewController == nil else { return false }
+        guard isViewLoaded && view.window != nil else { return false }
+        guard let dataStore else { return false }
+        let isLoggedIn = dataStore.authenticationManager.authStateIsPermanent
+        guard await WMFActivityTabDataController.shared.shouldShowReadingChallengeAnnouncement(isLoggedIn: isLoggedIn) else { return false }
+        presentReadingChallengeAnnouncement(dataStore: dataStore)
+        return true
+    }
+
+    @objc func presentReadingChallengeAnnouncementFromWidget() {
```
tonisevener: This might be unused.

Wikipedia/Code/NSUserActivity+WMFExtensions.m
```diff
@@ -235,7 +235,7 @@ - (WMFUserActivityType)wmf_type {
         } else if ([page isEqualToString:@"Activity"]) {
             return WMFUserActivityTypeActivity;
         } else if ([page isEqualToString:@"NotificationSettings"]) {
```
tonisevener: I think this case was deleted in the `liquid-glass` branch, could be that it accidentally returned in a merge somewhere.

Wikipedia/Code/ReadingChallengeAnnouncementCoordinator.swift
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    var learnMoreURL: URL? {
+        var languageCodeSuffix = ""
+        if let primaryAppLanguageCode = dataStore.languageLinkController.appLanguage?.languageCode {
+            languageCodeSuffix = primaryAppLanguageCode
+        }
+
+        return URL(
+            string: "https://www.mediawiki.org/wiki/Special:MyLanguage/Wikimedia_Apps/Team/25th_Birthday_Reading_Challenge?uselang=\(languageCodeSuffix)"
```
tonisevener: There's a newish helper method for making urls like this. It probably resolves to the same thing but I'd like us to try and use this more:

```
let url = WMFProject.mediawiki.translatedHelpURL(pathComponents: ["Wikimedia Apps", "Team", "25th Birthday Reading Challenge"], section: nil, language: appLanguage)
```

Wikipedia/Code/ReadingChallengeWidgetAnnouncementCoordinator.swift
```diff
@@ -0,0 +1,42 @@
...
+    init(presentingViewController: UIViewController & WMFFeatureAnnouncing) {
+        self.presentingViewController = presentingViewController
+    }
+
+    func start() {
+        guard let presenting = presentingViewController else { return }
+        presenting.announceFeature(
+            viewModel: makeViewModel(),
+            sourceView: presenting.view,
+            sourceRect: CGRect(origin: presenting.view.center, size: .zero),
```
tonisevener: I haven't tested this, but just wanted to give a heads up: This `sourceRect` is needed for iPad, which is designed to have an arrow that points to a particular area that the feature announcement is announcing. Not sure if that applies here, but we may need to figure out something more specific to point to rather than the center of the presenting view.

Wikipedia/Code/ExploreViewController.swift
```diff
@@ -983,25 +985,48 @@ extension ExploreViewController {
...
+        }
+        let isLoggedIn = dataStore.authenticationManager.authStateIsPermanent
+        return await WMFActivityTabDataController.shared.shouldShowReadingChallengeAnnouncement(isLoggedIn: isLoggedIn)
+    }
+
+    private func presentReadingChallengeAnnouncement() {
+        guard let navigationController else { return }
+        let coordinator = ReadingChallengeAnnouncementCoordinator(navigationController: navigationController, dataStore: dataStore, theme: theme)
+        self.readingChallengeCoordinator = coordinator
+        coordinator.start()
```
tonisevener: After joining, I'm not seeing the followup widget announcement ONLY from Explore and Article. Activity tab works. This may be intentional but wanted to give you a heads up in case it's not.

tonisevener (commented):
Looking good! Some general comments here, let me know what you think. Also I notice when testing on iPhone 16e simulator, some UI is cut off:

Image

Wikipedia/Code/ExploreViewController.swift
l-olson1214: Good catch!!! I misread the ticket :) thank you!

Wikipedia/Code/ActivityTabViewController.swift
```diff
@@ -305,11 +305,12 @@ final class WMFActivityTabHostingController: WMFComponentHostingController<WMFAc
                 presentOnboarding()
                 ActivityTabFunnel.shared.logOnboardingDidAppear()
                 await dataController.setHasSeenActivityTab(true)
+            } else if await Date() < dataController.surveyEndDate ?? Date() {
```
tonisevener: (Observation from testing)

If I've already seen the survey, this conditional is still true, because it's only checking the date and not if I have seen the survey. Replacing this conditional with `dataController.shouldShowSurvey()` should cover both (internally it checks both visit count and seen flag).

Wikipedia/Code/ArticleViewController+Announcements.swift
```diff
@@ -159,11 +159,32 @@ extension ArticleViewController {
...
+        readingChallengeCoordinator = ReadingChallengeAnnouncementCoordinator(
+            navigationController: nav,
+            dataStore: dataStore,
+            theme: theme
+        )
+        readingChallengeCoordinator?.onEnroll = { [weak self] in
+            self?.presentReadingChallengeWidgetAnnouncementIfNeeded()
+        }
+        readingChallengeCoordinator?.onDismiss = { [weak self] in
+            self?.presentReadingChallengeWidgetAnnouncementIfNeeded()
```
tonisevener: Might be incorrect to have the followup present code here in onDismiss. I notice you aren't doing that in the equivalent Activity Tab [code](https://github.com/wikimedia/wikipedia-ios/pull/5755/changes#diff-67c266c3ab70f0f8eb4b54eac1b691ea816ef26f49b6de0ff2fe32c164e0451eR348). Deduping this presentation logic with Activity could help.

Also, I'm not sure if this is related, but I notice that the full page announcement continually presents if I tap X to dismiss it, it should only show once. I can repro from Explore and Article, though the [survey conditional](https://github.com/wikimedia/wikipedia-ios/pull/5755/changes#r3003245259) is preventing me from testing on Activity now.

https://github.com/user-attachments/assets/dede3b4b-79c1-4ac4-92f0-b73c9a842ee8

Wikipedia/Code/ExploreViewController.swift
```diff
@@ -1024,9 +1025,26 @@ extension ExploreViewController {
         guard let navigationController else { return }
         let coordinator = ReadingChallengeAnnouncementCoordinator(navigationController: navigationController, dataStore: dataStore, theme: theme)
         self.readingChallengeCoordinator = coordinator
+        coordinator.onEnroll = { [weak self] in
+            self?.presentReadingChallengeWidgetAnnouncementIfNeeded()
+        }
+        coordinator.onDismiss = { [weak self] in
+            self?.presentReadingChallengeWidgetAnnouncementIfNeeded()
```
tonisevener: Same here, I don't think we're supposed to present the followup onDismiss.

tonisevener (changes requested):
Just one more thing that I think should be fixed before we merge. When I close on the full page announcement it continues to present the next time, which I don't think is intentional. We should set a flag upon close so that it is not presented again. Not sure if it needs to be a new flag or if you can lean on your existing ones, up to you.

See video at https://github.com/wikimedia/wikipedia-ios/pull/5755#discussion_r3003250711.

tonisevener (approved):

PR #5766, opened 2026-03-31
T418954: Enrolled not started, Streak lost can restart

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T418954

### Notes
* 

### Test Steps
1. See previous steps for enrollment 
2. Without having read any articles today
3. Get to enrollment point - enroll, then go to widget and ensure it looks proper, buttons take you where expected, and you can start a streak
4. Then, let streak expire (you can skip to next day in device date) and make sure you come back to this (if time in challenge)

### Screenshots/Videos

https://github.com/user-attachments/assets/35aa9ee7-dc56-4670-b275-1075aa3c123d

### Checklist
Enrolled but not started
-[x] For users who have just joined the challenge, but have not read yet, show a prompt to start reading
-[x] Include Explore Wikipedia CTA
-[x] Leads to Explore (Fall-back to "Search" if user has "Explore" disabled)
-[x] Streak lost and can restart

The same state will also be used for users who lose their streak, and can still restart
-[x] Allow restarting if on or before May 31
-[x] When a streak resets to 0 and the current date is on or before May 31st, the system automatically begins listening for a new 25-day streak : no user action required
-[x] The module and widget silently reset the streak counter to 0
-[x] The user does not need to re-enroll or take any action to begin a new attempt


tonisevener (approved):
Working great!

PR #5770, opened 2026-03-31
Smooth over article toolbar transition

tonisevener:
**Phabricator:** N/A

### Notes
Something I noticed when debugging article toasts - we are losing the liquid glass bottom toolbar transition because we hide the toolbar upon Article disappearance no matter what. If we don't hide it when moving to another Article, the transition looks more blobby (though that may not be desired).

The more menu button changes are slighter than they look - when we need to update it (like with watch status), we are replacing the context menu rather than re-creating the entire more button again (which also caused transition weirdness).

Just a suggestion, there's no bug / design review task associated with this.

Old:
https://github.com/user-attachments/assets/6846f1a5-d60e-47dd-aea4-8aba12607c64

New:
https://github.com/user-attachments/assets/d4407dbe-3cbe-4b6d-be42-e220b77cb03c


mazevedofs (approved):
Looks much better!

PR #5773, opened 2026-04-01
Maybe fix Explore toggle crash

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T421867

### Notes
I couldn't reproduce this crash, so this is just a guess. `updateDefaultTab` could fire back-to-back from the KVO observation. This could cause the tab bar to configure multiple times during a modal dismissal, which could cause some weird UI crashes. Artificially calling `updateDefaultTab` twice in quick-succession in AppViewController's `observeValueForKeyPath` crashes with a stack trace that looks similar to the crash log.

Using an `isUpdatingDefaultTab` property here to gatekeep this process a little more seems to fix it.

### Test Steps
1. Test the enable/disable Explore feed toggle in Settings, confirm there are no regressions.


mazevedofs (approved):
I also can't repro, let's get it to QA to check

PR #5774, opened 2026-04-01
Update Localizable.strings

K5Dyce:
Proposing some edit(s).

**Phabricator:** 

### Notes
* 

### Test Steps
1.

### DIFF
--- a/Wikipedia/Localizations/en.lproj/Localizable.strings
+++ b/Wikipedia/Localizations/en.lproj/Localizable.strings
@@ -332,7 +332,7 @@
 "diff-thanks-login-title" = "Log in to send 'Thanks'";
 "diff-thanks-send-button-title" = "Send 'Thanks'";
 "diff-thanks-send-subtitle" = "'Thanks' are an easy way to show appreciation for an editor's work on Wikipedia. 'Thanks' cannot be undone and are publicly viewable.";
-"diff-thanks-send-title" = "Publicly send 'Thanks'";
+"diff-thanks-send-title" = "Publicly send 'Thanks'?";
 "diff-thanks-sent" = "Your 'Thanks' was sent to $1";
 "diff-thanks-sent-already" = "You’ve already sent a ‘Thanks’ for this edit";
 "diff-thanks-sent-cannot-unsend" = "Thanks cannot be unsent";


bri-tong:
## Code Review

### Summary

This PR adds a question mark to the English localization string `"Publicly send 'Thanks'"` → `"Publicly send 'Thanks'?"`, used as the heading of the Thanks confirmation/education panel (`ThankRevisionAuthorEducationPanelViewController`). While the UX intent is reasonable — turning a heading into a question for a confirmation dialog — **the change is made in the wrong file and will be overwritten on the next localization sync.**

---

### Blocking Issue

#### ❌ Wrong file modified — change will be lost

The file modified (`Wikipedia/Localizations/en.lproj/Localizable.strings`) is a **generated output** of the localization pipeline, not the source of truth. The pipeline works as follows:

1. **Source**: `WMFLocalizedString()` calls in Swift code (e.g., `Wikipedia/Code/Panels.swift` line 363)
2. **Extract**: `scripts/localization_extract` runs `xcrun extractLocStrings` → writes to `Wikipedia/iOS Native Localizations/en.lproj/Localizable.strings`
3. **Export**: `exportLocalizationsFromSourceCode()` in `Command Line Tools/Update Localizations/localization.swift` copies from `iOS Native Localizations/en.lproj/` → `Localizations/en.lproj/` (the file this PR edits)

The current source of truth in `Panels.swift:363` still reads:
```swift
heading = WMFLocalizedString("diff-thanks-send-title", value:"Publicly send 'Thanks'", comment:"Title for sending thanks panel.")
```

The `iOS Native Localizations/en.lproj/Localizable.strings` file (UTF-16LE) also confirms:
```
"diff-thanks-send-title" = "Publicly send 'Thanks'";
```

The next time the localization export runs, it will **overwrite this change** with the value from the Swift source code.

#### ✅ Correct fix

Update the `value:` parameter in `Wikipedia/Code/Panels.swift` line 363:
```swift
heading = WMFLocalizedString("diff-thanks-send-title", value:"Publicly send 'Thanks'?", comment:"Title for sending thanks panel.")
```
Then run the localization extraction/export tooling so both `.strings` files are updated consistently.

---

### Additional Considerations

- This string is translated into **65 other languages** via Translatewiki. Changing the English source semantics (statement → question) means translators will need to update their translations. Until they do, only the English version will have the question mark.
- The PR description is mostly empty — no notes, no test steps, no screenshots. It would help reviewers to understand the UX rationale for this change.

---

### Non-blocking Suggestions

- Consider updating the `qqq.lproj` translation documentation to note that this string is phrased as a question/confirmation prompt, so translators understand the intended tone.

PR #5776, opened 2026-04-01
Update Localizable.strings

K5Dyce:
Changing this heading to a question improves the UX because it clearly frames the panel as a confirmation dialog, helping the user understand they are about to perform an action.

**Phabricator:** 

### Screenshots/Videos


mazevedofs (changes requested):
hi @K5Dyce, thanks for your contribution. Copy changes need to be reviewed/ approved by our product manager. To give them visibility, could you create a ticket on our [backlog](https://phabricator.wikimedia.org/project/board/782/)? You can also refer to our [contribution guide](https://www.mediawiki.org/wiki/Wikimedia_Apps/Team/iOS#Contribute_to_development) if you need more info.

As for the change, updating this binary file `Wikipedia/iOS Native Localizations/en.lproj/Localizable.strings` will only update the English language and won't be persistent - this file is rewritten by our localization script everytime we have strings added or updated in the project.

You need to update the source for this localized string, which can be found here: 
https://github.com/wikimedia/wikipedia-ios/blob/d9944c795de01de3d9e8669d4751716467cedef7/Wikipedia/Code/Panels.swift#L363

As I said, this change needs the approval of other people in the team, so we can update you when we have an answer.

In the meantime, feel free to update the correct string, and include all the binary files that result of that change after building the project.

K5Dyce:
Thank you. I will do that.

PR #5779, opened 2026-04-03
Add Developer Toggles for Reading Challenge

l-olson1214:
**Phabricator:** 
part of overall reading challenge tickets

### Notes
* There is some other code in here (adds an image, updates state computation, localizes strings, etc) but this is not the conclusion state PR despite touching that, that will be fully finished in a different PR

### Test Steps
1. Go to developer settings and test the various settings
2. Turn ON Force Reading Challenge State
3. Select one of the states to toggle on - only one can be on at a time
4. Ensure it changes state
5. Make sure you can increment and decrement days (you cannot go outside of 1 and 25)

### Screenshots/Videos

https://github.com/user-attachments/assets/06cf89a9-73e1-45d2-9dcd-5723193a2f2b


WMFComponents/Sources/WMFComponents/Components/Developer Settings/WMFDeveloperSettingsViewModel.swift
```diff
@@ -91,6 +106,10 @@ import WMFData
             .sink { isSelected in WMFDeveloperSettingsDataController.shared.forceMaxArticleTabsTo5 = isSelected }
             .store(in: &subscribers)
 
+        enableMoreDynamicTabsV2GroupC.$isSelected
```
tonisevener: Might not be needed?

tonisevener (approved):

tonisevener (approved):

PR #5787, opened 2026-04-06
Fix notifications crash

tonisevener:
**Phabricator:** N/A

### Notes
There's a crash in the LG TestFlight external beta from the Notifications Center > More menu. This fixes it, though note that this notifications screen still has the https://phabricator.wikimedia.org/T421303 bug.

### Test Steps
1. Log in, go to notifications center.
2. Swipe notification, tap More.
3. Tap Notification Settings. Ensure app does not crash.


mazevedofs (approved):
Thanks for catching this one!

PR #5789, opened 2026-04-06
Fixes for nearby hang

tonisevener:
**Phabricator:** 

### Notes
This is the last actionable crash I could find from the LG Testflight external beta build. I don't think this one is widespread, so I don't think we have to merge this one now if we're unsure about it. I think users that hit this freeze do not have a location that is resolved for the Explore feed card (hang starts in WMFNearbyContentSource.m). There are two fixes here:

1. Removed a `dispatch_async(dispatch_get_main_queue(), ^{` wrapper in WMFNearbyContentSource's `loadNewContentInManagedObjectContext` method. None of the other content sources have this line, and the existing moc.perform calls should already happen on the correct thread.
2. Added timeout logic in LocationManager in case a location never returns.

### Test Steps
1. Light regression test of Explore feed location card (ensure you see it and can enable it). Note: for me this card disappears after I tap the enable button. This happens on the existing App Store build for me too, so I think it's an existing issue or perhaps I just live in a dead zone regarding nearby location suggestions.
2. Light regression on Places tab.

### DIFF
--- a/WMF Framework/LocationManager.swift
+++ b/WMF Framework/LocationManager.swift
@@ -52,6 +52,10 @@ final public class LocationManager: NSObject {
     /// Returns the current locationManager permission state.
     public var authorizationStatus: CLAuthorizationStatus { locationManager.authorizationStatus }
 
+    /// Timeout tracking vars
+    private var locationTimeoutItem: DispatchWorkItem?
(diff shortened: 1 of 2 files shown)


mazevedofs (approved):
Looks good!

PR #5801, opened 2026-04-13
Fix localization conflicts in Localizable.strings

mazevedofs:
**Phabricator:** N/A

### Notes
* 

### Test Steps
1. 

### Checklist


tonisevener (changes requested):
I think a good deal of the data was also deleted in these files:

iOS Native Localizations/de.lproj/Localizable.strings
iOS Native Localizations/de.lproj/Localizable.stringsdict

Can you go back in `main` to before the merge and copy those back over into this branch?

tonisevener (approved):

PR #5791, opened 2026-04-07
T418956: Completion of the Reading Challenge

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T418956

### Notes
* 

### Test Steps
1. Go into developer settings, toggle ON using hardcoded states
2. Test the different completion states
3. Ensure both sizes look correct

### Screenshots/Videos
Simulator Screenshot - iPhone 17 Pro Max liquid glass - 2026-04-07 at 11 28 35

### Checklist
Challenge complete:
-[x] When a streak of 25 consecutive days is recorded, the module immediately transitions to State: Challenge complete successfully
-[x] Completion can occur after May 31st: a user who joins May 31st has until June 25th to complete their streak (or June 26th if a streak freeze is applied - see Nice to Have)
-[x] Displays a "Collect your prize" CTA on the Widget
-[x] Tapping "Collect your prize" opens an in-app prize screen
-[] State persists until July 10th, after which the Widget is removed for all users** Cannot remove, will stay in completion state
-[x] Completion is a terminal state: the system stops listening for article opens once 25 days are reached
-[x] Tapping anywhere on the widget (other than the CTA) opens Explore

Challenge concluded: incomplete
-[x] For users who do not reach 25 days with a streak that started between 1-31 May, but did read for more than 1 day in a row, they should see State: Challenge concluded: incomplete / Streak lost and cannot restart
-[x] Show user's highest streak achieved up until that point
-[x] Persist until July 10th, after which the module is removed for all users
-[x] Tapping anywhere on the widget opens Explore

Challenge concluded: Streak lost and cannot restart
-[x] Disallow re-starting the challenge if after May 31
-[x] After May 31st, the system stops listening for new streaks for all enrolled users who have not reached 25 days
If a user loses their streak after May 31, they should see State: Challenge concluded: incomplete / Streak lost and cannot restart
-[x] Tapping anywhere on the widget opens Explore

Challenge concluded: No streak
-[x] For users who never joined, or joined and never read more than 1 day in a row starting with a start date of 1-31 May, they should see State: Challenge concluded: No streak
-[x] Persists until July 10th, after which the module is removed for all users
-[x] Tapping anywhere on the widget opens Explore

Prize screen
-[x] Prize screen can be triggered any time from the [Challenge completed state of the widget](https://www.figma.com/design/nnXvcESYGNFiMmxYA41vVG/WP25--Apps-Reading-Challenge?node-id=98-3576&t=6nrOkakt7XUobPUR-4)
[...]

PR #5805, opened 2026-04-13
Reading Challenge - Logged Out

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T418951

### Notes
* Adds logged-out treatment to challenge flows
* Also fixes the %% in the discount string

### Test Steps
1. Log out of the app, force the challenge completed state in the widget
2. Click the "Collect Prize"in the widget, you'll be taken to the activity tab
3. Log in
4. You should see the modal after logging in
5. Log out, fresh install
6. Click "Join" in the onboarding view
7. Log in
8. You should be taken to the activity tab

### Checklist


Wikipedia/Code/ActivityTabViewController.swift
```diff
@@ -234,13 +256,15 @@ final class WMFActivityTabHostingController: WMFComponentHostingController<WMFAc
...
             WMFToastPresenter.shared.dismissCurrentToast()
+            loginSuccessCompletion?()
         }
 
         loginCoordinator.createAccountSuccessCustomDismissBlock = {
             WMFToastPresenter.shared.dismissCurrentToast()
             if let createVC = nav.presentedViewController {
                 createVC.dismiss(animated: true)
             }
+            loginSuccessCompletion?()
```
tonisevener: Have you tested the create account flow? You might need to add this to the create account completion in this way:

```
if let createVC = nav.presentedViewController {
    createVC.dismiss(animated: true) {
         loginSuccessCompletion?()
    }
} else {
    loginSuccessCompletion?()
}
````

File: Wikipedia/Code/ReadingChallengeAnnouncementCoordinator.swift [hunk 2]
```diff
@@ -140,10 +141,28 @@ final class ReadingChallengeAnnouncementCoordinator: NSObject, Coordinator {
...
+        } else {
+            let alert = UIAlertController(
+                title: WMFLocalizedString("reading-challenge-login-title", value: "Log in to join the challenge", comment: "Title for alert that asks users to log in to join the reading challenge"),
+                message: WMFLocalizedString("reading-challenge-login-message", value: "Log in or create an account to track your progress towards the reading challenge.", comment: "Message for alert that asks users to log in to join the reading challenge"),
+                preferredStyle: .alert
+            )
+            alert.addAction(UIAlertAction(title: CommonStrings.joinLoginTitle, style: .default) { [weak self] _ in
+                guard let self else { return }
+                self.navigationController.presentedViewController?.dismiss(animated: true) {
+                    self.navigationController.wmf_showLoginViewController(category: .login, theme: self.theme, loginDismissedCompletion: {
```
tonisevener: Optional: We could use `LoginCoordinator` here instead.

Wikipedia/Code/WMFAppViewController.m
```diff
@@ -1274,11 +1274,18 @@ - (BOOL)processUserActivity:(NSUserActivity *)activity animated:(BOOL)animated c
             [self dismissPresentedViewControllers];
             [self showRandomArticleFromShortcutWithSiteURL:[self siteURL] animated:animated];
             break;
-        case WMFUserActivityTypeActivity:
+        case WMFUserActivityTypeActivity: {
             [self dismissPresentedViewControllers];
             [self setSelectedIndex:WMFAppTabTypeRecent];
             [self.currentTabNavigationController popToRootViewControllerAnimated:animated];
-            break;
+            BOOL shouldCollectPrize = [activity.userInfo[@"collectPrize"] boolValue];
```
tonisevener, line 1281: Thanks for catching this - I swear I thought this worked before. Looks like it was lost in https://github.com/wikimedia/wikipedia-ios/pull/5791/changes/588406a1b8b8156bfdd650d6e0f6b2787eb20208#diff-53892ba35877905bfabb1572472302074e7251ac44ee8ac623d2b0be877d4953.

tonisevener (changes requested):
Also maybe a bug? I can't get into this state so I can't test task requirements fully:

> Log out, fresh install
> Click "Join" in the onboarding view

> Make the full-page announcement visible to logged-out users with the same base behavior

I can't seem to trigger the announcement when logged out, either from Explore, Article, Activity, or "Join" in the widget. There must be some conditionals somewhere that we need to remove.

Wikipedia/Code/WMFAppViewController.m
l-olson1214: Sorry about that!! Thank you for catching!

Wikipedia/Code/ReadingChallengeAnnouncementCoordinator.swift
```diff
@@ -140,10 +141,30 @@ final class ReadingChallengeAnnouncementCoordinator: NSObject, Coordinator {
...
+                title: WMFLocalizedString("reading-challenge-login-title", value: "Log in to join the challenge", comment: "Title for alert that asks users to log in to join the reading challenge"),
+                message: WMFLocalizedString("reading-challenge-login-message", value: "Log in or create an account to track your progress towards the reading challenge.", comment: "Message for alert that asks users to log in to join the reading challenge"),
+                preferredStyle: .alert
+            )
+            alert.addAction(UIAlertAction(title: CommonStrings.joinLoginTitle, style: .default) { [weak self] _ in
+                guard let self else { return }
+                self.navigationController.presentedViewController?.dismiss(animated: true) {
+                    let loginCoordinator = LoginCoordinator(navigationController: self.navigationController, theme: self.theme, loggingCategory: .login)
+                    loginCoordinator.loginSuccessCompletion = {
+                        self.navigationController.tabBarController?.selectedIndex = 3
```
tonisevener, line 162: > On successful login or account creation, the user should be taken back to the Activity tab with the [Widget announcement](https://www.figma.com/design/nnXvcESYGNFiMmxYA41vVG/WP25--Apps-Reading-Challenge?node-id=382-2855&t=33Zrff8kZOyR1KaO-4) showing

I'm not seeing the followup widget announcement after you log in, I tried calling the same logic as the logged-in conditional above but it's still not working. Can you look into it?

```
loginCoordinator.loginSuccessCompletion = {
    self.enroll()
    self?.onEnroll?()
    self?.onDismiss?()
    }

```

For background, the widget announcement looks like this (taken from the logged-in announcement flow).

https://github.com/user-attachments/assets/8521744d-c518-41a7-b928-2db895972bfb

tonisevener (changes requested):

tonisevener (approved):
Going to try a little bit of refactoring of these announcements in my next PR, I plan to tackle the followup widget announcement issue here.

PR #5807, opened 2026-04-14
T418950 P0 Design Review

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T418950

### Notes
* ipad widget announcement rendering incorrectly  - updated fonts according to Sarah
* after joining challenge and adding widget, still stuck at "join challenge" variant - dev setting, updated so that join challenge updates the dev setting
* announcement not opening when tapping "join challenge" on widget - fixed this by removing the guard that prevented it from being reshown
* announcement not triggering immediately - activity tab survey that'll be gone

### Test Steps
1. Go through test with both dev settings and manually changing the dates
2. Ensure you can join challenge from widget
3. Ensure tapping join challenge from widget brings you to activity tab with the announcement showing
4. Ensure widget changes upon join

### Checklist
- [x] ipad widget announcement rendering incorrectly 
- [x] after joining challenge and adding widget, still stuck at "join challenge" variant
- [x] announcement not opening when tapping "join challenge" on widget
- [x] announcement not triggering immediately


Wikipedia/Code/ActivityTabViewController.swift
```diff
@@ -354,7 +357,7 @@ final class WMFActivityTabHostingController: WMFComponentHostingController<WMFAc
         guard isViewLoaded && view.window != nil else { return false }
         guard let dataStore else { return false }
         let isLoggedIn = dataStore.authenticationManager.authStateIsPermanent
-        guard await WMFActivityTabDataController.shared.shouldShowReadingChallengeAnnouncement(isLoggedIn: isLoggedIn) else { return false }
+        guard await WMFActivityTabDataController.shared.shouldShowReadingChallengeAnnouncement_IgnoreHasSeen(isLoggedIn: isLoggedIn) else { return false }
```
tonisevener, line 360: Seems like we _shouldn't_ ignore hasSeen here, right? I'm seeing the announcement on every appearance of Activity tab now (no dev settings are set). 

https://github.com/user-attachments/assets/3e555444-e66a-494d-8dde-293b547fd9b0

tonisevener (changes requested):

tonisevener (approved):
One possible bug (see comment) but I'm going to work on a followup that might help with this.

PR #5808, opened 2026-04-14
Widget ui cleanup

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T418955
https://phabricator.wikimedia.org/T418952

### Notes
* ipad large variant contents are being cut off across stages - fixed with smaller fonts + padding
* make sure referencing correct type tokens, and scaling down when device sizes are smaller (see @william's implementation) - fixed with smaller fonts
* Updated close icon

### Test Steps
1. 

### Checklist
- [ ] Checked against Swift 6 strict concurrency

PR #5809, opened 2026-04-15
Reading challenge cleanup

tonisevener:
**Phabricator:** N/A

Hopefully fixes some announcement issues seen in https://github.com/wikimedia/wikipedia-ios/pull/5808 and 
https://github.com/wikimedia/wikipedia-ios/pull/5805.

### Notes
Two major pieces to this:

1. I reworked the developer config to allow overriding the current date. This overrides the current date when a page view is logged, as well as when the reading widget states are calculated. This is a way for us to work around the network connection issues when setting our device date to the future. I also added a reset button that clears out all of the reading challenge user defaults and clears out the page views, so we don't have to reinstall as much to test announcements.

I did remove all of specific state toggles in developer settings, so we have to act as a user to get through each of the states. We can add them back in if we really need it, but I think it might be good for us to test as close to a real user as possible.

2. I combined both the full page reading challenge announcement and widget announcement into one coordinator. I also moved a lot of the gatekeeping logic like checking for the hasSeen User Defaults flag and the config dates into the coordinator. Since this flow is re-used across Explore, Article, and Activity, I think sharing more of this logic through the coordinator leaves less room for bugs from only one presentation context.

Note I could not test the collect prize flow (completed state widget no longer has the button, see screenshot). I think that widget button may have been removed in recent merges.

Screenshot 2026-04-15 at 12 49 49 AM

### Test Steps
1. Fresh install.
2. Go to developer settings, there is now a new toggle at the top called "enable developer mode". Enable it
3. Shortcut to developer settings now shows in Profile
4. Install widget. Confirm state looks right.
5. Go to developer settings, scroll to the bottom, change override date to May 1.
6. Confirm widget shows "Join challenge" state.
7. Tap it, confirm you are taken to the announcement modal. Confirm joining challenge changes widget state.
8. Visit article, confirm widget changes.
9. Go to dev settings, change date to May 2nd. Confirm widget state changes.
10. Tap Random widget button. Confirm deep link to article. Confirm widget state changes.
11. Repeat steps 9-10 a few times.

### Checklist


l-olson1214 (approved):
I'm okay with this but I wonder if we should discuss with design / make the toggles accessible so they can easily switch them, or do you think it's better for them to also go through as close to a real user as possible?

tonisevener:
> I'm okay with this but I wonder if we should discuss with design / make the toggles accessible so they can easily switch them, or do you think it's better for them to also go through as close to a real user as possible?

@l-olson1214 Agreed, will work on bringing those back!

l-olson1214 (approved):

PR #5821, opened 2026-04-21
Allow pinch to zoom when reading articles

folecr:
Overrides user-scalable=no in the article webview

**Phabricator:** 
https://phabricator.wikimedia.org/T424079

### Notes

Allow pinch to zoom in the article webview

I frequently use pinch to zoom on the web (both desktop and mobile) to zoom into article text or images without relayout of the entire article webpage but the mobile app does not allow usage of this gesture. This PR allows using pinch-to-zoom in the mobile app exclusively in the article webview.

Since iOS 10+ Safari allows pinch to zoom even with `user-scalable=no` in the HTML. WKWebView however respects `user-scalable=no` and does not allow zooming by default. This change sets webview's configuration to `ignoresViewportScaleLimits` to match behavior of Safari and allow pinch to zoom. Note W3C recommendation https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/meta/name/viewport#user-scalable.

A broader, perhaps more correct fix, is to set `user-scalable=yes` in the content being served. It looks like `user-scalable` in the HTML content being served is being set by https://github.com/wikimedia/mediawiki-services-mobileapps/blob/master/lib/transformations/pcs/head.js#L71 introduced in change [b3bed0a05861401f83d84e93d436d08bb77da1c4](https://github.com/wikimedia/mediawiki-services-mobileapps/commit/b3bed0a05861401f83d84e93d436d08bb77da1c4)

If there is interest in enabling pinch to zoom, I could potentially make a change in that repo if there is consensus that it is the right spot, instead of this repo `wikipedia-ios`. Please see Note 2 below regarding Android.

* Note 1 : I did not find the current zoom level used in wikipedia-ios. There is an old commit https://github.com/wikimedia/wikipedia-ios/commit/a376a1030bcae896768e1565006eb23d7ca9c524 that uses the zoom level in `getElementRect()`. From a quick search it looks like `getElementRect()` itself is not used in wikipedia-ios. So, this change should not have any impact other than allowing the user to zoom in with gestures.
* Note 2 : Associated Android PR. On Android the webview cannot be configured to ignore viewport scale limits so the corresponding change on Android will require modifying HTML content before loading in webview. The Android PR will depend on the decision whether to make the change in PCS to serve pages with `user-scalable=yes`. If the served pages cannot be changed, then for the Android app, the loaded content will have to override user-scalable before parsing the HTML in webview.

### Test Steps

Test iOS wikipedia app with this change. Pinch to zoom in article webviews matches behavior in Safari.

### DIFF
--- a/Wikipedia/Code/ArticleViewController.swift
+++ b/Wikipedia/Code/ArticleViewController.swift
@@ -226,6 +226,7 @@ class ArticleViewController: ThemeableViewController, UIScrollViewDelegate, WMFN
     lazy var webViewConfiguration: WKWebViewConfiguration = {
         let configuration = WKWebViewConfiguration()
         configuration.setURLSchemeHandler(schemeHandler, forURLScheme: schemeHandler.scheme)
+        configuration.ignoresViewportScaleLimits = true
         return configuration
     }()
 
--- a/Wikipedia/Code/ArticleViewController.swift
+++ b/Wikipedia/Code/ArticleViewController.swift
@@ -224,6 +224,7 @@ class ArticleViewController: ThemeableViewController, UIScrollViewDelegate, WMFN
     lazy var webViewConfiguration: WKWebViewConfiguration = {
         let configuration = WKWebViewConfiguration()
         configuration.setURLSchemeHandler(schemeHandler, forURLScheme: schemeHandler.scheme)
+        configuration.ignoresViewportScaleLimits = true
         return configuration
     }()
 
--- a/Wikipedia/Code/ArticleViewController.swift
+++ b/Wikipedia/Code/ArticleViewController.swift
@@ -226,7 +226,11 @@ class ArticleViewController: ThemeableViewController, UIScrollViewDelegate, WMFN
     lazy var webViewConfiguration: WKWebViewConfiguration = {
         let configuration = WKWebViewConfiguration()
         configuration.setURLSchemeHandler(schemeHandler, forURLScheme: schemeHandler.scheme)
-        configuration.ignoresViewportScaleLimits = true
+
+        if WMFDeveloperSettingsDataController.shared.allowGestureZoomArticleWebview {
+            configuration.ignoresViewportScaleLimits = true
+        }
+
         return configuration
     }()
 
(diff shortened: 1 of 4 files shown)


Wikipedia/Code/ArticleViewController.swift
```diff
@@ -224,6 +224,7 @@ class ArticleViewController: ThemeableViewController, UIScrollViewDelegate, WMFN
     lazy var webViewConfiguration: WKWebViewConfiguration = {
         let configuration = WKWebViewConfiguration()
         configuration.setURLSchemeHandler(schemeHandler, forURLScheme: schemeHandler.scheme)
+        configuration.ignoresViewportScaleLimits = true
```
tonisevener: Hi @folecr! Thanks for the update. Our team would like to gate this behind a developer settings flag and test for a while before committing to this change in behavior for everyone. Can you wrap this up in a developer settings flag?

To get to our developer settings menu, go to app Settings > About > Tap the top "About" title 7 times, then our secret menu will appear :)

You can add a toggle in this form. You can follow the pattern of the other toggles (such as "Force email auth") to see how they are added to the dev settings view and update their underlying data in WMFDeveloperSettingsDataController.

And then ideally this line will be wrapped up in a conditional like this:

```
if WMFDeveloperSettingsDataController.shared.enablePinchToZoomArticle {
    configuration.ignoresViewportScaleLimits = true
}
```

Let us know if you have any issues, thanks!

tonisevener (changes requested):

folecr:
@tonisevener Thanks for the review. I've updated the PR with a developer settings toggle. Please let me know if there are further changes you suggest. Thanks.

tonisevener (approved):
Thanks @folecr! This looks good, can you fix the merge conflicts?

folecr:
updated and fixed the merge conflicts

folecr:
@tonisevener how do I request the workflows to run? thanks

tonisevener:
> @tonisevener how do I request the workflows to run? thanks

We have to approve them to run on our side, thanks!

PR #5829, opened 2026-04-24
Track app open

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T423721

### Notes
* Adds app_open action, tracks foregrounding, deeplink and widget activations
* The task was a little vague, but I added a comment for Shay. We can verify the code-side of things for now

### Test Steps
1. Change logging to DDLogLevelAll
2. Run the app, verify foreground action source
3. Open a deeplink, verify deeplink action source
4. Verify widget action source - formatting I adopted is `widget_name_of_widget` 

### Checklist


Widgets/Widgets/PictureOfTheDayWidget.swift
```diff
@@ -37,9 +37,9 @@ final class PictureOfTheDayData {
 
     // MARK: Public
 
-    static func fetchPictureOfTheDayEntryData(usingCache: Bool = false, maxWidth: Int = 960, completion: @escaping (PictureOfTheDayEntry) -> Void) {
```
tonisevener, line 40: @mazevedofs Are these width changes intentional or was it a POTD merge conflict? It seems unrelated to instrumentation.

tonisevener (commented):
Will test tomorrow but a few things jumped out at me via code.

tonisevener:
A few things from testing!

1. Upon fresh install and launch, the server post for this fails, I think because the app languages are missing. Maybe the data source or languages aren't set up at the time of event generation. It seems to be a one-time thing though, subsequent launches work fine.

2. Top read article widget tap logs as `deepLink` instead of `widget_top_read`. If I tap just outside of an article in the widget (which only opens the app and does not send straight to a particular article), it does log as `widget_top_read`. Seems like it should probably be `widget_top_read` for widget article taps too.

3. Same goes for on this day widget ^ - article taps log as `deepLink`.

Widgets/Widgets/PictureOfTheDayWidget.swift
mazevedofs: This was a conflict, i'll clean it up!

mazevedofs:
@tonisevener Shay updated the task description with changes to the events sent, I'll update the PR it later today while I wait and I'll report back my progress

tonisevener (approved):

PR #5840, opened 2026-04-28
T269478: Use GHA artifacts instead for dynamic baseline

pvzig:
**Phabricator: T269478** 

### Notes
* Alternative codecov baseline approach. Instead of committing a `coverage-baseline.json` that requires committing to main exceptions, we'll use GHA artifacts

### Test Steps
N/A

### Checklist

### Screenshots/Videos
N/A


tonisevener (approved):
This is cool, thank you!

PR #5842, opened 2026-04-30
Fix Alemannisch incorrectly labeled as Albanian in Locale lookup

HakanIST:
## Summary

The iOS `Locale` framework follows ISO 639-3, where `als` is assigned to Albanian (Tosk). MediaWiki uses `als` for Alemannisch (Alemannic German), whose standard ISO code is `gsw`. This causes the app to display "Albanian" instead of the correct localized name ("Swiss German" / "Alemannic German").

## Changes

- **`wikipedia-languages.json`**: Add `altISOCode: "gsw"` and `localeOverrideCode: "gsw"` to the `als` entry
- **`Wikipedia.swift`**: Add optional `localeOverrideCode` field to the `Wikipedia` struct
- **`WikipediaLookup.swift`**: Use `localeOverrideCode` (when present) for `Locale` lookups instead of the MW language code
- **`WikipediaLanguageCommandLineUtilityAPI.swift`**: Add `als` guard block and pass `localeOverrideCode` in all `Wikipedia` struct initializations
- **`MWKLanguageLinkControllerTests.m`**: Add unit tests for `gsw→als` ISO mapping and a regression test verifying Alemannisch is not labeled as Albanian

## Design Decisions

`localeOverrideCode` is kept **separate from `altISOCode`** to avoid changing Locale behavior for the existing `no` entry (`altISOCode: "nb"`), which was intentionally designed for language link matching only (see #3886). Reusing `altISOCode` for Locale lookup would change Norwegian's display name from "Norwegian" to "Norwegian Bokmål" — an unintended side effect.

## Testing

- Xcode build: ✅ `BUILD SUCCEEDED`
- Swift typecheck: ✅
- JSON decode validation (358 languages): ✅
- Locale lookup: `als` → "Swiss German" ✅ (was "Albanian")
- No regression for `no` → still "Norwegian" ✅

## Related

- Fixes https://phabricator.wikimedia.org/T398296
- Also fixes https://phabricator.wikimedia.org/T414588 (the `altISOCode` provides the missing `gsw→als` mapping for `MWKLanguageLinkController`)


HakanIST:
Hi @tonisevener @l-olson1214, I submitted a fix for T398296 and T414588. It's a small change with unit tests included. Would appreciate a review when you have time. Thank you!

tonisevener (approved):
@HakanIST Sorry about the delay on this review - it's working well for me, thanks! Also thanks for taking care to avoid regressions in Norwegian.

PR #5850, opened 2026-05-04
Add new UITestConfiguration and clean up some old test config

pvzig:
**Phabricator: T269478** 

### Notes
- Language and theme are now defined only in the test plan. Added a linting step to the UITest GHA to prevent regressions on this.
- Added a new `UITestConfiguration` to be the home for any and all launch test configuration.
- Cleaned up scattered uses of `#if UITEST` preprocessor.

### Test Steps
N/A

### Checklist
- [X] Checked against Swift 6 strict concurrency

### Screenshots/Videos
N/A


WikipediaUITests/AppOnboardingUITests.swift
```diff
@@ -1,32 +1,19 @@
+import Foundation
 import XCTest
 
 final class AppOnboardingUITests: XCTestCase {
 
-    func testOnboardingLightMode() throws {
+    func testFirstLaunchShowsOnboardingSmoke() throws {
         let app = XCUIApplication()
-        XCUIDevice.shared.appearance = .light
```
tonisevener, line 7: Just wanted to note that App Onboarding is going to be an unusual case from the rest of the app UI tests because the "WMFAppThemeName" User Default is nil upon first launch when onboarding is displayed, so it basically falls back to picking either the [light or black theme ](https://github.com/wikimedia/wikipedia-ios/blob/811d2ad139a3cf9bc46807f2b1a2756fb630143a/Wikipedia/Code/NSUserDefaults%2BWMFExtensions.swift#L154) based on the system mode.

I suspect that's why the screenshots can seem a little funky for App Onboarding. Here I am forcing a Dark / Black theme here through the test plan, on a light-mode device, and it looks like the text is following the system theme instead of the forced theme. I can put up a PR fixing it - I think this theming bug would only show in UI tests but wanted to let you know what's happening if you found yourself chasing it down. 

Image

tonisevener (approved):
Looks good and feels cleaner, thanks! Just one note that's not a blocker.

WikipediaUITests/AppOnboardingUITests.swift
pvzig: Ah that makes sense and is a detail I missed, thanks. I'll see what our options are both in terms of testing that behavior and opting certain tests out of all themes.

PR #5851, opened 2026-05-05
T418486: Add OTD DataController

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T418486

### Notes
* Languages? Copilot caught a difference but I'm not sure

### Test Steps
1. Ensure tests pass / checks pass
2. Look through files, check for any oddities 

### Checklist


WMFData/Tests/WMFDataTests/WMFOnThisDayDataControllerTests.swift
```diff
@@ -0,0 +1,261 @@
...
+    // MARK: - fetchOnThisDay – happy path
+
+    func testFetchOnThisDayParsesEventsCount() throws {
+        let controller = try makeController(mockJSONResource: "onthisday-events-02-21-get")
+        let expectation = expectation(description: "fetchOnThisDay completes")
+
+        controller.fetchOnThisDay(project: .wikipedia(.init(languageCode: "en", languageVariantCode: nil)), month: 2, day: 21) { result in
+            switch result {
+            case .success(let response):
+                XCTAssertEqual(response.events.count, 3)
```
tonisevener: I think tests would run faster if you checked the whole response at the same time, instead of several separate tests here. So something like:

```
case .success(let response):
    XCTAssertEqual(response.events.count, 3)
    XCTAssertEqual(response.events[0].text, "Wikipedia, a free wiki content encyclopedia, goes online.")
    XCTAssertEqual(response.events[0].year, 2001)
/// etc.
```

WMFData/Sources/WMFData/Data Controllers/Shared/WMFOnThisDayDataController.swift
```diff
@@ -50,17 +40,17 @@ public final class WMFOnThisDayDataController {
         completion: @escaping (Result<WMFOnThisDayResponse, Error>) -> Void
     ) {
         guard isSupported(project: project) else {
-            completion(.failure(WMFOnThisDayDataControllerError.unsupportedProject))
+            completion(.failure(WMFDataControllerError.basicServiceUnavailable))
```
tonisevener: We should maybe keep this one `WMFOnThisDayDataControllerError.unsupportedProject`, it matches what is actually happening here.

WMFData/Tests/WMFDataTests/WMFOnThisDayDataControllerTests.swift
```diff
@@ -11,251 +11,28 @@ final class WMFOnThisDayDataControllerTests: XCTestCase {
...
-            switch result {
-            case .success:
-                XCTFail("Expected unsupportedProject error, got success")
-            case .failure(let error):
-                XCTAssertEqual(error as? WMFOnThisDayDataControllerError, .unsupportedProject)
-            }
-            expectation.fulfill()
-        }
-
-        waitForExpectations(timeout: 2)
```
tonisevener: You should still keep the unsupported projects test, those are great to have! Just make sure they are reworked to use `async`.

WMFData/Tests/WMFDataTests/WMFOnThisDayDataControllerTests.swift
```diff
@@ -11,251 +11,28 @@ final class WMFOnThisDayDataControllerTests: XCTestCase {
...
+    func testFetchOnThisDayParsesEventsCount() async throws {
```
tonisevener, line 14: (Optional) Maybe rename to `testFetchOnThisDayParsesEvents` since you're checking more than just the count?

tonisevener (approved):

PR #5852, opened 2026-05-05
Convert WMFAppViewController to Swift

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T423289

### Notes
This PR ports our Objective-C WMFAppViewController class into Swift. The first commit leans heavily on Copilot and the remainder were manual.

I'm deliberately not doing any additional cleanup to keep risks low and code review as easy as possible. Further cleanup can happen at a later stage.

### Test Steps
1. Light regression testing for app launch & deep links.

### Checklist


Wikipedia/Code/WMFAppViewController.swift
```diff
@@ -0,0 +1,2091 @@
...
+
+    private var _savedArticlesFetcher: SavedArticlesFetcher?
+    
+    private var unprocessedUserActivity: NSUserActivity?
+    private var unprocessedShortcutItem: UIApplicationShortcutItem?
+
+    private var backgroundTasks: [String: UIBackgroundTaskIdentifier] = [:]
+    private let backgroundTasksLock = NSLock()
+    
+    private var isWaitingToResumeApp: Bool = false
```
l-olson1214, line 73: Since these are all independent of each other, could we turn them into a state machine?

Wikipedia/Code/WMFAppViewController.swift
```diff
@@ -0,0 +1,2091 @@
...
+
+    private var _settingsNavigationController: WMFComponentNavigationController?
+
+    var readingListsAlertController: ReadingListsAlertController!
+    
+    var syncStartDate: Date?
+    
+    var savedTabBarItemProgressBadgeManager: SavedTabBarItemProgressBadgeManager?
+    
+    private var hasSyncErrorBeenShownThisSesssion: Bool = false
```
l-olson1214: Extra S

Wikipedia/Code/WMFAppViewController.swift
```diff
@@ -0,0 +1,2091 @@
+import UIKit
```
l-olson1214, line 1: Overall this is a pretty chunky file, I wonder if we could run it through Claude/something and have it break it down into smaller coordinators or something?

Wikipedia/Code/WMFAppViewController.swift
```diff
@@ -0,0 +1,2091 @@
...
+        SessionsFunnel.shared.appDidBecomeActive()
+        checkRemoteAppConfigIfNecessary()
+        updatePrimaryWikiHasTempAccountsStatusIfNecessary()
+        periodicWorkerController?.start()
+        savedArticlesFetcher?.start()
+        assignMoreDynamicTabsV2ExperimentIfNeeded()
+        AppIconUtility.shared.checkAndRevertIfExpired()
+    }
+
+    @objc func performTasksThatShouldOccurAfterAnnouncementsUpdated() {
```
l-olson1214: does this need objc?

Wikipedia/Code/WMFAppViewController.swift
tonisevener: @l-olson1214 That will be covered in https://phabricator.wikimedia.org/T423290!

l-olson1214 (approved):

Wikipedia/Code/WMFAppViewController.swift
tonisevener: @l-olson1214 I agree we should refactor / clean up further, but my goal here is to just do a direct port of the Objective-C file, which is the only way I could keep the LOE down to a minimum. Refactoring into state machine / coordinators introduces a risk of regressions which we won't have time to fix.

Wikipedia/Code/WMFAppViewController.swift
tonisevener: Fixed!

Wikipedia/Code/WMFAppViewController.swift
tonisevener: Fixed!

PR #5856, opened 2026-05-06
Use system color for article toolbar buttons

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T424238

### Notes
For liquid glass toolbar buttons that could display over images (mainly Article and Explore), we should lean on system colors  for the icons.  This is so that, when the system decides to switch the button mode, the icon color switches to match.

### Test Steps
1. Go to article "Apple silicon". I think this is the article from the Phab screenshot.
2. Confirm that as you scroll, article bottom toolbar and top nav buttons are readable when it scrolls over the image.

### Checklist

### Screenshots/Videos

Old (note profile button and bottom toolbar icons):
https://github.com/user-attachments/assets/e4623310-7f0c-420b-9725-dc566d45d763

New:
https://github.com/user-attachments/assets/b456c2db-f82e-401b-b091-19162bdfb75a


l-olson1214 (changes requested):
Failing tests

tonisevener:
@l-olson1214 Tests are fixed now!

l-olson1214 (approved):

PR #5859, opened 2026-05-07
Onboarding UI Tests

pvzig:
**Phabricator: T269478** 

### Notes
* Adds stable accessibility ID structure to be shared between UI tests and app code
* Adds general helpers for UI test execution along with onboarding-specific helpers
* Adds onboarding UI tests 

### Test Steps
N/A

### Checklist

### Screenshots/Videos
N/A

### DIFF
--- a/Wikipedia/Code/AccessibilityIdentifiers.swift
+++ b/Wikipedia/Code/AccessibilityIdentifiers.swift
@@ -1,0 +1,82 @@
+import Foundation
+
+/// Shared accessibility identifiers used by app code and UI tests.
+///
+/// These strings are an external UI-test contract. Keep them centralized here so app code
+/// and tests do not drift when identifiers are renamed or new elements are added.
+enum AccessibilityIdentifiers {
+    enum Explore {
+        static let view = "Explore View"
+    }
+
+    enum Profile {
+        static let button = "profile-button"
+        static let view = "Profile View"
+    }
+
+    enum Onboarding {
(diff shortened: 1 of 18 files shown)


Wikipedia/Code/AccessibilityIdentifiers.swift
```diff
@@ -0,0 +1,83 @@
+import Foundation
+
+/// Shared accessibility identifiers used by app code and UI tests.
+///
+/// These strings are an external UI-test contract. Keep them centralized here so app code
+/// and tests do not drift when identifiers are renamed or new elements are added.
+enum AccessibilityIdentifiers {
```
tonisevener, line 7: The placement of this file will make it inaccessible to UI elements in WMFComponents. So, for example, trying to change this line [here](https://github.com/wikimedia/wikipedia-ios/blob/43efa0f4c0e65abab6b333037ebc05b6036b789f/WMFComponents/Sources/WMFComponents/Components/Profile/WMFProfileHostingController.swift#L17) won't work:

```
view.accessibilityIdentifier = AccessibilityIdentifiers.Profile.view
```

So maybe you could move this into WMFComponents instead, then any legacy UI app-side can just import WMFComponents if it needs to set an accessibility identifier.

Not a blocker to this merge since it might be a while before you're testing WMFComponents, but wanted to give you a heads up.

tonisevener (commented):
Great to see this stuff! One heads up I wanted to give you, but I can merge if you want to tackle it later.

tonisevener (approved):

PR #5882, opened 2026-05-18
T269478: E2E/UI Test Split

pvzig:
**Phabricator: T269478** 

### Notes
* Add dedicated E2E test lane

### Test Steps
N/A

### Checklist

### Screenshots/Videos
N/A


.github/workflows/run_e2e_ui_tests.yml
```diff
@@ -0,0 +1,97 @@
+name: Run E2E Tests
+
+on:
+  pull_request:
+    branches:
+      - 'main'
```
mazevedofs, line 6: Is running the E2E tests on every PR the intended behavior?
Every change pushed to an open PR will trigger this test suite, which might be too much. I would suggested we keep it only the push to main config below. Let me know what you think.

mazevedofs (changes requested):

.github/workflows/run_e2e_ui_tests.yml
pvzig: I think a limited E2E test suite run against PRs to help prevent major regressions from landing in `main` is a good idea and something I'd recommend.

mazevedofs (approved):

PR #5885, opened 2026-05-19
T269478: UI test mocking infra

pvzig:
**Phabricator: T269478** 

### Notes
  - Adds a `SessionHTTPClient` transport boundary underneath `Session`. `Session` continues to own request
    construction, cache fallback behavior, cookies, and delegate callbacks while the underlying HTTP client can be
    swapped for UI tests.
  - Adds a fixture-backed UI-test HTTP client selected through `-WMFUITestHTTPClientProfile` `fixture-strict`.
  - Some bundled fixtures

### Test Steps
N/A

### Checklist

### Screenshots/Videos
N/A


tonisevener (approved):
Looks good @pvzig! I do have a question for how we might lean on this to test calls that originate out of WMFData (WMFBasicService). Will DM you about this.

PR #5922, opened 2026-06-03
Remove games feature flag

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T427923

### Notes
* Removed the feature flag checks
* Renamed the flag to V2 so we can reuse it
* Kept the games cache cleaning

### Test Steps
1. If on simulator, fresh install
2. Have the "Show games v2" flag off
3. Make sure card shows and you can play
4. Reset the game cache
5. Background and foreground app to see the announcement on explore feed 
6. Reset the game cache
7. Open an article and check you see the announcement

### Checklist


l-olson1214 (changes requested):
The text itself still says "v1" - just for clarity can we change it?
Screenshot 2026-06-03 at 10 45 37 AM

l-olson1214 (approved):

PR #5946, opened 2026-06-08
Stabilize unit tests: remove timing waits and silent-pass guards

davmonk:
## Summary

Two test-only reliability fixes to the existing unit suites. No production code changes.

### `WMFDonateViewModelTests` — remove wall-clock waits
Three tests selected a value, then waited via `DispatchQueue.main.asyncAfter(deadline: .now() + 1.0)` + `wait(for:timeout: 2.0)` before asserting. The `WMFDonateViewModel` button/textfield/transaction-fee propagation chain is **fully synchronous** (plain Combine `.sink` with `.dropFirst()` — no `debounce`, `throttle`, `Task`, or `DispatchQueue`), so the values are observable immediately after the mutation. The waits added ~3s of dead time per run and coupled the tests to wall-clock timing (a flake risk on loaded CI). They now assert synchronously and each run in ~0.001s (down from ~1s+).

### `WMFArticleDataControllerTests` — surface request-construction failures
Both tests built their `ArticleInfoRequest` with `try?` + `guard … else { return }`, which **silently passes** the test if construction ever throws (no assertion runs). Changed the two tests to `throws` and use `try`, so a construction failure is reported as a real failure. This matches the pattern already used in `WMFYearInReviewDataControllerCreateOrRetrieveTests`.

## Testing
- `xcodebuild test -scheme WMFData` (iPhone 16, iOS 18.5) → 122 tests pass, 0 failures.
- `xcodebuild test -scheme WMFComponents` (iPhone 16, iOS 18.5) → 206 tests pass, 0 failures; the 3 rewritten Donate tests run in 0.001s each.
- SwiftLint `--strict` on both changed files: 0 violations.

## Notes
- No associated Phabricator ticket — this is self-initiated test cleanup/stabilization.
- A follow-up could add a shared `XCTestCase` base that restores `WMFDataEnvironment.current` in `tearDown` (no test currently resets the global singleton); deferred here to keep this PR narrow.

### DIFF
--- a/WMFComponents/Tests/WMFComponentsTests/WMFDonateViewModelTests.swift
+++ b/WMFComponents/Tests/WMFComponentsTests/WMFDonateViewModelTests.swift
@@ -168,15 +168,10 @@ final class WMFDonateViewModelTests: XCTestCase {
         
         // Select amount button
         firstButtonVM.isSelected = true
-        
-        let expectation = XCTestExpectation(description: "Waiting for textfield amount to update")
-        
-        DispatchQueue.main.asyncAfter(deadline: .now() + 1.0) {
-            XCTAssertEqual(viewModel.textfieldViewModel.amount, firstButtonVM.amount)
-            expectation.fulfill()
-        }
-        
-        wait(for: [expectation], timeout: 2.0)
+
+        // The button-selection -> textfield update propagates synchronously through Combine,
+        // so the new amount is observable immediately without waiting.
+        XCTAssertEqual(viewModel.textfieldViewModel.amount, firstButtonVM.amount)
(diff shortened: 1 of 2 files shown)


davmonk:
Closing as superseded. The stabilization work this PR set out to do — removing the timing-based waits and silent-pass guards in WMFDonateViewModelTests and WMFArticleDataControllerTests — was independently landed on main by #5952 and #5955 (pvz/update-async-wait-tests), which additionally converted both suites to Swift Testing async/await. Merging main into this branch and resolving to main's version leaves the branch byte-identical to main, so there is nothing left to merge.

PR #5948, opened 2026-06-08
Add unit tests for Year in Review slide data controllers

davmonk:
## Summary

Adds unit coverage for the nine Year in Review slide data controllers, none of which were previously tested. They are near-identical implementations, so their per-slide constants and population rules are the behavior most at risk from copy-paste drift. Test-only — no production code changes.

`YearInReviewSlideDataControllersTests` covers, across all nine controllers:
- **`id` bindings** — each controller reports the expected `WMFYearInReviewPersonalizedSlideID` raw value (guards against a copy-pasted slide keeping the wrong id).
- **`containsPersonalizedNetworkData`** — pins which slides hold personalized network data and are therefore cleared on logout (save/view/edit count) vs. those that aren't.
- **`shouldFreeze`** — pins the freeze flag per slide.
- **`shouldPopulate`** — config-only slides populate when the config is active; the **view count** slide additionally requires a `userID`; the **edit count** slide additionally requires a `username`.

Tests reuse the existing `WMFFeatureConfigResponse.Common.YearInReview.testConfig` helper, which is active for the current date, so the `isActive` checks are deterministic.

## Testing
- `xcodebuild test -scheme WMFData` (iPhone 16, iOS 18.5) → 129 tests pass, 0 failures (7 new).
- SwiftLint `--strict` on the new file: 0 violations.

## Notes
- No associated Phabricator ticket — self-initiated coverage work.
- The `populateSlideData(in:)` / `makeCDSlide(in:)` data paths (which require a Core Data context and delegate fixtures) are not covered here; this PR focuses on the identity and population-decision contracts. Those data paths are a reasonable follow-up.


tonisevener:
Hi @davmonk, looks like there's a WMFData test failure on this branch. Can you take a look?

`WMFData/Tests/WMFDataTests/YearInReviewSlideDataControllersTests.swift mutates WMFDataEnvironment.current but does not use WMFDataTestFixture setup`

davmonk:
Thanks @tonisevener — good catch. The branch had picked up the new `lint-wmfdata-test-fixtures.sh` guard from main, but `YearInReviewSlideDataControllersTests` was still mutating `WMFDataEnvironment.current` with the old raw pattern.

Fixed in 3ff6a47: adopted `WMFDataTestFixture` (import `WMFDataTestSupport`, wrap `setUp` with `fixture.setUp()` + `resetWMFDataTestState()`, add a `tearDown` calling `fixture.tearDown()`), matching `WMFArticleTabsDataControllerTests`. Verified locally: the guard script now passes, all 7 tests pass (serial, iOS 18.5), and SwiftLint `--strict` is clean.

tonisevener (approved):

PR #5956, opened 2026-06-12
T413380: WMFFeedDataController

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T413380

### Notes
* Uses WMFFeedImageNew -> to be renamed once we deprecate current WMFFeedImage
* Grabs all potential data, may need to trim down in the future

### Test Steps
1. Run unit tests, ensure all pass

### Checklist


WMFData/Sources/WMFData/Data Controllers/Explore/WMFFeedDataController.swift
```diff
@@ -0,0 +1,59 @@
+import Foundation
+
+public final class WMFFeedDataController {
```
mazevedofs: Quick pre-review: make it an actor

WMFData/Sources/WMFData/Models/Home/WMFFeedResponse.swift
```diff
@@ -0,0 +1,232 @@
...
+    private enum PageKeys: String, CodingKey {
+        case page, revisions, edit, talk
+    }
+
+    public init(from decoder: Decoder) throws {
+        let top = try decoder.container(keyedBy: TopLevelKeys.self)
+
+        if top.contains(.desktop) {
+            let desktop = try top.nestedContainer(keyedBy: PageKeys.self, forKey: .desktop)
+            desktopPage = try desktop.decodeIfPresent(String.self, forKey: .page)
```
tonisevener: This flattening introduces brittleness - if for some reason this `try` fails (Say if the url is not a String type in the response), it would propagate up and the entire WMFFeedResponse would fail. I think let's simplify and keep the original `json` structure of separate objects, I don't think there's a need to custom flatten them at this point.

```
public struct WMFFeedContentURLs: Codable {
    public let desktop: WMFFeedContentURLGroup?
    public let mobile: WMFFeedContentURLGroup?
}

public struct WMFFeedContentURLGroup: Codable {
    public let page: String?
    public let revisions: String?
    public let edit: String?
    public let talk: String?
}
```

Then you can delete this `public init(from decoder: Decoder) throws {` and `public func encode(to encoder: Encoder) throws {` entirely.

WMFData/Sources/WMFDataMocks/Resources/feed-featured-2025-12-11-get.json
```diff
@@ -0,0 +1,187 @@
+{
```
tonisevener, line 1: Can you pull a more recent feed response? Looks like this is missing `news` which would be good to test.

https://en.wikipedia.org/api/rest_v1/feed/featured/2026/06/15 returns a `news` object.

tonisevener (changes requested):
Thanks! Just a few things.

tonisevener (approved):

PR #5960, opened 2026-06-16
Add Home Data Controller

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T413367

### Notes
Created a wrapper data controller + tests around WMFFeedDataController to call from the new Home tab. Calling it WMFHomeDataController since that is the name for the new feature, but we can switch naming back to Explore if folks feel strongly about it.

### Test Steps
1. No way to call from UI at this point, but ensure code + unit tests look correct.

### Checklist


WMFData/Sources/WMFData/Data Controllers/Home/WMFHomeDataController.swift
```diff
@@ -0,0 +1,67 @@
+import Foundation
+
+public final actor WMFHomeDataController {
+
+    private let feedDataController: any WMFFeedDataControlling
+
+    // Dates for which feed data has been fetched, in descending order (most recent first).
+    private var fetchedDates: [Date] = []
```
mazevedofs: Is there a risk we get savedDates for one language, and if the user switches languages, the date ref will be from another language, since it's a singleton?
Can we do `fetchedDates: [String: [Date]] = ["en":[date1, date2...]`?

WMFData/Sources/WMFData/Data Controllers/Home/WMFHomeDataController.swift
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        let response = try await feedDataController.fetchFeed(project: project, date: previousDate)
+        recordFetchedDate(previousDate)
+        return response
+    }
+
+    // MARK: - Private
+
+    private func recordFetchedDate(_ date: Date) {
```
mazevedofs: Re: previous comment -> this could be `recordFetchedDate(for language: String, _ date: Date) ` or something like that

mazevedofs (changes requested):

WMFData/Sources/WMFData/Data Controllers/Home/WMFHomeDataController.swift
tonisevener: @mazevedofs Great catch! I pushed a fix.

mazevedofs (approved):

PR #5964, opened 2026-06-16
Add unit tests for WMFImageDataController

davmonk:
Adds unit tests for the shared `WMFImageDataController`, taking the file from **0% → 100% line coverage**.

## What
10 tests covering every branch of both public methods, using a small purpose-built `WMFService` mock (no network):

- **`fetchImageData`** — success; in-memory cache hit (service hit only once on repeat requests for the same URL); independent caching per URL; `basicServiceUnavailable` error; service-failure propagation; and that failures are not cached.
- **`fetchImageInfo`** — success decoding into `WMFImageInfo`; `unexpectedResponse` on empty pages; `mediaWikiServiceUnavailable` error; empty-title `failureCreatingRequestURL`; and service-failure propagation.

## Notes
- Services are injected directly via `init` rather than mutating the global `WMFDataEnvironment`, so the tests hold no shared state and don't contribute to the existing parallel-run global-environment flakiness.
- Verified locally: full WMFData suite green across multiple runs, SwiftLint `--strict` clean.


mazevedofs (approved):
Thanks for your contribution!

PR #5968, opened 2026-06-16
Expand WMFPageViewsDataController test coverage

davmonk:
Adds tests for the previously-untested methods of `WMFPageViewsDataController` (against a temporary `WMFCoreDataStore`), raising the file to **~74% (416/559 lines)** coverage.

> **Updated (synced with main):** main removed the reading-challenge feature (`fetchReadingChallengeState`, `ReadingChallengeState`, and the `devReadingChallengeState`/`readingChallengeUserCompleted` user-defaults keys), so the reading-challenge tests this PR originally included were dropped — you can't test deleted code. The PageViews coverage below is unaffected.

## What
- `fetchMostRecentTime` — latest timestamp, and `nil` on an empty store
- `fetchTimelinePages` — descending timestamp order
- `addPageViewSeconds` + `fetchPageViewMinutes` — seconds summed into minutes
- `fetchPageViewDates` — bucketing by day, hour, and month

## Notes
- Date-bucketing assertions use local-calendar dates built the same way the code reads them, so they're stable across time zones.

## Known gap: `fetchLinkedPageViews` (investigated)
`fetchLinkedPageViews` is intentionally left uncovered. Investigation (full notes in [this comment](https://github.com/wikimedia/wikipedia-ios/pull/5968#issuecomment-4723273169)) showed the crash is **not** in the relationship walk: the method returns `[[CDPageView]]` — raw Core Data `NSManagedObject`s — and the crash occurs when the caller reads a property (`pageView.page?.title`) **off the context's queue**, which trips `-com.apple.CoreData.ConcurrencyDebug 1` (enabled in the WMFData test plan). The walk itself terminates and returns correctly; a control that only inspects array structure passes.

The method is `public` with **zero callers**, and is the only method on this controller returning managed objects instead of value types (cf. `fetchTimelinePages → [WMFPageWithTimestamp]`). It still exists on `main` as of this update. Recommended fix: remove it (dead code) or extract value types inside the `perform` block, after which the linked-chain test can be added back. Worth a Phabricator ticket (GitHub issues are disabled on this repo).

Verified locally: 10/10 tests pass across serial runs, SwiftLint `--strict` clean.


davmonk:
### Investigation notes: why `fetchLinkedPageViews()` is left uncovered

(GitHub issues are disabled on this repo, so recording the investigation here next to the in-code `NOTE`. Worth a Phabricator ticket.)

**Finding:** `WMFPageViewsDataController.fetchLinkedPageViews()` returns `[[CDPageView]]` — raw Core Data `NSManagedObject`s — out of its `context.perform` block. Any consumer that reads a property of those objects (e.g. `pageView.page?.title`) off the owning context's queue trips Core Data's multithreading assertion and **crashes**. The WMFData test plan runs with `-com.apple.CoreData.ConcurrencyDebug 1`, so this is a hard crash in tests.

**Evidence:**
- A test that builds a `Start → Middle → End` chain, calls `fetchLinkedPageViews()`, then reads `chains.first?.compactMap { $0.page?.title }` crashes the runner. Instrumented logging shows the method itself returns successfully (1 chain, length 3) — the crash is in the *property access afterward*, not the fetch/walk.
- A control that calls the same method but only inspects array structure (`chains.map { $0.count }`, never touching a managed-object property) passes. This isolates the cause to off-queue managed-object access; the `previousPageView`/`nextPageViews` inverse relationships are modeled correctly and the walk terminates on well-formed data.

**Also notable:** the method is `public` but has **zero callers** in the codebase, and is the only method on this controller that returns `NSManagedObject`s instead of value types (cf. `fetchTimelinePages()` → `[WMFPageWithTimestamp]`). That both violates the data-layer guidance (no managed objects across the boundary) and makes it a latent crash for any future caller.

**Recommended fix:** either remove it (currently dead), or refactor it to extract value types *inside* the `perform` block (e.g. `[[WMFPage]]` or a small struct). Once it returns Sendable value types, the linked-chain unit test can be added back.

davmonk:
Heads-up on a change from syncing this branch with `main`: main removed the reading-challenge feature entirely (`fetchReadingChallengeState`, the `ReadingChallengeState` enum, and the `devReadingChallengeState`/`readingChallengeUserCompleted` user-defaults keys — all gone from production source). The 8 reading-challenge tests this PR originally added no longer compile against that, so I removed them along with their `setUp` defaults-clearing.

The **10 remaining `WMFPageViewsDataController` tests** (add/delete/import/counts, `fetchMostRecentTime`, `fetchTimelinePages`, `fetchPageViewMinutes`, `fetchPageViewDates`) are unaffected and pass — verified locally (serial, iOS 18.5), SwiftLint `--strict` clean. File coverage is now ~74% (416/559 lines); the earlier ~87% figure included the removed tests.

Note the `fetchLinkedPageViews` finding above still stands — that method was **not** part of the removal and still exists on main. Description updated to match.

mazevedofs (approved):

PR #5979, opened 2026-06-19
Fix talk page links with apostrophes being truncated

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T429687

### Notes
* The href-parsing regex used a negative lookahead ([>"']) that treated a literal apostrophe as the end of the attribute value. Parsoid emits apostrophes literally in talk page link hrefs (e.g. href="./Ocean's_8"), so links like "New Year's Eve" were truncated to "./New_Year", sending users to the wrong article or failing to load.
Replace the regex with a quote-aware pattern that captures the value up to the matching delimiter (double-quoted, single-quoted, or unquoted), preserving apostrophes inside double-quoted values. Apply the same fix to the legacy ObjC parser and add regression tests.

### Test Steps
1. Go to a talk page that contains a link to a page whose title contains an apostrophe
2. Tap the link

### Checklist


Wikipedia/Code/NSString+WMFHTMLParsing.m
tonisevener: FYI the project still compiles if I delete this method entirely. Is there anything I might be missing?

tonisevener (commented):
Wondering if we might be able to delete the Objective-C method.

Wikipedia/Code/NSString+WMFHTMLParsing.m
mazevedofs: I was a bit afraid of deleting it, will do!

mazevedofs: done!

tonisevener (approved):

PR #5982, opened 2026-06-22
Restore translucent status bar background on iOS 26

pushpaktiwarii:
## Summary

Restore a themed translucent background behind the status bar on the iOS 26 code path.

## Problem

The top safe area overlay becomes fully transparent on iOS 26, causing content to appear directly beneath the status bar items while scrolling.

## Fix

Replace the fully transparent overlay with a themed translucent background to improve visual separation while preserving theme support.

Bug: T423637


mazevedofs:
Hi @pushpaktiwarii, thanks for your contribution.
Running your branch, the overlay only shows when the user is scrolling up, not when scrolling down in the article view.


https://github.com/user-attachments/assets/c96f4092-a49a-418f-9ff8-8b8c8f1170c1

pushpaktiwarii:
@mazevedofs 
Thanks for testing and for the video! I see what you mean. I'll investigate why the overlay is only appearing while scrolling up and update the patch so that it also appears correctly while scrolling down in the article view.

pushpaktiwarii:
@mazevedofs  Thanks for the feedback! I investigated the issue and pushed an updated patch. The overlay height is now recalculated in viewSafeAreaInsetsDidChange() instead of relying only on the initial setup, ensuring it uses the correct status bar height after the view is attached to the window hierarchy and as the safe area changes. Could you please take another look when you have a chance?

mazevedofs:
Hi @pushpaktiwarii, the latest changes added an opaque background to both the tab bar and navigation bar
The idea is to add the soft gradient we can see in the Apple News screenshot in the ticket https://phabricator.wikimedia.org/T423637
Current:
IMG_3026
IMG_3024

Expected:

image

pushpaktiwarii:
@mazevedofs thanks for the feedback!

I've just pushed a new commit to fix this. I replaced the opaque backgrounds with a soft gradient for both the UITabBar and UINavigationBar (fading into .clear), which perfectly mimics the Apple News style you mentioned.

Could you please review the latest changes? Let me know if it looks good now!

WMFComponents/Sources/WMFComponents/Base/Component Base Classes/WMFComponentNavigationController.swift
```diff
@@ -127,8 +127,10 @@ open class WMFComponentNavigationController: UINavigationController {
     private func applySystemGlassAppearance(customLargeTitleFont: UIFont?) {
         let appearance = UINavigationBarAppearance()
 
-        appearance.configureWithDefaultBackground()
+        appearance.configureWithTransparentBackground()
```
tonisevener: These changes are very wide-reaching, affecting every navigation bar in the app. Ideally we should leave this alone and only make targeted changes to ArticleViewController, since that is the only location with a gradient issue at the top. Liquid Glass naturally seems to add a correct gradient to navigation bars that do not hide upon scroll (most of the app), but it doesn't work on Article because ArticleViewController sets this [hidesNavigationBarOnScroll](https://github.com/wikimedia/wikipedia-ios/blob/f152e47a67ec67cc48efb46869d3ab65b607549a/Wikipedia/Code/ArticleViewController.swift#L907) property. It seems that when the navigation bar hides upon scroll, the top gradient disappears with it.

I am good with updating `topSafeAreaOverlayView`, that is the correct view to add a gradient to. But it should only be updated on ArticleViewController. Can you revert the other changes in this PR?

Wikipedia/Code/WMFSettingsViewController.m
```diff
@@ -100,6 +100,11 @@ - (void)viewWillAppear:(BOOL)animated {
     [self configureNavigationBarFromObjC];
 }
 
+- (void)viewSafeAreaInsetsDidChange {
```
tonisevener: Can you merged the latest from `main` into this PR? This WMFSettingsViewController.m file has since been deleted.

tonisevener (changes requested):
Hi @pushpaktiwarii! I took a pass and would like to adjust so that these changes are a little less risky. Let me know if you have any questions, thanks!

pushpaktiwarii:
Hi @tonisevener ! Thanks for the review. I have updated the PR accordingly:

1. Merged `main`: Synced the latest changes from `main` into this branch.
2. Reverted global changes: Reverted the wide-reaching changes in `WMFComponentNavigationController`, tab bar appearance, and other view controllers back to their default system background.
3. Targeted fix in `ArticleViewController`: Scoped the `topSafeAreaOverlayView` styling strictly to `ArticleViewController` so top content remains covered when the navigation bar hides on scroll.

Let me know if everything looks good now!

mazevedofs:
@pushpaktiwarii could you resolve the conflicts in the branch? It's preventing us from triggering tests, thanks!

pushpaktiwarii:
@mazevedofs  I've resolved the branch conflicts and pushed an updated patch based on the latest main. The changes are now limited to ArticleViewController. Tests should be able to run now.

PR #5984, opened 2026-06-22
Home settings

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T427927

### Notes
* Adds the settings, persists them on User Defaults
* Separates the home settings from the settings coordinator

### Test Steps
1. Turn on the feature flag
2. Make sure you see the settings match the ticket
3. Make sure the toggle positions persist between launches
4. Make sure the for you tab button works 

### Checklist


Wikipedia/Code/HomeFeedSettingsCoordinator.swift
```diff
@@ -0,0 +1,171 @@
...
+        }, didTapLanguages: { [weak self] in
+            self?.showLanguages()
+        })
+        return WMFHomeFeedWhatsDrivingSettingsViewController(viewModel: viewModel, closeButtonHandler: closeButtonHandler)
+    }
+
+    // MARK: - Sub-flows
+
+    private func showCommunityModulesSettings() {
+        let viewModel = WMFHomeFeedCommunitySettingsViewModel(
```
tonisevener: (Optional) I envision the view model as being free to access the underlying data controller directly, rather than have the coordinator manage that connection. The coordinator manages navigation, but I don't think it should have to manage communication with the data controller. Let me know what you think!

Wikipedia/Code/SettingsCoordinator.swift
```diff
@@ -589,8 +590,9 @@ final class SettingsCoordinator: Coordinator, SettingsCoordinatorDelegate {
             return
         }
 
-        let homeFeedSettingsVC = WMFHomeFeedSettingsViewController(title: CommonStrings.homeFeedTitle)
-        settingsNav.pushViewController(homeFeedSettingsVC, animated: true)
+        let coordinator = HomeFeedSettingsCoordinator(navigationController: settingsNav, theme: theme)
+        self.homeFeedSettingsCoordinator = coordinator
```
tonisevener, line 594: Should we nil this out at some point?

WMFComponents/Sources/WMFComponents/Components/Settings/Home/Main/WMFHomeFeedSettingsViewModel.swift
```diff
@@ -0,0 +1,30 @@
...
+
+        let communityHeader = WMFLocalizedString("home-feed-settings-community-section-title", value: "Community", comment: "Section header for Community settings in Home feed settings.")
+        let forYouHeader = WMFLocalizedString("home-feed-settings-for-you-section-title", value: "For you", comment: "Section header for For You settings in Home feed settings.")
+
+        self.sections = [
+            SettingsSection(header: communityHeader, footer: nil, items: [
+                SettingsItem(image: nil, color: nil, title: modulesTitle, subtitle: communityModulesSubtitle, accessory: .chevron(label: nil), action: didTapCommunityModules)
+            ]),
+            SettingsSection(header: forYouHeader, footer: nil, items: [
+                SettingsItem(image: nil, color: nil, title: modulesTitle, subtitle: communityModulesSubtitle, accessory: .chevron(label: nil), action: didTapForYouModules),
```
tonisevener: Need a different subtitle here - "Turn on or off 'For you' modules.

WMFComponents/Sources/WMFComponents/Components/Home/WMFHomeView.swift
tonisevener: Very minor - can you swap these two around? In designs "Community" displays as the leading tab.

Wikipedia/Code/SettingsCoordinator.swift
mazevedofs: Not really, that would deallocate this coordinator prematurely. Other than settings coordinator itself, there's no other strong reference to this coordinator, so i think we're okay. 
That's the same pattern we use in Profile Coordinator for example:

https://github.com/wikimedia/wikipedia-ios/blob/0f4b257748008c8453cd063b0366e97d0dfea604/Wikipedia/Code/ProfileCoordinator.swift#L189-L199

WMFComponents/Sources/WMFComponents/Components/Settings/Home/Community/WMFHomeFeedCommunitySettingsViewModel.swift
```diff
@@ -0,0 +1,124 @@
...
+            case .featuredArticle: self.homeDataController.setCommunityFeaturedArticleIsOn(isOn)
+            case .topRead: self.homeDataController.setCommunityTopReadIsOn(isOn)
+            case .inTheNews: self.homeDataController.setCommunityInTheNewsIsOn(isOn)
+            case .onThisDay: self.homeDataController.setCommunityOnThisDayIsOn(isOn)
+            case .pictureOfTheDay: self.homeDataController.setCommunityPictureOfTheDayIsOn(isOn)
+            }
+        }
+    }
+
+    var sections: [SettingsSection] {
```
tonisevener, line 48: I noticed the toggle animation wasn't as smooth, and digging in, the reason is because this section property is computed. Toggling calls this computed property, which creates new section models with new IDs, which causes the SwiftUI list to draw out again.

Instead creating a stored sections property, then assigning these upon init restores the animation.

```
public private(set) var sections: [SettingsSection] = []
...
public init(homeDataController: WMFHomeDataController = .shared) {
   ...
   self.sections = buildSections()
   ...
}
...

private func buildSections() -> [SettingsSection] {
    let featuredArticle = SettingsItem(
    ...
    
}
```

WMFComponents/Sources/WMFComponents/Components/Settings/Home/For You/WMFHomeFeedForYouSettingsViewModel.swift
```diff
@@ -0,0 +1,106 @@
...
+            guard let self else { return }
+            switch module {
+            case .basedOnYourInterests: self.homeDataController.setForYouBasedOnInterestsIsOn(isOn)
+            case .becauseYouRead: self.homeDataController.setForYouBecauseYouReadIsOn(isOn)
+            case .continueReading: self.homeDataController.setForYouContinueReadingIsOn(isOn)
+            }
+        }
+    }
+
+    var sections: [SettingsSection] {
```
tonisevener, line 42: See [earlier comment](https://github.com/wikimedia/wikipedia-ios/pull/5984/changes#r3468532804) to restore toggling animation in here.

tonisevener (approved):
Approved. Did notice something that could be fixed now or later when we wrap up these views.

WMFComponents/Sources/WMFComponents/Components/Settings/Home/Community/WMFHomeFeedCommunitySettingsViewModel.swift
mazevedofs: I can make a follow-up PR for this so we can merge this one, I think the final dsign will be very similar so it's worth fixing it now vs later

PR #6002, opened 2026-07-06
T427936: Support hCaptcha in DiscussionTools API edits

kostajh:
**Phabricator:** https://phabricator.wikimedia.org/T427936

Why:

- Talk-page replies and new topics posted via discussiontoolsedit should support hCaptcha.

What:

- Detect the challenge in TalkPageFetcher, surfacing the server suggested sitekey and always-challenge flag
- present the existing WMFHCaptchaViewController (adding siteKeyOverride so the token is solved against the key the server verifies)
- resubmit the edit with the token as `captchaword`, plus wgConfirmEditForceShowCaptcha when an interactive always-challenge is required

Assisted-by: Claude Opus 4.8

### DIFF
--- a/Wikipedia/Code/TalkPageFetcher.swift
+++ b/Wikipedia/Code/TalkPageFetcher.swift
@@ -186,13 +186,13 @@ class TalkPageFetcher: Fetcher {
         }
     }
     
-    func postReply(talkPageTitle: String, siteURL: URL, commentId: String, comment: String, completion: @escaping (Result<Void, Error>) -> Void) {
+    func postReply(talkPageTitle: String, siteURL: URL, commentId: String, comment: String, hCaptchaToken: String? = nil, forceShowCaptcha: Bool = false, completion: @escaping (Result<Void, Error>) -> Void) {
         guard let title = talkPageTitle.denormalizedPageTitle else {
             completion(.failure(RequestError.invalidParameters))
             return
         }
-        
-        let params = ["action": "discussiontoolsedit",
+
+        var params = ["action": "discussiontoolsedit",
                       "paction": "addcomment",
                       "page": title,
                       "matags": WMFEditTag.appTalkReply.rawValue,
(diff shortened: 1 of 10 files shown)


Wikipedia/Code/TalkPageViewController.swift
```diff
@@ -1224,25 +1238,63 @@ extension TalkPageViewController: TalkPageReplyComposeDelegate {
...
+    /// Presents an hCaptcha challenge, reporting the token via `onSuccess` (to retry) or `onError`.
+    fileprivate func presentHCaptchaChallenge(on presenter: UIViewController, siteKey: String?, forceShowCaptcha: Bool, onSuccess: @escaping (String) -> Void, onError: @escaping () -> Void) {
+        let hcaptchaVC = WMFHCaptchaViewController()
+        hcaptchaVC.theme = theme
+        if forceShowCaptcha {
+            hcaptchaVC.siteKey = siteKey
+        } else {
+            let editApiKey = WMFDeveloperSettingsDataController.shared.loadFeatureConfig()?.ios.hCaptcha?.editApiKey
+            hcaptchaVC.siteKey = editApiKey?.isEmpty == false ? editApiKey : nil
+        }
```
ealbizzati, line 1250: if the non force path (else) if the remote config doesn't have an editApiKey yet, siteKey stays nil. WMFHCaptchaViewController falls back to config.apiKey. A token solved against that key would fail (that's my assumption) verification, and since failure comes back as another captcha challenge, the user would endup in a loop with no error message. 

Could that happen?

Wikipedia/Code/TalkPageViewController.swift
```diff
@@ -1224,25 +1238,63 @@ extension TalkPageViewController: TalkPageReplyComposeDelegate {
...
+                onSuccess(token)
+            }
+        }
+
+        hcaptchaVC.errorAction = { [weak hcaptchaVC] error in
+            guard let hcaptchaVC else { return }
+            hcaptchaVC.dismiss(animated: true) {
+                WMFToastManager.sharedInstance.showErrorAlert(error, sticky: true, dismissPreviousToasts: true)
+                onError()
+            }
```
ealbizzati, line 1266: should we avoid showing an error if the user is dismissing the captcha voluntarily? I mean, it isn't an error from the user's perspective, right?

Wikipedia/Code/TalkPageFetcher.swift
```diff
@@ -186,68 +186,98 @@ class TalkPageFetcher: Fetcher {
...
+        guard let hCaptchaToken else {
+            return
+        }
+        params["captchaword"] = hCaptchaToken
+        if forceShowCaptcha {
+            params["wgConfirmEditForceShowCaptcha"] = "1"
+        }
+    }
+
+    func evaluateResponse(_ error: Error?, _ result: [String : Any]?, completion: @escaping (Result<Void, Error>) -> Void) {
```
ealbizzati, line 247: heads up that my pr #6062 is also touching this. it propagates the newrevid in the completion

ealbizzati (commented):
Hi @kostajh left you a couple questions here. let me know what you think.

PR #6004, opened 2026-07-08
Integrate hCaptcha into the Login flow.

dbrant:
Mirror the account-creation hCaptcha wiring for login: when a login attempt fails with a captcha-related message code and the server offers an hCaptcha challenge (via authmanagerinfo), present the hCaptcha overlay, then resubmit the login with the resulting token in the captchaWord field.

https://phabricator.wikimedia.org/T430799


mazevedofs (approved):
Looks good, thanks for doing this! Ill get a build out so Anthony can test more extensively

PR #6008, opened 2026-07-12
T364154 Fix: New feature announcement overflows on iPad for Image Recs

pathaderavi:
**Phabricator:** 

#### Issue

- Long texts in the Popover `WMFFeatureAnnouncementView` in languages like Portuguese causes issue as shown in screenshot as the CTA is partially cut.

image

### Cause:

1. Earlier, the HostingViewController was fixed to a specific height, which SwiftUI View was trying to accommodate toe establish a height with
2. The hardcoded backgroundImage height calculator was relying on available height
3. Since SwiftUI view is wrapped in GeometryReader which has nested imageViews, the heights don't resolve so CTA can appear uncut

### Solution
1. Use PreferenceKey to report the height back to the HostingViewController to define preferred content size
2. Instead of calculating dynamically height, maintain it to 1.5 times to imageView height we have hardcoded so it has appropriate margined background 

### Test Steps
1. Adde hardcoded strings as shown in https://phabricator.wikimedia.org/T364154, and hardcoded bar button item to left to check the issue
5. Add solution to test the fix with long texts

### Solution

1. Check the height using PreferenceKey so that adapted height can be reported to HostingViewController
3. Use this reported height in preferredContentSize

Simulator Screenshot - iPad Air 11-inch (M4) - 2026-07-12 at 14 28 32

### Checklist


WMFComponents/Sources/WMFComponents/Components/Shared/Feature Announcement/WMFFeatureAnnouncementView.swift
```diff
@@ -1,51 +1,75 @@
...
+        static let contentHorizontalPadding: CGFloat = 22
+        static let containerHorizontalPadding: CGFloat = 10
+        static let contentTopPadding: CGFloat = 10
+
+        // Foreground (icon) image
+        static let foregroundImageWidth: CGFloat = 132
+        static let foregroundImageHeight: CGFloat = 118
+
+        // Background image (height is derived from the foreground icon height)
+        static let backgroundImageHeightRatio: CGFloat = 1.5
```
pathaderavi, line 24: This is the most important part to get agreement on. For me 1.5 multiplier seemed appropriate for background image to foreground image's height. But we can change it we want more height on top/bottom for background image

tonisevener (approved):
Thank you for this fix @pathaderavi! It's working well for me. We don't have any active features currently using this component but we will make any design tweaks (if needed) once we use it again.

PR #6010, opened 2026-07-14
T368100: Add impression counter to event instrumentation for iOS Image Recs

pathaderavi:
Task: Add impression counter to event instrumentation for iOS Image Recs

**Phabricator:** : https://phabricator.wikimedia.org/T368100

### Notes
* The current analytics events set doesn't have on appear for image recommendation 
* Added a new event and called whenever it appears, also ensured that it only gets called once 

### Test Steps
1. Hard code appearance of cell with `Add an Image` cell
2. Check the logging event 

### Checklist

### Screenshots/Videos

image


mazevedofs (approved):
Thanks for your contribution!

PR #6013, opened 2026-07-16
T284872: Observe update notifications for all articles

gemedet:
Bug: T284872

### Notes
* This PR configures the notification observer of `WMFArticleUpdated` to allow all senders, so that variants and redirects stay up to date.

### Test Steps
1. Open an article through a redirect link (e.g., "Relative_humidity")
2. Save or unsave the article

### Checklist

### Video
https://github.com/user-attachments/assets/750efa82-dc6d-45b1-851e-78f9151b1c22


Wikipedia/Code/ArticleViewController.swift
```diff
@@ -1581,6 +1582,7 @@ private extension ArticleViewController {
     func setupToolbar() {
         toolbarController = ArticleToolbarController(delegate: self)
         toolbarController?.apply(theme: theme)
+        toolbarController?.setToolbarButtons(enabled: false)
```
mazevedofs, line 1585: Hi @gemedet! 
The toolbar is disabled here and only enabled again inside the `articleLoadWaitGroup` completion block (`setupArticleLoadWaitGroup`), which runs only on a successful load.

The failure path, `handleArticleLoadFailure`, never leaves the group and never re-enables the toolbar. If the article fails to load, the entire toolbar stays disabled.
The vurrent behavior is to have the toolbar always enabled, so this is introducing a possible regression.

Was there a specific need in this bugfix to add the toolbar disabling?

mazevedofs (changes requested):

Wikipedia/Code/ArticleViewController.swift
gemedet: Hi @mazevedofs!

In my testing, `handleArticleLoadFailure()` calls `updateRefreshOverlay(visible: false)` which calls `setToolbarButtons(enabled: true)`, let me know if that's not expected.

My goal was just to avoid the situation where you can save the redirect url before the article finishes loading (see video). If that's out of scope, I can def revert it.

https://github.com/user-attachments/assets/db1df8b5-6975-47da-81b6-1a3a993d41b6

mazevedofs: I think you're right! It does get updated on a failure path. In that case, it looks good.

mazevedofs (approved):

PR #6014, opened 2026-07-17
Explore feed in community tab

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T432471

### Notes

- Embed the legacy Explore feed inside the Home tab's Community segment (phase 1 interim content).
- Add home-phase-2 developer flag + shared isCommunityFeedMode gate; Home tab rebuilds live when it toggles.
- Hide the language picker on the Community segment while the embedded feed is showing.
- Settings (phase 1): show two rows — For You (new home settings) + Community (legacy feed settings).
- Settings (phase 2): collapse back to a single Home feed row; flags off keeps the original Explore feed row.
- Gate the new home settings' Community section behind home-phase-2.
- Remove the "Explore tab" master toggle so the feed can't be turned off in community mode.
- Show an empty state with a "Customize the feed" button when all cards are hidden.
- Rename "Explore feed" → "Community feed" across user-facing strings, gated to community mode with new keys (originals kept for baseline).
- Remove the duplicate "What's driving your feed" link row from the For You modules screen (and drop SettingsItem's unused .link style).

### Test Steps

1. Turn on dev settings for Enable home tab
2. First tab is Home; Community segment shows the feed, no language button. Switch to For You -> language button appears; back to Community -> language button is gone.
3. Open a card's three-dot menu and the feed settings — the wording says Community, not Explore.
4. In feed settings, turn every card off → no "turn off" alert; feed shows an empty state with a "Customize the feed" button. Re-enable a card → feed returns.
5. Settings lists For You + Community rows.
6. Switch theme light/dark → feed re-themes correctly.
7. Turn on the enable home feed and enable phase 2 toggles
8. Community shows the new native feed; Settings collapses to one Home feed row (with a Community section inside).
9. Turn both toggles off. Explore tab as before; Settings shows the single Explore feed row; all wording says Explore.

### Checklist


WMFComponents/Sources/WMFComponents/Components/Settings/WMFSettingsView.swift
```diff
@@ -22,8 +22,7 @@ struct SettingsRow: View {
             }
             VStack(alignment: .leading, spacing: 2) {
                 Text(item.title)
-                    .font(Font(WMFFont.for(item.titleStyle == .link ? .boldSubheadline : .body)))
-                    .foregroundColor(item.titleStyle == .link ? Color(uiColor: theme.link) : nil)
```
l-olson1214, line 26: Did we want to remove foreground color here?

l-olson1214 (approved):

WMFComponents/Sources/WMFComponents/Components/Settings/WMFSettingsView.swift
mazevedofs: Yes, we removed a style that was created especially for the "what's driving you feed", but Sarah removed it. So this foreground color is not needed.

PR #6025, opened 2026-07-22
T430692: Fix widget images exceeding WidgetKit memory and archival limits

ealbizzati:
**Phabricator:** 
https://phabricator.wikimedia.org/T430692

### Notes
After fixing the picture of the day, I noticed the same issue on featured articles whenever the image was portrait with too large content. Same approach as the POTD fix — and while validating with all widgets installed, a few more oversized images showed up in the same shared process, so this PR bounds all of them:

- featured article: tfa thumbnail was decoded at full size with UIImage(data:) in the view; portrait thumbs exceed WidgetKit's archival size cap and the background image gets silently dropped. Now decoded with UIImage.downsampled at entry creation, sized to where it's displayed.
- top read: same archival cap issue at systemSmall (330x495 thumbs). Decoded at display size (240px).
- bundled assets: search widget globe (2244x2048, rendered at 70pt) resized to 240px; POTD sample fallback resized to 800px.
- POTD: cache fast path never matched (resized URL wasn't recorded), so every invocation re-fetched. Also small family now requests 500px instead of 960px.
- debug assert: scoped the sqlite check to the shared app container (it tripped on CFNetwork's own db).

Measured with os_proc_available_memory(): full reload batch with six widgets keeps ~20MB headroom; retained memory between batches dropped from ~10MB to ~1MB.

### Test Steps
1. On a day with a portrait featured article image (like 2026/07/22), add the featured article widget large — it won't load on main, loads on this branch.
2. Top read small: thumbnails show (on main they're dropped, with Widget archival failed in the console).

### Checklist

### Screenshots/Videos

fixed issue:
Screenshot 2026-07-22 at 12 40 46 PM


micgro42:
(note that this has the wrong Phabricator task linked: https://phabricator.wikimedia.org/T430697#12148888)

tonisevener (approved):
Works well, thanks!

PR #6029, opened 2026-07-23
T427673: For You and T432169: Dynamic Color

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T427673
https://phabricator.wikimedia.org/T432169

### Notes
* 

### Test Steps
1. With the dev settings for new home feed on
2. Go through onboarding, select interests
3. Navigate to articles for more than 10s
4. Navigate to For You
5. Ensure that up to the first 3 vertical scrolls are randomly selected interests, and you're able to swipe horizontally to see similar content
6. Then if you have any because you read
7. Then if you had any continue reading
8. Then if you had any more interests 
9. For testing with special images selected by Sarah: go to dev settings and toggle ON color test, then refresh For You

### Checklist

### Card Layout
- [x] Each card is full-screen immersive with only the bottom navigation bar persisting below card content
- [x] Wikipedia wordmark displayed at top left of every card
- [x] Tab switcher (Community / For You) accessible at the top of every card
- [x] Language switcher dropdown accessible at the top of every card
  - [x] Lists all user-configured languages
  - [x] Selecting a language refreshes the feed to that language
  - [x] 'Manage languages' option at the bottom navigates to language settings
- [x] Profile button accessible at the top of every card with badge for unread notifications
- [x] Tabs accessible at the top of every card
- [x] Three-dot more menu on each card with the following options:
  - [x] Save/Remove from Saved — saves article to reading list
  - [x] Share — opens share dialog
  - [x] Hide module — turns off the entire module in Explore Feed settings
  - [x] Hide this card — removes article suggestion and remembers state locally indefinitely
  - [x] Customize Interests — navigates to For You module settings

### Card Variants
- [x] All variants display: reason label, article title, card position indicator, three-dot menu
- [x] **Variant 1 (Balanced)** — 1st and 4th card in stack: full image background + short text excerpt
- [x] **Variant 2 (Image-focused)** — 2nd card in stack: full image background + article description (omitted if none exists)
- [x] **Variant 3 (Text-focused)** — 3rd card in stack: no image + text extract
- [x] 4th card uses random color variation as background per Figma
- [x] Fallback: if no image available, background color shown with modified card design per [Figma](https://www.figma.com/design/VwBaV9KZ7LWfOY2dt1BJyQ/explore-feed-explorations?node-id=4795-94506&t=ELodzU7oCq3WFcRb-4)
- [x] Tapping a card navigates to the full article

### Behavior
- [x] Vertical swipe (bottom-to-top) advances to next module; (top-to-bottom) goes back
- [x] Horizontal swipe (right-to-left) advances to next card in module; (left-to-right) returns to previous
- [x] Scrolling stops when user reaches end of feed
- [x] Feed automatically refreshes once per calendar day
- [x] No automatic re-refresh after daily refresh unless user manually triggers
- [x] Pull-to-refresh supported
  - [x] On refresh, history and interests are re-evaluated (cleared history not used)
- [x] Feed remembers user's position when returning from another tab
- [x] Clear error state shown when feed cannot load due to no internet connection

### Cross-Module Checks
- [x] No duplicate articles within the same carousel/module on the same day (e.g. Continue Reading & Saved Articles)
- [x] No duplicate articles across different modules on the same day (e.g. Because You Read & Because of Your Interest)
- [x] Articles hidden via "Hide this card" remain hidden indefinitely

### Module Order
- [x] Based on Your Interests — first 3 interests
- [x] Because You Read
- [x] Continue Reading
- [x] Based on Your Interests — remaining interests

### Screenshots/Videos
Screenshot 2026-07-23 at 11 25 04 AM

https://github.com/user-attachments/assets/5fc980ad-4d43-4bc9-bc3b-ca0432b345c7

https://github.com/user-attachments/assets/c731237b-7ac3-4bcd-a763-781d58574b8d

Screenshot 2026-07-23 at 11 32 08 AM

https://github.com/user-attachments/assets/d7ed82ea-d828-40a8-8a8c-44ba5264dbd3

Simulator Screenshot - iPhone 17 Pro Max liquid glass - 2026-07-23 at 15 56 30


mazevedofs:
@l-olson1214 could you fix the conflicts so the test suite can run?

WMFData/Sources/WMFData/Data Controllers/Developer Settings/WMFDeveloperSettingsDataController.swift
```diff
@@ -40,6 +41,11 @@ public protocol WMFDeveloperSettingsDataControlling: AnyObject {
 
     private var userDefaultsStore: WMFKeyValueStore? { WMFDataEnvironment.current.userDefaultsStore }
     
+    public var isUsingColorTestForYou: Bool {
```
mazevedofs: I don't think this dev setting is necessary; let's remove it

WMFComponents/Sources/WMFComponents/Components/Settings/Home/For You/Whats Driving/Your Interests/WMFHomeFeedInterestsSettingsViewController.swift
```diff
@@ -29,6 +33,13 @@ public final class WMFHomeFeedInterestsSettingsViewController: WMFComponentHosti
 
     private func configureNavigationBar() {
         let titleConfig = WMFNavigationBarTitleConfig(title: viewModel.title, customView: nil, alignment: .centerCompact)
-        configureNavigationBar(titleConfig: titleConfig, closeButtonConfig: nil, profileButtonConfig: nil, tabsButtonConfig: nil, searchBarConfig: nil, hideNavigationBarOnScroll: false)
+        let closeConfig = closeButtonHandler.map { _ in
```
mazevedofs, line 39: Selecting interests after onboarding doesn't load them into the feed.

If you skip selecting interests during onboarding and do that through the empty view, they never reload the view in spite of being persisted. That works currently on the stub UI + onboarding.

Plus, now we get this black bar behind the community feed.

https://github.com/user-attachments/assets/13c3681a-f5c0-4ff9-ad6d-41ec31ebc28c

WMFData/Sources/WMFData/Data Controllers/Home/WMFHomeDataController.swift
```diff
@@ -478,3 +478,66 @@ public enum WMFHomeDataControllerError: LocalizedError {
         }
     }
 }
+
+extension WMFForYouResponse {
```
mazevedofs: What is this extension? 

If it's a test suite helper, it shouldn't live in the class. Move it to a test file.

WMFComponents/Sources/WMFComponents/Components/Home/WMFForYouViewModel.swift
```diff
@@ -24,34 +32,109 @@ public struct WMFForYouModuleVisibility {
...
+            let header = WMFForYouHeaderLabel(
+                prefix: WMFLocalizedString("for-you-header-interest-article-prefix", value: "Because of your interest: ", comment: "Prefix for a For You feed card header based on an article the user has shown interest in."),
+                boldSuffix: $0.pageInterest.title
+            )
+            return WMFForYouPageViewModel(module: .basedOnInterests, headerLabel: header, articles: deduplicated($0.articles))
         }
         let becauseYouReadPage: [WMFForYouPageViewModel] = response.becauseYouReadArticles.map {
-            [WMFForYouPageViewModel(module: .becauseYouRead, headerLabel: "Because you read: \($0.recentlyRead.title)", articles: $0.articles)]
+            let header = WMFForYouHeaderLabel(
+                symbolName: "clock",
```
mazevedofs: Use our component here

WMFComponents/Sources/WMFComponents/Components/Home/WMFForYouViewModel.swift
```diff
@@ -24,34 +32,109 @@ public struct WMFForYouModuleVisibility {
...
+
+    public let emptyTitle = WMFLocalizedString("for-you-empty-title", value: "Nothing here yet", comment: "Title shown on the For You tab when there is no content to display.")
+    public let emptySubtitle = WMFLocalizedString("for-you-empty-subtitle", value: "Add interests to get personalized article recommendations.", comment: "Subtitle shown on the For You tab empty state encouraging the user to add interests.")
+    public let emptyButtonTitle = WMFLocalizedString("for-you-empty-button", value: "Choose your interests", comment: "Button on the For You empty state that opens the interests customization screen.")
+
+    public init(
+        response: WMFForYouResponse,
+        moduleVisibility: WMFForYouModuleVisibility = WMFForYouModuleVisibility(basedOnInterests: true, becauseYouRead: true, continueReading: true),
+        hiddenCardKeys: Set<String> = []
+    ) {
```
mazevedofs, line 92: Any reason to have everything happening inside the init? Can't we break down this work?

WMFComponents/Sources/WMFComponents/Components/Home/WMFHomeView.swift
```diff
@@ -73,46 +92,84 @@ public struct WMFHomeView: View {
...
-                onCustomizeInterests: { viewModel.didTapCustomizeInterests?() }
-            )
+            WMFForYouView(viewModel: forYouViewModel)
+                .ignoresSafeArea()
         } else if viewModel.isLoadingForYou {
             Spacer()
             ProgressView()
             Spacer()
+        } else if viewModel.forYouFeedError != nil {
+            VStack(spacing: 16) {
```
mazevedofs: Can you create reusable component like WMFEmptyView?

WMFComponents/Sources/WMFComponents/Components/Home/WMFHomeViewModel.swift
```diff
@@ -82,18 +126,30 @@ public final class WMFHomeViewModel: ObservableObject {
...
-            hiddenCardKeys.append(card.hideKey)
+            hiddenCardKeys.insert(card.hideKey)
+            forYouViewModel?.hiddenCardKeys.insert(card.hideKey)
         }
     }
 
     public func refreshForYouFeed() async {
+        isUsingColorTestForYou = WMFDeveloperSettingsDataController.shared.isUsingColorTestForYou
+
+        if isUsingColorTestForYou {
```
mazevedofs: We can also remove this check

Wikipedia/Code/HomeViewController.swift
```diff
@@ -57,13 +64,103 @@ final class HomeViewController: UIViewController, WMFNavigationBarConfiguring, T
...
+        viewModel.didShareForYouCard = { [weak self] article in
+            self?.shareForYouArticle(article)
+        }
+        viewModel.didTapUnsaveForYouCard = { [weak self] article in
+            self?.unsaveForYouArticle(article)
+        }
+
+        tabObservation = viewModel.$selectedTab.sink { [weak self] tab in
+            DispatchQueue.main.async {
+                self?.updateNavigationBarAppearance(for: tab)
```
mazevedofs: Let's avoid using `DispatchQueue.main.async`; let's use newer concurrency APIs.

Also, `sink` publishers are not sendable, which might be an issue. We're dealing with all UI work here (VM, VC, navigation bar), but it's important to check if it never crosses to a non-UI boundary. That could result in a strict concurrency error.

WMFComponents/Sources/WMFComponents/Components/Home/WMFHomeView.swift
```diff
@@ -12,12 +12,18 @@ public struct WMFHomeView: View {
...
+
     public var body: some View {
         mainContent
-            .environment(\.colorScheme, theme.preferredColorScheme)
-            .task {
-                viewModel.loadCurrentTabFeedIfNeeded()
-            }
+            .task { viewModel.loadCurrentTabFeedIfNeeded() }
             .onChange(of: viewModel.selectedTab) { _ in
                 viewModel.loadCurrentTabFeedIfNeeded()
```
mazevedofs, line 48: This introduces a warning
`'onChange(of:perform:)' was deprecated in iOS 17.0: Use `onChange` with a two or zero parameter action closure instead.`

WMFComponents/Sources/WMFComponents/Components/Home/WMFForYouView.swift
```diff
@@ -73,72 +273,204 @@ private struct WMFForYouPageView: View {
...
+                        }
+
+                        Spacer().frame(height: 16)
+
+                        WMFForYouMiniCard(
+                            label: viewModel.miniCardLabel,
+                            title: viewModel.title,
+                            description: viewModel.description,
+                            uiImage: viewModel.uiImage,
+                            onMenu: onSaveCard
```
mazevedofs: This 3-dot menu saves the article directly. Should it open a menu like in the other types of cards?

WMFComponents/Sources/WMFComponents/Components/Home/WMFForYouView.swift
```diff
@@ -10,40 +33,121 @@ public struct WMFForYouView: View {
...
+        .scrollBounceBehavior(.basedOnSize)
+    }
+
+    // MARK: - Empty State
+
+    private var emptyState: some View {
+        VStack(spacing: 0) {
+            Spacer()
+
+            Image(systemName: "sparkles")
```
mazevedofs: Let's use our component

WMFComponents/Sources/WMFComponents/Components/Home/WMFForYouView.swift
```diff
@@ -10,40 +33,121 @@ public struct WMFForYouView: View {
...
+                .font(.system(size: 48, weight: .light))
```
mazevedofs: Let's use a WMFFont, if not in the project, you can add it

WMFComponents/Sources/WMFComponents/Components/Home/WMFForYouView.swift
```diff
@@ -57,14 +161,110 @@ private struct WMFForYouPageView: View {
...
+        return scene?.windows.first?.safeAreaInsets.bottom ?? 0
+    }
 
     var body: some View {
-        TabView {
-            ForEach(articleViewModels) { article in
-                WMFForYouArticleCardView(viewModel: article, theme: theme, onHideModule: onHideModule, onHideCard: { onHideCard(article) }, onCustomizeInterests: onCustomizeInterests)
+        ScrollView(.horizontal, showsIndicators: false) {
+            LazyHStack(spacing: 0) {
+                ForEach(Array(articleViewModels.enumerated()), id: \.offset) { index, article in
```
mazevedofs: Using `.offset` as the ID here might  be the root cause of some UI weirdness we see like [T431637](https://phabricator.wikimedia.org/T431637)

Identity is array position, not a unique article. When a card is hidden, the array shrinks, so every later card slides into a slot SwiftUI already considers "the same view" and just shifts its contents, and doesn't reload the UI. 
Also, variant is derived from index, so each later card's layout style and background color will be reassigned randomly and not based on the actual card type needed.

It's worth a shot using another type of identifier here (hideKey is an excellent choice) to check if it solves some UI update issues. You'll need to adjust currentPage to match hideKey so the dots keep working correctly.

WMFComponents/Sources/WMFComponents/Components/Home/WMFHomeView.swift
```diff
@@ -26,12 +32,14 @@ public struct WMFHomeView: View {
...
+            ZStack(alignment: .top) {
                 forYouTabContent
+                    .ignoresSafeArea()
+                headerBar(isForYou: true)
+                    .padding(.top, safeAreaTop + 52)
             }
-            .frame(maxWidth: .infinity, maxHeight: .infinity)
-            .background(Color(uiColor: theme.paperBackground))
+            .ignoresSafeArea()
+            .environment(\.colorScheme, .dark)
```
mazevedofs, line 64: This might be the cause of this issue. The removal of `.environment(\.colorScheme, theme.preferredColorScheme)` leaves the community tab inheriting the theme from the system (which, in this case, is dark mode).

This color scheme is applied to the whole VC, not just the For You tab.
Image

Wikipedia/Code/HomeViewController.swift
```diff
@@ -57,13 +64,103 @@ final class HomeViewController: UIViewController, WMFNavigationBarConfiguring, T
...
+
+        UISegmentedControl.appearance(whenContainedInInstancesOf: [WMFHomeHostingController.self]).backgroundColor = .clear
         reloadLanguages()
+        
+        viewModel.isArticleSaved = { [weak self] card in
+            guard let self,
+                  let siteURL = card.project.siteURL,
+                  var articleURL = siteURL.wmf_URL(withTitle: card.title) else { return false }
+            articleURL.wmf_languageVariantCode = card.project.languageVariantCode
+            return dataStore.savedPageList.isAnyVariantSaved(articleURL)
```
mazevedofs: This is only called in view did load, so the saving experience is inconsistent.

https://github.com/user-attachments/assets/fb7e7f76-2459-494c-9a86-6b77ebae0186

Wikipedia/Code/HomeViewController.swift
```diff
@@ -96,12 +193,11 @@ final class HomeViewController: UIViewController, WMFNavigationBarConfiguring, T
     }
 
     // MARK: - What's Driving (test deep-link)
-
-    // TODO: Temporary. Presents "What's driving your feed" modally to test the settings entry point. You can delete if you're working on implementing the feed.
+    
     private var homeFeedSettingsCoordinator: HomeFeedSettingsCoordinator?
     private func presentWhatsDrivingTest() {
```
mazevedofs: Since you're reusing this func, rename it and update the comment above.

WMFComponents/Sources/WMFComponents/Components/Home/WMFForYouViewModel.swift
```diff
@@ -79,36 +162,213 @@ public final class WMFForYouPageViewModel: ObservableObject, Identifiable {
...
+
+    public var hideCardTitle: String {
+        CommonStrings.hideCardTitle
+    }
+
+    public let hideModuleTitle = WMFLocalizedString("for-you-menu-hide-module", value: "Hide module", comment: "Menu action to hide the entire For You feed module that contains this card.")
+
+    public let customizeInterestsTitle = WMFLocalizedString("for-you-menu-customize-interests", value: "Customize interests", comment: "Menu action to open the interests customization screen from a For You feed card.")
+
+    public let miniCardLabel = WMFLocalizedString("for-you-mini-card-label", value: "Mini card", comment: "Small label displayed above the title on the mini card variant of the For You feed card.")
```
mazevedofs: This label doesn't exist on Figma. The card only has title and subtitle.

WMFComponents/Sources/WMFComponents/Components/Home/WMFForYouView.swift
```diff
@@ -73,72 +273,204 @@ private struct WMFForYouPageView: View {
...
+                // MARK: Variant 3: Text-focused (also used as fallback when no image)
+                case .textFocused:
+                    VStack(alignment: .leading, spacing: 0) {
+                        HStack(alignment: .top, spacing: 12) {
+                            Text(viewModel.extract ?? viewModel.title)
+                                .font(Font(WMFFont.for(.georgiaTitle1)))
+                                .foregroundStyle(.white)
+                                .lineLimit(8)
+                                .frame(maxWidth: .infinity, alignment: .leading)
+                            menuView
```
mazevedofs: According to Figma, this variant only has the 3-dot menu on the mini card

Image

WMFComponents/Sources/WMFComponents/Components/Home/WMFHomeView.swift
mazevedofs: Might also be the cause of this tab bar bug 

https://github.com/user-attachments/assets/8a8f7901-be1b-43a7-b066-d4e234cfd81f

WMFComponents/Sources/WMFComponents/Components/Home/WMFHomeViewModel.swift
```diff
@@ -108,22 +164,37 @@ public final class WMFHomeViewModel: ObservableObject {
...
+            forYouViewModel = WMFForYouViewModel(response: .colorTest)
+            refreshSavedStates()
+            isLoadingForYou = false
+            return
+        }
+
+        guard let language = selectedLanguage else {
+            isLoadingForYou = false
+            return
```
mazevedofs, line 217: Loading state after picking a new language has an incorrect frame; the loading animation is almost out of the window

https://github.com/user-attachments/assets/92bb2e44-7a68-4831-87e5-cd133d4bbf8e

WMFComponents/Sources/WMFComponents/Components/Home/WMFForYouView.swift
```diff
@@ -57,14 +161,110 @@ private struct WMFForYouPageView: View {
...
+                        .fill(index == (currentPage ?? 0) ? Color.white : Color.white.opacity(0.4))
+                        .frame(
+                            width: index == (currentPage ?? 0) ? 8 : 7,
+                            height: index == (currentPage ?? 0) ? 8 : 7
+                        )
+                        .animation(.spring(response: 0.3, dampingFraction: 0.7), value: currentPage)
+                }
             }
+            .padding(.vertical, 20)
+            .padding(.bottom, windowSafeAreaBottom + 49 + 12)
```
mazevedofs: Why 49 + 12?

WMFComponents/Sources/WMFComponents/Components/Home/WMFForYouView.swift
```diff
@@ -73,72 +273,204 @@ private struct WMFForYouPageView: View {
...
-                        .font(Font(WMFFont.for(.subheadline)))
-                        .foregroundStyle(.white.opacity(0.9))
-                        .lineLimit(3)
-                        .shadow(radius: 2)
-                }
-            }
-            .padding(.horizontal, 20)
-            .padding(.bottom, 44)
+    private var dotsAndTabBarHeight: CGFloat {
+        windowSafeAreaBottom + 49 + 12 + 20 + 8 + 20
```
mazevedofs: Why 49 + 12 + 20 + 8 + 20?

WMFComponents/Sources/WMFComponents/Components/Home/WMFForYouView.swift
```diff
@@ -10,40 +33,121 @@ public struct WMFForYouView: View {
...
-            ScrollView(.vertical, showsIndicators: false) {
-                LazyVStack(spacing: 0) {
-                    ForEach(viewModel.pages.filter { moduleVisibility.isVisible($0.module) }) { page in
-                        let visibleArticles = page.articleViewModels.filter { !hiddenCardKeys.contains($0.hideKey) }
-                        if !visibleArticles.isEmpty {
-                            WMFForYouPageView(articleViewModels: visibleArticles, theme: theme, onHideModule: { onHideModule(page.module) }, onHideCard: onHideCard, onCustomizeInterests: onCustomizeInterests)
-                                .frame(width: geometry.size.width, height: geometry.size.height)
-                        }
+        if visiblePages.isEmpty {
+            emptyState
```
mazevedofs, line 102: The visible pages logic only checks for modules, not individual articles/cards. Hiding all articles results in a blank screen, not an empty state. You should also check for:

` let visibleArticles = page.articleViewModels.filter { !viewModel.hiddenCardKeys.contains($0.hideKey) } ` 

Image

WMFComponents/Tests/WMFComponentsTests/WMFHomeViewModelTests.swift
```diff
@@ -119,10 +134,19 @@ final class WMFHomeViewModelTests: XCTestCase {
...
+        let cardVM = WMFForYouArticleCardViewModel(article: article, headerLabel: header)
         XCTAssertEqual(cardVM.hideKey, "for_you_\(project.id)_Octopus")
     }
 
+    func testHideForYouCardIdempotent() {
+        let (vm, _) = makeViewModel()
+        let cardVM = makeForYouCardViewModel()
+        vm.hideForYouCard(cardVM)
+        vm.hideForYouCard(cardVM)
+        XCTAssertEqual(vm.hiddenCardKeys.count, 1)
```
mazevedofs: hideForYouCard adds items to a **Set** `hiddenCardKeys`

Sets don't allow duplicate items, so this test is testing Swift Foundation Sets. We can delete it.

WMFComponents/Tests/WMFComponentsTests/WMFHomeViewModelTests.swift
```diff
@@ -208,4 +230,18 @@ final class WMFHomeViewModelTests: XCTestCase {
...
+        let english = WMFLanguage(languageCode: "en", languageVariantCode: nil)
+        let spanish = WMFLanguage(languageCode: "es", languageVariantCode: nil)
+
+        vm.selectedLanguage = english
+        vm.forYouFeedError = URLError(.notConnectedToInternet)
+        XCTAssertNotNil(vm.forYouFeedError)
+
+        vm.selectedLanguage = spanish
+
+        XCTAssertNil(vm.forYouFeedError)
```
mazevedofs, line 364: If it's a no internet error, should changing the language clear it? It'll cause another unsuccessful API call and another error, so it shouldn't be nil

WMFComponents/Sources/WMFComponents/Components/Home/WMFForYouViewModel.swift
```diff
@@ -24,34 +32,109 @@ public struct WMFForYouModuleVisibility {
...
+            let header = WMFForYouHeaderLabel(
+                prefix: WMFLocalizedString("for-you-header-interest-topic-prefix", value: "Because of your interest: ", comment: "Prefix for a For You feed card header based on an interest topic."),
+                boldSuffix: $0.topic.displayName
+            )
+            return WMFForYouPageViewModel(module: .basedOnInterests, headerLabel: header, articles: deduplicated($0.articles))
         }
         let relatedPages = response.interestPageRelatedArticles.map {
-            WMFForYouPageViewModel(module: .basedOnInterests, headerLabel: "Interest Article: \($0.pageInterest.title)", articles: $0.articles)
+            let header = WMFForYouHeaderLabel(
+                prefix: WMFLocalizedString("for-you-header-interest-article-prefix", value: "Because of your interest: ", comment: "Prefix for a For You feed card header based on an article the user has shown interest in."),
```
mazevedofs: Identical to 

`WMFLocalizedString("for-you-header-interest-topic-prefix", value: "Because of your interest: ", comment: "...")` 
You can use only one of them, it's the same file

WMFComponents/Sources/WMFComponents/Components/Home/WMFForYouViewModel.swift
```diff
@@ -79,36 +162,213 @@ public final class WMFForYouPageViewModel: ObservableObject, Identifiable {
...
+            let scale = 0.25 / maxComponent
+            r *= scale; g *= scale; b *= scale
+        }
+        return (r, g, b)
+    }
+}
+
+
+extension String {
+    public var normalize: String {
```
mazevedofs: Move to the string extension file

Check if it's really working/being used

Image

WMFComponents/Sources/WMFComponents/Components/Home/WMFForYouView.swift
```diff
@@ -73,72 +273,204 @@ private struct WMFForYouPageView: View {
...
-        }
-        .onAppear {
-            viewModel.load()
+            .clipped()
+            .contentShape(Rectangle())
+            .onTapGesture { onTapCard() }
+            .onAppear { viewModel.load() }
+            .overlay {
+                if viewModel.loadState == .loading {
+                    Color.clear
```
mazevedofs: Is the idea to have it empty if it's still loading?

WMFComponents/Sources/WMFComponents/Components/Home/WMFForYouViewModel.swift
```diff
@@ -79,36 +162,213 @@ public final class WMFForYouPageViewModel: ObservableObject, Identifiable {
...
 
     deinit {
         loadTask?.cancel()
     }
 }
+
+// MARK: - WCAG color sampling
+
+extension UIImage {
+    func accessibleSampledColor() -> Color? {
```
mazevedofs: This func runs on the main thread, which is another probable cause of the delay in image loading and the weirdness in the card type animation.

Can we run this on a background thread? This porcess need info from the main thread, but it doesn't need to happen on the main thread.

It can process all this in a less busy lane and just return the RGB back. 

PS: you can't send a UIImage to a background thread, but you can send it as data.

WMFComponents/Sources/WMFComponents/Components/Home/WMFForYouView.swift
```diff
@@ -73,72 +273,204 @@ private struct WMFForYouPageView: View {
...
-            .padding(.horizontal, 20)
-            .padding(.bottom, 44)
+    private var dotsAndTabBarHeight: CGFloat {
+        windowSafeAreaBottom + 49 + 12 + 20 + 8 + 20
+    }
+
+    /// If the assigned variant requires an image but none is available, fall back to textFocused.
+    private var effectiveVariant: WMFForYouCardVariant {
+        if variant != .textFocused && viewModel.uiImage == nil {
+            return .textFocused
```
mazevedofs, line 374: Another possible culprit for the image/style flickering:

Since the image takes a while to load, but we do have the info that an article should have an image, you can check if `imageURL != nil` instead; you'd have to update the view model and pass it down.

Then the card variant wouldn't update on the fly. You'd only need to change the card format after an image loading failure, which is less common.

mazevedofs (commented):
Some other comments:
in iOS 18, it looks weird with the white tab bar, plus the pill format is weird

Image

tonisevener (approved):

tonisevener:
A couple of things, can be done in followup:

I think the Community / For You segmented control does not scale with dynamic type:
Screenshot 2026-08-12 at 4 43 14 PM

We may want to add some sort of background color or shadow to the iOS17 & iOS18 top navigation bar buttons (W icon, tabs and profile icons).

Screenshot 2026-08-12 at 4 50 54 PM

PR #6030, opened 2026-07-24
Update url banner params

voyagerfan:
**Phabricator:** 
https://phabricator.wikimedia.org/T415791

### Notes
* The AppsCampaignConfig.json file will be bumped to version 3 which will include a parameterized URL string to be populated by the app.
* This PR updates `webViewURL` to replace values in the parameterized string.

### Test Steps
1. If current time & date is not between August 3, 2026, at 09:00:00 AM PST and August 31, 2026, at 4:59:00 PM PST, set the emulator date/time by updating the System date and time. Go to Settings -> General -> Date & Time. Disable "Set time and date automatically". Set time/date anytime between aforementioned date/time range.
2. Run the app in Xcode on the Staging scheme in simulator. The Staging scheme is set up to point to the Test Wiki config.

Screenshot 2026-07-23 at 11 25 56 AM

3. Set the simulator region to Brazil by going to the Settings App (in the simulator) -> General -> Language & Region
4. Go back to the app. When you land on Explore, background the app, foreground, then pull to refresh on Explore. This ensures the fundraising config is fetched.
5. Visit an article on EN Wikipedia. You should see a campaign modal appear
DonateModal

6. Click on donate now to see the next modal:
AppleOrWeb

7. Click on "other payment method" to be taken to the web view.
webview

### Checklist

### Screenshots/Videos


Wikipedia/Code/DonateCoordinator.swift
```diff
@@ -86,21 +86,30 @@ class DonateCoordinator: Coordinator {
...
-            urlString = "https://donate.wikimedia.org/?wmf_medium=WikipediaApp&wmf_campaign=iOS&wmf_source=\(metricsID)&uselang=<langcode>"
-            urlString = urlString.replacingOccurrences(of: "<langcode>", with: languageCode)
+            parameterizedUrl = "https://donate.wikimedia.org/?wmf_medium=WikipediaApp&wmf_campaign=$platform;&wmf_source=$formattedId;&uselang=$language;&app_install_id=$appInstallId;&app_version=$appVersion;"
         }
-
-        let appVersion = Bundle.main.wmf_appVersion()
-        let appInstallID: String? = try? WMFDataEnvironment.current.crossProcessUserDefaultsStore?.load(key: WMFUserDefaultsKey.appInstallID.rawValue)
-        return URL(string: urlString)?.appendingAppVersionAndAppInstallID(appVersion: appVersion, appInstallID: appInstallID)
+        
+        let finalUrlString = parameterizedUrl
```
voyagerfan: Performing the variable replacement in the coordinator is not ideal. Url string usage is dependent on the `source`. I'm open to any other ideas.

Wikipedia/Code/DonateCoordinator.swift
```diff
@@ -86,21 +86,30 @@ class DonateCoordinator: Coordinator {
     private var webViewURL: URL? {
 
         guard let metricsID,
-              let languageCode else {
+              let languageCode,
+              let country = Locale.current.region?.identifier,
+              let appVersion = Bundle.main.wmf_appVersion(),
+              let appInstallID: String = try? WMFDataEnvironment.current.crossProcessUserDefaultsStore?.load(key: WMFUserDefaultsKey.appInstallID.rawValue)
```
voyagerfan: Note: changing appInstallId call from `String?` to `String`

Wikipedia/Code/DonateCoordinator.swift
tonisevener: I think it's okay for now. This DonateCoordinator is a shared class called from both article campaign models, year in review donate buttons, and our profile menu. So it is helpful that this logic can be shared across those features. Something you could do is write an extension method, similar to the old `appendingAppVersionAndAppInstallID`. Then we could potentially re-use in the future as-needed outside of DonateCoordinator.

```
extension URL {
    func replacingDonateParameters(languageCode: String?) -> URL? {

       let languageCode,
              let country = Locale.current.region?.identifier,
              let appVersion = Bundle.main.wmf_appVersion(),
              let appInstallID: String = try? WMFDataEnvironment.current.crossProcessUserDefaultsStore?.load(key: WMFUserDefaultsKey.appInstallID.rawValue)

        let urlString = self.absoluteString.
            .replacingOccurrences(of: "$formattedId;", with: metricsID)
            .replacingOccurrences(of: "$country;", with: country)
            .replacingOccurrences(of: "$language;", with: languageCode)
            .replacingOccurrences(of: "$appVersion;", with: appVersion)
            .replacingOccurrences(of: "$appInstallId;", with: appInstallID)
            .replacingOccurrences(of: "$platform;", with: "iOS")

        return URL(string: urlString)
    }
}
```

Then at the call site:

```
private var webViewURL: URL? {

    var url: URL?
    if case .articleCampaignModal(_, _, let articleCampaignDonateURL) = source {
          url = articleCampaignDonateURL.replacingDonateParameters(languageCode: languageCode)
    } else {
          url = URL(string:"https://donate.wikimedia.org/?wmf_medium=WikipediaApp&wmf_campaign=$platform;&wmf_source=$formattedId;&uselang=$language;&app_install_id=$appInstallId;&app_version=$appVersion;")?.replacingDonateParameters(languageCode: languageCode)
    }

    return url
```

tonisevener (commented):
Looking good, note I pushed up a couple of commits that removed some unneeded code that we discovered in our meeting. Refactoring to URL extension method is optional.

Going to see if I can test through the sandbox this afternoon, then I'll approve after that.

tonisevener (approved):

PR #6043, opened 2026-07-30
T433597: Fix find in page return key on iOS 26

woahitsraj:
Phabricator ticket: https://phabricator.wikimedia.org/T433597

## What
On iOS 26, tapping the keyboard's blue "Search" (return) key in the article find-in-page bar did nothing, leaving the keyboard covering the article and its highlighted matches. On iOS <26 the return key dismisses the keyboard so results are visible and navigable.

## Why
The find-in-page rewrite for iOS 26 split `WMFFindAndReplaceKeyboardBar.xib` into duplicate legacy + modern view hierarchies, and only the legacy text field got its `delegate` outlet connected. The modern field's delegate was unconnected, so `textFieldShouldReturn` → `keyboardBarDidTapReturn` never fired.

The modern close (✓), clear (✗), and `textFieldDidChange:` actions were also wired to the First Responder placeholder (`destination="-2"`) instead of the view. They currently function via responder-chain fallback, but are retargeted to the view here for parity with the legacy hierarchy.

## Changes
- `Wikipedia/Code/WMFFindAndReplaceKeyboardBar.xib`: connect the modern text field's `delegate` outlet to the bar; retarget the three modern actions from the First Responder proxy to the view (mirroring the legacy wiring).
- `WikipediaUITests/Robots/ArticleRobot.swift`: `findInArticle` now presses return and asserts the keyboard dismisses, so this regression stays covered (note the find test currently only runs in the nightly fixture lane, not the PR E2E lane).

## Testing
`ArticleControlsUITests/testArticleFindInArticleSearchesAndCloses` passes with this change in all four `UITests` test-plan configurations — English (Light), German (Dark), Hebrew (Sepia), Vietnamese (Black) — on iPhone 17 Pro / iOS 26.5, plus a pass of the English (Light, E2E) live-network configuration. `scripts/lint-ui-tests.sh` passes. The iOS <26 legacy layout wiring is untouched.


mazevedofs:
@woahitsraj thanks for your contribution!

mazevedofs (approved):

PR #6045, opened 2026-07-31
Fix inflated reading time in the Activity tab

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T429908

### Notes
* The app kept a reading-time stopwatch running on **every article the user had opened**, not just the one on screen. Articles stack up as you follow links and each of those screens stays in memory, so every time the app came back to the foreground all of them started timing again — a user with five articles open had every minute in the app counted five times. That is why the total could exceed the number of hours in the period: it was counting app time repeatedly, not reading time.
* Fix: only the article actually on screen counts reading time.
* Also fixed two smaller issues in the same code: an interval could be recorded twice if the user left an article and backgrounded the app at the same moment, and the "week" figure actually covered eight days.
* Affected most people who read more than one article in a sitting. The overcount scales with how many articles are left open, so light readers saw nearly correct numbers and heavy readers saw wildly inflated ones.
* Time spent with the app backgrounded is not counted. The stopwatch stops when the app stops being active and a fresh one starts on return, so the gap is never bridged. There is a test for exactly this.
* **No upper bound on a reading session.** A three hour read records three hours. A cap was considered and rejected: it would truncate genuine long reads, and the scenario it was guarding against is narrow — `UIApplicationSupportsMultipleScenes` is false, so a second window parked beside another app cannot happen. What remains is an article left on screen with the device kept awake, which auto-lock closes in the ordinary case (the app never disables the idle timer). Accepted trade-off.
* **Includes a one-time data cleanup.** The bad totals were saved as they happened, so they are already on device — fixing the measurement does not remove them. Left alone the weekly figure would stay wrong for another week, and Year in Review for up to a year. On first launch after updating, stored values above one hour on a single page view are trimmed to one hour. This check will run once, date-gated to Aug 31st.
* That cleanup ceiling is deliberately stricter than what live measurement now allows. Every row it touches was written by the buggy code, so trimming a genuine long read among them is a one-time loss on data already known to be contaminated, and it buys nobody opening the app to an implausible total. The constant is named `inflatedPageViewSecondsCeiling` so it cannot be misread as a rule about live measurement.
* **The cleanup never runs after 31 August 2026.** A once-per-install flag alone would not have been enough: a device first reaching this build well after release — a fresh install, or a launch where Core Data was not ready the first few times — would have applied the one hour ceiling to reading time the fixed code recorded correctly, and cut a genuine long read. The ceiling is only defensible against rows the buggy code wrote, so it is now bounded to the release window as well as to once per install.
* Consequence of that bound: anyone who does not update until after 31 August keeps their inflated history. Their weekly figure will still look right, because the old rows fall outside the seven day window by then, but their Year in Review 2026 will include the inflated numbers.
* The cleanup removes the extremes, it does not restore true totals. Overcounting spread thinly across many articles stays under the ceiling and survives, and the real values are not recoverable — only a running total was ever stored. Expect numbers after the cleanup to be believable rather than exact, becoming accurate as the old data ages out over the following week.

**Testability.** The timing rules lived as loose state on `ArticleViewController`, which made the fix effectively untestable: instantiating one needs an `articleURL`, an `MWKDataStore` and a theme, and the behaviour depends on two app-wide notifications. Deleting the on-screen guard would not have failed a single test.

* Moved the rules into `WMFReadingIntervalTracker` — a value type, no UIKit, takes dates rather than reading a clock. The view controller extension is now a thin adapter: forward the lifecycle event, persist whatever seconds come back.
* One deliberate behaviour change: an interval opens on `viewDidAppear` rather than waiting for the page view row to exist. The row is only needed to attribute the seconds at close time, so this no longer loses the gap between the article appearing and its content finishing its load.
* `WMFActivityTabDataController` now accepts an injected Core Data store, so the seven day window has a real test.
* Reading time had no test coverage at all before this. Added 23 test cases across three new Swift Testing suites; the whole suite is green (178 XCTest, 103 Swift Testing).
* Verified the tests actually bite: reverting the on-screen guard fails three of them, including the two that state the reported bug directly (an off-screen article recording nothing across foreground cycles, and two live trackers where only the visible one accumulates). The failure output quantifies the bug — 5400 seconds of phantom reading from one off-screen article over three foreground cycles.
* Also covered: long intervals reported at full length (90 min, 3 h, 8 h), live measurement explicitly ignoring the cleanup ceiling, the double-record race, a backwards device clock, content load ordering, and both sides of the cleanup cutoff. `now` is injected into the cleanup so the cutoff is tested rather than dependent on when the suite runs.
* Two test bugs found and fixed on the way, both worth knowing beyond this PR. `#expect` inside a `viewContext.perform` closure is not attributed to the running test — a real failure was reported against «unknown» while the suite line still read "passed"; values are now read out of the closure and asserted outside it. And `Int64?` compared against an untyped literal expression fails inside `#expect` while printing both sides as equal; stored seconds are now unwrapped with `#require` and compared against an explicit `Int64`. Other suites comparing Core Data integer attributes this way may be silently affected.
* Written with Swift Testing to match the rest of WMFDataTests. The four page view tests live in their own suite rather than the pre-existing XCTest class, since one type cannot mix the two frameworks — so `WMFPageViewsDataControllerTests` stays untouched.
* Not covered: the adapter wiring itself — that `viewDidAppear` calls `trackArticleDidAppear`, and that the notification observers point at the right methods. Four one-line call sites in `ArticleViewController`, verifiable by reading; testing them means instantiating the view controller, which is what this refactor was avoiding.

### Test Steps
1. Open five or six articles by following links, without backing out.
2. Background and foreground the app three times, waiting about a minute each time.
3. Go to the Activity tab and check "Time spent reading". It should only reflect foreground time on the article that was visible — not roughly (time elapsed × articles open), which is what happens before this change.
4. Confirm the weekly figure covers seven days rather than eight.
5. On a device carrying pre-fix data, confirm the cleanup runs once on first launch and not again on subsequent launches (look for the "Clamped inflated reading time on N page views" log line).

### Checklist


tonisevener (approved):
Good catch on these! I agree with this approach.

PR #6046, opened 2026-07-31
T424264: Always apply safe area insets to the source editor text view.

ealbizzati:
**Phabricator:** 

### Notes
- text view underlaps the translucent nav bar and relied on uikit's automatic content inset adjustment, but `.automatic` only applies safe area inset when the content is scrollable. short sections rendered tucked in the navigation bar with no way to scroll them into view. 
- Fix is setting ` contentInsetAdjustmentBehavior = .always`  on the editor textview, so the safe area inset is applied regardless of content size. The keyboard bottom inset math in `WMFSourceEditorViewController.updateInsets`. is unaffected since manual content insets are added on top of the safe area.

### Test Steps
1. open a short article or section.
2. confirm the first line renders below the navigation bar instead of tucked under it.
3. open a long article in the editor and confirm there is no double spacing at top.
4. type at the end of the text with the keyboard open and confirm the content scrolls above the keyboard
5. tap the find icon in the toolbar and confirm no extra gap below the find bar.

### Checklist

### Screenshots/Videos
the issue:

https://github.com/user-attachments/assets/ddffd67f-2489-4a60-9c23-5fbc7df029aa

with the fix:

https://github.com/user-attachments/assets/510675ca-701a-4e4f-aa2c-ba73a35c9c06


WMFComponents/Sources/WMFComponents/Components/Editors/Source Editor/WMFSourceEditorTextFrameworkMediator.swift
```diff
@@ -101,6 +101,8 @@ final class WMFSourceEditorTextFrameworkMediator: NSObject {
         
         textView.textContainerInset = .init(top: 16, left: 8, bottom: 16, right: 8)
         textView.translatesAutoresizingMaskIntoConstraints = false
+        textView.contentInsetAdjustmentBehavior = .always
+        textView.alwaysBounceVertical = true
```
ealbizzati, line 105: the alwaysBounceVertical isn't needed, I just thought it `felt` better. but I can remove it.

tonisevener (approved):
Thanks!

PR #6051, opened 2026-08-03
Add GitHub Action deploy workflows

tonisevener:
**Phabricator:** N/A

### Notes
These are new workflows to deploy the app from GitHub Actions instead of Xcode Cloud. Marking this as WIP while I finish up documentation on it, then will reference that documentation here. I would like us to start using this for the next release duty (it was unofficially used for the last one while I built these up).

Note that the commits are very meandering, but I didn't want to clean them up since a lot of them have tags that I want to keep in the remote repository.

Google Doc of documentation / edit suggestions is now open: https://docs.google.com/document/d/1-wASFeLRTID2s8QeF5sO9ScK8FbJ6XVWHmxKeYAZ3g4/edit?usp=sharing

We can move to a more open format for documentation once this is settled.

### Test Steps
1. This has already been tested to deploy release 2026.07.31 (6137), so just code review needed at this point.

### Checklist


.github/workflows/deploy_appstore.yml
```diff
@@ -1,12 +1,290 @@
...
+      # 5. Tag the build
+      # -------------------------------------------------------
+      - name: Tag build
+        env:
+          BUILD: ${{ steps.versioning.outputs.build }}
+        run: |
+          git config user.name "github-actions"
+          git config user.email "actions@github.com"
+          git tag "betas/$BUILD"
+          git push origin "betas/$BUILD"
```
ealbizzati, line 160: the build number is computed on line 30, based on existing tags, but the tag is not pushed until here. 

the git tags are the only record of which numbers are taken, so there's a 20-ish(?) minute window here where the build number is taken, but nothing on the repo says so. 

if there's ever a failure between upload and creating a tag, App Store connect will see build N+1 but tags will only exist until build N. and the script will never be able to self heal from it. should we create the tag on the same step we're computing the build number? (could also delete on failure)

I also think we could have issues with 2 concurrent runs, but I don't know how often we'd go into that case... 
but both issues could be prevented if we claim the tag before the archive with a retry.

also the tag push with an existing name won't always fail (only if the tag points to a different commit). 
if both runs are on the same SHA, the push is a no-op and exits 0. but if this was using gh we'd get a 422 "Reference already exists" whenever the tag already exists, regardless of the SHA. worth noting 422 is overloaded on that endpoint though (rulesets can also 422 with Reference update failed), so a retry would need to match on the message, not just the status.
so now I'm second guessing this approach because of having to match with the message. what do you think?

.github/workflows/deploy_testflight_internal.yml
```diff
@@ -1,7 +1,15 @@
 name: Internal TestFlight
 
 on:
+  schedule:
+    - cron: '0 10 * * *' # run at 10AM UTC
```
ealbizzati, line 5: this runs at the same time as tag_latest_beta.yml, which kicks off the xcode cloud nightly build, and  with the ci_pre_xcodebuild.sh change both now derive the build number from the same computed property. are we turning the xcode cloud nightly off as part of this, or should these be staggered? as-is I think they'd race for the same number every night.

.github/workflows/deploy_appstore.yml
```diff
@@ -1,12 +1,290 @@
...
+      submit_for_review:
+        description: 'Submit to App Store Review after uploading?'
+        required: false
+        type: boolean
+        default: false
+      version_override:
+        description: "Override the version (YYYY.MM.DD) instead of using today's date - use this to reuse an existing version instead of getting a new one. Leave blank to use today's date."
+        required: false
+        type: string
+        default: ''
```
ealbizzati, line 15: nit: missing `permissions: contents: write` here and for the other three deploy workflows. release_wrap_up.yml declares it. 

today it works off the repo default, but if that's ever tightened these fail after the upload has already gone through.

ealbizzati (commented):
Just leaving those 3 questions, let me know what you think.

.github/workflows/deploy_appstore.yml
tonisevener: Good catch - I always avoided pushing the tag before the build just in case building failed, but if we can catch the failure and delete it, that should be fine. Will open up a followup PR with that change if that's ok.

.github/workflows/deploy_testflight_internal.yml
tonisevener: Yeah, I'm planning on disabling the xcode nightly build once this gets merged.

.github/workflows/deploy_appstore.yml
tonisevener: Good catch, will address in a followup PR, thanks.

l-olson1214 (approved):

PR #6053, opened 2026-08-05
T427577: One time onboarding

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T427577

### Notes
* 

### Test Steps
1. 

### Checklist

- [x] Interstitial is shown on first app open after updating to the version containing the new Home feed
- [x] Interstitial is only shown to existing users, not new installs
- [x] Interstitial is only displayed when the Home tab is visible (not over an article, modal, or other location)
- [x] Interstitial is shown at most once — dismissing it or backgrounding the app during it prevents it from reappearing on subsequent opens
- [x] "Customize my feed" button launches the condensed onboarding flow (Interests Selection → Feed Preference → Feed Loading screens only, skipping welcome and privacy screens)
- [x] "Set it up for me" button dismisses the interstitial and takes the user directly to the feed
- [x] After "Set it up for me", the active tab defaults to Community
- [x] After "Set it up for me", the For You tab is pre-populated in the background using existing reading history (e.g. Because you read, Continue reading) so it is ready when the user switches to it
- [x] After completing the condensed "Customize my feed" flow, the user lands in the Home feed with their selections applied
- [x] After completing the condensed flow, the active tab reflects the feed preference the user chose (For You or Community)

### Screenshots/Videos
Simulator Screenshot - iPhone 17 Pro Max liquid glass - 2026-07-31 at 16 12 07

https://github.com/user-attachments/assets/656647a5-2699-4b69-813b-999ed55d2d0c
Screenshot 2026-08-05 at 1 09 36 PM
Screenshot 2026-08-05 at 1 11 20 PM

Simulator Screenshot - iPhone 16e 26 liquid glass - 2026-08-05 at 13 13 07

https://github.com/user-attachments/assets/70501125-bb77-43d7-8379-1c68c01f2b2b


Wikipedia/Code/WMFAppViewController.swift
```diff
@@ -1092,6 +1093,64 @@ final class WMFAppViewController: UITabBarController, AppTabBarDelegate {
...
+        // Make sure it's on the home tab and doesn't have any modals/etc
+        guard selectedIndex == WMFAppTabType.main.rawValue else { return }
+        guard let homeNav = viewControllers?[WMFAppTabType.main.rawValue] as? UINavigationController,
+              homeNav.viewControllers.count == 1 else { return }
+        guard presentedViewController == nil else { return }
+        
+        let defaults = UserDefaults.standard
+        let isNewInstall = defaults.bool(forKey: WMFUserDefaultsKey.didSendNewInstallOnboardingStartEvent.rawValue)
+        let alwaysShow = WMFDeveloperSettingsDataController.shared.alwaysShowNewOnboarding
+        let hasSeen = defaults.bool(forKey: WMFUserDefaultsKey.hasSeenNewHomeOnboarding.rawValue)
```
tonisevener: Can you flow these through WMFHomeDataController and use WMFUserDefaultsStore instead? Ideal for me if UserDefaults is accessed through the data layer.

```
let isNewInstall = WMFHomeDataController.shared.didSendNewInstallOnboardingStartEvent
etc.
```

Wikipedia/Code/WMFAppViewController.swift
```diff
@@ -1092,6 +1093,64 @@ final class WMFAppViewController: UITabBarController, AppTabBarDelegate {
...
+    
+    private func presentOneTimeHomeOnboardingIfNeeded() {
+        // Make sure it's on the home tab and doesn't have any modals/etc
+        guard selectedIndex == WMFAppTabType.main.rawValue else { return }
+        guard let homeNav = viewControllers?[WMFAppTabType.main.rawValue] as? UINavigationController,
+              homeNav.viewControllers.count == 1 else { return }
+        guard presentedViewController == nil else { return }
+        
+        let defaults = UserDefaults.standard
+        let isNewInstall = defaults.bool(forKey: WMFUserDefaultsKey.didSendNewInstallOnboardingStartEvent.rawValue)
```
tonisevener: UserDefaults `bool(forKey:)` has a tricky gotcha. If a value doesn't exist at all for this key, it defaults to false. This is incorrect in this case, because a fresh install would not have any value stored for this key, but because it defaults to false, it would evaluate to `isNewInstall = false`, which triggers this modal.

This is why it's handy to avoid using UserDefaults directly and instead run it through WMFUserDefaultsStore. Its load() method returns an optional, so you are forced to reckon with it not existing and setting a default yourself. I think in this case, if it doesn't exist, it should default to `true`.

```
let isNewInstall: Bool = WMFDataEnvironment.current.userDefaultsStore.load(key: WMFUserDefaultsKey.didSendNewInstallOnboardingStartEvent.rawValue)) ?? true
```

But (see earlier comment), ideally this code is accessed through WMFHomeDataController.

Wikipedia/Code/WMFAppViewController.swift
```diff
@@ -1092,6 +1093,64 @@ final class WMFAppViewController: UITabBarController, AppTabBarDelegate {
...
+        let alwaysShow = WMFDeveloperSettingsDataController.shared.alwaysShowNewOnboarding
+        let hasSeen = defaults.bool(forKey: WMFUserDefaultsKey.hasSeenNewHomeOnboarding.rawValue)
+
+        guard !isNewInstall && (alwaysShow || !hasSeen) else { return }
```
tonisevener: You're missing a check for the feature flag (`enableHomeTab` or `enableHomePhase2`, I'm not familiar with this feature anymore to know the difference 🙂 ).

WMFComponents/Sources/WMFComponents/Components/App Onboarding/One Time Onboarding/WMFOneTimeOnboardingView.swift
```diff
@@ -0,0 +1,89 @@
+import SwiftUI
+import WMFData
+
+public struct WMFOneTimeOnboardingView: View {
```
tonisevener: Would it be possible to lean on our existing WMFOnboardingView component instead of writing a new one? It looks very similar. We should avoid spinning off feature-specific components. Plus then we wouldn't have to test the layout as heavily, knowing it's already proven itself in other features.

Image

tonisevener (changes requested):
There's a bug here where this displays on fresh installs, so if we merge, production users would see it before the feature is complete. I think my comments below explain why.

Wikipedia/Code/WMFAppViewController.swift
mazevedofs: @tonisevener sorry about the terrible names 😅 
`enableHomeTab` gives you the explore feed inside the community tab
`enableHomePhase2` has the all new community tab, no more old explore. 

I can rename the flags so it's easy to understand!

WMFComponents/Sources/WMFComponents/Components/App Onboarding/One Time Onboarding/WMFOneTimeOnboardingView.swift
l-olson1214: Good catch!! Thank you

Wikipedia/Code/WMFAppViewController.swift
tonisevener: No problem! They're fine as-is.

WMFComponents/Sources/WMFComponents/Components/App Onboarding/One Time Onboarding/WMFOneTimeOnboardingView.swift
```diff
@@ -6,76 +6,28 @@ public struct WMFOneTimeOnboardingView: View {
...
-                    .font(Font(WMFFont.for(.semiboldSubheadline)))
-                    .foregroundStyle(Color(uiColor: theme.text))
-                Text(item.body)
-                    .font(Font(WMFFont.for(.callout)))
-                    .foregroundStyle(Color(uiColor: theme.inputAccessoryButtonTint))
-            }
-            .frame(maxWidth: .infinity, alignment: .leading)
-        }
-        .accessibilityElement(children: .combine)
+        WMFOnboardingView(
```
tonisevener: (Optional) Would it be possible to use this component raw instead of wrapping it up in the `WMFOneTimeOnboardingView`? Not sure what the wrapper buys us, and less code is better than more code.

Wikipedia/Code/WMFAppViewController.swift
```diff
@@ -1092,6 +1093,64 @@ final class WMFAppViewController: UITabBarController, AppTabBarDelegate {
...
+    private var homeFeedSettingsCoordinator: HomeFeedSettingsCoordinator?
+    
+    private func presentOneTimeHomeOnboardingIfNeeded() {
+        // Make sure it's on the home tab and doesn't have any modals/etc
+        guard selectedIndex == WMFAppTabType.main.rawValue else { return }
+        guard let homeNav = viewControllers?[WMFAppTabType.main.rawValue] as? UINavigationController,
+              homeNav.viewControllers.count == 1 else { return }
+        guard presentedViewController == nil else { return }
+
+        let isNewInstall = WMFHomeDataController.shared.didSendNewInstallOnboardingStartEvent()
```
tonisevener: Upon further thought, I think `didSendNewInstallOnboardingStartEvent` is not the right key to check for new installs. It is set to true whenever app onboarding appears from a fresh install. You can think about a user going through this flow:

1. Fresh install before Home feature is released
2. Legacy app onboarding displays, didSendNewInstallOnboardingStartEvent is set to true.
3. User later launches in new app version when Home feature released 
4. didSendNewInstallOnboardingStartEvent is _still_ set to true (because it is a persisted user default), meaning `isNewInstall` is evaluated to true here. The user would not ever see this one-time onboarding and would guard out at line 1111, which is a bug.

Can you find a different way to present this one-time onboarding for users that did not see app onboarding? I notice there's a `didShowOnboarding` flag passed into the `presentOnboardingIfNeeded` completion. Maybe you could check that before calling presentOneTimeHomeOnboardingIfNeeded.

```
private func resumeApp(_ completion: (() -> Void)?) {
        presentOnboardingIfNeeded { didShowOnboarding in
        ....
              self.finishResumingApp(didShowOnboarding)
         ....
}

private func finishResumingApp(didShowOnboarding: Bool) {
    ....
    if (!didShowOnboarding) {
        presentOneTimeHomeOnboardingIfNeeded()
    }
    ....
}
```

That approach is a bit spaghetti-code though, open to other suggestions.

Wikipedia/Code/WMFAppViewController.swift
tonisevener: My thinking has evolved on this one - See comment at https://github.com/wikimedia/wikipedia-ios/pull/6053/changes#r3762698026.

Wikipedia/Code/WMFAppViewController.swift
l-olson1214: Give that a shot!! I think this is working - if you have always show, it'll show right after the other one

tonisevener: Still unsure on this one, but let's move on and test it more fully later once our A/B test task is in. Right now we have to lean on the "force onboarding" flag and I think as long as we lean on that, it's not really a true test. Will note in https://phabricator.wikimedia.org/T427688 for us to re-check this without the force onboarding flag.

**Reasoning:**
The key you are leaning on now is set [immediately after app onboarding](                [self?.setDidShowOnboarding()](https://github.com/wikimedia/wikipedia-ios/blob/aaea33c0e90c706b75df97708afb58878eb508a5/Wikipedia/Code/WMFAppViewController.swift#L1536)), before `oneTimeOnboardingViewController` is ever called. So [at this point](https://github.com/wikimedia/wikipedia-ios/pull/6053/changes/2d21294ab57d541d469bb2166eeb689d8a651052#diff-1f33b03cccbc1216fb904fd8c2cc6369a0afa7ba3ed9adcea7bb872e3d49bd5fR1108), the flag will be set to true. This means for a fresh install user:

1. They go through app onboarding, Self.wmfDidShowOnboarding is set to true.
2. They then call presentOneTimeHomeOnboardingIfNeeded, which checks the flag. If true, it assumes they are an existing user, so they see the one time onboarding in the same session.

This would be double-onboarding situation to users.

We can address later on if it's an issue, this is safe to merge since it's behind a feature flag.

tonisevener (approved):

tonisevener (approved):

PR #6054, opened 2026-08-05
Persist article topics alongside reading history

dbrant:
Parse the topics reported in the Page Content Service's `final_setup` message, which were previously discarded along with the rest of that message's payload, and persist them in `WMFData`.

Topics are stored one row per (article, topic) pair with their confidence score, keyed on the existing CDPage so they relate to reading history, saved info and tabs. Model bumped to WMFData 8; the change is additive so it migrates lightweight.

On article load an article's topics are replaced wholesale in a single transaction, after dropping empty names, de-duplicating by name keeping the highest score, ordering by score and capping at five. An empty payload means "no topic data" rather than "topics were removed", so it leaves stored topics alone.

Aggregates reduce history to distinct articles before totalling, so an article read repeatedly counts once. Unlike Android, seconds are recorded per page view rather than per article on iOS, so they are summed per article and then attributed to each of its topics once.

Both paths that clear reading history now clear topics too, hence `deleteAllPageViewsAndCategories` -> `deleteAllPageViewsCategoriesAndTopics`.

https://phabricator.wikimedia.org/T429634


WMFData/Sources/WMFData/Data Controllers/Shared/WMFPageTopicsDataController.swift
```diff
@@ -0,0 +1,270 @@
+import Foundation
+import CoreData
+
+/// Persists the topics an article belongs to, keyed on the same `CDPage` that reading history hangs off of,
+/// so that topics can be aggregated over a period of reading history.
+// @unchecked Sendable: coreDataStore is an immutable let; all mutations go through backgroundContext.perform { }
+public final class WMFPageTopicsDataController: @unchecked Sendable {
+
+    /// Maximum number of topics persisted for a single article.
+    static let maximumTopicsPerPage = 5
```
l-olson1214, line 9: Why are we limiting it to 5 out of curiosity? Didn't see it in the ticket, thank you!

l-olson1214 (approved):

mazevedofs (approved):
Looks good, I pushed a small non-functional fix

PR #6056, opened 2026-08-06
T433115: Images preferred 

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T433115

### Notes
* Needs #6029 merged first, then updated to main branch base

### Test Steps
1. Load For You
2. Make sure articles have images in the proper indexes mostly (more often than not, not every article will have an image so we balance)

### Checklist

### Screenshots/Videos

https://github.com/user-attachments/assets/fdf948b9-d199-4cf1-bf64-e6b8bcdce3f1


WMFData/Sources/WMFData/Data Controllers/Home/WMFHomeDataController.swift
```diff
@@ -196,7 +196,12 @@ public final actor WMFHomeDataController {
             for topic in topics {
                 group.addTask {
                     let articles = try await self.fetchArticles(for: topic, project: project)
-                    let mapped = articles.shuffled().prefix(4).map { WMFForYouArticle(title: $0.title, project: project) }
+                    let mapped = articles
+                        .shuffled()
```
mazevedofs: Does suflling here to sort it afterwards have any effect?

WMFData/Sources/WMFData/Data Controllers/Home/WMFHomeDataController.swift
```diff
@@ -196,7 +196,12 @@ public final actor WMFHomeDataController {
             for topic in topics {
                 group.addTask {
                     let articles = try await self.fetchArticles(for: topic, project: project)
-                    let mapped = articles.shuffled().prefix(4).map { WMFForYouArticle(title: $0.title, project: project) }
+                    let mapped = articles
```
mazevedofs: You're adding this change only to `fetchForYouInterestTopicRandomArticles`, but the ticket doesn't mention limiting it to random. I believe we should apply the logic to all fetchers.

mazevedofs: You can create a helper function to do this sorting, add unit tests to it, and reuse in every fetch.

mazevedofs (changes requested):

WMFData/Sources/WMFData/Data Controllers/Home/WMFHomeDataController.swift
```diff
@@ -196,7 +196,8 @@ public final actor WMFHomeDataController {
             for topic in topics {
                 group.addTask {
                     let articles = try await self.fetchArticles(for: topic, project: project)
-                    let mapped = articles.shuffled().prefix(4).map { WMFForYouArticle(title: $0.title, project: project) }
+                    let mapped = await self.assignCardSlots(articles.shuffled())
```
mazevedofs: Here you're shuffling the articles to reorder them down the road using the images. It has no effect, from your assignCardSlots function:

>  // Slot layout: 0 = image, 1 = image, 2 = text, 3 = image

WMFData/Sources/WMFData/Data Controllers/Home/WMFHomeDataController.swift
```diff
@@ -205,6 +206,14 @@ public final actor WMFHomeDataController {
             return results
         }
     }
+    
+    /// Sorts articles so that those with thumbnails come first, using the decoded
+    /// `index` field as a stable tiebreaker within each group.
+    private func sortedByThumbnailPriority(_ articles: [WMFRandomArticle]) -> [WMFRandomArticle] {
```
mazevedofs: This is not being used

WMFData/Sources/WMFData/Data Controllers/Home/WMFHomeDataController.swift
```diff
@@ -225,6 +234,32 @@ public final actor WMFHomeDataController {
...
+                return imageQueue.next() ?? textQueue.next()
+            } else {
+                return textQueue.next() ?? imageQueue.next()
+            }
+        }
+
+        return [
+            next(preferImage: true),   // slot 0 — image
+            next(preferImage: true),   // slot 1 — image
+            next(preferImage: false),  // slot 2 — text
```
mazevedofs: The requirements say to "Preference articles with images for the first cards in each module", any reason to add a text-only card? Even if there are other image cards?

WMFData/Sources/WMFData/Data Controllers/Home/WMFHomeDataController.swift
```diff
@@ -225,6 +234,32 @@ public final actor WMFHomeDataController {
...
+        let withThumbnail = articles.filter { $0.thumbnail != nil }
+            .sorted { ($0.index ?? Int.max) < ($1.index ?? Int.max) }
+        let withoutThumbnail = articles.filter { $0.thumbnail == nil }
+            .sorted { ($0.index ?? Int.max) < ($1.index ?? Int.max) }
+
+        // Slot layout: 0 = image, 1 = image, 2 = text, 3 = image
+        var imageQueue = withThumbnail.makeIterator()
+        var textQueue = withoutThumbnail.makeIterator()
+
+        func next(preferImage: Bool) -> WMFRandomArticle? {
```
mazevedofs: Can we rename it to `nextImage`. Would be interesting to remove it from this other func and unit test it

WMFData/Sources/WMFData/Data Controllers/Home/WMFHomeDataController.swift
```diff
@@ -225,6 +234,32 @@ public final actor WMFHomeDataController {
             return results
         }
     }
+    
+    private func assignCardSlots(_ articles: [WMFRandomArticle]) -> [WMFRandomArticle] {
```
mazevedofs: You updated several tests unrelated to this ticket, but no tests were added for the new code. 
Can we add some unit tests?

WMFData/Sources/WMFData/Data Controllers/Home/WMFHomeDataController.swift
```diff
@@ -196,7 +196,8 @@ public final actor WMFHomeDataController {
             for topic in topics {
                 group.addTask {
                     let articles = try await self.fetchArticles(for: topic, project: project)
-                    let mapped = articles.shuffled().prefix(4).map { WMFForYouArticle(title: $0.title, project: project) }
+                    let mapped = await self.assignCardSlots(articles.shuffled())
+                        .map { WMFForYouArticle(title: $0.title, project: project) }
```
mazevedofs, line 218: This is still being added only to `fetchForYouInterestTopicRandomArticles`. The ticket asks to add it to the other fetchers for each module?

mazevedofs (changes requested):

WMFData/Sources/WMFData/Data Controllers/Home/WMFHomeDataController.swift
l-olson1214: I think the third one was the one that never has an image in the card, but let me see if this is forcing something weird

l-olson1214: Updated so we can have images or text in that one! Didn't intend to totally filter out images there, thank you for the catch!

WMFData/Sources/WMFData/Data Controllers/Home/WMFHomeDataController.swift
l-olson1214: Ended up getting rid of it

WMFData/Sources/WMFData/Data Controllers/Home/WMFHomeDataController.swift
l-olson1214: Added to all! Removed shuffle, created a second of the same func to fit the different types of pages and renamed for convenience

WMFData/Sources/WMFData/Data Controllers/Home/WMFHomeDataController.swift
l-olson1214: I think I ended up with a few funky things - fixed the unrelated tests back to normal!

WMFData/Sources/WMFData/Data Controllers/Home/WMFHomeDataController.swift
mazevedofs: we're still missing it in `fetchForYouContinueReading`

mazevedofs (changes requested):

WMFData/Sources/WMFData/Data Controllers/Home/WMFHomeDataController.swift
mazevedofs: I'm leaving this one to be reviewed by product

mazevedofs (approved):

PR #6057, opened 2026-08-07
T300050: Save article url in context menu

gemedet:
Bug: T300050

### Notes
* This PR uses `article.url` instead of `articleURL` when saving from the context menu.

### Test Steps
1. Open an article that contains a redirect link
2. Long press to open the context menu 
3. Save the redirected article

### Checklist

### Videos
https://github.com/user-attachments/assets/1f3b0363-9b8e-42e4-b1e5-817b5c77ec53


ealbizzati (approved):
Looks good, thanks.

verified that article.url is the redirect-resolved url. Beyond the state mismatch, this also removes the race described in T300050.

PR #6058, opened 2026-08-07
T427675: because you read

l-olson1214:
**Phabricator:** https://phabricator.wikimedia.org/T427675

### Notes
* Based on #6056 and #6029 (#6029 being priority 1, then #6056 being priority 2)
* Added disambiguation calls similar to Android
* Cannot filter out every single list (noted here on Slack: https://wikimedia.slack.com/archives/C0BB50D9TGD/p1786130244898889?thread_ts=1786117411.713379&cid=C0BB50D9TGD) 

### Test Steps
1. With a significantly viewed/saved article, open app without home enabled
2. Ensure you see Because You Read
3. Switch to Home
4. Ensure Because You Read is gone
5. Go to For You tab
6. Scroll down to Because You Read (past interests)
7. Feel free to add new articles by viewing them (note: 60s from the past month generally, but if that returns none then it checks for 10s views)
8. Feel free to refresh
9. Ensure in community settings, you cannot toggle on/off because you read (it doesn't exist)

### Checklist
- [x] The module selects the user's most recently read article (for more than 10 seconds) as the basis for recommendations
- [x] Avoid using a non-mainspace page as the seed article, and avoid suggesting non-mainspace pages as recommendations for Because You Read
- [x] 4 recommended article cards are displayed in a horizontal stack
- [x] Refresh the MoreLike suggestions daily
- [x] Only suggest articles in the main namespace, and do not suggest disambiguation articles
- [x] The reason label on each card reads "Because you read [Article Title]"
- [x] If the user has no reading history, this module is hidden from the feed until they've read at least one article
- [x] Exclude disambiguation pages and main page
- [x] Remove the existing "Because You Read" module and associated settings from the legacy explore feed (Community) for users in the B group of the experiment. Users in Control should maintain access.

#### Nice to Have

- [x] **Freshness:** Instead of just taking the last article read, for every new explore day, take all articles that were read for 1+ minutes in the last month and randomly pick one as the seed

### Screenshots/Videos
simulator_screenshot_A8CBFA56-20CE-40AE-A7DE-09CDE151995F

https://github.com/user-attachments/assets/f54ee458-8491-4284-b05d-256df93ddfb4

Simulator Screenshot - iPhone 17 Pro Max liquid glass - 2026-08-07 at 15 37 45


WMFData/Sources/WMFData/Data Controllers/Shared/WMFRelatedPagesDataController.swift
```diff
@@ -89,7 +95,10 @@ public actor WMFRelatedPagesDataController {
             return []
         }
 
-        return pages.map { page in
+        return pages.filter { page in
+            guard filterDisambiguationPages else { return true }
+            return page.pageprops?.disambiguation == nil
+        }.map { page in
```
mazevedofs, line 101: Need to filter out main pages as well

WMFData/Sources/WMFData/Data Controllers/Shared/WMFRelatedPagesDataController.swift
```diff
@@ -33,16 +33,21 @@ public actor WMFRelatedPagesDataController {
...
             let extract: String?
+            let pageprops: PageProps?
 
             struct Thumbnail: Codable {
                 let source: String?
             }
+
+            struct PageProps: Codable {
+                let disambiguation: String?
+            }
```
mazevedofs, line 45: You're asking main page from the api, but not usning to filter out pages
`"ppprop": "mainpage|disambiguation",`

mazevedofs (changes requested):

WMFData/Sources/WMFData/Data Controllers/Home/WMFHomeDataController.swift
```diff
@@ -245,14 +244,63 @@ public final actor WMFHomeDataController {
...
+            next(preferImage: false),  // slot 2 — text
+            next(preferImage: true)   // slot 3 — image
+        ].compactMap { $0 }
+    }
 
     private func fetchForYouBecauseYouReadArticles(project: WMFProject) async throws -> WMFForYouBecauseYouReadArticles? {
         guard let pageViewsDataController else { return nil }
-        let pages = try await pageViewsDataController.fetchRecentlyReadPages(project: project, minimumSeconds: 10)
+        // T427675:  Ensure freshness: instead of just taking the last article read, for every new explore day, take all articles that were read for 1+ minutes in the last month and randomly pick one as the seed
+        var pages = try await pageViewsDataController.fetchRecentlyReadPages(project: project, minimumSeconds: 60, mainNamespaceOnly: true)
```
mazevedofs, line 297: Does it have a way to track used seeds? So we can really ensure freshness.

WMFData/Sources/WMFData/Data Controllers/Shared/WMFRelatedPagesDataController.swift
```diff
@@ -89,7 +96,11 @@ public actor WMFRelatedPagesDataController {
             return []
         }
 
-        return pages.map { page in
+        return pages.filter { page in
+            guard page.pageprops?.mainpage == nil else { return false }
+            guard filterDisambiguationPages else { return true }
```
mazevedofs: WE'll always have to filter disambiguation pages out, so we can actually remove this guard 

`guard filterDisambiguationPages else { return true }` 


and this method argument

`filterDisambiguationPages: Bool = false`

mazevedofs (changes requested):
Almost there!

mazevedofs:
@l-olson1214 any reason not to check it against Swift 6 strict concurrency?

mazevedofs (approved):

PR #6059, opened 2026-08-09
Update Claude/Copilot build docs to a simulator destination that exists

IlyaSilantyev:
**Phabricator:**

### Notes
* The documented `xcodebuild` destination (`iPhone 16, OS=18.6`) in `.github/copilot-instructions.md` (symlinked as `CLAUDE.md`) isn't installed on current dev machines, so the build commands in the doc fail as written.
* Switched all three destinations to `iPhone 17 Pro, OS=latest` so the doc tracks whatever simulator runtime is actually installed, rather than going stale on the next Xcode update.
* Also fixed a pre-existing copy/paste bug: the WMFComponents build command was building the `WMFData` scheme instead of `WMFComponents`.
* Added a one-line fallback note pointing to `xcrun simctl list devices available` in case `iPhone 17 Pro` isn't present on a future machine.

### Test Steps
1. Ran `xcodebuild -scheme Wikipedia -project Wikipedia.xcodeproj -destination 'platform=iOS Simulator,name=iPhone 17 Pro,OS=latest' build` — succeeded.
2. Confirmed `iPhone 17 Pro` resolves via `xcodebuild -showdestinations -scheme Wikipedia -project Wikipedia.xcodeproj`.
3. Docs-only change; no app code touched.

### Checklist
- [x] Checked against Swift 6 strict concurrency (not applicable — documentation-only change)

### Screenshots/Videos
N/A (documentation-only change)


mazevedofs (approved):
Thanks for the contribution

PR #6063, opened 2026-08-12
T434229: Refactor choose editor sheet ahead of the editing preferences screen (1/2)

ealbizzati:
**Phabricator:** 
https://phabricator.wikimedia.org/T434229

### Notes
* Refactor choose editor sheet ahead of the editing preferences screen
	* replaces duplicate EditMode with WMFData's WMFEditMode, and extracts the option copy and the options card so the editing preferences settings screen can reuse them.

### Test Steps
1. Just guaranteeing modal still works the same way.

### Checklist

### Screenshots/Videos
- no visual changes.

### DIFF
--- a/WMFComponents/Sources/WMFComponents/Components/Shared/WMFEditModeCopy.swift
+++ b/WMFComponents/Sources/WMFComponents/Components/Shared/WMFEditModeCopy.swift
@@ -1,0 +1,28 @@
+import Foundation
+import WMFData
+import WMFNativeLocalizations
+
+enum WMFEditModeCopy {
+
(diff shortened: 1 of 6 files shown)


WMFComponents/Sources/WMFComponents/Components/Shared/WMFEditModeCopy.swift
```diff
@@ -0,0 +1,28 @@
+import Foundation
+import WMFData
+import WMFNativeLocalizations
+
+enum WMFEditModeCopy {
```
tonisevener, line 5: (Optional) Just FYI there is a CommonStrings.swift file that you could also use, for strings that are re-used in multiple places. We're not consistent with it and that file itself is getting unwieldy. So I'm open to more feature-specific namespace containers like this, just wanted to call it out.

tonisevener (approved):

PR #6067, opened 2026-08-13
T434821: Recognize new thumbnail domain

Ladsgroup:
We are migrating from upload.wikimedia.org to thumb.wikimedia.org (see T427465) which means MediaWiki will start advertising this domain (same exact path) instead of the upload one for thumbnails.

**Phabricator:** 
https://phabricator.wikimedia.org/T434821
### Notes
* Part of PP3.6.2

The new domain is not set up yet but this change is quite straightforward.


mazevedofs (approved):
Looks good, thanks!

PR #6068, opened 2026-08-13
T424181: Ensure Watchlist diff thanker completes on the main thread

fellipecaetano:
**Phabricator:** https://phabricator.wikimedia.org/T424181

### Notes

`DiffThanker` performs its network request asynchronously and invokes the configured completion handler on the same background queue as the network request. In this case (Watchlist "Thank" completion), the handler shows a toast, which is UI work that must be performed on the main thread. This pull request now ensures this.

### Test Steps

1. Open an article revision in the Watchlist.
2. Tap the revision author's menu and select **Thank**.
3. Confirm a toast appears without crashing.

Before:

https://github.com/user-attachments/assets/c66920ad-f2e1-4cd8-93af-4e9a9a693f8a 

After:

https://github.com/user-attachments/assets/2c473893-6e86-480e-9945-ed362c9b1bba

### Automated Testing

No test code was added. The affected code directly instantiates `DiffThanker` and presents toasts through the global `WMFToastManager`, so testing the full callback-to-UI path would require introducing additional dependency injection.


Wikipedia/Code/WMFAppViewController+Extensions.swift
```diff
@@ -369,13 +369,15 @@ extension WMFAppViewController: WMFWatchlistDelegate {
...
                 let diffThanker = DiffThanker()
                 diffThanker.thank(siteURL: siteURL, rev: Int(revisionID), completion: { result in
-                    switch result {
-                    case .success:
-                        let successfulThanks = WMFLocalizedString("watchlist-thanks-success", value: "Your ‘Thanks’ was sent to %@", comment: "Message displayed in a toast on successful thanking of user in Watchlist view. %@ is replaced with the user being thanked.")
-                        let successMessage = String.localizedStringWithFormat(successfulThanks, username)
-                        WMFToastManager.sharedInstance.showRichToast(successMessage, subtitle: nil, image: UIImage(named: "watchlist-thanks-checkmark"), dismissPreviousToasts: true)
-                    case .failure(let failure):
-                        WMFToastManager.sharedInstance.showRichToast(failure.localizedDescription, subtitle: nil, image: nil, dismissPreviousToasts: true)
+                    DispatchQueue.main.async {
```
mazevedofs: We're moving on towards more modern concurrency APIs, could you switch to 

`Task { @MainActor in` or something similar? Thanks

mazevedofs (changes requested):

Wikipedia/Code/WMFAppViewController+Extensions.swift
fellipecaetano: done!

mazevedofs (approved):
Looks good, thanks!

PR #6069, opened 2026-08-13
T434236: Add visual editor return journey to parse visual editor return URLs

ealbizzati:
> [!WARNING]
> returntoapp temporarily removed. Added a TODO mark to find it whenever we can add it back

**Phabricator:** 
https://phabricator.wikimedia.org/T434236

* Adds `VisualEditorReturnJourney`, a small value type that recognizes the URL the web hands back after a Visual Editor session (`wikipedia://{lang}.wikipedia.org/wiki/{title}?saved={bool}&revision={id}`, contract from T422075) and strips the return query items from the article URL. Mid-edit returns carrying our own outbound markers (`veaction`/`returntoapp`, e.g. tapping Safari's native app banner from the editing screen) are treated as not saved. Covered by unit tests.
* If the linked article is already on screen, `saved=true` refreshes it in place at the published revision (reusing `waitForNewContentAndRefresh`, the same mechanism the native editor uses post-save — preserves scroll position and bypasses stale caches); `saved=false` leaves the article untouched. With no revision available, the existing ETag polling handles replication lag.
* If the article is not on screen, it is pushed onto the current tab with the revision applied to the mobile-html request (`page/mobile-html/{title}/{revision}`). `revisionID` is threaded through `LinkCoordinator` → `ArticleCoordinator` → `ArticleViewController`'s initial load. Cold start works through the existing `unprocessedUserActivity` queue.
* ⚠️ Beyond ticket scope: any article deep link matching the visible article now brings the app to the foreground (scrolling to the URL fragment when present) instead of pushing a duplicate into a new tab. This also affects repeated generic deep links and widget taps.
* Outbound fix: the visual editor handoff now sends `useformat=mobile` — without it, tablets get the desktop site, which lacks MobileFrontend, the extension that owns the redirect back to the app.
* Known limitations: no fallback if PCS hasn't rendered the new revision yet (error surfaces, pull-to-refresh recovers); a post-publish return with no query markers (e.g. native Safari banner after the blue banner ships, T434258) is indistinguishable from a generic deep link.

### Test Steps
1. Enable the Visual Editing Journey flag in Developer Settings.
2. With an article (e.g. Cat) on screen, run:
   `xcrun simctl openurl booted "wikipedia://en.wikipedia.org/wiki/Cat?saved=true&revision={old revision id}"`
   → the article refreshes in place at that revision, no duplicate push, no new tab.
3. Repeat with `?saved=false` → nothing happens, the article stays as-is.
4. Navigate to a different article, run step 2's command again → Cat is pushed onto the current tab (no new tab created).
5. Terminate the app, run step 2's command → app launches directly into the article.
6. With Cat on screen, run:
   `xcrun simctl openurl booted "wikipedia://en.wikipedia.org/wiki/Cat#Etymology_and_naming"`
   → no duplicate push; the on-screen article scrolls to the section.
7. End-to-end (auto-redirect is live in production): open Cat → Edit article → visual editor (opens Safari) → publish an edit → the web redirects back and the app refreshes the article showing the new edit. On an iPad, verify Safari opens the mobile site (`useformat=mobile`).

### Checklist

### Screenshots/Videos

https://github.com/user-attachments/assets/bfe82c5a-2169-4cd8-9260-ad86a88a14ab


Wikipedia/Code/VisualEditorReturnJourney.swift
ealbizzati, line 1: Dedicated type so the web URL contract has one unit-testable home. 

Failable init doubles as the routing guard (nil = not a VE return, regular routing untouched), and articleURL is returned with the return params stripped so they never leak past routing. 

If the contract needs an update, this should be the only file that needs to changes.

Wikipedia/Code/VisualEditorReturnJourney.swift
```diff
@@ -0,0 +1,75 @@
...
+    /// The newly published revision, when the web included one. Always nil for abandoned edits.
+    let revisionID: UInt64?
+
+    /// The article URL with the return journey query items stripped.
+    let articleURL: URL
+
+    private static let savedQueryItemName = "saved"
+    private static let revisionQueryItemName = "revision"
+    private static let visualEditorActionQueryItemName = "veaction"
+    private static let returnToAppQueryItemName = "returntoapp"
```
ealbizzati, line 21: Until the tap-to-return banner (blue banner) ships, the saved/revision refresh path in this PR is dormant: deployed but unreachable from the web. 

Restoring the return journey later is a one-line change (see `TODO` on `ArticleViewController+Editing.swift:95`).

tonisevener (approved):
Looks good, thanks!

PR #6071, opened 2026-08-14
T427676: Continue Reading

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T427676

### Notes
* The one thing I'm unsure of was the Obj-C to remove Continue Reading from Community, everything seems like it's working but as Obj-C I'd like extra scrutiny 

### Test Steps
(note, original Continue Reading in Explore required 24hrs+ so if you have that, test continue reading prior to fresh installing)
1. Fresh install, enable home/home phase 2
2. Do not open any articles, ensure For You doesn't show a continue reading or saved
3. Go through articles 60s+ +/ save articles
4. Make sure your saved articles are your most recent, no more than 3 show up
5. Make sure your continue reading article is one you've read in the past month (if there's only one it'll be that)
6. Make sure you do not see Continue Reading in Community if home is enabled
7. If you read / save / unsave articles and pull to refresh, ensure updates

### Checklist
- [x] The first card in this module is always the article the user most recently read for more than 1 minute
- [x] Avoid using a non-mainspace page for Continue reading
- [x] The first card displays: reason label ('Continue reading'), article title, card position, three-dot more menu, and a short text excerpt.
- [x] If the user has saved articles on their reading list, up to 3 additional cards are shown, one per saved article, most recently saved [irregardless of list]
- [x] Reading list cards display: reason label ('From your reading list')
- [x] The article's top image is used as a full-bleed background. If no image exists, a plain color background is shown.
- [x] Tapping any card navigates to the article
- [x] If the user has no reading history and no reading list items, this module is hidden from the feed.
- [x] If the user only has a continue reading article but no reading list items, the module shows just 1 card
- [x] The user updates their history or unsaves items from saved and then refreshes the feed, this module should also update to reflect those changes [checked with Haley, daily update or via refresh)
- [x] Remove the existing "Continue reading" module and associated settings from the legacy explore feed (Community) for users in the B group of the experiment. Users in Control should maintain access. (note, removed for anyone with home/home phase 2,
Nice to have
- [x] Upon open, scrolls to the user's last read position.
- [x] Ensure freshness: instead of just taking the last article read, for every new explore day, take all articles that were read for 1+ minutes in the last month and randomly pick one as the seed
- [x] Reusable components: None needed here
- [x] Common Strings
- [x] Using WMFSFSymbols

### Screenshots/Videos
Simulator Screenshot - iPhone 17 Pro Max liquid glass - 2026-08-14 at 12 09 25
Simulator Screenshot - iPhone 17 Pro Max liquid glass - 2026-08-14 at 12 09 19
Simulator Screenshot - iPhone 17 Pro Max liquid glass - 2026-08-14 at 12 09 22
Simulator Screenshot - iPhone 17 Pro Max liquid glass - 2026-08-14 at 12 09 28


mazevedofs (changes requested):
Some issues I found:

- Still need to remove the settings item from the explore tab if the card is present in the for you feed.
- The methods  `assignRelatedPageCardSlots` and `assignRandomArticleCardSlots` are unused, replaced by `assignCardSlots`
- Saved articles don't respect feed language:
In the print, my feed is in PT and the only article in my saved is in EN, it shows anyway.

Image

Wikipedia/Code/WMFContinueReadingContentSource.m
```diff
@@ -34,18 +35,36 @@ - (void)stopUpdating {
...
 
     [moc performBlock:^{
-        
+        // Check Home flags before the lastRead guard so that stale Continue Reading
+        // groups are removed even when the user has no reading history.
+        // Read through WMFDeveloperSettingsDataController so the JSON-encoded Data
+        // values stored by WMFUserDefaultsStore are decoded correctly — NSUserDefaults
+        // boolForKey: returns NO for those.
+        WMFDeveloperSettingsDataController *devSettings = WMFDeveloperSettingsDataController.shared;
+        if (devSettings.enableHomeTab || devSettings.enableHomePhase2) {
```
mazevedofs: No need to check devSettings.enableHomePhase2, enableHomeTab is the gate here. Enabling just enableHomePhase2 has no effect

WMFData/Sources/WMFData/Data Controllers/Home/WMFHomeDataController.swift
```diff
@@ -390,38 +390,23 @@ public final actor WMFHomeDataController {
...
-        let withoutThumbnail = articles.filter { $0.thumbnailURL == nil }
-
-        var imageQueue = withThumbnail.makeIterator()
-        var textQueue = withoutThumbnail.makeIterator()
-
-        return (0..<4).compactMap { _ in imageQueue.next() ?? textQueue.next() }
+        let saved = try await savedArticlesDataController.fetchRecentlySavedArticles(limit: 3)
+        let fromReadingList = saved.compactMap { item -> WMFForYouArticle? in
+            guard let itemProject = WMFProject(id: item.page.projectID),
+                  itemProject.languageCode == project.languageCode else { return nil }
```
mazevedofs, line 398: You're limiting fetching to three before knowing if you have enough viable candidates

Let's say I saved articles in this order = FR, EN, EN, EN, EN, FR, FR

You're fetching only the top three, and then filtering by language.

If your feed in is FR, you get one. If your feed is E, you get two. We could be presentig more articles to the user.

You can either add a language predicate to your fetching to make sure you retrieve the most recent three for a given language.

mazevedofs (changes requested):

WMFData/Sources/WMFData/Data Controllers/Home/WMFHomeDataController.swift
```diff
@@ -390,38 +392,23 @@ public final actor WMFHomeDataController {
...
+        }
+        if let seed = pages.randomElement() {
+            return WMFForYouContinueReading(
+                continueReadingArticle: WMFForYouArticle(title: seed.title, project: project),
+                fromReadingListArticles: fromReadingList
+            )
+        } else {
+            return WMFForYouContinueReading(
+                continueReadingArticle: nil,
+                fromReadingListArticles: fromReadingList
```
mazevedofs, line 409: Here you allow returning nil for continue reading, but in this line https://github.com/wikimedia/wikipedia-ios/pull/6071/changes#diff-3f368dde8d66186f0999d40116cc5f4449f1380d927676929ec0b5679ad6174aR172 you gate the module on having a valid continue reading article, so this else statement is not really used. 
Not sure if the requirements allow for no continue reading + only saved pages. If it does, you need to change that gate. Otherwise, you can just return nil

mazevedofs (changes requested):

WMFData/Sources/WMFData/Data Controllers/Home/WMFHomeDataController.swift
l-olson1214:
```
if let continueReadingArticle = continueReading.continueReadingArticle {
            let continueHeader = WMFForYouHeaderLabel(
                symbol: .newspaper,
                format: CommonStrings.continueReadingTitle,
                highlight: continueReadingArticle.title.normalizedForDisplay
            )
            cards.append(WMFForYouArticleCardViewModel(article: continueReadingArticle, headerLabel: continueHeader))
            deduplicator.markUsed(continueReadingArticle)
        }

        let savedFormat = WMFLocalizedString("for-you-header-saved-article", value: "From your reading list", comment: "Header on a For You feed card showing an article from the user's reading list.")
        let savedCards = deduplicator.removingDuplicates(from: continueReading.fromReadingListArticles).map { article in
            let header = WMFForYouHeaderLabel(symbol: .bookmarkFill, format: savedFormat, highlight: article.title.normalizedForDisplay)
            return WMFForYouArticleCardViewModel(article: article, headerLabel: header)
        }
        cards.append(contentsOf: savedCards)
```


I don't return immediately, and I grab the saved ones and append them after, so they're adding even if there's no continue reading

mazevedofs (approved):

mazevedofs (approved):

PR #6073, opened 2026-08-14
Allow scrolling to top on Home feed

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T427578

### Notes
* Add a scroll-to-top request to each Home feed, to use with W and Home tab buttons
Hide the W tooltip when the Home tab is on

### Test Steps

1. In the For You segment swipe up through 2–3 modules.
2. Tap the W or Home tab button. -> jumps back to the first module.
3. Switch to Community. ->  Community is where you left it, not scrolled to top.
4. Do the same for Community. For you shouldn't change.

### Checklist


WMFComponents/Sources/WMFComponents/Components/Home/WMFHomeViewModel.swift
```diff
@@ -76,6 +76,18 @@ public final class WMFHomeViewModel: ObservableObject {
...
 
+    @Published public private(set) var forYouScrollToTopRequestID: Int = 0
+    @Published public private(set) var communityScrollToTopRequestID: Int = 0
+
+    public func scrollSelectedFeedToTop() {
+        switch selectedTab {
+        case .forYou:
+            forYouScrollToTopRequestID += 1
+        case .community:
+            communityScrollToTopRequestID += 1
```
l-olson1214, line 87: A few curiosities here - I was unable to get a breakpoint to hit on lines 86/87/and I added a print to 88 to test

I'm never actually hitting community, and when I'm in community and I press the W, it's not actually calling scrollSelectedFeedToTop, not 100% sure what's going on but will keep investigating. It works, it's just weird 

Also, with the += 1, will we ever have an issue with it getting to be really large numbers, or why would we do += 1 as opposed to 1 vs 0 or something? Thanks!

l-olson1214 (changes requested):

WMFComponents/Sources/WMFComponents/Components/Home/WMFHomeViewModel.swift
mazevedofs: 1 - Line 87 hits with the new community feed, the behavior you see for the old Explore is the pre-existing behavior for the old Explore feed, which we maintained. It goes through `ExploreViewController.scrollToTop()` via `HomeViewController.scrollSelectedFeedToTop()`. 

Screenshot 2026-08-17 at 13 43 58

2 - Swift Int is [64-bit on iPhones](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/thebasics/#Int), which means a user can tap each one 9,223,372,036,854,775,807 (2⁶³ − 1) times before it overflows. This is an in-memory state that will reset when the app relaunches.

l-olson1214 (approved):

PR #6022, opened 2026-07-21
T432495: Add one-time new install onboarding start event for legacy onboarding

ealbizzati:
**Phabricator:** 
T432495

### Notes
- Add one-time new install onboarding start event for legacy onboarding

### Test Steps
1. delete the app install
2. reinstall the app and run
3. check debugger for: `new_install_onboarding_start`

then rerun the app:
1. check debugger for: `new_install_onboarding_start` (this time it should NOT show up)

### Checklist
- [x] Checked against Swift 6 strict concurrency

### Screenshots/Videos
``` 
🧑‍🍳TestKitchen: Scheduling event to be sent to /analytics/product_metrics/app/base/2.0.0:

$schema: /analytics/product_metrics/app/base/2.0.0
action: app_open
action_source: new_install_onboarding_start
agent: {
    "app_install_id" = "DC4E9546-5593-4802-856B-F4D307ACA50F";
    "app_theme" = default;
    "app_version" = 0;
    "app_version_name" = "WikipediaApp/8.3.0.0";
    "client_platform" = ios;
    "client_platform_family" = app;
    "device_family" = Phone;
    "device_language" = "en-CA";
    "release_status" = debug;
}
dt: 2026-07-21T19:43:37.319Z
instrument_name: apps-authentication
mediawiki: {
    database = enwiki;
}
meta: {
    stream = "product_metrics.app_base";
}
performer: {
    "is_logged_in" = 0;
    "is_temp" = 0;
    "language_groups" = en;
    "language_primary" = en;
    "session_id" = 6c2516adaca6715292c3;
}
```

PR #6076, opened 2026-08-18
T427688: Home A/B Test

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T427688

### Notes
* Analytics to be done in a later ticket

### Test Steps
//To test updating from old://
Fresh install on MAIN branch (or prior version) OR have prior version installed 
Not deleting old version, update to new one (run new branch)
You will be assigned either control or groupB
If you are assigned control:
- Make sure you see explore as usual, no new onboarding
If you are assigned groupB: 
- Make sure you see the new one time onboarding and are able to go through it
- Make sure you don’t see it on subsequent installs 
- Make sure you see Community and For You

//To test fresh install://
New device/fresh install (fully delete any old versions)
If you're assigned control:
- Make sure you go through the legacy onboarding 
- Make sure you see explore
- Make sure you don't see for you/ community
If you're assigned group B
- Make sure you see the NEW onboarding 
- Make sure you do NOT see the one time onboarding 
- Make sure you see for you/ community / etc

### Checklist
- [x] The release with the Explore Feed revamp should only be made availalbe to 50% of iOS app users
A: Control (Explore tab, no changes to feed)
- [x]  Existing onboarding
B: Variant with redesigned feed
- [x] Renamed to Home
- [x] Updated onboarding with Interest selection for new installs and existing app users
- [x] “For you” feed with new language selector
- [x] Additional settings for "For you"

- [x] Existing Explore Feed wrapped into “Community”, With any modules that have been added to "For you" removed
- [x] Audience: All languages, logged-in and logged-out
- [x] Test should have no end-date. We will create a scaling task to end the test.

### Screenshots/Videos

from updating as old version:
groupB
https://github.com/user-attachments/assets/05c69aa6-3b02-4403-8bf7-cdda43beb99f

control
https://github.com/user-attachments/assets/380a4ff9-ba96-45b9-863d-1a767dc3f681

fresh install:
groupB
https://github.com/user-attachments/assets/3e522a69-f8eb-46b0-9ae0-66cb5d6521db

control
https://github.com/user-attachments/assets/908bebec-9675-47e5-92f2-5103cbecb5b2

more fresh install
https://github.com/user-attachments/assets/f449716a-06d6-4b96-a38f-fcc28cea2774
https://github.com/user-attachments/assets/725f00e1-44f4-4ffc-a378-6613d967b260


Wikipedia/Code/WMFAppViewController.swift
```diff
@@ -935,7 +936,7 @@ final class WMFAppViewController: UITabBarController, AppTabBarDelegate {
         // default can't set this flag directly — write it through the data controller instead.
         // The flag persists across launches, so apply the argument in both directions.
         if UserDefaults.standard.object(forKey: wmfEnableHomeTabForTesting) != nil {
-            WMFDeveloperSettingsDataController.shared.enableHomeTab = UserDefaults.standard.bool(forKey: wmfEnableHomeTabForTesting)
+            WMFDeveloperSettingsDataController.shared.enableHomePhase2 = UserDefaults.standard.bool(forKey: wmfEnableHomeTabForTesting)
```
l-olson1214, line 939: Unsure on this - to ask

WMFData/Sources/WMFData/Data Controllers/Developer Settings/WMFDeveloperSettingsDataController.swift
```diff
@@ -109,15 +97,11 @@ public protocol WMFDeveloperSettingsDataControlling: AnyObject {
     /// True while the legacy Explore feed backs the Home tab's Community segment (home tab on, phase 2
     /// off). In this mode the feed is presented as the "Community feed" throughout the UI.
     public var isCommunityFeedMode: Bool {
-        enableHomeTab && !enableHomePhase2
+        !enableHomePhase2
```
l-olson1214: Double check on this

WMFData/Sources/WMFData/Data Controllers/Home/WMFHomeDataController.swift
```diff
@@ -22,15 +32,71 @@ public final actor WMFHomeDataController {
     // Dates for which feed data has been fetched per project, in descending order (most recent first).
     private var communityFetchedDates: [WMFProject: [Date]] = [:]
 
-    public static let shared = WMFHomeDataController()
+    @objc public static let shared = WMFHomeDataController()
+
+    nonisolated(unsafe) private let experimentsDataController: WMFExperimentsDataController?
```
l-olson1214: To ask / look into

l-olson1214: [I believe, according to what I found] Unsafe - it's only in the init, which is how unsafe is supposed to be used - Swift guarantees init will be finished first

WMFData/Sources/WMFData/Data Controllers/Home/WMFHomeDataController.swift
```diff
@@ -1,6 +1,16 @@
 import Foundation
 
-public final actor WMFHomeDataController {
+@objc public final actor WMFHomeDataController {
+    
+    public enum CustomError: Error {
+        case missingExperimentsDataController
+        case unexpectedAssignment
+    }
```
tonisevener: Looks like this isn't used.

Wikipedia/Code/WMFAppViewController.swift
```diff
@@ -1102,16 +1107,13 @@ final class WMFAppViewController: UITabBarController, AppTabBarDelegate {
...
+        guard WMFHomeDataController.shared.isHomeTabGroupB else { return }
 
-        // An existing user is one who has already completed ANY prior onboarding flow.
-        // New installs will not have this key set yet; they see full app onboarding instead.
         let isExistingUser = UserDefaults.standard.bool(forKey: Self.wmfDidShowOnboarding)
-        let hasSeen = WMFHomeDataController.shared.hasSeenNewHomeOnboarding()
-        let alwaysShow = WMFDeveloperSettingsDataController.shared.alwaysShowNewOnboarding
-        guard WMFDeveloperSettingsDataController.shared.enableHomeTab || WMFDeveloperSettingsDataController.shared.enableHomePhase2 else { return }
-        guard isExistingUser && (alwaysShow || !hasSeen) else { return }
+        let hasSeenNewOnboarding = WMFHomeDataController.shared.didSendNewInstallOnboardingStartEvent()
```
tonisevener: Let's discuss this one further, I recall not wanting to use it at all in my last review. https://github.com/wikimedia/wikipedia-ios/pull/6053#discussion_r3762698026

Wikipedia/Code/WMFAppViewController.swift
```diff
@@ -1197,6 +1199,7 @@ final class WMFAppViewController: UITabBarController, AppTabBarDelegate {
         DispatchQueue.main.async {
             self.present(onboardingVC, animated: true)
         }
+        WMFHomeDataController.shared.setHasSeenOneTimeOnboarding(true)
```
tonisevener: Can you move this up into the completion block of the presentation above instead?

```
        DispatchQueue.main.async {
            self.present(onboardingVC, animated: true) {
                WMFHomeDataController.shared.setHasSeenOneTimeOnboarding(true)
            }
        }
```

The reason:

If for some reason UIKit decides it can't present the one-time onboarding (like if it is already presenting a modal...very possible because we have so many modals that fire in an unstructured manner), the system will fail to present this one-time onboarding and it will NOT call the present completion block. So by setting this flag in the completion block, you know for sure that it has actually successfully displayed the one-time onboarding.

If the system decides it can't present, it will not call the completion block, the flag will not be set, and it will attempt again on the next app launch. That is the behavior we want in case something goes awry with this presentation attempt. (This is what I had to do to fix a lot of the reading challenge presentation issues).

Wikipedia/Code/WMFAppViewController.swift
tonisevener: I still don't think we should use this flag (reminder of my comment in https://github.com/wikimedia/wikipedia-ios/pull/6053#discussion_r3762698026). It was created in https://github.com/wikimedia/wikipedia-ios/pull/6022 and was only ever intended for sending a logging event (thus the name). I don't think we should mix purposes here.

If you need a flag to check if they have seen app onboarding yet, maybe you could create a new one? Or lean on the legacy `UserDefaults.standard.set(NSNumber(value: true), forKey: Self.wmfDidShowOnboarding)`, which `isExistingUser` may already cover.

tonisevener: One bug I am seeing:

1. Go to main branch, fresh install. You should see legacy app onboarding. Tap through it.
2. Switch back to this branch. Run again (this simulates an app version upgrade). IF you are assigned groupB, you should see one-time onboarding, but you do not.

I suspect it's this line, referencing `didSendNewInstallOnboardingStartEvent`. It is set in the legacy onboarding (WMFWelcomeInitialViewController). I still don't think we should lean on this flag, it was originally intended for sending an event / instrumentation in the legacy onboarding (see https://github.com/wikimedia/wikipedia-ios/pull/6022). I think using it for the purposes of displaying this new onboarding is causing the logic to get muddy.

This is the bug I mentioned in https://github.com/wikimedia/wikipedia-ios/pull/6053#discussion_r3762698026.

WMFData/Sources/WMFData/Data Controllers/Developer Settings/WMFDeveloperSettingsDataController.swift
```diff
@@ -4,7 +4,6 @@ public protocol WMFDeveloperSettingsDataControlling: AnyObject {
     func loadFeatureConfig() -> WMFFeatureConfigResponse?
     var enableMoreDynamicTabsV2GroupC: Bool { get }
     var forceMaxArticleTabsTo5: Bool { get }
-    var enableHomeTab: Bool { get }
```
tonisevener, line 7: I'm glad you removed the feature flag for this! Keeping it would have made it so much harder to test and reason about.

Wikipedia/Code/WMFContinueReadingContentSource.m
```diff
@@ -40,8 +40,7 @@ - (void)loadNewContentInManagedObjectContext:(NSManagedObjectContext *)moc force
         // Read through WMFDeveloperSettingsDataController so the JSON-encoded Data
         // values stored by WMFUserDefaultsStore are decoded correctly — NSUserDefaults
         // boolForKey: returns NO for those.
-        WMFDeveloperSettingsDataController *devSettings = WMFDeveloperSettingsDataController.shared;
-        if (devSettings.enableHomeTab) {
+        if (WMFHomeDataController.shared.isHomeTabGroupB) {
```
tonisevener, line 43: This looks good, just curious about this Phab task requirement for the group B assignment:

> Existing Explore Feed wrapped into “Community”, With any modules that have been added to "For you" removed

Does that only mean this continue reading card, or do they also want the "Because you read" card (AKA related pages card) removed from the Community tab? I get a bit mixed up on how to trigger both of those, but it makes me think `WMFRelatedPagesContentSource` might need a similar chunk of code as this, removing the content group.

tonisevener (changes requested):
Very close and exciting! These are some things I noticed, lmk what you think.

Wikipedia/Code/WMFContinueReadingContentSource.m
l-olson1214: Good catch! Added to the BYR one, too

We'll have to keep an eye out as we continue adding things to make sure we gate them off

Wikipedia/Code/WMFAppViewController.swift
l-olson1214: Okay sweet, updated that to a different one

Thought I had been able to get the logic, but the naming kept messing me up 😅 
simulator_screenshot_E930F8D1-2B48-42FD-9250-BF648B86BC11


https://github.com/user-attachments/assets/c2df85d6-4ff9-4b46-81ab-ff319f159075
Screenshot 2026-08-21 at 12 48 32 PM

Wikipedia/Code/WMFAppViewController.swift
tonisevener: Close, right now you have it as:

```
DispatchQueue.main.async {
      self.present(onboardingVC, animated: true)
      WMFHomeDataController.shared.setHasSeenOneTimeOnboarding(true)
}
```

It needs to be:

```
DispatchQueue.main.async {
      self.present(onboardingVC, animated: true) {
              WMFHomeDataController.shared.setHasSeenOneTimeOnboarding(true)
      }
}
```

So that the flag isn't set until the presentation completion block fires.

WMFData/Sources/WMFData/Store/WMFUserDefaultsKey.swift
```diff
@@ -27,6 +25,8 @@ public enum WMFUserDefaultsKey: String {
     case developerSettingsShowYiR2025 = "dev-settings-yir-show-v3"
     case developerSettingsYiRV3LoginExperimentControl = "dev-settings-yir-login-experiment-control"
     case developerSettingsYiRV3LoginExperimentB = "dev-settings-yir-login-experiment-b"
+    case developerSettingsHomeTabExperimentControl = "dev-settings-home-tab-experiment-control"
+    case developerSettingsHomeTabExperimentGroupB = "dev-settings-home-tab-experiment-group-b"
```
tonisevener: Found a couple more bits of code that are unused, can you delete?

Wikipedia/Code/WMFAppViewController.swift
l-olson1214: 🤦 Dang! My bad, pushed up a fix, sorry about that

Closures

Wikipedia/Code/WMFAppViewController.swift
```diff
@@ -1102,16 +1107,15 @@ final class WMFAppViewController: UITabBarController, AppTabBarDelegate {
...
 
-        // An existing user is one who has already completed ANY prior onboarding flow.
-        // New installs will not have this key set yet; they see full app onboarding instead.
         let isExistingUser = UserDefaults.standard.bool(forKey: Self.wmfDidShowOnboarding)
-        let hasSeen = WMFHomeDataController.shared.hasSeenNewHomeOnboarding()
-        let alwaysShow = WMFDeveloperSettingsDataController.shared.alwaysShowNewOnboarding
-        guard WMFDeveloperSettingsDataController.shared.enableHomeTab || WMFDeveloperSettingsDataController.shared.enableHomePhase2 else { return }
-        guard isExistingUser && (alwaysShow || !hasSeen) else { return }
+        // New user default, rely on old one initially (above) - check did see onboarding but NOT new onboarding, make sure they haven't seen one time onboarding yet
+        let hasSeenLegacyOnboarding = WMFHomeDataController.shared.hasSeenLegacyOnboarding()
```
tonisevener: I actually think you don't need `hasSeenLegacyOnboarding`. You are never setting it (it is always false, `setHasSeenLegacyOnboarding` is never called) yet it seems to be testing well for me. I think this is because your A/B test guard above (`isHomeTabGroupB`) plus checking the legacy `Self.wmfDidShowOnboarding` is enough of a check.

So can you remove the mentions of `hasSeenLegacyOnboarding` here and the `hasSeenLegacyOnboarding` and `setHasSeenLegacyOnboarding` home controller methods?

tonisevener (approved):

PR #6079, opened 2026-08-18
Enhance Phabricator workflow to handle PR edits and track new task links

josephseddon:
**Phabricator:** 

### Notes
* Extended the GitHub Actions workflow to trigger on PR edits in addition to opens and closes
* Added logic to detect when a PR body edit introduces new Phabricator task links
* Only posts notifications for newly-linked tasks on edit events, avoiding duplicate notifications for already-linked tasks
* Improved task ID extraction with a reusable `task_ids()` function that handles both "Bug: T###" trailers and Phabricator URL formats
* Enhanced environment variable handling and improved shell script robustness (proper quoting, null checks, set operations)

### Test Steps
1. Open a PR without any task references
2. Edit the PR body to add a task reference (e.g., "Bug: T12345")
3. Verify that a comment is posted to the task linking the PR
4. Edit the PR body again without changing task references
5. Verify that no new comment is posted
6. Edit the PR body to add an additional task reference
7. Verify that only the newly-added task receives a notification

### Checklist

https://claude.ai/code/session_01XyimsJL4PymoNhpfWeQ5DG

### DIFF
--- a/.github/workflows/post_phab.yml
+++ b/.github/workflows/post_phab.yml
@@ -18,9 +18,12 @@ jobs:
         BODY_CHANGES: ${{ toJSON(github.event.changes.body) }}
         PHAB_API_TOKEN: ${{ secrets.PHAB_BOT_API_KEY }}
       run: |
-        # Task IDs referenced at the start of a line, either as a "Bug: T123" trailer
-        # or as a phabricator.wikimedia.org URL (optionally behind the "**Phabricator:**"
-        # label from the PR template).
+        # Task IDs referenced at the start of a line, in either shape:
+        #   - a "Bug:" or "Phabricator:" label, optionally markdown-bolded, followed by
+        #     a bare T123 or a phabricator.wikimedia.org URL
+        #   - a bare phabricator.wikimedia.org URL with no label
+        # A bare "T123" with neither a label nor a URL deliberately does NOT match --
+        # that would be too loose to be safe in prose.
         #
         # The match is case-insensitive, so IDs are upper-cased before `sort -u` and
         # before the added-task diff below -- otherwise "T123" and "t123" count as two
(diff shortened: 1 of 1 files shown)


.github/workflows/post_phab.yml
```diff
@@ -2,24 +2,69 @@ name: Post to Phabricator
...
-          echo "Processing: $line"
+        # Task IDs referenced at the start of a line, either as a "Bug: T123" trailer
+        # or as a phabricator.wikimedia.org URL (optionally behind the "**Phabricator:**"
+        # label from the PR template).
+        task_ids() {
+          printf '%s' "$1" \
+            | tr -d '\r' \
+            | grep -oEi "(^Bug:[[:space:]]*T[0-9]+)|(^([*]*phabricator[*]*:[*]*[[:space:]]*)?https://phabricator\.wikimedia\.org/T[0-9]+)" \
+            | grep -oEi "T[0-9]+" \
+            | sort -u
```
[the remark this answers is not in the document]
josephseddon: Applied in f7b4071d, with one correction to the premise: this step does **not** run with `pipefail` today.

`-o pipefail` is only added when a step sets `shell: bash` explicitly. This step sets no `shell:` key, so it gets the default `bash -e {0}`. The run of this very PR confirms it:

```
shell: /usr/bin/bash -e {0}
```

That run also exercised the case described as terminating the step — `opened` with no task reference in the body — and it completed successfully in 3s. Under `bash -e` the pipeline's status comes from `sort`, not `grep`, so the assignment already yielded an empty set.

Reproducing both shells locally against the extracted step:

| | task-less body | edit adding first task |
|---|---|---|
| `bash -e` (actual) | exit 0 | exit 0, posts T123 |
| `bash -eo pipefail` | exit 1 | exit 1, posts nothing |

So the failure is real but latent — one innocuous-looking `shell: bash` away. Worth guarding regardless, so the pipeline now ends `| sort -u || true` with a comment recording why it isn't redundant. Verified correct under both shells afterwards.

---
_Generated by [Claude Code](https://claude.ai/code)_

.github/workflows/post_phab.yml
```diff
@@ -2,24 +2,80 @@ name: Post to Phabricator
...
-    types: [opened, closed]
+    types: [opened, edited, closed]
 
 jobs:
   post_to_phab:
     runs-on: ubuntu-latest
     steps:
-    - name: Post to Phabricator when pull request is opened or closed
-      if: ${{ github.event_name == 'pull_request' && (github.event.action == 'opened' || github.event.action == 'closed') }}
+    - name: Post to Phabricator when pull request is opened, linked, or closed
```
mazevedofs, line 11: Question: does it mean if we edit a PR description N times, it'll post N updates to the task? 
Usually not an issue, just curiosity.

josephseddon: What I asked from it is to post only if a ticket is added. Most of the time you'll only add one ticket so you should only generate one additional update. If you add a second ticket link or replace it in a subsequent edit then that should in theory generate an additional comment. If you kept adding tickets to every update of the description then in theory it I it could be N times 

If ends up being noisy, I'll make some additional tweaks.

PR #6081, opened 2026-08-19
T435382: Fix crash when tapping a paragraph moved marker in revision diff

ealbizzati:
**Phabricator:** https://phabricator.wikimedia.org/T435382

### Notes
* Since the July 2024 diff redesign, list items live in section 2 of the diff collection view — sections 0 and 1 contain no items and exist only to host the two header supplementary views (a flow layout allows one header per section). `didTapItem` predates that change and still built its scroll target with `section: 0`, so tapping a "paragraph moved" marker always raised `NSInternalInconsistencyException` in `scrollToItem` and crashed the app.
* This introduces a `Section` enum and uses it everywhere the section layout is assumed. Besides fixing the crash, it restores the visible-cell shortcut and the fine-grained `scrollToChangeItem` offset, which also depended on the correct section.

### Test Steps
1. Open a diff containing a moved paragraph, e.g. `https://en.wikipedia.org/wiki/Special:MobileDiff/1369045204...1370193452` (via `xcrun simctl openurl booted`, or through article history / watchlist)
2. Tap a "Paragraph moved down/up" marker
3. Before: crash. After: the list scrolls to the other half of the move and centers it

### Checklist

### Screenshots/Videos
N/A — crash fix, no visual changes.

### DIFF
--- a/Wikipedia/Code/DiffListViewController.swift
+++ b/Wikipedia/Code/DiffListViewController.swift
@@ -7,6 +7,12 @@ class DiffListViewController: ThemeableViewController {
     
     fileprivate static let headerReuseIdentifier = "DiffHeaderView"
     fileprivate static let headerExtendedReuseIdentifier = "DiffHeaderExtendedView"
+
+    enum Section: Int, CaseIterable {
+        case header
+        case extendedHeader
+        case items
+    }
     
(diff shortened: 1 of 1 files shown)


mazevedofs (approved):
Looks great!

PR #6082, opened 2026-08-19
T431043: Topics Localizations

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T431043

### Notes
* run `swift scripts/GenerateTopicsStrings.swift` or 
* `swift scripts/GenerateTopicsStrings.swift [languagecode] [--stdout]`
* languageCode will specify for one language, and --stdout will allow to print the changes that will be made prior to making them

### Test Steps
1. Look through localizations, ensure at least the main languages have topics localized
2. Ensure no other localizations have been altered
3. Be amazed, ooooh, ahhhh

### Checklist
- [x] Checked against Swift 6 strict concurrency [script - does not conform/apply to actors/etc]

### Screenshots/Videos
Screenshot 2026-08-19 at 1 57 19 PM
Screenshot 2026-08-19 at 1 57 37 PM
Screenshot 2026-08-19 at 1 58 00 PM
Screenshot 2026-08-19 at 1 56 28 PM
Screenshot 2026-08-19 at 1 56 59 PM


WMFLocalizations/GenerateTopicsStrings.swift
```diff
@@ -0,0 +1,387 @@
...
+]
+
+// MARK: - Language mapping
+//
+// Maps MediaWiki language codes to .lproj folder names where the two differ.
+// Mirrors the logic in WMFLocalizedString.swift so written keys land in the
+// bundle the runtime reads. Falls back to the MediaWiki code itself when no
+// entry is present.
+
+let mediawikiToLproj: [String: String] = [
```
mazevedofs, line 71: The mediawikiToLproj table gives the wrong folder names. The script
writes to the Wikipedia/Localizations directory. The folders in this
directory have MediaWiki language codes. The table gives app bundle
names. resolveTargetLanguages compares the two names character by
character. The comparison fails, and the script writes no topic strings. You can check the produced localizable.strings folders.

The script writes no topic strings for these three locales:
  - pt-br. The table gives pt-BR. The folder name is pt-br.
  - sr-ec. The table gives sr. The folder name is sr-ec.
  - ku-latn. The table gives ku. The folder name is ku-latn.
The files in the branch show the result. The pt-br, sr-EC, and ku-latn
bundles have 0 topic keys. The pt, sr-Latn, and uz bundles have 39
topic keys.

The folders in Wikipedia/Localizations always have
MediaWiki codes. Only the import script twnLocaleToAppStoreConnectLocale changes a code.

WMFLocalizations/GenerateTopicsStrings.swift
```diff
@@ -0,0 +1,387 @@
...
+//   "some-key" = "some value";
+// When a key matches one of our managed keys, record which line it's on
+// (including the optional comment line immediately above it) so we can
+// replace just that entry in place. Everything else is left untouched.
+
+func injectIntoStringsFile(at url: URL, translations: [String: String]) throws {
+    var encoding: String.Encoding = .utf8
+    let existing: String
+    if url.isFileURL, FileManager.default.fileExists(atPath: url.path) {
+        existing = (try? String(contentsOf: url, usedEncoding: &encoding)) ?? ""
```
mazevedofs, line 249: This line uses `try?` to read the file. A read failure gives an empty string. The function then finds no managed keys.
It puts all 39 topic entries in the result. Line 310 writes the result to the same file. The file then has only the 39 topic entries. 
To fix this change `try?` to `try` on line 249.

WMFLocalizations/GenerateTopicsStrings.swift
```diff
@@ -0,0 +1,387 @@
+#!/usr/bin/env swift
+
```
mazevedofs, line 2: he script wrote topic strings for 27 locales that the app cannot use.
These locales have a folder in Wikipedia/Localizations, but they have no folder in the Resources directory. The import script calls `localeIsAvailable` on line 422 of  `localization.swift` for each locale.
This function takes the characters before the hyphen, and it looks for
them in the locales set on line 393. The set has the Apple locale
identifiers and the names of the existing Resources folders. For these 27 locales, the function gives `false`, and the import script writes no app bundle.
  ```
bcl, be-tarask, blk, bto, bug-bugi, dga, dty, fit, frp, gom-latn,
  hyw, kaa, kjp, ksw, li, mad, mhr, nah, nap, olo, pnb, rki, rue,
  skr-arab, sms, sro, wuu-hans
```
One locale in this list fails for a different reason. The prefix of
be-tarask is be, and be is a correct Apple identifier. The
localesUnsupportedByAppStoreConnect set on line 414 blocks be-tarask.

The folders are not new, but most of the content is new. 

Before this branch, 20 of these folders had only an InfoPlist.strings file. The
script added a Localizable.strings file to these 20 folders. The other 7 folders already had this file. TranslateWiki keeps InfoPlist strings for more locales than the main string catalog.

Check https://github.com/wikimedia/wikipedia-ios/blob/main/Command%20Line%20Tools/Update%20Localizations/localization.swift for more info on how the app writes bundles.

File: WMFLocalizations/GenerateTopicsStrings.swift [hunk 4]
```diff
@@ -0,0 +1,387 @@
...
+// GenerateTopicStrings.swift
+//
+// Fetches localized article topic strings from the MediaWiki allmessages API
+// and injects them into the matching .lproj/Localizable.strings files inside
+// the WMFNativeLocalizations Resources directory.
+//
+// Run from anywhere inside the repo:
```
mazevedofs, line 9: The branch adds 23 new Localizable.stringsdict files to the Resources directory. Each file is a copy of the English file, and the 23 files have 33,258 lines. This is 64 percent of the lines in the branch. Each of these 23 locales receives 39 topic strings and 1,446 lines of English text. The import script makes these copies on purpose. Line 499 of `localization.swift` writes the English plural dictionary when a locale has no plural messages.
These 23 locales had no bundle in the app before this branch.
Each locale now shows 39 topic labels in its own language.
The app shows English text for all the other strings. 
We should think if it's worth adding so many repeated English strings to languages that had no translations before. It's likely they won't receive translations in the future, and only the topics will be translated.

These are the 23 locales:
```
  be, dsb, dz, en-gb, fy, gu, inh, ky, mg, mni, ms-arab, mt, rm, scn,
  shi, si, smn, szl, ti, tok, tt-cyrl, yue-hant, zgh
```

WMFLocalizations/GenerateTopicsStrings.swift
```diff
@@ -0,0 +1,387 @@
...
+            try FileManager.default.createDirectory(at: lprojURL, withIntermediateDirectories: true)
+            try injectIntoStringsFile(at: stringsURL, translations: translations)
+
+            print("Updated \(stringsURL.path) (\(translations.count)/\(msgKeys.count) keys)")
+
+            if offset < languages.count - 1 {
+                Thread.sleep(forTimeInterval: 0.5)
+            }
+        } catch {
+            fputs("Skipping '\(languageCode)': \(error.localizedDescription)\n", stderr)
```
mazevedofs, line 381: The loop on line 350 catches an error for each language, and line 381 writes the message. The loop then continues, and the script counts no failures. If all 159 languages fail, the script still exits with code 0. WE cannot see the difference between a complete run and an empty run. 
Just checking if this is by design/desired.

WMFLocalizations/GenerateTopicsStrings.swift
```diff
@@ -0,0 +1,387 @@
...
+            let lprojURL = resourcesURL.appendingPathComponent("\(folder).lproj")
+            let stringsURL = lprojURL.appendingPathComponent("Localizable.strings")
+
+            try FileManager.default.createDirectory(at: lprojURL, withIntermediateDirectories: true)
+            try injectIntoStringsFile(at: stringsURL, translations: translations)
+
+            print("Updated \(stringsURL.path) (\(translations.count)/\(msgKeys.count) keys)")
+
+            if offset < languages.count - 1 {
+                Thread.sleep(forTimeInterval: 0.5)
```
mazevedofs, line 378: The script does not wait after an error, so HTTP 429 response removes the delay.
The script then sends the next request immediately.
if the wait is desired, move it out of the do/catch block

WMFLocalizations/GenerateTopicsStrings.swift
```diff
@@ -0,0 +1,387 @@
...
+        } else if let data {
+            result = .success(data)
+        }
+        semaphore.signal()
+    }.resume()
+
+    semaphore.wait()
+
+    let data = try result.get()
+    return try JSONSerialization.jsonObject(with: data)
```
mazevedofs, line 166: Errors are being sent here as `data`, but the error code is never read down the road; custom errors are being used like `Unexpected allmessages response shape` instead of the API error message. You can combine both for a better idea of the error, especially if it's a rate-limiting error.
Or you can read the error before the guard in line 202 to handle it in a better way

WMFLocalizations/GenerateTopicsStrings.swift
```diff
@@ -0,0 +1,387 @@
...
+
+func fetchJSON(url: URL, queryItems: [URLQueryItem]) throws -> Any {
+    var components = URLComponents(url: url, resolvingAgainstBaseURL: false)
+    components?.queryItems = queryItems
+    guard let requestURL = components?.url else {
+        throw ScriptError("Could not build request URL.")
+    }
+
+    var result: Result<Data, Error> = .failure(ScriptError("Request never completed."))
+    let semaphore = DispatchSemaphore(value: 0)
```
mazevedofs, line 152: This file is just a script, but it actually introduces Swift 6 issues.

The semaphore blocks a thread while it waits. The closure on line 154 writes to the result variable from a different thread. 
We also get errors in lines 325 and 330 for Swift 6 issues. 
This pattern might work, and it's more modern (you might need to fold in the error handling from another comment):

```suggestion
    func fetchJSON(url: URL, queryItems: [URLQueryItem]) async throws -> Any {
    var components = URLComponents(url: url, resolvingAgainstBaseURL: false)
    components?.queryItems = queryItems
    guard let requestURL = components?.url else {
        throw ScriptError("Could not build request URL.")
    }
    let (data, response) = try await session.data(from: requestURL)
    if let http = response as? HTTPURLResponse, !(200..<300).contains(http.statusCode) {
        throw ScriptError("HTTP \(http.statusCode) for \(requestURL.absoluteString).")
    }
    return try JSONSerialization.jsonObject(with: data)
}
```

mazevedofs (changes requested):
Maybe we should delete 
https://github.com/wikimedia/wikipedia-ios/blob/main/WMFData/Sources/WMFData/Models/Shared/WMFArticleTopic.swift

mazevedofs (changes requested):
Also, can we move this script to `Command Line Tools/`

WMFLocalizations/GenerateTopicsStrings.swift
l-olson1214: We now exit on partial failure with 1!

WMFLocalizations/GenerateTopicsStrings.swift
l-olson1214: Removed! Thank you!!

mazevedofs (changes requested):
The only thing missing is removing
https://github.com/wikimedia/wikipedia-ios/blob/main/WMFData/Sources/WMFData/Models/Shared/WMFArticleTopic.swift

this is now dead code, but still generate strings that are not used anywhere, so it's good remove this file from the code and the keys from translate wiki.
let's add some red lines to this diff!

mazevedofs (approved):

PR #6084, opened 2026-08-19
T435346: Add Force Fundraising Campaign Banner developer settings [boiler plate work]

ealbizzati:
**Phabricator:** 
https://phabricator.wikimedia.org/T435346

* Boilerplate/tooling for the Donation Reminder work: testing the banner flows currently requires setting the simulator region, deleting the app, and reopening articles until the campaign config is cached.
* Adds a "Fundraising" section to Developer Settings with two things:
  * **Force Fundraising Campaign Banner**: presents the banner on any article view, ignoring country, date window, prompt state ("maybe later" / "already donated"), opt-out, and donation history gates. The production path is untouched — the flag branches once at the top of `loadActiveCampaignAsset` into a development-only method. Still requires an active campaign in the remote config (real asset, not mocked).
  * **Clear banner prompt state and donation history**: resets "maybe later" / "already donated" and the local donation history (250-day gate), so the real flow can be re-tested without reinstalling the app. Follows the `clearGamesPersistence()` pattern.

### Test Steps
1. Build and run, enable Developer Settings, toggle "Force Fundraising Campaign Banner".
2. Open any article (any language, any region) — the banner presents. Tap "I already donated"; open another article — banner presents again.
3. Toggle force off, tap "Clear banner prompt state and donation history" — with a matching campaign (region/language), the banner presents once more under production rules.

### Checklist

### Screenshots/Videos

https://github.com/user-attachments/assets/63b5a97c-0aaf-44c3-9d09-06f370e0adcb

### DIFF
--- a/WMFComponents/Sources/WMFComponents/Components/Developer Settings/WMFDeveloperSettingsView.swift
+++ b/WMFComponents/Sources/WMFComponents/Components/Developer Settings/WMFDeveloperSettingsView.swift
@@ -35,6 +35,14 @@ struct WMFDeveloperSettingsView: View {
                 }
             }
 
+            Section {
+                Toggle("Force Fundraising Campaign Banner", isOn: $viewModel.forceFundraisingCampaignBanner)
+            } header: {
+                Text("Fundraising")
+            } footer: {
+                Text("Force ignores country and language settings. Only works if there is an active campaign.")
+            }
+
             ForEach(viewModel.formViewModel.sections) { section in
                 if let selectSection = section as? WMFFormSectionSelectViewModel {
                     WMFFormSectionSelectView(viewModel: selectSection)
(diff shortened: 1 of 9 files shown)


tonisevener (approved):
Looks good, this has really been needed for a while.

I do have one suggestion - currently, the fundraising config doesn't even fetch for users that have fresh installed. See the guard in this area of code - these old content source files are called when the Explore feed is refreshed - https://github.com/wikimedia/wikipedia-ios/blob/29321fc78b8095550f5bc59974d6bead57e39bf6/Wikipedia/Code/WMFAnnouncementsContentSource.m#L59-L71

So right now in this branch, if you fresh install, then turn the force flag on, it still won't show. You have to background, foreground, pull to refresh Explore, for the config to fetch. Then the banner shows. Not a huge deal and I can still approve, but just one last little bit of annoyance that the force flag currently doesn't fix. It would be great if we could do something like this:

1. Move the fundraising config fetch out of WMFAnnouncementsContentSource entirely. We will soon move away from the legacy Explore feed code anyways, so that class's days are numbered. Maybe fetch fundraising configs somewhere in the launch sequence?
2. Instead add the wmf_appResignActiveDate guard check to ArticleViewController+Announcements.swift. There you can have the force developer flag ignore it.

tonisevener (approved):

PR #6093, opened 2026-08-21
Swift 6 migration — Phase 4: WMFData strict-concurrency burn-down complete + CI gate

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T389255

## Swift 6 migration — Phase 4: WMFData has zero strict-concurrency warnings

This PR is part of the Swift 6 migration. The migration changes the leaf
packages first. This PR completes the WMFData burn-down. The count of warnings was approximately 127 at the start. The count is now zero. The CI gate now monitors WMFData and WMFComponents. The gate stops each PR that adds a concurrency warning to these two packages.

### WMFData
- `WMFService` and `WMFURLSession` are now `Sendable` protocols. Their
  completions are `@Sendable`. This is necessary for two reasons. Data
  controllers hold service instances and use them from many tasks and queues.
  URLSession calls the completions from its delegate queue. The decodable
  service methods now specify `T: Decodable & Sendable`. The response models
  are now `Sendable`.
- `WMFLockIsolated` is a new utility. It is a box that holds mutable state
  behind a lock. Some classes must stay subclasses of `NSObject` because
  Objective-C code uses them. These classes cannot become actors. These
  controllers now keep all mutable state in `WMFLockIsolated` boxes: developer
  settings, donate, fundraising campaign, temp accounts, growth tasks, and
  watchlist. Each `@unchecked Sendable` conformance has a comment that gives
  the reason. The lock makes the conformance safe.
- This PR repairs two data races. The donate-config fetch and the watchlist
  fetch each use a `DispatchGroup` to do two or more requests at the same time.
  Before this PR, the callbacks wrote to shared local variables at the same
  time. Now the callbacks write to one lock-protected box.

### WMFComponents (related changes)
- The data controller callbacks are now `@Sendable`. They do not run on the
  main actor. The watchlist, donate, and image recommendations view models now
  move to the main actor in these callbacks. Then they change the view model
  state. The completion callbacks of these view models now have the type
  `@MainActor @Sendable`. The image recommendations fetch does not use
  `DispatchQueue.main.async` now. It uses `Task { @MainActor in }`.
- The WMFComponents warning count stays at zero.

### App code
- `MediaWikiFetcher` is now `@unchecked Sendable`. A comment gives the reason.
  The full examination of this class will occur in the WMF Framework phase.

### CI
- The Strict Concurrency Check workflow now includes WMFData in
  `GATED_SCHEMES`. The job fails if a PR adds a concurrency warning to WMFData
  or WMFComponents sources.

### Verification
- A clean build shows zero warnings for WMFData and zero warnings for
  WMFComponents. The count procedure is the same as the CI procedure. The
  environment was Xcode 26.6 and an iPhone 17 simulator with iOS 26.5.
- All 202 WMFData unit tests passed. All 258 WMFComponents unit tests passed.
  The Wikipedia app scheme compiled correctly.
- The script `scripts/lint-wmfdata-test-fixtures.sh` gave a pass result.

### Notes for reviewers
- The language modes did not change. The packages stay in mode `.v5`. In this
  mode, violations are warnings. The `.v6` change for WMFComponents stays on
  hold because of swiftlang/swift#87316.
- The next phase adds WMF Framework to the workflow matrix in report-only mode.
  Then its burn-down starts.

### DIFF
--- a/WMFData/Sources/WMFData/Data Controllers/Donor Experience/WMFDonateDataController.swift
+++ b/WMFData/Sources/WMFData/Data Controllers/Donor Experience/WMFDonateDataController.swift
@@ -1,15 +1,33 @@
 import Foundation
 import Contacts
 
-@objc final public class WMFDonateDataController: NSObject {
-    
+// @unchecked Sendable: must stay an NSObject subclass for Obj-C callers, so it
+// cannot be an actor. All mutable state lives in WMFLockIsolated boxes below.
+@objc final public class WMFDonateDataController: NSObject, @unchecked Sendable {
+
     // MARK: - Properties
-    
-    var service: WMFService?
-    var sharedCacheStore: WMFKeyValueStore?
-    
-    private var donateConfig: WMFDonateConfig?
-    private var paymentMethods: WMFPaymentMethods?
+
+    private let _service: WMFLockIsolated<WMFService?>
+    var service: WMFService? {
+        get { _service.value }
+        set { _service.value = newValue }
+    }
+    private let _sharedCacheStore: WMFLockIsolated<WMFKeyValueStore?>
+    var sharedCacheStore: WMFKeyValueStore? {
+        get { _sharedCacheStore.value }
+        set { _sharedCacheStore.value = newValue }
+    }
+
+    private let _donateConfig = WMFLockIsolated<WMFDonateConfig?>(nil)
+    private var donateConfig: WMFDonateConfig? {
+        get { _donateConfig.value }
+        set { _donateConfig.value = newValue }
+    }
+    private let _paymentMethods = WMFLockIsolated<WMFPaymentMethods?>(nil)
+    private var paymentMethods: WMFPaymentMethods? {
+        get { _paymentMethods.value }
+        set { _paymentMethods.value = newValue }
+    }
     
     private let cacheDirectoryName = WMFSharedCacheDirectoryNames.donorExperience.rawValue
     private let cacheDonateConfigContainerFileName = "AppsDonationConfig"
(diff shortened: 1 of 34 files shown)


WMFData/Sources/WMFData/Data Controllers/Donor Experience/WMFDonateDataController.swift
```diff
@@ -1,15 +1,33 @@
...
+    private let _donateConfig = WMFLockIsolated<WMFDonateConfig?>(nil)
+    private var donateConfig: WMFDonateConfig? {
+        get { _donateConfig.value }
+        set { _donateConfig.value = newValue }
+    }
+    private let _paymentMethods = WMFLockIsolated<WMFPaymentMethods?>(nil)
+    private var paymentMethods: WMFPaymentMethods? {
+        get { _paymentMethods.value }
+        set { _paymentMethods.value = newValue }
+    }
```
ealbizzati: low/zero risk today

`donateConfig` and `paymentMethods` live in two separate `WMFLockIsolated` boxes, so the pair is never read or written atomically.
every current `loadConfigs` caller is in the main thread, and main already serializes read/write.

Issue would only show up if we ever call loadConfigs off the main thread.

WMFData/Sources/WMFData/Utility/WMFLockIsolated.swift
```diff
@@ -0,0 +1,38 @@
+import Foundation
+
+/// A minimal lock-guarded box for mutable state on classes that must stay
+/// `NSObject`-based (Obj-C visible singletons) and therefore cannot become actors.
+/// Holding all mutable stored properties in `WMFLockIsolated` boxes is what makes
+/// those classes' `@unchecked Sendable` conformances sound — every read/write is
+/// serialized behind the lock. Keep accesses coarse: read once into a local,
+/// compute, then write back, rather than many fine-grained hops.
+final class WMFLockIsolated<Value>: @unchecked Sendable {
```
ealbizzati: should this have Value constraint with Sendable??
it's a zero risk today because the stores that use it are UserDefaults (which is thread safe) and file cache (which is stateless). WMFFundraisingCampaignConfig is value-only structure (just missing the annotation)

the fix is small but would require more files touched. we could do it in a follow up too.

ealbizzati (approved):
lgtm.
tested a few flows I thought could be impacted: 
donation flow, donation campaign banner, watchlist

PR #6106, opened 2026-08-24
T435348: Show the in-article donation reminder card (2/3)

ealbizzati:
**Phabricator:** 
https://phabricator.wikimedia.org/T435348

* **Reminder cycle (data layer):** article reads are derived from the existing `CDPageView` reading history — a read counts when it lasts 5+ seconds (1 second in DEBUG builds, matching Android). The cycle mirrors Android's `DonationReminderHelper`: max 2 impressions per cycle, at most one per day, "Not now" closes the current window, and a new window (with fresh impressions) opens every time the user reaches the article goal again. No new counters are stored — the cycle "resets" by moving the anchor date in the reminder's `Progress`.
* **In-article card:** injected into the article web view above the title, using the same technique as Article as a Living Document (#3721): HTML built natively, inserted via the messaging controller, styled from `styleoverrides.css` with per-theme selectors, taps routed back through anchor hrefs intercepted in `handleLink`. First milestone shows "You've read X articles since you pledged Y!"; later milestones show "You've read another X articles!" (selected via `goalReachedCount`, same as Android).
* "Yes, donate now" routes to the existing donate payment sheet via a new `DonateCoordinator` source; the pledge-specific payment options land in the next PR. Funnel logging for the new source is intentionally empty — instrumentation is T435547.
* Everything is gated behind the `enableDonationReminder` developer flag. A new "Bypass Reminder Daily Limit" developer setting lets QA see repeat impressions without changing the device date.

### Test Steps
1. Enable developer mode, then turn on **Enable Donation Reminder** and **Force Fundraising Campaign Banner** in Profile → Developer settings.
2. Create a pledge (goal: 5 articles): article → banner → Maybe Later → setup screen → Confirm. (Setup screen comes from the base branch stack.)
3. Read 5 articles for at least ~2 seconds each, then open another article: the reminder card appears above the title. Check light/dark themes.
4. Tap **Not now**: card is removed and a toast promises the next reminder. Read 5 more articles: the card returns with the "another" heading.
5. Tap **Yes, donate now**: the standard donate payment sheet appears.
6. Turn on **Bypass Reminder Daily Limit** to see repeat impressions on the same day; with it off, a second impression waits for the next calendar day.
7. Turn **Enable Donation Reminder** off: no card anywhere. 

### Checklist

### Screenshots/Videos

Simulator Screenshot - iPhone 17 - 2026-08-24 at 10 08 07
Simulator Screenshot - iPhone 17 - 2026-08-24 at 12 08 39
Simulator Screenshot - iPhone 17 - 2026-08-24 at 17 13 24


Wikipedia/Code/ArticleViewController+DonationReminder.swift
```diff
@@ -0,0 +1,142 @@
...
+    private func didTapDonationReminderNotNow() {
+        messagingController.removeDonationReminderCard()
+
+        guard let reminder = WMFDonationReminderDataController.shared.loadReminder(),
+              case .articlesRead(count: let articlesReadGoal) = reminder.trigger
+        else { return }
+
+        WMFDonationReminderDataController.shared.recordFollowUpReminderNotNow()
+
+        let toastFormat = WMFLocalizedString("donation-reminder-card-not-now-toast", value: "We will remind you again after you’ve read %1$@ articles.", comment: "Toast shown after the user dismisses the in-article donation reminder card. %1$@ is the number of articles until the next reminder.")
```
tonisevener: Try the translatewiki plural format for this copy:

```
let toastFormat = WMFLocalizedString("donation-reminder-card-not-now-toast", value: "We will remind you again after you’ve read {{PLURAL:%1$d|%1$d article|%1$d articles}}.", comment: "Toast shown after the user dismisses the in-article donation reminder card. %1$d is the number of articles until the next reminder.")
```

There are a few more instances in this PR that could use plural as well.

Wikipedia/Code/ArticleViewController+DonationReminder.swift
```diff
@@ -0,0 +1,142 @@
...
+
+        if href.hasSuffix("#wmf-donation-reminder-not-now") {
+            didTapDonationReminderNotNow()
+            return true
+        }
+
+        return false
+    }
+
+    private func didTapDonationReminderDonate() {
```
tonisevener, line 44: Should we flag something here so that they don't see this cycle's donation reminder again? Similar to `recordFollowUpReminderNotNow`. I notice I see the donation ask again on the next article view, even if I donate (will send a video).

Wikipedia/Code/ArticleViewController+DonationReminder.swift
ealbizzati: applied the plural format to all four strings with article counts

Wikipedia/Code/ArticleViewController+DonationReminder.swift
ealbizzati: done. tapping "donate now" updates `timesReminderShown`.  so the current cycle is done and the card doesn't come back until next milestone is hit (or following day - whichever happens last). We don't close the card on that tap, though, as it should stay there until the next view appearance.

Wikipedia/Code/ArticleViewController+DonationReminder.swift
tonisevener: Thanks - I think it might be a more flexible for translators to translate the entire string together rather than interpolated as two separate translations like you did. Sometimes these approaches come back with complaints, that is, "5 articles" may need to be written out differently depending on the contextual string it gets inserted into. To avoid that I think it's better to add the PLURAL piece directly to each message.

ealbizzati: That makes sense. I moved the plural syntax into each full string and dropped the shared frafgment.

tonisevener (approved):

PR #6110, opened 2026-08-25
T435343: Add developer setting to fetch donate configs from Test Wiki

ealbizzati:
**Phabricator:** 
https://phabricator.wikimedia.org/T435343

### Notes
* Adds a "Use Test Wiki Donate Configs" developer setting that fetches the donate and fundraising campaign configs from test.wikipedia.org instead of donate.wikimedia.org (@tonisevener's suggestion, replacing the rarely used Staging scheme for campaign testing). The URL builders already supported the staging host: the toggle just routes these two fetches through it.
* Scoped to the two config fetches only; everything else stays on the session's service environment.

### Test Steps
1. Enable developer mode → Developer settings → turn on **Use Test Wiki Donate Configs**.
2. Background and foreground the app to refetch, then open an article: the campaign banner reflects `MediaWiki:AppsCampaignConfig.json` from test.wikipedia.org (publish a test campaign there to see it).
3. Toggle off and refetch: production configs again.

### Checklist

### Screenshots/Videos
n/a


WMFData/Sources/WMFData/Data Controllers/Donor Experience/WMFFundraisingCampaignDataController.swift
```diff
@@ -143,7 +143,8 @@ import Foundation
             return
         }
         
-        guard let url = URL.fundraisingCampaignConfigURL() else {
+        guard let url = URL.fundraisingCampaignConfigURL(environment: WMFDeveloperSettingsDataController.shared.donateConfigsServiceEnvironment)
```
tonisevener, line 146: It doesn't look like the test config is fetching for me after I toggle this, I think it's a caching issue.

It looks like the MediaWiki:AppsCampaignConfig is cached for a few hours and it doesn't seem like there's a way around it for someone testing. See `checkRemoteAppConfigIfNecessary` in WMFAppViewController as the entry point to fetching that config.

It makes this toggle a little less useful, because it initially fetches the prod config, then a tester would need to wait 3 hours before their test toggle will take effect.

ealbizzati: yeah..  I see what's happening. it's the 3 hour throttling or something. I'll think about this for a little while more.

tonisevener (approved):

PR #6115, opened 2026-08-26
Fix UI tests

tonisevener:
**Phabricator:** N/A

### Notes
This might fix our UI tests - a lot of them assumed the old Explore feed, and that assumption was incorrect once we merged the Home A/B test. For this I am hooking into the existing `wmfEnableHomeTabForTesting` flag and am forcing the experiment assignment depending on the result.

I am also updating the UI testing GitHub actions to allow us to run against any tag. The old structure was a little more rigid in that it would run only against `releases/...` tags. Now we can input `betas/nnnn` tags if we want, which allows us to run tests pre-release. We could consider loosening this up further.

### Test Steps
1. Ensure all unit tests pass. E2E check against this PR should pass now.

### Checklist


.github/workflows/run_ui_tests.yml
```diff
@@ -1,10 +1,15 @@
...
-      - nightly-ui-tests
   workflow_dispatch:
+    inputs:
+      tag:
+        description: "Tag to test against"
+        required: true
+        type: string
+  push:
+    tags:
+      - 'betas/*'
```
ealbizzati: Will this ever trigger the ui tests?

`deploy_testflight_internal.yml` uses GITHUB_TOKEN to push the tags, but [GitHub's docs](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/trigger-a-workflow#triggering-a-workflow-from-a-workflow) say that events triggered by GITHUB_TOKEN will not create a new workflow run.

this isn't really a blocker because we already don't run the e2e tests when creating beta, and we need the fix for the tests so... just a heads up that this might not get triggered.

ealbizzati (approved):
looks good to me

.github/workflows/run_ui_tests.yml
tonisevener: Excellent catch, I have run into that very situation before. Maybe we just set a cron to run this nightly instead? I think that was the original goal for this workflow.

PR #6116, opened 2026-08-26
For You feed loading adjustments 

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T432926

### Notes
* Adds a `WMFForYouModulePreloader`, which frontloads the summaries, the images, and the sampled colours for the next module
* `WMFImageColorSampler` now decodes an image sample rather than the whole image (not a crop, but a smaller version)
* Added some caching from article info downloaded in interest selection to make the preference screen have all the data it needs when loading

### Test Steps
1. Fresh install the app until you're sorted into the experiment
2. Choose several interests
3. Verify the feed preference screen loads quickly
4. Verify the For You feed loads quicker than before 

### Checklist


tonisevener (commented):
I pushed up a couple of of tweaks from watching the interests selection screen traffic. I noticed sometimes toggling the topics chip caused a duplicate call in quick succession, and the related pages call when toggling an article sometimes fetched a different article's data. My commits should fix those.

One thing I'm concerned about (unrelated to your changes here) - right now we allow an unbounded number of interest articles, and it seems like For You can fetch all of them. If a power user adds 100s of interest articles, all of their "morelike:" calls will be fetched concurrently when For You loads, which could cause a really long spinner and probably way too much data than they could ever scroll through. I thought I remembered we capped these at some point. It looks like that cap requirement was undone [here](https://phabricator.wikimedia.org/T427674#12091804):

Old requirement:
> Each day, freshly pick seed interests from the users's selections, up to 10 interests in total: up to 5 articles, and up to 5 topics

New requirement:
> Randomly select 3 of their interests to display first in the feed.
> The rest of the interests should come at the end of the feed, after "Continue reading"
> If the user has N interests, they have N versions of this module to swipe through everyday

I'm not sure if removing the cap was intentional, but I think we should bring some form of it back. Note that there's a set number of topics (around 30), so that's not quite unbounded, but fetching 30 topic calls when For You loads should probably also not be allowed.

I did a search on Android and they still have the original cap ([1](https://github.com/wikimedia/apps-android-wikipedia/blob/12d1dd2bffea3db7ead5a69b2025bcbf287fe64d/app/src/main/java/org/wikipedia/feed/HomeViewModel.kt#L663), [2](https://github.com/wikimedia/apps-android-wikipedia/blob/12d1dd2bffea3db7ead5a69b2025bcbf287fe64d/app/src/main/java/org/wikipedia/feed/HomeViewModel.kt#L687))

tonisevener (changes requested):

mazevedofs:
> I pushed up a couple of of tweaks from watching the interests selection screen traffic. I noticed sometimes toggling the topics chip caused a duplicate call in quick succession, and the related pages call when toggling an article sometimes fetched a different article's data. My commits should fix those.
> 

Thanks!


> One thing I'm concerned about (unrelated to your changes here) [...] 
> I did a search on Android and they still have the original cap ([1](https://github.com/wikimedia/apps-android-wikipedia/blob/12d1dd2bffea3db7ead5a69b2025bcbf287fe64d/app/src/main/java/org/wikipedia/feed/HomeViewModel.kt#L663), [2](https://github.com/wikimedia/apps-android-wikipedia/blob/12d1dd2bffea3db7ead5a69b2025bcbf287fe64d/app/src/main/java/org/wikipedia/feed/HomeViewModel.kt#L687))

Good catch! I remember removing that because I thought it made sense, but I didn't think about this side effect

tonisevener (approved):
Worked well, thanks!

PR #6118, opened 2026-08-26
Fix invalid batch update crash in CollectionViewUpdater

ealbizzati:
Two fetched results change cycles can occur in one main queue drain. Cycle one calls reloadData() and no layout pass runs before cycle two. The collection view reads the data source lazily and moves to the post-change state. Cycle two then applies batch updates against stale counts and UIKit throws NSInternalInconsistencyException.

Compare the item counts of each collection view section with the previous section counts before granular updates run. Fall back to reloadData() when the counts do not match.

Add a regression test that reproduces the crash.

**Phabricator:** 

### Notes
* 

### Test Steps
1. 

### Checklist


ealbizzati:
Superseded by #6141. The branch was renamed to `T436533` to match https://phabricator.wikimedia.org/T436533, which closed this PR. Same commits and same diff continue in #6141.

PR #6126, opened 2026-08-27
Improve HTTP error handling in Session network requests

josephseddon:
### Notes
* Split HTTP response validation into two separate guard statements for better error reporting
* When an HTTP response has a non-200 status code, now returns a `RequestError.http(statusCode)` instead of `nil`
* Applied consistently to two request handling methods in the Session class

### Test Steps
1. Verify that network requests with non-200 HTTP status codes now properly report the error via the completion handler
2. Confirm that requests with valid 200 responses continue to work as expected
3. Test that invalid response types (non-HTTPURLResponse) still return nil error as before

### Checklist

https://claude.ai/code/session_01XoXAuCUJsJEttNPdMqzGMQ

### DIFF
--- a/Wikipedia/Code/Session.swift
+++ b/Wikipedia/Code/Session.swift
@@ -461,8 +461,12 @@ public class Session: NSObject {
                 completionHandler(nil, response, error)
                 return
             }
-            guard let httpResponse = response as? HTTPURLResponse, httpResponse.statusCode == 200 else {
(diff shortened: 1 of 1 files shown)


tonisevener:
FYI going to cherry pick this into the [first PR](https://github.com/wikimedia/wikipedia-ios/pull/6121), just so we can review them all together.

PR #6129, opened 2026-08-27
T436271: Add app install ID to client error events

ealbizzati:
**Phabricator:** 
https://phabricator.wikimedia.org/T436271

### Notes
* **Client error events:** the install ID travels inside `error_context` (`error_context.app_install_id`). It cannot go at the top level: the `/mediawiki/client/error/2.0.0` schema declares `additionalProperties: false`, so switching to the full event body (`needsMinimal: false`) would get every event rejected by EventGate with a 400 — `error_context` is the schema's designated field for arbitrary extra data (string map, max 32 keys).
* Fixed the type of `Event.errorContext` from `String?` to `[String: String]?`: the schema requires an object. Latent until now because we always sent `nil`.
* **Launch log:** the install ID is logged once at launch, after `evaluateAppInstallAndSessionIDs()` generates/migrates it. `DDLogWarn` on purpose: the default log threshold (`WMFLogging.h`) is warning, and the line must reach the `console.log` that ships with the "Export user data" package.
* **Support email:** the install ID is appended to the `mailto:` body in `HelpViewController.sendEmail()`, below the app version. The user sees it in the draft before sending. The Explore feed and Activity tab feedback emails were left out of scope — happy to add them here or in a follow-up if we want them too.
* **Developer settings:** new "App install ID" section at the top of the screen, with tap-to-copy
* **Drive-by fix:** `evaluateAppInstallAndSessionIDs()` loaded the session ID with the app install ID key, so the legacy session ID migration never ran. The fix is retroactively a no-op: for users who already went through the broken path, `UserSession` has long since generated a session ID, and the migration only writes when the destination is empty.
* **Post-deploy step:** the `error_context.app_install_id` field is new to the Logstash index — someone with access needs to refresh the field list on the index pattern (Management → Index Patterns) so autocomplete/aggregations work. Asked in #talk-to-sre.

### Test Steps
1. Launch the app and check the console for the `App install ID: <uuid>` line at startup.
2. Open a nonexistent article (e.g. `wikipedia://en.wikipedia.org/wiki/Some_Nonexistent_Title_123` from Safari in the simulator). With `ddLogLevel` set to `DDLogLevelAll` locally, confirm the scheduled client error event includes `error_context` with `app_install_id` matching the launch log.
3. Verified end to end in production Logstash: event visible with `error_context.app_install_id`, `http.status_code: 404` and `error_class: SessionDelegate` (screenshot below).
4. Settings → About → enable developer mode → Developer settings: the "App install ID" section shows the same UUID; tapping copies it (toast confirms) and pasting elsewhere matches.
5. Settings → Help → email button: the draft body contains `App install ID: <uuid>` below the version line.

### Checklist

### Screenshots/Videos
Screenshot 2026-08-27 at 1 59 54 PM
Screenshot 2026-08-27 at 1 12 08 PM
Screenshot 2026-08-27 at 1 25 47 PM

### DIFF
--- a/WMFComponents/Sources/WMFComponents/Components/Developer Settings/WMFDeveloperSettingsView.swift
+++ b/WMFComponents/Sources/WMFComponents/Components/Developer Settings/WMFDeveloperSettingsView.swift
@@ -17,6 +17,21 @@ struct WMFDeveloperSettingsView: View {
                 Toggle("Enable Developer Mode", isOn: $viewModel.enableDeveloperMode)
             }
 
+            Section(header: Text("App install ID"), footer: Text("Tap to copy. Use it to find this install's errors in Logstash.")) {
+                Button {
+                    viewModel.copyAppInstallID()
+                } label: {
+                    HStack {
+                        Text(viewModel.appInstallID ?? "Unavailable")
+                            .font(Font(WMFFont.for(.callout)))
+                        Spacer()
+                        if let copyIcon = WMFSFSymbolIcon.for(symbol: .docOnDoc) {
+                            Image(uiImage: copyIcon)
+                        }
+                    }
+                }
+            }
+
             Section(header: Text("Games")) {
                 Toggle("Show Games Version 2", isOn: $viewModel.showGamesV2)
                 Button {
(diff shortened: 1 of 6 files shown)


tonisevener (approved):
Works great, thank you!

PR #6134, opened 2026-08-29
Add appInstallId query parameter to outbound visual editor URL

voyagerfan:
**Phabricator:** https://phabricator.wikimedia.org/T435362

### Notes
* This PR appends the app's unique install id (appInstallid) as a key-value pair to the query string for visual editor.

### Test Steps
1. 

### Checklist

### DIFF
--- a/Wikipedia/Code/ArticleViewController+Editing.swift
+++ b/Wikipedia/Code/ArticleViewController+Editing.swift
@@ -103,6 +103,10 @@ extension ArticleViewController {
         if let sectionID {
             queryItems.append(URLQueryItem(name: "section", value: String(sectionID)))
         }
+        if let appInstallId: String = try? WMFDataEnvironment.current.crossProcessUserDefaultsStore?.load(key: WMFUserDefaultsKey.appInstallID.rawValue), !appInstallId.isEmpty {
+            queryItems.append(URLQueryItem(name: "appinstallid", value: appInstallId))
+        }
+        
         components?.queryItems = queryItems
         navigate(to: components?.url, useSafari: true)
     }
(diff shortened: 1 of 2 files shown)


Wikipedia/Code/ArticleViewController+Editing.swift
ealbizzati, line 1: I think I lacked some documentation when I changed this method. 

`VisualEditorReturnJourney` strips the known VE query items when it rebuilds the article URL (VisualEditorReturnJourney.swift:57-67), but `appinstallid` is not in that set. 

Its own doc comment (lines 5-6) covers the case where the incoming URL is the editing URL and that URL now carries `appinstallid`. 

So veaction makes `isEditingContext true`, `saved false`, and `returnJourney.articleURL` comes back with `?appinstallid=<uuid>` still attached, then gets handed to LinkCoordinator in WMFAppViewController+Extensions.swift:98.

It's not a crash, and the refresh in place will still work, but the article url will still hold the appinstallid param and anything the user shares from it will have it. That's why the strip list exist. 

The fix is pretty straightforward on `VisualEditorReturnJourney.swift`:
```swift
    private static let appInstallIDQueryItemName = "appinstallid"
    ...

        let visualEditorQueryItemNames: Set<String> = [
            Self.savedQueryItemName,
            ....
            Self.useFormatQueryItemName,
            Self.appInstallIDQueryItemName
        ]
```

let me know if the comment is confusing or if you need some help.

ealbizzati (approved):

Wikipedia/Code/ArticleViewController+Editing.swift
voyagerfan: Thanks for the explanation and approving!

PR #6135, opened 2026-08-29
Make saved-article rate limiting detectable and act on it

josephseddon:
SavedArticlesFetcher already pauses all downloading when it sees a 429, but
nothing on the download path ever produced that error, so the handling was
unreachable:

- CacheFetching.dataForURLRequest flattened every non-200 into
  RequestError.unexpectedResponse.
- Session.jsonDecodableTask reports a non-200 as (nil, response, nil) — a nil
  error — which trackedJSONDecodableTask also turned into .unexpectedResponse.

Every real rate limit therefore looked like an ordinary failure: the article was
branded .apiFailed, its backoff escalated, and the fetcher moved on to the next
article and kept going.

Carry the status code through both paths instead. trackedJSONDecodableTask is
changed rather than Session.jsonDecodableTask so the blast radius stays on the
three ArticleFetcher call sites this concerns, rather than the dozen unrelated
callers that currently rely on the silent nil-error behaviour.

Add RequestError.rateLimited, which carries the server's Retry-After, and honour
it when pausing — clamped so downloading resumes within a session, and jittered
so devices throttled together do not all return in the same second. Both
.rateLimited and a bare .http(429) are recognised, via httpStatusCode.

Also close the one HTTP-logging gap on this path. Session.handleResponse covers
the JSON task methods and SessionDelegate covers the delegate-driven tasks the
SchemeHandler uses, but neither covers the permanent-cache downloads:
Session.dataTask(with:completionHandler:) does not call handleResponse, and its
completion-handler task suppresses the data-delegate callbacks. Log 429 and 5xx
from there under a distinct source so download-driven throttling can be told
apart from throttling while reading an article. 404s are deliberately excluded —
this path sees routine ones for resources an article no longer references.

Co-Authored-By: Claude <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01CjLA9cfAZ4qJL6qeFSNsJ1

### DIFF
--- a/WMF Framework/ArticleCacheController.swift
+++ b/WMF Framework/ArticleCacheController.swift
@@ -2,14 +2,14 @@ import Foundation
 
 public final class ArticleCacheController: CacheController {
     
-    init(moc: NSManagedObjectContext, imageCacheController: ImageCacheController, session: Session, configuration: Configuration, preferredLanguageDelegate: WMFPreferredLanguageInfoProvider) {
+    init(moc: NSManagedObjectContext, imageCacheController: ImageCacheController, session: Session, configuration: Configuration, preferredLanguageDelegate: WMFPreferredLanguageInfoProvider, throttle: CacheRequestThrottle = CacheRequestThrottle()) {
         let articleFetcher = ArticleFetcher(session: session, configuration: configuration)
         let imageInfoFetcher = MWKImageInfoFetcher(session: session, configuration: configuration)
         imageInfoFetcher.preferredLanguageDelegate = preferredLanguageDelegate
         let cacheFileWriter = CacheFileWriter(fetcher: articleFetcher)
-        
+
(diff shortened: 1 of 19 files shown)


josephseddon:
Closing in favour of three correctly scoped, stacked PRs.

This PR was opened for the rate-limit detection work alone, but subsequent commits landed on the same branch, so it grew to carry all five commits (19 files, ~1400 additions) across three unrelated concerns. Splitting it:

| PR | Scope |
|---|---|
| #6136 | Make saved-article rate limiting detectable and act on it |
| #6137 | Bound concurrent downloads per article in the permanent cache |
| #6138 | Fetch saved-article image metadata in batches instead of one call per image |

They are stacked in that order, because they genuinely depend on each other: #6137's throttle is unsafe without the `group.notify` correctness fix it carries, and #6138 needs both #6136's `RequestError.rateLimited` and #6137's `Session.cacheResponse` fix.

@copilot-pull-request-reviewer's three findings here have been carried forward into the relevant PR descriptions as known follow-ups rather than being lost with this thread — in particular the `failedKeys.first` masking, which lands in #6137's code and can still defeat #6136's pause.


---
_Generated by [Claude Code](https://claude.ai/code)_

PR #6141, opened 2026-08-31
T436533: Fix invalid batch update crash in CollectionViewUpdater

ealbizzati:
**Phabricator:** https://phabricator.wikimedia.org/T436533

Recreated from #6118, which GitHub closed when the branch was renamed to match the task number. Same commits, same diff.

### Notes
* Two `NSFetchedResultsController` change cycles can run in one main queue drain, because the view context merges changes from the feed background context. Cycle one takes the `reloadData()` path and no layout pass runs before cycle two. The collection view reads the data source lazily, so it is already in the post-change state. Cycle two applies granular updates against counts the collection view has absorbed, and UIKit throws `NSInternalInconsistencyException`.
* The guard in `controllerDidChangeContent` compared section counts only. It now also compares `collectionView.numberOfItems(inSection:)` of every section with `previousSectionCounts`, and falls back to `reloadData()` when they do not match.
* The broken precondition produces two different UIKit assertions. A delete in cycle two hits the delete index validation. An insert hits the generic batch update validation. Production reports on build 6160 show both. One guard covers both.
* Only the Explore feed reaches this path, because it sets `isSlidingNewContentInFromTheTopEnabled = true`.
* This is not a recent regression. Workarounds for the same class of crash exist from 2018 and from T253762 in 2022. Both check section counts, not item counts.
* Known limit: the guard validates the pre-update state. It does not validate that the change set reconciles previous counts with new counts. No crash report shows that case.

### Test Steps
1. Run `WikipediaUnitTests/Code/CollectionViewUpdaterCrashTests.swift`. All three tests must pass.
2. `testTwoChangeCyclesInSameRunLoopPassMustNotThrow` covers the delete assertion. On `main` it fails with "attempt to delete item 2 from section 1 which only contains 2 items before the update".
3. `testTwoChangeCyclesWithInsertInSameRunLoopPassMustNotThrow` covers the generic assertion. On `main` it fails with "Invalid batch updates detected".
4. `testSingleRolloverChangeCycleMustNotThrow` is the control. It passes with and without the fix, which shows that single change cycles keep using granular updates.
5. Manual check: open the Explore feed, pull to refresh several times, and switch between the For You and Community segments on the Home tab. The feed must animate in new content as before.

### Checklist

### Screenshots/Videos


mazevedofs (approved):
🎉

ealbizzati:
Addressed both review comments in 4bdebcdf4c.

**On the control test:** Good catch. The control asserted final counts only, so it would also have passed if the updater always fell back to `reloadData()`. Added a `CallCountingCollectionView` that counts `reloadData()` and `performBatchUpdates(_:completion:)`, with the counters reset after setup. The control now asserts one batch update and zero reloads; the two crash scenarios assert zero batch updates and two reloads. That distinguishes the paths directly instead of inferring them.

**On strict concurrency:** Annotated the UIKit-facing helpers and the test methods with `@MainActor`. The delegate witnesses stay nonisolated to match `CollectionViewUpdaterDelegate`.

I could not measure this file under `SWIFT_STRICT_CONCURRENCY=complete`: the app target does not build in that mode today, failing earlier in `WMF Framework/ColumnarCollectionViewLayoutInfo.swift:17` on pre-existing code, so nothing downstream is analyzed. For context, the strict-concurrency workflow currently covers `WMFData` and `WMFComponents` only, with `WMFComponents` as the sole gate. The app target and this test target are not checked yet, and this is the first test file in that target to carry the annotations.

PR #6144, opened 2026-09-01
T427677: Swipe gesture overlay

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T427677

### Notes
* 

### Test Steps
1. With group b enabled

Open For You
Ensure you see the swipe gesture
Ensure it lasts 2s
(may need fresh install to test swipe to dismiss earlier, since it only shows once)

Ensure you are able to swipe to dismiss earlier if wanted
Ensure it doesn't show up again after first show

### Checklist
- [x] Communicate to users that they should swipe up for new content
- [x] Must not last more than 2 seconds, and should be easy to dismiss or dismiss automatically
- [x] Show a maximum of 1 time to user
- [x] Show on first impression of the "For you" feed (note: shows on first AVAILABLE impression, since we launched FY without this) 

### Screenshots/Videos

https://github.com/user-attachments/assets/3aff9bec-19bd-4df8-beb8-689540373e98

https://github.com/user-attachments/assets/4aa51e58-cbaf-4dce-9593-78a7db49e6c1

https://github.com/user-attachments/assets/556dfe50-eaac-4983-90ff-44180da04ae9


WMFComponents/Sources/WMFComponents/Components/Home/WMFForYouView.swift
```diff
@@ -175,6 +198,82 @@ public struct WMFForYouView: View {
...
+                        // A downward swipe only dismisses: above the first module there is nothing.
+                    } else {
+                        let isNextCard = layoutDirection == .rightToLeft
+                            ? translation.width > 0
+                            : translation.width < 0
+                        swipeOnboardingCardStep = isNextCard ? 1 : -1
+                    }
+                }
+        )
+        .onTapGesture { dismissSwipeOnboarding() }
```
ealbizzati, line 272: I'm not sure whether "a tap on the hint dismisses it instead of opening the article" is a hard requirement. If it isn't, I think the overlay could let the touch through instead of taking it and replaying the swipe.


```suggestion
        .allowsHitTesting(false)
```

WMFComponents/Sources/WMFComponents/Components/Home/WMFForYouView.swift
ealbizzati, line 399:
```suggestion
                    dismissSwipeOnboarding()
                    guard !hasReportedCurrentDrag else { return }
```

with that

swipeOnboardingCardStep, moveCard(by:), the layoutDirection read and the per-page onChange can be removed.

tradeoff: a tap during the 2 s reaches the card and opens the article (the 2 s auto-dismiss still applies), and a pull-down during the hint triggers refresh, same as without the hint.

ealbizzati (approved):
left a couple comments as suggestions. but I don't think those are blockers.

PR #6148, opened 2026-09-01
Evergreen account creation data and presentation logic

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T436006 https://phabricator.wikimedia.org/T436010

### Notes
* Adds the data controller and the presentation logic for the feature
* Adds some instrumentation based on the Android work [here](https://github.com/wikimedia/apps-android-wikipedia/pull/6832/changes/94ba98762a1cecc79e86d8aedc4f44940bfb0f6c) and Shay's [comment](https://wikimedia.slack.com/archives/C0BB50D9TGD/p1788291728080909?thread_ts=1788262116.346099&cid=C0BB50D9TGD)
* Slide content/vpice over to be done in another task

### Test Steps
Caution: Move the date forward only. If you move the date backward, the app counts the days incorrectly.

1. Use a device.
2. On the device, open Settings. Select General. Select Date & Time.
3. Set Set Automatically to off.
4. Move the date forward in days.

Test 1: An install from before this change

1. Install the TestFlight version of the app.
2. Read three articles.
3. Set the date forward 8 days.
4. Open the TestFlight version. Read one article.
5. Set the date forward 1 more day.
6. Install the new version of the app.
7. Open the app.
8. The prompt shows immediately. You do not wait.

Test 2: A new install

1. Install the app.
2. Read three articles. Save one article.
3. Result: The prompt does not show.
4. Set the date to day 8.
5. Open the app.
6. Result: The prompt does not show. This result is correct. The app does not count today on the same day.
7. Set the date to day 9.
8. Open the app.
9. The prompt shows on the Explore tab.
10. Read the four slides.
11. Each slide shows your own numbers. For example, See your 3 reading days come together in Year in Review.
12. A slide with the number 1 uses the singular word. For example, 1 saved article.
13. A slide with no data shows no number.

### Checklist


WMFData/Sources/WMFData/Data Controllers/Evergreen Account Creation/WMFEvergreenAccountCreationDataController.swift
```diff
@@ -0,0 +1,307 @@
...
+    // MARK: - Eligibility
+
+    /// - Parameters:
+    ///   - context: The surface about to show the prompt.
+    ///   - hasPermanentAccount: Whether the reader is logged in to a real account, which lives
+    ///     outside WMFData. A temporary account does not count: those readers have nothing to sync
+    ///     yet, so the prompt is still for them.
+    ///   - isAnotherPromptVisible: Whether a tooltip or prompt is already on screen. This prompt
+    ///     ranks below all of them and never shares the screen with one.
+    public func shouldShowPrompt(in context: PresentationContext, hasPermanentAccount: Bool, isAnotherPromptVisible: Bool) -> Bool {
```
l-olson1214, line 163: This isn't called anywhere in the base (tests only)

Wikipedia/Code/EvergreenAccountCreationCoordinator.swift
```diff
@@ -0,0 +1,311 @@
...
+    @discardableResult
+    func start() -> Bool {
+        Task { await startIfEligible() }
+        return true
+    }
+
+    /// Presents the prompt only if the reader is eligible right now. Returns whether it presented,
+    /// so a caller sequencing prompts knows whether the screen is taken.
+    @discardableResult
+    func startIfEligible() async -> Bool {
```
l-olson1214: We might need to add something like 
`guard await dataController.shouldShowPrompt(in: context, hasPermanentAccount: dataStore.authenticationManager.authStateIsPermanent, isAnotherPromptVisible: presenter.presentedViewController != nil) else { return false }`
here, right now it's showing on every single launch

Wikipedia/Code/EvergreenAccountCreationCoordinator.swift
```diff
@@ -0,0 +1,311 @@
...
+        self.dataStore = dataStore
+        self.context = context
+    }
+
+    // MARK: - Presentation
+
+    @discardableResult
+    func start() -> Bool {
+        Task { await startIfEligible() }
+        return true
```
l-olson1214, line 52: should we return whatever value we get from startifeligible here?

l-olson1214 (changes requested):

Wikipedia/Code/EvergreenAccountCreationCoordinator.swift
mazevedofs: I had some test code removed, good to test again

WMFData/Sources/WMFData/Data Controllers/Evergreen Account Creation/WMFEvergreenAccountCreationDataController.swift
mazevedofs: yeah sorry I had some test code pushed, can you try again?

Wikipedia/Code/EvergreenAccountCreationCoordinator.swift
mazevedofs: Not necessarily. `startIfEligible()` is async and `start()` is not, so `start()` cannot return its result without a wait on the main actor, which would be blocking. Nothing is reading the value returned from this function, that's why we have it marked as `@discardableResult`. I removed the return value from `startIfEligible` because it's not used

l-olson1214 (approved):

PR #6153, opened 2026-09-01
T435348: Apply design review fixes to the follow-up reminder card

ealbizzati:
**Phabricator:** https://phabricator.wikimedia.org/T435348

### Notes
* Design review fixes for the follow-up reminder card: 16pt of spacing between the card and the lead image, the card copy uses the article text size (and scales with the reader's text size setting), and the dismissal toast has no icon and auto dismisses.
* Adds a **Use Hardcoded Payment Methods** developer setting that skips the getPaymentMethods API call, for QA when the payments API rate limits the device.

### Test Steps
1. With a saved reminder and the reading goal reached, open an article. The card sits 16pt below the lead image, aligned with the article text, and its copy matches the article text size.
2. Tap **Not now**. The toast shows without an icon and dismisses on its own.
3. In Developer Settings → Fundraising, enable **Use Hardcoded Payment Methods** and tap **Yes, donate now** on the card. The Apple Pay prompt opens even when the payments API fails.

### Checklist


WMFData/Sources/WMFData/Data Controllers/Developer Settings/WMFDeveloperSettingsDataController.swift
```diff
@@ -134,6 +134,29 @@ public protocol WMFDeveloperSettingsDataControlling: AnyObject {
...
+                refetchDonateConfigs()
+            }
+        }
+    }
+
+    public var hardcodedPaymentMethodsOverride: WMFPaymentMethods? {
+        guard useHardcodedPaymentMethods,
+              let fileURL = Bundle.module.url(forResource: "donate-hardcoded-payment-methods", withExtension: "json"),
+              let data = try? Data(contentsOf: fileURL) else {
+            return nil
```
tonisevener, line 154: I am hitting this nil here when I run via simulator and the flag is on, `let data = try? Data(contentsOf: fileURL)` is failing on me, `useHardcodedPaymentMethods` is evaluating to true. Let me know what you think. Everything else looks good.

ealbizzati: That's weird. I verified end to end on my side. With the dev settings ON, the cached PaymentMethods.json contains exactly the donate-hardcoded-payment-methods.json's list. Maybe it's a stale SPM build  not regenerating the WMFData bundle? 

Could you try a clean build folder?

tonisevener: That fixed it, thanks!

tonisevener (approved):

tonisevener (approved):

PR #6161, opened 2026-09-03
Donation reminder instrumentation - Phase 1

tonisevener:
**Phabricator:** https://phabricator.wikimedia.org/T435547

### Notes
Donation reminder instrumentation for everything in the deck except the "end of experiment" slides. It may be better to review all at once. I had to undo and rework a fair amount of stuff in later commits.

### Test Steps
1. Set WMFLogging.h = DDLogLevelAll

(Optional)
It's easier to view these events in the console if you comment out the `print()` statements in TestKitchenAdapter.swift, and wrap the DDLogDebug statements in a stream conditional in EventPlatformClient.swift:

```
if stream == .appDonorExperience {
    DLogDebug("\n\n📊EPC: Scheduling event to be sent to \(config.destination_event_service):")
    DDLogDebug("\(printablePayload)")
}
```

2. Watch the console and trigger events. The banner (and associated donate form flows) should log events with campaign IDs that look like `enNL_NL_2026_08_Control_remind{ABC}_iOS`. The donation reminder (and associated donate form flows from there) should log events with campaign IDs that look like: `enNL_appmenu_remind{BC}_iOS`

### Checklist


Wikipedia/Code/ArticleViewController+Announcements.swift
```diff
@@ -27,10 +27,43 @@ extension ArticleViewController {
...
+            }
+            
+            var donateSource: DonateCoordinator.Source = .articleCampaignModal(articleURL, activeCampaignAsset.metricsID, donateURL)
+            
+            // Setup donation reminder experiment if needed
+            if WMFDeveloperSettingsDataController.shared.enableDonationReminder {
+                
+                guard activeCampaignAsset.id == WMFDonationReminderDataController.experimentCampaignID else {
+                    return
+                }
```
ealbizzati, line 42: the new guard activeCampaignAsset.id == experimentCampaignID else { return } means that if the active campaign is not the experiment campaign, no banner shows at all.

right now that's fine, but on the PR i'm gonna add removing the FF, this guard will run for everyone, when the NL campaign ends and a new campaign starts, users in this app version will not see any new campaign banners.

would it be better to default to .articleCampaignModal flow instead of returning?

ealbizzati (approved):
Talked IRL (aka: slack).
The campaign guard structure mentioned in the comment will land on my follow up PR, removing FF and updating the campaign ID for the experiment

PR #6164, opened 2026-09-03
Replace CIDetector face detection with the Vision framework

mazevedofs:
**Phabricator:** https://phabricator.wikimedia.org/T229534

### Notes
- Use VNDetectFaceRectanglesRequest to find faces in article images. Vision returns normalized bounding boxes and reads the image orientation. The system selects the compute unit, so the onGPU parameter is not necessary. Remove it from all call sites.
- Add WMFFaceDetectionCache.swift to the WMF framework. The class keeps the same Objective-C interface. The cache sorts the faces by size and returns the largest face first. If Vision fails, the cache returns no face. The caller can still show the image.
- Remove the CIDetector, CIContext, and CIImage helper categories. Remove the UIImageView category that selected the GPU or the CPU from the application state.
- Add unit tests for the coordinate conversion, the cache, the failure fallback, and cancellation. Vision does not run in the unit test host, so the tests inject a detection function.

### Test Steps

1. Open a biography article, for example "Angela Merkel", and check that the lead image is cropped around the face.
2. Open an article with a non-portrait lead image, for example "Huhu beetle", and check that the image is top-aligned.
3. Open an article with a group photo, for example "The Beatles", and check that the crop centers on the largest face.
4. Go to the Explore feed and check that portrait thumbnails in "In the news" and "Top read" cards show faces, not cut-off heads.
5. Scroll the Explore feed fast for several screens and check that thumbnails do not flicker or swap between cells.
6. Open Saved and check that saved article thumbnails and reading list grid images crop around faces.
7. Open Places, tap a pin for a person-related article, and check the callout thumbnail.
8. Open a talk page and an article history diff for a biography and check the header image crop.
9. Open the editor, tap "Insert media", search for "portrait", and check that result thumbnails are not face-cropped (detection is off there).
10. Open the same biography article twice and check that the second load shows the crop instantly (cached result).
11. Send the app to the background, return, and open a new biography article to check that detection still works.
12. Add the Continue Reading widget with a biography as the last read article and check the widget thumbnail.
13. Switch to dark theme and check that face-cropped images render the same.
14. Check the console for Face detection failed warnings during the session; none are expected on a device.

### Checklist


ealbizzati:
This PR also removes the code path behind a crash group we have in the Organizer

**Signature**
`EXC_BAD_ACCESS (SIGSEGV)` in `libfaceCore.dylib` → `IntegralImage::setAnglePrivate` → `__bzero` with a NULL destination (~883 KB), inside a `dispatch_apply` opened by Apple's face detector. Our entry frame is `-[CIDetector(WMFFaceDetection) wmf_detectFeaturesInImage:options:failure:success:]_block_invoke` (`CIDetector+WMFFaceDetection.m:51`).

**Why this PR fixes it**
The stack shows `VNFaceDetectorPrivateRevisionLegacyFaceCore`: `CIDetector` routes to Vision's revision 1, which the SDK header marks deprecated since iOS 16. `VNDetectFaceRectanglesRequest` defaults to revision 3 with our deployment target, which is the ML-based detector and never touches `IntegralImage`. Deleting `CIDetector+WMFFaceDetection` removes the crashing path entirely.

**Repro context from the crash comments**
Tapping through chains of article links, the random article dice, and pull-to-refresh on the home/community feed. All of them run face detection on lead images (1280px) or Explore card thumbnails. Not deterministic, it's an allocation failure or race inside Apple's library.

One optional suggestion: pin `request.revision = VNDetectFaceRectanglesRequestRevision3` in `WMFFaceDetector` so the legacy revision can never come back through an SDK default change.

Crash report: incident F153150B-B659-476B-BCF9-974498437C71 (2026-08-30, iPhone18,3, iOS 26.6.1, build 2026.08.28 (6160)).

Screenshot 2026-09-03 at 9 47 48 PM

WMF Framework/WMFFaceDetectionAdapter.swift
```diff
@@ -0,0 +1,59 @@
...
+    /// The adapter calls the callbacks on the main actor.
+    @objc(detectFaceBoundsInImage:URL:failure:success:)
+    public func detectFaceBounds(in image: UIImage, url: URL?, failure: @escaping (Error) -> Void, success: @escaping (NSValue?) -> Void) {
+        Task {
+            do {
+                let faceBounds = try await cache.faceBounds(in: image, for: url)
+                success(faceBounds.map { NSValue(cgRect: $0) })
+            } catch is CancellationError {
+                // The caller cancelled the request. Call no callback.
+            } catch WMFFaceDetectionError.missingURL {
```
ealbizzati: `activeDetectionsKeyedByURL (WMFFaceDetectionCache.swift:80)` is keyed by the URL only and `cancelFaceDetection(for:)` cancels whatever task is registered under that url without checking who started it. 

2 image views showing the same url can cancel each other (one example would be 2 reading lists with same article saved). 

When one of the cells is recycled, prepareForReuse -> wmf_reset -> wmf_cancelImageDownload cancels the detection the other cell is waiting for. 

This catch then swallows the cancellation error with no call back, so that view never reaches `self.image = image`  and shows an empty thumbnail until the row is reconfigured. It's an edge case since this is a race condition. But this allows the view to fallback to the "bad crop" (without catering on the face) that we currently have.

```suggestion
            } catch is CancellationError {    
                // Another image view that shares this URL may have cancelled the task.
                // Report "no face" so this caller still shows its image.
                success(nil)
            } catch WMFFaceDetectionError.missingURL {
```

ealbizzati (commented):
there's a one edge case comment with a suggestion. see if that makes sense. if it doesn't let me know and I can approved, because other than that it looks good to me.

ealbizzati (approved):

PR #6171, opened 2026-09-04
T430742: P0 fixes

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T430742

### Notes
* Fix scroll notes from Sarah 
* Fix scaling notes from Sarah

### Test Steps
1. Load up a few articles
2. Scroll through to end of feed on FY
3. scroll beyond it, slowly
4. Make sure the text doesn't get weirdly small
5. make sure when you scroll it drops you at the beginning of the feed

### Checklist

### Screenshots/Videos

https://github.com/user-attachments/assets/a798a789-159a-4e70-b1a8-3a32fbb61fe7

### DIFF
--- a/WMFComponents/Sources/WMFComponents/Components/Home/WMFForYouEndOfFeedCardView.swift
+++ b/WMFComponents/Sources/WMFComponents/Components/Home/WMFForYouEndOfFeedCardView.swift
@@ -68,10 +68,18 @@ struct WMFForYouEndOfFeedCardView: View {
     var body: some View {
         GeometryReader { geometry in
             VStack(alignment: .leading, spacing: 32) {
+                // The text keeps its place at the bottom of the card and the illustration sits in
+                // the middle of whatever room is left above it. On a phone there is none, the
+                // spacers collapse, and the layout is the stack it was before. On an iPad the
+                // illustration no longer hangs off the bottom with the screen empty above it.
+                Spacer(minLength: 0)
+
                 illustration
-                    .frame(maxWidth: .infinity)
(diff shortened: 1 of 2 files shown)


WMFComponents/Sources/WMFComponents/Components/Home/WMFForYouEndOfFeedCardView.swift
```diff
@@ -108,14 +108,29 @@ struct WMFForYouEndOfFeedCardView: View {
                     )
                     .minimumScaleFactor(0.3)
                 }
+                // Two lines for every one of them: the title, the three sentences and both link
+                // rows. Applied to the stack so it reaches each `Text` inside it.
+                .lineLimit(2)
```
ealbizzati, line 113: do we really need this, if you have the minimumScaleFactor above to 0.3 and 0.25 on the fields? I'm just wondering if in different languages this might truncate the subtitle which, in English, already has 2 lines.

ealbizzati (commented):
just that one comment, but other than that it looks good to me. let me know what you decide and I can approve it.

WMFComponents/Sources/WMFComponents/Components/Home/WMFForYouEndOfFeedCardView.swift
ealbizzati: talked with @l-olson1214 and the goal is specifically to never go beyond 2 lines, even if that get's cut off, indeed. matching android's behaviour.

ealbizzati (approved):

PR #6183, opened 2026-09-10
T437342: Activity + YiR

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T437342

### Notes
* Config is not in place, references 2025 - you must utilize dev settings at this point
* The only difference between low data and data rich is logged in/out at this moment, will update as we begin to collect data

### Test Steps
1. Explore both on logged in and logged out
2. Navigate to developer settings, enable year in review entry point for 2026
3. Notice icon on activity tab
4. Tab into activity tab
5. If logged OUT, ensure you see the low data version
6. If logged IN, ensure you see the high data version (note: these will have more complex differentiation later on)
7. Tap anywhere on the card
8. Ensure it opens to (currently the old) year in review
9. Fresh install and try with the other authentication state
10. Check settings, ensure year in review entry point is no longer there

### Checklist
- [x] Move the permanent entry point for Year in Review into the Activity tab for logged-in and logged-out users
- [x] Low-data users see copy variant "Our Year in Review is here"
- [x] Data-rich users see copy variant "Your Year in Review is here"
- [x] Tapping the entry point launches directly into the appropriate flow: personalized flow for data-rich users, collective flow for low-data/logged-out users, per the Shared Logic task's data/low-data split
- [x] Remove previous entry point under "More"
- [x] Show an active badge on the Activity tab for logged-in and logged-out users once Year in Review is ready
- [x] Active state should be removed as soon as the user taps into Activity
- [x] Entry point should remain visible/accessible until the end of Year in Review's availability period

### Screenshots/Videos

https://github.com/user-attachments/assets/fd74dc02-f34b-4d6d-a9c7-8705d5f438bd (updated to do onappear of activity tab since video)

Simulator Screenshot - iPhone 16 Pro 26 5 - 2026-09-10 at 15 42 37


Wikipedia/Code/ActivityTabViewController.swift
```diff
@@ -118,6 +118,26 @@ final class WMFActivityTabHostingController: WMFComponentHostingController<WMFAc
...
+
+        if viewModel.yearInReviewViewModel == nil {
+            let yirViewModel = WMFActivityTabYearInReviewViewModel()
+            yirViewModel.onTap = { [weak self] in
+                self?.yirCoordinator?.start()
+            }
+            viewModel.yearInReviewViewModel = yirViewModel
+        }
+
+        // TODO: Shared Logic task supplies the real data-rich / low-data split
```
mazevedofs, line 137: Do we need this TODO, given that we will work on the new YiR?

WMFData/Sources/WMFData/Data Controllers/Developer Settings/WMFDeveloperSettingsDataController.swift
```diff
@@ -79,6 +80,21 @@ public protocol WMFDeveloperSettingsDataControlling: AnyObject {
         get { (try? userDefaultsStore?.load(key: WMFUserDefaultsKey.developerSettingsShowYiR2025.rawValue)) ?? false }
         set { try? userDefaultsStore?.save(key: WMFUserDefaultsKey.developerSettingsShowYiR2025.rawValue, value: newValue) }
     }
+    
+    // 2026 YIR
+    /// Debugging convenience: when true, the Year in Review entry point ignores the settings
+    /// toggle, the remote config's active window, and the suppressed-country list, so it presents
+    /// before the year's config exists remotely.
+    public var forceYiREntryPoint: Bool {
```
mazevedofs, line 88: Can we rename it to show YiR 2026?

WMFComponents/Sources/WMFComponents/Components/Developer Settings/WMFDeveloperSettingsViewModel.swift
```diff
@@ -118,13 +118,15 @@ import WMFData
         let forceHcaptchaChallenge = WMFFormItemSelectViewModel(title: "Force hCaptcha Challenge", isSelected: WMFDeveloperSettingsDataController.shared.forceHCaptchaChallenge)
         let allowGestureZoomArticleWebview = WMFFormItemSelectViewModel(title: "Allow pinch to zoom when reading articles", isSelected: WMFDeveloperSettingsDataController.shared.allowGestureZoomArticleWebview)
         let enableHomePhase2 = WMFFormItemSelectViewModel(title: "Enable Home Phase 2", isSelected: WMFDeveloperSettingsDataController.shared.enableHomePhase2)
+        let forceYiREntryPoint = WMFFormItemSelectViewModel(title: "Force Year in Review Entry Point 2026", isSelected: WMFDeveloperSettingsDataController.shared.forceYiREntryPoint)
```
mazevedofs, line 121: Can we rename it "Show Year in Review 2026"? We'll probably use it for more than the entry point

WMFData/Sources/WMFData/Notifications/WMFNSNotification.swift
```diff
@@ -16,6 +16,7 @@ public enum WMFNSNotification {
     public static let forYouModuleVisibilityDidChange = Notification.Name(WMFNotificationName.forYouModuleVisibilityDidChange.rawValue)
     public static let forYouInterestsDidChange = Notification.Name(WMFNotificationName.forYouInterestsDidChange.rawValue)
     public static let pageViewHistoryDidChange = Notification.Name(WMFNotificationName.pageViwHistoryDidChange.rawValue)
+    public static let yearInReviewActivityTabBadgeNeedsUpdate = Notification.Name(WMFNotificationName.yiractivity2026.rawValue)
```
mazevedofs, line 19: Format in camel case in `yiractivity2026`

Wikipedia/Code/WMFAppViewController.swift
```diff
@@ -468,6 +475,7 @@ final class WMFAppViewController: UITabBarController, AppTabBarDelegate {
         SessionsFunnel.shared.appDidBecomeActive()
         startEvergreenAccountCreationSession()
         checkRemoteAppConfigIfNecessary()
+        updateActivityTabYearInReviewBadge()
```
mazevedofs, line 478: We also need to add this call in `updateDefaultTab()` in this file

WMFComponents/Sources/WMFComponents/Components/Activity Tab/WMFActivityTabView.swift
```diff
@@ -36,13 +52,38 @@ public struct WMFActivityTabView: View {
                 }
             }
         }
+        .safeAreaInset(edge: .bottom) {
+            // Logged in, the card sits at the top instead — see loggedInList and
+            // customizedEmptyState.
+            if usesPinnedYearInReviewCard, hasYearInReviewCard {
```
mazevedofs, line 58: This is outside the progress view, so the user can get a loading view with the card pinned at the bottom. This doesn't happen on the logged-in branch, the card doesn't show when the view is loading because they're inside the same branch. 
I think both behaviors can be fine, it'd be good to have them consistent

WMFComponents/Sources/WMFComponents/Components/Activity Tab/WMFActivityTabView.swift
```diff
@@ -36,13 +52,38 @@ public struct WMFActivityTabView: View {
...
+                    .padding(.bottom, 16)
+            }
+        }
         .onAppear {
             viewModel.fetchData(fromAppearance: true)
         }
     }
 
+    @ViewBuilder
+    private var yearInReviewCard: some View {
```
mazevedofs, line 69: Any reason to have this being rendered in 5 different places instead of two (top/bottom case)?

We have this called in lines 59, 79, 245, 266, and 596 in this file.

Maybe something like this could help simplify this logic? The code repetition can lead to edge-case bugs like the loading one


```suggestion
    private enum YearInReviewCardPlacement {
    case none, top, inlineBottom, pinnedBottom
}

private var yearInReviewCardPlacement: YearInReviewCardPlacement {
    guard !viewModel.isLoading,
          viewModel.yearInReviewViewModel != nil else { return .none }

    if viewModel.authenticationState == .loggedIn { return .top }

    return dynamicTypeSize.isAccessibilitySize ? .inlineBottom : .pinnedBottom
}
```

Wikipedia/Code/ActivityTabViewController.swift
l-olson1214: Can definitely remove it! I just wanted to note that we had only some of the logic in there

WMFComponents/Sources/WMFComponents/Components/Activity Tab/WMFActivityTabView.swift
l-olson1214: Yeah! I think we still end up needing more than top/bottom because we have
- safeAreaInset (line ~53) — logged out, normal text size. Pinned above the tab bar, outside the content. Applies to both logged-out containers.
- loggedInList (line ~73) — logged in with at least one module on. Section at the top of the List.
- loggedOutList, empty branch (line ~245) — logged out, no sections, accessibility text size. Plain child in the VStack inside the ScrollView.
- loggedOutList, list branch (line ~266) — logged out, has sections, accessibility text size. Section at the bottom of the List.
- customizedEmptyState (line ~596) — logged in, all modules off. Plain child in the VStack inside the ScrollView.

.top and .inlineBottom each appear in two containers (a List Section vs a plain VStack), which need different modifiers — listRowInsets, listRowBackground, and top vs bottom padding, and then we have the loading state, but the loading state is a really good callout - fixed!

WMFComponents/Sources/WMFComponents/Components/Activity Tab/WMFActivityTabView.swift
l-olson1214: Thank you! Went with hiding during loading!

Wikipedia/Code/WMFAppViewController.swift
l-olson1214: Added!!

mazevedofs (changes requested):
Most of it looks good, just this item in the PR test section:
> Check settings, ensure year in review entry point is no longer there

Not sure that's what you mean, but that's the only option under Settings.
Image

mazevedofs (approved):

PR #6184, opened 2026-09-11
T437359: Add the semantic search experiment and its developer settings

ealbizzati:
**Phabricator:** 

### Notes
* It adds the experiment and the developer settings, and no UI.
* `WMFExperimentsDataController` gets the `semanticSearch` experiment: a 50/50 split between `SemanticSearch_Control` and `SemanticSearch_GroupB`. The bucket names still need to be aligned with Android and with the TestKitchen instrument.
* `WMFSemanticSearchDataController` assigns the bucket on the first eligible search, not on launch, so readers who never search in a target language stay out of the denominator. The entry point is available when four gates pass: the developer toggle, the target language, and the Group B bucket.
* The target languages come from `iosv1.semanticSearchLanguages` in the remote feature config. The key does not exist yet, so the controller falls back to `fr`, `ar` and `ja` (will we use the existing hybridSearch?). When we turn the feature on, the fallback goes away, an empty or missing array means off, and the developer settings below are removed.
* The "Enable Semantic Search" developer toggle keeps everything hidden in production. Without it no gate is evaluated and no bucket is assigned.
* The "Force Experiment Group" picker overrides the persisted bucket at read time and bypasses the language gate, so we can test in English. Switching it back to Off restores the persisted bucket.

### Test Steps
1. Not much to test, since this is an enabler PR.

### Checklist

### Screenshots/Videos
Simulator Screenshot - iPhone 17 - 2026-09-11 at 09 14 01

### DIFF
--- a/WMFData/Sources/WMFData/Models/Developer Settings/WMFFeatureConfigResponse.swift
+++ b/WMFData/Sources/WMFData/Models/Developer Settings/WMFFeatureConfigResponse.swift
@@ -106,6 +106,7 @@ public struct WMFFeatureConfigResponse: Codable {
     public struct IOS: Codable {
         public let hCaptcha: HCaptcha?
         public let visualEditorEnabled: Bool?
+        public let semanticSearchLanguages: [String]?
         
         public struct HCaptcha: Codable {
             public let baseURL: String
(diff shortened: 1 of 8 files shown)


WMFData/Sources/WMFData/Models/Developer Settings/WMFFeatureConfigResponse.swift
```diff
@@ -106,6 +106,7 @@ public struct WMFFeatureConfigResponse: Codable {
     public struct IOS: Codable {
         public let hCaptcha: HCaptcha?
         public let visualEditorEnabled: Bool?
+        public let semanticSearchLanguages: [String]?
```
ealbizzati: I don't like this being optional. but I kind of need it because of Codable conformance. 

Should I add the `init(from:)` and do the `decodeIfPresent ?? []` to avoid the optional array? I'm on the fence but accepting a second opinion

tonisevener (approved):

PR #6189, opened 2026-09-14
Prune persisted state of retired experiments

ealbizzati:
**Phabricator:** https://phabricator.wikimedia.org/T437359

### Context
While adding the semantic search experiment we looked at what happens to an experiment's persisted state once the experiment ends. The answer was: nothing. `WMFExperimentsDataController` writes a bucket file and a percentage file per experiment into the `Experiments` directory of the shared container, and the only code that ever removes them is the "clear" button in Developer Settings. The four experiments removed so far (article search bar, more dynamic tabs v1, and the activity tab twice) deleted their code and left their files behind on every device that took part.

Leftover files cost nothing in space, but they are a trap: if a new experiment reuses the file names of a retired one, readers who took part in the old experiment enter the new one with their old assignment, and nobody notices. Android ran into exactly this with the donation reminder experiment. iOS has already come close: the activity tab experiment was created twice with the same `activityTabBucket` file name, and only escaped because the bucket values were renamed between the two runs.

We wanted a fix that does not depend on anyone remembering to clean up when an experiment ends, and that does not touch the assignment of readers who are in an experiment right now.

### Notes
* `WMFKeyValueStore` gets `keys(inDirectory:)`, so the controller can list the `Experiments` directory instead of keeping a list of historical file names.
* `pruneRetiredExperiments` removes every file in that directory that no `Experiment` case uses. It runs from the WMFData housekeeping task, next to the Core Data pruning. Retiring an experiment stays what it is today, removing the case; the next housekeeping run does the rest.
* A signature file is persisted next to each bucket: the bucket file name plus the bucket values of the experiment. On read, a bucket whose signature differs from the current experiment is discarded and rolled again, so a reused file name can no longer carry an old assignment, even for readers who skipped the versions in between. A bucket persisted before signatures existed keeps its value and gets the signature on first read.
* The on-disk format does not change. Bucket and percentage files stay as they are, so older app versions keep working after a downgrade.
* `WMFExperimentsDataController` becomes `public` only to expose `pruneRetiredExperiments` to the app target.
* Out of scope: feature keys in `UserDefaults` left behind by removed features. They have no common prefix, so a generic prune cannot tell them apart from the rest of the app's defaults.

### Test Steps
1. Run the `WMFData` tests. `WMFExperimentsDataControllerTests` covers the signature backfill, a bucket from another experiment, a stale signature, reset, pruning, and `updatingTheAppKeepsEveryPersistedBucket`, which simulates a reader updating the app in the middle of every active experiment, in every bucket, with the percentages the features use.
2. On a device or simulator that already has an experiment assignment (for example, Home tab), install this build and confirm the assignment does not change. Developer Settings still shows the same forced or persisted group.
3. Optional: in the app group container, add a stray file to the `Experiments` directory, trigger the housekeeping background task, and confirm the stray file is gone while the files of active experiments remain.

### Checklist

### Screenshots/Videos
n/a


l-olson1214 (approved):
LGTM!! Definitely want Toni and Marina's opinions though!

mazevedofs (approved):

PR #6196, opened 2026-09-17
T438064: Set up YIR Announcement Architecture

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T438064

### Notes
* 

### Test Steps
1. Enable Show YIR 2026 and Show YIR Announcement in Developer settings
2. Go to home / for you 
3. Background
4. Foreground
5. Ensure you see the *old* YIR announcement (for now) - will be updated in second part of ticket
6. Ensure dismissing ticket shows toast
7. With developer settings you will be able to see it multiple times upon foreground, that's expected

### Checklist
- [x] Full-screen announcement shown once Year in Review is live and available to the user
- [x] Announcement appears once per users, on launch/first relevant app open after release, per the Shared Logic task's availability rules (respecting suppressed countries, toggle in settings, etc.)
- [x] Priority-level should be the same as previous year's announcementEnsure we are not suppressing the announcement for users who saw Year in Review in 2025
- [x] Do not show the announcement when a user has arrived to article view from a deep link
- [x] Dismissing the announcement doesn't block access, users can still reach Year in Review via the Activity tab entry point afterward
- [x] After a user dismisses the prompt, show toast/snackbar "You can access your Year in Review later in Activity."

### Screenshots/Videos

https://github.com/user-attachments/assets/83106b11-6f7e-43af-8f8e-c58bee626de4

https://github.com/user-attachments/assets/ec741bba-5c37-4422-a746-80deef036029


Wikipedia/Code/ArticleViewController+Announcements.swift
```diff
@@ -6,6 +6,11 @@ import WMFNativeLocalizations
 
 extension ArticleViewController {
 
+    /// Set when the fundraising banner shows. The Year in Review announcement waits for the next
+    /// app open rather than appearing behind it, so a user eligible for both never gets them back
+    /// to back. Session-scoped, never persisted.
+    static var didShowFundraisingBannerThisSession = false
```
mazevedofs, line 12: I think it's safer to rely on the same logic as `presentModalsIfNeeded`. I don't see this prop being reset, and fundraising banners can be shown several times.

Wikipedia/Code/HomeViewController.swift
```diff
@@ -461,6 +490,14 @@ extension HomeViewController: LogoutCoordinatorDelegate {
     }
 }
 
+extension HomeViewController: YearInReviewAnnouncementPresenting {
+    // TODO: confirm with the data team whether the Home tab needs its own value here. Sending
+    // `explore_prompt` keeps the funnel to values it already knows, but it under-reports the
+    // For You and home phase 2 surfaces.
+    var yirAnnouncementCoordinator: YearInReviewCoordinator? { yirCoordinator }
+    var yirAnnouncementLoggingID: String { "explore_prompt" }
```
mazevedofs: We don't have this requirement in the task, I think it can result in polluted data to add this here

Wikipedia/Code/YearInReviewAnnouncementPresenting.swift
```diff
@@ -0,0 +1,87 @@
...
+    var yirDataController: WMFYearInReviewDataController? { get }
+
+    /// The coordinator that presents the announcement. Each surface keeps its own, tied to its
+    /// navigation controller.
+    var yirAnnouncementCoordinator: YearInReviewCoordinator? { get }
+
+    /// Identifies the surface in the intro slide's logging.
+    var yirAnnouncementLoggingID: String { get }
+
+    func updateProfileButton()
```
mazevedofs: This func needs to stay out of here because the compiler reads it as all the pre-existing funs with the same name now conforming to this protocol.  Moving all the older `updateProfileButton` here will cause problems: user notification related updates are handled by this button, not only YiR.

Wikipedia/Code/YearInReviewAnnouncementPresenting.swift
```diff
@@ -0,0 +1,87 @@
...
+/// `ExploreViewController` owns it while the Explore feed is the visible feed. In the Home tab that
+/// is not always true: the embedded Explore feed is only on screen while the Community segment is
+/// selected, and with home phase 2 enabled it is never created at all. `HomeViewController` covers
+/// those cases, so the announcement still lands on the first relevant app open.
+///
+/// Both surfaces run the same guards and mark the same once-per-user flag, so moving between them
+/// cannot produce a second showing.
+@MainActor
+protocol YearInReviewAnnouncementPresenting: UIViewController {
+    var yirDataController: WMFYearInReviewDataController? { get }
```
mazevedofs: Any reason this is a requirement and not implemented in the body of the protocol? It's a singleton; you can just have an instance here.

WMFData/Sources/WMFData/Data Controllers/Developer Settings/WMFDeveloperSettingsDataController.swift
```diff
@@ -39,65 +40,117 @@ public protocol WMFDeveloperSettingsDataControlling: AnyObject {
...
+    }
+
+    /// Debugging convenience: while on, the 2026 Year in Review announcement ignores every gate —
+    /// the remote config, its active window, the opt-out toggle, suppressed countries and the
+    /// once-per-user state — so it presents before a 2026 config exists remotely and can be
+    /// retriggered without reinstalling. Named `force` for the same reason as `forceYiR2026`:
+    /// nothing below it is respected.
+    public var forceYiR2026Announcement: Bool {
+        get { loadFlag(.developerSettingsForceYiR2026Announcement) }
+        set { saveFlag(.developerSettingsForceYiR2026Announcement, newValue) }
```
mazevedofs, line 136: Maybe keep this bit? Not sure it was needed for the task

`if oldValue != newValue {
                NotificationCenter.default.post(name: WMFNSNotification.yearInReviewActivityTabBadgeNeedsUpdate, object: nil)
            }`

mazevedofs (changes requested):

mazevedofs (approved):

PR #6198, opened 2026-09-17
T437359: Replace the search results list with WMFSearchResultsView

ealbizzati:
**Phabricator:** 

- Wires `WMFSearchResultsView` (from #6194) into `SearchResultsViewController` and deletes `SearchResultsListViewController`. The search screen keeps `UISearchController`, the language bar and the legacy `WMFSearchFetcher`; only the results list changed.
- `SearchResultsViewController+ResultsList` holds the app-side glue: builds the view model, maps `MWKSearchResult` into the component's `SearchResult`, and handles tap (funnel + `saveLastSearch`), open, open in new tab, open in background tab, save/unsave (`ReadingListsAlertController` + `ReadingListsFunnel`) and share.
- `SearchResultsMapper` is a new stateless struct with the mapping and the "Redirected from" description logic, extracted so it could be unit tested (`SearchResultsMapperTests`, 9 tests, no network). The suite is serialized because `NSLocale.wmf_locale(for:)` mutates an unsynchronized static cache; worth a follow-up.
- Saved state: read through `dataStore.savedPageList` when results arrive and refreshed on `WMFArticleUpdated`, same as the deleted controller did.
- "Open in background tab" reproduces the peek behavior (tab limit → append to the current tab + toast; otherwise `createArticleTab(setAsCurrent: false)` + `ArticleTabsFunnel`).
- Context menu matches the peek: "View on a map" appears when the result has coordinates and opens Places; the tab actions only show for article URLs (`LinkCoordinator.destination`); the menu actions log the same `ArticleTabsFunnel` long press events the peek logged. Swipe actions keep the `ReadingListsFunnel` save/unsave events with position, like the old list.
- `overrideUserInterfaceStyle` is now set in `apply(theme:)` before the view loads. Without it the `UISearchController` results view kept the system appearance, which showed as light-gray separators in Black and a flash on open.
- The list background is `paperBackground` (like History and Saved). In the Black theme the area behind the status bar and language bar is now black instead of gray, because the iOS 26 scroll edge effect shows the list background instead of the old `midBackground`. Flagging in case design wants the gray back.
- Second commit removes `setTitleHTML(_:boldedString:)` and `boldFont` from `ArticleCollectionViewCell`; the search list was the only caller.

### Test Steps
1. Open Search and type a term with many results (e.g. "Tim cook"). Results appear under the language bar and the keyboard stays up while scrolling.
2. Type a term with a redirect (e.g. "Tim cook" → "The Subdudes"). The row shows "Redirected from: Tim Cook (musician)".
3. Tap a result. The article opens and the search term is saved in Recently searched.
4. Swipe a row left: Share and Save/Unsave appear. Save one, open the article, unsave it there, come back: the row reflects the new state.
5. Long press a row: preview with summary image and extract, plus Open, Open in new tab, Open in background tab, Save/Unsave, Share. Open in background tab creates a tab without leaving Search.
6. Search "Eiffel" and long press the tower result: "View on a map" appears and opens Places on it. Long press a result outside the main namespace (e.g. a "Talk:" page): no tab actions.
7. Switch language in the bar. Results reload for the new language; for an RTL language the rows flip.
8. Search for gibberish: "No results found". Turn Wi-Fi off and search: no-internet empty state.
9. Switch themes (Light, Sepia, Dark, Black) with results on screen. Rows, separators and swipe buttons follow the theme.
10. iPad: rows use the readable width like the previous list.
11. Run `SearchResultsMapperTests` in the Wikipedia test plan and `WMFSearchResultsViewModelTests` in WMFComponents.

### Checklist

### Screenshots/Videos
| Before migration | After Migration |
|---|---|
| preview-old | preview-new |
| actions-old | actions-new |
| list-old | list-new |


Wikipedia/Code/SearchResultsViewController+ResultsList.swift
```diff
@@ -0,0 +1,173 @@
...
+import WMFData
+import WMFNativeLocalizations
+import CocoaLumberjackSwift
+
+extension SearchResultsViewController {
+
+    typealias SearchResult = WMFSearchResultsViewModel.SearchResult
+
+    private static let noSearchResultsMessage = WMFLocalizedString("empty-no-search-results-message", value: "No results found", comment: "Shown when there are no search results")
+    private static let viewOnMapActionTitle = WMFLocalizedString("page-location", value: "View on a map", comment: "Label for button used to show an article on the map")
```
tonisevener: This key is declared twice in the codebase, not sure how it'll handle. Can you move it to CommonStrings?

WMFComponents/Sources/WMFComponents/Components/Search/Results/WMFSearchResultsViewModel.swift
```diff
@@ -279,16 +300,22 @@ public final class WMFSearchResultsViewModel: ObservableObject {
...
-        
-        shareAction(result, index, geometryFrames[result.id])
+
+        saveOrUnsaveAction(result, index, source)
+    }
+
+    func share(_ result: SearchResult, source: ActionSource) {
+        guard let index = index(of: result) else { return }
+
+        shareAction(result, index, geometryFrames[result.id], source)
```
tonisevener, line 318: I noticed none of the share sheets (both from the swipe menu and long press) are aligning correctly on iPad. That's what these geometryFrames are supposed to do, I beleive. It seemed like this was partially broken before, where it was only broken on long press, but now both share sites are broken. Just want to note, it can be looked into in a followup (or, we make a separate task and prioritize later).

Old Behavior:
Image

tonisevener (approved):
Looking good! Just a couple of minor things.

WMFComponents/Sources/WMFComponents/Components/Search/Results/WMFSearchResultsViewModel.swift
ealbizzati: nice catch. Fixed the issue for new search results and for History which also had the same bug

Simulator Screenshot - iPad Air 13-inch (M4) - 2026-09-18 at 14 30 31

tonisevener (approved):

PR #6203, opened 2026-09-18
T437359: Add the semantic search entry point to search results

ealbizzati:
**Phabricator:** 

### Notes
- Adds the Dive entry point from T437032 as the first row of the search results list: the reader's query as title, the explanatory line, the Beta badge, an info control, a hide control and the illustration from the Figma node. Same separator as the result rows. Copy is the iOS line recommended in the ticket.
- The entry point renders only when the search language is a target language (`fr`, `ar`, `ja` until the remote `iosv1.semanticSearchLanguages` key exists), the reader is in the experiment group, and it was not hidden. The bucket rolls on the first eligible search. The forced bucket in developer settings no longer bypasses the language gate, so testing needs a search in a target language.
- The copy of the card follows the language selected for the search, not the app language, through `WMFLocalizedString(languageCode:)`. `CommonStrings.betaLabel` became a function with an optional language code for the same reason; its two other callers only gained parentheses.
- "Try it now" shows during the first search session with the entry point and never again, regardless of language. Tapping the card does not remove it; a new session or a language change does. Both the hidden state and the first use persist in user defaults; developer settings got a "Reset entry point" button. The route back from Settings is US-8 (T437046), not this PR.
- Tap and info are wired as closures with no action yet: tap is US-4, info is US-3.
- Accessibility: the card is one VoiceOver element (label = Beta, query, description, Try it now; hint describes the activation) with the info and hide controls as custom actions, like the swipe actions on the result rows. `WMFBetaBadge` became a single accessibility element, which also affects the donation reminder card.
- Two VoiceOver regressions from the SwiftUI migration fixed in the last commit: the language bar now comes before the results list in the reading order, and after the search key dismisses the keyboard the focus moves to the first row (the card or the first result) instead of the row closest to the search field. The second one is driven by `searchBarSearchButtonClicked`, so tapping a result does not trigger it.
- Not covered: the entry point does not show when the lexical search has zero results, because the empty state replaces the list. The illustration has fixed colors, so its white page stays white in Dark and Black; a per-appearance asset can be added if design provides one.

### Test Steps
1. Developer settings: turn on "Enable Semantic Search", set "Force Experiment Group" to Group B, tap "Reset entry point".
2. Search in Français, Arabic or Japanese from the language bar. The card shows above the results with "Try it now". Keep typing: the card stays and the query updates.
3. Clear the field and search again: the card comes back without "Try it now". "Reset entry point" brings it back.
4. Switch to a non target language: the card disappears before the new results load. Switch back: it returns.
5. Tap the X: the card disappears and stays hidden across searches and languages. "Reset entry point" restores it.
6. Search in Arabic with the app in another language: the card mirrors, illustration on the left, and its copy follows the search language once translations exist (English fallback today).
7. VoiceOver: focus the card, hear the label and hint, swipe up for the two actions. Swipe left from the first row: the language bar. Type a term and tap the search key: focus lands on the first row.
8. Run `WMFSemanticSearchDataControllerTests` (WMFData), `WMFSearchResultsViewModelTests` and `WMFSemanticSearchEntryPointViewModelTests` (WMFComponents).

### Checklist

### Screenshots/Videos

Search for Tim cook-2 Search for Tim cook-3


Wikipedia/Code/SearchResultsViewController+ResultsList.swift
```diff
@@ -84,6 +84,75 @@ extension SearchResultsViewController {
         self.searchResultsByArticleURL = searchResultsByArticleURL
 
         resultsViewModel.showResults(results, searchTerm: searchResults.searchTerm, project: mapper.project)
+        if results.isEmpty {
+            resultsViewModel.hideEntryPoint()
```
tonisevener: Minor/Optional - maybe rename to `hideSemanticSearchEntryPoint` / `updateSemanticSearchEntryPoint...`?  It's longer but it would make it more clear to the reader.

tonisevener (approved):

tonisevener (approved):

Wikipedia/Code/SearchResultsViewController+ResultsList.swift
ealbizzati: done. ty

tonisevener (approved):

PR #6204, opened 2026-09-18
T437342: Entry Point P0

l-olson1214:
**Phabricator:** 
https://phabricator.wikimedia.org/T437342

### Notes
Spacing approved by Sarah here (https://wikimedia.slack.com/archives/C0BB50D9TGD/p1789754063960539) 
Saved shows as red, so it's okay for both icons to be
Also adds dev settings for toggling between data rich and low data

### Test Steps
1. 

### Checklist
- [x] too much space above logged-in version across both ipad and iphone

### Screenshots/Videos
image


WMFComponents/Sources/WMFComponents/Components/Developer Settings/WMFDeveloperSettingsViewModel.swift
```diff
@@ -110,6 +110,11 @@ import WMFData
     @objc public init(localizedStrings: WMFDeveloperSettingsLocalizedStrings) {
         self.localizedStrings = localizedStrings
 
+        // Year in Review owns the forced data state, so these two items read and write it through
+        // WMFYearInReviewDataController rather than developer settings. It has no shared instance
+        // and its init throws, so one is held for the lifetime of the sinks below.
+        let yirDataController = try? WMFYearInReviewDataController()
```
mazevedofs, line 116: I don't think we should have this decision point here. We can read the specific toggle somewhere and decide it where the toggle is called.

WMFData/Sources/WMFData/Data Controllers/Year In Review/WMFYearInReviewDataController.swift
```diff
@@ -23,6 +23,29 @@ import CoreData
 
     private let service = WMFDataEnvironment.current.mediaWikiService
     private var dataPopulationBackgroundTaskID: UIBackgroundTaskIdentifier = .invalid
+    
+    /// Which Year in Review experience to force, regardless of how much personalized data the
```
mazevedofs, line 27: Even with the toggle forcing the experience, down the line, if the user does not have enough data, it'd be impossible to test.
I get you're using this now to differentiate the intro slide text, but soon people will try to use this toggle to force an experience, and when it comes back incomplete, that will be seen as a bug and cause issues. 
I'm not telling you to back down from this decision now, but I think all this logic belongs in https://phabricator.wikimedia.org/T438490 when these decisions will be best designed. Honestly, I don't think that the rich/low data requirement should be in this ticket, but I wasn't given time to review it before adding it to the sprint

mazevedofs (changes requested):
I'm requesting changes so we can discuss it before this work is merged.

mazevedofs (approved):
I'm creating a new phab ticket for the outstanding points, thanks!

